PR #27340, opened 2024-02-05
perf(redraw): only redraw Visual area when cursor has moved

### LINKED ISSUE #20235: Invoke cb of `nvim_set_decoration_provider` over and over in visual mode
### Neovim version (nvim -v)

NVIM v0.8.0-dev-1106-g6cc03b9c9

no

### Operating system/version

ArchLinux

### Terminal name/version

Alacritty

alacritty

### Installation

build from repo

`cat mini.lua`
```lua
local api = vim.api
vim.o.cmdheight = 2
_G.cnt = 0
local ns = api.nvim_create_namespace('test')
api.nvim_set_decoration_provider(ns, {
    on_win = function(...)
        vim.schedule(function() end)
        _G.cnt = _G.cnt + 1
    end
})
api.nvim_feedkeys('v', 'n', false)
vim.defer_fn(function()
    local esc = api.nvim_replace_termcodes('<Esc>', true, false, true)
    api.nvim_feedkeys(esc, 'n', false)
    print(_G.cnt)
end, 3000)
```
1. nvim --clean +'so mini.lua'

### Expected behavior

Small enough cnt to output.

### Actual behavior

image

zeertzjq:
Fix #20235

### DIFF
--- a/src/nvim/move.c
+++ b/src/nvim/move.c
@@ -147,6 +147,9 @@ static void redraw_for_cursorcolumn(win_T *wp)
       // When 'cursorcolumn' is set or "CurSearch" is in use
       // need to redraw with UPD_SOME_VALID.
       redraw_later(wp, UPD_SOME_VALID);
+    } else if (VIsual_active) {
+      // In Visual mode need to redraw with UPD_INVERTED.
+      redraw_later(wp, UPD_INVERTED);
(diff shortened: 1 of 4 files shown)


gabrielmfern:
Does this make it better when editing very large files? Seems like something I wanted for a really long time

zeertzjq:
No

Issue #26980, opened 2024-01-10
Repeated `ext_hlstate` and missing ui flush when deleting lines


fredizzimo:
### Problem

When the `ext_hlstate` extension is enabled, and more than around 500 lines are deleted using `indent-blankline.nvim` or`mini.indentscope`, the `highlight_set` event is repeated many times and `flush` is never called.

This causes ui clients like Neovide to stop updating.

The original issue is reported here, see that for more information and log files from Neovide
* https://github.com/neovide/neovide/issues/2217

### Steps to reproduce

1. Create a file `minimal.lua`
```lua
-- Run this file as `nvim --clean -u minimal.lua`

for name, url in pairs{
  -- ADD PLUGINS _NECESSARY_ TO REPRODUCE THE ISSUE, e.g:
  -- some_plugin = 'https://github.com/author/plugin.nvim'
  mini_indentscope = 'https://github.com/echasnovski/mini.indentscope'
} do
  local install_path = vim.fn.fnamemodify('nvim_issue/'..name, ':p')
  if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system { 'git', 'clone', '--depth=1', url, install_path }
  end
  vim.opt.runtimepath:append(install_path)
end

-- ADD INIT.LUA SETTINGS _NECESSARY_ FOR REPRODUCING THE ISSUE
require('mini.indentscope').setup({})
```

2. run `neovide -- --clean -u minimal.lua
3. open a big source file
4. delete more than 500 lines

### Expected behavior

The lines should be deleted normally

### Neovim version (nvim -v)

0.9.5

### Vim (not Nvim) behaves the same?

N/A

### Operating system/version

Linux

### Terminal name/version

N/A

### $TERM environment variable

N/A

### Installation

Pacman

fredizzimo:
Some additional notes
* The `ext_hlstate` extension has to be enabled, without that you can delete as many lines you want
* `hlchunk.nvim` does not cause the same problem
* It was also reported that joining lines can cause the issue.
* A neovide log can be seen here https://pastebin.pl/view/raw/f9f1428f, the relevant parts are at the end before the keypresses.

tim-harding:
I just disabled `ext_hlstate` in Neophyte after encountering the same issue with `indent-blankline`. The logs are attached. I get a few hundred `hl_attr_define` events at line 4990 and then no further events are received and the program appears to hang. 
[log.txt](https://github.com/neovim/neovim/files/13924831/log.txt)

[...]

PR #27348, opened 2024-02-05
fix: crashes with large msgpack messages

fredizzimo:
This fixes the crashing for:
* https://github.com/neovim/neovim/issues/26980

But the other problems of  `ext_hlstate` as described in the issue are still there. Additionally, deleting a lot of lines like that gets very slow quickly, since it sends one message for each line deleted and the size of the message keeps growing.

The fix was done by determining the needed buffer space first, instead of trying to revert the effect of `prepare_call` if message does not fit. The previous code did not revert the full state, which caused corrupted messages to be sent. So, rather than trying to fix all of that, with fragile and hard to read code as a result, the code is now much simpler, although slightly slower.

During testing of this, I also noticed another issue, the messagepack session used in the test were receiving corrupted data, and caused the tests to fail, even before the actual bug I intended to fix was observed. This happened already with `num_lines = 100` in the `does not hang when combining too many highlights`  test. That turned out to be caused by a bug in `lmpack_session_receive`, which did not take the start position into account when determinging the remaining message size.

I also added a couple of extra tests for `ext_hlstate`, which did not seem to be properly tested elsewhere.

### DIFF
--- a/src/nvim/api/ui.c
+++ b/src/nvim/api/ui.c
@@ -585,7 +585,6 @@ static inline int write_cb(void *vdata, const char *buf, size_t len)
 
   data->pack_totlen += len;
   if (!data->temp_buf && UI_BUF_SIZE - BUF_POS(data) < len) {
-    data->buf_overflow = true;
     return 0;
   }
 
@@ -595,12 +594,40 @@ static inline int write_cb(void *vdata, const char *buf, size_t len)
   return 0;
 }
 
-static bool prepare_call(UI *ui, const char *name)
+static inline int size_cb(void *vdata, const char *buf, size_t len)
+{
+  UIData *data = (UIData *)vdata;
+  if (!buf) {
+    return 0;
+  }
+
+  data->pack_totlen += len;
+  return 0;
+}
+
+static void prepare_call(UI *ui, const char *name, size_t size_needed)
 {
   UIData *data = ui->data;
+  size_t name_len = strlen(name);
+  const size_t overhead = name_len + 20;
+  bool oversized_message = size_needed + overhead > UI_BUF_SIZE;
 
-  if (BUF_POS(data) > UI_BUF_SIZE - EVENT_BUF_SIZE) {
+  if (oversized_message || BUF_POS(data) > UI_BUF_SIZE - size_needed - overhead) {
     remote_ui_flush_buf(ui);
+  }
+
+  if (oversized_message) {
+    // TODO(bfredl): manually testable by setting UI_BUF_SIZE to 1024 (mode_info_set)
+    data->temp_buf = xmalloc(20 + name_len + size_needed);
+    data->buf_wptr = data->temp_buf;
+    char **buf = &data->buf_wptr;
+    mpack_array(buf, 3);
+    mpack_uint(buf, 2);
+    mpack_str(buf, S_LEN("redraw"));
+    mpack_array(buf, 1);
+    mpack_array(buf, 2);
+    mpack_str(buf, name, name_len);
+    return;
   }
 
   // To optimize data transfer(especially for "grid_line"), we bundle adjacent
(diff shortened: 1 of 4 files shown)


clason:
`mpack` is a vendored dependency; if this is a bug, please consider PRing the fix upstream (https://github.com/libmpack/libmpack).

fredizzimo:
> `mpack` is a vendored dependency; if this is a bug, please consider PRing the fix upstream

Ok, I will look into that, the repo has not been updated in 6 years though. And there seem to be a lot of commits directly to it here, so I assumed it's unmaintained.

fredizzimo:
The bug is actually in https://github.com/libmpack/libmpack-lua, which has been updated more recently.

clason:
Right, sorry, wrong link!

fredizzimo:
Ok, I will try to create a failing test there first.

fredizzimo:
I created the upstream PR here https://github.com/libmpack/libmpack-lua/pull/33

And also rebased this to make the fix exactly the same.

justinmk (approved):
Thank you, this is great.

fredizzimo:
Another Neovide issue that this fixes is this one
* https://github.com/neovide/neovide/issues/2353

In that case the `ext_hlstate` messages don't get that huge, but the frequency of them probably causes the pipes to fill up, so Neovim starts to build bigger packets internally.

@bfred, @famiu 
I this PR could be changed to something closer to the original implementation if needed. But we need to figure out exactly what needs to be restored, and guarantee that it doesn't break again if the implementation is changed. 

At least `ncalls` and `ncalls_pos` were not restored properly before, but there might be other fields that need to be restored as well. I did try to restore those originally but due to the `libmpack-lua` bug I was never able to get it to work. And by the time that was fixed I had already implemented the current solution and did not feel like going back to the fragile old solution, which needed to keep track of internals across different functions.

src/nvim/api/ui.c
```diff
@@ -615,64 +642,42 @@ static bool prepare_call(UI *ui, const char *name)
...
-  bool new_event = prepare_call(ui, name);
 
   msgpack_packer pac;
   data->pack_totlen = 0;
-  data->buf_overflow = false;
+  // First determine the needed size
+  msgpack_packer_init(&pac, data, size_cb);
+  msgpack_rpc_from_array(args, &pac);
+  // Then send the actual message
+  prepare_call(ui, name, data->pack_totlen);
```
bfredl, line 672: Hmm, this works as a quick fix, but it shouldn't be needed to pack things _twice_ to figure out the buffer sizing. we want to make it safe to just simply flush the buffer at any time and then mark that we are in a state where we cannot add more events (even arglists to the same event name) to the secondary buffer(s) needed to complete the large event.

fredizzimo: Well, the old version did actually pack twice with memcpy  if things did not fit too. This version just determines the size on the first pack and doesn't do any actual packing. So, the overhead is just shifted to be slightly more expensive in the normal case, and less expensive in the worst case. And while I haven't done any performance profiling, the cache misses probably dominate here, and with the cache filled at the second pass it will be much faster than the first pass. I personally care most about the worst-case performance, since that could show up as frame drops in Neovide, while the normal case is always fast enough by a huge margin to not matter at all for any practical purposes.

If anything is going to be changed, I think it should be to remove the allocation of extra buffers and flush on the fly. There's nothing in the RPC protocol that requires us to allocate buffers that big. We only need to know the number of array elements, the length of the strings and so on before writing them and never the complete byte size. The stream can be flushed any time, even in the middle of a string, therefore eliminating the need for additional buffering.

But that requires bigger changes for very little benefit and increased code complexity, unless the profiling data says it's a problem. Your suggestion also probably increases at least the code complexity and development time, which at least I don't have any extra to spare.

Finally, all those micro-optimizations pales in comparison to the actual problem of sending `N(N+1) / 2` `hl_attr_define` events, where N is the number of deleted lines. That can be fixed to only require `N`, but even that's a waste for what I need. And that's why I rather spend my time working on an alternative for our use cases that doesn't need that. https://github.com/neovim/neovim/issues/23610#issuecomment-1929366416

bfredl: > If anything is going to be changed, I think it should be to remove the allocation of extra buffers and flush on the fly. There's nothing in the RPC protocol that requires us to allocate buffers that big. We only need to know the number of array elements, the length of the strings and so on before writing them and never the complete byte size. The stream can be flushed any time, even in the middle of a string, therefore eliminating the need for additional buffering.

Sure it is more something this I had in mind. I'd be happy to look at this myself, as I don't think this requires much extra complexity at all: we just need to represent the state where the "ncalls" value already has been flushed away, and thus it cannot be changed (and thus we need to start on a new `"redraw"` event for the next event, even with the same event name). Optimization should just be generally tuned to use medium sized buffers (and batching is then invoked when we can pack really small events into the _lagom_ sized buffer)

That being said, I'm fine with merging this as is as a stop-gap as-is, as it might take some time until I get into look at that and not crashing is of course priority one (especially with good testing, as you add).

PR #27349, opened 2024-02-05
feat(lua): highlight.on_yank use win-local highlight

altermo:
Currently, `highlight.on_yank` will do buffer-local highlighting, this PR will make it window scoped.

Some additional changes
* fix: when yanking in the buffer, moving to another buffer, and yanking before the original buffer highlight disappears, the original buffer highlight won't disappear on timeout.

### DIFF
--- a/runtime/lua/vim/highlight.lua
+++ b/runtime/lua/vim/highlight.lua
@@ -55,6 +55,7 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
   local regtype = opts.regtype or 'v'
   local inclusive = opts.inclusive or false
   local priority = opts.priority or M.priorities.user
+  local scoped = opts._scoped or false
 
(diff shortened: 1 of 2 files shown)


clason:
I have to admit I'm not too fond of adding a `matchadd` hack and would rather wait until we have window-local extmarks (if ever).

The other changes are welcome, though.

runtime/lua/vim/highlight.lua
```diff
@@ -76,8 +76,36 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
   end
 end
 
+local function scoped_range(bufnr, ns, higroup, start, finish, opts)
```
zeertzjq, line 79: Maybe it's better to adding scope support to `vim.highlight.range()` instead of duplicating the code?

clason (approved):

Issue #27310, opened 2024-02-02
Suggestion: Explain linux installation in more detail in INSTALL.md


JamesC01:
### Problem

I'm really not sure where to put this, since it isn't a bug, but it also isn't a feature. Right now I feel like the INSTALL.md file doesn't explain enough about installing on linux. It doesn't really explain much outside of using the appimage, and running it by typing "./nvim.appimage", and installing it from tarball, and typing "./nvim-linux64/bin/nvim" when you want to run it.
This is not what most users want. The top of the file even mentions that to run neovim, you type "nvim." But if a user follows the installation instructions for tarball or appimage, this just won't work correctly.

When someone follows an installation guide, it should give them all they need to get the program functioning like usual, but at the moment, it leaves off a few steps needed to get the program running by typing "nvim". I've found some workarounds myself, but it's not easy to find out if you're doing something the "right" way (that's why I'm not changing INSTALL.md myself, I'm just not sure of the proper conventional way of doing this), so if it was included in the install instructions, I think it would be very helpful.

It does explain how to expose the _extracted_ appimage globally, but that is noted as an "optional" thing, and it makes it seem like you should only do this if you _can't_ already run the appimage directly.

### Expected behavior

It should explain the exact commands to install the program from tarball or appimage, and get it to run in the expected way, by typing "nvim" anywhere, instead of having to go into the folder, or typing a big long path every time you want to use the program. It should mention how to do this in the conventional way, since there's so many different ways of doing this, just going with the most typical option would be useful (I'm not sure of it.) After a quick search, it seems one convetional way is to put the main folder in /usr/opt and create a symlink from the nvim bin folder to allow starting it with just the command, but I think someone who knows better than me should write it. Maybe for debian systems, converting it to a .deb and using dpkg might be possible?

justinmk:
If #27350 isn't enough then this is "wontfix".

justinmk:
> When someone follows an installation guide, it should give them all they need to get the program functioning like usual, but at the moment, it leaves off a few steps needed to get the program running by typing "nvim".

Mainly because the binaries were never intended, and still aren't intended, and never will be intended, to be the primary install option. They are merely best-effort escape hatches. The primary, recommended way to install is via your system package manager or by building from source.

justinmk:
#27350 may or may not be merged but I'm closing this anyway, for the above reason.

PR #27350, opened 2024-02-05
docs(install): add more specific Linux install instructions

PurityLake:
Refers to #27310 

Adds more specific instructions on how to install neovim for AppImage and pre-built archives.

If I am missing any information or there are parts I should elaborate more on please let me know and I will update the file,

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,6 +93,23 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides pre-buil
 
 ## Linux
 
+### Pre-built archives
+
+The [Releases](https://github.com/neovim/neovim/releases) page provides pre-built binaries for Linux systems. There are a few ways to install this, the way listed below can be changed to a different directory than the one used here (e.g. `$HOME/.local/bin ` if you do not have root privileges)
+
+```sh
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
+# this is to clean up the files if you have previously installed it here before
+sudo rm -rf /opt/nvim
+sudo tar -C /opt -xzf nvim-linux64.tar.gz
+```
+
+After this you will need to tell your terminal where to find the `nvim` executable. Add the following line to your `~/.bashrc`:
+
+    export PATH="$PATH:/opt/nvim-linux64.tar.gz"
+
+Restart your terminal or run `source ~/.bashrc` and you can then run `nvim`
+
 ### AppImage ("universal" Linux package)
 
 The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppImage](https://appimage.org) that runs on most Linux systems. No installation is needed, just download `nvim.appimage` and run it. (It might not work if your Linux distribution is more than 4 years old.)
@@ -111,7 +128,19 @@ sudo mv squashfs-root /
 sudo ln -s /squashfs-root/AppRun /usr/bin/nvim
 nvim
 ```
- 
+
+If you are able to run the AppImage without extracting the files, a common way to install and run AppImage files is to do the following:
+
+    curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim.appimage
+    chmod u+x nvim.appimage
+    mkdir -p $HOME/.local/bin
+    mv nvim.appimage $HOME/.local/bin
+    ln -s $HOME/.local/bin/nvim.appimage $HOME/.local/bin/nvim
+    nvim
+
+The directory `$HOME/.local/bin` is a place where a user can place user specific or private binaries.
+This can be useful if you do not have root privileges with your current account. If your terminal can't find the executable straight away close and reopen it.
+
 ### Arch Linux
 
 Neovim can be installed from the community repository:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -117,11 +117,12 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppI
 
 To expose nvim globally:
 
-    mkdir -p $HOME/.local/bin
-    mv nvim.appimage $HOME/.local/bin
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -93,6 +93,23 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides pre-buil
...
+```sh
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
+# this is to clean up the files if you have previously installed it here before
+sudo rm -rf /opt/nvim
+sudo tar -C /opt -xzf nvim-linux64.tar.gz
+```
+
+After this you will need to tell your terminal where to find the `nvim` executable. Add the following line to your `~/.bashrc`:
+
+    export PATH="$PATH:/opt/nvim-linux64.tar.gz"
```
mike325, line 109: You are not appending the correct path of the uncompressed tar, it should not include the `.tar.gz` extension

PurityLake: Ah yes, you're right, dunno how I missed that. I'll fix that, thank you

clason:
This is way too chatty. This is a reference documentation, not a user manual. It's better to keep the text as simple and straightforward as possible, also with non-English speakers in mind.

PurityLake:
Ok fair point, I'll rewrite it with that in mind

INSTALL.md
```diff
@@ -93,6 +93,23 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides pre-buil
 
 ## Linux
 
+### Pre-built archives
+
+The [Releases](https://github.com/neovim/neovim/releases) page provides pre-built binaries for Linux systems. There are a few ways to install this, the way listed below can be changed to a different directory than the one used here (e.g. `$HOME/.local/bin ` if you do not have root privileges)
```
agarie, line 98:
```suggestion
The [Releases](https://github.com/neovim/neovim/releases) page provides pre-built binaries for Linux systems.
```

I think this information can be included in comments in the next code block :)

INSTALL.md
```diff
@@ -93,6 +93,23 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides pre-buil
...
+
+```sh
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
+# this is to clean up the files if you have previously installed it here before
+sudo rm -rf /opt/nvim
```
agarie, line 103:
```suggestion
# You can install in another directory instead of /opt, e.g. ~/.local/bin
sudo rm -rf /opt/nvim # remove previous installation
```

PurityLake:
@clason  I have altered the wording, I made it more terse but kept much of the same commands. I trimmed and rearranged the AppImage instructions to follow on from the previous set of commands so `curl` isn't repeated.

INSTALL.md
```diff
@@ -101,6 +115,14 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppI
     chmod u+x nvim.appimage
     ./nvim.appimage
 
+To expose nvim globally:
+
+    mkdir -p $HOME/.local/bin
+    mv nvim.appimage $HOME/.local/bin
+    ln -s $HOME/.local/bin/nvim.appimage $HOME/.local/bin/nvim
+    nvim
```
justinmk, line 123: Not sure why we're telling people how to use `mv` and `ln -s`. The docs need to stay focused.

The `/opt/` suggestion for the other case in this PR is ok I guess, it saves some decisions for the user. But this section needs to be deleted, also because it distracts from the "squashfs" followup below.

JamesC01: I saw `ln` being used when I searched what the typical ways of installing programs without a package manager were. If it should be used, then I think it's useful to include the commands, since the `nvim` command should be exposed globally (for most people.) And it's nice to list a working command for people who want to get up and running.

justinmk: > Could probably just remove the ln command, as using mv to remove the .appimage would be fine

yeah that seems reasonable and avoids some steps.

INSTALL.md
```diff
@@ -93,6 +93,20 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides pre-buil
 
 ## Linux
 
+### Pre-built archives
+
+The [Releases](https://github.com/neovim/neovim/releases) page provides pre-built binaries for Linux systems.
```
justinmk, line 98: What are the limitations of the tarball? surely it runs on fewer systems than the appimage. Not sure why that isn't specified in the releases page.

By the way, this PR duplicates the notes already on the releases page.

PurityLake: Guess it'd depend on what system the tarballs are generated on, is that done via CI?

justinmk: oh I remember now, the tarbar is just a copy of the files from the appimage. sigh.

INSTALL.md
```diff
@@ -93,6 +93,20 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides pre-buil
...
+
+```sh
+curl -LO https://github.com/neovim/neovim/releases/latest/download/nvim-linux64.tar.gz
+sudo rm -rf /opt/nvim
+sudo tar -C /opt -xzf nvim-linux64.tar.gz
```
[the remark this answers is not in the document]
PurityLake: Would you like me to make another PR to address these concerns? Or can this PR be reused?

INSTALL.md
```diff
@@ -101,6 +115,15 @@ The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppI
     chmod u+x nvim.appimage
     ./nvim.appimage
 
+To expose nvim globally:
+
+    mkdir -p /opt/nvim
+    mv nvim.appimage /opt/nvim/nvim
+
+And the following line to `~/.bashrc`:
```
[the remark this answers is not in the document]
justinmk: Providing steps for every shell is not in scope. Users need to understand how to use their shell.

PR #23428, opened 2023-05-01
fix(tui): grid_clear properly clears the screen

luukvbaal:
Problem:    When setting a shell size smaller than the containing
            terminal window through `:winsize` or `:set lines/columns`
            the screen is not properly cleared.
Solution:   Clear the tui dimensions rather than the grid dimensions.

PR #27352, opened 2024-02-05
fix(tui): `space_buf` overflow when clearing screen

seandewar:
Problem: `tui->space_buf` may be smaller than the width of the TUI or widest grid, causing an overflow when calling `tui_grid_clear` (which also clears the TUI's screen after #23428).

Solution: resize `space_buf` to be wide enough to fit the TUI and widest grid.

Didn't bother shrinking the allocation if the max widths decrease.

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -137,6 +137,7 @@ struct TUIData {
     int sync;
   } unibi_ext;
   char *space_buf;
+  size_t space_buf_len;
   bool stopped;
   int seen_error_exit;
   int width;
@@ -1055,10 +1056,7 @@ void tui_grid_resize(TUIData *tui, Integer g, Integer width, Integer height)
 {
   UGrid *grid = &tui->grid;
   ugrid_resize(grid, (int)width, (int)height);
-
-  xfree(tui->space_buf);
-  tui->space_buf = xmalloc((size_t)width * sizeof(*tui->space_buf));
-  memset(tui->space_buf, ' ', (size_t)width);
+  ensure_space_buf_size(tui, (size_t)width);
 
   // resize might not always be followed by a clear before flush
   // so clip the invalid region
@@ -1642,6 +1640,15 @@ static void invalidate(TUIData *tui, int top, int bot, int left, int right)
   }
 }
 
+static void ensure_space_buf_size(TUIData *tui, size_t len)
+{
+  if (len > tui->space_buf_len) {
+    tui->space_buf = xrealloc(tui->space_buf, len * sizeof *tui->space_buf);
+    memset(tui->space_buf + tui->space_buf_len, ' ', len - tui->space_buf_len);
+    tui->space_buf_len = len;
+  }
+}
+
 /// Tries to get the user's wanted dimensions (columns and rows) for the entire
 /// application (i.e., the host terminal).
 void tui_guess_size(TUIData *tui)
@@ -1678,6 +1685,7 @@ void tui_guess_size(TUIData *tui)
 
   tui->width = width;
   tui->height = height;
+  ensure_space_buf_size(tui, (size_t)tui->width);
 
   // Redraw on SIGWINCH event if size didn't change. #23411
   ui_client_set_size(width, height);
--- a/test/functional/terminal/tui_spec.lua
+++ b/test/functional/terminal/tui_spec.lua
@@ -2095,6 +2095,48 @@ describe('TUI', function()
     ]],
     }
   end)
+
+  it('no heap-buffer-overflow when changing &columns', function()
+    -- Set a different bg colour and change $TERM to something dumber so the `print_spaces()`
+    -- codepath in `clear_region()` is hit.
+    local screen = thelpers.setup_child_nvim({
+      '-u',
+      'NONE',
+      '-i',
+      'NONE',
+      '--cmd',
+      'set notermguicolors | highlight Normal ctermbg=red',
+      '--cmd',
+      'call setline(1, ["a"->repeat(&columns)])',
+    }, { env = { TERM = 'ansi' } })
+
+    screen:expect {
+      grid = [[
+      aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa|
+      ~                                                 |*3
+      [No Name] [+]                   1,1            All|
+                                                        |
+      -- TERMINAL --                                    |
+    ]],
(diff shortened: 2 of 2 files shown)


test/functional/terminal/tui_spec.lua
```diff
@@ -1919,6 +1919,39 @@ describe('TUI', function()
...
     end)
   end)
+
+  it('no heap-buffer-overflow when changing &columns', function()
+    -- Ensure a bg colour is set to trigger `print_spaces()` codepath in `clear_region()`.
+    child_session:request('nvim_command', 'highlight Normal ctermbg=red guibg=red')
+
+    feed_data(":call setline(1,['a'->repeat(&columns)])\n")
+    screen:expect {
+      grid = [[
```
zeertzjq, line 1929:
```suggestion
    feed_data(":call setline(1, 'a'->repeat(&columns))\n")
    screen:expect([[
```

test/functional/terminal/tui_spec.lua
```diff
@@ -1919,6 +1919,39 @@ describe('TUI', function()
...
+      aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa|
+      ~                                                 |*3
+      [No Name] [+]                                     |
+      :call setline(1,['a'->repeat(&columns)])          |
+      -- TERMINAL --                                    |
+    ]],
+      attr_ids = {},
+    }
```
zeertzjq, line 1937:
```suggestion
      :call setline(1, 'a'->repeat(&columns))           |
      -- TERMINAL --                                    |
    ]])
```

test/functional/terminal/tui_spec.lua
```diff
@@ -1919,6 +1919,39 @@ describe('TUI', function()
...
+
+    feed_data(':set columns=12\n')
+    screen:expect {
+      grid = [[
```
zeertzjq, line 1941:
```suggestion
    screen:expect([[
```

test/functional/terminal/tui_spec.lua
```diff
@@ -1919,6 +1919,39 @@ describe('TUI', function()
...
+      aaaaaaaaaaaa                                      |*4
+      < Name] [+]                                       |
+                                                        |
+      -- TERMINAL --                                    |
+    ]],
+      attr_ids = {},
+    }
```
zeertzjq, line 1948:
```suggestion
    ]])
```

test/functional/terminal/tui_spec.lua
zeertzjq: Or maybe `child_session:request()` can be used instead of `feed_data()`?

test/functional/terminal/tui_spec.lua
```diff
@@ -1919,6 +1919,39 @@ describe('TUI', function()
       eq({ 'https://example.com', '' }, exec_lua([[return _G.urls]]))
     end)
   end)
+
+  it('no heap-buffer-overflow when changing &columns', function()
+    -- Ensure a bg colour is set to trigger `print_spaces()` codepath in `clear_region()`.
+    child_session:request('nvim_command', 'highlight Normal ctermbg=red guibg=red')
+
+    feed_data(":call setline(1,['a'->repeat(&columns)])\n")
```
zeertzjq, line 1927: Maybe `child_session:request('nvim_command', ...)` can also be used instead of `feed_data()` here?

src/nvim/tui/tui.c
```diff
@@ -137,6 +137,7 @@ struct TUIData {
     int sync;
   } unibi_ext;
   char *space_buf;
+  int space_buf_len;
```
zeertzjq, line 140: I think this can be `size_t` to avoid casts.

src/nvim/tui/tui.c
```diff
@@ -1642,6 +1640,15 @@ static void invalidate(TUIData *tui, int top, int bot, int left, int right)
   }
 }
 
+static void ensure_space_buf_size(TUIData *tui, int len)
```
zeertzjq, line 1643:
```suggestion
static void ensure_space_buf_size(TUIData *tui, size_t len)
```

src/nvim/tui/tui.c
```diff
@@ -1642,6 +1640,15 @@ static void invalidate(TUIData *tui, int top, int bot, int left, int right)
...
+{
+  if (len > tui->space_buf_len) {
+    tui->space_buf = xrealloc(tui->space_buf, (size_t)len * sizeof *tui->space_buf);
+    memset(tui->space_buf, ' ', (size_t)len);
```
zeertzjq, line 1647: The `memset()` doesn't use `sizeof`, so I think the `xrealloc()` shouldn't use `sizeof` either.
```suggestion
    tui->space_buf = xrealloc(tui->space_buf, len);
    memset(tui->space_buf, ' ', len);
```

src/nvim/tui/tui.c
```diff
@@ -1055,10 +1056,7 @@ void tui_grid_resize(TUIData *tui, Integer g, Integer width, Integer height)
 {
   UGrid *grid = &tui->grid;
   ugrid_resize(grid, (int)width, (int)height);
-
-  xfree(tui->space_buf);
-  tui->space_buf = xmalloc((size_t)width * sizeof(*tui->space_buf));
-  memset(tui->space_buf, ' ', (size_t)width);
+  ensure_space_buf_size(tui, (int)width);
```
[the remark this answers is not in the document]
seandewar: `Integer` is `int64_t`, so the cast'll be needed for signedness I suppose

seandewar:
Ugh, GitHub was showing that comment as duplicated, but apparently it wasn't... :facepalm:

zeertzjq (commented):
It seems that the `print_spaces()` branch is never reached inside Nvim's `:terminal` (as it has `back_color_erase`), so the test doesn't fail without the fix. Maybe it needs a different `TERM` variable (e.g. `ansi`) to reach the branch?

seandewar:
That's unfortunate, I was afraid that might happen :cry:.

It seems to fail for me locally even without the ASAN enabled, as the screen state differs, but also fails with ASAN with `ASAN_OPTIONS=log_path:...` and `LOG_DIR=...` set. This might be due to my particular environment, though...

zeertzjq (approved):

seandewar:
OK, I think the `$TERM` debacle has been resolved, which seems to cause the test to fail on the CI now. :innocent:

PR #26813, opened 2023-12-31
perf: make screen size and position calculations more efficient

vanaigr:
NOTE: all referenced commits have been shifted by one, first commit now contains benchmarks.

### 1. Breakindent and strcmp()
`win_lbr_chartabsize()` recomputes `breakindent` of a line up to 2 times in order to determine the size of the character. [The function for computing breakindent](https://github.com/neovim/neovim/blob/0a598c13b1863ba1aff378027c4376e3ab7048ee/src/nvim/indent.c#L785) caches the last result, but as part of checking if the cached value could be used it strcmp's the given line to the cached (copied) line. Since `win_lbr_chartabsize()` is often called in a loop and none of the parameters that affect `breakindent` change, this results in strcmp being called for 2 identical lines (worst case for strcmp) for every character in the line starting from the first wrapped row (even though the value is almost never affects the final size of that character).

The first commit fixes the ussue of repeatedly comparing the lines by caching the indentation value in the `chartabsize_T` struct. I also decided to preserve the size of the structure by removing `cts_has_virt_text` and changing `cts_row` to `virt_row` since it is used only for virtual text and can be set to negative value if the line does not contain virtual text.

### 2. Unnecessary loop
`getvcol()` computes virtual column until the last position by iterating until the current position reaches the first byte of the character containing the last position. To compute the first byte of the character, it uses an extra loop at the start to find if the position is past the end of the line and sets it to the end of the line before computing the index of its first byte.

The second commit substitutes comparison with the first byte of the last position with checking if next position is greater than the last position, which allows removing the loop at the start.

### 3. Non-ascii characters are decoded multiple times.
[...]

PR #18690, opened 2022-05-22
fix(charset): don't consider chars above 0x10FFFF printable

bfredl:
attempt at fixes #18689

PR #27353, opened 2024-02-05
perf: improve utf_char2cells() performance

vanaigr:
`utf_char2cells()` calls `utf_printable()` twice (sometimes indirectly, through `vim_isprintc()`) for characters >= 128. The function can be refactored to call to it only once.

`utf_printable()` uses binary search on ranges of unprintable characters to determine if a given character is printable. Since there are only 9 ranges, and the first range contains only one character, binary search can be replaced with SSE2 SIMD comparisons that check 8 ranges at a time, and the first range is checked separately.
SSE2 is enabled by default in GCC, Clang and MSVC for x86-64.

I tested the function and it returns the same results for all 2^32 values.

#### Benchmarks:

I modified the benchmarks from #26813 to better reflect the performance difference and added a new benchmark that measures 3 byte utf-8 separately.

||ascii|2 byte|3 byte|random|
|-|-|-|-|-|
|before|14|85|194|170|
|after|13|83|172|154|


src/nvim/mbyte.c
```diff
@@ -1092,9 +1090,52 @@ bool utf_iscomposing(int c)
...
+
+  __m128i value = _mm_set1_epi16((int16_t)c);
+
+  // using _mm_cmplt_epi16() is less optimal, since it would require
+  // swapping operands (sse2 only has cmpgt instruction),
+  // and only the second operand can be a memory location
+
+  // character is printable when c is above/below both bounds
+  return _mm_movemask_epi8(_mm_cmpgt_epi16(value, hi))
+          == _mm_movemask_epi8(_mm_cmpgt_epi16(value, lo));
```
zeertzjq, line 1130:
```suggestion
         == _mm_movemask_epi8(_mm_cmpgt_epi16(value, lo));
```

zeertzjq:
You may also want to take #18690 into consideration.

VanaIgr:
Should I make characters above 10FFFF unprintable?

zeertzjq:
Hmm, actually it's not hard to add that check later, so not really necessary.

src/nvim/mbyte.c
```diff
@@ -1092,9 +1090,52 @@ bool utf_iscomposing(int c)
...
+    return false;
+  }
+  if (c < 0x180b || c > 0xffff) {
+    return true;
+  }
+
+# define L(v) ((int16_t)((v) - 1))  // lower bound (exclusive)
+# define H(v) ((int16_t)(v))  // upper bound (inclusive)
+
+  // bundaries of unprintable characters
```
zeertzjq, line 1112:
```suggestion
  // boundaries of unprintable characters
```

src/nvim/mbyte.c
```diff
@@ -1092,9 +1090,52 @@ bool utf_iscomposing(int c)
...
+// Only for characters of 0x100 and above!
+bool utf_printable(int c)
+  FUNC_ATTR_CONST
+{
+  if (c == 0x70f) {
+    return false;
+  }
+  if (c < 0x180b || c > 0xffff) {
+    return true;
+  }
```
zeertzjq, line 1107: Will this be faster as smaller codepoints are more common?
```suggestion
  if (c < 0x180b || c > 0xffff) {
    return true;
  }
  if (c == 0x70f) {
    return false;
  }
```
Also codepoints above `0xffff` are not that common, so maybe removing the `c > 0xffff` condition here (as the SIMD comparisons below also cover it) will be faster?

src/nvim/mbyte.c
zeertzjq, line 1112:
```suggestion
  // boundaries of unprintable characters
```

src/nvim/mbyte.c
zeertzjq, line 1107: This seems to be faster as the result of the `c != 0x070f` comparison is used directly as the return value:
```suggestion
  if (c < 0x180b || c > 0xffff) {
    return c != 0x070f;
  }
```
Also, codepoints above `0xffff` should be less common than codepoints below `0xffff`, so maybe it will be faster to remove this `c > 0xffff` check and instead let the SIMD comparisons below cover it?
```suggestion
  if (c < 0x180b) {
    return c != 0x070f;
  }
```

VanaIgr: I used 16 bit comparisons to fit 8 values in __m128i, so the value of `c` needs to be between 0 and FFFF.

zeertzjq: Oh right.

src/nvim/mbyte.c
```diff
@@ -1092,9 +1090,52 @@ bool utf_iscomposing(int c)
...
+{
+  if (c < 0x180B || c > 0xFFFF) {
+    return c != 0x70F;
+  }
+
+# define L(v) ((int16_t)((v) - 1))  // lower bound (exclusive)
+# define H(v) ((int16_t)(v))  // upper bound (inclusive)
+
+  // Boundaries of unprintable characters.
+  // Some values are negative when converted to int16_t.
```
zeertzjq, line 1110: Trailing space
```suggestion
  // Some values are negative when converted to int16_t.
```

ii14 (approved):

marcelbeumer:
Thanks so much everyone!

PR #27358, opened 2024-02-06
fix(event-loop): process input before events

zeertzjq:
Problem:
When `nvim_input` is followed immediately by non-fast events on RPC, both
events and input are available after the polling done by the `os_inchar()`
in `state_enter()`, but `state_enter()` then chooses to process events even
if input is available, which is inconsistent with `state_handle_k_event()`
that stops processing events once input is available.

Solution:
Also check for available input after the `os_inchar()` in `state_enter()`.


zeertzjq:
This likely fixes the root caused of the race between `nvim_input` and `nvim_paste` as mentioned in #27356.

PR #27362, opened 2024-02-06
ci(release): bump mac runner version to macos-12

dundargoc:


clason (approved):
Why not `-latest`?

dundargoc:
Just following our MAINTAIN policy of explicitly using the oldest non-deprecated version. `-latest` might change version before macos-12 is deprecated.

In practice I don't think it matters tbh, I don't see a real risk of using `-latest`. Shall I use that instead?

clason:
Nah, just wondering (and good to have it made explicit).

dundargoc:
> I think we can, however, write in `news.txt` that the minimum version has been raised.

Eh, sure. If you wanna make a PR go for it.

> Because of the same reason, running CI on latest macOS may not be a good idea because it would result in having too high macOS version (14.0).

Ummm, what? Too high for what exactly? Are you talking about release CI or test CI?

wookayin:
Sorry for the confusion. I meant release CI because this PR was about bumping versions for release CI.

Actually the "macos-latest" runner points to macOS 12 as of today (not the actual *latest* macOS), not 14, but this could change if we didn't explicit. There are macos-latest-large (macOS 13) and macos-14 runners btw; my main point was either we explicitly configure the minimum target version, or have it implicitly managed by pinning the macos runner version.

dundargoc:
Oh, you thought `latest` meant latest. I guess that's a fair assumption to make if you're not used to the label shenanigans github pulls of from time to time. Both me and clason knew latest meant 12, which is why I was so confused when you brought up 14 lol.

> but this could change if we didn't explicit

Eh, I know, I mentioned the exact point lol. Why I said it doesn't matter much is that these runner versions require regular oversight and change, regardles if they're set to latest or to 12. But yes, 12 is theoretically more correct which is why I set it to that.

clason:
> but this could change

This _will_ change in Q2.

Issue #25696, opened 2023-10-18
Snippet API roadmap


MariaSolOs:
### Problem

Foo

### Expected behavior

Bar

telemachus:
Sorry if this is not the right place, but should the current version of `vim.snippet.expand` support snippets like `${1:${TM_FILENAME_BASE:placeholder}}`? Looking at the roadmap here, I think that the answer is yes since "nested placeholders" is marked as completed. But I don't think that snippets like that work currently. (In experiments, if I use `${1:${TM_FILENAME_BASE:placeholder}}`, the tabstop is empty rather than filled with either `TM_FILENAME_BASE` or `placeholder`. Things like `${1:${TM_FILENAME_BASE}}` and `${1:$TM_FILENAME_BASE}` also leave the tabstop empty.)

MariaSolOs:
@telemachus that sounds like a bug. Could you please open a separate issue with repro steps and minimal setup?

juru1234:
@MariaSolOs I'm using the native snippet support without any additional plugins and I really like it. But jumping from one snippet part to the next with Tab is kinda slow because the cursor always starts again at the beginning of the line. Is this a known problem and will this be improved in future releases?

PR #25301, opened 2023-09-22
feat(lsp): snippet API

MariaSolOs:
Defines the minimal API for LSP snippet support. The initial design is:
1. `vim.snippet.expand` receives the snippet text and parses it using the LPEG grammar.
2. The snippet's children are transformed as follows:
    - Known variables are replaced by their values (e.g. `$TM_FILENAME` becomes `foo.txt`), falling back to their defaults when provided.
    - When a variable is unknown, the name of the variable is inserted and it is transformed into a placeholder.
    - Placeholders have their text[^1] inserted.
    - With choice nodes, the first value is inserted (see below for future work).
    - Text nodes are inserted as given.
3. Based on `expandtab` and the indentation of the current line, the snippet is indented accordingly. 
4. The snippet text is set in the current buffer.
5. To "fill out" the snippet, tabstops can be navigated using `vim.snippet.jump(direction)` (using `1` for jumping forward and `-1` for backwards). `vim.snippet.jumpable(direction)` can also be used before jumping to check if the jump in the given direction is valid.
6. The snippet session is cancelled when reaching the final tabstop (`$0`) or when the cursor moves outside the snippet's tabstops. Tabstops can be customized with the `SnippetTabstop` highlight group.

Note that the following features remain yet to be implemented in more mature versions of the snippet API:
- Variable transformations via regexes.
- Choice placeholders, in which the user can select from a given set of options to insert.
- Nested placeholders. Right now the value of a placeholder is simply converted to a string and inserted.

Pending discussion items in this PR:
- [X] _My use of `Range4` despite it being a "Treesitter" type_: Seems like we're fine with this given that [we plan on refactoring ranges/positions anyway](https://github.com/neovim/neovim/issues/25509).
- [x] _Currently `vim.snippet.jump` uses `vim.snippet.jumpable` to check that the movement is valid. Do we want to instead ask the user to invoke `jumpable` first and remove the check?_ No harm in keeping the check.
- [x] _I'm not 100% sure about the name of `vim.snippet.cancel`, since sometimes it does cancel the snippet session but it's also invoked when the snippet navigation is complete. Do we want to give it another name?_ Looks like we're all happy with `vim.snippet.exit`.

[^1]: When I say that the text of a node is inserted, I don't mean the raw table string, but the output from [`Node:__tostring()`](https://github.com/neovim/neovim/blob/a041d8ac118f45b9e20d4975737f25a548cb6337/runtime/lua/vim/lsp/_snippet_grammar.lua#L85).

PR #27375, opened 2024-02-07
Allow template variable values to be passed to `vim.snippet.expand()`

refractalize:
This is an extension to the existing `vim.snippet.expand()` api introduced in #25301 (see also: #25696)

The idea with this PR is that the caller can specify the values of variables defined in the templates by passing them in a second `options` argument to `vim.snippet.expand()`:

```lua
vim.snippet.expand(snippet, {
  variables: {
    A = "value for $A",
    B = function(name)
      return "value for $" .. name
    end
  }
})
```

The logic is as follows:
* If the variable is not defined, the variable's default (as defined in the template) is used, falling back to an empty string. **NOTE** This is different to the original behaviour, where an undefined variable would become a tabstop. This seemed awkward, and would override other existing tabstops of the same index. For e.g. `$A` would clobber `$1` if it came first in the template.
* If the variable is a function, the function is called with the variable's name as the first argument. The result is used as the variable's value. If the function returns `nil`, then the variable's default is used, falling back to an empty string.
* If the variable is a string, the string is used as the variable's value, or if `nil` falling back to the variable's default, and then to an empty string.

One usecase for this is to implement `TM_SELECTED_TEXT` in userland, the simplest approach being something like:

```lua
vim.snippet.expand(snippet, {
  variables: {
    TM_SELECTED_TEXT = function(name)
      return vim.fn.getreg('"')
    end
  }
})
```

**NOTE**: The use of`vim.fn.getreg('"')` here is pretty simplistic and has some odd edge cases, but it's good to use as an example here.

- [ ] test coverage

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -4245,7 +4245,7 @@ vim.snippet.active()                                    *vim.snippet.active()*
 vim.snippet.exit()                                        *vim.snippet.exit()*
     Exits the current snippet.
 
-vim.snippet.expand({input})                             *vim.snippet.expand()*
+vim.snippet.expand({input}, {opts})                     *vim.snippet.expand()*
     Expands the given snippet text. Refer to https://microsoft.github.io/language-server-protocol/specification/#snippet_syntax for
     the specification of valid input.
 
@@ -4253,6 +4253,14 @@ vim.snippet.expand({input})                             *vim.snippet.expand()*
 
     Parameters: ~
       • {input}  (`string`)
+      • {opts}   (`table?`) Options table with keys:
+                 • `variables` (`table`): table containing values for the
+                   template variables. Each of the entries can be either a
+                   string, or a function that returns a string. In either case
+                   if the result is `nil` then the template variables's
+                   default will be used, or an empty string. The function
+                   recieves two arguments, the name of the variable and the
+                   default value from the template.
 
 vim.snippet.jump({direction})                             *vim.snippet.jump()*
     Jumps within the active snippet in the given direction. If the jump isn't
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -10,47 +10,27 @@ local function cursor_pos()
   return cursor[1] - 1, cursor[2]
 end
 
---- Resolves variables (like `$name` or `${name:default}`) as follows:
---- - When a variable is unknown (i.e.: its name is not recognized in any of the cases below), return `nil`.
---- - When a variable isn't set, return its default (if any) or an empty string.
+--- Resolves variables (like `$name` or `${name:default}`) by looking up the value of the variable
+--- in the `variables` options table.
+---
+--- - When the variable's value is a function, the function's return value is the variable's value.
+--- - When the variable's value is a string, the string is the variable's value.
+--- - In either case, if the result is still `nil`, use the default value.
 ---
 --- Note that in some cases, the default is ignored since it's not clear how to distinguish an empty
 --- value from an unset value (e.g.: `TM_CURRENT_LINE`).
 ---
 --- @param var string
 --- @param default string
---- @return string?
-local function resolve_variable(var, default)
-  --- @param str string
-  --- @return string
-  local function expand_or_default(str)
-    local expansion = vim.fn.expand(str) --[[@as string]]
-    return expansion == '' and default or expansion
-  end
+--- @return string
+local function resolve_variable(var, default, variables)
+  local variable_value = variables[var] --[[@type (string?)|fun(string): string?]]
 
-  if var == 'TM_SELECTED_TEXT' then
-    -- Snippets are expanded in insert mode only, so there's no selection.
-    return default
-  elseif var == 'TM_CURRENT_LINE' then
-    return vim.api.nvim_get_current_line()
-  elseif var == 'TM_CURRENT_WORD' then
-    return expand_or_default('<cword>')
-  elseif var == 'TM_LINE_INDEX' then
-    return tostring(vim.fn.line('.') - 1)
-  elseif var == 'TM_LINE_NUMBER' then
-    return tostring(vim.fn.line('.'))
-  elseif var == 'TM_FILENAME' then
-    return expand_or_default('%:t')
-  elseif var == 'TM_FILENAME_BASE' then
-    return expand_or_default('%:t:r')
-  elseif var == 'TM_DIRECTORY' then
-    return expand_or_default('%:p:h:t')
-  elseif var == 'TM_FILEPATH' then
-    return expand_or_default('%:p')
+  if type(variable_value) == 'function' then
+    return variable_value(var) or default
+  else
+    return variable_value or default
   end
-
-  -- Unknown variable.
-  return nil
 end
 
 --- Transforms the given text into an array of lines (so no line contains `\n`).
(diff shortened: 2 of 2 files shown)


refractalize:
CC @MariaSolOs

justinmk:
Is this part of the LSP snippet spec? If it isn't, could be a forward-compatibility risk.

MariaSolOs:
I'm not a fan of this feature. The LSP [limits the variables that can be used in snippets](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax), all of which were handled in the initial PR.

Moreover, in the case one argues that the spec has room for ambiguity and that extra variables should be supported, even VS Code limits such extra variables from a [predefined list](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables). I don't think a minimal LSP snippet implementation should support this kind of dynamically defined snippet specific variables.

For the functionality that's appropriate for core, I think we should limit snippet support to what the LSP defines, and leaving this kind of functionality for snippet libraries to build on top of.

refractalize:
> I'm not a fan of this feature. The LSP [limits the variables that can be used in snippets](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax), all of which were handled in the initial PR.

Except `TM_SELECTED_TEXT`...

Would you prefer that we only allow the value of `TM_SELECTED_TEXT` to be passed in and nothing else?

refractalize:
Alternatively we provide an implementation for `TM_SELECTED_TEXT`, i.e. we grab the previously visual text, or whatever text that was deleted due to a `c` normal command (for e.g.) Frankly there's quite a lot of scope for UX preferences and options that would make this more complicated than simply allowing it to be passed in by the caller.

justinmk:
> The LSP [limits the variables that can be used in snippets](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax), all of which were handled in the initial PR.

Is this different than the `Variable transformations via regexes` item in https://github.com/neovim/neovim/issues/25696 ?

clason:
Yes, providing a robust implementation of TM_SELECTED_TEXT would be in scope and welcome. But not at the price of adding off-spec features. The point of this API is expanding snippets returned as completion items, and nothing else.

MariaSolOs:
> > I'm not a fan of this feature. The LSP [limits the variables that can be used in snippets](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax), all of which were handled in the initial PR.
> 
> Except `TM_SELECTED_TEXT`...
> 
> Would you prefer that we only allow the value of `TM_SELECTED_TEXT` to be passed in and nothing else?

This has been brought up in the past. Refer to https://github.com/neovim/neovim/pull/25301#discussion_r1348178743 and https://github.com/neovim/neovim/pull/25301/files#r1360184463.

refractalize:
> The point of this API is expanding snippets returned as completion items, and nothing else.

I didn't realise this when I saw `vim.snippet`, it just looked so useful regardless of LSP. I get where you're all coming from now. Cheers.

refractalize:
Hi again, before I jump into implementation of `TM_SELECTED_TEXT` I wanted to run a few ideas by you.

From the point of view of UX, we obviously have a few different approaches, getting the visual selection was mentioned in the comments in #25301 - but this issues in that it's not clear how current the visual selection is, it might be several edits out of date by the time the snippet is expanded which could in the very least be quite surprising to the user, not to mention never being able to "unset" the visual selection so `TM_SELECTED_TEXT` will never fall back to its default.

I have experimented with a different approach which is this: If you make a change edit, that is one of `c{motion}` or `{Visual}c` or likewise with it's `s` counterparts, then we can capture the deleted text and use this as the `TM_SELECTED_TEXT`. Now that you're in insert mode, you can invoke the LSP completion and snippet as normal. However, if you simply enter insert mode without deleting text, i.e. with `i`, then `TM_SELECTED_TEXT` would fall back to it's default in the template, or `''`.

This is actually very similar to the UX in VS Code. If you highlight some text and start typing the beginning of the snippet and accept the snippet completion, the originally highlighted text (that was effectively deleted when you started typing) is now used as the `TM_SELECTED_TEXT`.

This would be implemented as the following:

```lua
local selected_text = nil

vim.api.nvim_create_autocmd("TextYankPost", {
  pattern = "*",
  callback = function(e)
    if vim.v.event.operator == "c" then
      selected_text = vim.fn.join(vim.v.event.regcontents, "\n")
    end
  end,
})

vim.api.nvim_create_autocmd("InsertLeave", {
  pattern = "*",
  callback = function(e)
    selected_text = nil
  end,
})
```

That is, when we see a yank, but the operator was a change (triggered by `c` or `s` normal commands) then we capture the deleted text until the end of the insert mode. There might be a more elegant way of doing this of course.

The other aspect of this is how this behaviour might be configured, or indeed "started" if we want to leave it off by default. I'm less sure about this, but I propose one of the following:

1. We provide a user settable `vim.snippet.get_selected_text` function that would be a user (or plugin) specified implementation for getting the selected text. This approach would leave the implementation to plugins (using perhaps the above approach, or other). `TM_SELECTED_TEXT` would effectively be unset by default.
2. We prefer the above "c operator" approach, and ship this as the default. My concern here is that we might be incurring a small performance hit with these autocmds if the feature is never used.
3. We have an explicit opt-in on the "c operator" approach which can be turned on with something like `vim.snippet.setup({ capture_selected_text = true })` or similar.

I have no idea which (or if any) of these approaches are vaguely workable - very happy to discuss.

MariaSolOs:
> it's not clear how current the visual selection is, it might be several edits out of date by the time the snippet is expanded which could in the very least be quite surprising to the user, not to mention never being able to "unset" the visual selection so `TM_SELECTED_TEXT` will never fall back to its default.

This is exactly why we decided not to support this variable. No matter what we do, I feel like someone will find the behaviour unexpected and counterintuitive.

> I have experimented with a different approach which is this: If you make a change edit, that is one of `c{motion}` or `{Visual}c` or likewise with it's `s` counterparts, then we can capture the deleted text and use this as the `TM_SELECTED_TEXT`. Now that you're in insert mode, you can invoke the LSP completion and snippet as normal. However, if you simply enter insert mode without deleting text, i.e. with `i`, then `TM_SELECTED_TEXT` would fall back to it's default in the template, or `''`.

> This is actually very similar to the UX in VS Code. If you highlight some text and start typing the beginning of the snippet and accept the snippet completion, the originally highlighted text (that was effectively deleted when you started typing) is now used as the `TM_SELECTED_TEXT`.

I think this is a good idea and I'm okay with seeing a draft implementation for it. I still think someone won't like this and prefer a different heuristic, so I'm unsure...

That being said, I'm unsure of how many LSP servers actually use that variable in their snippets and perhaps we're overthinking the problem. It looks like something a user would include in custom snippets, so one can argue that if someone writes a snippet with that variable, they know how to use it.

@refractalize I would suggest putting this proposal in a new issue for your ideas not to get lost in the discussion here and for others to weigh in :)

refractalize:
@MariaSolOs thanks for considering all this.

> That being said, I'm unsure of how many LSP servers actually use that variable in their snippets and perhaps we're overthinking the problem. It looks like something a user would include in custom snippets, so one can argue that if someone writes a snippet with that variable, they know how to use it.

BTW I think you're right on this.

My original intention here was to make the `vim.snippet` API generally useful outside of the scope of LSP, useful as a snippet engine capable of expanding snippets in the way VS Code does (with it's additional variables such as `$UUID` and of course `$TM_SELECTED_TEXT`). I realise there are countless other snippet engines that can do this very thing, but since `vim.snippet` is now in core, it seemed we were only a couple of commits away from Neovim being more or less self sufficient in this regard.

I realise now that this wasn't the intention for `vim.snippet` (at least for the moment), since it's not 100% clear that any implementation of TM_SELECTED_TEXT would be universally acceptable, and like you say, and since it's not absolutely critical then I'll stop here.

Thanks for your feedback and consideration in any case.

PR #27388, opened 2024-02-08
vim-patch:9.1.{0082,0083}: redrawing can be improved

zeertzjq:
#### vim-patch:9.1.0082: Redrawing can be improved when deleting lines with 'cursorline'

Problem:  Redrawing can be improved when deleting lines with 'cursorline'.
Solution: Use smarter invalidation and adjustment.  Remove unnecessary
          UPD_VALID as it is already set at the top of the loop.  Make
          the test for vim/vim#4862 fail without the fix.
          (zeertzjq)

closes: vim/vim#13986

https://github.com/vim/vim/commit/7ce34c9a947b17a8b5e81e7c2335a63552182d10

#### vim-patch:9.1.0083: Redrawing can be improved when deleting lines with 'number'

Problem:  Redrawing can be improved when inserting/deleting lines with 'number'.
Solution: Only redraw the number column of lines below changed lines.
          Add a test as this wasn't previously tested.
          (zeertzjq)

closes: vim/vim#13985

https://github.com/vim/vim/commit/ae07ebc04b0726e12b1af39d52e01d86ae79ef0a


src/nvim/drawscreen.c
```diff
@@ -2326,9 +2320,12 @@ static void win_update(win_T *wp)
...
       lnum += foldinfo.fi_lines + 1;
     } else {
-      if (wp->w_p_rnu && wp->w_last_cursor_lnum_rnu != wp->w_cursor.lnum) {
-        // 'relativenumber' set and cursor moved vertically: The
-        // text doesn't need to be drawn, but the number column does.
+      // If:
+      // - 'number' is set and below inserted/deleted lines, or
+      // - 'relativenumber' is set and cursor moved vertically,
+      // the text doesn't need to be redrawn, but the number column does.
+      if ((wp->w_p_nu && mod_top != 0 && lnum >= mod_bot && buf->b_mod_xlines != 0)
```
luukvbaal, line 2327: Should `b_mod_xlines` be reset somewhere after redrawing? Currently we always update the number column upon cursor movement after lines have been inserted/deleted, until an edit is made that doesn't insert/delete lines.

zeertzjq: Hmm, a buffer can be shown in multiple windows, so maybe that should be done in `update_screen()`.

PR #27391, opened 2024-02-08
refactor(api): refactor more api functions to use arena return

bfredl:
Currently having two separate memory strategies for API return values is a bit unnecessary, and mostly a consequence of converting the hot spot cases which needed it first. But there is really no downside to using arena everywhere (which implies also directly using strings which are allocated earlier or even statically, without copy). 

There only restriction is we need to know the size of arrays in advance, but this info can often be passed on from some earlier stage if it is missing.

This collects some "small" cases. The more complex stuff will get a PR each, some which might use the new `Dict(foo)` return pattern as well.

### DIFF
--- a/src/nvim/generators/gen_api_dispatch.lua
+++ b/src/nvim/generators/gen_api_dispatch.lua
@@ -535,7 +535,6 @@ for i = 1, #functions do
     end
 
     -- function call
-    local call_args = table.concat(args, ', ')
     output:write('\n  ')
     if fn.return_type ~= 'void' then
       -- has a return value, prefix the call with a declaration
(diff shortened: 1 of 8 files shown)


src/nvim/generators/gen_api_dispatch.lua
```diff
@@ -545,58 +544,40 @@ for i = 1, #functions do
...
 
+    if fn.can_fail then
+      table.insert(call_args, 'error')
+    end
+
+    output:write(table.concat(call_args, ', '))
+    output:write(');\n')
+
     if fn.can_fail then
       -- if the function can fail, also pass a pointer to the local error object
```
justinmk, line 576: outdated comment?

PR #27394, opened 2024-02-08
build: stop installing parser.h from treesitter

dundargoc:


clason (approved):
Upstream has clarified that these should always be included from the parser's header file, not the library.

PR #25550, opened 2023-10-08
feat(api): make nvim_open_win support normal (non-floating) windows

willothy:
Goal: Make `nvim_open_win` and `nvim_win_set_config` support opening splits, and moving floating windows into splits.

Changes to `FloatConfig`:
- The `win` prop is now also used to determine which window should be split
  - Zero should split the current window, non-zero should split a specific window, and nil should create a top-level split.
  - A side-affect of this is that split windows can now be moved with `nvim_win_set_config`.
- Added `vertical` and `split` props for controlling split position

Additionally, `win_split_ins` now returns the `win_T*` pointer to the newly created window on success, and `NULL` on failure, as it can initialize windows and it is useful (required for this PR) to have a pointer to the new window after creating it.

**Needs Discussion**:
- Removing `relative = ""` entirely, or leaving it for compatibility
- Using `win=-1`  vs `win=nil` for top-level splits (would allow for `win=nil` to split the current window) ([related](https://github.com/neovim/neovim/pull/25550#issuecomment-1756655060))

TODO:
- [x] Tests
- [x] Docs
- [x] Create `split` prop
- [x] Add `vertical` boolean prop as an alias for `split="left"` / `split="below"`
- [x] Validation for updated win configs
- [x] Handle case of window already being a normal window
- [x] Update `nvim_win_get_config`
  - [x] Add `width`, `height`, and `split` props
- [x] Non-floats should use `relative=nil`, not `relative=""`
- [x] Apply width / height props to created splits
- [x]  Update current win split direction better in `set_config` ([see here](https://github.com/neovim/neovim/pull/25550#discussion_r1368034099))
- [x] More tests
  - [x] Opening windows in non-current tabpage
  - [x] Moving windows to other tabpages
  - [x] Ensure set_config being called with the retval of get_config is a no-op

https://github.com/neovim/neovim/assets/38540736/6a8e2d38-d0b4-4de5-bcca-e5d605f5609f

fixes #14315 (maybe)

PR #27397, opened 2024-02-09
refactor: rename `FloatConfig` to `WinConfig`

willothy:
`FloatConfig` is no longer used only for floats, so the name is counterintuitive.

Follows up on discussion in #25550.

Needs discussion: Should `win_T->w_float_config` be renamed to `w_config`? Left that out of this PR for now as I'm not sure.


justinmk:
> Needs discussion: Should `win_T->w_float_config` be renamed to `w_config`?

IMO yes. Could be another followup PR to avoid blocking this one.

justinmk (approved):

willothy:
> > Needs discussion: Should `win_T->w_float_config` be renamed to `w_config`?
> 
> IMO yes. Could be another followup PR to avoid blocking this one.

I agree, sounds good. Will make that one today :)

PR #26790, opened 2023-12-29
fix(startuptime): multiprocess startuptime

pabloariasal:
Running the TUI as an external process (commit https://github.com/neovim/neovim/commit/24488169564c39a506c235bf6a33b8e23a8cb528) causes issues with the reporting of startup times (`--startuptime`), as there are now two different processes running concurrently, each writing its relative times to the same file simultaneously (see issue #23036).

This can potentially mess up the formatting of the file and makes its interpretation very hard as it interleaves two startup sequences into the same timeline.

There are multiple solutions to this problem:

### 1. Only report the startup times of the "main" process (recognized by `--embed` flag)

Pros:
* The output file will contain a single timeline (only main process writes to the file)
* Easiest to implement

Cons:
* we will lose the startup times of the TUI process

### 2. Combine the output of both processes into single timeline

Pros:
* single timeline for both processes

Cons:
* hard two implement as each process starts at a different time and has a different clock (hence, can't be modelled by same timeline)

```
clock  event
0.0     main startes
0.5     main does something
0.0     tui starts
0.7     main does something
0.3     tui does something
```

### 3. Each process has its own timeline

Pros:
* relatively easy to implement
* no loss of information

Cons:
* Multiple timelines instead of one

This MR implements option 3, meaning that we report the timeline of each process independently.
It introduces the following changes:

1. each process writes the startup times to a memory buffer first *no concurrent writing takes place)
2. when the process has finished with the initialization the buffer is flushed and appended to the report
3. the report now consists of different sections, one for each process

The new report looks like this:

```
--- Startup times for TUI process ---

times in msec
 clock   self+sourced   self:  sourced script
 clock   elapsed:              other lines

000.006  000.006: --- NVIM STARTING ---
000.428  000.422: event init
000.728  000.301: early init
...
005.880  000.713: init highlight
005.882  000.002: --- NVIM STARTED ---

--- Startup times for Main process ---

times in msec
 clock   self+sourced   self:  sourced script
 clock   elapsed:              other lines

000.006  000.006: --- NVIM STARTING ---
000.409  000.403: event init
000.557  000.148: early init
000.633  000.077: locale set
...
014.383  000.430: first screen update
014.387  000.003: --- NVIM STARTED ---
```

### Implementation notes

Originally I tried to use [open_memstream](https://man7.org/linux/man-pages/man3/open_memstream.3.html) which allocates a dynamic memory buffer where each process can write its the startup times, but I had the sad realization that it doesn't exist on windows....

[...]

PR #27406, opened 2024-02-09
refactor: reduce some globals

justinmk:
ref https://github.com/neovim/neovim/pull/26790


src/nvim/profile.c
```diff
@@ -45,6 +45,7 @@ typedef struct {
 #define PRL_ITEM(si, idx)     (((sn_prl_T *)(si)->sn_prl_ga.ga_data)[(idx)])
 
 static proftime_T prof_wait_time;
+static char *startuptime_buf INIT(= NULL); // --startuptime buffer
```
zeertzjq, line 48:
```suggestion
static char *startuptime_buf; // --startuptime buffer
```

PR #24975, opened 2023-09-01
fix(lsp): robust rename()

tomtomjhj:
* Match the prefix of the list of the path components.
* Directly rename the existing buffers without deleting them, so that states like undo, window layout, etc are preserved.
* Don't write if not writable.
* buffer:// was removed in 236c20795eb9f11e21e0719b735ea741711acc08.
* Remove redundant custom bufwinid function.
* Fix tests.

PR #27408, opened 2024-02-10
fix(lsp): rename: load and list new buffer if attached to window

przepompownia:
Problem: renamed buffer isn't visible on `:ls` (if existed in some window before) and cannot be present as an alternate file. 
Solution: load it and switch `buflisted` to `true` if the old buffer was attached to a window.


wookayin:
A similar (but more comprehensive) attempt: #24975

przepompownia:
@tomtomjhj I created this PR just to fix buffer visibility on `:ls` (and switching between buffers). I converted it back to draft for looking what happens in your change (and possibly replacing `bufload` and setting `buflisted` by something that also implies sending possible missing client notification). 

It would be good to make some minor fixes before your PR is ready.

wookayin:
I wouldn't be a primary reviewer for this, but please also describe in the commit and the PR: what problem this PR tries to fix, and write tests if possible.

test/functional/plugin/lsp_spec.lua
```diff
@@ -2443,6 +2443,90 @@ describe('LSP', function()
       eq(0, lines)
       os.remove(new)
     end)
+    it('Ensures if the new buffer remains unlisted if the old was not in window before', function()
```
mfussenegger, line 2446:
```suggestion
    it('new buffer remains unlisted if the old was not in window before', function()
```

test/functional/plugin/lsp_spec.lua
```diff
@@ -2443,6 +2443,90 @@ describe('LSP', function()
...
+        vim.lsp.util.rename(old, new)
+        return vim.bo[newbufnr].buflisted
+      ]],
+        old,
+        new
+      )
+      eq(false, buflisted)
+      os.remove(new)
+    end)
+    it('Ensures if the new buffer remains unloaded if the old was not in window before', function()
```
mfussenegger, line 2466: I think we can merge this into the previous case. "new buffer remains unloaded and unlisted if the old was not shown in a window"

test/functional/plugin/lsp_spec.lua
```diff
@@ -2443,6 +2443,90 @@ describe('LSP', function()
...
+        local newbufnr = vim.fn.bufadd(new)
+        return vim.bo[newbufnr].buflisted
+      ]],
+        old,
+        new
+      )
+      eq(true, buflisted)
+      os.remove(new)
+    end)
+    it('Ensures if the new buffer is loaded if the old was in window before', function()
```
mfussenegger, line 2508: Same with this two cases too, they could be merged into a single case "renamed buffer is loaded and listed if it was in a window before"

mfussenegger (commented):
Left some minor comments about the tests. Otherwise lgtm.

Please ping me if you push changes so I can merge.

test/functional/plugin/lsp_spec.lua
przepompownia: Honestly, initially I broke it into separate cases because I couldn't find how to return double value by `exec_lua`. Please suggest an alternative (if exists) to returning table and then unpacking it.

wookayin: what's the issue about returning a table and unpacking? You can also simply do `assert(buflisted)` inside exec_lua to avoid communication over RPC.

mfussenegger: I think you can return a table like this:

```lua
local result = exec_lua [[
 ...
 return {
   listed = buflisted,
   loaded = vim.api.nvim_buf_is_loaded(...)
 }
]]
```

przepompownia: Looks simpler than triple `exec_lua` however thanks for all suggestions.

mfussenegger:
> @mfussenegger do I need to squash the commits (because of `lintcommit`) or it can be squashed at merge?

I can squash at merge

PR #27402, opened 2024-02-09
docs(lpeg): merge upstream changes

MariaSolOs:
Adds the changes merged upstream: https://github.com/LuaCATS/lpeg/pull/9

Follow-up from some typos and code style nits I found in https://github.com/neovim/neovim/pull/26989.

PR #27410, opened 2024-02-10
ci(docs): bump doxygen to 1.9.5

wookayin:
Docs generation for lua-lpeg would need a recent version of doxygen
(1.9.5+) to avoid some issues and different output. (see #27402)
Hence, the minimum required version for doxygen is bumped from 1.9.0 to 1.9.5.

Bumps doxygen on CI from 1.9.1 (latest in ubuntu-22.04) to 1.9.5,
installed via tarball releases because apt packages are outdated.

### Remarks

Doxygen 1.9.1 is too old; released in January 2021.

Doxygen 1.9.5:

- https://github.com/doxygen/doxygen/releases/tag/Release_1_9_5
- https://www.doxygen.nl/manual/changelog.html#log_1_9_5

~~Doxygen 1.10~~:

- https://github.com/doxygen/doxygen/releases/tag/Release_1_10_0
- https://www.doxygen.nl/manual/changelog.html#log_1_10_0

Ubuntu/Debian

- https://packages.ubuntu.com/search?keywords=doxygen
- https://tracker.debian.org/pkg/doxygen

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,6 +3,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
+      - '.github/workflows/docs.yml'
       - 'src/nvim/api/*.[ch]'
       - 'src/nvim/eval.lua'
       - 'runtime/lua/**.lua'
@@ -16,13 +17,29 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+    env:
+      BIN_DIR: ${{ github.workspace }}/bin
+      DOXYGEN_URL: 'https://github.com/doxygen/doxygen/releases/download/Release_1_9_5/doxygen-1.9.5.linux.bin.tar.gz'
+      DOXYGEN_VERSION: 1.9.5
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
 
       - name: Install dependencies
         run: |
-          sudo apt-get install -y doxygen python3-msgpack
+          sudo apt-get install -y python3-msgpack
+
+      - name: Install doxygen
+        run: |
+          curl --retry 5 --silent --show-error --fail -L -o /tmp/doxygen.tar.gz "$DOXYGEN_URL"
+          mkdir -p "$BIN_DIR" /opt/doxygen
+          tar xfz /tmp/doxygen.tar.gz --strip-components=1 -C /opt/doxygen
+          ln -sfn /opt/doxygen/bin/doxygen "$BIN_DIR/doxygen"
+          doxygen_version="$(doxygen -v)"
+          echo "$doxygen_version" | grep -qF "$DOXYGEN_VERSION" || {
+            echo "Unexpected doxygen version: $doxygen_version";
+            exit 1;
+          }
 
       - name: Generate docs
         run: |
--- a/scripts/gen_vimdoc.py
+++ b/scripts/gen_vimdoc.py
@@ -59,7 +59,7 @@ Element = minidom.Element
 Document = minidom.Document
 
 MIN_PYTHON_VERSION = (3, 7)
-MIN_DOXYGEN_VERSION = (1, 9, 0)
+MIN_DOXYGEN_VERSION = (1, 9, 5)
 
 if sys.version_info < MIN_PYTHON_VERSION:
     print("requires Python {}.{}+".format(*MIN_PYTHON_VERSION))


MariaSolOs:
I'm extremely happy (and as always, impressed) by your investigation and solution, yet the fact that `apt-get` isn't installing the `doxygen` version that we expect feels... odd. Is the source feed from `apt-get` outdated? If that's the case, shouldn't we report the issue there instead?

Moreover, it looks like this PR would add the extra step of us keeping `DOXYGEN_URL` up to date with the latest release, which should probably be documented in `MAINTAIN.md`. Still, the existence of this bug makes me feel that something is wrong with our dependency sources, and we shouldn't introduce a hack as dependency consumers.

wookayin:
Ubuntu/Debian packages are often outdated and usually many versions behind the latest. I don't expect them to be up-to-date. Even neovim at Debian's is stuck at 0.7.0 despite many requests/complaints have been made..

With custom install (as we do for cmake), we will have a better control. Developers on macOS will benefit because doxygen would be usually installed via Homebrew, but for someone who are using Linux (esp. Ubuntu/Debian), yes you are right -- this discrepancy could cause some issue; docgen will produce a different output. I will add the docs, and also, we should enforce "doxygen > 1.9.5" in our docgen scripts to avoid that.

MariaSolOs:
> Ubuntu/Debian packages are often outdated and usually many versions behind the latest. I don't expect them to be up-to-date. Non-LTS versions would get higher versions for the doxygen package, but for the 22.04 LTS version it's unlikely. Even neovim at Debian's is stuck at 0.7.0 despite many requests/complaints have been made..
> 
> With custom install (as we do for cmake), we will have a better control. Developers on macOS will benefit because doxygen would be usually installed via Homebrew, but for someone who are using Linux (esp. Ubuntu/Debian), yes you are right -- this discrepancy could cause some issue; docgen will produce a different output. I will add the docs, and also, we should enforce "doxygen >= 1.9.5" in our docgen scripts to avoid that. This may cause some inconvenience on Ubuntu though.

I see, thank you for the info.

_Personally_, I think that the extra manual maintenance to make sure that we're up-to-date with the latest `doxygen` release is worth it. I still insist on proper documentation explaining this, and (if possible) an issue filed upstream reporting this versioning issue.

wookayin:
I agree with you. We don't have to use the latest version of doxygen at all times, but when we run into any new issues (different behaviors and output), we could bump up the minimum version requirement in the future. So at this point, it would be 1.9.5.

MariaSolOs:
> I agree with you. We don't have to use the latest version of doxygen at all times, but when we run into any new issues (different behaviors and output), we could bump up the minimum version requirement in the future. So at this point, it would be 1.9.5.

This sounds good to me. I still want to wait to listen for what other maintainers have to say.

MariaSolOs (approved):

.github/workflows/docs.yml
```diff
@@ -16,13 +17,29 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+    env:
+      BIN_DIR: ${{ github.workspace }}/bin
+      DOXYGEN_URL: 'https://github.com/doxygen/doxygen/releases/download/Release_1_10_0/doxygen-1.10.0.linux.bin.tar.gz'
+      DOXYGEN_VERSION: 1.10.0
```
dundargoc, line 23: I think this is sus. I know `1.9.5` is minimum but I think we should keep the minimum and the version we download in sync if we're going with this route. It make more sense to me to download 1.9.5.

.github/workflows/docs.yml
```diff
@@ -3,6 +3,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
+      - '.github/workflows/docs.yml'
```
dundargoc, line 6: Nice! I completely missed this.

CONTRIBUTING.md
```diff
@@ -271,6 +271,8 @@ To validate the documentation files, run:
 make lintdoc
 ```
 
+Note: doxygen >= 1.9.5 is required.
```
dundargoc, line 274: Would it be too lazy to say something like "The minimum version is documented at the top scripts/gen_vimdoc.py" or something? We don't really document the minimum uncrustify version and such either. Just an idea, feels like it'd require less maintenance.

dundargoc:
Tbf we don't have to continually update the version to use the latest. If a version we set works for us then I don't think we need to update for the sake of it.

.github/workflows/docs.yml
```diff
@@ -16,13 +17,29 @@ jobs:
...
+        run: |
+          curl --retry 5 --silent --show-error --fail -L -o /tmp/doxygen.tar.gz "$DOXYGEN_URL"
+          mkdir -p "$BIN_DIR" /opt/doxygen
+          tar xfz /tmp/doxygen.tar.gz --strip-components=1 -C /opt/doxygen
+          ln -sfn /opt/doxygen/bin/doxygen "$BIN_DIR/doxygen"
+          doxygen_version="$(doxygen -v)"
+          echo "$doxygen_version" | grep -qF "$DOXYGEN_VERSION" || {
+            echo "Unexpected doxygen version: $doxygen_version";
+            exit 1;
+          }
```
dundargoc, line 42: Humor me, can you test this and check that it doesn't take too long? I'm wondering if the extra CI time is worth the simplicity.

```suggestion
- uses: cachix/install-nix-action@v25
- run: nix-env -iA doxygen -f https://github.com/NixOS/nixpkgs/archive/9957cd48326fe8dbd52fdc50dd2502307f188b0d.tar.gz
```

.github/workflows/docs.yml
```diff
@@ -16,13 +17,29 @@ jobs:
...
+      DOXYGEN_URL: 'https://github.com/doxygen/doxygen/releases/download/Release_1_10_0/doxygen-1.10.0.linux.bin.tar.gz'
+      DOXYGEN_VERSION: 1.10.0
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
 
       - name: Install dependencies
         run: |
-          sudo apt-get install -y doxygen python3-msgpack
+          sudo apt-get install -y python3-msgpack
```
justinmk, line 30: could we perhaps fail CI if apt gets >= 1.9.5, and print a message that reminds us to switch back to apt instead of building from the tarball?

CONTRIBUTING.md
wookayin: I agree! We could just drop this line and minimize maintenance points, because anyway users will get error messages when running `make doc`.

.github/workflows/docs.yml
wookayin: Yes, makes sense! Either download the latest (so that CI can let people know when something starts to break) or pin at the minimum version.

.github/workflows/docs.yml
wookayin: Actually not too bad: https://github.com/wookayin/neovim/actions/runs/7856673696/job/21439682545?pr=3

took 4s for installing nix and 19s for installing doxygen via nix. The baseline (tarball) is 2s. So the additional cost is around 20 seconds.

dundargoc: Eh, unsure if it's worth it. Up to you. Anyway, thanks for testing it.

dundargoc:
Superseded by https://github.com/neovim/neovim/pull/27480. If that gets blocked for some reason or other we can resurrect this again.

Issue #14567, opened 2021-05-16
tab drop: TabLeave and TabEnter are triggered unnecessary number of times 


przepompownia:
<details><summary>`nvim --version`</summary>
<p>

```
NVIM v0.5.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -g -O2 -fdebug-prefix-map=/build/neovim-JcImo2/neovim-0.5.0+ubuntu2+git202105161736-dab6b08a1-d569569c9=. -fstack-protector-strong -Wformat -Werror=format-security -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/build/neovim-JcImo2/neovim-0.5.0+ubuntu2+git202105161736-dab6b08a1-d569569c9/build/config -I/build/neovim-JcImo2/neovim-0.5.0+ubuntu2+git202105161736-dab6b08a1-d569569c9/src -I/build/neovim-JcImo2/neovim-0.5.0+ubuntu2+git202105161736-dab6b08a1-d569569c9/.deps/usr/include -I/usr/include -I/build/neovim-JcImo2/neovim-0.5.0+ubuntu2+git202105161736-dab6b08a1-d569569c9/build/src/nvim/auto -I/build/neovim-JcImo2/neovim-0.5.0+ubuntu2+git202105161736-dab6b08a1-d569569c9/build/include
Compiled by buildd@lgw01-amd64-050


Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```
</p>
</details>
- Operating system/version: Linux _ 5.10.0-6-amd64 #1 SMP Debian 5.10.28-1 (2021-04-09) x86_64 GNU/Linux
- Terminal name/version:
[...]

PR #27411, opened 2024-02-10
vim-patch:9.1.0087: Restoring lastused_tabpage too early in do_arg_all()

### LINKED ISSUE #14657: position in the changelist is reset to 0 when opening a buffer in floating windows
There's seems to be an [issue](https://github.com/folke/zen-mode.nvim/issues/15) with changelists in buffers that are opened in floating windows.

1. open any file and make some edits, so that you have a changelist
2. open that buffer in a floating window (using the `nvim_open_win` or [zen-mode](https://github.com/folke/zen-mode.nvim))
3. `g;` now gives the error `E662: at start of changelist`

glepnir:
Problem:  Restore lastused_tabpage too early in do_arg_all() function it
          will change later in the function.
Solution: Restore lastused_tabpage a bit later, when being done with
          tabpages (glepnir)

closes: vim/vim#13992

https://github.com/vim/vim/commit/2975a54f285e5b4bf026c1dc706b5d90777d64e7

### DIFF
--- a/src/nvim/arglist.c
+++ b/src/nvim/arglist.c
@@ -1093,11 +1093,6 @@ static void do_arg_all(int count, int forceit, int keep_tabs)
   // When the ":tab" modifier was used do this for all tab pages.
   arg_all_close_unused_windows(&aall);
 
-  // Now set the last used tabpage to where we started.
-  if (valid_tabpage(new_lu_tp)) {
-    lastused_tabpage = new_lu_tp;
-  }
-
   // Open a window for files in the argument list that don't have one.
   // ARGCOUNT may change while doing this, because of autocommands.
   if (count > aall.opened_len || count <= 0) {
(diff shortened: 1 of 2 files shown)


zeertzjq:
This doesn't fix #14567

glepnir:
oops.. i just test the case that he mentioned in vim  issue comment 

> but it fails with another simple test. In the above example please run tab drop f4 and type g<Tab>. For me, the last used tab contains f3, not expected f1.

this should be fixed , but the issue not this ?

justinmk:
Anyone know if  there's an open issue for this: https://github.com/tpope/vim-fugitive/pull/2262#issuecomment-1924273614
Seems like `settabvar()` sets `lastused_tabpage`, which is a bug imo.

zeertzjq:
Fixed in #27450

PR #27414, opened 2024-02-10
refactor(api): use an arena for nvim_buf_get_lines and buffer updates

bfredl:
Refactor some earlier "temporary Array" code in buffer_updates.c to use the modern style of MAXSIZE_TEMP_ARRAY and ADD_C

### DIFF
--- a/src/nvim/api/buffer.c
+++ b/src/nvim/api/buffer.c
@@ -264,6 +264,7 @@ ArrayOf(String) nvim_buf_get_lines(uint64_t channel_id,
                                    Integer start,
                                    Integer end,
                                    Boolean strict_indexing,
+                                   Arena *arena,
                                    lua_State *lstate,
                                    Error *err)
   FUNC_API_SINCE(1)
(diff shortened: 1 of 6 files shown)


src/nvim/api/buffer.c
```diff
@@ -295,18 +296,10 @@ ArrayOf(String) nvim_buf_get_lines(uint64_t channel_id,
...
-  if (!buf_collect_lines(buf, size, (linenr_T)start, 0, (channel_id != VIML_INTERNAL_CALL), &rv,
-                         lstate, err)) {
-    goto end;
-  }
-
-end:
-  if (ERROR_SET(err)) {
-    api_free_array(rv);
-    rv.items = NULL;
-  }
```
justinmk, line 309: Being able to skip this boilerplate is a nice quality-of-life gain from the Arena pattern.

src/nvim/api/buffer.c
```diff
@@ -1413,14 +1399,15 @@ static inline void init_line_array(lua_State *lstate, Array *a, size_t size)
 /// @param a           Array to push onto when not using Lua
 /// @param s           String to push
 /// @param len         Size of string
-/// @param idx         0-based index to place s
+/// @param idx         0-based index to place s, only used for lua
```
justinmk, line 1402:
```suggestion
/// @param idx         0-based index to place s, only used for Lua
```

src/nvim/api/buffer.c
```diff
@@ -1450,27 +1437,17 @@ static void push_linestr(lua_State *lstate, Array *a, const char *s, size_t len,
 /// @param n Number of lines to collect
 /// @param replace_nl Replace newlines ("\n") with NUL
 /// @param start Line number to start from
-/// @param start_idx First index to push to
+/// @param start_idx First index to push to (only for lua)
```
justinmk, line 1440:
```suggestion
/// @param start_idx First index to push to (only for Lua)
```

PR #27418, opened 2024-02-10
vim-patch:9.1.{0089,0093}: qsort() comparison function fixes

zeertzjq:
#### vim-patch:9.1.0089: qsort() comparison functions should be transitive

Problem:  qsort() comparison functions should be transitive
Solution: Do not subtract values, but rather use explicit comparisons

Improve qsort() comparison functions

There has been a recent report on qsort() causing out-of-bounds read &
write in glibc for non transitive comparison functions
https://www.qualys.com/2024/01/30/qsort.txt

Even so the bug is in glibc's implementation of the qsort() algorithm,
it's bad style to just use substraction for the comparison functions,
which may cause overflow issues and as hinted at in OpenBSD's manual
page for qsort(): "It is almost always an error to use subtraction to
compute the return value of the comparison function."

So check the qsort() comparison functions and change them to be safe.

closes: vim/vim#13980

https://github.com/vim/vim/commit/e06e43766500ecb4cd1031fa16cf9cbebdb222c1

#### vim-patch:9.1.0093: Still a qsort() comparison function that returns result of subtraction

Problem:  Still a qsort() comparison function fuzzy_match_item_compare()
          that returns result of subtraction (after 9.1.0089).
Solution: Use an explicit comparison instead of subtraction.
          (zeertzjq)

closes: vim/vim#14004

https://github.com/vim/vim/commit/77078276bfe695070441a1bbdc02949d31de8922

### DIFF
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -673,10 +673,20 @@ int sign_item_cmp(const void *p1, const void *p2)
 {
   const SignItem *s1 = (SignItem *)p1;
   const SignItem *s2 = (SignItem *)p2;
-  int n = s2->sh->priority - s1->sh->priority;
 
-  return n ? n : (n = (int)(s2->id - s1->id))
-         ? n : (s2->sh->sign_add_id - s1->sh->sign_add_id);
+  if (s1->sh->priority != s2->sh->priority) {
+    return s1->sh->priority < s2->sh->priority ? 1 : -1;
+  }
+
+  if (s1->id != s2->id) {
+    return s1->id < s2->id ? 1 : -1;
+  }
+
+  if (s1->sh->sign_add_id != s2->sh->sign_add_id) {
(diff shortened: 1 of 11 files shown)


src/nvim/decoration.c
```diff
@@ -673,10 +673,20 @@ int sign_item_cmp(const void *p1, const void *p2)
...
+  if (s1->sh->priority != s2->sh->priority) {
+    return s1->sh->priority < s2->sh->priority ? 1 : -1;
+  }
+
+  if (s1->id != s2->id) {
+    return s1->id < s2->id ? 1 : -1;
+  }
+
+  if (s1->sh->sign_add_id != s2->sh->sign_add_id) {
+    return s1->sh->sign_add_id > s2->sh->sign_add_id ? 1 : -1;
```
luukvbaal, line 686: Regressed here, fixed in #27431.

PR #24764, opened 2023-08-17
feat(ui): forward :intro message properly to external UI:s

bfredl:
The intro message doesn't work very well with multigrid UI:s The message is plainly drawn on the `default_grid` which is kinda ok in TUI but doesn't provide any useful context for external UI:s.

Introduce a new `msg_intro` event with is used when `ext_multigrid` and/or `ext_messages` is set.

Fixes #24705

PR #27423, opened 2024-02-11
fix(messages): use message grid for :intro

zeertzjq:
TODO:
- cmdline area is scrolled up, not cleared
- not fully drawn when 'cmdheight' is very large
- ext_messages? #24764


zeertzjq:
Calling `msg_scroll_flush()` before drawing the lines (i.e. drawing the lines without throttling) will lead to fewer problems, but the UI will receive a lot of blank lines.

src/nvim/version.c
Hanymato, line 1: 7ee02ad56efea08710c922b9c1e614fc26883119

PR #27425, opened 2024-02-11
docs: describe interaction between editorconfig and ftplugin

znd4:
Hi, I'm not sure if the behavior I'm documenting here is actually intentional, or is meant to be guaranteed to users, but it does seem intuitive, and does match the behavior I've seen when experimenting, so I decided to create a PR in case it's worth mentioning in the docs

### DIFF
--- a/runtime/doc/editorconfig.txt
+++ b/runtime/doc/editorconfig.txt
@@ -86,4 +86,11 @@ Example: >lua
       vim.b[bufnr].foo = val
     end
 <
+
+EditorConfig overrides filetype plugins           *ftplugin-vs-editorconfig*
+
+Nvim's EditorConfig resolution runs after |ftplugin| processing. This means
+that EditorConfig settings override global filetype settings
+(such as those in the |after-directory|).
+
  vim:tw=78:ts=8:et:sw=4:ft=help:norl:
--- a/runtime/doc/editorconfig.txt
+++ b/runtime/doc/editorconfig.txt
@@ -6,7 +6,7 @@
 
 EditorConfig integration                                        *editorconfig*
 
-Nvim supports EditorConfig. When a file is opened, Nvim searches all parent
+Nvim supports EditorConfig. When a file is opened, after running |ftplugin|s, Nvim searches all parent
 directories of that file for ".editorconfig" files, parses them, and applies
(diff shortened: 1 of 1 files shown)


runtime/doc/editorconfig.txt
```diff
@@ -86,4 +86,11 @@ Example: >lua
       vim.b[bufnr].foo = val
     end
 <
+
+EditorConfig overrides filetype plugins           *ftplugin-vs-editorconfig*
```
znd4, line 90: I'm also not sure about this help tag. Maybe `editorconfig` should come first? Maybe `*-and-*`?

wookayin: editorconfig-vs-ftplugin would be better.

wookayin:
> if the behavior I'm documenting here is actually intentional, or is meant to be guaranteed to users, but it does seem intuitive

FYI, Editorconfig runs on 'BufNewFile', 'BufRead', 'BufFilePost', and the current behavior and ordering is due to BufRead (BufReadPost) taking place *after* FileType, which makes sense because editorconfig can update per-project configs (indentation, etc.), overriding global, default indentation configs.

runtime/doc/editorconfig.txt
```diff
@@ -86,4 +86,11 @@ Example: >lua
       vim.b[bufnr].foo = val
     end
 <
+
+EditorConfig overrides filetype plugins           *editorconfig-vs-ftplugin*
```
gpanders, line 90: I don't think this needs a separate section or its own tag. I would just update the sentence near the beginning of the help file.

For example, change "When a file is opened" to "When a file is opened, after running |ftplugin|s,".

znd4: > For example, change \"When a file is opened\" to \"When a file is opened, after running |ftplugin|s,\".

Ooh, nice. I have much to learn about technical documentation.

gpanders:
@znd4 Is this ready for review?

wookayin:
One note is that `after/ftplugin` would be executed after editorconfig, and it'd be even better if we can include this information as well in the documentation.

dundargoc:
Superseded by https://github.com/neovim/neovim/pull/28736

PR #27426, opened 2024-02-11
vim-patch:6d91227267a5

clason:
runtime(gpg): Mark dangerous use-embedded-filename with WarningMsg

The syntax highlighter is likely to encourage people to use the listed
commands.

But `use-embedded-filename` is a dangerous option that can cause GnuPG
to write arbitrary data to arbitrary files whenever GnuPG encounters
malicious data.

GnuPG upstream explicitly warns against using this option:

https://dev.gnupg.org/T4500
https://dev.gnupg.org/T6972

However, since this is a valid option, we cannot just drop it from the
syntax script. Instead, let's mark it with the WarningMsg highlighting
to make it obvious, that this option is different (and should not be
used for security reasons).

closes: vim/vim#13961

https://github.com/vim/vim/commit/6d91227267a5f7e21fc9b30443687618e92751bf

Co-authored-by: Daniel Kahn Gillmor <dkg@fifthhorseman.net>

### DIFF
--- a/runtime/syntax/gpg.vim
+++ b/runtime/syntax/gpg.vim
@@ -1,9 +1,13 @@
 " Vim syntax file
 " Language:		gpg(1) configuration file
+" Maintainer: This file is looking for a maintainer!
 " Previous Maintainer:	Nikolai Weibull <now@bitwi.se>
-" Latest Revision:	2010-10-14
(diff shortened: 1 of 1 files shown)


clason:
🙄

seandewar (approved):

PR #27429, opened 2024-02-11
docs: correct on_key docs

altermo:


zeertzjq:
`Passing in nil when {ns} is specified` maybe?

PR #25826, opened 2023-10-29
refactor(decorations): break up Decoration struct into smaller pieces

bfredl:
Remove the monolithic Decoration struct. Before this change, each extmark could either represent just a hl_id + priority value as a inline decoration, or it would take a pointer to this monolitic 112 byte struct which has to be allocated.

This separates the decorations into two pieces: DecorSignHighlight for signs, highlights and simple set-flag decorations (like spell, ui-watched), and DecorVirtText for virtual text and lines.

The main separation here is whether they are expected to allocate more memory. Currently this is not really true as sign text has to be an allocated string, but the plan is to get rid of this eventually (it can just be an array of two schar_T:s). Further refactors are expected to improve the representation of each decoration kind individually. The goal of this particular PR is to get things started by cutting the Gordian knot which was the monolithic struct Decoration.

Now, each extmark can either contain chained indices/pointers to these kinds of objects, or it can fit a subset of DecorSignHighlight inline.

The point of this change is not only to make simple decorations smaller in memory. In fact, the main motivation is to later allow them to grow **larger**, but on a dynamic, on demand fashion. As a simple example, it would be possible to augment highlights to take a list of multiple `hl_group`:s, which then would trivially map to a chain of multiple DecorSignHighlight entries.

One small feature improvement included with this refactor itself, is that the restriction that extmarks cannot be removed inside a decoration provider has been lifted. These are instead safely lifetime extended on a "to free" list until the current iteration of screen drawing is done.

PR #27431, opened 2024-02-11
fix(column): use a single path for sign sorting

luukvbaal:
Problem:  #25826 added a (duplicate) sign comparison function, which was modified and strayed from the original in #27418.
Solution: Merge the two functions and add display test that actually tests for this order in addition to the legacy tests (which test the builtin-functions return order).

### DIFF
--- a/src/nvim/sign.c
+++ b/src/nvim/sign.c
@@ -168,7 +168,7 @@ static int buf_findsign(buf_T *buf, int id, char *group)
 }
 
 /// qsort() function to sort signs by line number, priority, id and recency.
-int sign_cmp(const void *p1, const void *p2)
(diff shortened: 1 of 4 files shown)


src/nvim/sign.c
```diff
@@ -168,7 +168,7 @@ static int buf_findsign(buf_T *buf, int id, char *group)
 }
 
 /// qsort() function to sort signs by line number, priority, id and recency.
-int sign_cmp(const void *p1, const void *p2)
+int sign_row_cmp(const void *p1, const void *p2)
```
zeertzjq, line 171:
```suggestion
static int sign_row_cmp(const void *p1, const void *p2)
```

PR #27052, opened 2024-01-17
fix(column): remove sign from line it was previously on with undo

luukvbaal:

PR #27437, opened 2024-02-12
fix(extmarks): redraw pre-undo position

### LINKED ISSUE #27368: Virtual text shown on two lines after delete and undo
### Problem

Virtual text shown on two lines after delete and undo:

https://github.com/neovim/neovim/assets/35768171/509dc921-049b-455b-9958-5b01e69146c2

### Steps to reproduce

1. Run `nvim --clean`
2. Source the following file:
```vim
call setline(1, ['aaa', 'bbb', 'ccc', 'ddd'])
let ns = nvim_create_namespace('')
call nvim_buf_set_extmark(0, ns, 1, 0, #{virt_text: [['foo']]})
```
3. Press `j`
4. Press `dd`
5. Press `u`
6. Virtual text is shown on both line 2 and line 3

### Expected behavior

Virtual text is only shown on line 2, which is the behavior after pressing `<C-L>`

### Neovim version (nvim -v)

v0.10.0-dev-2294+gc0b99bb1de

N/A (deletes text property with line)

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.31.0

xterm-kitty

### Installation

build from repo

zeertzjq:
Problem:
Virtual text not redrawn properly after undo moves its extmark.

Solution:
Redraw the moved extmark's pre-undo position.

Fix #27368

### DIFF
--- a/src/nvim/undo.c
+++ b/src/nvim/undo.c
@@ -2390,9 +2390,7 @@ static void u_undoredo(bool undo, bool do_buf_event)
     // When text has been changed, possibly the start of the next line
     // may have SpellCap that should be removed or it needs to be
     // displayed.  Schedule the next line for redrawing just in case.
-    // Also just in case the line had a sign which needs to be removed.
-    if ((spell_check_window(curwin) || buf_meta_total(curbuf, kMTMetaSignText))
-        && bot <= curbuf->b_ml.ml_line_count) {
+    if (spell_check_window(curwin) && bot <= curbuf->b_ml.ml_line_count) {
       redrawWinline(curwin, bot);
     }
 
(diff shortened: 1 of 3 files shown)


zeertzjq:
Note: both this PR and #27052 causes a full screen redraw, because both `b_mod_set` and `w_redraw_top` are set.

src/nvim/undo.c
```diff
@@ -2391,8 +2391,7 @@ static void u_undoredo(bool undo, bool do_buf_event)
     // may have SpellCap that should be removed or it needs to be
     // displayed.  Schedule the next line for redrawing just in case.
     // Also just in case the line had a sign which needs to be removed.
```
luukvbaal, line 2393:
```suggestion
```

bfredl (approved):

zeertzjq:
#27442 will fix the full window redraw.

PR #27446, opened 2024-02-12
feat(lsp): add option to clear default input string on `vim.lsp.buf.rename`

Akmadan23:
A really simple feature that's nice to have in my opinion...


mfussenegger (commented):
Isn't this roughly the same as calling `vim.lsp.buf.rename(vim.fn.input('New Name: '))` ? Not sure if a new option is needed.

justinmk:
Absolutely not. Adding random flags for every little thing is never acceptable. Please don't send PRs like this.

Issue #25509, opened 2023-10-05
Lua: vim.Position / vim.Range abstraction


justinmk:
### Problem

- There are various different ways of representing a position (e.g. `:help api-indexing`).
- Comparing or reasoning about positions requires boilerplate / utility functions such as:
    - https://github.com/neovim/neovim/pull/24499#issuecomment-1658916388
    - https://github.com/vscode-neovim/vscode-neovim/pull/1490/files#diff-356c71d28d1ad83654bec96ec39481938dd06e6589a05d781ed2ab028d9296b6R70-R84

Copying @mfussenegger summary:

> cursor/position related functions:
> 
> * `vim.region(..., pos1, pos2, ...)`
>   * pos1 & pos2: `integer[] (line, column)` `(0,0)` indexed byte positions
> * `vim.inspect_pos(..., row, col, ...)` 0 based
> * `vim.show_pos(..., row, col, ...)` 0 based
> * `vim.highlight.range(... start, finish, ...)` - `{line, col}` tuple, doesn't mention if `(0, 0)` or something else?
> * `vim.treesitter.get_node_text(node, ...)`
> * `vim.fn.cursor(lnum, col, off)` `(1, 1)` - With `0` meaning stay in current line or column
> * `vim.fn.getpos(expr)` - result is `[bufnr, lnum, col, off]` (1, 1) based;
> * `vim.fn.getcurpos([{winid}])` like ^ but with an extra `curswant` in the result.
> * `vim.fn.getline(lnum, end)` - 1 based
> * `nvim_win_set_cursor(win, pos)` -  `pos` is `(row, col)` tuple  with `(1, 0)` indexing
> * `nvim_win_get_cursor(win)` - Mirrors `nvim_win_set_cursor`
> * `nvim_buf_get_lines(..., start, end, ...)` - 0 based, end-exclusive
> * `nvim_buf_get_lines(..., start_row, start_col, end_row, end_col...)` - 0 based, row indices end-inclusive, column indices end-exclusive
> 
> LSP structures:
> 
> ```
> Position
>    line: integer 0 based
>    character: integer 0 base  (but encoding dependent
> ```
> 
> ```
> Range
>     start: Position
>     end: Position
> ```

### Expected behavior

Implement `vim.pos.Pos` which represents a normalized position.

`vim.pos()` creates a position:

```
vim.pos{
  kind?: 'mark' | 'extmark' | 'cursor' | ...
  col: number
  row: number
}
```

[...]

PR #22598, opened 2023-03-10
feat(lua): vim.func.on_func()

justinmk:
## Problem:

- No easy way to hook into:
    - vim.paste
    - vim.ui.x
    - LSP request/response handlers. (No mention of "before" or "after" in `:help lsp`...)
- Hooking into functions is not ergonomic.
    - Not easy to remove temporary hooks (and restore the old one(s)).
    - No visbility/tracing for debugging/performance/checkhealth.

## Solution:
Introduce `vim.on_fun()`, a generic way to "hook into" any function before/after it is called.

This is a powerful escape-hatch to avoid needing to add formal events (autocmds) for every conceivable thing. And it leverages basic Lua features without becoming an "inner platform": it doesn't require consumers to buy-into the platform, they can continue to treat "hooked" functions like normal functions.

Example:

    vim.on_fun(vim.lsp.handlers, 'textDocument/definition', function(fn, args)
      vim.notify('before, yay')
      local rv = fn(unpack(args))
      vim.notify('after, yay')
      return rv
    end)

Fixes #20568
Fixes #22075
Fixes #22323
Fixes #24229
Related: #13977

## Todo

- naming suggestions? (`on_fn`, `on_call`, ...)?
- ✅  implement `:unhook()` ?
- ✅  some way to have visibility into how many of these damn things are hooked into a given function? `on_fun()` has the same risk as "autocmd spam" where a user/plugin accidentally, redundantly registers a handler and overloads the editor.

PR #27447, opened 2024-02-12
feat(lsp): allow passing position to lsp.buf functions

### LINKED ISSUE #27400: Ability to pass position parameters to lsp.buf functions
### Problem

At the moment, `lsp.buf.references` [requires my cursor to be on the word for which I want to get references (calls the make_position_params function)](https://github.com/rudiejd/neovim/blob/bcf96538143564bc8de26c5db28df7cfbaf8025a/runtime/lua/vim/lsp/util.lua#L1961). If I want to request references for a symbol that isn't under my cursor, I first have to navigate there then navigate back. This also applies to `lsp.buf.implementation`, `lsp.buf.definition`, etc. 

### Expected behavior

In Lua, I should be able to pass my own [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams) to my LSP server so that I can utilize all of these functions for any symbol in the buffer, not just the one under my cursor. 

Right now, my alternative options for this seem to be:
1. Do `lsp.buf_request` in lua and pass whatever parameters I want
2. Change the cursor to the position of my request, make the request, and change it back

If adding the ability to pass position isn't desired for any of the `lsp.buf` module functions, I can go with (1), but to me this seems like a feature that should be standard.

rudiejd:
Closes https://github.com/neovim/neovim/issues/27400 

I considered also making `window` an option that could be passed here, but that seems like it could run against the grain of this module being requests for the "current buffer"

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -422,6 +422,10 @@ The following changes to existing APIs or features add new behavior.
 
 • |--startuptime| reports the startup times for both processes (TUI + server) as separate sections.
 
+* |vim.lsp.start()| now maps |K| to use |vim.lsp.buf.hover()| if the server
+  supports it, unless |'keywordprg'| was customized before calling
+  |vim.lsp.start()|.
+
 ==============================================================================
 REMOVED FEATURES                                                 *news-removed*
 
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -41,11 +41,11 @@ end
 --- Displays hover information about the symbol under the cursor in a floating
 --- window. Calling the function twice will jump into the floating window.
 ---@param options table|nil additional options
----     - row: 1-indexed row number of position, defaults to current line
----     - col:0-indexed column number of position, defaults to current character
+---     - row: 0-indexed row number of position, defaults to current line
+---     - col: 0-indexed column number of position, defaults to current character
 function M.hover(options)
   options = options or {}
-  local params = make_position_params(options.row, options.col)
+  local params = util.make_position_params(nil, nil, options.row, options.col)
   request(ms.textDocument_hover, params)
 end
 
@@ -68,11 +68,11 @@ end
 ---     - reuse_win: (boolean) Jump to existing window if buffer is already open.
 ---     - on_list: (function) |lsp-on-list-handler| replacing the default handler.
 ---        Called for any non-empty result.
----     - row: 1-indexed row number of position, defaults to current line
----     - col:0-indexed column number of position, defaults to current character
+---     - row: 0-indexed row number of position, defaults to current line
+---     - col: 0-indexed column number of position, defaults to current character
 function M.declaration(options)
   options = options or {}
-  local params = make_position_params(options.row, options.col)
+  local params = util.make_position_params(nil, nil, options.row, options.col)
   request_with_options(ms.textDocument_declaration, params, options)
 end
 
@@ -82,11 +82,11 @@ end
 ---     - reuse_win: (boolean) Jump to existing window if buffer is already open.
 ---     - on_list: (function) |lsp-on-list-handler| replacing the default handler.
 ---       Called for any non-empty result.
----     - row: 1-indexed row number of position, defaults to current line
----     - col:0-indexed column number of position, defaults to current character
+---     - row: 0-indexed row number of position, defaults to current line
+---     - col: 0-indexed column number of position, defaults to current character
 function M.definition(options)
   options = options or {}
-  local params = make_position_params(options.row, options.col)
+  local params = util.make_position_params(nil, nil, options.row, options.col)
   request_with_options(ms.textDocument_definition, params, options)
 end
 
@@ -96,11 +96,11 @@ end
 ---     - reuse_win: (boolean) Jump to existing window if buffer is already open.
 ---     - on_list: (function) |lsp-on-list-handler| replacing the default handler.
 ---       Called for any non-empty result.
----     - row: 1-indexed row number of position, defaults to current line
----     - col:0-indexed column number of position, defaults to current character
+---     - row: 0-indexed row number of position, defaults to current line
+---     - col: 0-indexed column number of position, defaults to current character
 function M.type_definition(options)
   options = options or {}
-  local params = make_position_params(options.row, options.col)
+  local params = util.make_position_params(nil, nil, options.row, options.col)
   request_with_options(ms.textDocument_typeDefinition, params, options)
 end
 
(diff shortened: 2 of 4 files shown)


runtime/lua/vim/lsp/util.lua
```diff
@@ -1956,18 +1961,19 @@ local function make_position_param(window, offset_encoding)
...
 --- Creates a `TextDocumentPositionParams` object for the current buffer and cursor position.
----
+---@param row integer|nil 1-indexed row number of position, defaults to current line
+---@param col integer|nil 0-indexed column number of position, defaults to current character
 ---@param window integer|nil: window handle or 0 for current, defaults to current
 ---@param offset_encoding string|nil utf-8|utf-16|utf-32|nil defaults to `offset_encoding` of first client of buffer of `window`
 ---@return table `TextDocumentPositionParams` object
 ---@see https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams
-function M.make_position_params(window, offset_encoding)
+function M.make_position_params(row, col, window, offset_encoding)
```
wookayin, line 1970: We should not break the function by changing the signature.

runtime/lua/vim/lsp/util.lua
```diff
@@ -1939,10 +1939,15 @@ end
 
 ---@param window integer|nil: window handle or 0 for current, defaults to current
 ---@param offset_encoding string utf-8|utf-16|utf-32|nil defaults to `offset_encoding` of first client of buffer of `window`
-local function make_position_param(window, offset_encoding)
+---@param row integer|nil 1-indexed row number of position, defaults to current line
+---@param col integer|nil 0-indexed column number of position, defaults to current character
+---@return table position line and character number within buffer
+local function make_position_param(window, offset_encoding, row, col)
```
wookayin, line 1945: I don't think using 1,0-indexing is a good idea for LSP APIs. see `|api-indexing|`. I think using 0-based indexing for all row and col would be better. The LSP protocol uses 0-based indexing for the Range representations.

The row, col indexing used for this function is only due to `nvim_win_get_cursor` and other mark-related APIs specifically.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -40,8 +40,12 @@ end
 
 --- Displays hover information about the symbol under the cursor in a floating
 --- window. Calling the function twice will jump into the floating window.
-function M.hover()
-  local params = util.make_position_params()
+---@param options table|nil additional options
```
wookayin, line 43:
```suggestion
---@param opts? table additional options
```
Not only this place.

Also whitespace is not consistent on `col`.

runtime/lua/vim/lsp/util.lua
```diff
@@ -1939,10 +1939,15 @@ end
...
   window = window or 0
   local buf = api.nvim_win_get_buf(window)
-  local row, col = unpack(api.nvim_win_get_cursor(window))
+  if not (row or col) then
+    row, col = unpack(api.nvim_win_get_cursor(window))
+  end
```
wookayin, line 1950: This does not align with what is documented about the default values of `row` and `col`.

wookayin (changes requested):

runtime/lua/vim/lsp/util.lua
rudiejd: d'oh, I forget the util module is public... I got annoyed by repeatedly passing `(nil, nil, row, col)`, though I suppose I could get past that by just passing a table as args

runtime/lua/vim/lsp/util.lua
rudiejd: Gotcha, thanks for the context. I can make this 0 based for input without breaking the existing behavior

runtime/lua/vim/lsp/buf.lua
rudiejd: I originally named this as `opts`, but wanted consistency with the arguments that other members of this module already have. Should everything be `opts?`?

runtime/lua/vim/lsp/util.lua
wookayin: Actually.... I just realized that `buf.format()` and `make_given_range_params` is using 1,0-indexing 🤔 (which I'd want to change and break but it's difficult to change at this point). Probably wasn't carefully chosen in retrospect or there could be some reason, e.g. wanted to avoid conversion from the callsite; Some other core devs

runtime/lua/vim/lsp/buf.lua
wookayin: I see. Options would be fine then. Either we change all or remain consistent.

runtime/doc/news.txt
```diff
@@ -422,6 +422,10 @@ The following changes to existing APIs or features add new behavior.
 
 • |--startuptime| reports the startup times for both processes (TUI + server) as separate sections.
 
+* |vim.lsp.start()| now maps |K| to use |vim.lsp.buf.hover()| if the server
```
MariaSolOs, line 425: This change looks unrelated to this PR (?).

MariaSolOs (commented):
I'm not a fan of this change. I feel like we should invest in a more flexible and general solution like https://github.com/neovim/neovim/pull/22598 for LSP-handler customization.

wookayin:
> I feel like we should invest in a more flexible and general solution like #22598 for LSP-handler customization.

I think this PR is not about LSP *handlers*, but is about adding position parameters to the "request" functions other than the current cursor location. But I can (partially) agree with you, in a sense that rather than too specific parameters for every use cases we would want to have more **general** LSP request functions that would allow injection of any additional LSP request parameters at ease. For example, we may also want to have `buf = ...`, `win = ...`, as well as cursor locations. Ideally, it would be even better if users can build such LSP parameters (e.g. [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams)) *outside* using some utility APIs, and pass these tables to inject them into LSP requests being sent.

MariaSolOs:
> I think this PR is not about LSP _handlers_, but is about adding position parameters to the "request" functions other than the current cursor location. But I can (partially) agree with you, in a sense that rather than too specific parameters for every use cases we would want to have more **general** LSP request functions that would allow injection of any additional LSP request parameters at ease. For example, we may also want to have `buf = ...`, `win = ...`, as well as cursor locations. Ideally, it would be even better if users can build such LSP parameters (e.g. [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams)) _outside_ using some utility APIs, and pass these tables to inject them into LSP requests being sent.

Well I count this as overloading/overriding/decoration of handlers ;)

But yes, I agree with you. Perhaps expansing `vim.lsp.with` to accommodate for these scenarios would be nice, because having custom options (with several common properties) for every LSP handler doesn't feel very maintainable to me.

rudiejd:
To me,  #22598 and `vim.lsp.with` don't seem like suitable substitutes for this. As @wookayin mentioned, this is more about changing an individual request the changing the handlers. I don't think users should have to override handlers to pass positions to them.

My use case for this would be searching for referenced of a symbol in my buffer that I'm not moused over, so the position could change every time I want to call the handler. I'd also like to be able to navigate to the definition of something that isn't under my cursor without polluting the jumplist.

I agree about the options being cumbersome here, though. Would it be helpful to have a common type for `TextDocumentPositionParams` that gets passed as an optional argument instead of an `options` argument? I originally went with the latter because some functions already had optional arguments, but the former seems more maintainable.

mfussenegger:
I think I still don't really understand the use-case for this.

If you're programmatically making requests, I'd think you also want to deal with the result in some special way?
If that's the case we've `client.request` and `vim.lsp.buf_request_all` where you can create requests however you want. Creating a table with a couple properties is imho not that complicated that it justifies various options and various helper functions.

And if we're going to extend this in some way, we should make sure it'll compose with whatever is happening with https://github.com/neovim/neovim/issues/25509

runtime/lua/vim/lsp/buf.lua
justinmk: > Should everything be `opts?`?

in favor of that

justinmk:
> My use case for this would be searching for referenced of a symbol in my buffer that I'm not moused over, so the position could change every time I want to call the handler. 

Placing the cursor programmatically works there. 

> I'd also like to be able to navigate to the definition of something that isn't under my cursor without polluting the jumplist.

`:help cursor()` says "Does not change the jumplist."

In general I agree that APIs should always take a position (#25509) instead of inspecting the cursor, but these `vim.lsp.buf` APIs are mostly for "interactive" tasks (e.g. `rename()` moves the cursor).

justinmk:
> I actually managed to accomplish what I wanted to do with `client.request`

Can you share a code sample in https://github.com/neovim/neovim/issues/27400

Issue #27369, opened 2024-02-07
Virtual text with left gravity shown on two lines after pasting above


zeertzjq:
### Problem

Virtual text with left gravity shown on two lines after pasting above:

https://github.com/neovim/neovim/assets/35768171/a734b112-b4f5-43dc-901e-b4a50f7509ab

### Steps to reproduce

1. Run `nvim --clean`
2. Source the following file:
```vim
call setline(1, ['aaa', 'bbb', 'ccc', 'ddd'])
let ns = nvim_create_namespace('')
call nvim_buf_set_extmark(0, ns, 1, 0, #{virt_text: [['foo']], right_gravity: v:false})
```
3. Press `yyp`
4. Virtual text is shown on both line 2 and line 3

### Expected behavior

Virtual text is only shown on one line. After pressing `<C-L>` it is shown on line 2, which may or may not be the expected behavior.

### Neovim version (nvim -v)

v0.10.0-dev-2294+gc0b99bb1de 

### Vim (not Nvim) behaves the same?

N/A

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.31.0

### $TERM environment variable

xterm-kitty

### Installation

build from repo

zeertzjq:
This bug only happens if the extmark is at the beginning of the line. Otherwise, the virtual text is only shown on line 3.

PR #27453, opened 2024-02-13
fix(extmarks): redraw new position after deleting lines

### LINKED ISSUE #27370: Wrong display when deleting line with inline virtual text causes another line to wrap
### Problem

Wrong display when deleting line with inline virtual text causes another line to wrap:

https://github.com/neovim/neovim/assets/35768171/44818e37-325c-41d4-afc3-69a102b0a9bc

### Steps to reproduce

1. Run `nvim --clean`
2. Source the following file:
```vim
call setline(1, ['aaa', 'bbb', 'ccc', 'ddd', 'eee'])
call setline(3, repeat('c', winwidth(0) - 1))
let ns = nvim_create_namespace('')
call nvim_buf_set_extmark(0, ns, 1, 0, #{virt_text: [['!!!']], virt_text_pos: 'inline'})
```
3. Press `j`
4. Press `dd`
5. The last two lines are both displayed as `eee`

### Expected behavior

The last two lines are displayed as `ddd` and `eee`

### Neovim version (nvim -v)

v0.10.0-dev-2294+gc0b99bb1de   

N/A (deletes text property with line)

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.31.0

xterm-kitty

### Installation

build from repo

zeertzjq:
Problem:
Virtual text not redrawn properly when its extmark's line is deleted.

Solution:
Redraw new positions of extmarks in the range when deleting lines.

Fix #27370

### DIFF
--- a/src/nvim/change.c
+++ b/src/nvim/change.c
@@ -167,6 +167,13 @@ void changed_internal(buf_T *buf)
 static void changed_lines_invalidate_win(win_T *wp, linenr_T lnum, colnr_T col, linenr_T lnume,
                                          linenr_T xtra)
 {
+  // If lines haven been inserted/deleted and the buffer has virt_lines,
+  // invalidate the line after the changed lines as some virt_lines may
+  // now be drawn above a different line.
+  if (xtra != 0 && buf_meta_total(wp->w_buffer, kMTMetaLines) > 0) {
+    lnume++;
+  }
+
   // If the changed line is in a range of previously folded lines,
   // compare with the first line in that range.
   if (wp->w_cursor.lnum <= lnum) {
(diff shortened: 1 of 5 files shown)


src/nvim/change.c
```diff
@@ -518,6 +526,13 @@ void deleted_lines_mark(linenr_T lnum, int count)
 /// @param xtra number of extra lines (negative when deleting)
 void changed_lines_redraw_buf(buf_T *buf, linenr_T lnum, linenr_T lnume, linenr_T xtra)
 {
+  // If lines haven been inserted/deleted and the buffer has virt_lines,
+  // mark the line after the changed lines to be redrawn as some virt_lines may
+  // now be drawn above a different line.
+  if (xtra != 0 && buf_meta_total(buf, kMTMetaLines) > 0) {
+    lnume++;
+  }
+
```
zeertzjq, line 535: This change actually reduces redraw when deleting a line with virtual lines attached, because it makes the range here the same as `changed_lines_invalidate_buf()`, which is a situation that `win_update()` handles better.

zeertzjq:
#27370 happens because splicing moves an extmark to a line that didn't change. This PR fixes it by redrawing the new positions of extmarks in the deleted range at places where this may happen. IDK if there is a more general solution.
#27369 happens because splicing moves an extmark from a line that didn't change. IDK how to fix that.

zeertzjq:
Oh, this PR also doesn't work for extmarks that have disabled `undo_restore`.

PR #27457, opened 2024-02-13
fix(vim.is_callable): use `rawget()` to retrieve `__call` from mt

tmillr:
LuaJIT uses a raw get to retrieve `__call` from a metatable to determine if a table is callable.

### DIFF
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -862,7 +862,7 @@ function vim.is_callable(f)
   if m == nil then
     return false
   end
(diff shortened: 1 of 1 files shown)


bfredl:
> LuaJIT uses a raw get to retrieve __call from a metatable to determine if a table is callable.

Also puc lua 5.1 (it is not luajit specific). As noted, test would be good.

tmillr:
Ok I added some tests. Sorry for the delay.

dundargoc:
The reopen button is disabled 🤷. I guess you need open a new PR.

PR #27460, opened 2024-02-13
refactor(eval): use arena when converting typvals to Object

bfredl:


src/nvim/mapping.c
```diff
@@ -2305,13 +2306,13 @@ void f_mapset(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
   LuaRef rhs_lua = LUA_NOREF;
   dictitem_T *callback_di = tv_dict_find(d, S_LEN("callback"));
   if (callback_di != NULL) {
-    Object callback_obj = vim_to_object(&callback_di->di_tv);
-    if (callback_obj.type == kObjectTypeLuaRef && callback_obj.data.luaref != LUA_NOREF) {
-      rhs_lua = callback_obj.data.luaref;
-      orig_rhs = "";
-      callback_obj.data.luaref = LUA_NOREF;
+    if (callback_di->di_tv.v_type == VAR_FUNC) {
```
zeertzjq, line 2309: `tv_is_func()`?

bfredl:
Note: this contains two a bit ugly but  _temporary_ changes which can be reverted
once the Arena vs no-Arena distinction in API wrappers has been removed.

Both nlua_push_Object and object_to_vim_take_luaref() has been changed
to take the object argument as a pointer. this is not going to be
necessary once these are only used with arena (or not at all) allocated
Objects.

src/nvim/api/private/converter.c
```diff
@@ -19,6 +19,8 @@
 /// Helper structure for vim_to_object
 typedef struct {
   kvec_withinit_t(Object, 2) stack;  ///< Object stack.
+  Arena *arena;  ///< arena where objects will be allocatod
```
zeertzjq, line 22:
```suggestion
  Arena *arena;  ///< arena where objects will be allocated
```

PR #27148, opened 2024-01-23
Add nil check in rpc.lua to handle nil json body

1337jazz:
When using Omnisharp LSP there seems to be occasions where the following error is encountered when the JSON body is null for whatever reason:

```
Error  10:19:49 msg_show.lua_error Error executing luv callback:
/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:386: attempt to index local 'decoded' (a nil value)
stack traceback:
	/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:386: in function 'handle_body'
	/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:726: in function 'handle_body'
	/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:260: in function </usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:244>
```

This PR aims to address in a simple (or perhaps naive?) way.

PR #27478, opened 2024-02-15
Check if json is decoded before continuing

djgalvan:
Problem:
`decoded` is used throughout `Client:hand_body(body)` without first checking if the json content was decoded first, leading to errors.

Solution:
Check that the json content was decoded before continuing.

Thoughts:
The json content is expected to be decoded throughout the function, the function is used while parsing in `create_read_loop`. Logging would lead to spam while waiting for the content to be decoded.

I'm looking over the commit guidelines and will update appropriately

### DIFF
--- a/runtime/lua/vim/lsp/rpc.lua
+++ b/runtime/lua/vim/lsp/rpc.lua
@@ -397,6 +397,9 @@ end
 ---@package
 function Client:handle_body(body)
   local ok, decoded = pcall(vim.json.decode, body, { luanil = { object = true } })
+  if not decoded then
+    return
+  end
   if not ok then
     self:on_error(M.client_errors.INVALID_SERVER_JSON, decoded)
     return


clason:
What problem does this solve? (Please always describe your PR -- and commit! -- in the form "Problem: ... Solution: ...". It's easy to make a three-line change; it's hard to reason about the consequences, especially three years down the line...)

mfussenegger:
See also https://github.com/neovim/neovim/pull/27148 

This is likely the wrong fix, but without explanation we can't tell.

djgalvan:
Improper commit

justinmk:
you can force-push to update this PR. that's preferred rather than creating a new one.

clason:
>  leading to errors.

What kind of errors? Can you be specific? Or is this "drive-by pessimization" (excuse the wording)?

djgalvan:
> > leading to errors.
> 
> What kind of errors? Can you be specific? Is there a concrete and reproducible example? Or is this "drive-by pessimization" (excuse the wording)?

Sure:

Error executing luv callback:
.../david/Local/nvim/share/nvim/runtime/lua/vim/lsp/rpc.lua:406: attempt to index local 'decoded' (a nil value)
stack traceback:
        .../david/Local/nvim/share/nvim/runtime/lua/vim/lsp/rpc.lua:406: in function 'handle_body'
        .../david/Local/nvim/share/nvim/runtime/lua/vim/lsp/rpc.lua:787: in function 'handle_body'
        .../david/Local/nvim/share/nvim/runtime/lua/vim/lsp/rpc.lua:260: in function <.../david/Local/nvim/share/nvim/runtime/lua/vim/lsp/rpc.lua:2
44>

I have no clue what "drive-by pessimization" means.


> you can force-push to update this PR. that's preferred rather than creating a new one.

I was trying that, but some more testing, it seemingly didn't fix the issue. Only present one MacOSx, I'm digging.

clason:
Can you also share _how_ you got the error?

djgalvan:
I would think any variation of:

```
if not decoded then
   return
end
```

```
if decoded == nil then
  return
end
```

```
if decoded == nil
   do return end
end
```

Should work, unsure of that last one. The function seemingly continues to run. There's something I'm not understanding with terminating the function after the check

djgalvan:
My config is [here](https://github.com/djgalvan/nvim) if you want to check. It's a mess, don't judge. That issue is closed so I'm focused on what's happening here.

djgalvan:
Error doesn't happen on Windows either. Something funky on MacOS, I'll leave that for more capable hands.

Issue #18909, opened 2022-06-09
Mounted paths in Windows (`\\Mac\Home\projects\distant`) cause invalid string values with LSP servers


chipsenkbeil:

chipsenkbeil:
I'm guessing that the mounted path is actually a network path on Windows.  `\\Mac\Home` is a specific directory mounted into my VM using Windows networking. So my guess is that any network path is going to fail with neovim. I tested both `rls` and `rust-analyzer` on VS Code using my network path and they work, so it is possible to have a language server work with a network path! 😄 

<img width="540" alt="image" src="https://user-images.githubusercontent.com/2481802/172864348-06d58355-47c5-4c3f-8898-d9ea5295ad4b.png">

FrancescoLuzzi:
`nvim version 0.9.2`
I've encountered the same issue on windows 10, but I want to add something to the context:
- opening a file directly, will attach the client (e.g. `nvim main.go`)
- opening the current folder in neovim then `explicitly` opening the file will attach the client (e.g. `nvim . -> :e main.go`)
- opening the current folder the opening a file with something like `nvim-tree`, won't attach the client correctly

I'll try to reproduce this behaviour on another windows machine😃

PR #27482, opened 2024-02-15
fix(uri): correctly convert windows UNC path to URI

FrancescoLuzzi:
Detect and convert windows UNC paths to URI and back
Related to #18909

### DIFF
--- a/runtime/lua/vim/uri.lua
+++ b/runtime/lua/vim/uri.lua
@@ -9,6 +9,12 @@ local schar = string.char
 local tohex = require('bit').tohex
 local URI_SCHEME_PATTERN = '^([a-zA-Z]+[a-zA-Z0-9.+-]*):.*'
 local WINDOWS_URI_SCHEME_PATTERN = '^([a-zA-Z]+[a-zA-Z0-9.+-]*):[a-zA-Z]:.*'
+-- https://learn.microsoft.com/en-us/troubleshoot/windows-server/identity/naming-conventions-for-computer-domain-site-ou
+-- added also "." to the regex to handle \\wsl.localhost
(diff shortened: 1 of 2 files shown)


justinmk:
Thanks for looking into this. https://github.com/neovim/neovim/pull/28203 added UNC support to normalize(). Can normalize() be used to simplify the code here?

runtime/lua/vim/uri.lua
```diff
@@ -110,10 +123,14 @@ function M.uri_to_fname(uri)
   end
   uri = M.uri_decode(uri)
   --TODO improve this.
-  if is_windows_file_uri(uri) then
+  if is_windows_volume_uri(uri) then
     uri = uri:gsub('^file:/+', ''):gsub('/', '\\')
   else
-    uri = uri:gsub('^file:/+', '/') ---@type string
+    if IS_WINDOWS and uri:match('^file://[a-zA-Z.-]') ~= nil then -- handle UNC file uri in windows
+      uri = uri:gsub('^file:/+', '//'):gsub('/', '\\')
```
justinmk, line 130: This is an example of logic that could be removed if we used `vim.normalize()` earlier in this function.

IOW: check for `file:` early in this function, and if true, use `vim.normalize()` on the rest of the path. What other logic would that allow us to remove, in uri.lua ?

FrancescoLuzzi:
Just to be sure, using `vim.fs.normalize` will change the output of `vim.uri_to_fname`:
all `\` will become `/`.
If this is intended, I'm off to the races.

FrancescoLuzzi:
I went with a more os aware approach and used `vim.fs.normalize`

Let me know if any changes are needed

PR #27484, opened 2024-02-15
docs: document breaking change for nvim_create_autocmd callback

gpanders:
https://github.com/neovim/neovim/pull/27428 changed the semantics of callbacks passed to nvim_buf_attach and nvim_create_autocmd such that any truthy value will detach the callback/delete the autocommand (rather than just the literal boolean value `true`). Update the documentation accordingly and add an entry to `news.txt`.

cc @bfredl

### DIFF
--- a/src/nvim/api/buffer.c
+++ b/src/nvim/api/buffer.c
@@ -125,7 +125,7 @@ Integer nvim_buf_line_count(Buffer buffer, Error *err)
 ///             - on_bytes: Lua callback invoked on change.
 ///               This callback receives more granular information about the
 ///               change compared to on_lines.
-///               Return `true` to detach.
+///               Return a truthy value (not `false` or `nil`) to detach.
 ///               Args:
 ///               - the string "bytes"
 ///               - buffer handle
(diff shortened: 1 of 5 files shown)


bfredl:
it should be changed back (and tested)

bfredl:
after a bit more discussion, the past behavior of `nvim_create_autocmd` was actually inconsistent with `nvim_buf_attach` and similar functions. perhaps we should rip of the bandaid and accept a breaking change to make it consistent (but still document it, of course).

gpanders:
I was mistaken that this is a breaking change for `nvim_buf_attach`: this is in fact how those callbacks have always worked. It is only `nvim_create_autocmd()` that has had its behavior changed.

The fact that `nvim_create_autocmd` behaved differently from `nvim_buf_attach` in the first place was not intentional so can be considered a bug. Making the two consistent with each other is a good fix. I've updated the news entry accordingly.

src/nvim/api/buffer.c
```diff
@@ -125,7 +125,7 @@ Integer nvim_buf_line_count(Buffer buffer, Error *err)
 ///             - on_bytes: Lua callback invoked on change.
 ///               This callback receives more granular information about the
 ///               change compared to on_lines.
-///               Return `true` to detach.
+///               Return a truthy value (not `false` or `nil`) to detach.
```
zeertzjq, line 128: There is another mention at `on_lines`

gpanders: 73c87c5aead2c21eec0889f9d86952a0bee4f8aa

PR #27495, opened 2024-02-16
build: set deps default build type to Release

gpanders:
Debugging dependencies is rare so a Debug build type is usually not needed. In cases where it _is_ needed it is easy to rebuild in Debug mode. But since Release builds are more common, it makes more sense as a default.

For Neovim itself we stick with a Debug build as a default, since rebuilding and debugging is done _much_ more frequently than with dependencies (which we _mostly_ expect to "just work").

### DIFF
--- a/cmake/Util.cmake
+++ b/cmake/Util.cmake
@@ -148,8 +148,8 @@ function(add_glob_target)
   add_custom_target(${ARG_TARGET} DEPENDS ${touch_list})
 endfunction()
 
-# Set default build type to Debug. Also limit the list of allowable build types
-# to the ones defined in variable allowableBuildTypes.
+# Set default build type to BUILD_TYPE. Also limit the list of allowable build
+# types to the ones defined in variable allowableBuildTypes.
 #
 # The correct way to specify build type (for example Release) for
 # single-configuration generators (Make and Ninja) is to run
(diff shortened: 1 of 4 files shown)


dundargoc (approved):
Looks good. Nice.

cmake/Util.cmake
```diff
@@ -165,21 +165,22 @@ endfunction()
 #
 # Passing CMAKE_BUILD_TYPE for multi-config generators will not only not be
 # used, but also generate a warning for the user.
-function(set_default_buildtype)
-  set(allowableBuildTypes Debug Release MinSizeRel RelWithDebInfo)
+function(set_default_buildtype BUILD_TYPE)
+  set(allowableBuildTypes ${BUILD_TYPE} Debug Release MinSizeRel RelWithDebInfo)
```
dundargoc, line 169: Bit unsure, should we maybe write a small comment that this is for multi-config generators (since the default behavior for them is to use the first item in `CMAKE_CONFIGURATION_TYPES`)? Up to you, just wanted to mention it.

gpanders: What if we only modify `allowableBuildTypes` in the `if(isMultiConfig)` branch?

dundargoc: Hmm, that could work. Good thinking.

cmake/Util.cmake
```diff
@@ -165,21 +165,27 @@ endfunction()
...
     endif()
   else()
     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "${allowableBuildTypes}")
     if(NOT CMAKE_BUILD_TYPE)
-      message(STATUS "CMAKE_BUILD_TYPE not specified, default is 'Debug'")
-      set(CMAKE_BUILD_TYPE Debug CACHE STRING "Choose the type of build" FORCE)
+      message(STATUS "CMAKE_BUILD_TYPE not specified, default is '${BUILD_TYPE}'")
+      set(CMAKE_BUILD_TYPE ${BUILD_TYPE} CACHE STRING "Choose the type of build" FORCE)
     elseif(NOT CMAKE_BUILD_TYPE IN_LIST allowableBuildTypes)
       message(FATAL_ERROR "Invalid build type: ${CMAKE_BUILD_TYPE}")
```
dundargoc, line 190: Hmm, is this still needed?

gpanders: This check happens if someone sets `CMAKE_BUILD_TYPE` themselves, e.g. `make CMAKE_BUILD_TYPE=Blorf`

dundargoc: Wait, you're right. OK thanks for checking.

dundargoc (approved):

zeertzjq:
Won't this prevent stacktrace across LuaJIT/Luv/libvterm boundary?

gpanders:
It only affects the default. If you need stack traces you can still build with RelWithDebInfo or Debug.

Presumably people were not using the default Debug build type before because of how slow it is, so if they were setting the build type to RelWithDebInfo anyway they can still do so.

zeertzjq:
Then what is the point of setting deps default build type to Release if most people are going to override it anyway?

zeertzjq:
Oh, actually the stacktraces still work. They just have `??` in functions from deps.

gpanders:
> Then what is the point of setting deps default build type to Release if most people are going to override it anyway?

The reasons I explained in the commit message. `Release` is more common for deps than `Debug`. Most people override it because the current default is bad. With a better default, fewer people will need to override it.

PR #27501, opened 2024-02-16
fix(api): don't use stl 'fillchar' for "use_statuscol_lnum"

luukvbaal:
Problem: nvim_eval_statusline() uses "stl" from 'fillchars' with "use_statuscol_lnum".
Solution: Re-order fillchar else chain.

### DIFF
--- a/test/functional/api/vim_spec.lua
+++ b/test/functional/api/vim_spec.lua
@@ -3806,7 +3806,8 @@ describe('API', function()
       it('works with statuscolumn', function()
         exec([[
           let &stc='%C%s%=%l '
-          set cul nu nuw=3 scl=yes:2 fdc=2
+          " should not use "stl" from 'fillchars'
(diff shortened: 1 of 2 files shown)


luukvbaal:
Pardon the somewhat unnecessary churn but I think the else chain makes more sense like this.

test/functional/api/vim_spec.lua
```diff
@@ -3806,7 +3806,7 @@ describe('API', function()
       it('works with statuscolumn', function()
         exec([[
           let &stc='%C%s%=%l '
-          set cul nu nuw=3 scl=yes:2 fdc=2
+          set cul nu nuw=3 scl=yes:2 fdc=2 fillchars=stl:#
```
zeertzjq, line 3809: Add a comment above?

zeertzjq (approved):

zeertzjq (approved):

b0o:
Thank you @luukvbaal!

PR #27276, opened 2024-01-31
ci: also test on macOS M1

clason:
GH Actions run macos-14 on M1 now, giving us our first ARM build.

Test both arches for now; we can eventually drop Intel in favor of ARM to reduce the test load (and will have to, once Apple drops support for Intel completely). M1 runners seem about twice as fast, at least at this moment:

* macos-12:
  - Build 48s
  - functionaltest 7m 7s

* macos-14:
  - Build 24s
  - functionaltest 5m 5s

unittest failure is relevant and due to arch difference, see #26145 

(oldttest failure is also relevant and due to building without `libintl`: one is straight locale failure (though behind `has('gettext')` guard?), the other is in `Wildmode_test()` and unclear)

**Question:** Do we want to move the build/release tests to M1 (once we get the unittest fixed)?

@dundargoc

PR #27505, opened 2024-02-17
build(macos): release workflow on M1

clason:
Follow-up to #27276 

Run the release workflow on macos-14 to use faster M1 runners.

Lock the deployment target to the oldest supported version (11.0,
due to libuv support) instead of relying on the host OS version.


clason:
https://github.com/clason/neovim/actions/runs/7941606470

dundargoc (approved):
Looks good. Bonus points for testing the workflow.

PR #27510, opened 2024-02-17
fix(lsp): support workspace/configuation with no section

tpope:
The [spec for `workspace/configuration`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) marks the `section` property of each item in `items` optional. Therefore, I believe it violates the spec to skip over items without a section, because then the length of the results won't match the length of a valid `items` input. The spec does not elaborate on _what_ to return in this case, but I don't think it would be controversial to say that returning the full configuration, as done for an empty string, is the most natural interpretation.

That empty string case, by the way, was initially [added in response](https://github.com/neovim/nvim-lspconfig/commit/5da124fc82c24592abc3a97c68363bb4e85e3c85) to a real world implementation requesting it. I don't have a similar real world implementation to point to for the omitted `section`, but I would note that `getConfiguration()` from `vscode-languageserver-node` [defaults to a request with no section](https://github.com/Microsoft/vscode-languageserver-node/blob/d859bb14d1bcb3923eecaf0ef587e55c48502ccc/server/src/common/configuration.ts#L24-L26) when called with no arguments. I surmise that this is intended as a way to retrieve the full configuration.

### DIFF
--- a/runtime/lua/vim/lsp/handlers.lua
+++ b/runtime/lua/vim/lsp/handlers.lua
@@ -206,6 +206,9 @@ M[ms.workspace_configuration] = function(_, result, ctx)
         value = vim.NIL
       end
       table.insert(response, value)
+    else
+      -- If no section is provided, return settings as is
+      table.insert(response, client.settings)
     end
   end
   return response
--- a/test/functional/fixtures/fake-lsp-server.lua
+++ b/test/functional/fixtures/fake-lsp-server.lua
@@ -149,9 +149,16 @@ function tests.check_workspace_configuration()
           { section = 'testSetting2' },
           { section = 'test.Setting3' },
           { section = 'test.Setting4' },
+          {},
+          { section = '' },
         },
       })
-      expect_notification('workspace/configuration', { true, false, 'nested', vim.NIL })
+      local all = {
+        testSetting1 = true,
+        testSetting2 = false,
+        test = { Setting3 = 'nested' },
+      }
+      expect_notification('workspace/configuration', { true, false, 'nested', vim.NIL, all, all })
       notify('shutdown')
     end,
   }
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -502,6 +502,8 @@ describe('LSP', function()
               { section = 'testSetting2' },
               { section = 'test.Setting3' },
               { section = 'test.Setting4' },
+              {},
+              { section = '' },
             },
           },
           { method = 'workspace/configuration', client_id = 1 },


tpope:
Forgive me, accidentally skipped the "draft" step to iron out the checks. This should be ready to go now.

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -196,7 +196,7 @@ M[ms.workspace_configuration] = function(_, result, ctx)
 
   local response = {}
   for _, item in ipairs(result.items) do
-    if item.section then
+    if item.section and item.section ~= vim.NIL then
```
MariaSolOs, line 199: I don't think that we need to check for `~= vim.NIL`, since `section: nil` isn't allowed by the protocol.

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -206,6 +206,9 @@ M[ms.workspace_configuration] = function(_, result, ctx)
         value = vim.NIL
       end
       table.insert(response, value)
+    else
+      -- If no section is provided, return settings as is
```
MariaSolOs, line 210: I think this is reasonable. However, I do believe that we should create an issue [upstream](https://github.com/microsoft/language-server-protocol) asking for this clarification to be made on the specs.

MariaSolOs (approved):

runtime/lua/vim/lsp/handlers.lua
tpope: It is indeed unnecessary for spec compliance. I was just trying to be a bit more robust the face of wonky JSON serialization.

MariaSolOs: I don't think such robustness is needed ;) Besides, this check would support misbehaving clients.

bfredl: IIRC, we will delete JSON keys that are `null` in dictinonaries ( due to ` pcall(vim.json.decode, body, { luanil = { object = true } }` in `lsp/rpc.lua` ), so we don't need to check for `vim.NIL` explicitly regardless.

MariaSolOs:
> I don't have a similar real world implementation to point to for the omitted `section`,

[Here](https://github.com/microsoft/vscode-eslint/blob/89d07f2eee7e363b90bc6b69ea41e5f8bdb55e2f/client/src/client.ts#L602) is one from `vscode-eslint`. The middleware handler actually skips those items with a section defined.

tpope:
> > I don't have a similar real world implementation to point to for the omitted `section`,
> 
> [Here](https://github.com/microsoft/vscode-eslint/blob/89d07f2eee7e363b90bc6b69ea41e5f8bdb55e2f/client/src/client.ts#L602) is one from `vscode-eslint`. The middleware handler actually skips those items with a section defined.

Good find. I would not the intent of this client code appears to be to implement just enough of the spec to appease the [corresponding server code](https://github.com/microsoft/vscode-eslint/blob/be30e933c56ea6e8e579808d5f7c7b205c8e16cc/server/src/eslintServer.ts#L668). And that server code isn't omitting the section, it's sending `''`. So I don't know that we should read to much into it, as much as I would like to.

runtime/lua/vim/lsp/handlers.lua
mfussenegger: +1, see also https://github.com/microsoft/language-server-protocol/issues/972

danielwe:
> I don't have a similar real world implementation to point to for the omitted `section`

One example is JETLS, a new language server for Julia: https://github.com/aviatesk/JETLS.jl/blob/a451764a903328d504abf95e926d283a2525a166/src/workspace-configuration.jl#L53

Relevant discussion and open PR on their end: aviatesk/JETLS.jl#482 and aviatesk/JETLS.jl#483.

See also the Helix client for an implementation aligned with this PR, returning the full config when `section` is omitted: https://github.com/helix-editor/helix/blob/99ebccbfc5f6b414e442e72ca6ce4fd75d2d8f0b/helix-term/src/application.rs#L1033-L1041

I appreciate the desire for a clarification from upstream, but in the meantime, isn't this the most plausible interpretation? If the server sent a configuration request, surely it would like to receive some configuration.

lewis6991 (approved):

MariaSolOs (approved):

runtime/lua/vim/lsp/handlers.lua
justinmk: > we will delete JSON keys that are `null` in objects ( due to ` pcall(vim.json.decode, body, { luanil = { object = true } }` in `lsp/rpc.lua` ), so we don't need to check for `vim.NIL` explicitly regardless

For reference, that was dropped in https://github.com/neovim/neovim/pull/34849, "json null" now sets `vim.NIL`.

PR #27511, opened 2024-02-17
refactor(api): use arena for mappings, autocmd, channel info

bfredl:


src/nvim/mapping.c
```diff
@@ -2396,14 +2395,14 @@ void f_maplist(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
...
         xfree(lhs);
 
-        Dictionary dict = mapblock_fill_dict(mp,
-                                             did_simplify ? keys_buf : NULL,
-                                             buffer_local, abbr, true);
+        Arena arena = ARENA_EMPTY;
+        Dictionary dict = mapblock_fill_dict(mp, did_simplify ? keys_buf : NULL,
+                                             buffer_local, abbr, true, &arena);
         typval_T d = TV_INITIAL_VALUE;
         object_to_vim(DICTIONARY_OBJ(dict), &d, NULL);
```
zeertzjq, line 2402: Cannot use `object_to_vim_take_luaref()` here?

bfredl: done

src/nvim/api/private/helpers.c
```diff
@@ -581,6 +581,15 @@ String arena_string(Arena *arena, String str)
   }
 }
 
+Array arena_take_arraybuilder(Arena *arena, ArrayBuilder *arr)
+{
+  Array ret = arena_array(arena, kv_size(*arr));
+  ret.size = kv_size(*arr);
+  memcpy((ret).items, (*arr).items, sizeof((ret).items[0]) * (ret).size);
```
zeertzjq, line 588:
```suggestion
  memcpy(ret.items, (*arr).items, sizeof(ret.items[0]) * ret.size);
```

PR #27522, opened 2024-02-18
feat(tabline): middle mouse button now closes tab

### LINKED ISSUE #20372: Close a tab via middle click
### Feature already in Vim?

Yes, Vim 9.0.0597

### Feature description

vim/vim@3bdef10dc1c836e5e5a6fdcc1c21155cbd80d798 added support for closing tabs using the middle mouse button.
That seems like a convenient feature, especially for closing background tabs.

NachoNievaG:
Fix #20372

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -316,6 +316,8 @@ The following new APIs and features were added.
 
 • Added |nvim_tabpage_set_win()| to set the current window of a tabpage.
 
+• Clicking on a tabpage in the tabline with the middle mouse button closes it.
+
 ==============================================================================
 CHANGED FEATURES                                                 *news-changed*
 
--- a/src/nvim/mouse.c
+++ b/src/nvim/mouse.c
@@ -136,6 +136,26 @@ static void move_tab_to_mouse(void)
     tabpage_move(tabnr);
   }
 }
+/// Close the current or specified tab page.
+///
+/// @param c1  tabpage number, or 999 for the current tabpage
+static void mouse_tab_close(int c1)
+{
+  tabpage_T *tp;
+
+  if (c1 == 999) {
+    tp = curtab;
+  } else {
+    tp = find_tabpage(c1);
+  }
+  if (tp == curtab) {
+    if (first_tabpage->tp_next != NULL) {
+      tabpage_close(false);
+    }
+  } else if (tp != NULL) {
+    tabpage_close_other(tp, false);
+  }
+}
 
 static bool got_click = false;  // got a click some time back
 
@@ -487,43 +507,32 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
       if (is_click && cmdwin_type == 0 && mouse_col < Columns) {
         in_tab_line = true;
         c1 = tab_page_click_defs[mouse_col].tabnr;
+
         switch (tab_page_click_defs[mouse_col].type) {
         case kStlClickDisabled:
           break;
-        case kStlClickTabClose: {
-          tabpage_T *tp;
-
-          // Close the current or specified tab page.
-          if (c1 == 999) {
-            tp = curtab;
-          } else {
-            tp = find_tabpage(c1);
-          }
-          if (tp == curtab) {
-            if (first_tabpage->tp_next != NULL) {
-              tabpage_close(false);
-            }
-          } else if (tp != NULL) {
-            tabpage_close_other(tp, false);
-          }
-          break;
-        }
         case kStlClickTabSwitch:
-          if ((mod_mask & MOD_MASK_MULTI_CLICK) == MOD_MASK_2CLICK) {
-            // double click opens new page
-            end_visual_mode();
-            tabpage_new();
-            tabpage_move(c1 == 0 ? 9999 : c1 - 1);
-          } else {
-            // Go to specified tab page, or next one if not clicking
-            // on a label.
-            goto_tabpage(c1);
-
-            // It's like clicking on the status line of a window.
-            if (curwin != old_curwin) {
+          if (which_button != MOUSE_MIDDLE) {
+            if ((mod_mask & MOD_MASK_MULTI_CLICK) == MOD_MASK_2CLICK) {
+              // double click opens new page
               end_visual_mode();
+              tabpage_new();
+              tabpage_move(c1 == 0 ? 9999 : c1 - 1);
+            } else {
+              // Go to specified tab page, or next one if not clicking
+              // on a label.
+              goto_tabpage(c1);
+
+              // It's like clicking on the status line of a window.
+              if (curwin != old_curwin) {
+                end_visual_mode();
+              }
             }
+            break;
           }
+          FALLTHROUGH;
+        case kStlClickTabClose:
+          mouse_tab_close(c1);
           break;
         case kStlClickFuncRun:
           call_click_def_func(tab_page_click_defs, mouse_col, which_button);
--- a/test/functional/ui/tabline_spec.lua
+++ b/test/functional/ui/tabline_spec.lua
@@ -184,4 +184,18 @@ describe('tabline', function()
     ]],
     }
   end)
+
+  it('middle-click closes tab', function()
+    command('tabnew')
+    command('tabnew')
+    command('tabnew')
+    command('tabprev')
+    eq({ 3, 4 }, api.nvim_eval('[tabpagenr(), tabpagenr("$")]'))
+    api.nvim_input_mouse('middle', 'press', '', 0, 0, 1)
+    eq({ 2, 3 }, api.nvim_eval('[tabpagenr(), tabpagenr("$")]'))
+    api.nvim_input_mouse('middle', 'press', '', 0, 0, 20)
+    eq({ 2, 2 }, api.nvim_eval('[tabpagenr(), tabpagenr("$")]'))
+    api.nvim_input_mouse('middle', 'press', '', 0, 0, 1)
+    eq({ 1, 1 }, api.nvim_eval('[tabpagenr(), tabpagenr("$")]'))
+  end)
 end)


test/functional/ui/tabline_spec.lua
```diff
@@ -184,4 +184,29 @@ describe('tabline', function()
     ]],
     }
   end)
+
+  it('middle closes tab', function()
```
justinmk, line 188:
```suggestion
  it('middle-click closes tab', function()
```

test/functional/ui/tabline_spec.lua
```diff
@@ -184,4 +184,29 @@ describe('tabline', function()
...
+    insert('tab1')
+    command('tabnew')
+    insert('tab2')
+    command('tabprev')
+    api.nvim_set_option_value('tabline', '%1T口口%2Ta' .. ('b'):rep(38) .. '%999Xc', {})
```
justinmk, line 193: is this relevant to this test?

src/nvim/mouse.c
```diff
@@ -412,6 +412,18 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
 
   // Middle mouse button does a 'put' of the selected text
   if (which_button == MOUSE_MIDDLE) {
+    // handled outside any specific mode
```
justinmk, line 415: why? (the comment should give a hint about why, also)

src/nvim/mouse.c
```diff
@@ -412,6 +412,18 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
...
+      if (mouse_grid <= 1 && mouse_row == 0 && firstwin->w_winrow > 0) {
+        if (is_click && cmdwin_type == 0 && mouse_col < Columns) {
+          in_tab_line = true;
+          c1 = tab_page_click_defs[mouse_col].tabnr;
+          handle_tabclose(c1);
+        }
+        return false;
+      }
+    }
     if (State == MODE_NORMAL) {
```
justinmk, line 427: use a blank line to separate this "group" of code, just like the other cases in this same function

src/nvim/mouse.c
```diff
@@ -412,6 +412,18 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
 
   // Middle mouse button does a 'put' of the selected text
   if (which_button == MOUSE_MIDDLE) {
+    // handled outside any specific mode
+    if (tab_page_click_defs != NULL) {  // only when initialized
```
justinmk, line 416: can/should this live in the other block that checks this same condition, on line 476 (new: 488)

src/nvim/mouse.c
```diff
@@ -412,6 +412,18 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
...
+      // Check for clicking in the tab page line.
+      if (mouse_grid <= 1 && mouse_row == 0 && firstwin->w_winrow > 0) {
```
justinmk, line 418: can/should this live in the other block that checks this same condition(s), on line 476 (new: 488)

src/nvim/mouse.c
```diff
@@ -966,6 +964,25 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
   return moved;
 }
 
+void handle_tabclose(int c1)
```
justinmk, line 967: use naming conventions that match surrounding code, if not the entire codebase
```suggestion
void do_tabclose(int c1)
```

src/nvim/mouse.c
NachoNievaG: I understand that it's supposed to close the tab, despite the current mode.
AFAIK there is no specific behavior regarding middle mouse clicking on a tab in a specific mode.
Please, correct me if I'm wrong :)

src/nvim/mouse.c
justinmk: why not move this code: https://github.com/neovim/neovim/blob/ce5a9bfe7e537c81d34bd4a27fc6638f20114e67/src/nvim/mouse.c#L413-L414

_after_ this code:
https://github.com/neovim/neovim/blob/ce5a9bfe7e537c81d34bd4a27fc6638f20114e67/src/nvim/mouse.c#L476

and put the middle-click handling in there? special-casing stuff in random places makes the logic very hard to follow.

src/nvim/mouse.c
justinmk: `<MiddleMouse>` can have mode-specific mappings like any other keycode. So I don't see why it ignore modes. OTOH, this might be moot, see https://github.com/neovim/neovim/pull/27522#discussion_r1493824043

NachoNievaG: Oh, i see.
Then it would be better to enclose the "click on tab with middle mouse to close" scenario, so it would be still be able to work on mode-specific mappings.
Great catch, thank you.

test/functional/ui/tabline_spec.lua
NachoNievaG: It is not, I believe I found a better way to test this.

NachoNievaG:
@justinmk I manage to reduce the amount of failing tests (8 to 2) given that the mod keys + middle click experience was affected.
Yet I can manage to find a way to fix this test:
https://github.com/neovim/neovim/blob/ce5a9bfe7e537c81d34bd4a27fc6638f20114e67/test/functional/ui/mouse_spec.lua#L502
Also the test workflow is a bit confusing, so I'm not entirely sure what the test scope is supposed to be.
Do you have any context on this?
I believe every other scenario is fixed.
thanks!

src/nvim/mouse.c
```diff
@@ -487,25 +487,22 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
...
         c1 = tab_page_click_defs[mouse_col].tabnr;
+
+        // Check for modifiers when clicking the tab line.
+        // Middle mouse button in the tab line: Close the tab page.
+        if (which_button == MOUSE_MIDDLE
+            && !(mod_mask &
+                 (MOD_MASK_MULTI_CLICK | MOD_MASK_SHIFT | MOD_MASK_CTRL | MOD_MASK_ALT |
+                  MOD_MASK_META))) {
+          do_tabclose(c1);
+          return false;
```
seandewar, line 498:
```suggestion
          return true;
```

Nit: doesn't matter here since middle mouse in insert mode (where `start_arrow()` applies) puts instead, but maybe it's best to stay consistent with the other code that returns `true` below?

seandewar:
> Also the test workflow is a bit confusing, so I'm not entirely sure what the test scope is supposed to be.

That test is checking whether `%@` items in a custom `'tabline'` ("execute function labels" in `:h 'stl'`) receive the middle mouse click events. Because your changes close the tabpage when that happens, the label isn't receiving the event anymore.

zeertzjq (changes requested):
This can be simply achieved using a fallthrough:
```diff
diff --git a/src/nvim/mouse.c b/src/nvim/mouse.c
index f4d1be5b2f..07bde24ec6 100644
--- a/src/nvim/mouse.c
+++ b/src/nvim/mouse.c
@@ -490,6 +490,26 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
         switch (tab_page_click_defs[mouse_col].type) {
         case kStlClickDisabled:
           break;
+        case kStlClickTabSwitch:
+          if (which_button != MOUSE_MIDDLE) {
+            if ((mod_mask & MOD_MASK_MULTI_CLICK) == MOD_MASK_2CLICK) {
+              // double click opens new page
+              end_visual_mode();
+              tabpage_new();
+              tabpage_move(c1 == 0 ? 9999 : c1 - 1);
+            } else {
+              // Go to specified tab page, or next one if not clicking
+              // on a label.
+              goto_tabpage(c1);
+
+              // It's like clicking on the status line of a window.
+              if (curwin != old_curwin) {
+                end_visual_mode();
+              }
+            }
+            break;
+          }
+          FALLTHROUGH;
         case kStlClickTabClose: {
           tabpage_T *tp;
 
@@ -508,23 +528,6 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
           }
           break;
         }
-        case kStlClickTabSwitch:
-          if ((mod_mask & MOD_MASK_MULTI_CLICK) == MOD_MASK_2CLICK) {
-            // double click opens new page
-            end_visual_mode();
-            tabpage_new();
-            tabpage_move(c1 == 0 ? 9999 : c1 - 1);
-          } else {
-            // Go to specified tab page, or next one if not clicking
-            // on a label.
-            goto_tabpage(c1);
-
-            // It's like clicking on the status line of a window.
-            if (curwin != old_curwin) {
-              end_visual_mode();
-            }
-          }
-          break;
         case kStlClickFuncRun:
           call_click_def_func(tab_page_click_defs, mouse_col, which_button);
           break;
```

src/nvim/mouse.c
```diff
@@ -1133,6 +1122,24 @@ bool is_mouse_key(int c)
          || c == K_X2RELEASE;
 }
 
+static void do_tabclose(int c1)
```
zeertzjq, line 1125: This function name is confusing as there is already a `:tabclose` Ex command. I think the function name should contain the word "mouse". Its position in the file is also strange. I think it should be after `move_tab_to_mouse()`.

runtime/doc/news.txt
```diff
@@ -316,6 +316,8 @@ The following new APIs and features were added.
 
 • Added |nvim_tabpage_set_win()| to set the current window of a tabpage.
 
+• Middle mouse button closes tabs.
```
zeertzjq, line 319: Middle mouse button on what?

src/nvim/mouse.c
NachoNievaG: Agree, would `mouse_tab_close()`  be more suitable?

runtime/doc/news.txt
zeertzjq, line 319: Middle mouse button clicking on what?

src/nvim/mouse.c
zeertzjq: Yes, sounds good.

test/functional/ui/tabline_spec.lua
```diff
@@ -184,4 +184,17 @@ describe('tabline', function()
     ]],
     }
   end)
+  it('middle-click closes tab', function()
```
zeertzjq, line 187:
```suggestion

  it('middle-click closes tab', function()
```

runtime/doc/news.txt
NachoNievaG: `• Clicking on a tab with the middle mouse button closes it.`?

zeertzjq:
```suggestion
• Clicking on a tabpage in the tabline with the middle mouse button closes it.
```

src/nvim/mouse.c
```diff
@@ -1133,6 +1122,24 @@ bool is_mouse_key(int c)
          || c == K_X2RELEASE;
 }
 
+static void mouse_tab_close(int c1)
```
zeertzjq, line 1125: I think this function should be moved below `move_tab_to_mouse()` so that it is grouped with more similar functions.

zeertzjq (approved):

src/nvim/mouse.c
```diff
@@ -137,6 +137,24 @@ static void move_tab_to_mouse(void)
   }
 }
 
+static void mouse_tab_close(int c1)
```
zeertzjq, line 140:
```suggestion
/// Close the current or specified tab page.
static void mouse_tab_close(int c1)
```

PR #27523, opened 2024-02-18
feat(treesitter): add lang arg to :InspectTree

ribru17:
Gives `:InspectTree` an optional argument representing a specific parser to show in the tree, for easier debugging. I'm not sure how to add proper documentation for this, it seems the documentation for the command itself is not fully complete, as it doesn't show that it takes a count.

### DIFF
--- a/runtime/plugin/nvim.lua
+++ b/runtime/plugin/nvim.lua
@@ -7,17 +7,26 @@ vim.api.nvim_create_user_command('Inspect', function(cmd)
 end, { desc = 'Inspect highlights and extmarks at the cursor', bang = true })
 
 vim.api.nvim_create_user_command('InspectTree', function(cmd)
+  ---@type string
+  local lang = cmd.fargs and cmd.fargs[1]
   if cmd.mods ~= '' or cmd.count ~= 0 then
     local count = cmd.count ~= 0 and cmd.count or ''
     local new = cmd.mods ~= '' and 'new' or 'vnew'
 
     vim.treesitter.inspect_tree({
       command = ('%s %s%s'):format(cmd.mods, count, new),
+      lang = lang,
     })
   else
-    vim.treesitter.inspect_tree()
+    vim.treesitter.inspect_tree({
+      lang = lang,
+    })
   end
-end, { desc = 'Inspect treesitter language tree for buffer', count = true })
+end, {
+  desc = 'Inspect treesitter language tree for buffer',
+  count = true,
+  nargs = '?',
+})
 
 vim.api.nvim_create_user_command('EditQuery', function(cmd)
   vim.treesitter.query.edit(cmd.fargs[1])


runtime/plugin/nvim.lua
```diff
@@ -7,17 +7,26 @@ vim.api.nvim_create_user_command('Inspect', function(cmd)
...
 
     vim.treesitter.inspect_tree({
       command = ('%s %s%s'):format(cmd.mods, count, new),
+      lang = lang,
     })
   else
-    vim.treesitter.inspect_tree()
+    vim.treesitter.inspect_tree({
+      lang = lang,
+    })
```
MariaSolOs, line 23: This could be simplified to:
```lua
  if cmd.mods ~= '' or cmd.count ~= 0 then
    local count = cmd.count ~= 0 and cmd.count or ''
    local new = cmd.mods ~= '' and 'new' or 'vnew'

    vim.treesitter.inspect_tree({
      command = ('%s %s%s'):format(cmd.mods, count, new),
      lang = cmd.fargs and cmd.fargs[1],
    })
  else
    vim.treesitter.inspect_tree()
```

ribru17: Hmm... in my testing I couldn't get this to work; something like `:InspectTree markdown_inline` will still just show the regular markdown tree

clason: Markdown is special; you'll _never_ have `markdown_inline` root tree.

gpanders:
What is the motivating use case for this? You mentioned "easier debugging", could you elaborate?

When `:InspectTree` is used in a buffer, that buffer already has a filetype. What does specifying an arbitrary language buy us? I tested this out just now by running `:InspectTree lua` in a C buffer and, predictably, the results were not useful.

clason:
The main use (for me) would be for languages whose name differs from the filetype (e.g., `latex` parser for `tex` filetypes), but arguably that should be fixed by adding smarter language resolution.

There _may_ be a use case for only showing injected (or root) trees, so, e.g., `:InspectTree vimdoc` on a help file would omit injected languages. But I posit this is rare enough not to need syntactic sugar.

(That being said, this PR seems harmless, as it simply exposes the existing underlying option in a straightforward manner.)

dundargoc:
Could this not be achieved by changing the filetype of the current buffer instead?

clason:
Not in my case, since `latex` is not a valid filetype.

dundargoc:
Oh. I forgot parsers and vim filetypes doesn't necessarily need to have parity. Thanks.

gpanders:
> (That being said, this PR seems harmless, as it simply exposes the existing underlying option in a straightforward manner.)

My concern is that we will lock in the argument to the `:InspectTree` command which _could_ be used for a different, more useful, as yet unknown use case in the future.

`vim.treesitter.inspect_tree()` is exposed intentionally to allow this kind of flexibility. Is overriding the language needed often enough to warrant it being part of the command itself?

If we made it an optional argument e.g. `:InspectTree ++lang={lang}` or similar, my reservations would be addressed, but that is of course a less simple change.

clason:
> vim.treesitter.inspect_tree() is exposed intentionally to allow this kind of flexibility. Is overriding the language needed often enough to warrant it being part of the command itself?

And does it need to be _manually_, instead making `:InspectTree`/`inspect_tree()` smarter and using the improved language resolution methods? I think this should be done first, in any case.

ribru17:
> And does it need to be _manually_, instead of making `:InspectTree`/`inspect_tree()` smarter and using the improved language resolution methods? I think this should be done first, in any case.

Shall I close in favor of this?

clason:
If you want to tackle this, that would be great!

ribru17:
Challenge accepted

PR #27530, opened 2024-02-19
test: add test case validate completed_item when new leader add

glepnir:
Problem: missing test case the completed_item validate when new leader.

Solution: add a test case.

### DIFF
--- a/test/functional/editor/completion_spec.lua
+++ b/test/functional/editor/completion_spec.lua
@@ -1052,21 +1052,21 @@ describe('completion', function()
 
   it('set completed_item correct when new leader add', function()
     source([[
-        funct Omni_test(findstart, base)
-          if a:findstart
-            return col(".")
-          endif
-          return [#{word: "one"}, #{word: "two"}, #{word: "five"}]
-        endfunc
-        set omnifunc=Omni_test
-        set complete=.
-        function! OnChange()
-          let g:event = copy(v:event)
-          let g:item = get(v:event, 'completed_item', {})
-          let g:word = get(g:item, 'word', v:null)
-        endfunction
-        autocmd CompleteChanged * call OnChange()
-      ]])
+      funct Omni_test(findstart, base)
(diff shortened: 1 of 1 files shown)


test/functional/editor/completion_spec.lua
```diff
@@ -1050,6 +1050,47 @@ describe('completion', function()
...
+        endfunc
+        set omnifunc=Omni_test
+        set complete=.
+        function! OnChange()
+          let g:event = copy(v:event)
+          let g:item = get(v:event, 'completed_item', {})
+          let g:word = get(g:item, 'word', v:null)
+        endfunction
+        autocmd CompleteChanged * call OnChange()
+      ]])
```
zeertzjq, line 1069:
```suggestion
      funct Omni_test(findstart, base)
        if a:findstart
          return col(".")
        endif
        return [#{word: "one"}, #{word: "two"}, #{word: "five"}]
      endfunc
      set omnifunc=Omni_test
      set complete=.
      function! OnChange()
        let g:event = copy(v:event)
        let g:item = get(v:event, 'completed_item', {})
        let g:word = get(g:item, 'word', v:null)
      endfunction
      autocmd CompleteChanged * call OnChange()
    ]])
```

glepnir: seems like can't use stylua on test file T.T

test/functional/editor/completion_spec.lua
```diff
@@ -1050,6 +1050,47 @@ describe('completion', function()
     feed('<esc>')
   end)
 
+  it('set completed_item correct when new leader add', function()
```
zeertzjq, line 1053:
```suggestion
  it('sets completed_item correctly when adding new leader', function()
```

zeertzjq:
Isn't this the same as the test in #27332?

glepnir:
same. but lua with screen expect. don't needed ?

PR #27537, opened 2024-02-19
fix(runtime): validate $VIMRUNTIME on startup

### LINKED ISSUE #27527: improve Lua loader error for inaccessible $VIMRUNTIME
### Problem

If `$VIMRUNTIME` is inaccessible, using a builtin Lua module fails with its usual error:

```
E5108: Error executing lua vim/_init_packages.lua:0: module 'vim.version' not found:
	no field package.preload['vim.version']
	no file './vim/version.lua'
	no file '/Users/runner/work/neovim/neovim/.deps/usr/share/luajit-2.1/vim/version.lua'
        ...
```

but none of the listed paths mention the `$VIMRUNTIME` directory. This can be very confusing for users trying to troubleshoot an issue.

This situation is not uncommon, for example it can happen if a user is trying `nvim` from the `Downloads` folder on macOS, which applications can't access by default. Real-world use-case is for embedders such as vscode-neovim, which may have a (runnable) absolute path to `~/Downloads/bin/nvim` without having access to `~/Downloads/`: https://github.com/vscode-neovim/vscode-neovim/issues/1815

## Steps to reproduce

Steps vary per OS. Steps for macOS:

1. download the [latest Nvim](https://github.com/neovim/neovim/releases) to your `~/Downloads` folder.
1. ensure that your terminal app does _not_ have access permissions to `Downloads`:
    - image
1. in your terminal app, run `nvim` from your downloads folder.
   ```
   ~/Downloads/nvim-macos/bin/nvim
   ```
1. Try to use a non-builtin Lua runtime file:
   ```
   :lua =vim.version
   ```
1. Try `!ls $VIMRUNTIME`:
   ```
   :!ls $VIMRUNTIME
   ls: /Users/xx/Downloads/t/nvim-macos/share/nvim/runtime: Operation not permitted
   shell returned 1
   ```

### Expected behavior

The main use-case is embedders. `:checkhealth` already warns about invalid `$VIMRUNTIME`.

- Lua loader error should say if it can't access `$VIMRUNTIME`, or include it in the "failed to access paths" list?

glepnir:
Problem: When $VIMRUNTIME is inaccessible, there shows a confusing
E484 without explaining the root cause.

Solution: Validate $VIMRUNTIME by sourcing filetype.lua before
enabling filetype/syntax plugins. Show E5009 if it fails.

Fix #27527

### DIFF
--- a/runtime/lua/vim/_core/util.lua
+++ b/runtime/lua/vim/_core/util.lua
@@ -99,4 +99,16 @@ function M.term_exitcode()
   return ''
 end
 
+--- Returns false and notifies if $VIMRUNTIME is inaccessible or invalid.
+---
+--- @return boolean
+function M._validate_vimruntime()
+  local rt = vim.fn.getenv('VIMRUNTIME')
+  if not rt or rt == '' or not vim.uv.fs_stat(rt .. '/filetype.lua') then
+    vim.notify(('E5009: Invalid $VIMRUNTIME: %q'):format(rt), vim.log.levels.ERROR)
+    return false
+  end
+  return true
+end
+
 return M
--- a/src/nvim/main.c
+++ b/src/nvim/main.c
@@ -495,11 +495,17 @@ int main(int argc, char **argv)
 
   // If using the runtime (-u is not NONE), enable syntax & filetype plugins.
   if (!vimrc_none || params.clean) {
-    // Sources filetype.lua unless the user explicitly disabled it with :filetype off.
-    filetype_maybe_enable();
-    // Sources syntax/syntax.vim. We do this *after* the user startup scripts so that users can
-    // disable syntax highlighting with `:syntax off` if they wish.
-    syn_maybe_enable();
+    Error err = ERROR_INIT;
+    Object result = NLUA_EXEC_STATIC("return require('vim._core.util')._validate_vimruntime()",
+                                     (Array)ARRAY_DICT_INIT, kRetNilBool, NULL, &err);
+    api_clear_error(&err);
+    if (LUARET_TRUTHY(result)) {
+      // Sources filetype.lua unless the user explicitly disabled it with :filetype off.
+      filetype_maybe_enable();
+      // Sources syntax/syntax.vim. We do this *after* the user startup scripts so that users can
+      // disable syntax highlighting with `:syntax off` if they wish.
+      syn_maybe_enable();
+    }
   }
 
   set_vim_var_nr(VV_VIM_DID_INIT, 1);
--- a/test/functional/core/startup_spec.lua
+++ b/test/functional/core/startup_spec.lua
@@ -1784,6 +1784,21 @@ describe('runtime:', function()
     command('edit FTDETECT')
     eq('SsABab', eval('g:aseq'))
   end)
+
+  it('shows E5009 when $VIMRUNTIME is inaccessible #27527', function()
+    local logfile = t.tmpname()
+    clear({
+      args_rm = { '-u' },
+      args = { '--clean' },
+      env = { VIMRUNTIME = '/foo/bar', NVIM_LOG_FILE = logfile },
+    })
+    matches('E5009: Invalid %$VIMRUNTIME', exec_capture('messages'))
+    if t.is_os('win') then
+      t.assert_log('neovim%.ico not found', logfile)
+    else
+      t.assert_nolog('.', logfile)
+    end
+  end)
 end)
 
 describe('user session', function()


runtime/lua/vim/_init_packages.lua
```diff
@@ -40,6 +40,21 @@ function vim._load_package(name)
...
+
+  if vim.startswith(basename, 'vim/') then
+    local vimruntime = os.getenv('VIMRUNTIME')
+    if vimruntime then
+      local accessible = vim.fn.isdirectory(vimruntime) == 1
+        and vim.fn.filereadable(vim.fn.expand(vimruntime)) == 1
+      if not accessible then
+        error('$VIMRUNTIME is not accessible or has insufficient permissions')
+      else
+        error('Module or shared library not found for ' .. name)
```
bfredl, line 52: when we don't know any specific error, we should return `nil` (the package loader will let the search continue and display the final error mesage)

glepnir: if we want validate it in main function i think loader no need change then ..

runtime/lua/vim/_init_packages.lua
```diff
@@ -40,6 +40,21 @@ function vim._load_package(name)
     local f, err = package.loadlib(found[1], 'luaopen_' .. modname:gsub('%.', '_'))
     return f or error(err)
   end
+
+  if vim.startswith(basename, 'vim/') then
+    local vimruntime = os.getenv('VIMRUNTIME')
+    if vimruntime then
+      local accessible = vim.fn.isdirectory(vimruntime) == 1
```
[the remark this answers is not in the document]
glepnir:
```diff
diff --git a/src/nvim/main.c b/src/nvim/main.c
index e18c56187..cee1be4bb 100644
--- a/src/nvim/main.c
+++ b/src/nvim/main.c
@@ -600,6 +600,7 @@ int main(int argc, char **argv)
     do_autocmd_uienter_all();
     TIME_MSG("UIEnter autocommands");
   }
+  validate_vimruntime_env();
 
 #ifdef MSWIN
   if (use_remote_ui) {
diff --git a/src/nvim/runtime.c b/src/nvim/runtime.c
index 44fcc29ad..e92aa5376 100644
--- a/src/nvim/runtime.c
+++ b/src/nvim/runtime.c
@@ -2871,3 +2871,22 @@ bool script_autoload(const char *const name, const size_t name_len, const bool
 r
   xfree(tofree);
   return ret;
 }
+
+void validate_vimruntime_env(void)
+{
+  const char *vimruntime_env = vim_getenv("VIMRUNTIME");
+  if (vimruntime_env == NULL) {
+    emsg(_("E5009: $VIMRUNTIME is empty or unset"));
+    return;
+  }
+
+  if (!os_isdir(vimruntime_env)) {
+    semsg(_("E5009: Invalid $VIMRUNTIME: %s"), vimruntime_env);
+    return;
+  }
+
+  if (!os_file_is_readable(vimruntime_env)) {
+    semsg(_("E5009: Invalid $VIMRUNTIME: %s is not accessible or has insufficient permissions"),
+          vimruntime_env);
+  }
+}
```

justinmk:
I wouldn't expect that checking `os_isdir` is necessary. Instead, just try to source a harmless, common file that we know is in the runtime, and handle the error. If it fails in any way, show an error explaining that VIMRUNTIME is invalid.

test/functional/core/startup_spec.lua
```diff
@@ -1737,6 +1737,23 @@ describe('runtime:', function()
     command('edit FTDETECT')
     eq('SsABab', eval('g:aseq'))
   end)
+
+  it('shows E5009 when $VIMRUNTIME is inaccessible #27527', function()
```
justinmk, line 1788: this test somehow causes an unexpected log message, which fails Windows CI:

    WRN 2026-02-21T10:03:00.619 c/T238.13292.0 os_icon_init:87: neovim.ico not found: \\.\pipe\T238.13292.0


I re-ran CI to see if it's flaky, but even if it passes next time, it's still suspicious that the failure was associated with this new test...

src/nvim/main.c
```diff
@@ -462,11 +462,14 @@ int main(int argc, char **argv)
 
   // If using the runtime (-u is not NONE), enable syntax & filetype plugins.
   if (!vimrc_none || params.clean) {
-    // Sources filetype.lua unless the user explicitly disabled it with :filetype off.
-    filetype_maybe_enable();
-    // Sources syntax/syntax.vim. We do this *after* the user startup scripts so that users can
-    // disable syntax highlighting with `:syntax off` if they wish.
-    syn_maybe_enable();
+    bool vimruntime_valid = validate_vimruntime_env();
+    if (vimruntime_valid) {
```
justinmk:
```suggestion
    if (validate_vimruntime_env()) {
```

justinmk:
Now that we have `vim/_core/` , is there any reason not to implement this in Lua? I see [previous discussion](https://github.com/neovim/neovim/pull/27537#discussion_r1494955969) about about `_init_packages.lua` , but I don't think that was intended to suggest it must live in C? cc @bfredl 

 could live in `vim/_core/editor.lua` or a new file like `vim/_core/startup.lua` ?

bfredl:
My comment was really about what it was used _for_ (by when it was triggered) , not the implementation language. Lua has been an implantation language of core since forever.

test/functional/core/startup_spec.lua
zeertzjq: Why is it unexpected?

justinmk: "unexpected" because CI expects the logfile to be empty. 

but yes it makes sense that the icon can't be find in this test case.
@glepnir the test should set NVIM_LOG_FILE and call `t.assert_log('neovim.ico not found'...)`

runtime/lua/vim/_core/editor.lua
```diff
@@ -1302,6 +1302,15 @@ function vim.deprecate(name, alternative, version, plugin, backtrace)
   end
 end
 
+function vim._validate_vimruntime()
```
justinmk: let's avoid adding more things to `vim.` . we have `vim/_core/` , we can create new modules as needed or choose a more appropriate existing module in there. i guess `vim/_core/util.lua` is ok for now.

runtime/lua/vim/_core/editor.lua
```diff
@@ -1302,6 +1302,15 @@ function vim.deprecate(name, alternative, version, plugin, backtrace)
...
+  local rt = os.getenv('VIMRUNTIME')
```
justinmk: VIMRUNTIME [is special](https://github.com/neovim/neovim/blob/ce5c7111f4050f168ddf4806287b4c384264b0fa/src/nvim/os/env.c#L969) so we should probably go through `vim.fn.getenv()` here. Unless `os.getenv()` already does that?

src/nvim/main.c
```diff
@@ -462,11 +462,17 @@ int main(int argc, char **argv)
 
   // If using the runtime (-u is not NONE), enable syntax & filetype plugins.
   if (!vimrc_none || params.clean) {
-    // Sources filetype.lua unless the user explicitly disabled it with :filetype off.
-    filetype_maybe_enable();
-    // Sources syntax/syntax.vim. We do this *after* the user startup scripts so that users can
-    // disable syntax highlighting with `:syntax off` if they wish.
-    syn_maybe_enable();
+    Error err = ERROR_INIT;
+    Object result = NLUA_EXEC_STATIC("return vim._validate_vimruntime()",
```
justinmk:
```suggestion
    Object result = NLUA_EXEC_STATIC("return require('vim._core.util._validate_vimruntime()')",
```

runtime/lua/vim/_core/util.lua
```diff
@@ -81,4 +81,13 @@ function M.source_is_lua(bufnr, line1, line2)
   return lang_tree:lang() == 'lua'
 end
 
+function M._validate_vimruntime()
+  local rt = vim.fn.getenv('VIMRUNTIME')
+  if rt and rt ~= '' and not vim.uv.fs_stat(rt .. '/filetype.lua') then
+    vim.notify('E5009: Invalid $VIMRUNTIME: ' .. rt, vim.log.levels.ERROR)
```
justinmk: if VIMRUNTIME is empty string that seems like an invalid case, no?
note: intentionally check `rt == ''` because a random `/filetype.lua` file in the filesystem root is not a "valid" runtime.


```suggestion
  if not rt or rt == '' or not vim.uv.fs_stat(rt .. '/filetype.lua') then
    vim.notify(('E5009: Invalid $VIMRUNTIME: %q'):format(rt), vim.log.levels.ERROR)
```

test/functional/core/startup_spec.lua
```diff
@@ -1737,6 +1737,28 @@ describe('runtime:', function()
...
+    clear()
+    local logfile = t.tmpname()
+    local screen = Screen.new(60, 7)
+    fn.jobstart({
+      nvim_prog,
+      '-u',
+      'NONE',
+      '--clean',
```
justinmk: since the test is already calling `clear()`, can we use `clear{env=...}` instead of `jobstart` ?

test/functional/core/startup_spec.lua
```diff
@@ -1737,6 +1737,28 @@ describe('runtime:', function()
...
+    }, {
+      term = true,
+      env = {
+        VIMRUNTIME = '/foo/bar',
+        NVIM_LOG_FILE = logfile,
+      },
+    })
+    screen:expect({ any = 'E5009: Invalid %$VIMRUNTIME: /foo/bar' })
+    if t.is_os('win') then
+      t.assert_log('neovim.ico not found', logfile)
```
justinmk: to avoid hidden problems on non-Windows, we need to assert that the logfile is empty on non-Windows. i think this should work (didn't test though):

    t.assert_nolog('.', logfile)

test/functional/core/startup_spec.lua
```diff
@@ -1737,6 +1737,21 @@ describe('runtime:', function()
...
+  it('shows E5009 when $VIMRUNTIME is inaccessible #27527', function()
+    local logfile = t.tmpname()
+    clear({
+      args_rm = { '-u' },
+      args = { '--clean' },
+      env = { VIMRUNTIME = '/foo/bar', NVIM_LOG_FILE = logfile, },
+    })
+    matches('E5009: Invalid %$VIMRUNTIME', exec_capture('messages'))
+    if t.is_os('win') then
+      t.assert_log('neovim.ico not found', logfile)
```
zeertzjq:
```suggestion
      t.assert_log('neovim%.ico not found', logfile)
```

Issue #25714, opened 2023-10-19
Add support for LSP completion side effects (snippet expansion, command execution, additionalTextEdits)


mfussenegger:
### Problem

The language server protocol specifies that completion items can contain:

- snippets
- `additionalTextEdits` (e.g used to add import statements when selecting a completion candidate)
- A `command` (used to execute client side or server side commands - for all sorts of things). For example, vscode defines a `editor.action.triggerParameterHints` client command to trigger a  signature help immediately after a completion candidate got selected.

Currently neovim core handles none of that, and instead delegates to completion plugins like nvim-cmp, nvim-lsp-compl or others.

### Expected behavior

Have this functionality built-in.

With https://github.com/neovim/neovim/pull/25301 the groundwork for snippet expansion is done.
The next step is to integrate it with the completion, omnifunc. For that we'll need to setup a `CompleteDone` autocmd similar to:

https://github.com/mfussenegger/nvim-lsp-compl/blob/1642f1c57a83a0b630fdf7f339f4a5550a766407/lua/lsp_compl.lua#L565-L668

Open Questions:

- [ ] Opt-in or opt-out. Enabling it by default could conflict with above established completion plugins. On the other hand, opt-in is another hurdle in getting started from scratch
- [ ] Snippet expansion trigger - implicit/explicit? (For me, always implicit expanding would make this unusable)

Proposal:

- Make it opt-in via a `vim.lsp.completion.attach(client, bufnr, [opts])` method that can be called in a `LspAttach` autocmd. It would sets up a `CompleteDone` autocmd for the given buffer and client.
- Add a `vim.lsp.completion.accept()` function that if called causes the next accempted completion candidate to expand a snippet, if the completion item contains one. To be used in a mapping like:

```
vim.keymap.set('i', '<CR>', function()
  return vim.lsp.completion.accept() and '<c-y>' or '<CR>'
end, { expr = true })
```

`opts` relate a bit to the auto-completion aspect and https://github.com/neovim/neovim/issues/25670, and might not be necessary in a first step.

## Related issues

- https://github.com/neovim/neovim/issues/13199
- https://github.com/neovim/neovim/issues/12310

justinmk:
> Currently neovim core handles none of that, and instead delegates to completion plugins like nvim-cmp, nvim-lsp-compl or others.

Can we extend the builtin completion concepts rather than add a LSP-specific interface?

> `vim.lsp.completion.attach(client, bufnr, [opts])`

[...]

PR #26724, opened 2023-12-23
fix(gen_lsp.lua): improve type name, and fix wrong type inheritance

wookayin:
Note: this PR consists of several commits, which were made in order to make diff atomic and PR easier to read/review, orthogonal to the upstream LSP model changes.

Follow-up after #23750.
Fixes #26617.

### Style improvements:

1. Anonymous classes derived from `StructureLiteralType` should have a
   better name. The class name can be also nested. Examples:

```diff
----@field serverInfo? anonym1
+---@field serverInfo? lsp._anonym1.serverInfo
```
```diff
----@field insertTextModeSupport? anonym26
+---@field insertTextModeSupport? lsp._anonym26.completionItem.insertTextModeSupport
```

2. Add one separate empty line before each `@field` definition. Without
   these, empty lines the doc can look confusing because descriptions
   also may contain empty lines. See `lsp.CompletionItem` for example:

```lua
---The kind of this completion item. Based of the kind
---an icon is chosen by the editor.
---@field kind? lsp.CompletionItemKind
---Tags for this completion item.
---
---@since 3.15.0
---@field tags? lsp.CompletionItemTag[]
```

   It might feel like "Tags for this completion item" belongs to `kind`,
   not `tags` due to the lack of separator blank lines. The following
   (after this commit) should look much better:

```diff
 ---The kind of this completion item. Based of the kind
 ---an icon is chosen by the editor.
 ---@field kind? lsp.CompletionItemKind
+---
 ---Tags for this completion item.
 ---
 ---@since 3.15.0
 ---@field tags? lsp.CompletionItemTag[]
```

3. Escape some LSP-specific annotations that can't be recognized by
   lua-ls. It'd be better to make them visible in LSP hover doc windows.

   Example: `@sample ...`.

### Fixes:

1. A type may extend from more than one base types (as well as mixin
   types). Previously only the first base class was being considered,
   resulting incomplete base classes for `@class` definitions.

   Example: `InlayHintOptions` (should have both of `resolveProvider`
   and `workDoneProgress`, the latter is from `WorkDoneProgressOptions`)

```diff
----@class lsp.InlayHintOptions
+---@class lsp.InlayHintOptions: lsp.WorkDoneProgressOptions
```

2. Remove `<200b>` (zero-width space) unicode characters.

3. Add the missing newline at EOF.

PR #27542, opened 2024-02-19
feat(lsp): support for completion itemDefaults

MariaSolOs:
Part of #25714.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -125,6 +125,15 @@ The following changes may require adaptations in user config or plugins.
 • Returning any truthy value from a callback passed to |nvim_create_autocmd()|
   (rather than just `true`) will delete the autocommand.
 
+• |vim.lsp.util.extract_completion_items()| will no longer return reliable
+  results, since it does not apply `itemDefaults` when its input is a
+  `CompletionList`.
+  Moreover, since support for LSP `completionList.itemDefaults` was added,
+  some third party plugins might be negatively impacted in case the language
+  servers support the feature but the plugin does not.
+  If necessary, the respective capability can be
+  removed when calling |vim.lsp.protocol.make_client_capabilities()|.
+
 ==============================================================================
 BREAKING CHANGES IN HEAD                                    *news-breaking-dev*
 
@@ -213,6 +222,11 @@ The following new APIs and features were added.
     the original LSP `Location` or `LocationLink`.
   • Added support for connecting to servers using named pipes (Windows) or
     unix domain sockets (Unix) via |vim.lsp.rpc.domain_socket_connect()|.
+  • Added support for `completionList.itemDefaults`, reducing overhead when
+    computing completion items where properties often share the same value
+    (e.g. `commitCharacters`). Note that this might affect plugins and
+    language servers that don't support the feature, and in such cases the
+    respective capability can be unset.
 
 • Treesitter
   • Bundled parsers and queries (highlight, folds) for Markdown, Python, and
--- a/runtime/lua/vim/lsp/_completion.lua
+++ b/runtime/lua/vim/lsp/_completion.lua
@@ -4,6 +4,8 @@ local lsp = vim.lsp
 local protocol = lsp.protocol
 local ms = protocol.Methods
 
+--- @alias vim.lsp.CompletionResult lsp.CompletionList | lsp.CompletionItem[]
+
 ---@param input string unparsed snippet
 ---@return string parsed snippet
 local function parse_snippet(input)
@@ -37,7 +39,7 @@ local function get_completion_word(item)
   return item.label
 end
 
----@param result lsp.CompletionList|lsp.CompletionItem[]
+---@param result vim.lsp.CompletionResult
 ---@return lsp.CompletionItem[]
 local function get_items(result)
   if result.items then
@@ -49,7 +51,7 @@ end
 --- Turns the result of a `textDocument/completion` request into vim-compatible
 --- |complete-items|.
 ---
----@param result lsp.CompletionList|lsp.CompletionItem[] Result of `textDocument/completion`
+---@param result vim.lsp.CompletionResult Result of `textDocument/completion`
 ---@param prefix string prefix to filter the completion items
 ---@return table[]
 ---@see complete-items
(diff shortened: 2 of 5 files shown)


runtime/lua/vim/lsp/_completion.lua
```diff
@@ -37,19 +37,49 @@ local function get_completion_word(item)
...
+    end
+  end
+end
+
+---@param result vim.lsp.CompletionResult
 ---@return lsp.CompletionItem[]
 local function get_items(result)
   if result.items then
+    for _, item in ipairs(result.items) do
+      ---@diagnostic disable-next-line: param-type-mismatch
```
MariaSolOs, line 70: This is needed because the type of `result.itemDefaults` is [anonymous](https://github.com/neovim/neovim/blob/8952a89db588db10a9dba16356f9bbd35ca5fabb/runtime/lua/vim/lsp/_meta/protocol.lua#L4776), and using an autogenerated `_anonym` type feels flaky. This is why I manually define `lsp.ItemDefaults` in `completion.lua`, which duplicates the definition :(

Happy to hear for better ideas here.

justinmk: > using an autogenerated `_anonym` type feels flaky.

We could give a better name to the generated "anon" types maybe. I don't see why we would manually define generated types, since we are generating things to avoid that.

MariaSolOs: Okay. I don't want to block this PR waiting for the change, so I'll use the anonymous type and update it when we use a better name. WDYT @justinmk?

runtime/lua/vim/lsp/completion.lua
```diff
@@ -0,0 +1,11 @@
+local M = {}
+
+--- @alias vim.lsp.CompletionResult lsp.CompletionList | lsp.CompletionItem[]
```
mfussenegger, line 3: Why the separate file for this? Can this go into _completion?

If we're going to have more public API I think we can then rename that module. Having both _completion and completion shouldn't be necessary(?)

runtime/lua/vim/lsp/protocol.lua
```diff
@@ -764,6 +764,14 @@ function protocol.make_client_capabilities()
             return res
           end)(),
         },
+        completionList = {
+          itemDefaults = {
```
mfussenegger, line 768: Note that this is indirectly breaking one of the still exposed lsp.utils methods that was recently deprecated. Part of the reason why I suggested to add itemDefaults later and why I wanted to fast track the removal of those methods.

I'm fine pushing ahead with this anyway, but if we do we could at least mention it in the news.txt

runtime/lua/vim/lsp/_completion.lua
mfussenegger: See the discussion in https://github.com/neovim/neovim/pull/26724#discussion_r1436821011
Ideally we'd get rid of the anonymous types, until we have that I think it's fine to declare them manually

runtime/lua/vim/lsp/completion.lua
MariaSolOs: It's because the work of #27542 will be done in `completion.lua`, and because I feel these are "public" types it made sense to put it here, unless we want to initially put all of the completion work in the private module.

runtime/lua/vim/lsp/_completion.lua
MariaSolOs: So are you okay with the manual declaration done in this PR, or would you rather use the poorly named generated one (like @justinmk I believe prefers)?

runtime/lua/vim/lsp/protocol.lua
mfussenegger: `extract_completion_items` if called without the default application will likely no longer return usable results.

runtime/lua/vim/lsp/completion.lua
mfussenegger: > unless we want to initially put all of the completion work in the private module.

Either that, or we already rename `_completion` to `completion.lua`.
Not sure how we deal with the omnifunc, because it is also exposed under `vim.lsp.omnifunc`. Could rename it to `_omnifunc`, or deprecate `vim.lsp.omnifunc` in favour of `vim.lsp.completion.omnifunc`. No strong opinion on that.

What I'd like to avoid is having a public `completion.lua` that calls into a private `_completion.lua`.

runtime/lua/vim/lsp/protocol.lua
MariaSolOs: I see. Okay yeah I'm not too worried given the util will be removed soon. I will document that in both the function and in `news.txt`.

runtime/lua/vim/lsp/completion.lua
MariaSolOs: Alright I'll move this to `_completion.lua` (we might rename it in #27542). The omnifunc issue might be a larger/longer discussion.

runtime/doc/news.txt
```diff
@@ -125,6 +125,10 @@ The following changes may require adaptations in user config or plugins.
 • Returning any truthy value from a callback passed to |nvim_create_autocmd()|
   (rather than just `true`) will delete the autocommand.
 
+• |vim.lsp.util.extract_completion_items()| will no longer return reliable
```
mfussenegger, line 128: This could also mention the added support for itemDefaults and maybe expand on the implications (lower completion overhead for servers supporting it, and that it could also break 3rd party plugins which don't support it - maybe with a hint how to remove the capability)

mfussenegger (approved):

runtime/doc/news.txt
```diff
@@ -125,6 +125,14 @@ The following changes may require adaptations in user config or plugins.
...
+  results, since it does not apply `itemDefaults` when its input is a
+  `CompletionList`.
+  Moreover, since support for LSP `completionList.itemDefaults` was added,
+  some third party plugins might be negatively impacted because of lack of
+  support by language servers. If necessary, the respective capability can be
```
mfussenegger, line 133: Small correction: They're negatively affected if the servers support it but the plugin doesn't.

MariaSolOs:
I think this is ready to go ;) @mfussenegger

clason:
Do you want to squash your commits?

PR #27364, opened 2024-02-06
docs: small fixes

dundargoc:

PR #27544, opened 2024-02-20
docs: document ttimeout being enabled

craigmac:
Was not documented in list of vim differences that this is enabled by default.


dundargoc:
Thank you for your contribution! We collect all typo fixes and similar small fixes into a single pull request and merge it once it gets big enough: https://github.com/neovim/neovim/pull/27364

PR #27547, opened 2024-02-20
feat(help): hide filename of "gO" outline using conceal

itchyny:
Help outlines, invoked by `gO`, displays the help section titles in the location list window. This feature is implemented by setting the buffer lines after opening the window, but this implementation breaks the assumption that the quickfix window texts are consistently constructed by the quickfix list items. I think we can use the conceal feature here. Using conceal here improves interoperability between quickfix plugins, and also simplifies the outline implementation.
Originally reported at https://github.com/itchyny/vim-qfedit/issues/12.

### DIFF
--- a/runtime/ftplugin/help.vim
+++ b/runtime/ftplugin/help.vim
@@ -71,7 +71,7 @@ if !exists('g:no_plugin_maps')
 
         if indent(lnum) <= indent(l)
           let level = has_section + has_sub_section
-          let add_text = matchstr(text, '\S.*')
+          let add_text = matchstr(text, '\S.\{-}\ze\s\=\~$')
         endif
       endif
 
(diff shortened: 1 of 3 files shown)


clason:
Thank you, but we don't maintain these plugins; please open the PR at vim/vim -- if they make the change, we will port it.

zeertzjq:
@clason The modified parts are Nvim-only.

clason:
Not all of them. (And why are these not documented, then?)

zeertzjq:
`:h gO`

clason:
I mean in the runtime files themselves. Local modifications need to be documented.

justinmk (approved):
Thank you! `gO` was a "minimum viable product", so improvements were expected and welcome.

clason:
I would still like to have seen these modifications clearly documented to prevent confusion in the future. This is adding needless friction to people (me) porting runtime file updates.

justinmk:
Yes, but that shouldn't block this PR.

PR #27552, opened 2024-02-20
refactor(api): make freeing of return-value opt-in instead of opt out.

bfredl:
As only a few API functions make use of explicit freeing of the return value, make it _opt-in_ instead. The arena is always present under the hood, so `Arena *arena` arg now doesn't mean anything other than getting access to this arena. Also it is in principle possible to return an allocated value while still using the arena as scratch space for other stuff (unlikely, but there no reason to not allow it).

### DIFF
--- a/src/nvim/api/deprecated.c
+++ b/src/nvim/api/deprecated.c
@@ -32,8 +32,8 @@
 /// @deprecated Use nvim_exec2() instead.
 /// @see nvim_exec2
 String nvim_exec(uint64_t channel_id, String src, Boolean output, Error *err)
-  FUNC_API_SINCE(7)
-  FUNC_API_DEPRECATED_SINCE(11)
+  FUNC_API_SINCE(7) FUNC_API_DEPRECATED_SINCE(11)
+  FUNC_API_RET_ALLOC
 {
   Dict(exec_opts) opts = { .output = output };
   return exec_impl(channel_id, src, &opts, err);
@@ -42,8 +42,8 @@ String nvim_exec(uint64_t channel_id, String src, Boolean output, Error *err)
 /// @deprecated
 /// @see nvim_exec2
 String nvim_command_output(uint64_t channel_id, String command, Error *err)
-  FUNC_API_SINCE(1)
-  FUNC_API_DEPRECATED_SINCE(7)
+  FUNC_API_SINCE(1) FUNC_API_DEPRECATED_SINCE(7)
+  FUNC_API_RET_ALLOC
 {
   Dict(exec_opts) opts = { .output = true };
   return exec_impl(channel_id, command, &opts, err);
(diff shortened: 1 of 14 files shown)


src/nvim/api/deprecated.c
```diff
@@ -44,6 +44,7 @@ String nvim_exec(uint64_t channel_id, String src, Boolean output, Error *err)
 String nvim_command_output(uint64_t channel_id, String command, Error *err)
   FUNC_API_SINCE(1)
   FUNC_API_DEPRECATED_SINCE(7)
+    FUNC_API_RET_ALLOC
```
justinmk, line 47: is uncrustify doing this weird indent?

src/nvim/generators/gen_api_dispatch.lua
```diff
@@ -60,7 +60,6 @@ local function add_function(fn)
       fn.parameters[#fn.parameters] = nil
     end
     if #fn.parameters ~= 0 and fn.parameters[#fn.parameters][1] == 'arena' then
-      -- return value is allocated in an arena
       fn.arena_return = true
```
justinmk, line 63: So this means the `Arena` parameter is still supported? Should we mention that in the docstring for `FUNC_API_RET_ALLOC` or somewhere else? When would that be used?

zeertzjq: When it needs to be passed to another function.

justinmk: Ok so this is similar to the `channel_id` magic parameter. I guess we currently don't have a place where those are documented.

bfredl: This change removes the magic behavior associated with `arena` and *changes* it into an actually documented attribute instead. This PR literally is that documentation addition.

justinmk: My main question, which I still have, is: the name `arena_return` implies that this is a way to return stuff, i.e. an alternative to `FUNC_API_RET_ALLOC`. So does it make sense to mention `arena` parameter in the docstring on `FUNC_API_RET_ALLOC` ?

src/nvim/func_attr.h
```diff
@@ -212,6 +212,8 @@
 #if defined(DEFINE_FUNC_ATTRIBUTES) || defined(DEFINE_EMPTY_ATTRIBUTES)
 /// Fast (non-deferred) API function.
 # define FUNC_API_FAST
+/// Return value needs to be freed
```
justinmk, line 215:
```suggestion
/// Return value must be freed. For non-allocating return, add an `Arena` parameter to the function signature.
```

bfredl: The mere _presence_ of an arena parameter does not change _any_ behavior anymore, so this would be quite misleading, as an API function can just as well return Object values which has a longer, external lifetime (like channel info is a good example). 

The point of an arena is just that the arena can be _used_ to allocate values which doesn't need to be xfree:d, which can be (part of) the return value, or other stuff besides that which just needs any temporary memory, which is why several `void` returning functions also use it.

src/nvim/generators/gen_api_dispatch.lua
bfredl: ye, the name `arena_return` should be changed.

PR #27559, opened 2024-02-21
doc: fix type annotation for lsp.WorkspaceEdit

stevearc:
A super easy one: the type annotation here is incorrect. LuaLS needs the parens to interpret this as a list of unions, rather than a union with the last item being a list.


MariaSolOs:
These types are autogenerated, so it looks like there's a bug in https://github.com/neovim/neovim/blob/master/scripts/gen_lsp.lua?

MariaSolOs:
@stevearc congrats! You found a bug in our generator script :D

#27560 fixes it. Since we don't want to change the generated code, are you okay with closing this PR and using that one instead?

stevearc:
SGTM! Thanks for the proper fix

PR #27562, opened 2024-02-21
perf(os/env): os_getenv() allocation when there is no env var set

bfredl:
os_getenv("FOO") caches the result when $FOO is set to something non-empty. However, when $FOO was not set, every new call to os_getenv("FOO") would allocate a temporary scratch buffer to immediately throw away.

This has an huge impact e.g. on logging which depends on potentially non-set env vars which are checked each time.

### DIFF
--- a/src/nvim/os/env.c
+++ b/src/nvim/os/env.c
@@ -61,8 +61,7 @@ static PMap(cstr_t) envmap = MAP_INIT;
 const char *os_getenv(const char *name)
   FUNC_ATTR_NONNULL_ALL
 {
-  char *e;
-  size_t size = 64;
+  char *e = NULL;
   if (name[0] == '\0') {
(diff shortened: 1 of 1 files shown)


src/nvim/os/env.c
```diff
@@ -78,23 +77,31 @@ const char *os_getenv(const char *name)
...
+    if (r != 0 || size == 0 || e[0] == '\0') {
+      xfree(e);
+      goto end;
+    }
+  } else if (r != 0 || size == 0 || buf[0] == '\0') {
     e = NULL;
     goto end;
+  } else {
+    // NB: `size` param of uv_os_getenv() includes the NUL-terminator,
+    // except when it does not include the NUL-terminator.
```
clason, line 96: 😆

PR #27566, opened 2024-02-21
build(deps): bump tree-sitter to v0.21.0

clason:


clason:
Max flubbed the release _again_... Let's wait a bit and see whether he starts taking the sledge hammer to things before merging.

justinmk:
is there something we can say in the commit message about changes that are relevant to us?

clason:
Hard to say, since changelogs are another thing Max is not interested in. I'll manually trawl the git log and see if something sticks out.

clason:
Nothing, really, for the lib. Relevant changes are in relation to WASM parsers, which we haven't really looked into yet (but which this release makes possible).

PR #26557, opened 2023-12-13
fix(defaults): background detection in tmux

tomtomjhj:
Wrap the query with passthrough sequence

PR #27571, opened 2024-02-22
fix(defaults): remove tmux background detection passthrough

zeertzjq:
There is now a new tmux 3.4 release that queries background color from
the parent terminal if background is not set in tmux, so removing the
passthrough still works when background is not set in tmux, and fixes
the incorrect detection when background is set in tmux.

This reverts PR #26557.


gpanders (approved):

przepompownia:
Thanks for the fix before I found time to report the problem for 3.4.

PR #27584, opened 2024-02-22
refactor(defaults): use getregion() for default * and # mappings

zeertzjq:


runtime/lua/vim/_defaults.lua
```diff
@@ -4,27 +4,9 @@ do
...
       assert(cmd == '/' or cmd == '?')
-      local region = vim.region(
-        0,
-        '.',
-        'v',
-        vim.api.nvim_get_mode().mode:sub(1, 1),
-        vim.o.selection == 'inclusive'
-      )
-      local chunks = region_chunks(region)
+      local chunks = vim.fn.getregion('.', 'v', vim.fn.mode())
```
justinmk, line 9: Does vim.region need an inclusive/exclusive parameter?

zeertzjq: `getregion()` falls back to the 'selection'

PR #27588, opened 2024-02-23
refactor(terminal): use SM 20 for force_crlf

zeertzjq:
Ref https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

    CSI Pm h  Set Mode (SM).
                Ps = 2  ⇒  Keyboard Action Mode (KAM).
                Ps = 4  ⇒  Insert Mode (IRM).
                Ps = 1 2  ⇒  Send/receive (SRM).
                Ps = 2 0  ⇒  Automatic Newline (LNM).


zeertzjq:
This doesn't seem correct, as SM 20 also changes the behavior of VT and FF.

PR #27593, opened 2024-02-23
docs(treesitter): fix ambiguous parameter description about `lang`

### LINKED ISSUE #27754: Wrong description about parameters of treesitter functions
### Problem

This is a documentation issue and not a bug, see https://github.com/neovim/neovim/pull/27593
I don't understand how I can fix it and I found that I forgot to create a feature branch to fix it, so turning it into an issue, I might close it in a few days, or if someone fixes it I'll close this PR right away.

### Steps to reproduce

nvim --clean +help vim.treesitter.get_parser()

### Expected behavior

There is a wrong description of the parameter lang. It is the name of the parser actually instead of the filetype, the wrong description may misguide users to pass vim.o.filetype to it, but it will not work in some cases, such as the filetype typescriptreact corresponds to parser tsx.

### Neovim version (nvim -v)

0.10.0 commit 0d553c8

no

### Operating system/version

Linux

### Terminal name/version

kitty

xterm-kitty

### Installation

Build from repo

ofseed:
There is a wrong description of the parameter `lang`. It is the name of the parser actually instead of the filetype, the wrong description may misguide users to pass `vim.o.filetype` to it, but it will not work in some cases, such as the filetype `typescriptreact` corresponds to parser `tsx`.

### DIFF
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -782,7 +782,7 @@ get_parser({bufnr}, {lang}, {opts})              *vim.treesitter.get_parser()*
     Parameters: ~
       • {bufnr}  (`integer?`) Buffer the parser should be tied to (default:
                  current buffer)
-      • {lang}   (`string?`) Filetype of this parser (default: buffer
+      • {lang}   (`string?`) Language of this parser (default: from buffer
                  filetype)
       • {opts}   (`table?`) Options to pass to the created language tree
 
@@ -829,7 +829,7 @@ inspect_tree({opts})                           *vim.treesitter.inspect_tree()*
       • {opts}  (`table?`) Optional options table with the following possible
                 keys:
                 • lang (string|nil): The language of the source buffer. If
-                  omitted, the filetype of the source buffer is used.
+                  omitted, detect from the filetype of the source buffer.
                 • bufnr (integer|nil): Buffer to draw the tree into. If
                   omitted, a new buffer is created.
                 • winid (integer|nil): Window id to display the tree buffer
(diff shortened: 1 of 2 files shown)


wookayin:
The title of commit and PR could be `docs`, not `doc`.

Looks good -- one note is that there are more similar descriptions in other places as well; e.g. `inspect_tree`.

ofseed:
No matter whether `filetype` should be accepted as `lang`, because these functions must accept `lang` as a parameter, I think the documentation should be clear to pass `lang` instead of `filetype`.

But I think we can be more lenient under the hood too, by modifying `get_parser` to accept filetype as well, we can make these functions work when passing filetype too.

There's already a private function called `resolve_lang` in `treesitter/languagetree.lua`, can I expose it as an API and use it in `get_parser` to ignore the differences between filetype and lang?

ofseed:
The documentation should be fixed now. Because they all use `get_parser` under the hood, they should have the same behavior.

ofseed:
>elsewhere it looks to be correctly distinguishing between the language of the parser and the filetype of the buffer.

These functions whose docsting I modified in this PR all use `get_parser`. So I think they all cannot distinguish the filetype and language of the parser, sometimes it works just because the parser name and the filetype have the same value.

And the default value of them exactly is the parser of the buffer, not the filetypes, so these are wrong anyway.

clason:
My point is that there is no such thing as "buffer language".

ofseed:
Indeed. Would this be better? At least it distinguishes that this is not the filetype.

clason (approved):

PR #27594, opened 2024-02-23
fix(defaults): validate 'channel' before responding to OSC request

### LINKED ISSUE #27587: OSC 11 request doesn't work in terminal opened by nvim_open_term
### Problem

OSC 11 request doesn't work in terminal opened by `nvim_open_term`. It causes an error and doesn't receive a response.

### Steps to reproduce

1. Run `nvim --clean`
2. Source the following Lua file:
```lua
_G.input = ''
local chan = vim.api.nvim_open_term(0, {
  on_input = function(_, _, _, data)
    _G.input = _G.input .. data
  end,
})
vim.api.nvim_chan_send(chan, '\027]11;?\027\\')
print(_G.input)
```
3. Error:
```
Error detected while processing TermRequest Autocommands for "*":
Error executing lua callback: vim/_defaults.lua:0: E900: Invalid channel id
stack traceback:
        [C]: in function 'nvim_chan_send'
```

### Expected behavior

No error and `_G.input` contains an OSC response.

### Neovim version (nvim -v)

v0.10.0-dev-2427+geb4783fb6c

N/A

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.31.0

xterm-kitty

### Installation

build from repo

gpanders:
Fixes: https://github.com/neovim/neovim/issues/27587

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -143,6 +143,10 @@ do
     group = nvim_terminal_augroup,
     desc = 'Respond to OSC foreground/background color requests',
     callback = function(args)
+      local channel = vim.bo[args.buf].channel
+      if channel == 0 then
+        return
+      end
       local fg_request = args.data == '\027]10;?'
       local bg_request = args.data == '\027]11;?'
       if fg_request or bg_request then
@@ -157,7 +161,6 @@ do
         end
         local command = fg_request and 10 or 11
         local data = string.format('\027]%d;rgb:%04x/%04x/%04x\007', command, red, green, blue)
-        local channel = vim.bo[args.buf].channel
         vim.api.nvim_chan_send(channel, data)
       end
     end,


zeertzjq:
This behavior may be unexpected. For `termopen()` 'channel' is input to the process, but the channel of `nvim_open_term()` is output to the terminal.

zeertzjq:
The behavior of the default `TermRequest` autocommand is still wrong. It triggers `TermRequest` again, and the `on_input` callback doesn't receive anything.

gpanders:
This is not the right change. The channel returned by `nvim_open_term` behaves differently with `nvim_chan_send` than the channel used with `jobstart` and `termopen`, so setting it to `'channel'` is misleading.

I think the proper solution to https://github.com/neovim/neovim/issues/27587 is simply to check if `'channel'` is 0 and exit early if so. I'm not sure what else there is to do. @zeertzjq do you agree?

zeertzjq:
I think that an OK solution when there is no way to retrieve the `on_input` callback. That should probably be a separate request.

zeertzjq (approved):

PR #27596, opened 2024-02-23
Fix: utf_cp_tail_off() considers trailing bytes beyond character end

vanaigr:
`utf_cp_tail_off()` currently counts the number of trailing UTF-8 bytes and checks if it is exactly equal to the expected number of trailing bytes. This causes all bytes of a correct UTF-8 character that is followed by trailing bytes to be considered illegal (with tail offset = 0).

This function also expects the string to be NUL-terminated, which is, as far as I know, not the case for Lua strings, but the function is used in `nlua_str_utf_end()`.


zeertzjq:
The same problems seem to also apply to `utf_cp_head_off()`, so I think it's better to fix both in one PR.

VanaIgr:
I replaced `utf_cp_tail_off()`, `utf_cp_head_off()` and `mb_off_next()` with `utf_cp_bounds()`.

PR #27599, opened 2024-02-24
refactor(fileio): remove API shell layer encouraging unnecessary allocations

bfredl:
Functions like `file_open_new()` and `file_open_fd_new()` which just is a wrapper around the real functions but with an extra xmalloc/xfree around is an anti-pattern. If the caller really needs to allocate a `FileDescriptor` as a heap object, it can do that directly.

FileDescriptor by itself is pretty much a pointer, or rather two: the OS fd index and a pointer to a buffer. So most of the time an extra pointer layer is just wasteful.

In the case of `scriptin[curscript]` in getchar.c, curscript=N used to mean in practice:

```
N+1 open scripts             when curscript>0
zero or one open scripts     when curscript==0
```

Which means scriptin[0] had to be compared to NULL to disambiguate the curscript=0 case.

Instead, use curscript==-1 to mean that are no script, then all pointer comparisons dissappear and we can just use an array of structs without extra pointers.


bfredl:
Got rid of `ShaDaWriteDef`

`shada_write()` is only ever used with FileDescriptor. We already have separate
 code paths for serializing shada data into memory, see `shada_encode_regs()` and friends.

src/nvim/getchar.c
```diff
@@ -1342,31 +1339,53 @@ void openscript(char *name, bool directly)
...
+    error = file_open(&scriptin[0], scriptin_name,
+                      kFileReadOnly|kFileNonBlocking, 0);
+  }
+  if (error) {
+    vim_snprintf(IObuff, IOSIZE,
+                 _("Cannot open for reading: \"%s\": %s\n"),
+                 scriptin_name, os_strerror(error));
+    fprintf(stderr, "%s", IObuff);
+    curscript--;
+    os_exit(2);
```
zeertzjq, line 1379: It may be more consistent to make `open_scriptin()` return a boolean value, and move the error message printing and `os_exit(2)` to `main()`.

src/nvim/shada.c
```diff
@@ -636,33 +616,12 @@ static int read_char(ShaDaReadDef *const sd_reader)
...
 {
   close_file(sd_reader->cookie);
-}
-
-/// Wrapper for closing file descriptors opened for writing
-static void close_sd_writer(ShaDaWriteDef *const sd_writer)
-  FUNC_ATTR_NONNULL_ALL
-{
-  close_file(sd_writer->cookie);
+  xfree(sd_reader->cookie);
```
zeertzjq, line 624:
```suggestion
  XFREE_CLEAR(sd_reader->cookie);
```

PR #27601, opened 2024-02-24
refactor(lsp): remove redundant code

tomtomjhj:
* use builtin function
* buffer:// was removed in 236c20795eb9f11e21e0719b735ea741711acc08.

(split out from #24975)


mfussenegger:
Thanks for splitting it up

PR #27603, opened 2024-02-24
fix(lsp): when renaming directory, check path prefix of buffer names

tomtomjhj:
Problem: When renaming `/path/to/dir`, buffers with names like `fern://drawer/file:///path/to/dir`, `/path/to/dir123` are deleted.

Solution: Match each path component.

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -639,13 +639,28 @@ function M.text_document_completion_list_to_complete_items(result, prefix)
   return vim.lsp._completion._lsp_to_complete_items(result, prefix)
 end
 
---- Get list of buffers for a directory
-local function get_dir_bufs(path)
-  path = path:gsub('([^%w])', '%%%1')
+local function path_components(path)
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/lsp/util.lua
```diff
@@ -639,13 +639,28 @@ function M.text_document_completion_list_to_complete_items(result, prefix)
...
+---@return integer[]
+local function get_bufs_with_prefix(prefix)
+  prefix = path_components(prefix)
   local buffers = {}
   for _, v in ipairs(vim.api.nvim_list_bufs()) do
-    local bufname = vim.api.nvim_buf_get_name(v)
-    if bufname:find(path) then
+    local bname = vim.api.nvim_buf_get_name(v)
+    local path = path_components(vim.fs.normalize(bname, { expand_env = false }))
+    if path_under_prefix(path, prefix) then
```
mfussenegger, line 663: Isn't this `vim.startswith` ?

mfussenegger (approved):

PR #27607, opened 2024-02-24
fix(lsp): use real bufnr when getting codelens

icholy:


tomtomjhj:
I suggest removing the metatable of `lens_cache_by_buf` since the `__index` metamethod is no longer needed.

icholy:
Ah I missed that metatable. This PR isn't required.

PR #27609, opened 2024-02-24
docs: fix several misleading and superfluous wordings

echasnovski:
In `nvim_buf_set_extmark()` the `cursorline_hl_group` is not used to highlight the actual line, but the placed sign text (similar to `culhl` in legacy signs).

First argument of `on_line` in `nvim_set_decoration_provider()` is indeed `"line"` and not `"win"`.


echasnovski:
Also removed `Note: ranges are unsupported and  decorations are only applied to start_row` as they are superfluous at this point.

PR #27614, opened 2024-02-24
fix(lsp): add assertion for explicit bufnr in apply_text_edits

### LINKED ISSUE #27613: LSP formatexpr does not properly save and restore marks
### Problem

Use the following mapping to reformat the entire buffer while keeping the cursor's position:

```vim
nnoremap gq<CR> mzgggq`z
```

Using this with `set formatexpr=v:lua.vim.lsp.formatexpr()` shows an error `E20: Mark not set` and does not correctly restore the cursor position.

local pattern = 'lua'
local root_markers = {'.git', '.editorconfig', '.luarc.json'}

  pattern = pattern,
  callback = function(args)
    vim.lsp.start({
      name = 'lua-language-server',
      cmd = cmd,
      root_dir = root_dir,
      settings = settings
    })
  end
})

vim.keymap.set('n', 'gq<CR>', 'mzgggq`z')

### Expected behavior

The cursor position does not move after formatting using the mapping above.

### Neovim version (nvim -v)

71384129f75462f14344df236cf0a22b83da4f4c

N/A

### Operating system/version

N/A

### Log file

_No response_

gpanders:
Assert that the buffer number passed to apply_text_edits is fully
resolved (not 0 or null). Pass the known buffer number to
apply_text_edits from lsp.formatexpr().

Fixes: https://github.com/neovim/neovim/issues/27613

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -1280,7 +1280,7 @@ function lsp.formatexpr(opts)
       local response =
         client.request_sync(ms.textDocument_rangeFormatting, params, timeout_ms, bufnr)
       if response and response.result then
-        lsp.util.apply_text_edits(response.result, 0, client.offset_encoding)
+        lsp.util.apply_text_edits(response.result, bufnr, client.offset_encoding)
         return 0
       end
     end
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -419,6 +419,11 @@ function M.apply_text_edits(text_edits, bufnr, offset_encoding)
   if not next(text_edits) then
     return
   end
+
+  if not bufnr or bufnr == 0 then
+    bufnr = api.nvim_get_current_buf()
+  end
+
   if not api.nvim_buf_is_loaded(bufnr) then
     vim.fn.bufload(bufnr)
   end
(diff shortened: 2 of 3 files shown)


runtime/lua/vim/lsp.lua
```diff
@@ -1280,7 +1280,7 @@ function lsp.formatexpr(opts)
       local response =
         client.request_sync(ms.textDocument_rangeFormatting, params, timeout_ms, bufnr)
       if response and response.result then
-        lsp.util.apply_text_edits(response.result, 0, client.offset_encoding)
+        lsp.util.apply_text_edits(response.result, bufnr, client.offset_encoding)
```
gpanders, line 1283: This is a bit of a redundant fix and is not strictly necessary since `apply_text_edits` will resolve a `0` buffer number, but since we already have the buffer number available anyway we might as well use it.

test/functional/plugin/lsp_spec.lua
```diff
@@ -1808,7 +1808,7 @@ describe('LSP', function()
         make_edit(4, 0, 5, 0, 'barfoo'),
       }
       eq(true, exec_lua('return vim.api.nvim_buf_set_mark(1, "a", 2, 1, {})'))
-      exec_lua('vim.lsp.util.apply_text_edits(...)', edits, 1, 'utf-16')
+      exec_lua('vim.lsp.util.apply_text_edits(...)', edits, 0, 'utf-16')
```
gpanders, line 1811: This change is sufficient to demonstrate the fix. With `0` this test fails on master.

runtime/lua/vim/lsp/util.lua
```diff
@@ -419,6 +419,11 @@ function M.apply_text_edits(text_edits, bufnr, offset_encoding)
   if not next(text_edits) then
     return
   end
+
+  if not bufnr or bufnr == 0 then
```
mfussenegger, line 423: The method here is intended to be called with a result from a language server for a specific document/ buffer, so I think at this point this should be a concrete buffer number already. Because of that I'd tend to turn this into an error with an assert.

(Also the type annotation doesn't imply nil)

MariaSolOs (approved):

runtime/lua/vim/lsp.lua
mfussenegger: +1 for the change. See my comment below I think this is actually the right fix

mfussenegger (approved):

PR #27618, opened 2024-02-25
fix(lsp): add snippet regression test

### LINKED ISSUE #25895: Snippet state is not updated when built-in completion popup menu is visible
### Problem

When completion popup menu is visible during entering tabstop text but no its entry is selected, snippet state seems to be not updated properly. This situation is common with autocompletion which uses built-in completion or if indeed user wants to use on demand built-in completion.

### Steps to reproduce

1. Create the following config file:
```lua
vim.keymap.set('i', '<C-l>', function() vim.snippet.jump(1) end)
vim.keymap.set('i', '<C-e>', function() vim.snippet.expand('$1 = function($2)\n$3\nend') end)
```
2. Start Neovim with `--clean` and this config file.
3. Start Insert mode and press `<C-e>` to expand the snippet.
4. Press `<C-n>` to show completion popup and `<C-p>` to make no item selected.
5. Press `<C-l>` to jump forward.
6. Type any character. Whole line up until cursor is highlighted as if entered text in tabstop. Pressing `<C-l>` again moves cursor to the start of the snippet.

### Expected behavior

Snippet navigation does not depend on whether popup is visible or not.

------

What I assume is that some state tracking is not triggered when popup menu is visible. There are several places which met the eye on the first sight:
- [These autocommands](https://github.com/neovim/neovim/blob/5cefec7349610853910c21a0215f85a4d47132d1/runtime/lua/vim/snippet.lua#L322) don't include `TextChangedP` (as `TextChangedI` is "Not triggered when the popup menu is visible."). Although, I've added it and didn't see positive changes, but maybe I am missing something.
- [These autocommands](https://github.com/neovim/neovim/blob/5cefec7349610853910c21a0215f85a4d47132d1/runtime/lua/vim/snippet.lua#L276) are not triggered when popup menu is visible (as per `:h CursorMovedI`). I would assume those should not be crucial here because cursor move is done with `vim.snippet.jump()`, but I might be missing something.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-1466+gec66a95fb-dirty

No

### Operating system/version

EndeavourOS Linux x86_64 (6.5.9-arch2-1)

### Terminal name/version

st-0.9

st-256color

### Installation

From source

MariaSolOs:
Adding a snippet test to prevent #25895 from happening in the future.

Note that the mentioned bug was (unintentionally) fixed on master before.

Fixes #25895.

### DIFF
--- a/test/functional/lua/snippet_spec.lua
+++ b/test/functional/lua/snippet_spec.lua
@@ -228,4 +228,16 @@ describe('vim.snippet', function()
     feed(',2')
     eq({ 'for i=1,10,2 do', '\t', 'end' }, buf_lines(0))
   end)
+
+  it('updates snippet state when built-in completion menu is visible', function()
+    test_expand_success({ '$1 = function($2)\n$3\nend' }, { ' = function()', '', 'end' })
+    -- Show the completion menu.
+    feed('<C-n>')
+    -- Make sure no item is selected.
+    feed('<C-p>')
+    -- Jump forward (the 2nd tabstop).
+    exec_lua('vim.snippet.jump(1)')
+    feed('foo')
+    eq({ ' = function(foo)', '', 'end' }, buf_lines(0))
+  end)
 end)


MariaSolOs:
Closes #25895.

PR #27621, opened 2024-02-25
fix(lsp): remove unnecessary file load/write when renaming

tomtomjhj:


runtime/lua/vim/lsp/util.lua
```diff
@@ -705,11 +705,12 @@ function M.rename(old_fname, new_fname, opts)
   end
 
   for _, b in ipairs(oldbufs) do
-    vim.fn.bufload(b)
     -- The there may be pending changes in the buffer
```
MariaSolOs, line 708: Since we're already changing this code, could we fix the typo in this comment?

MariaSolOs (commented):
@tomtomjhj Is there an associated bug? Could you please describe the problem and the fix?

tomtomjhj:
No, this is just a minor efficiency improvement. 

Previously `rename` would unconditionally read the to-be-renamed file from the disk and write it to the disk. This is redundant in some cases
* If the file is not already loaded, it's not attached to lsp client, so nvim doesn't need to care about this file. 
* If the file is loaded but has no change, it doesn't need to be written.

MariaSolOs:
@tomtomjhj got it. Even for improvements like this, make sure to explain the change in the description ;)

PR #27629, opened 2024-02-25
feat(tui): colorized underline in Windows Terminal

### LINKED ISSUE #26744: Implementing undercurl support in windows terminal
### Problem

Windows did not support underline syles other than the regular one. no undercurl or double line . but with the latest merge one month ago these has been implemented and planned for  https://github.com/microsoft/terminal/pull/16097.

in the latest canary you can see this in action using powershell >= 6  (Cmd as far as i know does not support escape chars. correct me if i am wrong )

using both 
```
Write-Host "`e[4:3mtest`e[0m"
Write-Host "`e[4:3;58:2::000:255:000mtest`e[0m"
```
will print both colored non colored undercurl 

image

but neovim in windows won't show these even when modifying the t_Ce and t_Cs variables

```
if vim.fn.has("termguicolors") == 1 then
    vim.o.termguicolors = true
    -- Enable undercurl
    vim.api.nvim_set_var('t_Cs', '\\e[4:3m')
    vim.api.nvim_set_var('t_Ce', '\\e[4:0m')
end
```

Can this be implemented early so when the final release is done this wil be ready?
the latest canary also supports all other styles of underline like dotted and double underlines.

### Expected behavior

to show undercurls and other styles of  underlines

ajmandourah:
Closes https://github.com/neovim/neovim/issues/26744

Windows did not support underline syles other than the regular one. no undercurl or double line . but with the latest merge these has been implemented and planned for https://github.com/microsoft/terminal/pull/16097.

in the latest Preview you can see this in action using powershell >= 6 (Cmd as far as i know does not support escape chars. correct me if i am wrong )

using both
```
Write-Host "`e[4:3mtest`e[0m"
Write-Host "`e[4:3;58:2::000:255:000mtest`e[0m"
```
will print both colored non colored undercurl

292871937-1c559811-e124-4354-8d67-cacc0f3f78a2

Adding the paramaters in the terminfo for vtpcon will give full support for underline style rendering in native windows. I assume this will also give support to other terminals other than windows terminal like wezterm and alacritty as  these are relient on windows terminal conptty implementation which should update as soon as the preview branch merge into the stable one. 

To test this out download the latest Windows terminal Preview from either github or Windows store.

image

Thanks to @tusharsnx and @kmoschcau for the inspiration.


src/nvim/tui/terminfo_defs.h
```diff
@@ -1,8 +1,14 @@
...
 // uncrustify:off
 
-// Generated by scripts/update_terminfo.sh and ncurses 6.4.20230520
+//
+// Generated by scripts/update_terminfo.sh and ncurses 6.3.20211021
+//
 
-#pragma once
+#ifndef NVIM_TUI_TERMINFO_DEFS_H
+#define NVIM_TUI_TERMINFO_DEFS_H
```
zeertzjq, line 11: ?

ajmandourah: Sorry seems like I mixed up the master br with the stable one as I could not compile the master one correctly with msvc for some reason hence I revert to stable.

Will fix this

src/nvim/tui/terminfo_defs.h
```diff
@@ -1,6 +1,6 @@
 // uncrustify:off
 
-// Generated by scripts/update_terminfo.sh and ncurses 6.4.20230520
+// Generated by scripts/update_terminfo.sh and ncurses 6.3.20211021
```
zeertzjq, line 3: Why is it still 6.3 here if you generated with updated ncurses? Although I ran `./scripts/update_terminfo.sh` on your branch and the only change is the version here.

ajmandourah: Honestly i don't know. I am using wsl to run the script, maybe something is wrong there. i will regenerate again with cygwin

zeertzjq: Hmm that seems to cause more problems. Let's revert the newest commit then.

scripts/windows.ti
```diff
@@ -59,6 +59,8 @@ vtpcon|ANIS emulation for console virtual terminal sequence with libuv,
 	sgr=\E[0%?%p1%p3%|%t;7%;%?%p2%t;4%;%?%p6%t;1%;m,
 	sgr0=\E[0m, smam@, smglp@, smgrp@, smglr@, smir@, smkx@, smm@, tbc@, u6@, u7@, u8@, u9@,
 	Cr@, Cs@, Ms@, XM@, kDC3@, kDC4@, kDC5@, kDC6@, kDC7@,
+        Smulx=\E[4:%p1%dm,
+        Setulc=\E[58::2::%p1%{65536}%/%d::%p1%{256}%/%{255}%&%d::%p1%{255}%&%d%;m,
```
zeertzjq, line 63:
```suggestion
	Smulx=\E[4:%p1%dm,
	Setulc=\E[58::2::%p1%{65536}%/%d::%p1%{256}%/%{255}%&%d::%p1%{255}%&%d%;m,
```

ajmandourah:
Thank you BTW for your time and for reviewing. And sorry for the mess created earlier on.

tusharsnx:
@ajmandourah Thanks for doing this! 


An important point we need to discuss is

Have you tested this on terminals that do not support colored underlines (`SGR 58/59`) and extended underline styles (`SGR 4:X`)?

The problem with these sub parameter (colon `:` syntax) based control sequences are, if we send the sequence to unsupported terminals, they'll not render anything.

Neovim, when not using `SGR 4:X` and `SGR 58`, sends standard `SGR 4` for all underlines (except double underline `SGR 21`), and don't even send `SGR 58` at all. This is important to keep unsupported terminals working as usual.

Unfortunately, feature detection in terminals isn't the best. You could've used [DECRQSS](https://vt100.net/docs/vt510-rm/DECRQSS.html) and [DECRPSS](https://vt100.net/docs/vt510-rm/DECRPSS.html), but that'll let you know the capability of backing conhost and not of the terminal user is interacting with 🥲

So, I think we should add a config like `compatibility.legacy_underline_styles` to help user revert to older behavior (i.e only `SGR 4` and no `SGR 4:X`). Sending `SGR 58` to unsupported terminals won't hurt I guess. I'm also not an nvim expert, so maybe there's already a way to turn off extended styles. In that case, mentioning that in the docs should be enough 🙂

ajmandourah:
@tusharsnx 

You are right.  here is a comparision

left: with current modification. current windows terminal (underline styles + colors not supported yet)
middle: with old terminfo. current windows terminal (underline styles + colors not supported yet)
right: with modifications. windows terminal preview (with support for underline styles and colors)

image

from my understanding other terminals like Putty for example have their own terminfo defined in the terminfo_defs.h I made sure not to touch those .

tusharsnx:
So, what we are looking for is to get _left_ to work as _middle_ with the current changeset. Can we edit `Smulx` (or something similar) to get left to still show all underlines?

ajmandourah:
> So, what we are looking for is to get _left_ to work as _middle_ with the current changeset. Can we (as a user) edit `Smulx` (or something similar) to get the terminal on the _left_ to show the missing underlines?

You are probably right but the question is if we really need to do so . I have just tried alacritty and wezterm, 2 of the most popular terminals for windows and they both shows the correct underline (middle). As Windows terminal will eventually will support underlines when they merge it into the stable which i recon to be soon, I doubt we will need to implement a feature which may not be used. 

image

again . other terminals like putty and conemu have their own terminfo defined in the header file so these probably wont be affected. did not try though .

tusharsnx:
> You are probably right but the question is if we really need to do so.

I think we need it because it's considered as a breaking change for unsupported terminals. This is probably why Neovim blocks this feature unless a terminal explicitly says (through terminfo, or similar methods) that it supports these extended control sequences.

> I doubt we will need to implement a feature which may not be used

This will used as a last resort for unsupported terminals. Though, we might not need it right now 🤔

*I'm okay with what core maintainers think is right for Neovim to do here.*

zeertzjq:
Enabling this unconditionally doesn't seem ideal. Is it possible for a program running in Windows Terminal to get the version of the terminal?

ajmandourah:
> Enabling this unconditionally doesn't seem ideal. Is it possible for a program running in Windows Terminal to get the version of the terminal?

There is `wt - v`. Will come back home and test it out.

ajmandourah:
After reading around in the Windows terminal (WT) repo, it seems like when running WT will set the env variable `%WT_SESSION%`. neovim can probably look it up and set the terminfo fallback to it.

ajmandourah:
Added another builtin fallback win-term terminfo which is mainly for WT.
I modified the terminfo.c with the following : 
```c++
} else if (terminfo_is_term_family(term, "vtpcon") && getenv("WT_SESSION")) {
  *termname = xstrdup("builtin_win_term");
  return unibi_from_mem((const char *)windows_terminal_terminfo,
                        sizeof windows_terminal_terminfo);
```
which specificly targeting WT. with this no more conflict is there with unsupported terminals for example the legacy command line .

image

Left: with the mentioned addition of a win-term terminfo fallback
Right: without (the vtpcon only modified terminfo)

notice that stylized underlines are not supported in the command prompt terminal .

src/nvim/tui/terminfo_defs.h
```diff
@@ -2118,6 +2118,210 @@ static const int8_t win32con_terminfo[] = {
...
 };
 
+// win-term|ANIS emulation for console virtual terminal sequence with libuv,
```
gpanders, line 2121:
```suggestion
// win-term|ANSI emulation for console virtual terminal sequence with libuv,
```

scripts/windows.ti
```diff
@@ -48,6 +48,29 @@ conemu|ANIS X3.64 and Xterm 256 colors for ConEmu with libuv,
...
+	kHOM3@, kHOM4@, kHOM5@, kHOM6@, kHOM7@,
+	kIC3@, kIC4@, kIC5@, kIC6@, kIC7@,
+	kLFT3@, kLFT4@, kLFT5@, kLFT6@, kLFT7@,
+	kNXT3@, kNXT4@, kNXT5@, kNXT6@, kNXT7@,
+	kPRV3@, kPRV4@, kPRV5@, kPRV6@, kPRV7@,
+	kRIT3@, kRIT4@, kRIT5@, kRIT6@, kRIT7@,
+	kUP3@, kUP4@, kUP5@, kUP6@, kUP7@, rmxx@, smxx@, xm@,
+	use=libuv+basekey, use=libuv+exkey, use=xterm+256color, use=xterm+sl, use=xterm-new,
+
+win-term|ANIS emulation for console virtual terminal sequence with libuv,
```
gpanders, line 73:
```suggestion
win-term|ANSI emulation for console virtual terminal sequence with libuv,
```

zeertzjq:
https://github.com/microsoft/terminal/issues/8303 mentions `ms-terminal` as the terminfo name.

ajmandourah:
> [microsoft/terminal#8303](https://github.com/microsoft/terminal/issues/8303) mentions `ms-terminal` as the terminfo name.

will change the name to ms-terminal.

on a side note. neovim still uses  vtpcon even for windows terminal . so technically we are virtualizing the terminal.

Kayzels:
I'm very new to Neovim, so unable to provide feedback on your commits.

However, I just tested this out on my machine. I can confirm it works perfectly in both Powershell 7 and in cmd. Colored and styled underlines show when AtlasEngine is enabled, and underlines without styles show when it is not.

image

scripts/windows.ti
ajmandourah: fixed

zeertzjq:
Another approach is to add a condition here instead of adding a terminfo entry:
```c
  // Extended underline.
  // terminfo will have Smulx for this (but no support for colors yet).
  tui->unibi_ext.set_underline_style = unibi_find_ext_str(ut, "Smulx");
  if (tui->unibi_ext.set_underline_style == -1) {
    int ext_bool_Su = unibi_find_ext_bool(ut, "Su");  // used by kitty
    if (vte_version >= 5102 || konsolev >= 221170
        || (ext_bool_Su != -1
            && unibi_get_ext_bool(ut, (size_t)ext_bool_Su))) {
      tui->unibi_ext.set_underline_style = (int)unibi_add_ext_str(ut, "ext.set_underline_style",
                                                                  "\x1b[4:%p1%dm");
    }
  }
```

ajmandourah:
> Another approach is to add a condition here instead of adding a terminfo entry:
> ```c
>   // Extended underline.
>   // terminfo will have Smulx for this (but no support for colors yet).
>   tui->unibi_ext.set_underline_style = unibi_find_ext_str(ut, "Smulx");
>   if (tui->unibi_ext.set_underline_style == -1) {
>     int ext_bool_Su = unibi_find_ext_bool(ut, "Su");  // used by kitty
>     if (vte_version >= 5102 || konsolev >= 221170
>         || (ext_bool_Su != -1
>             && unibi_get_ext_bool(ut, (size_t)ext_bool_Su))) {
>       tui->unibi_ext.set_underline_style = (int)unibi_add_ext_str(ut, "ext.set_underline_style",
>                                                                   "\x1b[4:%p1%dm");
>     }
>   }
> ```
> Since there are already several conditions, I'm not sure which solution is better.

Didn't go that deep to be honest. But it's as you said another possibility. Which can be added as failsafe.

I guess what really going to decide is the performance (which I think won't matter here as it is literally just a condition) and how extendable the approach in case new features are added. I don't know if there are other terminal features that WT does not support yet in comparison with Unix terminals.

I am open and up for suggestings.

ajmandourah:
after some testing it seems like adding Setulc caused some issue when scrolling around . was noticed when using nvim-tree plugin but I recon this will also affect multiple windows. Removing the parameter reserve the ability of windows terminal to render the underlines and the colors so only smlux was needed.

ajmandourah:
Is there anything left to be done here?

clason:
Fix the failing tests and squash your commits?

ajmandourah:
> Fix the failing tests and squash your commits?

thanks. I am new to git. will look it up

ajmandourah:
I tried changing getenv() to os_getenv() as recommended by lint. it compiles fine in visual studio but does not pass the test. any ideas to proceed?

src/nvim/tui/terminfo.c
```diff
@@ -120,6 +122,10 @@ static unibi_term *terminfo_builtin(const char *term, char **termname)
     *termname = xstrdup("builtin_conemu");
     return unibi_from_mem((const char *)conemu_terminfo,
                           sizeof conemu_terminfo);
+  } else if (terminfo_is_term_family(term, "vtpcon") && os_getenv("WT_SESSION")) {
```
tusharsnx, line 125: nit:
```suggestion
  } else if (terminfo_is_term_family(term, "vtpcon") && os_env_exists("WT_SESSION")) {
```

src/nvim/tui/terminfo.c
```diff
@@ -14,11 +14,13 @@
 #include "nvim/tui/terminfo.h"
 #include "nvim/tui/terminfo_defs.h"
 
+
 #ifdef __FreeBSD__
 # include "nvim/os/os.h"
 #endif
 
 #ifdef INCLUDE_GENERATED_DECLARATIONS
+# include "os/env.c.generated.h"
```
tusharsnx, line 23:
```suggestion

```

You should get rid of this include and instead remove the header gaurd (`#ifdef __FreeBSD__` and `#endif`) around the `nvim/os/os.h` include. That should include the required `os_getenv()` (or `os_env_exists()` as I suggested in another comment) correctly.

src/nvim/tui/terminfo.c
```diff
@@ -14,11 +14,13 @@
 #include "nvim/tui/terminfo.h"
 #include "nvim/tui/terminfo_defs.h"
 
+
```
tusharsnx, line 17:
```suggestion

```

Remove empty lines.

tusharsnx (commented):
Looks great.

You can apply the suggested changes and see if it helps.

src/nvim/tui/terminfo.c
ajmandourah: Thank you for your suggestions. will apply the changes and test it out

ajmandourah:
> Looks great.
> 
> You can apply the suggested changes and see if it helps.

it compiles in msvc. getting this on lint check

```
FAIL: /home/runner/work/neovim/neovim/src/nvim/tui/terminfo.c (File size changed from 8862 to 8861)
```

dundargoc:
Run `make format` locally.

ajmandourah:
> Run `make format` locally.

Thanks.
it passes lint now.

ajmandourah:
Now FreeBSD is complaingin but looks like an issue with the script.

zeertzjq:
There seems to still be some problems:
- The new terminfo entry is not added to `scripts/update_terminfo.sh`, so it will be removed when the script is next run.
- The new terminfo entry is mostly a copy of `vtpcon`, which may lead to future mistakes (e.g. only updating one entry when both should be updated).

So I'm starting to think https://github.com/neovim/neovim/pull/27629#issuecomment-1975770891 is a better approach. I've opened a PR #28035 for that.

zeertzjq:
Hmm, it seems that checking `WT_SESSION` environment variable still prevents undercurl from working in older Windows Terminal versions. I checked this in Windows Terminal 1.12.10983.0, and highlights with undercurl show neither underline or undercurl.

ajmandourah:
> Hmm, it seems that checking `WT_SESSION` environment variable still leads to worse behavior on older Windows Terminal versions. I checked this with Windows Terminal 1.12.10983.0, and a highlight group with undercurl shows neither an underline nor an undercurl, while without `WT_SESSION` it shows an underline.

Undercurl supported since 1.19. You are right older version will have this issue as there is no way to get the version installed through the environment variables. The only way I know of is through windows registry but this will be so messy and prone to changes from Microsoft side.

Updating windows terminal is the only way to go for now.
I will be on the lookout for ways to check what version is installed but seems like the devs are not interested into providing this information through env vars.

zeertzjq:
Also related: https://github.com/microsoft/terminal/issues/1040

PR #27632, opened 2024-02-25
fix(lsp): correct the error message's cmd on spawning

notomo:
This PR fixes to show cmd as string in error message.

📝 actual error message
```
Spawning language server with cmd: `table: 0x7fbe28326188` failed. The language server is either not installed, missing from PATH, or not executable.
```


MariaSolOs:
Maybe concatenating the command string array into a single string would be more readable?

notomo:
Thank you for reviewing!

I think the robust way is using vim.inspect.
The forllowin are reasons.
- to avoid error in error handling even if cmd is unexpected value
- to inform accurate cmd as it is

PR #27097, opened 2024-01-20
feat!: rewrite TOhtml in lua

altermo:
**The reasoning behind the rewrite**
1. Some neovim-specific decorations (mainly extmarks and treesitter) don't get handled by TOhtml.vim.
2. TOhtml.vim is written in Vim script, and to not make it slow, they needed to write some hard-to-understand code, and lua is so much faster, that there is not much need to use magic to write fast code.
3. TOhtml.vim is not updated that often (typically once every few years) so there's no need to worry about rewriting every new feature of TOhtml.vim in lua.

**What this rewrite is not**
As @clason described, this rewrite needs not to be a full rewrite of `TOhtml.vim` (with all the features), so it is not.

**What this rewrite is**
This rewrite has the purpose of creating a plugin which can convert a vim window with all it's decorations to html.

**The visual things that still don't work**
+ LSP semantic tokens that are not on the screen
+ Highlighted folds
+ More complex internal statuscol (number+relativenumer+{foldcolumn or signcolumn})
+ Right aligned extmarks
+ Extmarks' hl_mode
+ Spell
+ Priority for specific decorations
+ Less used highlight attributes (font, special color, reverse, underdouble, blend, ...)

Related:
+ https://github.com/neovim/neovim/issues/21629

PR #27656, opened 2024-02-28
fix(tohtml): set filetype of generated HTML to `html`

clason:
Problem: `:TOhtml` opens the generated HTML code in a split, meaning it
inherits the `help` filetype if a help buffer is to be converted.

Solution: Explicitly set the filetype to `html`.

Fixup for #27097


clason:
Note: The issue only seems to occur for `help` buffers; anything else already works.

PR #27660, opened 2024-02-28
test: silence expected errors

dundargoc:
This will remove unrelated errors in .nvimlog at the end of test output.

### DIFF
--- a/src/nvim/event/libuv_process.c
+++ b/src/nvim/event/libuv_process.c
@@ -90,7 +90,7 @@ int libuv_process_spawn(LibuvProcess *uvproc)
 
   int status;
   if ((status = uv_spawn(&proc->loop->uv, &uvproc->uv, &uvproc->uvopts))) {
-    ELOG("uv_spawn(%s) failed: %s", uvproc->uvopts.file, uv_strerror(status));
+    DLOG("uv_spawn(%s) failed: %s", uvproc->uvopts.file, uv_strerror(status));
     if (uvproc->uvopts.env) {
       os_free_fullenv(uvproc->uvopts.env);
     }
--- a/test/functional/autocmd/autocmd_oldtest_spec.lua
+++ b/test/functional/autocmd/autocmd_oldtest_spec.lua
@@ -7,9 +7,17 @@ local api = helpers.api
 local fn = helpers.fn
 local exec = helpers.exec
 local feed = helpers.feed
+local assert_log = helpers.assert_log
+local is_os = helpers.is_os
+
+local testlog = 'Xtest_autocmd_oldtest_log'
 
 describe('oldtests', function()
   before_each(clear)
+
+  after_each(function()
+    os.remove(testlog)
+  end)
 
   local exec_lines = function(str)
     return fn.split(fn.execute(str), '\n')
@@ -49,6 +57,7 @@ describe('oldtests', function()
   end)
 
   it('should fire on unload buf', function()
+    clear({ env = { NVIM_LOG_FILE = testlog } })
     fn.writefile({ 'Test file Xxx1' }, 'Xxx1')
     fn.writefile({ 'Test file Xxx2' }, 'Xxx2')
     local fname = 'Xtest_functional_autocmd_unload'
@@ -81,6 +90,10 @@ describe('oldtests', function()
     fn.delete('Xxx2')
     fn.delete(fname)
     fn.delete('Xout')
+
+    if is_os('win') then
+      assert_log('stream write failed. RPC canceled; closing channel', testlog)
+    end
   end)
 
   -- oldtest: Test_delete_ml_get_errors()
--- a/test/functional/options/defaults_spec.lua
+++ b/test/functional/options/defaults_spec.lua
@@ -865,6 +865,10 @@ describe('XDG defaults', function()
 end)
 
 describe('stdpath()', function()
+  after_each(function()
+    os.remove(testlog)
+  end)
+
   -- Windows appends 'nvim-data' instead of just 'nvim' to prevent collisions
   -- due to XDG_CONFIG_HOME, XDG_DATA_HOME and XDG_STATE_HOME being the same.
   local function maybe_data(name)
@@ -890,7 +894,7 @@ describe('stdpath()', function()
 
   it('reacts to $NVIM_APPNAME', function()
     local appname = 'NVIM_APPNAME_TEST' .. ('_'):rep(106)
-    clear({ env = { NVIM_APPNAME = appname } })
+    clear({ env = { NVIM_APPNAME = appname, NVIM_LOG_FILE = testlog } })
     eq(appname, fn.fnamemodify(fn.stdpath('config'), ':t'))
     eq(appname, fn.fnamemodify(fn.stdpath('cache'), ':t'))
     eq(maybe_data(appname), fn.fnamemodify(fn.stdpath('log'), ':t'))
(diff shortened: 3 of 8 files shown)


zeertzjq:
What is the definition of "expected errors"?

justinmk:
> What is the definition of "expected errors"?

Errors that are intentinoal/expected in the tests. To resolve this, tests should define a local NVIM_LOG_FILE and use `assert_log` to check that the expected messages were logged.

Changing `ELOG` to `ILOG` isn't generally the right way to fix this, unless a particular message truly isn't considered an error.

dundargoc:
> To resolve this, tests should define a local NVIM_LOG_FILE and use `assert_log` to check that the expected messages were logged.

Yeah, I'm working on it. That's what I did but some problems showed up so it's still work in progress.
 
> Changing `ELOG` to `ILOG` isn't generally the right way to fix this, unless a particular message truly isn't considered an error.

The problem with these specific logs was that we're already testing the error output. That and some tests straight up fail if we change the log file.

dundargoc:
OK, this is a new one. There is an expected error on mac and linux but not windows, so if I assert the log the tests will fail on windows. @justinmk what do?

src/nvim/event/libuv_process.c
```diff
@@ -90,7 +90,7 @@ int libuv_process_spawn(LibuvProcess *uvproc)
 
   int status;
   if ((status = uv_spawn(&proc->loop->uv, &uvproc->uv, &uvproc->uvopts))) {
-    ELOG("uv_spawn(%s) failed: %s", uvproc->uvopts.file, uv_strerror(status));
+    DLOG("uv_spawn(%s) failed: %s", uvproc->uvopts.file, uv_strerror(status));
```
justinmk, line 93: Can this be `ILOG` ? Why would we want to ignore this error?

test/functional/autocmd/autocmd_oldtest_spec.lua
```diff
@@ -81,6 +90,10 @@ describe('oldtests', function()
     fn.delete('Xxx2')
     fn.delete(fname)
     fn.delete('Xout')
+
+    if is_os('win') then
+      assert_log('stream write failed. RPC canceled; closing channel', testlog)
```
justinmk, line 95: This is definitely an improvement, thanks. Eventually, do we still want to fix this at a lower level? I.e. if this error truly doesn't matter then we shouldn't log it (for specific cases on Windows).

test/functional/options/defaults_spec.lua
```diff
@@ -928,6 +932,9 @@ describe('stdpath()', function()
     -- Valid appnames:
     test_appname('a/b', 0)
     test_appname('a/b\\c', 0)
+    if not is_os('win') then
+      assert_log('Failed to start server: no such file or directory:', testlog)
```
justinmk, line 936: 👍  NVIM_APPNAME does not affect server name on Windows.

test/functional/terminal/tui_spec.lua
```diff
@@ -3120,6 +3127,9 @@ describe('TUI as a client', function()
 
     client_super:close()
     server:close()
+    if is_os('mac') then
+      assert_log('uv_tty_set_mode failed: Unknown system error %-102', testlog)
```
justinmk, line 3131: Could be a bug. But at least now it's "documented" in our tests.

justinmk:
Thanks for this. Now should we _fail_ CI if the logs are _not_ empty?

P.S. before/after examples are helpful for this kind of change (though the "after" in this case is "empty"). 

before:

```
------------------------------------------------------------------------------
$NVIM_LOG_FILE: /Users/runner/work/neovim/neovim/build/.nvimlog
(last 100 lines)
ERR 2024-03-28T21:59:26.081 nvim.14952.0 chan_close_with_error:560: RPC: failed to decode msgpack
ERR 2024-03-28T21:59:26.082 nvim.14952.0 chan_close_with_error:560: RPC: (null)
ERR 2024-03-28T21:59:27.057 T1409.15103.0 libuv_process_spawn:93: uv_spawn(/bin/pwd) failed: permission denied
ERR 2024-03-28T21:59:32.183 T1451.16112.0 libuv_process_spawn:93: uv_spawn(nosuchshell) failed: no such file or directory
ERR 2024-03-28T21:59:32.184 T1451.16112.0 libuv_process_spawn:93: uv_spawn(nosuchshell) failed: no such file or directory
ERR 2024-03-28T21:59:35.382 T1517.16548.0/c os_set_cloexec:493: Failed to get flags on descriptor 3: Bad file descriptor
WRN 2024-03-28T22:01:07.537 T3802.20024.0/c server_start:163: Failed to start server: no such file or directory: /Users/runner/work/neovim/neovim/build/Xtest_tmpdir/a%b.runner/rGXrSF/a/
WRN 2024-03-28T22:01:07.547 T3802.20024.0/c server_start:163: Failed to start server: no such file or directory: /Users/runner/work/neovim/neovim/build/Xtest_tmpdir/a%b%c.runner/GQgM6J/a/
ERR 2024-03-28T22:02:23.005 nvim.23439.0 tui_mode_change:1284: uv_tty_set_mode failed: Unknown system error -102
ERR 2024-03-28T22:02:34.372 T5248.23805.0 decor_provider_error:38: error in provider ns1.start: Error executing lua: [string "<nvim>"]:4: Foo
stack traceback:
	[C]: in function 'error'
	[string "<nvim>"]:4: in function <[string "<nvim>"]:3>
ERR 2024-03-28T22:02:34.372 T5248.23805.0 decor_provider_error:38: error in provider ns1.end: Error executing lua: [string "<nvim>"]:4: Foo
stack traceback:
	[C]: in function 'error'
	[string "<nvim>"]:4: in function <[string "<nvim>"]:3>
ERR 2024-03-28T22:02:44.192 T5422.24446.0 os_set_cloexec:493: Failed to get flags on descriptor 3: Bad file descriptor
ERR 2024-03-28T22:03:58.598 T7038.27015.0 libuv_process_spawn:93: uv_spawn(fake_shell) failed: no such file or directory
```

test/functional/options/defaults_spec.lua
dundargoc: Uhh, isn't that the same thing? Maybe the intent is clearer you mean?

test/functional/autocmd/autocmd_oldtest_spec.lua
dundargoc: I have no clue, I'm not sure what's happening here currently.

test/functional/options/defaults_spec.lua
justinmk: yes, by "instead" I meant we could change the test code (not the behavior).

```
if is_os('win') then
  assert_nolog(...)
else
  assert_log('Failed to start server: no such file or directory:', testlog)
```

dundargoc:
> P.S. before/after examples are helpful for this kind of change (though the "after" in this case is "empty").

Eh? I'll keep it in mind for future, but if I had read that snippet in the commit log I'd probably think "no one asked, why are you telling me this?"

src/nvim/event/libuv_process.c
dundargoc: Changed to `ILOG` in https://github.com/neovim/neovim/commit/ae28ef327e02ac87ef26f941c401312ed0462d8c.

test/functional/autocmd/autocmd_oldtest_spec.lua
dundargoc: Hmm, what makes this specific error different compared to the other errors we "ignore" with `assert_log`. Aka, what makes this particular error different?

PR #27662, opened 2024-02-28
vim-patch:9.1.0142: getregion() can be improved

zeertzjq:
#### vim-patch:9.1.0142: getregion() can be improved

Problem:  getregion() can be improved (after v9.1.120)
Solution: change getregion() implementation to use pos as lists and
          one optional {opt} dictionary (Shougo Matsushita)

Note: The following is a breaking change!

Currently, the getregion() function (included as of patch v9.1.120) takes
3 arguments: the first 2 arguments are strings, describing a position,
arg3 is the type string.

However, that is slightly inflexible, there is no way to specify
additional arguments. So let's instead change the function signature to:

getregion(pos1, pos2 [, {Dict}]) where both pos1 and pos2 are lists.
This is slightly cleaner, and gives us the flexibility to specify
additional arguments as key/value pairs to the optional Dict arg.

Now it supports the "type" key to specify the selection type
(characterwise, blockwise or linewise) and now in addition one can also
define the selection type, independently of what the 'selection' option
actually is.

Technically, this is a breaking change, but since the getregion()
Vimscript function is still quite new, this should be fine.

closes: vim/vim#14090

https://github.com/vim/vim/commit/19b718828d8d5fab52d94c6cdba694641879ab38

Co-authored-by: Shougo Matsushita <Shougo.Matsu@gmail.com>

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -2918,14 +2918,25 @@ getreginfo([{regname}])                                           *getreginfo()*
 		If {regname} is not specified, |v:register| is used.
 		The returned Dictionary can be passed to |setreg()|.
 
-getregion({pos1}, {pos2}, {type})                                  *getregion()*
-		Returns the list of strings from {pos1} to {pos2} as if it's
-		selected in visual mode of {type}.
-		For possible values of {pos1} and {pos2} see |line()|.
-		{type} is the selection type:
-			"v" for |charwise| mode
-			"V" for |linewise| mode
-			"<CTRL-V>" for |blockwise-visual| mode
+getregion({pos1}, {pos2} [, {opts}])                               *getregion()*
+		Returns the list of strings from {pos1} to {pos2} in current
+		buffer.
+
+		{pos1} and {pos2} must both be |List|s with four numbers.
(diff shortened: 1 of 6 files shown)


clason:
_now_ can we replace `vim.region`? 🥺

zeertzjq:
I think its blockwise part can be replaced with `getregion()` as it requires getting lines from the buffer anyway. Maybe charwise part too. But for the linewise part `vim.region()` is faster as it doesn't get text from the buffer.

clason:
Yeah, linewise is fine. It's only blockwise that runs into problems due to how Vim and Lua count multibyte character widths differently.

zeertzjq:
> I think its blockwise part can be replaced with `getregion()` as it requires getting lines from the buffer anyway. Maybe charwise part too. But for the linewise part `vim.region()` is faster as it doesn't get text from the buffer.

Oh, actually for the blockwise part there is still a whitespace padding problem: when only a part of a TAB is selected, `getregion()` returns spaces for those cells, which take more bytes than a TAB. That's probably still better than the current behavior, and the padding behavior may be changed by an optional argument in the future.

PR #27667, opened 2024-02-29
docs: fix minor grammar mistake

exformation:
I wanted to make it "preventing autocommands from being set multiple times" but that would be passive voice. Hopefully this wording is okay.


clason:
Thank you, but I actually (unsurprisingly prefer) the original phrasing (your breaks from "for organizing" with "to prevent").

PR #27673, opened 2024-02-29
fix(lsp): fold non-list-like iterator for protocol values

### LINKED ISSUE #27669: 853f647 breaks textDocument.completion.completionItemKind.valueSet (at least for gopls)
### Problem

Commit 853f647 refactors how the values for `textDocument.completion.completionItemKind.valueSet` are calculated. Before this commit the value was
```
valueSet = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25 }
```
i.e. a simple array. With this commit it changed into a complex type:
```
valueSet = { { 1, "Text",
  <metatable> = <1>{}
}, { 2, "Method",
  <metatable> = <table 1>
}, { 3, "Function",
  <metatable> = <table 1>
}, { 4, "Constructor",
  <metatable> = <table 1>
}, { 5, "Field",
  <metatable> = <table 1>
}, { 6, "Variable",
  <metatable> = <table 1>
}, { 7, "Class",
  <metatable> = <table 1>
}, { 8, "Interface",
  <metatable> = <table 1>
}, { 9, "Module",
  <metatable> = <table 1>
}, { 10, "Property",
  <metatable> = <table 1>
}, { 11, "Unit",
  <metatable> = <table 1>
}, { 12, "Value",
  <metatable> = <table 1>
}, { 13, "Enum",
  <metatable> = <table 1>
}, { 14, "Keyword",
  <metatable> = <table 1>
}, { 15, "Snippet",
  <metatable> = <table 1>
}, { 16, "Color",
  <metatable> = <table 1>
}, { 17, "File",
  <metatable> = <table 1>
}, { 18, "Reference",
  <metatable> = <table 1>
}, { 19, "Folder",
  <metatable> = <table 1>
}, { 20, "EnumMember",
  <metatable> = <table 1>
}, { 21, "Constant",
  <metatable> = <table 1>
}, { 22, "Struct",
  <metatable> = <table 1>
}, { 23, "Event",
  <metatable> = <table 1>
}, { 24, "Operator",
  <metatable> = <table 1>
}, { 25, "TypeParameter",
  <metatable> = <table 1>
} }
```
and when sending this to `gopls` it will respond with
```
RPC[Error] code_name = ParseError, message = "JSON RPC parse error: json: cannot unmarshal array into Go struct field
 ClientSymbolKindOptions.capabilities.workspace.symbol.symbolKind.valueSet of type protocol.SymbolKind
```

P.S.: This most likely also affects the other `valueSet` elements that use the new `get_value_set()` function but `gopls` will stop on the first error from the top.

```lua
local pattern = "go"
local cmd = { "gopls" }
local root_markers = { ".git", ".editorconfig" }

	pattern = pattern,
	callback = function(args)
		vim.lsp.start({
			name = "gopls",
			cmd = cmd,
			root_dir = root_dir,
			settings = settings,
		})
	end,
})
```

### Expected behavior

The language server should not complain about values sent to it and start regularly.

### Neovim version (nvim -v)

v0.10.0-dev-2480+g0eaae1bc0

gopls 0.15.1

### Operating system/version

Arch Linux

### Log file

_No response_

clason:
Fixup for #27628

Closes #27669

### DIFF
--- a/runtime/lua/vim/lsp/protocol.lua
+++ b/runtime/lua/vim/lsp/protocol.lua
@@ -1,9 +1,17 @@
 --- @diagnostic disable: duplicate-doc-alias
 
+-- TODO(clason) can be simplified after reverse lookup is removed
+---@param t table<any, any>
+---@return number[]
 local function get_value_set(t)
-  return vim.iter.filter(function(i)
-    return type(i) == 'number'
-  end, ipairs(t))
+  local result = {}
+  for _, v in pairs(t) do
+    if type(v) == 'number' then
+      table.insert(result, v)
+    end
+  end
+  table.sort(result)
+  return result
 end
 
 -- Protocol for the Microsoft Language Server Protocol (mslsp)


runtime/lua/vim/lsp/protocol.lua
```diff
@@ -1,9 +1,16 @@
 --- @diagnostic disable: duplicate-doc-alias
 
+-- TODO(clason) can be simplified after reverse lookup is removed
 local function get_value_set(t)
```
mfussenegger, line 4:
```suggestion
-- TODO(clason) can be simplified after reverse lookup is removed
---@param t table<any, any>
---@return number[]
local function get_value_set(t)
```

A plain loop variant would also work:

```
local function get_value_set(t)
  local result = {}
  for _, v in pairs(t) do
    if type(v) == "number" then
      table.insert(result, v)
    end
  end
  return result
end
```

But I guess either is fine.

mfussenegger (approved):

PR #27674, opened 2024-02-29
fix(snippet): correct indent with newline

glepnir:
Problem: snippet newline use before line indent after expand.

Solution: it should level + 1.

### DIFF
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -446,14 +446,18 @@ function M.expand(input)
       base_indent = base_indent .. (snippet_lines[#snippet_lines]:match('(^%s*)%S') or '') --- @type string
     end
 
+    local shiftwidth = vim.fn.shiftwidth()
+    local curbuf = vim.api.nvim_get_current_buf()
+    local expandtab = vim.bo[curbuf].expandtab
     local lines = vim.iter.map(function(i, line)
       -- Replace tabs by spaces.
-      if vim.o.expandtab then
-        line = line:gsub('\t', (' '):rep(vim.fn.shiftwidth())) --- @type string
+      if expandtab then
+        line = line:gsub('\t', (' '):rep(shiftwidth)) --- @type string
       end
       -- Add the base indentation.
       if i > 1 then
-        line = base_indent .. line
+        line = #line ~= 0 and base_indent .. line
+          or (expandtab and (' '):rep(shiftwidth) or '\t'):rep(vim.fn.indent('.') / shiftwidth + 1)
       end
       return line
     end, ipairs(text_to_lines(text)))
--- a/test/functional/lua/snippet_spec.lua
+++ b/test/functional/lua/snippet_spec.lua
@@ -5,6 +5,7 @@ local clear = helpers.clear
 local eq = helpers.eq
 local exec_lua = helpers.exec_lua
 local feed = helpers.feed
+local api = helpers.api
 local fn = helpers.fn
 local matches = helpers.matches
 local pcall_err = helpers.pcall_err
@@ -230,7 +231,7 @@ describe('vim.snippet', function()
   end)
 
   it('updates snippet state when built-in completion menu is visible', function()
-    test_expand_success({ '$1 = function($2)\n$3\nend' }, { ' = function()', '', 'end' })
+    test_expand_success({ '$1 = function($2)\nend' }, { ' = function()', 'end' })
     -- Show the completion menu.
     feed('<C-n>')
     -- Make sure no item is selected.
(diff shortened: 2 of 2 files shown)


clason:
Please don't make unrelated changes in a commit; it makes it harder to review. (If you want to include these changes here, at least put them in a separate `refactor` commit.)

runtime/lua/vim/snippet.lua
```diff
@@ -1,13 +1,14 @@
 local G = vim.lsp._snippet_grammar
-local snippet_group = vim.api.nvim_create_augroup('vim/snippet', {})
-local snippet_ns = vim.api.nvim_create_namespace('vim/snippet')
+local api = vim.api
```
clason, line 2: This is just adding noise with noticeable benefit.

test/functional/lua/snippet_spec.lua
```diff
@@ -230,14 +231,20 @@ describe('vim.snippet', function()
   end)
 
   it('updates snippet state when built-in completion menu is visible', function()
-    test_expand_success({ '$1 = function($2)\n$3\nend' }, { ' = function()', '', 'end' })
+    local curbuf = api.nvim_get_current_buf()
+    api.nvim_set_option_value('sw', 2, { buf = curbuf })
```
clason, line 235: This should be a separate test, with its own description.

runtime/lua/vim/snippet.lua
glepnir: onelese table lookup . it's somewhat trivial.

test/functional/lua/snippet_spec.lua
glepnir: I just feel this case is enough. But there is no harm in adding a newline of tabs and spaces.

clason: The point is about being explicit _what_ each test is testing. Testing multiple things at the same time means failures will be harder to interpret.

runtime/lua/vim/snippet.lua
clason: Yes, it's trivial, and the table lookup is not a bottleneck in this code (unless benchmarks say otherwise).

And now I have to wade through a long diff to find the ones that make the _actual_ change described in the PR.

test/functional/lua/snippet_spec.lua
glepnir: emm seems like my test case is missing, stash restore my case lol i will readd it.

runtime/lua/vim/snippet.lua
```diff
@@ -446,14 +446,18 @@ function M.expand(input)
       base_indent = base_indent .. (snippet_lines[#snippet_lines]:match('(^%s*)%S') or '') --- @type string
     end
 
+    local sw = vim.fn.shiftwidth()
```
MariaSolOs, line 449: Nit, but I prefer to keep the full-variable-name convention in this file:
```suggestion
    local shiftwidth = vim.fn.shiftwidth()
```

MariaSolOs, line 449: Why use `vim.fn.shiftwidth` here and not `vim.bo[curbuf].shiftwidth` (as we do for `expandtab`)?

test/functional/lua/snippet_spec.lua
```diff
@@ -230,14 +231,29 @@ describe('vim.snippet', function()
...
```
MariaSolOs, line 235: Look at the signature of `test_expand_success`. You can pass a settings parameter instead of setting the options like this.
https://github.com/neovim/neovim/blob/e1af3817958a2b9a818fcb10eec9f1197bd62dfd/test/functional/lua/snippet_spec.lua#L71

test/functional/lua/snippet_spec.lua
```diff
@@ -230,14 +231,29 @@ describe('vim.snippet', function()
   end)
 
   it('updates snippet state when built-in completion menu is visible', function()
```
MariaSolOs, line 233: Why are you modifying this test? These are _unit tests_ and each one should check a single functionality.

MariaSolOs (changes requested):

runtime/lua/vim/snippet.lua
lucario387: `vim.fn.shiftwidth` also handles if `expandtab` is not enabled

lucario387: so it will use `tabstop` or `shiftwidth` depending on the input

test/functional/lua/snippet_spec.lua
glepnir: because the snippet in this case also have newline

MariaSolOs: Most of them do, but this specific test should just check that the built-in completion doesn't affect snippet navigation. Adding extra settings and checks will make investigating any future test failures harder, since the unit test is verifying multiple things.

glepnir: i had remove the newline in this case because we do have a single it for newlines so no need change here

test/functional/lua/snippet_spec.lua
glepnir: oh right. i saw it :)

test/functional/lua/snippet_spec.lua
MariaSolOs: I still don't think there's need to modify this test :)

glepnir: Is newline important in this case? This is irrelevant if you just want to verify what happens when pum is visible. Otherwise, either add appropriate indent settings or modify newline's expect ❓

MariaSolOs: But _real life_ snippets have new lines. This was a regression test added when fixing https://github.com/neovim/neovim/issues/25895, and I used the same example that echasnovski used.

I just find it odd to change this test when the bugfix is unrelated to it. The test is passing, and checks a specific behaviour independent from this PR.

MariaSolOs (approved):

mfussenegger (approved):

PR #27680, opened 2024-02-29
Fix `cmake -G Xcode` for new Xcode build system

kaii-zen:
According to `BUILD.md`, this is supposed to work, but when I tried I got the following error:

```
CMake Error in src/nvim/CMakeLists.txt:
  The custom command generating

    /Users/kaiizen/Dev/github/kaii-zen/neovim/build/cmake.config/auto/versiondef.h

  is attached to multiple targets:

    nvim_bin
    libnvim
    generated-sources

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

Apparently CMake now defaults to Xcode's new build system; and, until Xcode 15, it was still possible to opt back to the old build system to avoid these errors. Xcode 15, however, no longer supports the old build system.

I managed to appease CMake by adding some `add_dependencies()` to fix the above error and the similar ones that followed and the resulting Xcode project seems to build successfully.

Since I'm not super familiar with CMake or the Neovim build, I kept all changes conditional to only when the Xcode generator is used, to avoid potentially affecting other use cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,11 @@ if(POLICY CMP0092)
   cmake_policy(SET CMP0092 NEW)
 endif()
 
+# Support generating for new Xcode build system
+if(CMAKE_GENERATOR STREQUAL "Xcode" AND POLICY CMP0114)
+  cmake_policy(SET CMP0114 NEW)
+endif()
+
 project(nvim C)
 
 if(POLICY CMP0135)
--- a/src/nvim/CMakeLists.txt
+++ b/src/nvim/CMakeLists.txt
@@ -902,6 +902,11 @@ add_custom_target(generated-sources DEPENDS
   ${NVIM_GENERATED_SOURCES}
 )
 
+if(CMAKE_GENERATOR STREQUAL "Xcode")
+  add_dependencies(nvim_bin generated-sources)
+  add_dependencies(libnvim generated-sources)
+endif()
+
 set(VIMDOC_FILES
   ${NVIM_RUNTIME_DIR}/doc/api.txt
   ${NVIM_RUNTIME_DIR}/doc/diagnostic.txt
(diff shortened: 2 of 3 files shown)


clason:
It works with the current CLI developer tools, at least for me. Can you test these?

kaii-zen:
> It works with the current CLI developer tools, at least for me. Can you test these?

You mean like this?

```
kaiizen@kaiibook build % sudo xcode-select --switch /Library/Developer/CommandLineTools

kaiizen@kaiibook build % cmake -G Xcode ..
-- The C compiler identification is unknown
CMake Error at CMakeLists.txt:19 (project):
  No CMAKE_C_COMPILER could be found.



-- Configuring incomplete, errors occurred!
```

clason:
Don't use the Xcode generator, then, of course.

dundargoc:
Changing target dependencies based on generator sounds like an incorrect approach. I'm not sure what cmake means by "common dependency", I'll need to investigate what they're referring to.

dundargoc:
OK, so it turns out that the xcode generator for some reason doesn't allow having multiple targets depend on the same generated file, so each `add_custom_command` will need their own dedicated target in order to make xcode work. This will likely require significant refactoring in order to work properly. I don't think this PR is the correct approach unfortunately.

I suggest we simply disallow the xcode generator with an error message for the time being.

kaii-zen:
> OK, so it turns out that the xcode generator for some reason doesn't allow having multiple targets depend on the same generated file, so each `add_custom_command` will need their own dedicated target in order to make xcode work. This will likely require significant refactoring in order to work properly. I don't think this PR is the correct approach unfortunately.
> 
> I suggest we simply disallow the xcode generator with an error message for the time being.

So it wants a single target depending on the `add_custom_command` and all other targets depending on that target?
I agree it doesn't make sense to alter the dependency graph for all build systems because Apple didn't think theirs through (hence the conditionals). Also, iirc, while diving into this rabbit hole I *think* I read somewhere that Xcode actually no longer has this limitation; if that's the case then this issue may simply go away once CMake catches up.

dundargoc:
> So it wants a single target depending on the add_custom_command and all other targets depending on that target?

Yes, that's how I interpret it.

dundargoc:
Superseded by https://github.com/neovim/neovim/pull/27715

PR #22673, opened 2023-03-15
fix(float): remove -1 in height clamp

Diablo-D3:
Problem: Clamp for height in floating windows enforced no more than editor height - 1, disallowing full editor height floating windows when using cmdheight=0

Solution: Clamp to full height, removing the -1. Tested to give the intended results with cmdheight=0, 1, or more than 1. This also inadvertently fixes a rendering error with cmdheight >1 where the bottom border would be overlapped by the cmdline.

PR #27687, opened 2024-03-01
fix(float): allow floating window in cmdline area

luukvbaal:
Problem:  Floats are arbitrarily positioned at 1 row above screen size.
Solution: Position at 1 row above 'cmdheight', only if window is hidden behind the message area.


clason:
Related: https://github.com/neovim/neovim/pull/22673

bfredl:
ye, with explicitly set zindex, allowing overlap with message area makes sense.

PR #27688, opened 2024-03-01
fix(lsp): defer writing error msgs

tomtomjhj:
Context:
Nvim catches errors from the user's `on_exit`callback and prints the error message.

Problem:
Printing the error message uses Nvim api functions. But `on_exit` runs in `:h lua-loop-callbacks` where most of `vim.api` is not allowed, so Nvim itself raises error.

Solution:
`vim.schedule()` the error reporting when necessary.

### DIFF
--- a/runtime/lua/vim/lsp/client.lua
+++ b/runtime/lua/vim/lsp/client.lua
@@ -603,8 +603,16 @@ local wait_result_reason = { [-1] = 'timeout', [-2] = 'interrupted', [-3] = 'err
 ---
 --- @param ... string List to write to the buffer
 local function err_message(...)
-  api.nvim_err_writeln(table.concat(vim.tbl_flatten({ ... })))
-  api.nvim_command('redraw')
+  local message = table.concat(vim.tbl_flatten({ ... }))
+  if vim.in_fast_event() then
+    vim.schedule(function()
+      api.nvim_err_writeln(message)
+      api.nvim_command('redraw')
+    end)
+  else
+    api.nvim_err_writeln(message)
+    api.nvim_command('redraw')
+  end
 end
 
 --- @private


mfussenegger (commented):
I think the other code paths that lead to `self:write_error` could have the same problem. E.g in rpc the `handle_body` method.

What do you think of instead integrating the `in_fast_event` check into the `write_error` method?

tomtomjhj:
Thanks for the comment. I additionally noticed that `err_message` is directly used in `Client:_on_error`, so I moved the check to `err_message` instead.

MariaSolOs:
> Thanks for the comment. I additionally noticed that `err_message` is directly used in `Client:_on_error`, so I moved the check to `err_message` instead.

Could logging also run `lua-loop-callbacks`? Although the default implementation doesn't seem to do so, perhaps someone could set a custom logger or we might change the core implementation, requiring this change to live in `Client:_on_error`.

tomtomjhj:
>  requiring this change to live in Client:_on_error

`Client:_run_callbacks` also calls `Client:write_error` (which calls `err_message`), so `in_fast_event` check should be added there too.  

> perhaps someone could set a custom logger or we might change the core implementation

IMO, the custom loggers should `vim.schedule` themselves, similarly to custom `on_exit`. But I don't have a strong preference on this.

runtime/lua/vim/lsp/client.lua
```diff
@@ -603,8 +603,16 @@ local wait_result_reason = { [-1] = 'timeout', [-2] = 'interrupted', [-3] = 'err
...
-  api.nvim_command('redraw')
+  local message = table.concat(vim.tbl_flatten({ ... }))
+  if vim.in_fast_event() then
+    vim.schedule(function()
+      api.nvim_err_writeln(message)
+      api.nvim_command('redraw')
+    end)
+  else
+    api.nvim_err_writeln(message)
+    api.nvim_command('redraw')
```
justinmk, line 614: instead of checking vim.in_fast_event(), why not always schedule() it?

tomtomjhj: Hmm, good point. 

`err_message` is called while handling errors from user-provided callbacks (`on_init`, ...). Some of them run while the context is not in_fast_event, and can make use of that fact, e.g., call `nvim_err_writeln` without deferring. In that case, the error messages (from user's code and `err_message`) can be in out of order. 

That's unlikely to be an actual problem, but in general I prefer to avoid deferral if possible.

btw there's a similar case here
https://github.com/neovim/neovim/blob/9319d97679053a219b1701d70abce414d0f6a8b0/runtime/lua/vim/lsp/log.lua#L25-L33

PR #27699, opened 2024-03-02
fix(lsp): create codelens request parameters for each buffer

cstoku:
The codelens `refresh()` processes the specified buffer or all loaded buffers.
However, it does not create a textDocument parameter for each buffer and does not specify the buffer when creating the textDocument parameter, so the URI of the current buffer is always specified.

To fix this, create a textDocument parameter for each buffer while respecting `bufnr` .

### DIFF
--- a/runtime/lua/vim/lsp/codelens.lua
+++ b/runtime/lua/vim/lsp/codelens.lua
@@ -299,12 +299,12 @@ function M.refresh(opts)
   local bufnr = opts.bufnr and resolve_bufnr(opts.bufnr)
   local buffers = bufnr and { bufnr }
     or vim.tbl_filter(api.nvim_buf_is_loaded, api.nvim_list_bufs())
-  local params = {
-    textDocument = util.make_text_document_params(),
-  }
 
   for _, buf in ipairs(buffers) do
     if not active_refreshes[buf] then
+      local params = {
+        textDocument = util.make_text_document_params(buf),
+      }
       active_refreshes[buf] = true
       vim.lsp.buf_request(buf, ms.textDocument_codeLens, params, M.on_codelens)
     end
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -3879,6 +3879,92 @@ describe('LSP', function()
         end,
       }
     end)
+
+    it('refresh multiple buffers', function()
+      local lens_title_per_fake_uri = {
+        ['file:///fake/uri1'] = 'Lens1',
+        ['file:///fake/uri2'] = 'Lens2',
+      }
+      clear()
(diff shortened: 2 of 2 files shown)


runtime/lua/vim/lsp/codelens.lua
```diff
@@ -299,12 +299,12 @@ function M.refresh(opts)
   local bufnr = opts.bufnr and resolve_bufnr(opts.bufnr)
   local buffers = bufnr and { bufnr }
     or vim.tbl_filter(api.nvim_buf_is_loaded, api.nvim_list_bufs())
-  local params = {
-    textDocument = util.make_text_document_params(),
-  }
 
   for _, buf in ipairs(buffers) do
     if not active_refreshes[buf] then
+      local params = {
```
MariaSolOs, line 305: Isn't this change creating `params` for non-active instead?

A motivational example (e.g. an issue where a server is doing something different and causing the described problem) would help.

cstoku: Suppose that `hoge.lua` is open with `bufnr=1` and `fuga.vim` with `bufnr=2`, and the current buffer is `bufnr=1`.
If we call `vim.lsp.codelens.refresh()`, it tries to refresh all loaded buffers.

In the code before fix, `vim.lsp.util.make_text_document_params()` does not specify `bufnr`, so it creates `params` for the current buffer `hoge.lua`.

Then, try to run `vim.lsp.buf_request` for each buffer, but `params` for `hoge.lua` is always passed and executed.

BTW, `active_refreshes` remembers the buffers that are being refreshed, to avoid duplicate refresh operations.

MariaSolOs: I see, I understand now.

MariaSolOs (commented):
Could you please add a test for this scenario?

cstoku:
I added the test.

test/functional/plugin/lsp_spec.lua
```diff
@@ -3879,6 +3879,73 @@ describe('LSP', function()
         end,
       }
     end)
+
+    it('refresh multiple buffers', function()
```
mfussenegger, line 3883: I think the test case could be simplified a bit by using the `_create_server` functionality instead of the `test_rpc_server`. See `Fallback to command execution on resolve error` for an example of how it is used.

Would be great if you could take a look at changing that. But if not I can also merge this as is, as it is already a clear improvement.

cstoku: Thanks for the review!
I'll try to fix it based on your review.

cstoku: @mfussenegger I fixed it.

PR #27710, opened 2024-03-02
fix(types): move IterMod type definition

### LINKED ISSUE #27711: Missing completion items when typing `vim.iter({})`
### Problem

When typing `vim.iter({}):`, completions for the module's functions are missing.

### Steps to reproduce

1. Write the following configuration to `repro.lua`:
```lua
local plugins = {
  lspconfig = 'https://github.com/neovim/nvim-lspconfig.git',
}

vim.env.PATH = vim.env.PATH .. ':' .. vim.fn.stdpath 'data' .. '/mason/bin'

for name, url in pairs(plugins) do
  local install_path = 'repro/' .. name
  end
  vim.o.runtimepath = install_path .. ',' .. vim.o.runtimepath
end

vim.bo.omnifunc = 'v:lua.vim.lsp.omnifunc'
require'lspconfig'.lua_ls.setup {}
```

2. Run `nvim --clean -u repro.lua repro.lua`.
3. Type `ggOvim.iter({}):<C-x><C-o>`.
4. Notice how no LSP completions (such as `any`, `map`, etc) are provided.

### Expected behavior

Completions for `vim.iter` to be provided.

### Neovim version (nvim -v)

v0.10.0-dev-dc8c086c7

No

### Operating system/version

Fedora Linux Asahi

### Terminal name/version

wezterm 20240226_174525_22424c32

wezterm

### Installation

From source

willothy:
Problem: `vim.iter`'s call operator is not being detected by lua_ls, so there's no completion or type hints for any `Iter` methods. 
Solution: Move type annotation to after the return statement so that it properly affects the value.

tldr makes `vim.iter()` actually return an `Iter`

fixes #27711

### DIFF
--- a/runtime/lua/vim/iter.lua
+++ b/runtime/lua/vim/iter.lua
@@ -1140,9 +1140,8 @@ function M.map(f, src, ...)
   return Iter.new(src, ...):map(f):totable()
 end
 
----@type IterMod
 return setmetatable(M, {
   __call = function(_, ...)
     return Iter.new(...)
   end,
-})
+}) --[[@as IterMod]]


willothy:
Hmm this seems to have messed up docs a bit, I will look into that.

MariaSolOs:
I guess this fixes #27711 ;) ?

willothy:
> I guess this fixes #27711 ;) ?

Ah, yes it does! I looked for an open issue but must've missed that one :)

MariaSolOs:
> > I guess this fixes #27711 ;) ?
> 
> 
> 
> Ah, yes it does! I looked for an open issue but must've missed that one :)
> 
> 
> 
> EDIT: lol I did not look at the number, no wonder I didn't see it

Lol yeah you're amazing but fixing future bugs might be a bit too much

willothy:
Haha thanks - it would definitely make things easier though!

Also, not sure if using `@as` is considered bad practice or not but it's the only change I tried that didn't mess up the doc generation. Various combinations of `@as` / my initial implementation of removing the blank line after the `IterMod` type def caused the module-level convenience functions to show up as `M:func` instead of `vim.iter.func`.

MariaSolOs:
> Also, not sure if using `@as` is considered bad practice or not but it's the only change I tried that didn't mess up the doc generation. 

I don't think it's considered bad practice per se. From my experience it's just usually a sign of missing previous type narrowing (e.g. an if-statement like `if foo ~= nil then ...`). Also there are usually better ways of pleasing the language server, like using `assert`. This does feel like a legit case to use it though.

> Various combinations of `@type` / my initial implementation of removing the blank line after the `IterMod` type def caused the module-level convenience functions to show up as `M:func` instead of `vim.iter.func`.

That's odd. @lewis6991 might know what's going on there.

willothy:
That's good to know, thanks!

I'm not familiar with the docgen implementation at all, but I'm curious so I'll look into it. It definitely seems like my initial implementation *should* work (it does fix the types, just messes up docs), so maybe this problem could be solved there instead to avoid running into the same issue again.

That could also be a followup though since this is a tiny change and does fix the types.

willothy:
Alright, I see why the docs weren't generated like I'd expected but not sure if it would actually be considered a problem. Tables with `---@class` annotations are treated differently than tables with `---@type` annotations, so the functions are added as methods instead of normal fields. 

Changing that might break docgen for other files so maybe it's best to just use `@as` in this case. Another option is using `---@type IterMod` above `local M = {}`, but that leads to 'inject field' warnings so not sure if that's any better.

MariaSolOs:
@willothy sounds like this might have been caused by #27644. Have you tried reverting that commit and checking if it still repros?

I'm also not sure if this is the intended design though.

willothy:
@MariaSolOs ah that's a good point! I'll try that out now :)

willothy:
I've reverted that commit and tried using the commit right before it, both still are lacking `vim.iter` types so it seems this was caused sometime earlier. Haven't figured out when the issue started yet.

PR #27712, opened 2024-03-02
fix(column): full redraw with 'stc, 'rnu' and inserted lines

### LINKED ISSUE #27709: When using a custom statuscolumn, an increase in the number width doesn't redraw lines properly
### Problem

When `numberwidth` is set to `3`, and a buffer is changed such that the width needs to be increased to `4`, _and_ a custom `statuscolumn` is set, the width change doesn't properly redraw all lines. The result is that the number/status column overlays one column of the buffer, until you somehow force a redraw or move the cursor around:

[Screencast from 2024-03-02 22-22-50.webm](https://github.com/neovim/neovim/assets/86065/21321001-8512-4f07-affe-71b6fb23ec05)

### Steps to reproduce

Create `repro.lua` with these contents:

```lua
vim.opt.numberwidth = 3
vim.opt.statuscolumn = '%s%=%{&relativenumber ? (v:virtnum == 0 ? (v:relnum > 0 ? v:relnum : v:lnum) : " ") : ""} '
vim.opt.relativenumber = true
vim.opt.number = true

local lines = {}

for i = 1, 99 do
	table.insert(lines, "line " .. i)
end

vim.api.nvim_buf_set_lines(0, 0, #lines - 1, false, lines)
```

Now run `nvim --clean -u repro.lua`, then:

1. Press `G` to go to the bottom of the buffer
2. Press `o` to create a new empty line
3. Observe how in the lines above the text changes from "line NN" to "ine NN"

When doing so, the following happens:

[asciicast](https://asciinema.org/a/nfBRPU3ZajtcNx5iMLuWBU10w)

`:redraw` doesn't help, but pressing `Control+L` does (which is what triggers the redraw at the end of the recording).

### Expected behavior

Lines are redrawn properly such that all buffer text remains visible.

### Neovim version (nvim -v)

b12d193b4a

No

### Operating system/version

Arch Linux

### Terminal name/version

Ghostty

xterm-ghostty

### Installation

AUR

luukvbaal:
Problem:  Text is not redrawn with 'relativenumber' when only the 'statuscolumn' is redrawn after inserted lines.
Solution: Force a full redraw if statuscolumn width changed.

Fix #27709

### DIFF
--- a/test/functional/ui/statuscolumn_spec.lua
+++ b/test/functional/ui/statuscolumn_spec.lua
@@ -927,4 +927,17 @@ describe('statuscolumn', function()
                                        |
     ]])
   end)
+
+  it('line increase properly redraws buffer text with relativenumber #27709', function()
+    screen:try_resize(33, 4)
+    command([[set rnu nuw=3 stc=%l\ ]])
+    command('call setline(1, range(1, 99))')
+    feed('Gyyp')
+    screen:expect([[
+      98  98                           |
+      99  99                           |
+      100 ^99                           |
+                                       |
+    ]])
+  end)
 end)
--- a/src/nvim/drawline.c
+++ b/src/nvim/drawline.c
@@ -581,7 +581,7 @@ static void draw_lnum_col(win_T *wp, winlinevars_T *wlv, int sign_num_attr, int 
 }
 
 /// Build and draw the 'statuscolumn' string for line "lnum" in window "wp".
-static void draw_statuscol(win_T *wp, winlinevars_T *wlv, linenr_T lnum, int virtnum,
+static void draw_statuscol(win_T *wp, winlinevars_T *wlv, linenr_T lnum, int virtnum, int col_rows,
                            statuscol_T *stcp)
 {
   // When called for the first non-filler row of line "lnum" set num v:vars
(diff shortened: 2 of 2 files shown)


test/functional/ui/statuscolumn_spec.lua
```diff
@@ -927,4 +927,17 @@ describe('statuscolumn', function()
                                        |
     ]])
   end)
+
+  it('line increase properly redraws buffer text with relativenumber', function()
+    screen:try_resize(33, 4)
+    command('set rnu nuw=3 stc=%l\\ ')
```
zeertzjq, line 933:
```suggestion
    command([[set rnu nuw=3 stc=%l\ ]])
```

test/functional/ui/statuscolumn_spec.lua
```diff
@@ -927,4 +927,17 @@ describe('statuscolumn', function()
                                        |
     ]])
   end)
+
+  it('line increase properly redraws buffer text with relativenumber', function()
```
gpanders, line 931:
```suggestion
  it('line increase properly redraws buffer text with relativenumber #27709', function()
```

zeertzjq (approved):

PR #27714, opened 2024-03-03
fix(treesitter): document more standard highlight groups

clason:
Problem: Not all standard treesitter groups are documented.

Solution: Document them all (without relying on fallback); add default link for new `*.builtin` groups to `Special` and `@keyword.type` to `Structure`. Remove `@markup.environment.*` which only made sense for LaTeX.

More `*.builtin` groups may follow in the future if they're added (consistently) in nvim-treesitter.

Also update all parsers to the latest version and queries from downstream.

Cf. https://github.com/nvim-treesitter/nvim-treesitter/pull/6232

@echasnovski 

TODO:
* [x] update `colors/vim.lua`
* [x] update queries


src/nvim/highlight_group.c
```diff
@@ -248,16 +249,18 @@ static const char *highlight_init_both[] = {
...
 
   "default link @function         Function",
   "default link @function.builtin Special",
 
   "default link @constructor Special",
   "default link @operator    Operator",
 
-  "default link @keyword Keyword",
+  "default link @keyword      Keyword",
+  "default link @keyword.type Structure",
```
echasnovski, line 263: Any particular reason to explicitly link `@keyword.type`? At least in default color scheme that would lead for them to not be bold and be plain `Nvim*Grey2`.

If there is no strong reason to link it, I'd leave `@keyword.type` fall back to `@keyword` as all other keywords.

clason: The links were meant mostly as illustration (and best-effort extrapolation). I defer to you for actual links; just make a suggestion and I'll commit it.

echasnovski: I'd remove explicit `@keyword.type` link and let it fall back to `@keyword`.

As other new groups have `builtin`, they should link to `Special` (as they already do, so nothing to do there).

clason:
@gpanders take note of the vimdoc queries, which use a different workaround to the `iter_matches`palooza.

runtime/queries/markdown_inline/highlights.scm
```diff
@@ -16,10 +16,10 @@
...
 
 ; Conceal codeblock and text style markers
-((code_span_delimiter) @markup.raw.delimiter
-  (#set! conceal ""))
-
-((emphasis_delimiter) @conceal
+([
+  (code_span_delimiter)
+  (emphasis_delimiter)
+] @conceal
```
gpanders, line 22: `@conceal` isn’t listed as a standard capture (unless I missed it). Is there some kind of exception in this case?

clason: Yes; it's a "special capture" like `@spell` (or, rather, like `@_foo`). See `:h treesitter-highlight-conceal`.

PR #27715, opened 2024-03-03
build: don't allow Xcode as generator

dundargoc:
Xcode does not allow having multiple targets depend on a custom command.
This limitation severely hinders its usability and complying with it
would likely require extensive refactoring and boilerplate. It makes
more sense to simply refer users to use "Ninja" or "Unix Makefiles"
instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,10 @@ if(POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+if(XCODE)
+  message(FATAL_ERROR [[Xcode generator is not supported. Use "Ninja" or "Unix Makefiles" instead.]])
+endif()
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ if(POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+if(XCODE)
+  message(FATAL_ERROR [[Xcode is not an allowed build generator at the moment. Use "Ninja" or "Unix Makefiles" instead.]])
```
clason, line 21:
```suggestion
  message(FATAL_ERROR [[Xcode generator is not supported. Use "Ninja" or "Unix Makefiles" instead.]])
```

clason (approved):

PR #27716, opened 2024-03-03
ci(windows): optionally test windows with ASAN

dundargoc:
Add the label `ci:windows-asan` to enable it.


dundargoc:
TODO: write real commit message + document in maintain

dundargoc:
This PR is "done" in that it does what is says on the tin. However, there are a bunch of tests that still fail which I suspect is due to the lower default stack size in windows. Solving that looks tough so I will leave that as an exercise to someone more motivated.

bfredl (approved):

PR #27718, opened 2024-03-03
Fix error when using custom treesitter directive

ErhanKurnaz:
I got the following error when I used my custom treesitter directive:

```
Error in decoration provider treesitter/highlighter.line:
Error executing lua: /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:741: attempt to get length of local 'v' (a boolean value)
stack traceback:
        /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:741: in function 'handler'
        /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:823: in function 'apply_directives'
        /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:898: in function 'iter'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:266: in function 'fn'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:209: in function 'for_each_highlight_state'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:251: in function 'on_line_impl'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:320: in function <...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:314>
```

through some printing I found out that only the keys that were numbers were actual tables. So only keys that are numbers can be used to take the length of. I also saw that we did a "if key is number" check in de `add_predicate` function so I just copied that. This seemed to fix the issue

### DIFF
--- a/runtime/lua/vim/treesitter/query.lua
+++ b/runtime/lua/vim/treesitter/query.lua
@@ -742,7 +742,9 @@ function M.add_directive(name, handler, opts)
     local function wrapper(match, ...)
       local m = {} ---@type table<integer, TSNode>
       for k, v in pairs(match) do
-        m[k] = v[#v]
+        if type(k) == 'number' then
+          m[k] = v[#v]
+        end
       end
       handler(m, ...)
     end


ErhanKurnaz:
For some extra context this was the custom directive code:

```lua
```

ErhanKurnaz:
And this is how I used it in a highlights query to add some highlighting in a Go format string:

```scheme
(
 (call_expression
   function: (selector_expression
               field: (field_identifier (#lua-match? "^.*f$")))
   arguments: (argument_list
                (interpreted_string_literal) @string.special
                (#lua-match? @string.special "%%#?%+?%.?%d* ?%w")
                (#offset-to-match! @string.special "%%#?%+?%.?%d* ?%w")))
)
```

clason:
We need more context: Please give a minimal and complete example we can test with `nvim --clean -u test.lua` on the latest(!) commit of `master`.

ErhanKurnaz:
[A gist to the minimal configuration](https://gist.github.com/ErhanKurnaz/74f5cc6cf5efe4aada6b910fba46cce1)

- put the highlights.scm under queries/go/highlights.scm
- open `test.go`
- `:TSInstall go`
- `:= vim.treesitter.start()`

and now you should see the error

clason:
And what version of Neovim are you testing on?

ErhanKurnaz:
NVIM v0.10.0-dev-2504+g3971797be
Build type: Release
LuaJIT 2.1.1707061634

lucario387:
This line just is wrong though

```
               field: (field_identifier (#lua-match? "^.*f$")))
```

ErhanKurnaz:
If I remove that line I still get the same error

clason:
This is almost certainly not the correct fix. The first step is to check the actual contents of the full `match` table in your situation and understand what is going wrong and why.

The second step (in parallel) is to use `all=true` in `add_directive` (which is the correct behavior and should be used in any new code since it will become the default).

ErhanKurnaz:
Ah I see. Yeah it works when I say that `all = true`. Then this pull request can be closed. Thanks for the help

PR #27727, opened 2024-03-04
docs(luaref): fix typo

WantenMN:
2.5.9 Function Definitions

lua-colonsyntax


clason:
I don't see any difference there?

WantenMN:
> I don't see any difference there?

There is an extra '(' symbol there.

PR #27728, opened 2024-03-04
fix(tohtml): replace hex escape with digit escape

altermo:
Lua 5.1 doesn't support hex escape.

The test didn't fail because it also used hex escape.


zeertzjq:
Hmm, it should work if you use these Unicode characters directly instead of specifying them as UTF-8 bytes.

zeertzjq (approved):

PR #27738, opened 2024-03-05
vim-patch:9.1.{0147,0149,0150,0152,0156},58f1e5c0893a: 'winfixbuf'

### LINKED ISSUE #12517: Sticky Buffers
### Environment
- `nvim --version`:
```
NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.4.3/src -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```
- `vim -u DEFAULTS` (version: ) behaves differently?: no
- Operating system/version: Arch linux
- Terminal name/version: alacritty 0.4.3
- `$TERM`: alacritty

### Actual behaviour

This is a suggestion to solve a general problem plugin developers encounter.

Consider the following scenario:

You have a file manager plugin installed (like NERDTree) and you wish to keep it's buffer through your nvim session.
The main problem is that if you try to open another file (For example with :FZF) you need to check if your focus is not on the file manager buffer or else this would happen:

(Taken from https://github.com/junegunn/fzf/issues/453)

The workaround is to use hacky vimscript to try detect the buffer name and then shift the focus:
```vim
nnoremap <silent> <expr> <Leader><Leader> (expand('%') =~ 'NERD_tree' ? "\<c-w>\<c-w>" : '').":FZF\<cr>"
```
But you'll have to use the prefix `(expand('%') =~ 'NERD_tree' ? "\<c-w>\<c-w>" : '')."` for each possible command you would like, which is tedious.
What if you would like to switch from NERDTree to other file manager? You would have to change everything to match the new buffer name.
This use case is also relevant when using a terminal buffer in normal mode, I don't want to accidently overwrite it with something else.

Plugin developers cannot make their plugin to act nicely with other plugins as fzf developer describes the problem perfectly in https://github.com/junegunn/fzf.vim/issues/326#issuecomment-282936932
### Reproduce with nvim -u NORC
```
nvim -u NORC
install fzf.vim and NERDTree
:NERDTREE (opens nerdtree)
Focus nerdtree buffer
:FZF and choose any file
nerdtree is gone

```
### Expected behaviour

Opening a file when the focus is on a buffer that I would like to keep, the file will be open in a different buffer or a vertical split if there are no more buffers.
Maybe marking buffers with sticky mode which can only be close only with :quit ?

```

1. nvim -u NORC
2. install fzf.vim and NERDTree
3. :NERDTREE (opens nerdtree)
4. Focus nerdtree buffer
5. Mark buffer as sticky
6. :FZF and choose any file
7. The file is opened in a split 
8. Focus nerdtree buffer
9. :FZF and choose any file
10. The file is opened in the same split as in step 7


```

Is there a better solution for this problem other than detecting buffer names?

I'm aware there are mapping to open files in vertical or horizontal split. But the main issue is that I constantly need to know where my focus is and be careful not to overwrite a buffer that I want to keep.

Thank you for your time.

ColinKennedy:
Close #12517

#### vim-patch:9.1.0147: Cannot keep a buffer focused in a window

Problem:  Cannot keep a buffer focused in a window
          (Amit Levy)
Solution: Add the 'winfixbuf' window-local option
          (Colin Kennedy)

closes: vim/vim#13903

https://github.com/vim/vim/commit/215703563757a4464907ead6fb9edaeb7f430bea

Co-authored-by: Colin Kennedy <colinvfx@gmail.com>

#### vim-patch:9.1.0149: null pointer member access when accessing 'winfixbuf' property

Problem:  qf_goto_win_with_qfl_file may check if prevwin has 'winfixbuf'
          set without checking if it's valid first.
Solution: Reverse the condition. Add a test, a modeline, and a missing
          CheckFeature. (Searn Dewar)

closes: vim/vim#14140

https://github.com/vim/vim/commit/5131f224da93f2e042a4b22545ef62b1b2ab8460

Co-authored-by: Sean Dewar <6256228+seandewar@users.noreply.github.com>

#### vim-patch:58f1e5c0893a

runtime(doc): Highlight the error message at *E1513*

https://github.com/vim/vim/commit/58f1e5c0893aed47d762f5698a1ebebc05ec9819

#### vim-patch:9.1.0150: Several minor 'winfixbuf' issues

Problem:  several minor 'winfixbuf' issues exist, mostly relating to the
	  quickfix list
Solution: address them and adjust tests. Retab and reflow a few things too.
	  (Sean Dewar)

Things touched include:

- Replace the semsgs with gettext'd emsgs.

- Handle window switching in ex_listdo properly, so curbuf and curwin
  are kept in-sync and trigger autocommands; handle those properly.

- Don't change the list entry index in qf_jump_edit_buffer if we fail
  due to 'wfb' (achieved by returning FAIL; QF_ABORT should only be used
  if the list was changed).

- Make qf_jump_edit_buffer actually switch to prevwin when using `:cXX`
  commands **outside** of the list window if 'wfb' is set in curwin.
  Handle autocommands properly in case they mess with the list.

  NOTE: previously, it seemed to split if 'wfb' was set, but do nothing
  and fail if prevwin is *valid*. This behaviour seemed strange, and maybe
  unintentional? Now it aligns more with what's described for the `:cXX`
  commands in the original PR description when used outside a list window,
  I think.

- In both functions, only consider prevwin if 'wfb' isn't set for it;
  fallback to splitting otherwise.

- Use win_split to split. Not sure if there was a specific reason for
  using ex_splitview. win_split is simpler and respects modifiers like
  :vertical that may have been used. Plus, its return value can be checked
  for setting opened_window in qf code (technically win_split_ins autocmds
  could immediately close it or change windows, in which the qf code might
  close some other window on failure; it's already the case elsewhere,
  though).

closes: vim/vim#14142

https://github.com/vim/vim/commit/4bb505e28cac0389561fff78d8bbe0319c2bcf2f

Co-authored-by: Sean Dewar <6256228+seandewar@users.noreply.github.com>


zeertzjq:
Hmm, `has('pythonx')` is always true, which seems like a bug. I'll open a PR.

ColinKennedy:
> Hmm, has('pythonx') is always true, which seems like a bug. I'll open a PR.

Yeah I was wondering about that one but figured it was just an intended difference between Neovim and Vim. Glad you caught it :)

ColinKennedy:
I've updated the PR, please let me know if there's any other changes needed and I can add them. Thank you!

ColinKennedy:
I recently found out about https://github.com/vim/vim/pull/14142 and https://github.com/vim/vim/commit/4bb505e28cac0389561fff78d8bbe0319c2bcf2f. I'll add this patch into this PR and update the title/description

.github/workflows/docs.yml
```diff
@@ -1,5 +1,8 @@
 name: docs
 on:
+  push:
+    branches:
+      - 'vim-9.1.0147-002'
```
clason, line 5: 👀 

(Convention is that such changes go in a separate `TEST COMMIT DELETE LATER` commit, not squashed in with the actual changes.)

ColinKennedy: Crap I forgot to remove that, sorry. I'll try to remember in the future or at least keep that commit separate, as you suggested

test/functional/options/winfixbuf_spec.lua
```diff
@@ -0,0 +1,54 @@
+local helpers = require('test.functional.helpers')(after_each)
+local clear = helpers.clear
+local exec_lua = helpers.exec_lua
+
+describe("Neovim API calls with 'winfixbuf'", function()
```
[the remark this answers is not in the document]
ColinKennedy: It's pretty late where I live so I will leave that reading for another day but for now I'll commit your suggestion!

.github/workflows/docs.yml
clason: No worries; that _is_ something that people do if they need to verify something specific on CI.

ColinKennedy:
Okay, this PR has all patches for `'winfixbuf'` that I could see using `vim-patch.sh`. It should be reviewable tests pass (at the time of writing, a couple are still running). Please let me know of any edits that I should make!

zeertzjq (approved):

PR #27756, opened 2024-03-06
fix(fs): use generics for better typing

altermo:
**Problem:**
When using `vim.fs.dirname` (or `vim.fs.basename`) the typing says that it may return `nil` even if `string` was passed in, which causes some unnecessary warnings.
For example, this is a type warning: `vim.notify(vim.fs.dirname('/home/user'))`

**Sulution:**
Use generics.

**Other changes:**
The docs now contain the information that `vim.fs.dirname` and `vim.fs.basename` may take `nil`.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2848,7 +2848,7 @@ vim.fs.basename({file})                                    *vim.fs.basename()*
     Return the basename of the given path
 
     Parameters: ~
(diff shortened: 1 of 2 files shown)


wookayin:
How about using `@overload fun(file: nil): nil` instead?

https://luals.github.io/wiki/annotations/#overload

PR #27760, opened 2024-03-07
fix(process): close handles and timer in pty_process_close()

zeertzjq:
This should prevent use-after-free on exit on Windows.


dundargoc:
Confirmed this fixes the ASAN problem

dundargoc (approved):

PR #27774, opened 2024-03-08
refactor: use libc stpcpy and stpncpy

IAKOBVS:
We check for strnlen which is POSIX 2008, and POSIX 2008 includes stpcpy and stpncpy.


dundargoc:
What's the benefit tho? xstpcpy should always work anyway

IAKOBVS:
> What's the benefit tho? xstpcpy should always work anyway

So that they do not scan the whole source before copying it to the destination, resulting in 2 * strlen(src) operations or being up to twice as slow as strcpy.

IAKOBVS:
The asan clang oldtest failure seems to be irrelevant as it's also present in other merged commits.

src/nvim/memory.c
```diff
@@ -333,43 +333,16 @@ size_t memcnt(const void *data, char c, size_t len)
...
-/// WARNING: If copying takes place between objects that overlap, the behavior
-/// is undefined.
-///
-/// Nvim version of POSIX 2008 stpcpy(3). We do not require POSIX 2008, so
-/// implement our own version.
-///
-/// @param dst
-/// @param src
+#ifndef HAVE_STPCPY
 char *xstpcpy(char *restrict dst, const char *restrict src)
```
justinmk, line 337: why delete the entire doc? at least a note like this is helpful:

    /// Nvim version of POSIX 2008 stpcpy(3).

justinmk:
> So that they do not scan the whole source before copying it to the destination, resulting in 2 * strlen(src) operations or being up to twice as slow as strcpy.

This is useful to mention in the PR description / commit message. Please always say what problem is being solved, it greatly helps reviewers and future devs.

Also very helpful to rebase your PRs every now and then.

src/nvim/memory.c
IAKOBVS: I was following xstrnlen which has no comments, presumably because it follows POSIX strnlen.

justinmk: xstrnlen could have a similiar docstring. A 1-line comment providing a reference is low-cost way to give hints.

asmodeus812:
> > What's the benefit tho? xstpcpy should always work anyway
> 
> So that they do not scan the whole source before copying it to the destination, resulting in 2 * strlen(src) operations or being up to twice as slow as strcpy.

Doubt on the 2*, since it is sequential string the hardware prefetcher will do its job, it might be also possible that the string is still in cache from the first count when doing the second passthrough count. There is likely some performance gain to be had here but certainly not 2 *.

justinmk (approved):

PR #27777, opened 2024-03-08
feat!: remove deprecated functions

dundargoc:


clason:
Mark as `feat!`, just to be safe?

lewis6991 (approved):

MariaSolOs (approved):

wookayin:
Would we also need to remove corresponding entries from doc/deprecated.txt?

dundargoc:
> Would we also need to remove corresponding entries from doc/deprecated.txt?

Pretty sure that's not a thing. I don't remember ever removing something from `deprecated.txt`. I mean, we could start doing it I guess, but I think it'd make more sense to do it in a separate, dedicated PR.

wookayin:
I'd suggest that we consider doing that as a routine release workflow --- it'd be helpful for [helpful.vim](https://github.com/tweekmonster/helpful.vim). I'm fine with either removing together here (for the atomicity) or as a separate PR.

PR #27789, opened 2024-03-09
fix(lsp): default link more LSP semantic token types

cajomar:
`@lsp.type.keyword` and `@lsp.type.string` weren't included in the list of default highlight links.
This isn't a problem for languages since they'll already have vim syntax files or treesitter support, but some DSLs will only lsp semantic highlighting implemented.
If these token types were intentionally omitted it's easy enough to add these links in one's config, but it seems reasonable to me to link them by default.


MariaSolOs:
@cajomar You will want to squash the `update doc` for the lintcommit check.

Also adding @clason and @echasnovski as reviewers as subject experts.

clason:
If it's in the spec, we should include it. Regarding the default colorscheme which is intentionally minimal, I defer to @echasnovski

echasnovski:
> If it's in the spec, we should include it.

If that is the case, then there are quite a few missing from current `@lsp.type` highlights. Based on [this part of LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens), there should also be (compared to the current `master`): `event`, `keyword`, `number`, `operator`, `regexp`, `string`. Should all of them be present (at least in documentation)?

echasnovski:
In this case, @cajomar, would you mind also adding extra 5 semantic tokens types to documentation and default color scheme? Please, use the following default highlight groups (while preserving alphabetical order and aligning):

```
@lsp.type.event    @type
@lsp.type.modifier @type.qualifier
@lsp.type.number   @number
@lsp.type.operator @operator
@lsp.type.regexp   @string.regexp
```

clason:
Note to self: `@type.qualifier 
` is being renamed to `modifier` in my open PR.

cajomar:
@echasnovski By adding it to the default color scheme do you mean just modifying `highlight_group.c` or is there a separate color file I need to edit?

echasnovski (approved):
Thanks!
Looks good to me.

> @echasnovski By adding it to the default color scheme do you mean just modifying `highlight_group.c` or is there a separate color file I need to edit?

No, no. 'highlight_group.c' is just right.

mfussenegger (commented):
The test failures look related?

Otherwise lgtm

echasnovski:
> The test failures look related?

Yeah, they are.

@cajomar, if you'd update test, that would be great. If you don't feel like learning how to update Neovim screenshot tests, you can revert to the previous state of your PR (with only `keyword` and `string` added) and I'll then add the rest in the follow up PR.

cajomar:
@echasnovski I updated the semantic tokens screenshot test.
The tokens in the expected screenshot are two columns off from the actual text input, I presume because rust-analyzer is auto-formatting to 4-space indentation. If that's unintended I can update the rust input to 4-space indentation and add string/macro highlighting tests as well.

Two other tests are failing, but they appear to fail on `master` as well and seem unrelated to my changes.

clason:
Yes, would be nice to change the tests to not be weird.

PR #27804, opened 2024-03-10
docs: mention `nvim_buf_set_name()` in `BufFilePre`/`BufFilePost`

echasnovski:
Problem: right now it is not entirely obvious that `nvim_buf_set_name()` also triggers events for renaming buffer.

Solution: mention it in help file.

### DIFF
--- a/runtime/doc/autocmd.txt
+++ b/runtime/doc/autocmd.txt
@@ -238,9 +238,11 @@ BufEnter			After entering (visiting, switching-to) a new
 							*BufFilePost*
 BufFilePost			After changing the name of the current buffer
 				with the ":file" or ":saveas" command.
+				Also during |nvim_buf_set_name()|.
 							*BufFilePre*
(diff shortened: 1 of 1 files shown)


echasnovski:
> E.g. it always creates an alternate buffer. I'm not sure it's wise to document every single side effect when most of them are probably unintended.

Those events are triggered during "...changing the name of the current buffer". `nvim_buf_set_name()` changes buffer name, so I'd assume that this is intended.

Originally I was searching for the way to act when buffer name is changed. But as from Lua this is mostly done with `nvim_buf_set_name()`, I did not find this events immediately. Hence adding these notes to the help.

echasnovski:
After digging deeper into help (not without a help of `:helpgrep`) I found out that `BufNew` is also for "After ... renaming an existing buffer". So although `nvim_buf_set_name()` does indeed trigger `BufFilePre` and `BufFilePost`, those are not the proper event to watch for rename of an existing buffer.

Issue #27811, opened 2024-03-11
ui2 (extui): ext_cmdline/message for the TUI: tracking issue


luukvbaal:
I will be working on an ext_cmdline/message implementation for the TUI. An internal C implementation was explored first, but dropped after coming to the conclusion that properly untangling the current cmdline from the message grid in C would require more work than leveraging and improving the already existing parallel ext_ code path.

**Goals/progress:**
- [x] Introduce an experimental ext_cmdline/messages UI: #27855.
- [ ] Remove redundant C components:
	- [ ] Unmaintained Vimscript parser and cmdline highlighter.
	- [ ] Message grid.

luukvbaal:
Okay, I have basically finished the necessary plumbing for an initial ext_cmdline/message implementation(#27855). Now I find myself wondering if it will really be worth the effort to try to spoof the current message grid implementation if we intend to move away from it later. My initial implementation is close but actually making all screen tests pass will require more work. Maybe we @neovim/core @neovim/padawans should try to settle on what the default cmdline/message should look like first.

I think there is merit in keeping the cmdline in the current 'cmdheight' area, as anything else might be too "unfamiliar"? If we don't care about that and can settle on a better alternative, I'm all for it. E.g. a floating window without borders on the last row (potentially on top of the statusline), or a floating window with borders positioned elsewhere.

I suppose the most obvious candidate for an alternative message window would be a floating window with or without borders in the bottom right corner.

Welcoming discussion.

przepompownia:
After merge #31167  I started to try what is currently possible without FFI hacks and what issues can wait for the user and store the result in https://github.com/przepompownia/mini-ui-messages-example. It has two init files with shared redirection logic: the first uses modified mini.notify, the second displays messages in a custom way. 

`n` in normal mode causes sending two `msg_show` events: the first one with empty `kind`, the second with  `search_count` (unless we have added `S` to `shortmess`). Probably it would be good to assign some special kind to the first (I don't know if it requires anything in Vim too).
 
On the other side, commands like `:=1`, `:={x=1, y=1}` results with `msg_show` events and empty `kind`. I guess it's impossible to distinguish such messages from the search output. I also didn't find any Vim option to silent search events completely.

przepompownia:
Let me link #22914 as possibly related - otherwise please mark it as offtopic.

justinmk:
The subissues is overflowing, shall we create a new parent issue?

luukvbaal:
[...]

PR #27813, opened 2024-03-11
feat(ext_messages): add hl_id to ext_messages chunks

### LINKED ISSUE #20310: Make it easier to retrieve highlight attributes during `vim.ui_attach` with `ext_messages`
### Feature already in Vim?

_No response_

### Feature description

When using `ext_messages`, I found no easy way to retrieve highlight attributes. `ext_hlstate` looks promising, but doesn't seem to be available within lua.

Would be great if there would be an API call to retrieve the HL attributes.

Current workaround I found by @notomo:

```lua
local M = {}

function M.setup()
  local ffi = require("ffi")
  local ok, err = pcall(
    ffi.cdef,
    [[typedef int32_t RgbValue;
      typedef struct attr_entry {
        int16_t rgb_ae_attr, cterm_ae_attr;
        RgbValue rgb_fg_color, rgb_bg_color, rgb_sp_color;
        int cterm_fg_color, cterm_bg_color;
        int hl_blend;
      } HlAttrs;
      HlAttrs syn_attr2entry(int attr);]]
  )
  if not ok and not err:find("redefine") then
    error(err)
  end
  M.attr2entry = ffi.C.syn_attr2entry
end

function M.attr2entry(attr_id)
  M.setup()
  return M.attr2entry(attr_id)
end

M.cache = {}

function M.get_hl_group(attr_id)
  return "NoiceAttr" .. tostring(attr_id)
end

function M.get_hl(attr_id)
  if not M.cache[attr_id] then
    local attrs = M.attr2entry(attr_id)
    local hl_group = M.get_hl_group(attr_id)
    vim.api.nvim_set_hl(0, hl_group, {
      fg = attrs.rgb_fg_color,
      bg = attrs.rgb_bg_color,
      sp = attrs.rgb_sp_color,
      bold = bit.band(attrs.rgb_ae_attr, 0x02),
      standout = bit.band(attrs.rgb_ae_attr, 0x0100),
      italic = bit.band(attrs.rgb_ae_attr, 0x04),
      underline = bit.band(attrs.rgb_ae_attr, 0x08),
      undercurl = bit.band(attrs.rgb_ae_attr, 0x10),
      nocombine = bit.band(attrs.rgb_ae_attr, 0x0200),
      reverse = bit.band(attrs.rgb_ae_attr, 0x01),
      blend = attrs.hl_blend ~= -1 and attrs.hl_blend or nil,
    })
    M.cache[attr_id] = hl_group
  end
  return M.cache[attr_id]
end

return M
```

luukvbaal:
**refactor(message): propagate highlight id instead of attrs**

Problem:  Highlight group id is not propagated to the end of the message call
          stack, where ext_messages are emitted.
Solution: Refactor message functions to pass along highlight group id
          instead of attr id.

**feat(ext_messages): add hl_id to ext_messages chunks**

Problem:  Ext_messages chunks only contain the highlight attr id, which
          is not very useful for vim.ui_attach().
Solotion: Add highlight group id to message chunks, which can easily be
          used to highlight text in the TUI through nvim_buf_set_extmark():
          hl_group = hl_id.

Resolve #20310

### DIFF
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -805,10 +805,11 @@ must handle.
 	    kinds as the empty kind.
 
 	content
-	    Array of `[attr_id, text_chunk]` tuples, building up the message
-	    text of chunks of different highlights. No extra spacing should be
-	    added between chunks, the `text_chunk` by itself contains any
-	    necessary whitespace. Messages can contain line breaks "\n".
+	    Array of `[attr_id, text_chunk, hl_id]` tuples, building up the
+	    message text of chunks of different highlights. No extra spacing
+	    should be added between chunks, the `text_chunk` by itself
+	    contains any necessary whitespace. Messages can contain line
+	    breaks "\n".
 
 	replace_last
 	    Decides how multiple messages should be displayed:
--- a/src/nvim/autocmd.c
+++ b/src/nvim/autocmd.c
@@ -220,14 +220,14 @@ static void au_show_for_event(int group, event_T event, const char *pat)
         // show the group name, if it's not the default group
         if (ac->pat->group != AUGROUP_DEFAULT) {
           if (last_group_name == NULL) {
-            msg_puts_attr(get_deleted_augroup(), HL_ATTR(HLF_E));
+            msg_puts_hl(get_deleted_augroup(), HLF_E + 1, false);
           } else {
-            msg_puts_attr(last_group_name, HL_ATTR(HLF_T));
+            msg_puts_hl(last_group_name, HLF_T + 1, false);
           }
           msg_puts("  ");
         }
         // show the event name
-        msg_puts_attr(event_nr2name(event), HL_ATTR(HLF_T));
+        msg_puts_hl(event_nr2name(event), HLF_T + 1, false);
       }
 
       // Show pattern only if it changed.
@@ -240,7 +240,7 @@ static void au_show_for_event(int group, event_T event, const char *pat)
         }
 
         msg_col = 4;
-        msg_outtrans(ac->pat->pat, 0);
+        msg_outtrans(ac->pat->pat, 0, false);
       }
 
       if (got_int) {
@@ -260,17 +260,17 @@ static void au_show_for_event(int group, event_T event, const char *pat)
         size_t msglen = 100;
         char *msg = xmallocz(msglen);
         if (ac->exec.type == CALLABLE_CB) {
-          msg_puts_attr(exec_to_string, HL_ATTR(HLF_8));
+          msg_puts_hl(exec_to_string, HLF_8 + 1, false);
           snprintf(msg, msglen, " [%s]", ac->desc);
         } else {
           snprintf(msg, msglen, "%s [%s]", exec_to_string, ac->desc);
         }
-        msg_outtrans(msg, 0);
+        msg_outtrans(msg, 0, false);
         XFREE_CLEAR(msg);
       } else if (ac->exec.type == CALLABLE_CB) {
-        msg_puts_attr(exec_to_string, HL_ATTR(HLF_8));
+        msg_puts_hl(exec_to_string, HLF_8 + 1, false);
       } else {
-        msg_outtrans(exec_to_string, 0);
+        msg_outtrans(exec_to_string, 0, false);
       }
       XFREE_CLEAR(exec_to_string);
       if (p_verbose > 0) {
--- a/src/nvim/ex_getln_defs.h
+++ b/src/nvim/ex_getln_defs.h
@@ -10,8 +10,8 @@
 /// Defines a region which has the same highlighting.
 typedef struct {
   int start;  ///< Colored chunk start.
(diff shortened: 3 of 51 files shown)


runtime/doc/ui.txt
```diff
@@ -197,6 +197,7 @@ the editor.
 
 	The following keys are deprecated:
 
+	undeprecate?
```
luukvbaal, line 200: Originally deprecated here: https://github.com/neovim/neovim/commit/fa4c2601000df2d792b0de865da3ac8c43ab723f. With this change the original reasoning no longer holds so undeprecate?

src/nvim/autocmd.c
```diff
@@ -219,14 +219,14 @@ static void au_show_for_event(int group, event_T event, const char *pat)
         // show the group name, if it's not the default group
         if (ac->pat->group != AUGROUP_DEFAULT) {
           if (last_group_name == NULL) {
-            msg_puts_attr(get_deleted_augroup(), HL_ATTR(HLF_E));
+            msg_puts_hl(get_deleted_augroup(), HLF_E + 1, false);
```
luukvbaal, line 222: Adding `+ 1` at each call site is unfortunate. Do we want to adapt the enum and account for it elsewhere?

zeertzjq:
What is the point of this?

luukvbaal (commented):
Started as an idea to be able to store message highlights as extmarks. Dropped it because it didn't seem worth the churn. Finished it because @bfredl suggested it would be a welcome change regardless.

Possibly resolves #20310.

luukvbaal:
> What is the point of this

Ran into this while still working on an internal replacement for the message grid. `extmark_set()` expects a highlight group ID, not attributes, so the ID needed to be reserved. See review comment.

Besides that, the point is that the highlight id is more useful than it's attributes for ext_messages consumers(#20310).

justinmk:
> What is the point of this?

A concise problem/solution description in each PR is very helpful. Even though https://github.com/neovim/neovim/issues/27811 also provides useful context, it's very helpful for each PR to say what problem it is solving.

luukvbaal:
@folke could you confirm that this would resolve the need for the workaround you mentioned in #20310?

If so, should I rebase this @zeertzjq @justinmk @bfredl or do you not think it's worth the churn?

folke:
Yes, looks great!

bfredl:
Looks good at a quick glance. Please go ahead and rebase and I will do a final review.

src/nvim/autocmd.c
bfredl: I'll adjust the enum to start with 1 in a follow up refactor, this is good enough for now.

src/nvim/ex_getln_defs.h
```diff
@@ -49,7 +49,7 @@ struct cmdline_info {
   int cmdfirstc;                ///< ':', '/', '?', '=', '>' or NUL
   int cmdindent;                ///< number of spaces before cmdline
   char *cmdprompt;              ///< message in front of cmdline
-  int cmdattr;                  ///< attributes for prompt
+  int cmdhlid;                  ///< highlight id for prompt
```
bfredl, line 52: `cmd_hl_id` (or maybe `cmdhl_id`). it is better to be consistent with always using `hl_id` elsewhere, never `hlid`

src/nvim/globals.h
```diff
@@ -156,7 +156,7 @@ EXTERN bool msg_scrolled_ign INIT( = false);
 EXTERN bool msg_did_scroll INIT( = false);
 
 EXTERN char *keep_msg INIT( = NULL);         // msg to be shown after redraw
-EXTERN int keep_msg_attr INIT( = 0);         // highlight attr for keep_msg
+EXTERN int keep_msg_hlid INIT( = 0);         // highlight id for keep_msg
```
bfredl, line 159: `hl_id`

src/nvim/statusline.c
```diff
@@ -582,9 +582,10 @@ void win_redr_ruler(win_T *wp)
 
   if (ui_has(kUIMessages) && !part_of_status) {
     MAXSIZE_TEMP_ARRAY(content, 1);
-    MAXSIZE_TEMP_ARRAY(chunk, 2);
+    MAXSIZE_TEMP_ARRAY(chunk, 3);
     ADD_C(chunk, INTEGER_OBJ(attr));
     ADD_C(chunk, CSTR_AS_OBJ(buffer));
+    ADD_C(chunk, INTEGER_OBJ(0));
```
bfredl, line 588: Hmm, we probably should either calculate the actual hl_id, or just leave this change out completely (we can add it later if it is actually needed)

bfredl (approved):
otherwise LGTM

src/nvim/statusline.c
luukvbaal: Right, think I just forgot about this one as a TODO, thanks. In the current logic it is always equal to `HLF_MSG`, added an assert to catch if that ever changes and more logic is needed.

PR #27830, opened 2024-03-12
refactor: add vim_str[n]cpy_up and use them in vim_str[n]save_up

IAKOBVS:
Current uses of vim_strup() calls memcpy()/strcpy() before calling vim_strup(). This results in 2 * strlen(string) operations.

We can trivially convert to lowercase while copying the string instead.

### DIFF
--- a/src/nvim/strings.c
+++ b/src/nvim/strings.c
@@ -346,6 +346,18 @@ void vim_strncpy_up(char *restrict dst, const char *restrict src, size_t n)
   *dst = '\0';
 }
 
+// memcpy (does not NUL-terminate) plus vim_strup.
+void vim_memcpy_up(char *restrict dst, const char *restrict src, size_t n)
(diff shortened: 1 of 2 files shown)


src/nvim/strings.c
```diff
@@ -324,6 +324,40 @@ void vim_strup(char *p)
...
+// memcpy (does not NUL-terminate) plus vim_strup.
+void vim_memcpy_up(char *restrict dst, const char *restrict src, size_t n)
+  FUNC_ATTR_NONNULL_ALL
+{
+  uint8_t c;
+  while (n--) {
+    c = (uint8_t)(*src++);
+    *dst++ = (char)(uint8_t)(c < 'a' || c > 'z' ? c : c - 0x20);
+  }
+  *dst = '\0';
```
zeertzjq, line 358:
```suggestion
```

zeertzjq (approved):

PR #27833, opened 2024-03-12
fix(tutor): set a value for ":syntax iskeyword"

En-En-Code:
(Believe this is the correct place for this PR. Although a runtime file, Vim does not use the `.tutor` extension for its tutor files and thus does not have a `runtime/syntax/tutor.vim` file.)

1. The colon (:) no longer counted as a keyword character for syntax highlighting. As a result, "Note:" now correctly displays as "**Note**:", etc. I choose the same set of keyword characters as `runtime/ftplugin/tutor.vim` for consistency. Supposedly the `iskeyword` setting set by that file influences the syntax highlighting, but when building from master, I tripped the "Tutor applies {unix:…,win:…} transform" functional test because of this inconsistency. I don't mind the reduncacy, but I am also all for a better fix if one exists.
2. While skimming over the tutor files to confirm item 1, I noticed the *Learning Vim in 2014* link did not extend to the "2014/" at the end of the line. The change to the regex I made works identically on links not ending in a "/", while allowing matches with the more general `[[:graph:]]` over the more restrictive `\>` in the character(s) preceding a "/". This keeps the regex quite close to the original, including allowing punctuation following a URI to not be matched whether or not the link ends in "/", but allows non-keyword characters to be at the end of a link if closed off by a "/". An additional functional test has been added to the same file as the test failed in item 1 to confirm this functionality.

### DIFF
--- a/runtime/syntax/tutor.vim
+++ b/runtime/syntax/tutor.vim
@@ -8,6 +8,8 @@ syn include @TUTORSHELL syntax/sh.vim
 unlet b:current_syntax
 syn include @VIMNORMAL syntax/vimnormal.vim
 
+syn iskeyword @,-,_,48-57
+
 syn match tutorLink /\[.\{-}\](.\{-})/ contains=tutorInlineNormal
 syn match tutorLinkBands /\[\|\]\|(\|)/ contained containedin=tutorLink,tutorLinkAnchor conceal
 syn match tutorLinkAnchor /(.\{-})/ contained containedin=tutorLink conceal
--- a/test/functional/plugin/tutor_spec.lua
+++ b/test/functional/plugin/tutor_spec.lua
@@ -12,7 +12,7 @@ describe(':Tutor', function()
     clear({ args = { '--clean' } })
     command('set cmdheight=0')
     command('Tutor')
-    screen = Screen.new(80, 30)
+    screen = Screen.new(81, 30)
     screen:set_default_attr_ids({
       [0] = { foreground = Screen.colors.DarkBlue, background = Screen.colors.Gray },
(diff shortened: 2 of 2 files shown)


seandewar:
To fix this for 0.10, I've pushed my suggested change and fixed the failing test.

I've left the `:syn include` for syntax/sh.vim alone; it's not so important to address now. (It can be obsoleted by a move to tree-sitter, anyways :wink:)

En-En-Code:
Thanks! I should have a little more confidence in myself instead of being scared to look at my notifications for nearly a week. :sweat_smile:
I'll take a bit to post-mortem. I do prefer adding the numerals to `iskeyword` as `@,-,_,48-57` is pretty common throughout the documentation, treats numerals the same as letters when adjacent to keywords, and also is a solution to the second problem, at least as far as its current usage is concerned. URL validation is something with enough edge cases and quirks that I personally do not find regex a particularly appealing solution outside of being a crude approximator, which is all it needs to be here.
I should have caught my own test case error. At one point I accidentally ran the test case with `sudo` which caused it to use my default shell (bash) over my personal shell (zsh) and fail due to how the text was wrapped in the former but truncated in the latter. Lesson: cross-platform, or even cross-shell maintenance can be tricky. A better solution would have been to reformat the tutor file itself so that the offending line did not go beyond 80 characters (considering that, IIRC, the default width of the PowerShell interface and GNOME's terminal emulator is 80).
As for developing the tree-sitter for such a simple filetype, I *might* be interested in doing it, as a good next step to learning about the technicals of Neovim, although unfortunately it is no where near my highest programming priority right now.

PR #27834, opened 2024-03-12
doc: add :lua to the list of commands that see :bar as their agrument

Sam-programs:


zeertzjq:
Then the list is still very incomplete...

dundargoc:
> Then the list is still very incomplete...

Wouldn't including this make the list less incomplete? Just making sure, would you prefer not adding this at all or do you mean we should make this list complete?

PR #27843, opened 2024-03-13
refactor: reduce strlen calls in strcat

IAKOBVS:
strcat is O(strlen(dst) + strlen(src)). Calling strcat on the same string is inefficient because it will always unnecessarily recalculate the length of the string from the start of the string.

We can make subsequent strcat calls O(strlen(src)) by caching the length of the string using stpcpy, saving the pointer to the end of the string and using it as dst for later calls.

Relevant: https://github.com/neovim/neovim/pull/27774


IAKOBVS:
The tsan clang functional test failure seems to be irrelevant as it's present in other merged commits.

justinmk:
Sounds good to me. We will likely want to wait until after release before merging this, to allow it to bake.

Meanwhile if you can rebase your PRs every now and then, that will also be very helpful.

justinmk:
Thanks for your attention to performance details :) 

It will help greatly if you only focus on codepaths that matter. `spell_read_aff` for example doesn't matter, it's not a bottleneck. So the reduced readability is not worth the (presumed) performance gain.

src/nvim/memory.h
```diff
@@ -2,12 +2,37 @@
...
+///
+/// WARNING: If copying takes place between objects that overlap, the behavior
+/// is undefined.
+///
+/// Nvim version of POSIX 2008 stpcpy(3). We do not require POSIX 2008, so
+/// implement our own version.
+///
+/// @param dst
+/// @param src
+static inline char *xstpcpy(char *restrict dst, const char *restrict src)
```
justinmk, line 30: why move this into the header?

justinmk:
As mentioned, `spell_read_aff` is not performance-sensitive, so the extra aliasing and indirection is not worth the trouble. If there are more targeted changes you want, please open a new PR.

PR #27854, opened 2024-03-14
fix(api): fix set_lines viewport adjustment, but this time good

### LINKED ISSUE #27720: `nvim_buf_set_lines` adjustment of cursor and of topline is inconsistent
### Problem

```lua
local lines = vim.api.nvim_buf_get_lines(0, 0, -1, false)
vim.api.nvim_buf_set_lines(0, 0, #lines, true, lines)
```

I believe the above code should do nothing to the buffer, but with the latest nightly, buffer is scrolled downwards resulting an inconsistent user experience.

I was not able to bisect the exact commit causing this problem, but I'm pretty sure this was caused by the series of refactoring https://github.com/neovim/neovim/pull/24889 done by @bfredl .

I found several similar issues, but they referred to the issue with some plugin or lsp.format, whereas my _steps to reproduce_ is reproducible without any setup (nvim -u NONE).

### Steps to reproduce

**No config file required.** (nvim -u NONE)

1. Open a file that has more lines than nvim's window height.
3. `G` to go to the bottom.
4. `H` to move the cursor to the top of the window.
5. Run the following code.

```lua
local lines = vim.api.nvim_buf_get_lines(0, 0, -1, false)
vim.api.nvim_buf_set_lines(0, 0, #lines, true, lines)
```

### Result

The buffer and cursor moves down and lands one line above the bottom `scrolloff`. The exact same behavior as doing `zb` after `step 3.`.

### Expected behavior

Do not scroll the buffer, and have no visual change in above case.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2504+g3971797be1

N/A

### Operating system/version

Linux

### Terminal name/version

Wezterm

tmux-256color

### Installation

Build from source

bfredl:
fixes #27720

### DIFF
--- a/src/nvim/mark.c
+++ b/src/nvim/mark.c
@@ -1243,11 +1243,11 @@ void mark_adjust_buf(buf_T *buf, linenr_T line1, linenr_T line2, linenr_T amount
       if (win != curwin || by_api) {
         if (win->w_topline >= line1 && win->w_topline <= line2) {
           if (amount == MAXLNUM) {                  // topline is deleted
-            if (line1 <= 1) {
-              win->w_topline = 1;
+            if (by_api && amount_after > line1 - line2 - 1) {
+              // api: if the deleted region was replaced with new contents, topline will
+              // get adjusted later as an effect of the adjusted cursor in fix_cursor()
             } else {
-              // api: if the deleted region was replaced with new contents, display that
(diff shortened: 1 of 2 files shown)


pysan3:
Thanks again for your amazing work @bfredl ❤️

Issue #22478, opened 2023-03-02
:messages query API, kill "Press ENTER" with fire


justinmk:
### Problem

"Press ENTER" prompt is obnoxious. Most messages do not require explicit user confirmation. No, not even error messages.

The prevalence of "Press ENTER" is mostly an accident because of vim's reliance on stale tty state to show a message.

### Expected behavior

`:messages` should be querable via an API.
All messages emitted by Nvim should have metadata allowing Nvim (and plugins etc) to better decide:

1. which messages genuinely need active user confirmation
2. which messages can be shown passively (in the cmdline/message area)
3. which messages need a pager
4. which messages should be shown for a minimum of X seconds, fading away after some time

### Message metadata

- result (needs pager -- or actual buffer/floating window)
- error (needs message, and be sticky; user cannot silence)
    - most errors don't need confirm. they can be "sticky" by showing them in the cmdline/message area (which may get overwritten by another message, then show `+3 errors` in 'showcmd')
- confirm (invokes "Press ENTER" or something like it)
- info (needs message, but user can decide how it happens)
- progress (be sticky and replaceable)

### Eliminate "Press ENTER" (for 99% of cases)

With the above, Nvim can do smart things like , instead of showing "Press ENTER", show `+3 messages` in the 'showcmd' area (e.g.) and the user can check them if they want (via a mapping like `g<` or `g>`).

"Press ENTER" should be very rare except for actually dangerous situations.
"Editing a readonly file" is not a dangerous situation.

GitMurf:
I apologize for commenting here if not appropriate (if there is a gh issue or discussion that is more appropriate, please point me there as happy to take conversation there)... but I am curious to know where things stand with the `:messages api` (specifically avoiding the "press Enter..." annoyance when setting cmdheight=0). Thanks!

voronind-com:
Any progress on this?

voronind-com:
I ended up patching nvim to massively de-annoy it. It does three things:
1. Disable the hit-enter popup completely.
2. Disable timed delays globally.
3. Always truncate messages so they don't clog the screen.

Now I can finally work w/o interruptions, and so do you: https://github.com/voronind-com/nix/blob/main/patch/nvim/PressEnter.patch

Shougo:
I have added `pressenter` option.

justinmk:
@luukvbaal 's work on the UI layer will help here.

PR #27855, opened 2024-03-14
feat(ui): ext_cmdline/messages for the TUI

luukvbaal:
Problem:  We have an unmaintained Vimscript parser and cmdline
highlighting mechanism, with which it is hard to leverage the
treesitter highlighter. Long messages result in a hit-enter-prompt.

Solution: Implement a vim.ui_attach() UI, that replaces the message
grid (orphaning some 3000+ LOC core C code). Introduce an experimental
vim._extui module, because removing the message grid at the same time is
too risky. The new UI leverages the bundled treesitter highlighter and
parser for Vimscript, as well as the matchparen plugin, to highlight the
cmdline. Messages are truncated in the cmdline area, or placed in a
floating message box in the bottom right corner. Special ("list_cmd")
messages and the message history are shown in a, "more prompt" (now a
fully interactive regular window). Various default UI elements ('showcmd',
'ruler') are also placed in the cmdline area, as virtual text.

`require('vim._extui').enable({})` enables the experimental UI.
`{ msg.pos = 'box' }` or `:set cmdheight=0` enables the message
box variant.

Followup:
  - Come to a consensus for how best to represent messages (by default).
  - Start removing message grid when this is deemed a successful replacement.
    When that is finished, make this new UI the default and update a lot of tests.

---
Given the following highlighted cmdline:
image

Pressing enter with `cfg.msg.pos = 'cmd'` yields a message in the cmdline with duplicate and overflow indicators. User can press <kbd>g<</kbd> to open the more-prompt, now in a (fully interactive) regular window.
image

`cfg.msg.pos = 'box'` yields:
image


clason:
> Come to a consensus for how best to represent messages, alternative:
Only move messages that do not fit the 'cmdheight' area to a dedicated
floating message box.

related: #22478 

TL;DR: we need to distinguish (at least) three kinds of messages
1. single line "status messages" (-> route to `vim.notify`; print on cmdline as default)
2. error/deprecation messages/stacktraces etc. that should not be ignored (-> ???, floating window sounds good but probably needs to go through a `vim.notify` hook for plugins)
3. long-form output like `:map` (-> full buffer for search, yank, etc.)

(Of course, the line between 2 and 3 is blurry -- for example, `:Inspect` output would fit better in a floating window, so expected length of output is pobably a better distinguisher, and a separate "priority" option should be applied to must-read error messages.)

luukvbaal:
My starting point (which I mentioned in https://github.com/neovim/neovim/issues/27811#issuecomment-1997091918, and is now present in the PR) is that _all_ messages go in a (disappearing) floating window. Only when they exceed some length are they moved to a persistent floating window (mimicking a more prompt).

If @neovim/core (or people who try out this feature) feel strongly about still placing messages that fit the 'cmdheight' area where we are used for them to be, we can change that. The logic for determining based on text height is present in this PR. My hope was that people can get used to even small messages appearing in the bottom right corner.

clason:
That's fine; I just wanted to link the context. The main points are that a) we need "message metadata" to allow different handling of messages (length is a good stand-in for an MVP, but maybe not enough in the long run) and b) we need to expose that to plugins like noice that want to shiny up the message UX. Again, not a blocker, just something to keep in mind.

justinmk:
>   When that is finished, make this new UI the default and update aloooot of tests.

Why shouldn't the default behavior continue to show the messages in the cmdline region? 

If a message exceeds (i.e. would trigger "press enter"), just truncate its display. Users can invoke a default mapping such as `g>` to see the pager.

Changing the placement of messages doesn't need to be part of this phase.

luukvbaal:
> Why shouldn't the default behavior continue to show the messages in the cmdline region?

Because we can (with ext_messages and floating windows) easily choose a representation that makes more sense.
The message box representation seemed natural to me, and is what I implemented.
At least with `:set cmdheight=0`, anything other than what we currently have makes more sense IMO.

But I can see how this would be contentious, which is why I asked for input in #27811.
If the old representation needs to stay with (`cmdheight > 0`) that could be done but if it were up to me I would try to avoid it to avoid having to maintain two separate representations.

justinmk:
Just tried this locally, and I'm very excited for it. The highlighting and pager are very cool.

> > Why shouldn't the default behavior continue to show the messages in the cmdline region?
> 
> Because we can (with ext_messages and floating windows) easily choose a representation that makes more sense.

It's not obvious that showing the result of `:ls` in a timed floating window at bottom-right is any better than placing it where it normally goes. Seems somewhat arbitrary. 

Main point I want to make is: this PR is fantastic, and I want to ship it ASAP. Mimicking the legacy behavior (to a reasonable extent) is a two-way door: we can always debate a more "fancy" layout later, it doesn't reduce complexity or risk to decide that *now*. It can easily be a later phase, so it should be.

runtime/lua/vim/_extui.lua
```diff
@@ -0,0 +1,109 @@
...
+      ext.msg.set_pos(ext.cmdheight, 0)
+    end,
+    desc = 'Manipulate cmdline and message window dimensions after shell resize.',
+  })
+
+  api.nvim_create_autocmd('OptionSet', {
+    group = ext.augroup,
+    pattern = { 'cmdheight', 'laststatus', 'termguicolors' },
+    callback = function(ev)
+      if ev.match == 'cmdheight' then
```
gpanders, line 66:
```suggestion
    callback = function(ev)
      ext.tab_check_wins()
      if ev.match == 'cmdheight' then
```

runtime/lua/vim/_extui.lua
```diff
@@ -0,0 +1,110 @@
...
+        -- 'cmdheight' set; (un)hide cmdline window and set its height.
+        ext.cmdheight = vim.v.option_new
+        local cfg = { height = math.max(ext.cmdheight, 1), hide = ext.cmdheight == 0 }
+        pcall(api.nvim_win_set_config, ext.wins[ext.tab].cmd, cfg)
+      elseif ev.match == 'termguicolors' then
+        -- 'termguicolors' toggled; add or remove border and set 'winblend' for message windows.
+        for _, tab in ipairs(api.nvim_list_tabpages()) do
+          local win = ext.wins[tab].msg
+          pcall(api.nvim_win_set_config, win, { border = vim.v.option_new and 'none' or 'single' })
+          pcall(api.nvim_set_option_value, 'winblend', vim.v.option_new and 30 or 0, { win = win })
```
gpanders, line 77: AFAICT winblend only has an effect when `Normal` has a guifg/bg color defined. I keep `Normal` cleared in my colorscheme (it just uses default fg/bg colors from the terminal) so these options make the message window have a plain black background with no border. In addition to setting these options when `'termguicolors'` is set I suggest also checking the values of the `Normal` highlight group.

luukvbaal: Well the idea was to have an invisible message box (`MsgArea` linked to `Normal` by default); it should not be black unless your terminal background is indeed black, or you define `MsgArea` with a black background.

`'winblend'` helps to discern text that runs behind the floating window, but that doesn't work with `'notermguicolors'`, in which case a border is needed. Otherwise I don't think a cleared `Normal` group is relevant; it should still give alpha to text below the float, but maybe I'm misunderstanding.

gpanders: >Clearing Normal should not yield black background unless your terminal background is indeed black, or you define MsgArea with a black background.

That's what I'm saying, this is not the case. The `'winblend'` option creates a black background when `Normal` is cleared. You can try it for yourself and see. I don't know if this is intended behavior or not (the compositor presumably needs to know the default fg/bg color to perform the blending which it doesn't know if `Normal` is cleared).

luukvbaal: I'm not sure what to try, could you share reproduction steps and or a screenshot of what you're seeing?

gpanders: With this PR display any kind of message (e.g. `Ctrl-G`) with `Normal` cleared and then again with `Normal` defined with a gui fg/bg color.

With Normal cleared, I see this:

Screenshot from 2025-01-03 09-24-24

With Normal defined with the same fg/bg colors as my terminal, I see this:

Screenshot from 2025-01-03 09-24-35

luukvbaal: Interesting, not sure what causes that; it doesn't reproduce from `build/bin/nvim --clean +"hi clear Normal" +"set tgc" +"lua require('_extui').enable(true)"` so more info is needed.

przepompownia:
Side effects with merged-in master branch here:

- `nvim --clean -V9 -c "lua require('vim._extui').enable(true)" runtime/lua/vim/lsp/buf.lua` - crashes after [q]uitting from initial (not redirected yet) messages:
```cpp
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007fce587dedef in __pthread_kill_internal (threadid=<optimized out>, signo=6) at ./nptl/pthread_kill.c:78
#2  0x00007fce5878ad02 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fce587734f0 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007fce58773418 in __assert_fail_base (fmt=0x7fce588f7ca0 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x5615277b57a0 "msg_ext_kind == kind",
    file=file@entry=0x5615277d76b0 "/home/arctgx/dev/external/neovim-dev/src/nvim/message.c", line=line@entry=315, function=function@entry=0x56152778daa8 <__PRETTY_FUNCTION__.17> "msg_multihl") at ./assert/assert.c:94
#5  0x00007fce58783612 in __assert_fail (assertion=assertion@entry=0x5615277b57a0 "msg_ext_kind == kind", file=file@entry=0x5615277d76b0 "/home/arctgx/dev/external/neovim-dev/src/nvim/message.c", line=line@entry=315,
    function=function@entry=0x56152778daa8 <__PRETTY_FUNCTION__.17> "msg_multihl") at ./assert/assert.c:103
#6  0x00005615274e898c in msg_multihl (hl_msg=..., kind=0x5615277a9a20 "echomsg", history=true, err=false) at /home/arctgx/dev/external/neovim-dev/src/nvim/message.c:315
#7  0x000056152734c87c in nvim_echo (chunks=..., history=<optimized out>, opts=0x7fff936a533e, err=0x7fff936a53e0) at /home/arctgx/dev/external/neovim-dev/src/nvim/api/vim.c:795
#8  0x000056152733c99d in handle_nvim_echo (channel_id=<optimized out>, args=..., arena=<optimized out>, error=0x7fff936a53e0) at /home/arctgx/dev/external/neovim-dev/build/src/nvim/auto/api/private/dispatch_wrappers.generated.h:7936
#9  0x0000561527503421 in request_event (argv=0x24f71) at /home/arctgx/dev/external/neovim-dev/src/nvim/msgpack_rpc/channel.c:366
#10 0x00005615275d6f06 in state_handle_k_event () at /home/arctgx/dev/external/neovim-dev/src/nvim/state.c:120
#11 0x00005615275126c4 in nv_event (cap=<optimized out>) at /home/arctgx/dev/external/neovim-dev/src/nvim/normal.c:6634
#12 0x00005615275068e4 in normal_execute (state=0x7fff936a5580, key=<optimized out>) at /home/arctgx/dev/external/neovim-dev/src/nvim/normal.c:1244
#13 0x00005615275d6d9d in state_enter (s=0x7fff936a5580) at /home/arctgx/dev/external/neovim-dev/src/nvim/state.c:102
#14 0x0000561527503bd4 in normal_enter (cmdwin=<optimized out>, noexmode=<optimized out>) at /home/arctgx/dev/external/neovim-dev/src/nvim/normal.c:522
#15 0x0000561527319157 in main (argc=<optimized out>, argv=<optimized out>) at /home/arctgx/dev/external/neovim-dev/src/nvim/main.c:651
```

- `nvim --clean -c "lua require('vim._extui').enable(true)" runtime/lua/vim/lsp/buf.lua -c 'se verbose=9'`
```cpp
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007f07527a2def in __pthread_kill_internal (threadid=<optimized out>, signo=6) at ./nptl/pthread_kill.c:78
#2  0x00007f075274ed02 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f07527374f0 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007f0752737418 in __assert_fail_base (fmt=0x7f07528bbca0 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x55d306d0194c "pos_delta >= 0",
    file=file@entry=0x55d306d236b0 ".../src/nvim/message.c", line=line@entry=2476, function=function@entry=0x55d306cd9ef0 <__PRETTY_FUNCTION__.0> "msg_scroll_flush") at ./assert/assert.c:94
#5  0x00007f0752747612 in __assert_fail (assertion=assertion@entry=0x55d306d0194c "pos_delta >= 0", file=file@entry=0x55d306d236b0 ".../src/nvim/message.c", line=line@entry=2476,
    function=function@entry=0x55d306cd9ef0 <__PRETTY_FUNCTION__.0> "msg_scroll_flush") at ./assert/assert.c:103
#6  0x000055d306a3abae in msg_scroll_flush () at .../src/nvim/message.c:2476
#7  0x000055d306b69254 in ui_flush () at .../src/nvim/ui.c:541
#8  0x000055d3068ef9ed in ui_flush () at .../src/nvim/drawscreen.c:404
#9  screen_resize (width=179, height=<optimized out>) at .../src/nvim/drawscreen.c:401
#10 screen_resize (width=<optimized out>, height=<optimized out>) at .../src/nvim/drawscreen.c:269
#11 0x000055d306b64161 in ui_refresh () at .../src/nvim/ui.c:245
#12 0x000055d306b64f98 in ui_remove_cb (ns_id=2, checkerr=true) at .../src/nvim/ui.c:799
#13 0x000055d306b22f06 in state_handle_k_event () at .../src/nvim/state.c:120
#14 0x000055d306a5e6c4 in nv_event (cap=<optimized out>) at .../src/nvim/normal.c:6634
#15 0x000055d306a528e4 in normal_execute (state=0x7ffe3ebf3720, key=<optimized out>) at .../src/nvim/normal.c:1244
#16 0x000055d306b22d9d in state_enter (s=0x7ffe3ebf3720) at .../src/nvim/state.c:102
#17 0x000055d306a4fbd4 in normal_enter (cmdwin=<optimized out>, noexmode=<optimized out>) at .../src/nvim/normal.c:522
#18 0x000055d306865157 in main (argc=<optimized out>, argv=<optimized out>) at .../src/nvim/main.c:651
```
Not sure if worth report while testing on combined commit. On the other side there are only Lua changes so probably both cases can be minimized to something reproducible on master.

Background: I wanted to check at first if WinScrolled autocmds are (rather unwantedly) triggered on the message window.

luukvbaal:
Thanks for reporting, I'm not too concerned with the second backtrace, which happens after the extui was already forcibly detached due to excessive errors. 'verbose' was not considered yet, good to look into thanks.

First backtrace hits an assert that we want to avoid (recently added to make sure), but is also not too critical. Both will  be easier resolve once the message grid is no more but I can check if there is an easy fix in the meantime.

przepompownia:
Now (after merging master again here) on `nvim --clean -c "lua require('vim._extui').enable(true)" runtime/lua/vim/lsp/buf.lua -c 'se verbose=9'` and hitting Enter enough I get 

```
Error executing vim.schedule lua callback: ...m-src-dev/share/nvim/runtime/lua/vim/_extui/messages.lua:276: Invalid window id:
1015
stack traceback:
        [C]: in function 'nvim_win_set_config'
        ...m-src-dev/share/nvim/runtime/lua/vim/_extui/messages.lua:276: in function 'win_set_pos'
        ...m-src-dev/share/nvim/runtime/lua/vim/_extui/messages.lua:299: in function 'set_pos'
        ...m-src-dev/share/nvim/runtime/lua/vim/_extui/messages.lua:104: in function 'show_msg'
```

przepompownia:
Let's get around the above issue with `verbose` to show the case with `WinScrolled`:

- `nvim --clean -c "lua require('vim._extui').enable(true)" -c 'au WinScrolled * echo rand()'`
- `:echo 'x'`

Is it possible to ignore autocommand events on the message window? I tried to wrap displaying message window and updating its buffer by `vim._with({noautocmd = true},`, but with no result. 

`WinScrolled` with (anti?)pattern `*` appears even in clean Nvim (`call s:Highlight_Matching_Pair()`), but also in plugins like nvim-tresitter-context.

luukvbaal:
Yes, autocommands firing for the extui buffers/windows is probably unwanted, at least publicly (namespaced autocommands/editor objects(window buffer) could still be useful for the event handler itself). Perhaps this warrants the more general API of marking a window as a UI element which I alluded to in #27984, #31269.

We can ourselves set `'eventignore' = "all"` and restore it in the UI callback, but that doesn't avoid firing `WinScrolled`; which fires after the UI callback has already returned (it's checked every event loop iteration, rather than fired when scrolling actually happens).

przepompownia:
Quick check without detailed debugging shows me that
- for some reason set `vim.go.eventignore = 'all'` (and empty after) directly works unlike from `vim._with`,
- `OptionSet` seems to be the edge case, for which I cannot avoid displaying verbose messages about firing that event,
- verbosity (at level 9 and probably not only) produces junky `\n` messages with no kind:
image

runtime/lua/vim/_extui/messages.lua
```diff
@@ -0,0 +1,306 @@
...
+        end_row = row,
+        hl_group = chunk[3],
+        invalidate = true,
+        undo_restore = false,
+        end_right_gravity = (i == #content), -- extend highlight for dupe/trunc indicator.
+      }
+    end
+  end
+
+  local trunc = target == 'cmd' and row + 1 - ext.cmdheight
```
chemio9, line 88: trunc might be `false` or a number

runtime/lua/vim/_extui/cmdline.lua
```diff
@@ -0,0 +1,144 @@
...
+---@param content CmdContent
+---@param prompt string
+local function set_text(content, prompt)
+  for _, chunk in ipairs(content) do
+    prompt = prompt .. chunk[2]
+  end
+  api.nvim_buf_set_lines(ext.bufs.cmd, row, -1, false, { prompt .. ' ' })
+end
+
+local promptlen = 0 -- Current length of the prompt, stored for use in "cmdline_pos"
```
przepompownia, line 24: Would it be possible (and would it make sense) to provide that length in `cmdline_pos` event to avoid storing it on the user side?

luukvbaal: Possible yes, but somewhat hard to describe what it is, and possibly not useful for external GUIs. Any thoughts @bfredl on adding what would essentially be:

    Byte-offset of `cmdline_show->prompt` plus current cursor position in the `cmdline_show->content`

That is what a `vim.ui_attach()` implementation needs to position the cursor. Whereas a GUI likely would use a self-calculated cell-offset (not sure)? Could add it to add to `cmdline_pos` despite that I guess since it is useful for at least some subscribers of the event.

luukvbaal: Possible yes, but I think a GUI would probably use a self-calculated cell-offset rather than this byte-offset (not sure). 

Do you think it would make sense to add the byte-offset of the `cmdline_pos->prompt` to the `cmdline_show` event as a third argument @bfredl?
That is what a `vim.ui_attach()` implementation needs to position the cursor.

przepompownia:
After switching to use ext_cmdline I noticed (both here and in my plugin) that wildmenu does not work. Are we forced to implement `ext_popupmenu` behavior too in that case? If so, it involves implementing custom ins-completion menu what I would like to avoid.

luukvbaal:
See #31269 @przepompownia.

przepompownia:
Is it currently possible to get the correct cmdline window height in multiline commands, e.g. `:if 1<CR>`? I see that the example is not related to `cmdline_block_*` event.

here:
image
built-in:
image
At this point, it's more out of curiosity than necessity.

luukvbaal:
> I see that the example is not related to cmdline_block_* event.

I think that is an oversight/bug. This pattern should emit `cmdline_block` events.

runtime/lua/vim/_extui/cmdline.lua
bfredl: hmm, I don't think it is necessary that _any_ internal implementation would implement prompts using string concatenation. 

Storing state between ui events is necessary in general, so storing `#promt` or whatever on the lua side shouldn't be a big deal.

luukvbaal:
Marking this as ready, but this still depends on #31269 and #32835.

runtime/lua/vim/_extui.lua
```diff
@@ -0,0 +1,98 @@
...
+  api.nvim__redraw({
+    flush = true,
+    cursor = handler == ext.cmd[event] and true or nil,
+    win = handler == ext.cmd[event] and ext.wins[ext.tab].cmd or nil,
+  })
+end
+local scheduled_ui_callback = vim.schedule_wrap(ui_callback)
+
+M.enable = function(opts)
+  ext.cfg = vim.tbl_deep_extend('keep', opts, ext.cfg)
```
przepompownia, line 23: `opts` seems to need validation (try opts = nil for example).

przepompownia:
Cannot keep `cmdheight=0`. Try `nvim --clean --cmd "lua require('vim._extui').enable({enable = true})" -c 'set cmdheight=0'`.

luukvbaal:
This module defines an `OptionSet` autocommand that forces `cfg.msg.pos = 'box'` with `'cmdheight' = 0`, but that doesn't fire during startup. For now either set cmdheight _before_ enabling extui or schedule either one of them.

przepompownia:
The recent commit fixed also invisible `echo` message (before I reported it) in the case mentioned above, thanks!

runtime/lua/vim/_extui.lua
```diff
@@ -0,0 +1,106 @@
...
+    end
+    for _, buf in pairs(ext.bufs) do
+      api.nvim_buf_delete(buf, {})
+    end
+    api.nvim_clear_autocmds({ group = ext.augroup })
+    vim.ui_detach(ext.ns)
+    return
+  end
+
+  api.nvim_set_hl(ext.ns, 'Normal', { link = 'MsgArea' })
```
przepompownia, line 40: I tried to link `Normal` differently in the history and the message window (to dim only disappearing messages) but currently it seems to be not simple from the user side (at first glance it requires `=require('vim._extui.shared').wins[tabnr].box` and set `winhighlight`) or I have overlooked something.

luukvbaal: Yes a new highlight group to apply to the message box is probably in order. Also not yet applying `MoreMsg` to the new more window.

przepompownia:
Now `nvim-src-clean -c "lua require('vim._extui').enable({enable = true})"` (without `set cmdheight`) the cmdline area is invisible (regardless of redrawing).

luukvbaal:
Please make sure your runtime actually points to this branch when testing this, because it isn't invisible on my end...

przepompownia:
NVIM v0.12.0-dev-180+gce26a54c4e, built again and installed after `make disclean` and manual `MAKE_INSTALL_PREFIX` subdirectory removal. 

Below: what I see after `:echom 1` and then `:mes`:

https://github.com/user-attachments/assets/5b0aa307-80a5-4361-9e9c-bea742ae185e

luukvbaal:
IDK @przepompownia cannot reproduce. Please try `make && VIMRUNTIME=$(git rev-parse --show-toplevel)/runtime build/bin/nvim --clean +"lua require('vim._extui').enable({})"`

przepompownia:
> IDK @przepompownia cannot reproduce. Please try `make && VIMRUNTIME=$(git rev-parse --show-toplevel)/runtime build/bin/nvim --clean +"lua require('vim._extui').enable({})"`

The same for me. 

On the other side, at the moment I see that the problem is related to hiding window logic: none of (commented below) conditions is satisfied (unconditional unhiding makes the cmdline window visible) 
image

luukvbaal:
If cmdheight > 0 the cmdline window is created with `hide = false` and should thus never have to be unhidden in the first place.

przepompownia:
`hide` is enabled inside `tab_check_wins` because `M.cmdheight == 0` (in `shared`).

For me, the `OptionSet` autocmd is fired for `termguicolors` (and then calls `tab_check_wins`) before `VimEnter`, so the command window isn't initialized again from `ui_callback`.

luukvbaal:
I think #33546 will avoid these problems.

przepompownia:
> I think #33546 will avoid these problems.

Good idea, regardless of how much it helps avoid the problem here (especially if we still will have to defer call ui_attach after VimEnter).

przepompownia:
The above problem seems to be fixed after #33546

przepompownia:
Additionally, I found a problem resulting from my outdated configuration. The repro can be reduced to
```sh
nvim --cmd 'au UIEnter * se ch=0' --cmd "lua require('vim._extui').enable({enable = true})"
```
In this case, the statusline will not be displayed until you enter the cmdline mode.

runtime/lua/vim/_extui/messages.lua
```diff
@@ -0,0 +1,394 @@
...
+    end
+    api.nvim_buf_set_lines(buf, 0, len, false, {})
+    self.count = self.count - 1
+    -- Resize or hide message box for removed message.
+    if self.count > 0 then
+      M.set_pos('box')
+    else
+      self.width = 1
+      M.prev_msg = ext.cfg.msg.pos == 'box' and '' or M.prev_msg
+      api.nvim_win_set_config(ext.wins[ext.tab].box, { hide = true })
```
przepompownia, line 55: Need to check `nvim_win_is_valid`: I have `vim.keymap.set('n', '<Esc>', vim.cmd.fclose)` and sometimes I hide the window before its timer expires.

runtime/lua/vim/_extui/messages.lua
```diff
@@ -0,0 +1,394 @@
...
+      api.nvim_buf_clear_namespace(ext.bufs.box, -1, 0, -1)
+    end
+  end, 4000)
```
przepompownia, line 58: It should be configurable.

przepompownia:
```lua
vim.schedule(function()
  vim.notify('w', vim.log.levels.WARN)
end)
vim.schedule(function()
  vim.notify('e', vim.log.levels.ERROR)
end)
```
with `set ch=0` displays only `e` for me. Both messages are visible by `:mes`.

luukvbaal:
Might not be an issue with the implementation here. For whatever reason `msg_show->replace_last == true` for the second message. 

The only place I know of where `replace_last` is useful for us is where Nvim sends two message events for an incomplete message to show busy state, later to be appended to (when writing a file).

But maybe we should just identify those two messages by it's content because I don't think we want to honor `replace_last` anywhere else. It's only useful to exactly mimic the current message grid implementation when it replaces a message that will not be scrolled.

przepompownia:
I didn't look how messages with `search_count`, `bufwrite`, `undo` and `completion` kinds are handled here but found `replace_last` flag useful for those kinds (and nothing else) when started writing my custom plugin.

przepompownia:
The recent commit makes the above case working as expected, thanks!

przepompownia:
Do you consider adding (optional?) LSP progress handler here or (if it could be out of scope of this PR) at least some user documentation how to implement it?

luukvbaal:
Definitely seems out of scope to me, I guess you want to re-use the message box for that?

przepompownia:
Yes, it would be a pity to waste this opportunity. At the first glance `M.show_msg` allows replacing the recent message, but there is no guarantee that it would be a progress message. I cannot find here a way how to access a message marked before as a special and update it (and probably display it at some fixed position, .e.g. lowest).

clason:
Definitely out of scope. Recall the stated problem:

>  We have an unmaintained Vimscript parser and cmdline
highlighting mechanism, with which it is hard to leverage the
treesitter highlighter. Long messages result in a hit-enter-prompt.

Let's get this solved; then we can go from there. This PR has been open for more than a year; please help get it done, not delay it further.

luukvbaal:
The cmdline shouldn't have a border (or respect 'winborder' for now, I think). If it does that's a bug, one that I haven't encountered.

I'll add a mapping for `ctrl-c`.

przepompownia:
This PR has been open for more than a year; please help get it done, not delay it further.

I also wouldn't mind merging this PR as soon as possible and can create a separate feature request for message updates.

bfredl (approved):
Should be low-risk to merge this now, and iterate and fix things on master.

We can decide when 0.12 is drawing closer whether it is ready to be advertised as a feature or remain a `WARNING: EXPERIMENTAL` thing for that release.

runtime/lua/vim/_extui.lua
```diff
@@ -0,0 +1,129 @@
...
+      scheduled_ui_callback(event, ...)
+    else
+      ui_callback(event, ...)
+    end
+  end)
+
+  api.nvim_set_hl(ext.ns, 'Normal', { link = 'MsgArea' })
+  api.nvim_set_hl(ext.ns, 'Search', { link = 'MsgArea' })
+  api.nvim_set_hl(ext.ns, 'CurSearch', { link = 'MsgArea' })
+  api.nvim_set_hl(ext.ns, 'IncSearch', { link = 'MsgArea' })
```
brianhuster, line 73: May I know why these hl groups are linked to 'MsgArea'? It makes me unable to see highlighting in messages window

luukvbaal: Yes needs more specificity. With how the various windows in this module evolved we probably only want these links in the cmdline window.

brianhuster: Could I send a PR to remove them? I just don't understand why they should be linked to MsgArea. Or are you planning to add `MsgSearch`, `MsgCurSearch`, `MsgIncSearch` highlight groups, so those lines are placeholders?

luukvbaal: They are needed in the cmdline window to avoid highlighting the entered search command itself.

luukvbaal: #33792

garydwatson:
this is probably the wrong place to say this seeing as how this is already merged... however... when enabling this feature by...
require('vim._extui').enable({}) enables the experimental UI.
{ msg.pos = 'box' } or :set cmdheight=0 enables the message

it completely breaks the :ls / :buffers command... the list of buffers has a paged view that can be scrolled up and down either line by line or a page at a time, and futher, you can hit ":" at any point while looking through the list to enter command mode so you could do a ":bd 1 7 8 9 42" while looking at the list...

if you replace that very fully functional view with this, then buffer navigation and management will effectively be broken.

justinmk:
> the :ls / :buffers command

Use `g<` to enter the pager. There will also be ways to change the behavior. And we're still working out the default behavior.

PR #27858, opened 2024-03-14
fix(ui): issues controlling cursor position with ext_cmdline

### LINKED ISSUE #20309: control the visible cursor with getchar(), vim.ui_attach+ext_cmdline, etc
### Neovim version (nvim -v)

nightly

yes

### Operating system/version

Linux

### Terminal name/version

kitty

xterm-kitty

### Installation

bob

I'm building a custom cmdline using `vim.ui_attach`.

Through the API, we get enough information to show a fake cursor at the command prompt, but the original cursor also stays on the screen where it was before entering the cmdline.

See screenshot below.

image

### Expected behavior

When implementing ext_cmdline, the cursor should be hidden between `cmdline_show` and `cmdline_hide` events.

### Actual behavior

The cursor is still there

### LINKED ISSUE #20793: cursor doesn't redraw with getchar
### Neovim version (nvim -v)

nightly

not relevant

### Operating system/version

linux

### Terminal name/version

kitty

xterm-kitty

### Installation

bob

`:luafile %` the file below

```lua
vim.fn.getcharstr()
vim.api.nvim_win_set_cursor(0, { 1, 10 })
-- after this, the cursor will not be shown at {1,10}
vim.fn.getcharstr()
-- now it will show at the correct position
print("done")
```

### Expected behavior

After the first `getcharstr`, the visible cursor should be at `{1,10}`

### Actual behavior

The cursor is still shown at the original position and will move to `{1,10}` after the second `getcharstr` finishes.

luukvbaal:
**fix(api): update grid cursor in nvim_win_set_cursor()**

Problem:  Cursor position set by nvim_win_set_cursor() is not reflected
          on the screen when followed by a blocking call like getchar().
Solution: Immediately update the cursor position on the grid.

Fix #20793

**fix(ui): ext_cmdline should not move cursor to curwin**

Problem:  The ext_cmdline cursor position on the screen seems to rely on
          an implicit assumption that the event listener implements a
          cmdline window that is made the current window which is
          problematic (e.g. breaks 'incsearch' in the actual current
          window).
Solution: Remove this assumption and allow nvim_win_set_cursor() to move
          the cursor on the screen to a non-current window (previous
          commit).

Fix #20309

### DIFF
--- a/src/nvim/api/window.c
+++ b/src/nvim/api/window.c
@@ -148,7 +148,7 @@ void nvim_win_set_cursor(Window window, ArrayOf(Integer, 2) pos, Error *err)
   switchwin_T switchwin;
   switch_win(&switchwin, win, NULL, true);
   update_topline(curwin);
-  validate_cursor(curwin);
+  setcursor_mayforce(true);
   restore_win(&switchwin, true);
 
   redraw_later(win, UPD_VALID);
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -767,7 +767,7 @@ static uint8_t *command_line_enter(int firstc, int count, int indent, bool clear
   }
 
   setmouse();
-  ui_cursor_shape();               // may show different cursor shape
+  setcursor();
 
   TryState tstate;
   Error err = ERROR_INIT;
@@ -927,7 +927,6 @@ static uint8_t *command_line_enter(int firstc, int count, int indent, bool clear
   }
   may_trigger_modechanged();
   setmouse();
-  ui_cursor_shape();            // may show different cursor shape
   sb_text_end_cmdline();
 
 theend:
@@ -3859,7 +3858,6 @@ void cursorcmd(void)
     if (ccline.redraw_state < kCmdRedrawPos) {
       ccline.redraw_state = kCmdRedrawPos;
     }
-    setcursor();
     return;
   }
 
@@ -4553,6 +4551,7 @@ static int open_cmdwin(void)
   State = save_State;
   may_trigger_modechanged();
   setmouse();
+  setcursor();
 
   return cmdwin_result;
 }
--- a/test/functional/api/window_spec.lua
+++ b/test/functional/api/window_spec.lua
@@ -219,6 +219,21 @@ describe('API/win', function()
 
       -- curwin didn't change back
       neq(win, curwin())
+
+      -- shows updated position after getchar() #20793
+      feed(':call getchar()<CR>')
+      api.nvim_win_set_cursor(win, { 1, 5 })
+      screen:expect {
+        grid = [[
+                                      |
+        {1:~                             }|*2
+        {2:[No Name]                     }|
+        prolo^gue                      |
+                                      |*2
+        {3:[No Name] [+]                 }|
+        :call getchar()               |
+      ]],
+      }
     end)
 
     it('remembers what column it wants to be in', function()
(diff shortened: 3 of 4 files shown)


src/nvim/api/window.c
```diff
@@ -134,25 +153,32 @@ void nvim_win_set_cursor(Window window, ArrayOf(Integer, 2) pos, Error *err)
...
+    validate_cursor(curwin);
+    if (opts->flush) {
+      setcursor_mayforce(true);
+    }
+    restore_win(&switchwin, true);
+    redraw_later(win, UPD_VALID);
+    win->w_redr_status = true;
+  } else {  // screen
+    ui_grid_cursor_goto(default_grid.handle, (int)row - 1, (int)col);
+    skip_cursor_update = true;
```
bfredl, line 180: just my immediate reaction, but this part seems to be _very_ different functionality, to the extent it should be a separate API call altogether.

luukvbaal: I see. Like I said I didn't even have a use case for it.
I just went "no opts array" -> nvim_set_cursor() -> add 2 lines for screen relative cursor.
If it isn't wanted perhaps this should just be nvim_set_cursor2()?

zeertzjq:
Use (1,0) indexing for screen position is strange, as it only makes sense for text position.

luukvbaal:
> Using (1,0) indexing for screen position is strange.

I thought so too but using different indexing for different options in the same functions was even stranger to me. Might be irrelevent if we're not adding `screen`.

luukvbaal:
> Might be irrelevent if we're not adding screen.

In that case we might as well forego this new API and just resolve this by having `nvim_win_set_cursor()` call `setcursor_mayforce()`. I don't think there is any harm in that?

src/nvim/ex_getln.c
```diff
@@ -767,7 +767,7 @@ static uint8_t *command_line_enter(int firstc, int count, int indent, bool clear
   }
 
   setmouse();
-  ui_cursor_shape();               // may show different cursor shape
+  setcursor();
```
luukvbaal, line 770: Instead, move the cursor back to curwin when leaving the cmdline (and remove duplicate call to `ui_cursor_shape()`).

zeertzjq: Is removing duplicate call to `ui_cursor_shape()` really necessary? It doesn't do anything if editor state hasn't changed.

luukvbaal: No but why would we keep it? It's called in setmouse() in the line above.

zeertzjq: Oh, I thought you were talking about `command_line_check()`.

bfredl (approved):
Yes, I think the behavior in the added tests make a lot of sense.

echasnovski:
I am not sure forcing cursor draw in not-current window is good behavior. This might lead to visually indicating that current window has changed when in reality it has not. And if it is not a huge deal in regular Nvim TUI, in Neovide it might lead to [this kind of behavior](https://github.com/neovide/neovide/issues/2466#issuecomment-2041117466).

The use case here is that user input is taken one by one via `getchar()` (technically, `getcharstr()`), while calling `nvim_win_set_cursor()` on the not current window which is never meant to be focused. The `nvim_win_set_cursor()` is usually needed to ensure the proper window view: either first line as top line ('mini.clue') or cursorline highlighting (which *I think* is what is done in 'nvim-cmp').

PR #27861, opened 2024-03-14
feat(man): allow opening pages in current window

### LINKED ISSUE #27860: Allow opening man pages in the current window
### Problem

Currently `:Man` forces using separate window unless the current buffer has alerady `man` filetype. It can be simply achieved by adding simple modifier to `man.open_page`, for example as a fake modifier:

```diff
diff --git a/runtime/lua/man.lua b/runtime/lua/man.lua
index ac15aff60..0657e4e9e 100644
--- a/runtime/lua/man.lua
+++ b/runtime/lua/man.lua
@@ -716,7 +716,7 @@ function M.open_page(count, smods, args)
   local target = ('%s(%s)'):format(name, sect)

   local ok, ret = pcall(function()
-    if smods.tab == -1 and find_man() then
+    if smods.curwin or (smods.tab == -1 and find_man()) then
       vim.cmd.tag({ target, mods = { silent = true, keepalt = true } })
     else
       smods.silent = true
```

In practice I'd like to open man page by nvim from the shell without using `MANPAGER` environment variable. 

By default I run `nvim` inside `systemd-run` container (I hope "container" it's the correct term) and have terminal related issues (probably because `man` is run outside the container, but `nvim` inside it). 

In practice, the number of lines are not correctly refreshed after zooming in and out in tmux window.

### Expected behavior

- allow opening man pages in the current window at least from `man.open_page`
- optionally extend `Man` command to allow use it too

przepompownia:
Fixes #27860

### DIFF
--- a/runtime/doc/filetype.txt
+++ b/runtime/doc/filetype.txt
@@ -660,7 +660,7 @@ soft wrapped as with `g:man_hardwrap=0`. You can set in your environment: >sh
      export MANWIDTH=999
 An alternative to using `MANPAGER` in shell can be redefined `man`, for example: >sh
      man() {
-	nvim "+hide Man $1"
+          nvim "+hide Man $1"
      }
 
 So `groff`'s pre-formatting output will be the same as with `g:man_hardwrap=0` i.e soft-wrapped.
--- a/runtime/doc/filetype.txt
+++ b/runtime/doc/filetype.txt
@@ -658,16 +658,16 @@ will inevitably display the manual page as it was passed to it from stdin. One
 of the caveats of this is that the width will _always_ be hard-wrapped and not
 soft wrapped as with `g:man_hardwrap=0`. You can set in your environment: >sh
      export MANWIDTH=999
-An alternative to using `MANPAGER` in shell can be redefined `man`, for example: >sh
-     man() {
-          nvim "+hide Man $1"
-     }
 
 So `groff`'s pre-formatting output will be the same as with `g:man_hardwrap=0` i.e soft-wrapped.
 
 To disable bold highlighting: >
      :highlight link manBold Normal
 
+An alternative to using `MANPAGER` in shell can be redefined `man`, for example: >sh
+     man() {
+          nvim "+hide Man $1"
+     }
 
 MARKDOWN                                                *ft-markdown-plugin*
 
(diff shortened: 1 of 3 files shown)


zeertzjq:
Adding an `smods` field that can only be used when calling `open_page` from Lua is a bit strange, given that `open_page()` is not a documented function and may be refactored in the future. How about making `:hide Man` open in current window instead? That may also be a bit strange, but at least `:Man` is a documented command.

przepompownia:
Honestly, I missed existence of `hide` among other command modifiers so thanks for the suggestion.

runtime/doc/filetype.txt
```diff
@@ -626,8 +626,10 @@ Man                       Open the manpage for the <cWORD> (man buffers)
                           or <cword> (non-man buffers) under the cursor.
 Man!                      Display the current buffer contents as a manpage.
 
-|:Man| accepts command modifiers. For example, to use a vertical split: >
+|:Man| accepts command modifiers. For example, to use a vertical split: >vim
      :vertical Man printf
+To reuse the current window: >vim
+     :hide Man man
```
zeertzjq, line 632:
```suggestion
     :hide Man printf
```

runtime/doc/news.txt
```diff
@@ -253,6 +253,7 @@ The following new APIs and features were added.
   • The `#set!` directive can set the "url" property of a node to have the
     node emit a hyperlink. Hyperlinks are UI specific: in the TUI, the OSC 8
     control sequence is used.
+  • |:Man| now supports `hide` modifier to open page in the current window
```
zeertzjq, line 256:
```suggestion

• |:Man| now supports `hide` modifier to open page in the current window
```

przepompownia:
Thanks!

runtime/doc/filetype.txt
```diff
@@ -647,21 +649,25 @@ Variables:
...
      export MANWIDTH=999
 
 So `groff`'s pre-formatting output will be the same as with `g:man_hardwrap=0` i.e soft-wrapped.
 
 To disable bold highlighting: >
      :highlight link manBold Normal
 
+An alternative to using `MANPAGER` in shell can be redefined `man`, for example: >sh
+     man() {
+          nvim "+hide Man $1"
```
bew, line 669: Note that by doing this you'll most probably not have man completions from your shell (or you'll have to add it manually)

przepompownia: Example for bash (5.2.21(1)-release): 

image

PR #21709, opened 2023-01-09
fix(ui): command line issues with external messages

luukvbaal:
Resolve https://github.com/neovim/neovim/issues/21653. Follow-up to https://github.com/neovim/neovim/pull/21580.

Issue #21653, opened 2023-01-05
Message specifications when saved have changed since #21580


yutkat:
### Neovim version (nvim -v)

v0.9.0-dev-1827+gae64772a8

### Vim (not Nvim) behaves the same?

no

### Operating system/version

Arch Linux

### Terminal name/version

wezterm 20230102-155114-67896f67

### $TERM environment variable

xterm-256color

### Installation

build from repo

### How to reproduce the issue

1. use this init.lua

```lua
vim.cmd [[syntax enable]]
vim.cmd [[filetype plugin indent on]]

local install_path = vim.fn.stdpath('data')..'/site/pack/packer/opt/packer.nvim'
if vim.fn.empty(vim.fn.glob(install_path)) == 1 then
  vim.api.nvim_command('!git clone https://github.com/wbthomason/packer.nvim '..install_path)
end
vim.cmd [[packadd packer.nvim]]

local function load_plugins()
  return require('packer').startup(function()
    use {'wbthomason/packer.nvim', opt = true}
    use({
      "folke/noice.nvim",
      config = function()
        require("noice").setup({})
      end,
      requires = {
        "MunifTanjim/nui.nvim",
        "rcarriga/nvim-notify",
      }
    })
  end)
end
load_plugins()
```

2. `:PackerSync` and `:qa`
3. open `nvim aaa.lua`
4. `:w`
5. Normally, the file name of the saved file(aaa.lua) should be displayed, but `<` is displayed.

If this added condition is removed, it will be normal. https://github.com/neovim/neovim/pull/21580/files#diff-93eb182156bc96b7817704f1db2a233ebdaf3d90a93979e8bb9ae1d3f5e8757aR3844

```c
&& p_ch > 0
```

Maybe it is a problem on the plugin side(noice.nvim), if so, I'm sorry.

### Expected behavior

Saved file names are displayed correctly.
![swappy-20230105-174251](https://user-images.githubusercontent.com/8683947/210739038-e4992163-32aa-44ca-a4f0-a1b31cee16b7.png)

### Actual behavior

`<` is displayed
![swappy-20230105-174313](https://user-images.githubusercontent.com/8683947/210739058-258f4260-3915-491a-8520-b89a1152974a.png)

bfredl:
>   https://github.com/neovim/neovim/pull/21580/files#diff-93eb182156bc96b7817704f1db2a233ebdaf3d90a93979e8bb9ae1d3f5e8757aR3844

[...]

PR #21580, opened 2022-12-29
fix(ui): command line issues with 'cmdheight' set to zero

luukvbaal:
**fix(ui): allow resize commands to set 'cmdheight' to 0**

Resolve https://github.com/neovim/neovim/issues/21558

**fix: issues with command line if ui elements are externalized**

Resolve https://github.com/neovim/neovim/issues/20888 and handle side effects for setting `'cmdheight'` to zero.

PR #27872, opened 2024-03-15
fix(ui): don't force 'cmdheight' to zero with ext_messages

luukvbaal:
Remove remaining code that prevents non-zero 'cmdheight' with ext_messages.
Followup to #27812, which was incomplete.

### DIFF
--- a/test/functional/ui/cmdline_spec.lua
+++ b/test/functional/ui/cmdline_spec.lua
@@ -1705,19 +1705,24 @@ describe('cmdheight=0', function()
     ]])
   end)
 
-  it('cannot be resized at all with external messages', function()
+  it('can be resized with external messages', function()
     clear()
     screen = new_screen({ rgb = true, ext_messages = true })
     command('set laststatus=2 mouse=a')
     command('resize -1')
     screen:expect([[
       ^                         |
+      {1:~                        }|*2
+      {3:[No Name]                }|
+                               |
+    ]])
+    api.nvim_input_mouse('left', 'press', '', 0, 3, 10)
+    poke_eventloop()
+    api.nvim_input_mouse('left', 'drag', '', 0, 4, 10)
+    screen:expect([[
+      ^                         |
       {1:~                        }|*3
       {3:[No Name]                }|
     ]])
-    api.nvim_input_mouse('left', 'press', '', 0, 6, 10)
-    poke_eventloop()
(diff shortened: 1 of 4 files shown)


luukvbaal:
Opening this as draft to discuss, can update failing tests if we do indeed want this.
Judging by previous recent PR's we do but TBH I forgot about previous context in #21580 and #21709 when I made #27812.

Opening a floating window in the 'cmdheight' area still seems like a valid ext_cmdline/messages use case to me. I don't think https://github.com/neovim/neovim/issues/21653#issuecomment-1375649070 is an issue either in that case (especially when the message grid is removed?).

test/functional/ui/cmdline_spec.lua
```diff
@@ -1713,20 +1713,4 @@ describe('cmdheight=0', function()
...
-    command('resize -1')
-    screen:expect([[
-      ^                         |
-      {1:~                        }|*3
-      {3:[No Name]                }|
-    ]])
-    api.nvim_input_mouse('left', 'press', '', 0, 6, 10)
-    poke_eventloop()
-    api.nvim_input_mouse('left', 'drag', '', 0, 5, 10)
-    screen:expect_unchanged()
```
bfredl, line 1730: better to update the test and its title.

luukvbaal: Sure, so do we follow through with allowing 'cmdheight' > 0 with ext_messages?

bfredl: yes, it should be allowed as it is required for The Plan, but it should still default to zero with an external ext_messages UI (otherwise it is just wasted space).

luukvbaal: I mean The Plan is still up in the air. My feeling now is that 100% mimicking the current message grid implementation to pass tests was a nice idea, but might not be worth the effort. So if we can agree on a different representation that does not use the 'cmdheight' area it might not be necessary for The Plan.

That doesn't change the fact that it can still be a valid use case, so I think we can follow through with this PR regardless.

> but it should still default to zero with an external ext_messages UI

Hmm yeah that makes sense since it's the cmdheight area is not in use unless the event handler uses it, so it should be responsible for making it non-zero. I just thought implicitly changing a (now valid) default option value was a little weird to begin with, but that is already documented. I can update the text.

PR #27874, opened 2024-03-15
fix(messages)!: vim.ui_attach message callbacks are unsafe

luukvbaal:
Problem:  Lua callbacks for "msg_show" events with vim.ui_attach() are
          executed when it is not safe.
Solution: Disallow non-fast API calls for "msg_show" event callbacks.
          Make sure fast APIs do not modify Nvim state.

### DIFF
--- a/src/nvim/lua/executor.c
+++ b/src/nvim/lua/executor.c
@@ -187,7 +187,7 @@ static void nlua_luv_error_event(void **argv)
   msg_ext_set_kind("lua_error");
   switch (type) {
   case kCallback:
-    semsg_multiline("Error executing luv callback:\n%s", error);
+    semsg_multiline("Error executing callback:\n%s", error);
     break;
   case kThread:
     semsg_multiline("Error in luv thread:\n%s", error);
@@ -201,13 +201,13 @@ static void nlua_luv_error_event(void **argv)
   xfree(error);
 }
 
-static int nlua_luv_cfpcall(lua_State *lstate, int nargs, int nresult, int flags)
+/// Execute callback in "fast" context. Used for luv and some vim.ui_event
+/// callbacks where using the API directly is not safe.
+static int nlua_fast_cfpcall(lua_State *lstate, int nargs, int nresult, int flags)
   FUNC_ATTR_NONNULL_ALL
 {
   int retval;
 
-  // luv callbacks might be executed at any os_breakcheck/line_breakcheck
-  // call, so using the API directly here is not safe.
   in_fast_callback++;
 
   int top = lua_gettop(lstate);
@@ -366,11 +366,13 @@ static int nlua_init_argv(lua_State *const L, char **argv, int argc, int lua_arg
 static void nlua_schedule_event(void **argv)
 {
   LuaRef cb = (LuaRef)(ptrdiff_t)argv[0];
+  uint32_t ns_id = (uint32_t)(ptrdiff_t)argv[1];
   lua_State *const lstate = global_lstate;
   nlua_pushref(lstate, cb);
   nlua_unref_global(lstate, cb);
   if (nlua_pcall(lstate, 0, 0)) {
     nlua_error(lstate, _("Error executing vim.schedule lua callback: %.*s"));
+    ui_remove_cb(ns_id, true);
   }
 }
 
@@ -392,8 +394,9 @@ static int nlua_schedule(lua_State *const lstate)
   }
 
   LuaRef cb = nlua_ref_global(lstate, 1);
-
-  multiqueue_put(main_loop.events, nlua_schedule_event, (void *)(ptrdiff_t)cb);
+  // Pass along UI event handler to disable on error.
+  multiqueue_put(main_loop.events, nlua_schedule_event, (void *)(ptrdiff_t)cb,
+                 (void *)(ptrdiff_t)ui_event_ns_id);
   return 0;
 }
 
@@ -425,7 +428,7 @@ static int nlua_wait(lua_State *lstate)
   FUNC_ATTR_NONNULL_ALL
 {
   if (in_fast_callback) {
-    return luaL_error(lstate, e_luv_api_disabled, "vim.wait");
+    return luaL_error(lstate, e_fast_api_disabled, "vim.wait");
   }
 
   intptr_t timeout = luaL_checkinteger(lstate, 1);
@@ -598,7 +601,7 @@ static void nlua_common_vim_init(lua_State *lstate, bool is_thread, bool is_stan
     luv_set_cthread(lstate, nlua_luv_thread_cfcpcall);
   } else {
     luv_set_loop(lstate, &main_loop.uv);
-    luv_set_callback(lstate, nlua_luv_cfpcall);
+    luv_set_callback(lstate, nlua_fast_cfpcall);
   }
   luaopen_luv(lstate);
   lua_pushvalue(lstate, -1);
@@ -724,7 +727,7 @@ static int nlua_ui_detach(lua_State *lstate)
     return luaL_error(lstate, "invalid ns_id");
   }
 
-  ui_remove_cb(ns_id);
+  ui_remove_cb(ns_id, false);
   return 0;
 }
 
@@ -1174,7 +1177,7 @@ int nlua_call(lua_State *lstate)
   size_t name_len;
   const char *name = luaL_checklstring(lstate, 1, &name_len);
   if (!nlua_is_deferred_safe() && !viml_func_is_fast(name)) {
-    return luaL_error(lstate, e_luv_api_disabled, "Vimscript function");
+    return luaL_error(lstate, e_fast_api_disabled, "Vimscript function");
   }
 
   int nargs = lua_gettop(lstate) - 1;
@@ -1231,7 +1234,7 @@ free_vim_args:
 static int nlua_rpcrequest(lua_State *lstate)
 {
   if (!nlua_is_deferred_safe()) {
-    return luaL_error(lstate, e_luv_api_disabled, "rpcrequest");
+    return luaL_error(lstate, e_fast_api_disabled, "rpcrequest");
   }
   return nlua_rpc(lstate, true);
 }
@@ -1594,6 +1597,12 @@ bool nlua_ref_is_function(LuaRef ref)
 Object nlua_call_ref(LuaRef ref, const char *name, Array args, LuaRetMode mode, Arena *arena,
                      Error *err)
 {
+  return nlua_call_ref_ctx(false, ref, name, args, mode, arena, err);
+}
+
+Object nlua_call_ref_ctx(bool fast, LuaRef ref, const char *name, Array args, LuaRetMode mode,
+                         Arena *arena, Error *err)
+{
   lua_State *const lstate = global_lstate;
   nlua_pushref(lstate, ref);
   int nargs = (int)args.size;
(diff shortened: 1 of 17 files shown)


zeertzjq:
Actually ext_messages callback may cause Nvim to crash in many cases, because code calling `*msg` functions don't expect it to change state.

luukvbaal:
Hmm, does it though? The messages are batched and (usually) flushed in normal_redraw() where changing state should be allowed?

But yeah if they are flushed where it isn't allowed I guess that should be prevented.

zeertzjq:
`msg_start()` calls `msg_ext_ui_flush()`

luukvbaal:
Okay, yeah msg_start seems kind of redundant in the situation where ext_messages is used in core and the message grid is removed.

src/nvim/eval/userfunc.c
```diff
@@ -2347,7 +2346,7 @@ void ex_function(exarg_T *eap)
   // Makes 'exe "func Test()\n...\nendfunc"' work.
   if (*p == '\n') {
     line_arg = p + 1;
-  } else if (*p != NUL && *p != '"' && !eap->skip && !did_emsg) {
+  } else if (*p != NUL && *p != '"' && !eap->skip && called_emsg > saved_called_emsg) {
```
zeertzjq, line 2349:
```suggestion
  } else if (*p != NUL && *p != '"' && !eap->skip && called_emsg == saved_called_emsg) {
```

zeertzjq:
> The handler may call api functions that reset it, leaving :function active after an error.

Is there at least an interactive example where this happens?

luukvbaal:
> > The handler may call api functions that reset it, leaving :function active after an error.
> 
> Is there at least an interactive example where this happens?

Still on mobile but if you checkout my ext PR and run `nvim --clean`, you should see the issue when trying to redefine an existing function.

zeertzjq:
I just found a strange example where using `called_emsg` will also not work:
```vim
func Fail()
  func Foo{'()'}bar()
  endfunc
endfunc

set formatexpr=Fail()
normal! gqq
```
This will cause a function with invalid name `Foo()bar` to be defined.

So may another solution is needed (e.g. explicit `goto`).

zeertzjq (dismissed):
I think this should be solved in API calls instead, because `did_emsg` and `called_emsg` are not necessarily equivalent. For example, `did_emsg` is not incremented with `:silent!`, while `called_emsg` is. If ext_messages callback makes the value of `did_emsg` unreliable then it may also cause other problems like duplicate error messages.

luukvbaal:
> Actually ext_messages callback may cause Nvim to crash in many cases, because code calling *msg functions don't expect it to change state.

> I think this should be solved in API calls instead

Should we try to avoid both issues at once by not flushing messages where it can lead to these problems?

I.e. instead of having msg_start/end() flush messages, insert some kind of special chunk to separate messages, or stash the events to emit them when it is safe. @bfredl

luukvbaal:
The current commit avoids the `:function` problem this PR originally tried to fix. Also fixes this crash for example:

```lua
nvim --clean -S test.lua
-- emit any message

-- test.lua
vim.ui_attach(vim.api.nvim_create_namespace(''), {ext_messages = true}, function(event, ...)
  if event == "msg_show" then
    vim.fn.getchar()
  end
end)
```

Would like to have some feedback before updating tests, which now receive multiple events per `screen:expect()`.

luukvbaal:
### Status:
Postponed to 0.11 cycle and considering to change the callback signature for "msg_show", or possibly add a new event if we want a deprecation cycle for "msg_show"? Not strictly necessary but "ext_messages" is still experimental. If we are throttling message events, it makes more sense to emit a single message event with an array of all the emitted messages rather than multiple events.

folke:
@luukvbaal makes sense!

I'll fix the Noice side when needed :)

bfredl:
Another approach for the unsafety we could take is something like the following. Instead of adding a layer of batching in the _c code_, what we will do is to mark some `ui_attach` callback as being fast-unsafe (similar as for low-level libluv calbacks). We will then need to add extra msg_ event which is is unrestricted, which marks the right spot for a core implementation to translate the collected data (now in the _lua code_, which can filter whatever data it needs to keep) into API calls to show windows as needed. 

The benefit is some preserved flexibility and consistency for other plugin consumers. some plugins might not need core API calls in the msg_show handler at all, but might show messages in some completely external way (by using libuv io, subprocesses etc), and for these there is no need for throttling.

The tricky part is when this extra notification needs to be done. one point is `msg_scroll_flush()`, but we also want to do it before `update_screen()` if there is both pending messages and other window changes, so all window redraws including message windows can then happen for the same `update_screen()` batch. 

And sorry for the delay in me responding to this kind of PR:s. Substantial enhancements to the `vim.ui_attach` mechanism will be one of my priorities now in the 0.11 cycle.

luukvbaal:
Thanks for your feedback @bfredl; that indeed sounds preferable to the route taken here. Feel free to close this PR when you intend to work on this particular issue or I may repurpose this PR to the proposed solution when I (hopefully) resume work on the ext_ui stuff.

(I'm still available for small fixes and stuff but I think I will need to commit to another stretch of full-time work to really try to drive the core ext UI home.)

luukvbaal:
@bfredl one issue I still see even with that approach is that some fast-safe api functions are perhaps not actually "safe" in this particular context.

I.e. the `:function` issue that originally prompted this PR is still reproducible with e.g. `nvim_get_runtime_file()` which is marked api-fast but also calls `try_end()` which modifies some global variables that messes with the `:function` logic. Granted that might be a rather niche issue but it at least indicates that it is not a complete solution. Although it could be that this isolates the problem to just that `did_emsg` issue. Presumably api-fast functions don't modify any/much other Nvim state.

luukvbaal:
https://github.com/neovim/neovim/blob/ba70404c558169b813f51d5a0bb71cd540555c93/src/nvim/api/private/helpers.c#L109-L112

It could be that interpreting this suggestion as applicable for api-fast functions resolves that issue. But IDK how we would go about enforcing that...

bfredl:
well, would need to fix that by getting rid of the TRY_WRAP in `get_nvim_runtime_file` then. I presume it might be some errors thrown by regexes we need to convert to error returns.

luukvbaal:
Yes that particular case is easy enough to resolve. I was moreso suggesting that we might look for a way to make sure things like this don't happen in other places/regresses i.e. some custom lint rule for api-fast functions.

luukvbaal:
> what we will do is to mark some ui_attach callback as being fast-unsafe (similar as for low-level libluv calbacks). We will then need to add extra msg_ event which is is unrestricted

Rebased and updated the PR to do this. Do we actually need a new event to notify message handlers when it is safe to display messages? The handler could just `vim.schedule()` inside the fast context right? Avoids the need to identify the appropriate place to send this notification which may be tricky as you said.

Ping @folke since I previously [informed](https://github.com/neovim/neovim/issues/28772#issuecomment-2119296186) you to the contrary. Is scheduling sufficient or do you think a new event would be helpful?

folke:
@luukvbaal yep, should be fine for noice. I've changed the implementation since and do more `nvim_redraw` flushes now which solves some of the issues I had.

I also retry failed attempts at rendering (new) windows during textlock and other conditions.

Regarding the redraw flushes, I honestly always thought that redraw was not possible during blocking events or when input was requested. It's only very recently that I learned that it's just the automatic flushing and internal redraws are disabled during those events. I didn't know I could still trigger manual redraw flushes. Kinda stupid I didn't know this before :)

src/nvim/lua/executor.c
```diff
@@ -219,8 +222,11 @@ static int nlua_luv_cfpcall(lua_State *lstate, int nargs, int nresult, int flags
     }
     const char *error = lua_tostring(lstate, -1);
 
-    multiqueue_put(main_loop.events, nlua_luv_error_event,
-                   xstrdup(error), (void *)(intptr_t)kCallback);
+    if (!event_error_queued) {
+      multiqueue_put(main_loop.events, nlua_luv_error_event,
+                     xstrdup(error), (void *)(intptr_t)kCallback);
+      event_error_queued = true;
```
bfredl, line 228: Coalescing multiple error callbacks into one might be good but then we need to preserve the individual `error` messages and display all of them at  once we are ready to handle the callback.

src/nvim/lua/executor.c
```diff
@@ -1605,9 +1611,10 @@ Object nlua_call_ref(LuaRef ref, const char *name, Array args, LuaRetMode mode,
     nlua_push_Object(lstate, &args.items[i], 0);
   }
 
-  if (nlua_pcall(lstate, nargs, 1)) {
-    // if err is passed, the caller will deal with the error.
-    if (err) {
+  bool fast = name && strcmp(name, "msg_show") == 0;
+  if (fast ? nlua_fast_cfpcall(lstate, nargs, 1, -1) : nlua_pcall(lstate, nargs, 1)) {
```
bfredl, line 1615: It probably will look better to structure this like `if (fast) { ... the fast stuff } else { the old code}`

src/nvim/lua/executor.c
```diff
@@ -1605,9 +1611,10 @@ Object nlua_call_ref(LuaRef ref, const char *name, Array args, LuaRetMode mode,
     nlua_push_Object(lstate, &args.items[i], 0);
   }
 
-  if (nlua_pcall(lstate, nargs, 1)) {
-    // if err is passed, the caller will deal with the error.
-    if (err) {
+  bool fast = name && strcmp(name, "msg_show") == 0;
```
bfredl, line 1614: To make this scalable we should add this as a flag to `api/ui_events.h` to mark the events which need this . I could implement that part myself on top of your work here.

src/nvim/lua/executor.c
luukvbaal: This is to avoid recursiveness. If one msg_show event handler itself _causes_ an error we need to avoid emitting that error multiple times, which presumably will be the same every time.

But yeah I suppose this solution might hide multiple different errors.

luukvbaal: This and possibly other recursion checks are no longer necessary because I added automatic detachment for when a vim.ui_attach callback produces excessive errors.

src/nvim/lua/executor.c
luukvbaal: I took a stab at this because I first implemented the automatic detachment per UI extension where further information from the generator was necessary in C. Later decided that wasn't really necessary. Hope what I ended up with isn't too far off what you had in mind.

bfredl (approved):
Nice work. We probably will need to adjust the details later, but this is a good start for error handling. (better than just deadlocking the instance:p)

folke:
I just updated Noice to work again, but one thing that I'm not sure how to fix is `msg_show` with `confirm` kind.
There's probably other related issues as well.

It seems that when a `confirm` is active, `scheduled` callbacks all wait to execute till the `confirm` is done.

This means, there's no way for noice to display the confirm message the user needs to handle.

Any ideas?

luukvbaal:
Right, so maybe we do need a new event to indicate pending messages may be displayed (https://github.com/neovim/neovim/pull/27874#issuecomment-2167705853).

Alternatively we could exclude some `msg_show` events, that we know to be safe (or must make safe because the event loop is delayed waiting for input) from the fast context. Although that would kind of defeat the purpose of blanket marking all msg_show events as fast in the first place.

PR #27875, opened 2024-03-15
docs: update fedora 39 build prerequisite

dominicracine:
Will trying to install Neovim on Fedora 39 while following the guidance I got this error:

```
"nvim.pot" "nvim.pot" 9358L, 181124B written[386/415] Generating cs.cp1250.moines

FAILED: src/nvim/po/cs.cp1250.mo /neovim/build/src/nvim/po/cs.cp1250.mo
cd /neovim/build/src/nvim/po && /usr/bin/msgfmt -o /neovim/build/src/nvim/po/cs.cp1250.mo /neovim/src/nvim/po/cs.cp1250.po
/neovim/src/nvim/po/cs.cp1250.po: warning: Charset "CP1250" is not supported. msgfmt relies on iconv(),
                                           and iconv() does not support "CP1250".
                                           Installing GNU libiconv and then reinstalling GNU gettext
                                           would fix this problem.
                                           Continuing anyway.
/usr/bin/msgfmt: Cannot convert from "CP1250" to "UTF-8". msgfmt relies on iconv(), and iconv() does not support this conversion.
[387/415] Generating cs.mo
FAILED: src/nvim/po/cs.mo /neovim/build/src/nvim/po/cs.mo
cd /neovim/build/src/nvim/po && /usr/bin/msgfmt -o /neovim/build/src/nvim/po/cs.mo /neovim/src/nvim/po/cs.po
/neovim/src/nvim/po/cs.po: warning: Charset "ISO-8859-2" is not supported. msgfmt relies on iconv(),
                                    and iconv() does not support "ISO-8859-2".
                                    Installing GNU libiconv and then reinstalling GNU gettext
                                    would fix this problem.
                                    Continuing anyway.
/usr/bin/msgfmt: Cannot convert from "ISO-8859-2" to "UTF-8". msgfmt relies on iconv(), and iconv() does not support this conversion.
```

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -293,10 +293,10 @@ Platform-specific requirements are listed below.
 sudo apt-get install ninja-build gettext cmake unzip curl build-essential
 ```
 
-### CentOS / RHEL / Fedora
+### RHEL / Fedora
 
 ```
-sudo dnf -y install ninja-build cmake gcc make unzip gettext curl
+sudo dnf -y install ninja-build cmake gcc make unzip gettext curl glibc-gconv-extra
 ```
 
 ### openSUSE


BUILD.md
```diff
@@ -299,6 +299,13 @@ sudo apt-get install ninja-build gettext cmake unzip curl build-essential
 sudo dnf -y install ninja-build cmake gcc make unzip gettext curl
 ```
 
+### Fedora 39
+
+```
+sudo dnf -y install ninja-build cmake gcc make unzip gettext curl && \
+sudo dnf groupinstall -y "Development Libraries"
+```
+
```
spacey-sooty, line 308: Right above this are Fedora requirements why make them seperate instead of updating existing ones?

dundargoc:
This solution is incorrect. Newer fedoras only need to install one more package: `glibc-gconv-extra`. `sudo dnf groupinstall -y "Development Libraries"` installs 180 packages, one of them which happens to be `glibc-gconv-extra`.

dundargoc:
Also this problem isn't exclusive to fedora 39, all fedoras from 35 and up has this problem, so the documentation should reflect that.

The background here is that glibc-gconv-extra was bundled in glibc in older fedoras but was separated for...reasons.

DominicRacine:
The current requirements in the doc worked fine with Fedora 38 on my side

DominicRacine:
I didn't add the CentOS and RHEL because I didn't test on those OS

DominicRacine:
I added it to the installation requirements and it worked!  `sudo dnf -y install ninja-build cmake gcc make unzip gettext curl glibc-gconv-extra`.

Should it also be added for CentOS and RHEL?

dundargoc:
Rethinking this I think trying to provide build instructions for all versions of all distros is a fools errand. I'd just add `glibc-gconv-extra` to the `### CentOS / RHEL / Fedora` section instead of creating one for "Fedora 39" only.

Also centos is ded. You can replace `### CentOS / RHEL / Fedora` to `### RHEL / Fedora`.

DominicRacine:
I updated the MR, let me know what you think

clason (approved):

PR #27884, opened 2024-03-16
feat: add TSLineHighlighter and attach it to the cmdwin

Sam-programs:
Works but i am not happy with the current implementation.

### DIFF
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -46,6 +46,7 @@
 #include "nvim/highlight_defs.h"
 #include "nvim/highlight_group.h"
 #include "nvim/keycodes.h"
+#include "nvim/lua/executor.h"
 #include "nvim/macros_defs.h"
 #include "nvim/map_defs.h"
 #include "nvim/mapping.h"
@@ -1193,7 +1194,7 @@ static int command_line_execute(VimState *state, int key)
       pum_want.active = false;
     }
 
-    if (!cmdline_was_last_drawn) {
+    if (!cmdline_was_last_drawn && !ui_has(kUICmdline)) {
       redrawcmdline();
     }
     return 1;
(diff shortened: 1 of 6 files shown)


clason:
Note: @luukvbaal is already working on this issue (with a different, better approach). Let's not duplicate efforts.

Sam-programs:
I thought luke was working on just the cmdline and messages, not the cmdwin.

clason:
I think that's part and parcel of the whole thing; we should avoid having different approaches for the two if not absolutely necessary. (If it is, we can always reopen this.)

luukvbaal:
To be honest, I hadn't given the cmdwin much thought until now.
There is some overlap in applying treesitter highlighting, but as the cmdline window doesn't use the message grid, it is kind of orthogonal to replacing the current cmdline/message area.
I don't think my work necessarily conflicts with this, or needs to re-implement the cmdline window in Lua.

clason:
Reopening this then. I'd like some discussion about the necessity/general usefulness of a line highlighter (and the significant refactor introduced here) first, though. The cmdwin is relatively niche, and _if_ we want to add syntax highlighting to it, I'd prefer it to be simpler and self-contained (e.g., using the string parser plus manual extmark application), as performance is not a huge concern here. Ideally, we can leverage some of the cmdline highlight infrastructure, so I definitely want to park this PR until the former is done.

@neovim/treesitter

src/nvim/ex_getln.c
```diff
@@ -4444,6 +4448,15 @@ static int open_cmdwin(void)
...
   int save_count = save_batch_count();
 
+  MAXSIZE_TEMP_DICT(opts, 1);
+  PUT_C(opts, "line_based", BOOLEAN_OBJ(true));
+  MAXSIZE_TEMP_ARRAY(args, 3);
+  ADD_C(args, INTEGER_OBJ(cmdwin_buf->handle));
+  ADD_C(args, CSTR_TO_OBJ("vim"));
+  ADD_C(args, DICTIONARY_OBJ(opts));
+  Error err = ERROR_INIT;
+  NLUA_EXEC_STATIC("vim.treesitter.start(...)", args, kRetNilBool, NULL, &err);
```
seandewar, line 4458: This probably should only apply to specific `histtypes`, like how the old logic set the cmdwin's `&filetype` before?

For example, you probably don't want a cmdwin opened for `input()` via CTRL-F to highlight everything as Vim script, as the input could be for any purpose.

justinmk:
Using treesitter for cmdwin is in scope, but not sure it's worth a lot of extra code. Instead I would start by enabling the `vim` (vimscript) parser in it.

PR #23002, opened 2023-04-10
test(lsp_spec): fix unstable tests for set_defaults

mliszcz:
In the `test_rpc_server` procedure, both `on_setup` and `on_init` callbacks can run concurrently in some scenarios. This caused some CI failures in tests for the LSP set_defaults feature.

This commit attempts to fix this by merging those two callbacks in the impacted tests.

See: https://github.com/neovim/neovim/actions/runs/4553550710/attempts/1

PR #22810, opened 2023-03-29
refactor(lsp): do not parse verbose output when overwriting options

mliszcz:
Now that all prerequisite are in place, we can finally stop parsing verbose set? output.

PR #27899, opened 2024-03-17
test(lsp): fix flaky basic_finish test

zeertzjq:
Problem:
As mentioned in #23002 on_setup and on_init are run concurrently.
However, in basic_finish tests on_setup must attach the client before
on_init finishes.  The other basic_finish test isn't flaky because it
makes an RPC request in on_init.

Solution:
Don't use on_setup in basic_finish tests.

### DIFF
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -282,7 +282,8 @@ describe('LSP', function()
       local client
       test_rpc_server {
         test_name = 'basic_finish',
-        on_setup = function()
+        on_init = function(_client)
+          client = _client
           exec_lua [[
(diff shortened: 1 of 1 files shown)


zeertzjq:
```
ERROR LSP: test/functional/fixtures/fake-lsp-server.lua:25: assert_eq failed: left == "\"shutdown\"", right == "\"textDocument/didOpen\""
```

zeertzjq:
The situation seems to be the same as the one mentioned in https://github.com/neovim/neovim/pull/22810#issuecomment-1489877926

> This looks like the LSP client got attached between calls in `on_setup`, but before `on_init` was called.

However in this case the client must be attached **before** `on_init` is called.

PR #27911, opened 2024-03-18
build: use `GIT_REPOSITORY` for local URLs

dundargoc:
`GIT_REPOSITORY` will cause cmake to rebuild if local dependency
changes, which isn't the case for `URL`.

Also document how to test a different commits of a dependency.


clason:
I think this needs a bit more instruction:
1. how to find out the SHA (`make cmake`, copy-paste the `actual` from the error message)
2. SHA is ignored for local resources
3. if you point the URL to a local checkout, you need to `make distclean` (or remove the correct `stamp` file -- that's the most non-trivial information to document here) before building, or any changes in the local directory will not be picked up because the build is cached.

CONTRIBUTING.md
```diff
@@ -318,6 +318,25 @@ types, etc. See [:help dev-lua-doc][dev-lua-doc].
       - Private functions usually should be underscore-prefixed (named "_foo", not "foo").
 - Mark deprecated functions with `@deprecated`.
 
+Third-party dependencies
+------------------------
+
+Sometimes problems occur due to dependencies of neovim. To build neovim using a
+different commit of a dependency change the appropriate URL in
+`cmake.deps/dept.txt`. For example, to use a different version of luajit
```
zeertzjq, line 326:
```suggestion
`cmake.deps/deps.txt`. For example, to use a different version of luajit
```

dundargoc:
Dependent on https://github.com/neovim/neovim/pull/27912

CONTRIBUTING.md
```diff
@@ -318,6 +318,24 @@ types, etc. See [:help dev-lua-doc][dev-lua-doc].
       - Private functions usually should be underscore-prefixed (named "_foo", not "foo").
 - Mark deprecated functions with `@deprecated`.
 
+Third-party dependencies
+------------------------
+
+To build neovim using a different commit of a dependency change the appropriate
```
justinmk, line 324:
```suggestion
To build Nvim using a different commit of a dependency change the appropriate
```

justinmk:
Will this also explain IGNORE_SHA?

dundargoc:
Nah I will explain it, I just need to merge that PR first

dundargoc:
> 3. if you point the URL to a local checkout, you need to `make distclean` (or remove the correct `stamp` file -- that's the most non-trivial information to document here) before building, or any changes in the local directory will not be picked up because the build is cached.

This is technically not needed if you replace `URL` with `GIT_REPOSITORY` for local dependencies as it will only rebuild if needed. We could check if `URL` points to a local directory and if so swap `URL` with `GIT_REPOSITORY` instead of documenting it. I'm kinda on the fence on this. Suggestions?

justinmk:
> This is technically not needed if you replace `URL` with `GIT_REPOSITORY` for local dependencies

What are the cases where it _wouldn't_ be a git repo? Would it simplify docs/code if we required that?

justinmk:
> Local checkout. _Very_ useful. Removing that is not acceptable.

Local checkout of a tarball? I was assuming `GIT_REPOSITORY` is for local dirs containing `.git/` .

clason:
> > 3. if you point the URL to a local checkout, you need to `make distclean` (or remove the correct `stamp` file -- that's the most non-trivial information to document here) before building, or any changes in the local directory will not be picked up because the build is cached.
> 
> This is technically not needed if you replace `URL` with `GIT_REPOSITORY` for local dependencies as it will only rebuild if needed. We could check if `URL` points to a local directory and if so swap `URL` with `GIT_REPOSITORY` instead of documenting it. I'm kinda on the fence on this. Suggestions?

Just document that "if you want to use a local checkout, replace `URL <url>` with `GIT_REPOSITORY <path>`".

dundargoc:
> What are the cases where it _wouldn't_ be a git repo? Would it simplify docs/code if we required that?

A tarball. Nah, unfortunately not if we want to use tarballs.

> Local checkout of a tarball? I was assuming GIT_REPOSITORY is for local dirs containing a .git/ subdir.

No, not at all. `URL` works for remote tarball but not remote git repo. `GIT_REPOSITORY` works for remote git repo, but not remote tarball. Both `URL` and `GIT_REPOSITORY` works for local repo, just that `GIT_REPOSITORY` always works correctly (aka that it's rebuilt when needed).

dundargoc:
> Just document that "if you want to use a local checkout, replace `URL <url>` with `GIT_REPOSITORY <path>`".

I want to test if we can accomplish this automatically. If it turns out that it doesn't work well I will go with this, thanks for the suggestion.

PR #27912, opened 2024-03-18
build: introduce variable DEPS_IGNORE_SHA for skipping dependency hash check

dundargoc:
This will reduce friction as developers no longer need to provide a hash
when testing out different commits.

To skip the hash check, set `DEPS_IGNORE_SHA` to `TRUE` in
`cmake.deps/CMakeLists.txt`.


cmake.deps/CMakeLists.txt
```diff
@@ -17,21 +17,13 @@ include(Deps)
...
 
-set(DEFAULT_MAKE_CFLAGS CFLAGS+=-g)
-
-check_c_compiler_flag(-Og HAS_OG_FLAG)
-if(HAS_OG_FLAG)
-  set(DEFAULT_MAKE_CFLAGS CFLAGS+=-Og ${DEFAULT_MAKE_CFLAGS})
-endif()
-
-set(DEPS_INCLUDE_FLAGS "-I${DEPS_INSTALL_DIR}/include -I${DEPS_INSTALL_DIR}/include/luajit-2.1")
+set(IGNORE_SHA FALSE)
```
justinmk, line 24: seems like a `DEPS_` prefix makes sense. We already have `DEPS_BUILD_DIR`, etc.
```suggestion
set(DEPS_IGNORE_SHA FALSE)
```

PR #27868, opened 2024-03-15
docs: fix typos

dundargoc:

PR #27925, opened 2024-03-19
docs: clarify what `nvim_buf_set_name()` does

echasnovski:
The main motivation for this update is to provide context as to how users can watch if the buffer changed name. Currently the most reliable approach is to use `BufFilePre` and `BufFilePost` events. Their triggered after `nvim_buf_set_name()` seems to be intended.

Alternative approach: #27804

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -2447,7 +2447,7 @@ nvim_buf_set_mark({buffer}, {name}, {line}, {col}, {opts})
       • |nvim_buf_get_mark()|
 
 nvim_buf_set_name({buffer}, {name})                      *nvim_buf_set_name()*
-    Sets the full file name for a buffer
+    Sets the full file name for a buffer, like |:file_f|
 
     Parameters: ~
       • {buffer}  Buffer handle, or 0 for current buffer
--- a/src/nvim/api/buffer.c
+++ b/src/nvim/api/buffer.c
@@ -968,7 +968,7 @@ String nvim_buf_get_name(Buffer buffer, Error *err)
   return cstr_as_string(buf->b_ffname);
 }
 
-/// Sets the full file name for a buffer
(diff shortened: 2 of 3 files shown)


runtime/doc/api.txt
```diff
@@ -2449,6 +2449,10 @@ nvim_buf_set_mark({buffer}, {name}, {line}, {col}, {opts})
 nvim_buf_set_name({buffer}, {name})                      *nvim_buf_set_name()*
     Sets the full file name for a buffer
 
+    Works similar to |:file_f|. If the buffer did have a name, that name
```
zeertzjq, line 2452:
```suggestion
    Works similarly to |:file_f|. If the buffer did have a name, that name
```

src/nvim/api/buffer.c
```diff
@@ -970,6 +970,10 @@ String nvim_buf_get_name(Buffer buffer, Error *err)
 
 /// Sets the full file name for a buffer
 ///
+/// Works similar to |:file_f|. If the buffer did have a name,
```
zeertzjq, line 973:
```suggestion
/// Works similarly to |:file_f|. If the buffer did have a name,
```

runtime/doc/api.txt
```diff
@@ -2449,6 +2449,10 @@ nvim_buf_set_mark({buffer}, {name}, {line}, {col}, {opts})
...
+    becomes the |alternate-file| name. An unlisted buffer is created to hold
+    the old name.
```
justinmk, line 2454: Do not repeat docs.

```suggestion
    Sets the full file name for a buffer, like |:file_f|.
```

justinmk (approved):

dundargoc:
Thank you for your contribution! We collect all typo fixes and similar small fixes into a single pull request and merge it once it gets big enough: https://github.com/neovim/neovim/pull/27868

PR #27930, opened 2024-03-19
fix(api): nvim_create_buf fixes

seandewar:
Problem: nvim_create_buf may leak memory or fail an assertion.

Solution: fix those thingies.


seandewar:
Wait a second, this assertion below maybe looks wrong too: https://github.com/neovim/neovim/blob/16a416cb3c17ed3a7f21d35da5d211fcad947768/src/nvim/api/vim.c#L993

Autocommands could surely name and set `&swapfile` for the new buffer to fail this assert?

justinmk (approved):

PR #27947, opened 2024-03-20
fix(cmdline): wrong 'incsearch' highlighting after :redraw

### LINKED ISSUE #17810: Wrong incsearch highlighting when calling nvim_buf_set_lines from timer in cmdline with conceallevel
### Neovim version (nvim -v)

v0.7.0-dev+1292-ge9b53f3fb

No

### Operating system/version

WSL Ubuntu

### Terminal name/version

wsltty

xterm-256color

### Installation

Build from source

```vim
" init.vim
autocmd CmdlineChanged * call timer_start(0, {-> Foo()})

let s:buf = nvim_create_buf(v:false, v:true)

function! Foo()
  call nvim_buf_set_lines(s:buf, 0, -1, v:true, [])
  redraw
endfunction
```
1. `nvim -u init.vim`
2. `:h<CR>` to open help file
3. `/with<CR>` search for text
4. `/Bar<CR>` search for another text

### Expected behavior

`incsearch` highlight works as expected.

### Actual behavior

When searching for `with`, the incsearch highlighting is missing.
When searching for `Bar`, incsearch highlights `with` instead of `Bar`.

This is similar to #14064 but with the additional step of `nvim_buf_set_lines()`.

Bisected to 5ab122917474b3f9e88be4ee88bc6d627980cfe0: `feat: add support for global statusline `

luukvbaal:
Problem:  Calling :redraw from a timer callback clears 'incsearch' highlighting.
Solution: Re-apply 'incsearch' highlighting if the screen was updated.

Fix #17810

### DIFF
--- a/test/functional/ui/searchhl_spec.lua
+++ b/test/functional/ui/searchhl_spec.lua
@@ -674,4 +674,18 @@ describe('search highlighting', function()
       :%g@a/b^                                 |
     ]])
   end)
+
+  it('incsearch is still visible after :redraw from K_EVENT', function()
+    fn.setline(1, { 'foo', 'bar' })
+    feed('/foo<CR>/bar')
+    screen:expect([[
+      foo                                     |
+      {3:bar}                                     |
+      {1:~                                       }|*4
+      /bar^                                    |
+    ]])
+    command('redraw!')
+    -- There is an intermediate state where :redraw! removes 'incsearch' highlight.
+    screen:expect_unchanged(true)
+  end)
 end)
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -1167,6 +1167,7 @@ static int command_line_execute(VimState *state, int key)
     return -1;  // get another key
   }
 
+  disptick_T display_tick_saved = display_tick;
   CommandLineState *s = (CommandLineState *)state;
   s->c = key;
 
@@ -1177,6 +1178,10 @@ static int command_line_execute(VimState *state, int key)
       do_cmdline(NULL, getcmdkeycmd, NULL, DOCMD_NOWAIT);
     } else {
       map_execute_lua(false);
+    }
+    // Re-apply 'incsearch' highlighting in case it was cleared.
+    if (display_tick > display_tick_saved && s->is_state.did_incsearch) {
+      may_do_incsearch_highlighting(s->firstc, s->count, &s->is_state);
     }
 
     // nvim_select_popupmenu_item() can be called from the handling of


luukvbaal:
(This also applies to ext_cmdline callbacks. Although I think Ideally the C code should call update_screen() after vim.ui_attach() ext_cmdline events, so that explicitly calling :redraw isn't necessary.)

zeertzjq:
> CmdlineChanged callback

I think you should say "timer callback" here.

test/functional/ui/searchhl_spec.lua
```diff
@@ -674,4 +674,20 @@ describe('search highlighting', function()
...
+    feed('/foo<CR>/bar')
+    screen:expect {
+      grid = [[
+      foo                                     |
+      {3:bar}                                     |
+      {1:~                                       }|*4
+      /bar^                                    |
+    ]],
+    }
+  end)
```
zeertzjq, line 692: This test gives warnings:
```
warning: Screen changes were received after the expected state. This indicates
indeterminism in the test. Try adding screen:expect(...) (or poke_eventloop())
between asynchronous (feed(), nvim_input()) and synchronous API calls.
  - Use screen:redraw_debug() to investigate; it may find relevant intermediate
    states that should be added to the test to make it more robust.
  - If the purpose of the test is to assert state after some user input sent
    with feed(), adding screen:expect() before the feed() will help to ensure
    the input is sent when Nvim is in a predictable state. This is preferable
    to poke_eventloop(), for being closer to real user interaction.
  - poke_eventloop() can trigger redraws and thus generate more indeterminism.
    Try removing poke_eventloop().
      
stack traceback:
	test/functional/ui/screen.lua:788: in function '_wait'
	test/functional/ui/screen.lua:470: in function 'expect'
	test/functional/ui/searchhl_spec.lua:684: in function <test/functional/ui/searchhl_spec.lua:678>
```
Maybe just use an RPC event instead?
```suggestion

  it('incsearch is still visible after :redraw from K_EVENT', function()
    fn.setline(1, { 'foo', 'bar' })
    feed('/foo<CR>/bar')
    screen:expect([[
      foo                                     |
      {3:bar}                                     |
      {1:~                                       }|*4
      /bar^                                    |
    ]])
    command('redraw!')
    -- There is an intermediate state where :redraw! removes 'incsearch' highlight.
    screen:expect_unchanged(true)
  end)
```

luukvbaal: Yeah I was still looking into it, thanks!

zeertzjq (approved):

PR #27963, opened 2024-03-21
fix(messages)!: decouple ext_messages from message grid

### LINKED ISSUE #20416: vim.ui_attach: detect if a message was sent because of a redraw (with `ext_messages`)
### Feature already in Vim?

_No response_

### Feature description

Whenever a `redraw` happens anywhere, the non-cleared messages will be resent.

This is problematic, if the code that handles messages, also does a redraw. Redraws in my code only happen when any queued messages have not been rendered yet and we're in some sort of blocking call.

This would effectively cause some sort of loop.

To circumvent this, I currently [wrap any redraw function](https://github.com/folke/noice.nvim/blob/dee23c21c3ff52119d8452b77814e6be75dd23bb/lua/noice/hacks.lua#L55) and when we're inside a redraw cycle, we simply ignore incoming messages.

Not sure what the best way would be to solve this for `ext_messages`.

luukvbaal:
Problem:  ext_messages is implemented to mimic the message grid
          implementation w.r.t. scrolling messages, clearing scrolled
          messages, hit-enter-prompts and replacing a previous message.
          Meanwhile, an ext_messages UI may not be implemented in a way
          where these events are wanted. Moreover, correctness of these
          events even assuming a "scrolled message" implementation
          depends on fragile "currently visible messages" global state,
          which already isn't correct after a previous message was
          supposed to have been overwritten (because that should not only
          happen when `msg_scroll == false`).

Solution: 
- No longer attempt to keep track of the currently visible
            messages: remove the `msg_ext(_history)_visible` variables.
            UIs may remove messages pre-emptively (timer based), or never
            show messages that don't fit a certain area in the first place.
- No longer emit the `msg(_history)_clear` events to clear
            "scrolled" messages. This opens up the `msg_clear` event to
            be emitted when messages should actually be cleared (e.g.
            when the screen is cleared). May also be useful to emit before
            the first message in an event loop cycle as a hint to the UI
            that it is a new batch of messages (vim._extui currently
            schedules an event to determine that).
- Set `replace_last` explicitly at the few callsites that want
            this to be set to true to replace an incomplete status message.
- Don't store a "keep" message to be re-emitted.

Resolve #20416

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -67,7 +67,10 @@ EDITOR
 
 EVENTS
 
-• todo
+• |ui-messages| no longer emits the `msg_show.return_prompt`, `msg_clear` and
+  `msg_history_clear` events. These events arbitrarily assume a message UI
+  mimicking the legacy message grid. Benefit: reduced UI event traffic and
+  more flexibility for UIs.
 
 HIGHLIGHTS
 
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -836,7 +836,6 @@ must handle.
 		"lua_error"	Error in |:lua| code
 		"lua_print"	|print()| from |:lua| code
 		"rpc_error"	Error response from |rpcrequest()|
-		"return_prompt"	|press-enter| prompt after a multiple messages
 		"quickfix"	Quickfix navigation message
 		"search_cmd"	Entered search command
 		"search_count"	Search count message ("S" flag of 'shortmess')
@@ -893,8 +892,5 @@ must handle.
 ["msg_history_show", entries] ~
 	Sent when |:messages| command is invoked. History is sent as a list of
 	entries, where each entry is a `[kind, content, append]` tuple.
-
-["msg_history_clear"] ~
-	Clear the |:messages| history.
 
  vim:tw=78:ts=8:noet:ft=help:norl:
--- a/src/nvim/api/ui_events.in.h
+++ b/src/nvim/api/ui_events.in.h
@@ -174,8 +174,6 @@ void msg_ruler(Array content)
   FUNC_API_SINCE(6) FUNC_API_REMOTE_ONLY;
 void msg_history_show(Array entries)
   FUNC_API_SINCE(6) FUNC_API_REMOTE_ONLY;
-void msg_history_clear(void)
-  FUNC_API_SINCE(10) FUNC_API_REMOTE_ONLY;
 
 // This UI event is currently undocumented.
 // - When the server needs to intentionally exit with an exit code, and there is no
--- a/src/nvim/bufwrite.c
+++ b/src/nvim/bufwrite.c
@@ -1763,7 +1763,6 @@ restore_backup:
     if (msg_add_fileformat(fileformat)) {
       insert_space = true;
     }
-    msg_ext_set_kind("bufwrite");
     msg_add_lines(insert_space, lnum, nchars);       // add line/char count
     if (!shortmess(SHM_WRITE)) {
       if (append) {
@@ -1773,6 +1772,8 @@ restore_backup:
       }
     }
 
+    msg_ext_set_kind("bufwrite");
+    msg_ext_overwrite = true;
     set_keep_msg(msg_trunc(IObuff, false, 0), 0);
   }
 
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -552,7 +552,6 @@ int update_screen(void)
   }
 
   win_ui_flush(true);
-  msg_ext_check_clear();
 
   // reset cmdline_row now (may have been changed temporarily)
   compute_cmdrow();
@@ -964,10 +963,6 @@ bool skip_showmode(void)
 int showmode(void)
 {
   int length = 0;
-
-  if (ui_has(kUIMessages) && clear_cmdline) {
-    msg_ext_clear(true);
-  }
 
   // Don't make non-flushed message part of the showmode.
   msg_ext_ui_flush();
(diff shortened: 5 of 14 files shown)


luukvbaal:
Low effort PR mainly intended to discuss the posed problem-solution statement.

justinmk:
Looks nice!

> Not emitting the hit-return message also seems fair game. The handler can decide and trigger a hit-enter prompt itself (e.g. based on the message kind or text height of the message).

Also sounds nice.

What are the downsides?

luukvbaal:
> What are the downsides?

From my POV there aren't any, besides changing behavior which clients may be relying on (not sure who else to ping to confirm that). I would say a client is in a better position to decide whether clearing/prompting is required than to take these suggestions from Nvim.

justinmk:
we can merge it early in 0.11 development cycle and find out

JeanMertz:
> we can merge it early in 0.11 development cycle and find out

Given that the 0.11 dev cycle is underway, should this be merged?

clason:
Needs a rebase. What is the plan here (and for message grid rework generally)? Are we confident about this being ready for 0.11, or do we want to bump to 0.12?

folke:
I'm personally a big fan of this PR.

Will reduce `ui_attach` events a lot. In Noice I currently need to do all sorts of work-arounds and duplicate event detection to make sure the ui doesn't get spammed.

luukvbaal:
> duplicate event detection to make sure the ui doesn't get spammed.

I still haven't actually found a clear reproducer where `keep_msg` is responsible for this, do you have one? As mentioned in https://github.com/neovim/neovim/issues/20416#issuecomment-2484076484, we may still want to remove `keep_msg`. But I think the scope of this PR is too big, and I've since found that `replace_last` actually is useful sometimes (I think noice also uses it?).

Can still go through with at least disabling `keep_msg` and the `hit-enter-prompt` under the given reasoning. I'm no longer sure the breaking event changes here are worth it.

luukvbaal:
> I just did some testing and it actually seems most of those issues with re-submitting events seems to be solved already.

Yeah recursive/redundant events were already reduced in various other PRs. The `keep_msg` itself is quite rare(hence my struggle to reproduce it😅).

> However, as soon as I have the cmdline open, I get spammed with cmdline_show events.

Does this patch improve that?
```diff
diff --git a/src/nvim/ex_getln.c b/src/nvim/ex_getln.c
index 1eecee2a38..bc99a18086 100644
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -3527,12 +3527,13 @@ void cmdline_ui_flush(void)
   CmdlineInfo *line = &ccline;
   while (level > 0 && line) {
     if (line->level == level) {
-      if (line->redraw_state == kCmdRedrawAll) {
+      CmdRedraw redraw_state = line->redraw_state;
+      line->redraw_state = kCmdRedrawNone;
+      if (redraw_state == kCmdRedrawAll) {
         ui_ext_cmdline_show(line);
-      } else if (line->redraw_state == kCmdRedrawPos) {
+      } else if (redraw_state == kCmdRedrawPos) {
         ui_call_cmdline_pos(line->cmdpos, line->level);
       }
-      line->redraw_state = kCmdRedrawNone;
       level--;
     }
     line = line->prev_ccline;
```

folke:
> Yeah recursive/redundant events were already reduced in various other PRs.

Should've checked that indeed :)

> Does this patch improve that?

Nope, still the same. `cmdline_show` events.

luukvbaal:
Depending on how you redraw, #32237 might help @folke. This completely eliminates redundant cmdline events for me, just a complete redraw (`<C-L>, :mode) still re-emits the cmdline state.

folke:
@luukvbaal yes, that fixes it indeed!

luukvbaal:
After looking into this again to resolve #34035, I found that "currently visible messages", which is used to determine when to emit these events, can already get out-of-sync with what it would be on the message grid. The fact that it doesn't work properly in it's current form increased my confidence in making a breaking change here (are there any UIs that tried to implement this exactly as the message grid or do they all just implement `msg_set_pos` events?), so I've rebased and reworked this.

justinmk (approved):
Let's try it! 🚀  The simplification is worth a minor breakage.

Rebasing should fix the broken windows CI.

runtime/doc/news.txt
```diff
@@ -67,7 +67,8 @@ EDITOR
 
 EVENTS
 
-• todo
+• |ui-messages| no longer emits the `msg_show.return_prompt`, `msg_clear` and
+  `msg_history_clear` events.
```
justinmk:
```suggestion
  `msg_history_clear` events. Performance benefit: reduces UI events traffic.
```

justinmk (approved):

runtime/doc/news.txt
```diff
@@ -67,7 +67,10 @@ EDITOR
 
 EVENTS
 
-• todo
+• |ui-messages| no longer emits the `msg_show.return_prompt`, `msg_clear` and
+  `msg_history_clear` events. These events arbitrarily assume a message UI
```
justinmk, line 71: What guidance can we give to UIs to adapt to this change? ref https://github.com/vscode-neovim/vscode-neovim/issues/2507

luukvbaal: The guidance is for UIs to determine themselves when to clear messages🫣 @xiyaowong already seems to have the right idea.

If a UI opens a new window to show messages, it makes sense to just clear messages after a few seconds. If a UI shows messages in a persistent area (cmdline), it makes sense to "clear" messages at the start of a new batch of messages. extui uses the next eventloop cycle to determine this (vim.schedule), like the message grid implementation does.

kaixuan188:
Using neovim-vscode, I wasted a ton of tokens because of this bug. I finally got Claude to help me find this PR... crash...

PR #27972, opened 2024-03-21
vim-patch:9.1.0196: filetype: support for gnuplot files is lacking

clason:
Problem:  filetype: support for gnuplot files is lacking
Solution: Also detect *.gnuplot files
          (RobbiZ98)

closes: vim/vim#14243

https://github.com/vim/vim/commit/3a6bd0c5c743bf69d2e8af4c8b3c6b2cb5f3631a

Co-authored-by: RobbiZ98 <113035863+RobbiZ98@users.noreply.github.com>


runtime/lua/vim/filetype.lua
```diff
@@ -466,6 +466,7 @@ local extension = {
   glsl = 'glsl',
   gn = 'gn',
   gni = 'gn',
+  gnuplot = 'gnuplot',
```
zeertzjq, line 469: Also need to remove line 1308

clason: No.

zeertzjq: Why? It's duplicate.

clason: No it's not. Test fails if I remove it.

zeertzjq: Oh, that seems to be a much bigger problem.

clason: No, just a standard issue that `extension` doesn't apply to `.extension` files?

zeertzjq: This means Nvim's behavior is different from Vim for many `.extension` files.

clason: Where this matters, we have added the `.extension` files separately. That's the whole point of the filetype tests.

zeertzjq: > Where this matters

Where does this not matter? It should matter for any extension.

clason: In practice. Nobody (sane) is working with literal `.tex` files.

zeertzjq: Yes, but that still matters, doesn't it? I think this needs more discussion in #27976.

PR #16955, opened 2022-01-06
feat(filetype.lua): fix .env file not detected

gsx64:
File type with `.env` extension such as `foo.env` is detected as normal, however `.env` file itself is not detected.
After adding `.env` entry inside filename table, `.env` file is detected as normal.

PR #16978, opened 2022-01-07
Filetype updates

gpanders:
Aggregation of filetype.lua updates

PR #27976, opened 2024-03-22
fix(filetype): don't use fnamemodify() with :e for extension

zeertzjq:
Use pattern matching instead, as fnamemodify() with :e produces an empty
string when the file name only has an extension, leading to differences
in behavior from Vim.

Related #16955 #27972

### DIFF
--- a/runtime/lua/vim/filetype.lua
+++ b/runtime/lua/vim/filetype.lua
@@ -1472,7 +1472,6 @@ local filename = {
   ['bash.bashrc'] = detect.bash,
   bashrc = detect.bash,
   ['.bashrc'] = detect.bash,
-  ['.env'] = detect.sh,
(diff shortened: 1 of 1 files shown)


gpanders:
Hmm I think I agree with [@clason’s comment](https://github.com/neovim/neovim/pull/27972#discussion_r1535188177). The cases where a file `.foo` has the same matching for `x.foo` seem rare enough to use file name matching. Though if this doesn’t regress anything I suppose there’s no harm? But I also don’t understand what problem (specifically) it solves.

zeertzjq:
Related https://github.com/neovim/neovim/pull/16978/commits/19864bd99529334909922e8d2a61a600fea7b29a

zeertzjq:
Also related: #16955

PR #27978, opened 2024-03-22
fix(float): combine bot border attr by stl attr when on it

glepnir:
Problem: when float bottom border on statusline and Floatborder backgournd is transparency i
t will override statusline highlight

Solution: bot border combine statusline attr when on it.

<details>
<summary>current behavior</summary>

old
</details>

<details>
<summary>with patch</summary>

new

</details>

### DIFF
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -119,6 +119,7 @@
 #include "nvim/version.h"
 #include "nvim/vim_defs.h"
 #include "nvim/window.h"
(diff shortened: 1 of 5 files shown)


bfredl:
This seems a bit ad-hoc? you could set `blend=30` or something on the `FloatBorder` highlight for a similar effect already.

glepnir:
yup very very special case . I tried blend it not works well.  the on stl check can also done in script and modified `FloatBorder` highlight . just though maybe we can handle this situation automatically

PR #27979, opened 2024-03-22
feat(ui): 'smoothscroll' marker for hidden virt line above first line

luukvbaal:
Problem:  There may be hidden virtual lines above the first line
          unbeknownst to the user.
Solution: Show the 'smoothscroll' marker in case there are hidden virtual
          lines above the first line. Scroll up when moving the cursor behind
          the marker and redraw the line when scrolling up by other means.

### DIFF
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -181,6 +181,9 @@ void buf_put_decor(buf_T *buf, DecorInline decor, int row, int row2)
       buf_put_decor_sh(buf, sh, row, row2);
       idx = sh->next;
     }
+    if (row == 0 && decor.data.ext.vt && decor.data.ext.vt->flags & kVTLinesAbove) {
+      adjust_topfill(false);
+    }
   }
 }
 
--- a/src/nvim/drawline.c
+++ b/src/nvim/drawline.c
@@ -103,7 +103,8 @@ typedef struct {
 
   hlf_T diff_hlf;            ///< type of diff highlighting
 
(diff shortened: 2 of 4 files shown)


zeertzjq:
Why only when above first line? That's inconsistent.

luukvbaal:
> Why only when above first line? That's inconsistent.

For all other lines the user already knows it is possible to scroll up.

> And... I don't think this is a good idea anyway.

Why?

zeertzjq:
For 'smoothscroll' the line isn't fully shown anyway, so it doesn't matter much, but for topfill the first line is usually fully shown without the marker, so the marker is quite disruptive.

luukvbaal:
I mean it's quite a niche situation already and it's only disruptive with `:se nonu nornu fdc=0 scl=no`. Not knowing that one can scroll up seems more disruptive to me.

zeertzjq:
> it's only disruptive with `:se nonu nornu fdc=0 scl=no`

I don't think so. Line number being invisible is also a big inconvenience.

luukvbaal:
I mainly opened this for consistency with Vim but upon further inspection it doesn't seem to work properly there anyways so perhaps that was premature. Vim shows the marker for hidden virtual lines anywhere in the buffer. But corrupts the screen with `:se nu`.

Oh well, I guess at least the ruler indicates that we are not at the top of the buffer in this situation.

zeertzjq:
Vim's "above" virtual text implementation leads several bugs related to virtual column, so matching Vim's behavior regarding this isn't always a good idea.

PR #27998, opened 2024-03-23
fix(vim.iter): use correct cmp function when truncating tail in `take`

### LINKED ISSUE #27997: vim.iter doesn't respect `take(n)` when called after `rev()`
### Problem

Calling `take` on a reversed iterator (`vim.iter`) is not respected. If `take` is called before `rev`, then it will be respected. If the iterator is unreversed (`iter:rev():rev()`) then `take` will also again be respected.

### Steps to reproduce

```lua
vim.iter({ 1, 2, 3 }):take(2)
-- GOOD: Produces { 1, 2 }

vim.iter({ 1, 2, 3 }):rev():rev():take(2)
-- GOOD: Produces { 1, 2 }

vim.iter({ 1, 2, 3 }):take(2):rev()
-- GOOD: Produces { 3, 2 }

vim.iter({ 1, 2, 3 }):rev():take(2)
-- PROBLEM: Produces { 3, 2, 1 }
```

### Expected behavior

```lua
vim.iter({ 1, 2, 3 }):rev():take(2)
-- EXPECTED: { 3, 2 }
```

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2671+gdc110cba3

N/A

### Operating system/version

macOS

### Terminal name/version

wezterm

xterm-256color

### Installation

bob

cbochs:
Fixes https://github.com/neovim/neovim/issues/27997

### DIFF
--- a/test/functional/lua/iter_spec.lua
+++ b/test/functional/lua/iter_spec.lua
@@ -139,6 +139,8 @@ describe('vim.iter', function()
 
   it('rev()', function()
     eq({ 3, 2, 1 }, vim.iter({ 1, 2, 3 }):rev():totable())
+    eq({ 3, 2 }, vim.iter({ 1, 2, 3 }):rev():take(2):totable())
+    eq({ 3, 2 }, vim.iter({ 1, 2, 3 }):take(2):rev():totable())
 
     local it = vim.iter(string.gmatch('abc', '%w'))
     matches('rev%(%) requires a list%-like table', pcall_err(it.rev, it))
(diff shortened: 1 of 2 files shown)


test/functional/lua/iter_spec.lua
```diff
@@ -139,6 +139,8 @@ describe('vim.iter', function()
 
   it('rev()', function()
     eq({ 3, 2, 1 }, vim.iter({ 1, 2, 3 }):rev():totable())
+    eq({ 3, 2 }, vim.iter({ 1, 2, 3 }):rev():take(2):totable())
```
gpanders, line 142: Thank you for adding a test case! I think these cases belong in the `take()` test block rather than `rev()` though.

cbochs: Moved!

gpanders (approved):

PR #27999, opened 2024-03-23
feat: bundle vim-commentary

dundargoc:
vim-commentary is a tried and tested minimal plugin that "feels" like an
obvious addition to core. Its minimal size makes it a low-risk addition
to core and we can always easily remove it the need arises.

Disclaimer: this is an early proposal to test the waters. I have not asked for permissions yet.


dundargoc:
I've talked to @echasnovski in private and he'd like to propose an alternative solution to this. I think I will close this for now and let him take ownership of this particular issue. We can always revive it's needed.

PR #28000, opened 2024-03-23
ci: provide separate macos releases for intel and arm

dundargoc:
This will immensely reduce the complexity required to support both
architectures, reduce overall lines of code and unblock follow-up
simplifications.


clason (approved):

dundargoc:
Todo:test releases

.github/workflows/release.yml
```diff
@@ -59,8 +59,12 @@ jobs:
           ./build/bin/nvim --version | head -n 3 >> $GITHUB_OUTPUT
           printf 'END\n' >> $GITHUB_OUTPUT
 
-  macOS:
-    runs-on: macos-14
+  macos:
```
clason, line 62: **nit:** Why that rename? `macOS` is the correct spelling.

clason:
I think that's missing some adjustments to the `publish` step: if you have two different builds, they should be uploaded separately (and marked as `Intel` and `M1`, not `macos-12` and `macos-14`).

Also, we need to adjust the release template to give correct download instructions.

dundargoc:
Ye, I know lol.

.github/workflows/release.yml
dundargoc: I dislike mixing upper case and lower case in code.

clason: As long as output remains correct, that is fine.

.github/workflows/release.yml
```diff
@@ -59,8 +59,17 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ macos-12, macos-14 ]
+        include:
+          - runner: macos-12
+            arch: x86_64
+          - runner: macos-14
+            arch: arm64
```
clason, line 71: These are the default (only) arches for these runners, but it's good to be explicit here 👍

.github/workflows/release.yml
```diff
@@ -72,17 +81,37 @@ jobs:
         run: printf 'CMAKE_BUILD_TYPE=Release\n' >> $GITHUB_ENV
       - if: github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && github.event.inputs.tag_name == 'nightly')
         run: printf 'CMAKE_BUILD_TYPE=RelWithDebInfo\n' >> $GITHUB_ENV
-      - name: Build universal binary
-        run: ./.github/scripts/build_universal_macos.sh
+
+      - name: Build deps
+        run: |
+          MACOSX_DEPLOYMENT_TARGET=11.0
```
clason, line 87:
```suggestion
          export MACOSX_DEPLOYMENT_TARGET=11.0
```
?

PR #28005, opened 2024-03-23
fix(edit): backspace adds extra spaces for inline virtual text

### LINKED ISSUE #26388: Strange white space padding when using  nvim_buf_set_extmark for inlay virtual text
### Problem

Please refer to the attached video
https://github.com/neovim/neovim/assets/1681295/8fe6b0af-719c-49fa-8398-1ca33a6ea3ba

### Steps to reproduce

install nvim head with brew
open following 1.lua code
```lua
local ns = vim.api.nvim_create_namespace('test')
vim.api.nvim_buf_set_extmark(0, ns, 0, 5, {
              virt_text_pos = 'inline',
              ephemeral = false,
              virt_text = {{'[virtual text]'}},
            })
```

```sh
nvim -u NORC 1.lua
```

Run `1.lua`  with `luafile %`
Change to insert mode, move cursor to `1:7` and press `<BS>`; 
`[virtual text]` shifted to right with space added.
Press `<Right>` the cursor stops on next character after virtual text `[virtual text]n|`

### Expected behavior

TBH, I am not sure what is the expected behavior. But I would expected the virtual text remain it's initial position.
Also If my cursor is around virtual text when press `<left>` and `<right>` it should stop on virtual text boundary   `|[virtual text]|` (the `|` stands for expected cursor position) 

### Neovim version (nvim -v)

NVIM v0.10.0-dev-748eae6

I do not know as the API may be unique to nvim

### Operating system/version

macOS 10.15.2

### Terminal name/version

kitty

kitty-256

### Installation

brew install --HEAD

vanaigr:
Fixes #26388.

The problem occurs because `want_vcol` is offset by the width of the virtual text, which increases the number of inserted spaces. Similar issue also happens for 'linebreak' and 'breakindent'.

This can be fixed by ignoring 'linebreak', 'breakindent' and inline virtual text when calculating virtual columns.

I also changed one of the conditions from `if (curwin->w_cursor.col == 0)` to `if (curwin->w_cursor.col <= revins_on)` since when `revins_on` is `true` and the cursor is at the start of the line, its `col` is 1. `vcol` and `want_vcol` are calculated for different lines in this case, which results in this behavior:
revins


test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
...
+    local feed_command = helpers.feed_command
+    local api = helpers.api
+    local function set_lines(line_b, line_e, ...)
+      api.nvim_buf_set_lines(0, line_b, line_e, true, { ... })
+    end
+    local function s(count)
+      return (' '):rep(count)
+    end
+
+    local function setup(width, height)
```
zeertzjq, line 235: `setup()` overshadows a global function.

test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
...
+      clear()
```
zeertzjq, line 236: This is not needed as `clear()` is already called in `before_each()`.

test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
     ]],
     }
   end)
+
+  describe('backspace', function()
+    local feed_command = helpers.feed_command
+    local api = helpers.api
```
zeertzjq, line 227: I think this should be moved to the top of the file.

test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
...
+    end
+    local function s(count)
+      return (' '):rep(count)
+    end
+
+    local function setup(width, height)
+      clear()
+      local screen = Screen.new(width, height + 1)
+      screen:attach()
+      feed_command('setl listchars=space:.,tab:>- list')
```
zeertzjq, line 239: Why not use `command()` instead of `feed_command()`?

test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
...
+        { 7, 17 },
+        { 2, 9 },
+        { 1, 8 }, -- last screen cell of first tab is at vcol 8
+      })
+    end)
+
+    it('works with breakindent', function()
+      setup(17, 3)
+      feed_command('setl ts=4 sw=4 bri briopt=min:5')
+      set_lines(0, 1, '\t' .. s(4) .. '\t' .. s(9) .. '\t ' .. 'a')
```
zeertzjq, line 318:
```suggestion
      set_lines(0, 1, '\t' .. s(4) .. '\t' .. s(9) .. '\t a')
```

test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
...
+        { 7, 9 },
+        { 4, 5 },
+        { 1, 1 },
+      })
+    end)
+
+    it('works in replace mode', function()
+      setup(50, 1)
+      feed_command('setl ts=8 sw=8 sts=8')
+      set_lines(0, 1, '\t' .. s(4) .. '\t' .. s(9) .. '\t ' .. 'a')
```
zeertzjq, line 303:
```suggestion
      set_lines(0, 1, '\t' .. s(4) .. '\t' .. s(9) .. '\t a')
```

test/functional/editor/mode_insert_spec.lua
```diff
@@ -221,4 +221,132 @@ describe('insert-mode', function()
...
+        { 6, 9 },
+        { 2, 5 },
+        { 1, 1 },
+      })
+    end)
+
+    it('works with inline virtual text', function()
+      setup(50, 1)
+      feed_command('setl ts=4 sw=4')
+      set_lines(0, 1, '\t' .. s(4) .. '\t' .. s(9) .. '\t ' .. 'a')
```
zeertzjq, line 335:
```suggestion
      set_lines(0, 1, '\t' .. s(4) .. '\t' .. s(9) .. '\t a')
```

src/nvim/edit.c
```diff
@@ -3752,7 +3735,7 @@ static bool ins_bs(int c, int mode, int *inserted_space_p)
   }
 
   // Delete newline!
-  if (curwin->w_cursor.col == 0) {
+  if (curwin->w_cursor.col <= revins_on) {
```
zeertzjq, line 3738: It seems that this change can cause Nvim to crash. Please remove this change from this PR.

src/nvim/edit.c
```diff
@@ -3846,42 +3829,68 @@ static bool ins_bs(int c, int mode, int *inserted_space_p)
...
-      // the previous character.
-      getvcol(curwin, &curwin->w_cursor, &vcol, NULL, NULL);
-      colnr_T start_vcol = vcol;
-      dec_cursor();
-      getvcol(curwin, &curwin->w_cursor, NULL, NULL, &want_vcol);
-      inc_cursor();
+
+      bool const use_ts = !curwin->w_p_list || curwin->w_p_lcs_chars.tab1;
+      int const lnum = curwin->w_cursor.lnum;
+      char *const line = ml_get_buf(curbuf, lnum);
```
zeertzjq, line 3841: The `lnum` variable seems unnecessary.
```suggestion
      char *const line = get_cursor_line_ptr();
```

src/nvim/edit.c
```diff
@@ -3846,42 +3829,68 @@ static bool ins_bs(int c, int mode, int *inserted_space_p)
...
+      char *const end_ptr = line + curwin->w_cursor.col;
+
+      colnr_T vcol = 0;
+      colnr_T space_vcol = 0;
+      StrCharInfo sci = utf_ptr2StrCharInfo(line);
+      StrCharInfo space_sci = sci;
+      bool prev_space = false;
+      do {
+        bool cur_space = ascii_isspace(sci.chr.value);
```
zeertzjq, line 3850: `ascii_iswhite()` should be used, not `ascii_isspace()`.
```suggestion
        bool cur_space = ascii_iswhite(sci.chr.value);
```

zeertzjq (approved):

PR #28014, opened 2024-03-24
fix(backup): E1509 when writing extended attr to symlink

### LINKED ISSUE #26473: Error E1509 when I write a file that is a symbolic link
### Problem

20231208_11h29m37s_grim

### Steps to reproduce

This happens without plugins and without a configuration file.

Create a example file:
`touch example.txt`

Create a symbolic link of the example file:
`ln -s example.txt example_link.txt`

Open the symbolic link:
`nvim example_link.txt`

Execute `:w`

### Expected behavior

Since there are no problems saving the changes to the file, the expected behavior is that there is no error message.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2289ca2 Build type: Release LuaJIT 2.1.1700008891

no, vim 9.0.2120

### Operating system/version

Fedora 38

### Terminal name/version

alacritty 0.12.3

alacritty

### Installation

build from source

synapse-overload:
Fixes: #26473

### DIFF
--- a/src/nvim/bufwrite.c
+++ b/src/nvim/bufwrite.c
@@ -900,6 +900,14 @@ static int buf_write_make_backup(char *fname, bool append, FileInfo *file_info_o
         // remove old backup, if present
         os_remove(*backupp);
 
+        // copy the file
+        if (os_copy(fname, *backupp, UV_FS_COPYFILE_FICLONE) != 0) {
+          *err = set_err(_("E509: Cannot create backup file (add ! to override)"));
+          XFREE_CLEAR(*backupp);
+          *backupp = NULL;
+          continue;
+        }
+
         // set file protection same as original file, but
         // strip s-bit.
         os_setperm(*backupp, perm & 0777);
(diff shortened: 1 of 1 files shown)


zeertzjq:
It's copied 7 lines below. They don't need to be copied twice. Although I'm a bit confused why this change fixes the problem.

synapse-overload:
The fix solves the issue because the attribute setting with setattx is done on a file that doesn't exist, for it to exist the copy has to be done first and only afterwards the setattx must be done, this is why the lines are copied lower after the copy file to the backup file which is the named the same as the symlink bit with a ~ at the end, the error with which it fails is 2, which is "no such file", and it's only triggered on round=1. The correct sequence is firsr copy file, then set attribute on it. The attribute is a selinux attribute, I debugged with gdb.

synapse-overload:
For extra info, the atrribute being set is "security.selinux" len=17. For normal, non-symlink files this attribute doesn't show up and I believe that's why the copy_xattr doesn't break in that case. My system is Fedora 39 with kernel 6+. I believe the other people that saw the bug are also on fedora.

zeertzjq:
It seems that the `os_setperm()` above doesn't do what it's expected to do either. Maybe the correct solution is to move the `os_copy()` just after the `os_remove()` instead.

synapse-overload:
I missed that, was trying to get my issue to work and to keep the changes minimal. The thing that threw me off was the for loop with round 0 and round 1 which I honestly didn't understand why it's done that way, for exactly 2 rounds hardcoded. But then I saw that the os_copy for the backup was done after the file extended attribute set and it was clear that the file must first physically exist before and 'set anything' is done on it.

Issue #21458, opened 2022-12-17
Implement 'mouseshape' option like Vim


sfsam:
### Feature description

Neovim has a help entry for the `mouseshape` option, but it is currently not implemented. It is useful in GUIs for visual feedback. For example, in Insert mode, the mouse shape is typically an I-Beam whereas it is an Arrow in Normal mode. When hovering over a split or status line, it becomes a left-right or up-down shape to indicate the element can be resized.

This option is implemented in Vim and is used by the MacVim GUI for this purpose. It would be good if Neovim also allowed this option to be set by the user and provided this information to GUIs so they can set the mouse pointer shape as appropriate.

Tmpod:
I'd like to work on this, but I'm unsure where to start. Could anyone give me some pointers? Thanks!

justinmk:
@Tmpod this commit added the 'mousescroll' option which gives a hint about related areas: https://github.com/neovim/neovim/commit/93c8fe77cbfd3934c7599196e573136f75f6e9af

Note that "GUI options" are emitted as UI events for UIs to listen to, via `"option_set"` UI event: https://github.com/neovim/neovim/blob/2dc439c672facaeb8e51ef6aa20efc0e7092eee2/runtime/doc/ui.txt#L63-L66

Tmpod:
Thank you, will look into those! :D

zeertzjq:
`mode_info_set` may be the actual UI event that needs change here.

gpanders:
The TUI can support this with the OSC 22 escape sequence, which is supported by at least Kitty and a few other terminals https://sw.kovidgoyal.net/kitty/pointer-shapes/

The list of cursor shapes supported by that protocol is [here](https://sw.kovidgoyal.net/kitty/pointer-shapes/#pointer-shape-names). Vim has a different list of names. The list in kitty's document is based on the `cursor` property from the [CSS standard](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).

I think we should not enforce a static list of names, but allow users to specify whichever shape name they want (linking to the kitty document above for examples). That way GUIs can implement their own shapes if they want to. But the TUI would only support the shapes listed by the kitty protocol.

PeterCardenas:
ended up going with using the existing `mode_info_set` and `mode_change` events here: https://github.com/neovim/neovim/pull/37186. pr is a bit rough, but would like feedback if this is right direction before cleaning it up (I should probably split up the API change and the TUI change?)

PR #28016, opened 2024-03-24
feat(UI): add ext_elementtype UI extension

fredizzimo:
This adds a new UI extension that contains adds element type information to the highlight attributes. This information can be used to for example detect floating window borders and winbars for a smooth scrolling implementation. Another example use case is to detect the splits, so that the mouse cursor can be changed when hovering it.

I guess that the original idea of `ext_hlstate` and it's `ui` key was similar, but currently that is useless since it does not encode the original highlight groups of linked groups. It also does not nearly capture the same information as this addition does.

See the documentation for further description.

A Neovide example implementation is here:
* https://github.com/neovide/neovide/pull/2438

Fixes https://github.com/neovim/neovim/issues/23610

NOTE: I have only implemented a few element types now, since I think it's better that we get a first implementation that fixes the immediate problem before making it perfect.

The screen.lua had to be reworked quite extensively to support this, because the tests themselves, are using a different format than the internal ui attribute definition. I hope that does not infer with the work that @bfredl is doing, I noticed that since I had to do some fixes myself, due to yesterday's refactoring.

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -179,6 +179,7 @@ Nvim exposes its API metadata as a Dictionary with these items:
                           describing the return value and parameters.
 - ui_events               |UI| event signatures
 - ui_options              Supported |ui-option|s
+- ui_element_tags         String to bit number mapping for |ui-elementtype|
 - {fn}.since              API level where function {fn} was introduced
 - {fn}.deprecated_since   API level where function {fn} was deprecated
 - types                   Custom handle types defined by Nvim
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -38,6 +38,8 @@ with these (optional) keys:
 
 							*ui-ext-options*
 - `ext_cmdline`		Externalize the cmdline. |ui-cmdline|
+- `ext_elementtype`	Gives details about UI element types. |ui-elementtype|
+			Sets `ext_linegrid` implicitly.
 - `ext_hlstate`		Detailed highlight state. |ui-hlstate|
 			Sets `ext_linegrid` implicitly.
 - `ext_linegrid`	Line-based grid events. |ui-linegrid|
@@ -308,7 +310,7 @@ numerical highlight ids to the actual attributes.
 	screen with changed background color itself.
 
 							*ui-event-hl_attr_define*
-["hl_attr_define", id, rgb_attr, cterm_attr, info] ~
+["hl_attr_define", id, rgb_attr, cterm_attr, info, element_tags] ~
 	Add a highlight with `id`  to the highlight table, with the
 	attributes specified by the `rgb_attr` and `cterm_attr` dicts, with the
 	following (all optional) keys.
@@ -356,6 +358,9 @@ numerical highlight ids to the actual attributes.
 
 	`info` is an empty array by default, and will be used by the
 	|ui-hlstate| extension explained below.
+
+	`element_tags` is an empty array by default, and will be used by the
+	|ui-elementtype| extension explained below.
 
 ["hl_group_set", name, hl_id] ~
 	The built-in highlight group `name` was set to use the attributes `hl_id`
@@ -575,6 +580,36 @@ differ, because the builtin group was linked to another group |:hi-link| , or
 because 'winhighlight' was used. UI items will be transmitted, even if the
 highlight group is cleared, so `ui_name` can always be used to reliably identify
 screen elements, even if no attributes have been applied.
+
+==============================================================================
+Details about UI element types				       *ui-elementtype*
+
+Activated by the `ext_elementtype` |ui-option|.
+Activates |ui-linegrid| implicitly.
+
+When this is set the `element_tags` parameter of |ui-event-hl_attr_define|
+contains semantic information about the type of UI element displayed.
+
+`element_tags` is a bitfield integer, with one more of the following tags set:
+    `StatusBar`    Status bars
+    `WinBar`       Winbars
+    `FloatBorder`  Floating window borders, including titles
+    `FloatTitle`   Floating window titles and footers
+    `Top`          A top element, like a floating window top border
+    `Bottom`       A bottom element
+    `Left`         A left element
+    `Right`        A right element
+    `HSplit`       A horizontal split, which also includes StatusBars, since they
+                 act as splits
+    `VSplit`       A vertical split
+
+The name to bit number mapping can be read from |api-metadata|, the bit number is simply the index.
+
+Multiple element tags can be set, so a floating window footer is represented as:
+    `FloatBorder` | `FloatTitle` | `Bottom`
+
+New element types can be added in the future, so UI implementations should be
+prepared to ignore the ones they don't handle.
 
 ==============================================================================
 Multigrid Events						 *ui-multigrid*
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -343,6 +343,9 @@ The following new APIs and features were added.
 
 • |extmarks| option `scoped`: only show the extmarks in its namespace's scope.
 
+• Added a new UI extension `ext_elementtype`, which makes it possible for UIs
+  to track UI elements like floating window borders. |ui-elementtype|
+
 ==============================================================================
 CHANGED FEATURES                                                 *news-changed*
 
--- a/src/nvim/api/ui.c
+++ b/src/nvim/api/ui.c
@@ -38,6 +38,7 @@
 #include "nvim/option.h"
 #include "nvim/types_defs.h"
 #include "nvim/ui.h"
+#include "nvim/ui_defs.h"
 
 #define BUF_POS(ui) ((size_t)((ui)->packer.ptr - (ui)->packer.startptr))
 
@@ -161,7 +162,7 @@ void nvim_ui_attach(uint64_t channel_id, Integer width, Integer height, Dictiona
     }
   }
 
-  if (ui->ui_ext[kUIHlState] || ui->ui_ext[kUIMultigrid]) {
+  if (ui->ui_ext[kUIHlState] || ui->ui_ext[kUIMultigrid] || ui->ui_ext[kUIElementType]) {
     ui->ui_ext[kUILinegrid] = true;
   }
 
@@ -672,7 +673,7 @@ void remote_ui_default_colors_set(RemoteUI *ui, Integer rgb_fg, Integer rgb_bg, 
 }
 
 void remote_ui_hl_attr_define(RemoteUI *ui, Integer id, HlAttrs rgb_attrs, HlAttrs cterm_attrs,
-                              Array info)
+                              Array info, Integer element_tags)
 {
   if (!ui->ui_ext[kUILinegrid]) {
     return;
@@ -700,6 +701,8 @@ void remote_ui_hl_attr_define(RemoteUI *ui, Integer id, HlAttrs rgb_attrs, HlAtt
   } else {
     ADD_C(args, ARRAY_OBJ((Array)ARRAY_DICT_INIT));
   }
+
+  ADD_C(args, INTEGER_OBJ(element_tags));
 
   push_call(ui, "hl_attr_define", args);
 }
--- a/src/nvim/api/ui.h
+++ b/src/nvim/api/ui.h
@@ -17,8 +17,22 @@ EXTERN const char *ui_ext_names[] INIT( = {
   "ext_linegrid",
   "ext_multigrid",
   "ext_hlstate",
+  "ext_elementtype",
   "ext_termcolors",
   "_debug_float",
+});
+
+EXTERN const char *ui_element_tag_names[] INIT( = {
+  "StatusBar",
+  "WinBar",
+  "FloatBorder",
+  "FloatTitle",
+  "Top",
+  "Bottom",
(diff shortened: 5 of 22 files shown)


bfredl:
I have not checked the details yet but if `ext_hlstate` has a bug in its implementation we could fix that bug instead of reinventing `ext_hlstate_but_hlstate_again`, perhaps?

fredizzimo:
I'm not sure if `ext_hlstate` can be fixed without completely breaking existing potential uses. Even if the `ui` key of `info` array actually referred to the original built-in highlight group, it still can't capture all the information that I capture here, it can only use actual highlight groups. And I'm not sure if that was the actual intention or not either. It also contains a lot of useless information, at least for Neovide, and currently I can't think of anything we would like to use it for, so in Neovide we only need this new `ext_elementype` extension.

That said, this new extension shares almost everything with `ext_hlstate`, including almost all of the implementation, but instead of extending `info`, it adds `element_tags`, so it's more like a sub-option of `ext_hlstate`. So mabye it's possible to combine them somehow. But if that's the case, I think we need to fix the current performance problems that `ext_hlstate`, but not `ext_elementtype` also suffers from, it sends N(N+1) / 2 `hl_attr_define events` when deleting lines.

bfredl:
missing info can be fixed by adding new ui highlight groups or splitting them (which can be done fully compatibly using default links).

"information overflow" shouldn't be a big problem in practice as UI:s can extract the useful info at _definition_ time (they don't need to rework anything per cell).

fredizzimo:
I'm not sure about mixing two different types of concepts here, just to get rid of an additional extension.

1. The highlight groups.
   * Have attributes assigned to them
   * Are sometimes combined, sometimes blended together
   * Can be linked to another group
   * Describes a cell's visual appearance like a CSS style
2. The element types
   * Are just tags, without any attributes
   * They are always additive, several types can be assigned to a cell
   * Are absolutely not allowed to be linked to something else
   * Describes the functionality of a cell, like a html element type

Perhaps the confusion comes from the fact that I added it to the `hl_attr_define` UI event? I could have created a similar event that shares over 90% of the implementation, but sharing code is not too easy do in C especially since I have very little practical straight C experience. Alternatively, the event could be named `cell_attr_define`, but we have to deal with backwards compatibility.

Or maybe you want me to keep the current structure and add the element type to the info array? But that does not make much sense IMO. It's already confusing with some fields valid for only some types, like `ui_name` only valid for the `ui` kind. I have honestly no idea what this even means, because for example a floating window title has `kind = syntax` and the `hi_name = Title`, while the border has `kind = ui`, `hi_name = FloatBorder` and `ui_name = FloatBorder`

Since the element type is a combination of tags, the bitfield representation I add here is much more compact and easier to parse and use.

Using actual highlight groups and links here makes even less sense. The `HLEntry` struct which the code deals with does not have any information about linked highlight groups, they are resolved way before that. And neither does the info array, you need a tree-like structure to represent the all the different highlight groups in a cell and their parents, an array is not enough.

But those are just the technical challenges, there are numerous design problems, that at least I can't solve, here are a few examples:

1. Let's say we have the two highlight groups `FloatBorder` and `FloatTitle`, and the user links them together, how can we ever know which one is which? Remember, the link can go in either direction.
2. If the user sets a custom highlight group for the floating window border. How do we represent that? Do we automatically also add a "magic" "FloatBorder" highlight that doesn't define any attributes to the cell in addition to the user highlight group? 
3. How do we represent combinations, `FloatBorder, FloatTitle, Top`? For the title at the top, which also happens to be logically a border. The user could specify highlights for all of them, which one is used? Or do we not support it at all? In this case the `FloatTitle` encodes everything, if we also have `FloatFooter`, but what about `FloatBorder, Top, Left`, that's useful information that won't be possible to encode.
4. Some of the more meta of the element types like `Top` doesn't even make sense as a highlight group, what does it mean to specify some highlight for `Top`?

I'm sure that all of this is "solvable", but at what cost? The current implementation is less than 100 lines of code if you remove the testing code, I'm sure any of the above implementations will be more complex, especially if we also have to fix the bugs in `ext_hlstate`. At Neovide we are overflooded with bugreports about crashes, slowdowns and poor scrolling behaviour, so we would rather see this fixed sooner than later, and really hope to get this for nvim 0.10. Also, in that sense, adding a new extension that is not used in the wild, is much safer than fixing some "bug" that users rely on.

> "information overflow" shouldn't be a big problem in practice as UI:s can extract the useful info at definition time (they don't need to rework anything per cell).

If information overload does not matter, why isn't `ext_hlstate` enabled by default then? The difference is just between getting an array of zero elements, and an array with information you don't care about, it wouldn't break any clients.

But it's not just about information overflow, it's also about performance. Neovide runs in real time, and for a smooth experience we should get responses as soon as possible, preferably within a frame of 8ms. Especially when scrolling it needs to be fast, if it starts taking more than the keyboard repeat rate of around 30 ms, all hope is lost and updates starts to stack on top of each other, and it starts to lag badly. We also can't afford millions of pointless `ext_hlstate` events when deleting text, but that's a bug and should be fixable. 

PS. It would also have been nice to have had this discussion before I started with the implementation, especially since @justinmk gave a thumbs up here https://github.com/neovim/neovim/issues/23610#issuecomment-1929366416.

runtime/doc/api.txt
```diff
@@ -179,6 +179,7 @@ Nvim exposes its API metadata as a Dictionary with these items:
                           describing the return value and parameters.
 - ui_events               |UI| event signatures
 - ui_options              Supported |ui-option|s
+- ui_element_tags         String to bit number mapping for |ui-elementtype|
```
justinmk, line 182: what is "tags" doing for us in this name
```suggestion
- ui_elements         String to bit number mapping for |ui-elementtype|
```

runtime/doc/ui.txt
```diff
@@ -38,6 +38,8 @@ with these (optional) keys:
 
 							*ui-ext-options*
 - `ext_cmdline`		Externalize the cmdline. |ui-cmdline|
+- `ext_elementtype`	Gives details about UI element types. |ui-elementtype|
```
justinmk, line 41: not clear why "types" is appended everywhere. It's emitted info about UI elements, right?

```suggestion
- `ext_elements`	Gives details about UI element. |ui-elementtype|
```

runtime/doc/ui.txt
```diff
@@ -576,6 +581,36 @@ because 'winhighlight' was used. UI items will be transmitted, even if the
...
+
+Activated by the `ext_elementtype` |ui-option|.
+Activates |ui-linegrid| implicitly.
+
+When this is set the `element_tags` parameter of |ui-event-hl_attr_define|
+contains semantic information about the type of UI element displayed.
+
+`element_tags` is a bitfield integer, with one more of the following tags set:
+    `StatusBar`    Status bars
+    `WinBar`       Winbars
```
justinmk, line 595: This doesn't explain at all why this new extension exists. All of the UI "ext" things literally exist to emit info about various UI widgets ("elements"). So this seems like an [inner platform](https://en.wikipedia.org/wiki/Inner-platform_effect), i.e. it's emitting UI info inside of the UI info system...

justinmk:
> We also can't afford millions of pointless `ext_hlstate` events when deleting text, but that's a bug and should be fixable.

Can we start with that?

> PS. It would also have been nice to have had this discussion before I started with the implementation, especially since @justinmk gave a thumbs up here [#23610 (comment)](https://github.com/neovim/neovim/issues/23610#issuecomment-1929366416).

PRs are part of the discussion.

runtime/doc/ui.txt
fredizzimo: The rest of the documentation does not really motivate each extension's existence either, so I didn't add that information to the documentation. But I can add a short description if needed.

The motivation is in the PR description and in the linked issue, but here's a longer one.

The immediate problem for Neovide is that UI elements like winbars, and borders are not detectable by the existing UI extensions, at least not in the general case, when the highlight group is something different than `FloatBorder` or `WinBar`. And they are part of the same grid as the buffer, so without being able to detect them it's not possible to know which parts of the grid should scroll.

I originally requested for them to be represented as separate grids, so we wouldn't need to care, but that was turned down by @bfredl here https://github.com/neovim/neovim/issues/23610#issuecomment-1867430580

But when I started to think more about it, I could think about plenty of more use cases. Changing the mouse shape when hovering elements is one of them, and that needs to be responsive, so a roundtrip to nvim is not an option. That's also one example, where only giving information about the borders as suggested, is not enough. You might want to change the cursor when hovering over URL links in the middle of the window for example.

Another use case is to render some of these elements in a more native style. For example, if we know that the type is s a border, then we could skip rendering the outer edges(background) of the glyphs. But to do that we need to know that it's a border, it could also be regular buffer text at the edge of the window. This is an alternative to having separate heavyweight extensions for everything, like `ext_tabline`, which also requires taking over the layout and a lot of other things.

BTW, I think UI elements is the correct term here https://en.wikipedia.org/wiki/List_of_graphical_user_interface_elements, for example a border is not of the widget subtype. But some of the elements are indeed widgets.

runtime/doc/ui.txt
fredizzimo: In my programmer's mind, we get information about the type of the UI element, not the actual UI element itself, or even details about a given element. But I agree it's a bit redundant, so I can remove it.

runtime/doc/api.txt
fredizzimo: Tags was added because by itself a single tag is not necessarily an UI element by itself, for example `Left` is one of the possible tags.

But if you prefer it can be changed to `ui_elements`

fredizzimo:
> >We also can't afford millions of pointless ext_hlstate events when deleting text, but that's a bug and should be fixable.

> Can we start with that?

Only if we agree on a design where `ext_hlstate` is important and needed by us in the future. Otherwise, I have plenty of reported bugs here that I rather fix and that affect both me and our users.

bfredl:
> but that was turned down by @bfredl here https://github.com/neovim/neovim/issues/23610#issuecomment-1867430580

It was """turned down"""" together with a concrete solution of the original problem which still stand as my suggestion.

fredizzimo:
> > but that was turned down by @bfredl here [#23610 (comment)](https://github.com/neovim/neovim/issues/23610#issuecomment-1867430580)
> 
> It was """turned down"""" together with a concrete solution of the original problem which still stand as my suggestion.

And I responded with this:
> I think we need to remain flexible and allow these "widgets" anywhere in the window, not just around the borders. Static information related to the window is also a little bit inflexible, for example, consider the floating window titles, they also need the exact position horizontally (to prevent one of errors, by having the clients calculate it themselves. There are also elements like splits, that does not belong to a window, but still, something that an UI might want to know about. For example, in Neovide we would like to display a different cursor when the mouse is over a split, once mouseshape support is implemented https://github.com/neovim/neovim/issues/21458.

Explaining why the proposed solution is not optimal and given that there were no further comments and a thumbs up by justinmk, I assumed that everyone agreed, especially since I said that I will start working on it.

fredizzimo:
This PR solves the original problem. I don't see why we need a duplicate solution for it.

vhakulinen:
As a fellow gui developer, this feature seems a bit out of place.

Many of the problems could be solved by diverging further from the TUI aspect. Winbar and statusline? Turn them off and create GUI specific feature. Floating window borders? Should be described as part of `win_float_pos` event. Splits? Ditch the base grid and nvim's window management and do it in the GUI (haven't tried it but its something I'm planning to do).

Maybe my approach for nvim GUI is different form that of neovide, but thats my 2 cents anyways.

Also, regarding the existing hl events: I've also found that they should include the hlgroup name and that there should be a API to subscribe to more hl groups that just he builtin ones.

runtime/doc/ui.txt
```diff
@@ -576,6 +581,36 @@ because 'winhighlight' was used. UI items will be transmitted, even if the
...
 screen elements, even if no attributes have been applied.
 
+==============================================================================
+Details about UI element types				       *ui-elementtype*
+
+Activated by the `ext_elementtype` |ui-option|.
+Activates |ui-linegrid| implicitly.
+
+When this is set the `element_tags` parameter of |ui-event-hl_attr_define|
+contains semantic information about the type of UI element displayed.
```
justinmk, line 591: Why make it optional? An extra field in `hl_attr_define` is not a breaking change per `:help api-contract`, so we can just always emit it.

But I do like the leverage gained by [bfredl's suggestion](https://github.com/neovim/neovim/pull/28016#issuecomment-2018857464) to use highlight groups for this purpose instead.

justinmk:
> I originally requested for them to be represented as separate grids ... but that was turned down by bfredl here [#23610 (comment)](https://github.com/neovim/neovim/issues/23610#issuecomment-1867430580)

>> What we need to supply the full information to the UI how many cells on the top, left, right, and bottom are being used as fixed areas instead of the main scrolling text

For reference, that's now implemented as `win_viewport_margins`: https://github.com/neovim/neovim/pull/28081

re https://github.com/neovim/neovim/pull/28016#issuecomment-2018857464 : 

>> missing info can be fixed by adding new ui highlight groups or splitting them (which can be done fully compatibly using default links).
>
> I'm not sure about mixing two different types of concepts here, just to get rid of an additional extension.

Highlight groups are a general mechanism for annotations. I wouldn't say this is a different concept.

PR #28018, opened 2024-03-24
refactor(options): make `'maxcombine'` an immutable option

famiu:
Problem: Currently, the `immutable` property of options can be applied for options that are hidden and options whose value simply can't be changed. Which is problematic when attempting to convert an option like `'maxcombine'` into an immutable option, because trying to `:set` an immutable option currently gives an error, which is only desired behavior for hidden options, not options that are actually immutable.

Solution: Separate the `immutable` property into two distinct `hidden` and `immutable` properties. Change all options with the `immutable` property to use the `hidden` property instead. Also add `p_mco` as an `immutable` option, as its value cannot be changed, and the underlying variable is not used anywhere.

### DIFF
--- a/src/nvim/generators/gen_options.lua
+++ b/src/nvim/generators/gen_options.lua
@@ -164,14 +164,19 @@ local function dump_option(i, o)
   if o.enable_if then
     w(get_cond(o.enable_if))
   end
+
+  -- Options cannot be both hidden and immutable.
+  assert(not o.hidden or not o.immutable)
+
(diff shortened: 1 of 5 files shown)


bfredl (approved):

zeertzjq:
I think these two names should be exchanged. Existing mentions of "hidden" options are like 'shellslash' on non-Windows: settings its value is ignored and its value is always 0, which is more similar to "immutable" in this PR.

PR #28025, opened 2024-03-25
fix(intro): still show intro message with floating window

luukvbaal:
Stop drawing the intro only after a split is opened.

### DIFF
--- a/test/functional/ui/messages_spec.lua
+++ b/test/functional/ui/messages_spec.lua
@@ -1604,13 +1604,14 @@ describe('ui/ext_messages', function()
       [5] = { foreground = Screen.colors.Blue1 },
       [6] = { reverse = true },
       [7] = { bold = true, reverse = true },
+      [8] = { background = Screen.colors.Plum1 },
     })
   end)
 
   it('supports intro screen', function()
     -- intro message is not externalized. But check that it still works.
     -- Note parts of it depends on version or is indeterministic. We ignore those parts.
-    screen:expect([[
+    local introscreen = [[
       ^                                                                                |
       {1:~                                                                               }|*4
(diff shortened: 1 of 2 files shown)


zeertzjq (commented):
I think you can first check `last_window(firstwin)`, and then only check the buffer of `firstwin`.

luukvbaal:
That would hide the intro upon opening a new tabpage, which we currently don't.
`one_window()` could work.

test/functional/ui/messages_spec.lua
```diff
@@ -1628,32 +1629,24 @@ describe('ui/ext_messages', function()
...
+
+    -- opening a float also does not
+    local win = api.nvim_open_win(0, false, {
+      relative = 'editor',
+      height = 1,
+      width = 1,
+      row = 5,
+      col = 0,
+    })
+    command('mode')
```
bfredl, line 1647: I'd prefer if the floating window was actually visible and then assert the new screenstate with intro message and float both visible, even if it means the `introscreen` de-duplication cannot be used. We don't make screen tests less robust to be able to code golf them.

test/functional/ui/messages_spec.lua
```diff
@@ -1628,32 +1629,24 @@ describe('ui/ext_messages', function()
...
+    screen:expect { unchanged = true, grid = introscreen, showmode = { { '-- INSERT --', 3 } } }
+    api.nvim_win_close(win, true)
```
bfredl, line 1649: then, needs an assert that closing the window returns to the previous state that only the intro message was visible.

Otherwise, the `expect` below doesn't verify that specifically `feed('x')` was what hid the intro message anymore.

PR #28044, opened 2024-03-26
vim-patch:9.1.{0211,0215}

luukvbaal:
**vim-patch:9.1.0211: page-wise scrolling does not support smooth-scrolling**

Problem:  Page-wise scrolling with Ctrl-F/Ctrl-B implements
          it's own logic to change the topline and cursor.
          In doing so, skipcol is not handled properly for
          'smoothscroll', and virtual lines.
Solution: Re-use the logic from Ctrl-E/Ctrl-Y while staying
          backward compatible as much as possible.

https://github.com/vim/vim/commit/b9f5b95b7bec2414a5a96010514702d99afea18e

**vim-patch:9.1.0215: Half-page scrolling does not support smooth-scrolling**

Problem:  Page-wise scrolling with Ctrl-D/Ctrl-U implements
          it's own logic to change the topline and cursor.
          More logic than necessary for scrolling with Ctrl-F/Ctrl-B
          was removed in patch 9.1.0211.
Solution: Re-use the logic from Ctrl-E/Ctrl-Y/Ctrl-F/Ctrl-B while
          staying backward compatible as much as possible.
          Restore some of the logic that determined how many lines will
          be scrolled (Luuk van Baal)

https://github.com/vim/vim/commit/5a2e3ec9ac72b6e644fea4ebba7e632498296e2f


zeertzjq:
Please use `w_height_inner` instead of `w_height`.

test/old/testdir/test_normal.vim
```diff
@@ -4182,20 +4184,34 @@ func Test_normal34_zet_large()
...
 endfunc
 
-" Test for { and } paragraph movements in a single line
-func Test_brace_single_line()
-  let text =<< trim [DATA]
-    foobar one two three
-  [DATA]
+" Test for { and } paragraph movements and Ctrl-B in buffer with a single line
+func Test_single_line_scroll()
+  CheckFeature textprop
```
zeertzjq, line 4189: This test seems strange. It tests two unrelated things together.

zeertzjq (approved):

Issue #28024, opened 2024-03-25
Add option to filter vim.diagnostic.goto_next/goto_prev by a list of namespaces


nhruo123:
### Problem

Currently we can't filter vim.diagnostic.goto_next/goto_prev with a list of namespaces, this could be very useful, for example when you want to create a separate bind for a specific diagnostic namespace and another for everything but that namespace. 

I use cspell with none-ls as my spell checker in neovim, but I don't want my spell errors to to share the same goto_next and goto_prev list as my LSP and linter diagnostics because it hard to move around the buffer with [d and ]d. With this addition it would be possible to list every diagnostic namespace but the cspell one and create separate bindings for jumping to spell diagnostics and the rest of my diagnostics.

### Expected behavior

Make `vim.diagnostic.GetOpts.namespace` accept an integer or a list of integers. And make sure that functions that use `vim.diagnostic.GetOpts` are updated accordingly.

PR #28045, opened 2024-03-26
feat(diagnostic): add support for many namespaces filtering in GetOpts

mayromr:
Implements #28024 by making `vim.diagnostic.GetOpts.namespace` `integer[]|integer`.
I am a very inexperienced with Neovim development so I am not sure if this is a breaking change or not (this type is never returned from Neovim API, but if someone uses this type in his API it could cause problems as now it could receive an array and not expect it). If it is a breaking change another approach could be adding a disallow List instead.


lewis6991 (approved):

gpanders:
Thank you!

PR #28050, opened 2024-03-26
refactor(options): require `enable_if = false` iff no variable

zeertzjq:
This makes grepping for unsupported options easier.


bfredl:
even tho it has no effect, it does make it more visible what options are disabled, I think (easier to grep for than the absence of 'varname').

zeertzjq:
Yes, so maybe make specifying both `varname` and `enable_if = false` an error instead?

justinmk (approved):

PR #28054, opened 2024-03-27
vim-patch:9.1.0208: winfixbuf does not allow to re-edit current buffer

### LINKED ISSUE #14237: 'winfixbuf' should allow :edit without args
## Problem

Raised by @tpope at https://github.com/tpope/vim-fugitive/issues/2272#issuecomment-2004992176 : 

> This option may be too aggressive. It prevents reloading the window by calling `:e` with no arguments. That doesn't change what buffer the window points to, but it is nonetheless blocked.

## Expected behavior

'winfixbuf' should not show an error if `:edit` is used without args to reload the buffer (which may be application-defined for non-file buffers).

cc @ColinKennedy

ColinKennedy:
#### vim-patch:9.1.0208: winfixbuf does not allow to re-edit current buffer

Problem:  winfixbuf does not allow to re-edit current buffer
          (Tim Pope, after v9.1.0147)
Solution: Explicitly allow :e even when 'winfixbuf' is set,
          since it just re-loads the current buffer
          (Colin Kennedy)

fixes: vim/vim#14237
closes: vim/vim#14286

https://github.com/vim/vim/commit/65e580bd5610465bb6b9c1a546b7a8d00c76aa47

Co-authored-by: Colin Kennedy <colinvfx@gmail.com>

### DIFF
--- a/src/nvim/ex_docmd.c
+++ b/src/nvim/ex_docmd.c
@@ -216,7 +216,6 @@ static void restore_dbg_stuff(struct dbg_stuff *dsp)
   current_exception = dsp->current_exception;
 }
 
-
 /// Check if ffname differs from fnum.
 /// fnum is a buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
 /// ffname is a full path to where a buffer lives on-disk or would live on-disk.
--- a/src/nvim/ex_docmd.c
+++ b/src/nvim/ex_docmd.c
@@ -216,6 +216,39 @@ static void restore_dbg_stuff(struct dbg_stuff *dsp)
   current_exception = dsp->current_exception;
 }
 
+
+/// Check if ffname differs from fnum.
+/// fnum is a buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
+/// ffname is a full path to where a buffer lives on-disk or would live on-disk.
+static bool is_other_file(int fnum, char *ffname)
+{
+  if (fnum != 0) {
+    if (fnum == curbuf->b_fnum) {
+      return false;
+    }
+
+    return true;
+  }
+
+  if (ffname == NULL) {
+    return true;
+  }
+
+  if (*ffname == NUL) {
+    return false;
+  }
+
+  if (!curbuf->file_id_valid
+      && curbuf->b_sfname != NULL
+      && *curbuf->b_sfname != NUL) {
+    // This occurs with unsaved buffers. In which case `ffname`
+    // actually corresponds to curbuf->b_sfname
+    return path_fnamecmp(ffname, curbuf->b_sfname) != 0;
+  }
+
+  return otherfile(ffname);
+}
+
 /// Repeatedly get commands for Ex mode, until the ":vi" command is given.
 void do_exmode(void)
 {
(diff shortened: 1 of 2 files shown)


src/nvim/ex_docmd.c
```diff
@@ -216,6 +216,39 @@ static void restore_dbg_stuff(struct dbg_stuff *dsp)
   current_exception = dsp->current_exception;
 }
 
+
```
zeertzjq, line 219:
```suggestion
```

src/nvim/ex_docmd.c
```diff
@@ -5371,11 +5403,12 @@ static void ex_find(exarg_T *eap)
 /// ":edit", ":badd", ":balt", ":visual".
 static void ex_edit(exarg_T *eap)
 {
-  // Exclude commands which keep the window's current buffer
+  char *ffname = eap->cmdidx == CMD_enew ? NULL : eap->arg;
+
```
zeertzjq, line 5407:
```suggestion

  // Exclude commands which keep the window's current buffer
```

zeertzjq (approved):

src/nvim/ex_docmd.c
```diff
@@ -216,6 +216,38 @@ static void restore_dbg_stuff(struct dbg_stuff *dsp)
   current_exception = dsp->current_exception;
 }
 
+/// Check if ffname differs from fnum.
+/// fnum is a buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
+/// ffname is a full path to where a buffer lives on-disk or would live on-disk.
```
justinmk, line 221: Should it check that ffname is actually a full path (or expand it)? Is `eap->arg` in `ex_edit` guaranteed to be a full path?

ColinKennedy: I'm not at my machine so I can't give you hard data but I remember ffname wasn't a silver bullet

These were the considerations that I had to make while writing the code, generally

- Does it live on disk
- What is the user's current `:pwd`
    - Are they referring to a buffer with a relative path that does or does not exist on-disk (if relative, what's the `:pwd`)
- Does `curbuf->fname` point to `curbuf->ffname` or `curbuf->sfname` and is `curbuf->ffname` the same as `curbuf->sname` currently

In the end I actually got all of the weirdest edge cases working in Neovim but I used a variable that Vim doesn't guarantee to exist. So I haven't submitted a PR yet. Luckily the unsolved cases are pretty edge-of-edge case so I figured we could take our time with solving them

justinmk: Yeah this is a pain [for plugins](https://github.com/justinmk/vim-dirvish/blob/b660af1fa07fe1d44d4eb3ea5242334f6c2766ca/autoload/dirvish.vim#L459-L470) too.

If there isn't already, we might want to move this function to a more re-usable place, for internal use at least. Externally, `vim.fs.normalize()` is supposed to help with comparing filepaths, but still doesn't cover the weird edge cases of Vim buffer equivalence, so we might need e.g. a `bufequal()` function.

ColinKennedy: That does sound like a `bufequal` function would be pretty useful. Probably worth mentioning an effort to standardize would likely run into https://github.com/vim/vim/pull/14286#issuecomment-2023134377 unless a workaround that works for Vim and Neovim is found. But FWIW I have all of that code working in Neovim so we could implement it whenever we like. Want me to make a `bufequal` GitHub issue on https://github.com/vim/vim or should we hold off?

justinmk: > Want me to make a `bufequal` GitHub issue on https://github.com/vim/vim

SGTM!

ColinKennedy: Added here: https://github.com/vim/vim/issues/14340

PR #28060, opened 2024-03-27
test: skip flaky test

dundargoc:


zeertzjq (changes requested):
Please see how other tests in this file are skipped.

dundargoc:
I thought I did. Multiple tests have been commented out: https://github.com/neovim/neovim/blob/master/test/old/testdir/test_perl.vim#L183-L185 which is why I thought this was how vim skipped tests.

dundargoc:
I've commented out only the `call` if that's what you man

zeertzjq (approved):

PR #28064, opened 2024-03-28
fix(fs): allow backslash characters in unix paths

### LINKED ISSUE #28061: `vim.fs.normalize` mangles unix paths with backslash
### Problem

Backslashes are valid characters in unix paths.
Current implementation of `vim.fs.normalize` converts all backslashes to forward slashes, resulting in invalid/wrong paths.

see: https://github.com/neovim/neovim/issues/27958#issuecomment-2022937067

### Steps to reproduce

`:lua =vim.fs.normalize([[/a/b\c]])` --> becomes `/a/b/c`

### Expected behavior

Backslashes should be untouched for unix paths.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2718+g1bfe6fdce

no, vim.fs

### Operating system/version

Linux archlinux 6.7.5-arch1-1

### Terminal name/version

kitty 0.31.0

tmux-256color

### Installation

build from repo

jamestrew:
Backslashes are valid characters in unix style paths.

This PR aims to fix the conversion of backslashes to forward slashes in several `vim.fs` functions when not on Windows. On Windows, backslashes will still be converted to forward slashes.

closes https://github.com/neovim/neovim/issues/28061

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -1,6 +1,7 @@
 local M = {}
 
 local iswin = vim.uv.os_uname().sysname == 'Windows_NT'
+local os_sep = iswin and '\\' or '/'
 
 --- Iterate over all the parents of the given path.
 ---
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/fs.lua
```diff
@@ -47,19 +48,20 @@ function M.dirname(file)
     return nil
   end
   vim.validate({ file = { file, 's' } })
-  if iswin and file:match('^%w:[\\/]?$') then
-    return (file:gsub('\\', '/'))
-  elseif not file:match('[\\/]') then
+  file = file:gsub(os_sep, '/') --[[@as string]]
```
gpanders, line 51: Worth gating this with an `if iswin` (or `if os_sep ~= '/'`)? For a long path, could potentially be wasteful on Unix.

runtime/lua/vim/fs.lua
```diff
@@ -72,10 +74,11 @@ function M.basename(file)
     return nil
   end
   vim.validate({ file = { file, 's' } })
-  if iswin and file:match('^%w:[\\/]?$') then
+  file = file:gsub(os_sep, '/') --[[@as string]]
```
gpanders, line 77: Same comment as above.

gpanders (approved):
Thanks. "Literally anything but NUL can be a path character" is the original sin of POSIX.

Test cases are what I really care about. `vim.fs.normalize` has proven how tricky it is to work with paths, so I'm glad that we're making the tests around this more comprehensive.

runtime/lua/vim/fs.lua
jamestrew: yeah sounds good! implemented

Issue #27938, opened 2024-03-19
Missing `work done progress token` from client init


vilelamarcospaulo:
### Problem

Right now, `clojure-lsp` relies on the microsoft lsp specification that delegates the `Work Done progress token` creation to the sender. 

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initiatingWorkDoneProgress
> Work Done progress can be initiated in two different ways:
> 
> 1. by the sender of a request (mostly clients) using the predefined workDoneToken property in the requests parameter literal. The document will refer to this kind of progress as client initiated progress.

The problem is their implementation ignores the `$/progress` publication when this is not present, this way none `lsp-status` tool  works correctly to give to user feedback, what is bad since we don't really know when the processing is finished and can take a while. 
Others LSP (luals, golsp, etc...) just sets a random value `(string | int)` when it's not present and everything is working.

But in this case (maybe other LSP also relies in this definitions) it's just is ignored. 

To resolve this, in my local neovim build updated the `function Client:initialize()` implementation to also sent the property `workDoneToken = "foobar"` inside the `initialize_params`. Since this is the same approach used by `vscode` and `emacs`.
After this change the `$/progress` starts to receive every progress update, thus `lsp-status` and `fidget` was working just fine.

### Steps to reproduce using "nvim -u minimal_init.lua"

```
vim.lsp.set_log_level("debug")
vim.lsp.start_client({
  name = 'my-server-name',
  cmd = { 'clojure-lsp' },
  root_dir = vim.fn.getcwd()
})

```

### Expected behavior

should contains the `$/progress` in the lsp debug logs

```
-- [DEBUG][2024-03-19 18:26:22] .../vim/lsp/rpc.lua:403	"rpc.receive"	{  jsonrpc = "2.0",  method = "$/progress",  params  = {    token = "foobar",    value = {      kind = "begin",      percentage = 0,      title = "clojure-lsp"    }  }}
-- [DEBUG][2024-03-19 18:26:22] .../vim/lsp/rpc.lua:403	"rpc.receive"	{  jsonrpc = "2.0",  method = "$/progress",  params = {    token = "foobar",    value = {      kind = "report",      message = "Finding kondo config",      percentage = 5    }  }}
```

### Neovim version (nvim -v)

Nvim 0.9.5

### Language server name/version

clojure-lsp/2024.02.01-11.01.59

### Operating system/version

macos/14.2.1  (sonoma)

[...]

PR #28078, opened 2024-03-28
feat(lsp): conditionally send a work done progress token on initialize

rudiejd:
Problem:
https://github.com/neovim/neovim/issues/27938 we don't send a progress token on initialize

Solution:
Conditionally send a progress token if a flag is set. 

Apparently there's no harm in sending one, but since the LSP doesn't require one to be sent, I'm guessing disabling this feature might come in handy in case some servers aren't expecting it on initialize.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1012,6 +1012,9 @@ Lua module: vim.lsp.client                                        *lsp-client*
                                   request before sending kill -15. If set to
                                   false, nvim exits immediately after sending
                                   the "shutdown" request to the server.
+                                • {progress_on_init}? (`boolean`, default:
+                                  `true`) Send a progress token on
+                                  initialization
       • {get_language_id}       (`fun(bufnr: integer, filetype: string): string`)
       • {capabilities}          (`lsp.ClientCapabilities`) The capabilities
                                 provided by the client (editor or tool)
@@ -1183,6 +1186,9 @@ Lua module: vim.lsp.client                                        *lsp-client*
                                 request before sending kill -15. If set to
                                 false, nvim exits immediately after sending
                                 the "shutdown" request to the server.
+                              • {progress_on_init}? (`boolean`, default:
+                                `true`) Send a progress token on
+                                initialization
       • {root_dir}?           (`string`) Directory where the LSP server will
                               base its workspaceFolders, rootUri, and rootPath
                               on initialization.
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -40,7 +40,7 @@ The following changes may require adaptations in user config or plugins.
     set selection=exclusive
     set selectmode=mouse,key
     set mousemodel=popup
-    set keymodel=startsel,stopsel
+    setn keymodel=startsel,stopsel
 <
 • When switching windows, |CursorMoved| autocommands trigger when Nvim is back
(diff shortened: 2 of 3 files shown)


clason:
In particular, the spec says:

> To keep the protocol backwards compatible servers are only allowed to use `window/workDoneProgress/create` request if the client signals corresponding support using the client capability `window.workDoneProgress` which is defined as follows: [...]

So we should make sure that we set this capability correctly, but otherwise expect servers to respect this correctly. We do _not_ add _any_ workarounds for misbehaving servers!

mfussenegger:
Wouldn't it make more sense to have this changed in the server? If the client doesn't report a token, they can generate one.

I'd also rather not have a new flag/setting for this, but if we set the token on the client side it could limit servers in regards how they can group notifications. Without this, they could use many different tokens, with this, they'd have to assign the same token for all progress relevant for initialize.

For current neovim core that shouldn't be an issue because there is no display that groups the notifications, but it could affect plugins in some negative way.


---

> So we should make sure that we set this capability correctly

We do:

https://github.com/neovim/neovim/blob/dde2cc65fd2ac89ad88b19df08dc03cf1da50316/runtime/lua/vim/lsp/protocol.lua#L847

rudiejd:
@mfussenegger I think the issue here is that, while the server can generate one, they're not required by the protocol to generate one. My interpretation is that the progress token can be generated on either side.

> I'd also rather not have a new flag/setting for this, but if we set the token on the client side it could limit servers in regards how they can group notifications. Without this, they could use many different tokens, with this, they'd have to assign the same token for all progress relevant for initialize.

yeah, that does seem like an argument in favor of a setting, albeit one that we default to true. I don't think you could avoid this by using a UUID either.

mfussenegger:
> while the server can generate one, they're not required by the protocol to generate one. My interpretation is that the progress token can be generated on either side.

Yes, either side can. But a client _not_ sending the token isn't a opt-out of progress reports if the `workDoneProgress` capability is set to `true`, it just means the server can decide on the tokens.
And having a fallback on the server side to generate a token (or many) if the client request doesn't provide one, and if the capability is set shouldn't have any negative impact for other clients or servers.
This is afaik also what most other servers do.

rudiejd:
>And having a fallback on the server side to generate a token (or many) if the client request doesn't provide one, and if the capability is set shouldn't have any negative impact for other clients or servers.
This is afaik also what most other servers do.

Maybe this is something that needs to be clarified with the spec. To me, it seems that it's unclear where the responsibility of creating the token lies

mfussenegger:
ok re-reading the spec again, specifically these two parts:

Under client initiated:
> There is no specific client capability signaling whether a client will send a progress token per request. The reason for this is that this is in many clients not a static aspect and might even change for every request instance for the same request type. So the capability is signal on every request instance by the presence of a workDoneToken property.

Server initiated:
> Servers can also initiate progress reporting using the window/workDoneProgress/create request. This is useful if the server needs to report progress outside of a request (for example the server needs to re-index a database).


So I think it would be fine to just always set a token on initialize, without any configuration or flags. Servers can still initiate new progress reports that are independent of the request.

mfussenegger:
Opened https://github.com/neovim/neovim/pull/28182

rudiejd:
thank you! I'm on vacation this week

PR #28080, opened 2024-03-28
fix(intro): link showing intro to state at start

echasnovski:
Problem: Current behavior of stateful intro message is too persistent.
  For example, it is still drawn if new empty buffer is shown in current
  window (either by explicitly setting it or after `tabnew`). Although
  the buffer is empty, the act of it being shown should be made visible.

Solution: Make intro message persist if all is true:
    - Current buffer is the same as it was just after start, i.e. empty
      nameless with initial handle (set to 1).
    - Current window is the same as it was just after start, i.e. single
      non-floating with initial handle (set to 1000).

### DIFF
--- a/src/nvim/version.c
+++ b/src/nvim/version.c
@@ -2715,9 +2715,11 @@ void list_version(void)
 /// Whether it still is not too late to show an intro message
 bool may_show_intro(void)
 {
-  return (buf_is_empty(firstwin->w_buffer)
-          && (firstwin->w_buffer->b_fname == NULL)
-          && one_window(firstwin)
+  return (buf_is_empty(curbuf)
+          && (curbuf->b_fname == NULL)
+          && (curbuf->handle == 1)
+          && (curwin->handle == LOWEST_WIN_ID)
+          && one_window(curwin)
           && (vim_strchr(p_shm, SHM_INTRO) == NULL));
 }
 
--- a/test/functional/ui/messages_spec.lua
+++ b/test/functional/ui/messages_spec.lua
@@ -1630,25 +1630,25 @@ describe('ui/ext_messages', function()
       {1:~                }type  :help iccf{5:<Enter>}       for information {1:                 }|
       {1:~                                                                               }|*5
     ]]
+    local showmode = { { '-- INSERT --', 3 } }
     screen:expect(introscreen)
 
     -- <c-l> (same as :mode) does _not_ clear intro message
     feed('<c-l>i')
-    screen:expect { grid = introscreen, showmode = { { '-- INSERT --', 3 } } }
+    screen:expect { grid = introscreen, showmode = showmode }
 
-    -- opening a float also does not
-    local win = api.nvim_open_win(api.nvim_create_buf(false, false), true, {
+    -- opening a float without focus also does not
+    local win = api.nvim_open_win(api.nvim_create_buf(false, false), false, {
       relative = 'editor',
       height = 1,
       width = 5,
       row = 1,
       col = 5,
     })
-    feed('float<esc><c-l>')
     screen:expect {
       grid = [[
-                                                                                      |
-      {1:~    }{8:floa^t}{1:                                                                      }|
+      ^                                                                                |
+      {1:~    }{8:     }{1:                                                                      }|
       {1:~                                                                               }|*3
       {MATCH:.*}|
       {1:~                                                                               }|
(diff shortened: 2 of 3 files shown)


src/nvim/version.c
```diff
@@ -2715,8 +2715,9 @@ void list_version(void)
 /// Whether it still is not too late to show an intro message
 bool may_show_intro(void)
 {
-  return (buf_is_empty(firstwin->w_buffer)
-          && (firstwin->w_buffer->b_fname == NULL)
+  return (buf_is_empty(curbuf)
+          && (curbuf->b_fname == NULL)
+          && (curbuf->handle == 1)
```
zeertzjq, line 2720: This will show intro if buffer 1 is shown in a floating window that has been entered, regardless of what the buffer in the non-floating window is.

echasnovski: Good point.
I guess the proper approach would be to also clear intro if current window is not the one as it was during start? Adding `&& (curwin->handle == 1000)` condition works. Is this an appropriate solution here?

echasnovski: This should be resolved now.

src/nvim/version.c
```diff
@@ -2715,8 +2715,10 @@ void list_version(void)
...
+          && (curwin->handle == 1000)
```
luukvbaal, line 2721:
```suggestion
          && (curwin->handle == LOWEST_WIN_ID)
```

echasnovski: Thanks!

src/nvim/version.c
```diff
@@ -2715,8 +2715,10 @@ void list_version(void)
 /// Whether it still is not too late to show an intro message
 bool may_show_intro(void)
 {
-  return (buf_is_empty(firstwin->w_buffer)
-          && (firstwin->w_buffer->b_fname == NULL)
+  return (buf_is_empty(curbuf)
+          && (curbuf->b_fname == NULL)
+          && (curbuf->handle == 1)
+          && (curwin->handle == LOWEST_WIN_ID)
```
zeertzjq, line 2721: This will show intro if the window 1000 was split, then made into a floating window using `nvim_win_set_config`.

echasnovski: Realized that adding `&& (!curwin->w_floating)` is working here and is aligned with the description of conditions ("...single non-floating...").

*Now* I hope there is no (reasonable) edge cases?

test/functional/ui/messages_spec.lua
```diff
@@ -1715,6 +1717,56 @@ describe('ui/ext_messages', function()
...
+    screen:expect {
+      grid = [[
+                                                                                      |
+      {1:~    }{8:^     }{1:                                                                      }|
+      {1:~                                                                               }|*22
+    ]],
+    }
+  end)
+
+  it('clears intro screen when initial window is converted to be floating', function()
```
luukvbaal, line 1754: This one also passes without the added `w_floating` condition. Because `may_show_floating()` is called before the last `nvim_set_current_win()`. This makes it more effective:
```diff
--- a/test/functional/ui/messages_spec.lua
+++ b/test/functional/ui/messages_spec.lua
@@ -1752,11 +1752,13 @@ describe('ui/ext_messages', function()
   end)

   it('clears intro screen when initial window is converted to be floating', function()
-    local init_win_id = api.nvim_get_current_win()
-    command('split')
-    local win_opts = { relative = 'editor', height = 1, width = 5, row = 1, col = 5 }
-    api.nvim_win_set_config(init_win_id, win_opts)
-    api.nvim_set_current_win(init_win_id)
+    exec_lua([[
+      local init_win_id = vim.api.nvim_get_current_win()
+      vim.cmd('split')
+      local win_opts = { relative = 'editor', height = 1, width = 5, row = 1, col = 5 }
+      vim.api.nvim_win_set_config(init_win_id, win_opts)
+      vim.api.nvim_set_current_win(init_win_id)
+    ]])
     screen:expect {
       grid = [[
```

Didn't check the other tests.

echasnovski: Ah, indeed, thanks!

I tried to make test failing without the `w_floating` check, but decided that manual check and this kind of test was enough. The `exec_lua()` approach is much better, obviously.

src/nvim/version.c
```diff
@@ -2715,8 +2715,11 @@ void list_version(void)
...
+          && (!curwin->w_floating)
           && one_window(firstwin)
```
zeertzjq, line 2723: This is equivalent to
```suggestion
          && one_window(curwin)
```

zeertzjq (approved):

PR #28081, opened 2024-03-28
feature(ui): indicate margins for the area used by win_viewport

bfredl:
Problem: using win_viewport for implementing smooth scrolling in an external UI might run into problems when winbar is used, as there is no indication that the entire grid is not used for scrolled buffer text.

Solution: add win_viewport_margins event. reserve possibility for margins in every direction, even though currently only top is used for winbar.

Addresses a part of #23610, in particular https://github.com/neovim/neovim/issues/23610#issuecomment-1868216660

### DIFF
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -656,6 +656,11 @@ tabs.
 	the screen and copy it to the viewport destination once `win_viewport`
 	is received.
 
+["win_viewport_margins", grid, win, top, bottom, left, right] ~
+	Indicates the margins of a window grid which are _not_ part of the
+	viewport as indicated by the `win_viewport` event. This happens
+	e.g. in the presence of 'winbar' and floating window borders.
+
 ["win_extmark", grid, win, ns_id, mark_id, row, col] ~
 	Updates the position of an extmark which is currently visible in a
 	window. Only emitted if the mark has the `ui_watched` attribute.
--- a/src/nvim/api/ui_events.in.h
+++ b/src/nvim/api/ui_events.in.h
@@ -118,6 +118,10 @@ void win_viewport(Integer grid, Window win, Integer topline, Integer botline, In
                   Integer curcol, Integer line_count, Integer scroll_delta)
   FUNC_API_SINCE(7) FUNC_API_CLIENT_IGNORE;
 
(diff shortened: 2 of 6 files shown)


Kethku:
This seems like a good step. Could it also be extended to specify margins when floating window borders are used or should that be done in a subsequent PR?

bfredl:
yes, it should. will add it here.

bfredl:
@Kethku added handling of float borders.

Kethku:
Awesome. Let me look into making a local build and integrating it in neovide just to double check this works how I think it does.

Kethku:
I struggled to get everything building from source on my windows machine. On inspection though I think that this PR will solve the gui smooth scrolling portion of #23610 though so don't block on me. Once this merges I will test it via the nightly build

akiyosi:
I would like to test this new event.
I understand that top can be shrunk by winbar, but I would like to have values other than top greater than 1 (i.e., a grid where, for example, grid is 2 and top, bottom, left, and right are all 1). Is it possible to create such a grid?

bfredl:
@akiyosi yes using a floating window with both borders and winbar

Kethku:
One gotcha that wasn't obvious at first is that the viewport margins are sometimes sent before the window position has been reported to the gui. @bfedrl Imho this was somewhat unexpected and should be represented in the documentation or delayed until the window is positioned/sized.

bfredl:
Are they both sent on the same batch before "flush" or does this happen with events in separate batches?

The idea is that window state is fully self consistent at points where "flush" is sent. As long as that is maintained, I don't see one order being more natural than the other.

Kethku:
I believe they are. Makes sense!

Kethku:
> Are they both sent on the same batch before "flush" or does this happen with events in separate batches?
> 
> The idea is that window state is fully self consistent at points where "flush" is sent. As long as that is maintained, I don't see one order being more natural than the other.

Ok looks like the initial ViewportMargin events are not sent in the same batch as the resize or position events which are used to initialize the window. This isn't a huge deal since the initial events are all empty, but is still unexpected

zeertzjq:
This should probably be in `grid_line` instead, because virtual line with `virt_lines_leftcol` do not have number column.

PR #28083, opened 2024-03-28
feat: allow opting in to builtin clipboard providers

### LINKED ISSUE #27698: FEATURE: An option to prioritize `:h clipboard-tool` over OSC 52 clipboard.
### Problem

As it says in `:h clipboard-osc52`, OSC 52 clipboard kicks in by default when I meet all of the following criteria.

```txt
  • Nvim is running in the |TUI|
  • |g:clipboard| is unset
  • 'clipboard' is not set to "unnamed" or "unnamedplus"
  • $SSH_TTY is set
```

I unfortunately meet all the criteria but I want to go through the clipboard detection listed in `:h clipboard-tool`.
My terminal does not support osc52 and I hop between OSs so I sometimes use `tmux` and sometimes `xsel` / `xlip` so I cannot hardcode something to `g:clipboard` as well.

### Expected behavior

Would it be possible to move ocs52 clipboard to the end of the `:h clipboard-tool` list, or provide an option to disable it manually?

I would appreciate having the clipboard-tools prioritize over osc52 as I said in the title, but simply having the option to disable it would also be helpful.

Considering that it already verifies the `SSH_TTY` envvar, could we add a custom envvar to disable it?

cc @gpanders

dundargoc:
Setting `vim.g.clipboard = false` will use the builtin clipboard
providers.

Closes https://github.com/neovim/neovim/issues/27698.

### DIFF
--- a/runtime/autoload/provider/clipboard.vim
+++ b/runtime/autoload/provider/clipboard.vim
@@ -67,7 +67,8 @@ function! provider#clipboard#Error() abort
 endfunction
 
 function! provider#clipboard#Executable() abort
-  if exists('g:clipboard')
+  " Setting g:clipboard to v:false explicitly opts-in to using the "builtin" clipboard providers below
+  if exists('g:clipboard') && g:clipboard isnot# v:false
     if type({}) isnot# type(g:clipboard)
           \ || type({}) isnot# type(get(g:clipboard, 'copy', v:null))
           \ || type({}) isnot# type(get(g:clipboard, 'paste', v:null))


runtime/autoload/provider/clipboard.vim
```diff
@@ -67,7 +67,8 @@ function! provider#clipboard#Error() abort
 endfunction
 
 function! provider#clipboard#Executable() abort
-  if exists('g:clipboard')
+  " Setting g:clipboard to v:false explicitly opts-in to using the "builtin" clipboard providers below
+  if exists('g:clipboard') && g:clipboard != v:false
```
gpanders, line 71: If `g:clipboard` is defined then this gives an error because we can't compare a Dictionary to boolean.

cc @seandewar

clason: `:h type()`?

gpanders, line 71:
```suggestion
  if exists('g:clipboard') && isnot# v:false
```

gpanders (approved):

PR #28086, opened 2024-03-28
fix(tui): don't use DECRQSS in screen or tmux

### LINKED ISSUE #28082: Tmux server exits when starting neovim
### Problem

As soon as I try to start neovim,  neovim fails to start and the tmux  server exits.

### Steps to reproduce

open a tmux session and start neovim.

The issues first appears with commit 4ee9e58056... from Mar 28.  The previous commit 7168000. didn't have this problem.

### Expected behavior

I expect neovim to start without crashing.

### Neovim version (nvim -v)

commit 4ee9e58056... 

-

### Operating system/version

macOS 14.4.1

### Terminal name/version

item 3.4.23

screen-256color

### Installation

build from repo

zeertzjq:
They behave strangely when receiving it.

Fix #28082

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -370,6 +370,8 @@ static void terminfo_start(TUIData *tui)
                               : (konsole ? 1 : 0);
   bool wezterm = strequal(termprg, "WezTerm");
   const char *weztermv = wezterm ? os_getenv("TERM_PROGRAM_VERSION") : NULL;
+  bool screen = terminfo_is_term_family(term, "screen");
+  bool tmux = terminfo_is_term_family(term, "tmux") || !!os_getenv("TMUX");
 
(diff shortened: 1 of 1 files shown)


wincent:
So while this fix avoids a crash in tmux, it also means that extended underline (eg. undercurl, coloring) won't work in Neovim in tmux. tmux itself [fixed the bug in May](https://github.com/tmux/tmux/commit/aa17f0e0c1c8b3f1d6fc8617613c74f07de66fae). Do we have a plan for somehow allowing Neovim users to use extended underline inside newer versions of tmux?

zeertzjq:
tmux does not support DECRQSS, so this PR doesn't not prevent extended underline from working in tmux. If terminfo for tmux contains the correct `Smulx` value (which is already the case in the latest ncurses version), extended underline will work in tmux.

PR #28090, opened 2024-03-29
feat: add filetypes

Freed-Wu:
I'll add some reviews to explain some filetypes.


lucario387:
This looks like 10 Vim PRs :)

clason:
Yes. New filetypes **must** go to Vim first; we will port the logic as a vim-patch. Custom additions are **not** wanted.

PR #28100, opened 2024-03-29
test(normal): port legacy normal test for page scrolling

luukvbaal:


test/functional/legacy/normal_spec.lua
```diff
@@ -29,6 +35,71 @@ describe('normal', function()
...
+    feed('yy100pG<C-L>')
+    api.nvim_buf_set_extmark(0, ns, 100, 0, { virt_lines = { { { '---', 'IncSearch' } } } })
+    screen:expect({
+      grid = [[
+        foobar one two three                    |*17
+        ^foobar one two three                    |
+                                                |
+      ]],
+    })
+    feed('<C-F>')
```
luukvbaal, line 85: Scrolling beyond the end of the buffer is different from the old behavior and slightly unexpected for (half)-page scrolling I think (more so than for CTRL-E/Y). I'll check if we can maintain the old behavior while scrolling down to reveal virtual lines.

zeertzjq (approved):

PR #26071, opened 2023-11-16
feat(column): add :redrawstatuscolumn to rebuild 'statuscolumn'

luukvbaal:
Problem:  No obvious way to force a 'statuscolumn' rebuild.
Solution: Let :redraw force a 'statuscolumn' rebuild.

PR #28101, opened 2024-03-29
feat(api): add nvim__redraw for more granular redrawing

### LINKED ISSUE #20793: cursor doesn't redraw with getchar

luukvbaal:
Experimental and subject to future changes.
Add a way to redraw certain elements that are not redrawn while Nvim is waiting
for input, or currently have no API to do so. This API covers all that can be
done with the :redraw* commands, in addition to the following new features:
- Immediately move the cursor to a (non-current) window.
- Target a specific window or buffer to mark for redraw.
- Mark a buffer range for redraw (replaces nvim__buf_redraw_range()).
- Redraw the 'statuscolumn'.

Resolve #20793
Closes #26071

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -1566,6 +1566,35 @@ nvim__invalidate_glyph_cache()                *nvim__invalidate_glyph_cache()*
     For testing. The condition in schar_cache_clear_if_full is hard to reach,
     so this function can be used to force a cache clear in a test.
 
+nvim__redraw({opts})                                          *nvim__redraw()*
+    EXPERIMENTAL: this API may change in the future.
+
+    Instruct Nvim to redraw various components.
+
+    Parameters: ~
+      • {opts}  Optional parameters.
+                • win: Target a specific |window-ID| as described below.
+                • buf: Target a specific buffer number as described below.
+                • flush: Update the screen with pending updates.
+                • valid: When present mark `win`, `buf`, or all windows for
+                  redraw. When `true`, only redraw changed lines (useful for
+                  decoration providers). When `false`, forcefully redraw.
+                • range: Redraw a range in `buf`, the buffer in `win` or the
+                  current buffer (useful for decoration providers). Expects a
+                  tuple `[first, last]` with the first and last line number of
+                  the range, 0-based end-exclusive |api-indexing|.
+                • cursor: Immediately update cursor position on the screen in
+                  `win` or the current window.
+                • statuscolumn: Redraw the 'statuscolumn' in `buf`, `win` or
+                  all windows.
+                • statusline: Redraw the 'statusline' in `buf`, `win` or all
+                  windows.
+                • winbar: Redraw the 'winbar' in `buf`, `win` or all windows.
+                • tabline: Redraw the 'tabline'.
+
+    See also: ~
+      • |:redraw|
+
 nvim__stats()                                                  *nvim__stats()*
     Gets internal stats.
 
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -50,6 +50,7 @@
 #include "nvim/mapping.h"
 #include "nvim/mark.h"
 #include "nvim/mark_defs.h"
+#include "nvim/math.h"
 #include "nvim/mbyte.h"
 #include "nvim/memline.h"
 #include "nvim/memory.h"
@@ -2304,4 +2305,159 @@ Dictionary nvim__complete_set(Integer index, Dict(complete_set) *opts, Arena *ar
     }
   }
   return rv;
+}
+
+static void redraw_status(win_T *wp, Dict(redraw) *opts, bool *flush)
+{
+  if (opts->statuscolumn && *wp->w_p_stc != NUL) {
+    wp->w_nrwidth_line_count = 0;
+    changed_window_setting(wp);
+  }
+  win_grid_alloc(wp);
+
+  // Flush later in case winbar was just hidden or shown for the first time, or
+  // statuscolumn is being drawn.
+  if (wp->w_lines_valid == 0) {
+    *flush = true;
+  }
+
+  // Mark for redraw in case flush will happen, otherwise redraw now.
+  if (*flush && (opts->statusline || opts->winbar)) {
+    wp->w_redr_status = true;
+  } else if (opts->statusline || opts->winbar) {
+    win_check_ns_hl(wp);
+    if (opts->winbar) {
+      win_redr_winbar(wp);
+    }
+    if (opts->statusline) {
+      win_redr_status(wp);
+    }
+    win_check_ns_hl(NULL);
+  }
+}
+
+/// EXPERIMENTAL: this API may change in the future.
+///
+/// Instruct Nvim to redraw various components.
+///
+/// @see |:redraw|
+///
+/// @param opts  Optional parameters.
+///               - win: Target a specific |window-ID| as described below.
+///               - buf: Target a specific buffer number as described below.
+///               - flush: Update the screen with pending updates.
+///               - valid: When present mark `win`, `buf`, or all windows for
+///                 redraw. When `true`, only redraw changed lines (useful for
+///                 decoration providers). When `false`, forcefully redraw.
+///               - range: Redraw a range in `buf`, the buffer in `win` or the
+///                 current buffer (useful for decoration providers). Expects a
+///                 tuple `[first, last]` with the first and last line number
+///                 of the range, 0-based end-exclusive |api-indexing|.
+///               - cursor: Immediately update cursor position on the screen in
+///                 `win` or the current window.
+///               - statuscolumn: Redraw the 'statuscolumn' in `buf`, `win` or
+///                 all windows.
+///               - statusline: Redraw the 'statusline' in `buf`, `win` or all
+///                 windows.
+///               - winbar: Redraw the 'winbar' in `buf`, `win` or all windows.
+///               - tabline: Redraw the 'tabline'.
+void nvim__redraw(Dict(redraw) *opts, Error *err)
+  FUNC_API_SINCE(12)
+{
+  win_T *win = NULL;
+  buf_T *buf = NULL;
+
+  if (HAS_KEY(opts, redraw, win)) {
+    win = find_window_by_handle(opts->win, err);
+    if (ERROR_SET(err)) {
+      return;
+    }
+  }
+
+  if (HAS_KEY(opts, redraw, buf)) {
+    VALIDATE(win == NULL, "%s", "cannot use both 'buf' and 'win'", {
+      return;
+    });
+    buf = find_buffer_by_handle(opts->buf, err);
+    if (ERROR_SET(err)) {
+      return;
+    }
+  }
+
+  int count = (win != NULL) + (buf != NULL);
+  VALIDATE(popcount(opts->is_set__redraw_) > count, "%s", "at least one action required", {
+    return;
+  });
+
+  if (HAS_KEY(opts, redraw, valid)) {
+    // UPD_VALID redraw type does not actually do anything on it's own. Setting
+    // it here without scrolling or changing buffer text seems pointless but
+    // the expectation is that this may be called by decoration providers whose
+    // "on_win" callback may set "w_redr_top/bot".
+    int type = opts->valid ? UPD_VALID : UPD_NOT_VALID;
+    if (win != NULL) {
+      redraw_later(win, type);
+    } else if (buf != NULL) {
+      redraw_buf_later(buf, type);
+    } else {
+      redraw_all_later(type);
+    }
+  }
+
+  if (HAS_KEY(opts, redraw, range)) {
+    VALIDATE(kv_size(opts->range) == 2
+             && kv_A(opts->range, 0).type == kObjectTypeInteger
+             && kv_A(opts->range, 1).type == kObjectTypeInteger
+             && kv_A(opts->range, 0).data.integer >= 0
+             && kv_A(opts->range, 1).data.integer >= -1,
+             "%s", "Invalid 'range': Expected 2-tuple of Integers", {
+      return;
+    });
+    linenr_T first = (linenr_T)kv_A(opts->range, 0).data.integer + 1;
+    linenr_T last = (linenr_T)kv_A(opts->range, 1).data.integer;
+    if (last < 0) {
+      last = buf->b_ml.ml_line_count;
+    }
+    redraw_buf_range_later(win ? win->w_buffer : (buf ? buf : curbuf), first, last);
+  }
+
+  if (opts->cursor) {
+    setcursor_mayforce(win ? win : curwin, true);
+  }
+
+  bool flush = opts->flush;
+  if (opts->tabline) {
+    // Flush later in case tabline was just hidden or shown for the first time.
+    if (redraw_tabline && firstwin->w_lines_valid == 0) {
+      flush = true;
+    } else {
+      draw_tabline();
+    }
+  }
+
+  bool save_lz = p_lz;
+  int save_rd = RedrawingDisabled;
+  RedrawingDisabled = 0;
+  p_lz = false;
+  if (opts->statuscolumn || opts->statusline || opts->winbar) {
+    if (win == NULL) {
+      FOR_ALL_WINDOWS_IN_TAB(wp, curtab) {
+        if (buf == NULL || wp->w_buffer == buf) {
+          redraw_status(wp, opts, &flush);
+        }
+      }
+    } else {
+      redraw_status(win, opts, &flush);
+    }
+  }
+
+  // Flush pending screen updates if "flush" or "clear" is true, or when
+  // redrawing a status component may have changed the grid dimensions.
+  if (flush && !cmdpreview) {
+    update_screen();
+  }
+  ui_flush();
+
+  RedrawingDisabled = save_rd;
+  p_lz = save_lz;
 }
--- a/test/functional/api/vim_spec.lua
+++ b/test/functional/api/vim_spec.lua
@@ -5016,4 +5016,216 @@ describe('API', function()
       eq(false, exec_lua('return _G.success'))
     end)
   end)
+
+  it('nvim__redraw', function()
+    local screen = Screen.new(60, 5)
+    screen:attach()
+    local win = api.nvim_get_current_win()
+    eq('at least one action required', pcall_err(api.nvim__redraw, {}))
+    eq('at least one action required', pcall_err(api.nvim__redraw, { buf = 0 }))
+    eq('at least one action required', pcall_err(api.nvim__redraw, { win = 0 }))
+    eq("cannot use both 'buf' and 'win'", pcall_err(api.nvim__redraw, { buf = 0, win = 0 }))
+    feed(':echo getchar()<CR>')
+    fn.setline(1, 'foobar')
+    command('vnew')
+    fn.setline(1, 'foobaz')
+    -- Can flush pending screen updates
+    api.nvim__redraw({ flush = true })
+    screen:expect({
+      grid = [[
+        foobaz                        │foobar                       |
+        {1:~                             }│{1:~                            }|*2
+        {3:[No Name] [+]                  }{2:[No Name] [+]                }|
+        ^:echo getchar()                                             |
+      ]],
+    })
+    -- Can update the grid cursor position #20793
+    api.nvim__redraw({ cursor = true })
+    screen:expect({
+      grid = [[
+        ^foobaz                        │foobar                       |
+        {1:~                             }│{1:~                            }|*2
+        {3:[No Name] [+]                  }{2:[No Name] [+]                }|
+        :echo getchar()                                             |
+      ]],
+    })
+    -- Also in non-current window
+    api.nvim__redraw({ cursor = true, win = win })
+    screen:expect({
+      grid = [[
+        foobaz                        │^foobar                       |
+        {1:~                             }│{1:~                            }|*2
+        {3:[No Name] [+]                  }{2:[No Name] [+]                }|
+        :echo getchar()                                             |
+      ]],
+    })
+    -- Can update the 'statusline' in a single window
+    api.nvim_set_option_value('statusline', 'statusline1', { win = 0 })
+    api.nvim_set_option_value('statusline', 'statusline2', { win = win })
+    api.nvim__redraw({ cursor = true, win = 0, statusline = true })
+    screen:expect({
+      grid = [[
+        ^foobaz                        │foobar                       |
+        {1:~                             }│{1:~                            }|*2
+        {3:statusline1                    }{2:[No Name] [+]                }|
+        :echo getchar()                                             |
+      ]],
+    })
+    api.nvim__redraw({ win = win, statusline = true })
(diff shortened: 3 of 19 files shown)


runtime/doc/api.txt
```diff
@@ -1307,6 +1307,32 @@ nvim_put({lines}, {type}, {after}, {follow})                      *nvim_put()*
                   |P|).
       • {follow}  If true place cursor at end of inserted text.
 
+nvim_redraw({opts})                                            *nvim_redraw()*
+    Instruct Nvim to redraw various elements.
+
+    Parameters: ~
+      • {opts}  Optional parameters.
+                • `update_screen`: Immediately update the screen. Useful when
```
clason, line 1315:
```suggestion
                • `screen`: Immediately update the screen. Useful when
```
update (=redraw) is implied

clason: In general, we should try to be consistent about "update" vs. "redraw" here (or, if there is a distinction, explain it).

luukvbaal: The implication of `update_screen` is that the actual buffer content in windows is redrawn, for which `update_screen()` is the only entry point internally. As opposed the various other "widgets" associated with a window which can also be targeted in this API and redrawn without redrawing the entire window, but would also be redrawn when `update_screen()` is called and they were pending for redraw.

Considering that, maybe this can be called `text` or `viewport` as it is referred to under `:h window`?

clason: Thanks for the explanation. So `flush` is (to me!) closest in meaning (but still needs an explanation at this point in the documentation, or at least a taglink), but I defer to @bfredl

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,169 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+///                 and `clear` updates, current buffer when 0 or omitted.
+void nvim_redraw(Dict(redraw) *opts, Error *err)
+  FUNC_API_SINCE(12)
+{
+  win_T *win = NULL;
+  buf_T *buf = NULL;
+  bool save_lz = p_lz;
+  int save_rd = RedrawingDisabled;
+  RedrawingDisabled = 0;
+  p_lz = false;
```
bfredl, line 2464: I think it should be enough to do these around `update_screen()` inside its if-statement.

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,169 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+    if (ERROR_SET(err)) {
+      goto theend;
+    }
+  }
+
+  if (opts->clear) {
+    if (win != NULL) {
+      redraw_later(win, UPD_NOT_VALID);
+    } else if (buf != NULL) {
+      redraw_buf_later(buf, UPD_NOT_VALID);
```
bfredl, line 2484: if we treat `win` and `buf` as mutually exclusive, we should throw an error when both `buf` and `win` are supplied.

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,169 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+  if (HAS_KEY(opts, redraw, range)) {
+    VALIDATE(kv_size(opts->range) == 2
+             && kv_A(opts->range, 0).type == kObjectTypeInteger
+             && kv_A(opts->range, 1).type == kObjectTypeInteger,
+             "%s", "Invalid 'range': expected 2-tuple of Lua numbers", {
+      goto theend;
+    });
+    Integer first = kv_A(opts->range, 0).data.integer;
+    Integer last = kv_A(opts->range, 1).data.integer;
+    nvim__buf_redraw_range(buf ? buf->handle : curbuf->handle, first, last, err);
```
bfredl, line 2499: then, if `win` is supplied but not `buf`, this either needs to throw an error or use the buffer of that window (ideally it would imply only redrawing the range for the specific window, but doing it for the buffer in the window is good enough).

src/nvim/api/vim.c
luukvbaal: `win_redr_status()` checks `!redrawing()` as well.

src/nvim/api/vim.c
luukvbaal: Yes, I considered exposing `win_upate()` to be able to target specific windows but I'm not sure that's a good idea. It would only be effective when `redrawing() == false` anyways, otherwise the rest of the windows would be updated at the next `update_screen()` anyways.

runtime/doc/api.txt
```diff
@@ -1307,6 +1307,34 @@ nvim_put({lines}, {type}, {after}, {follow})                      *nvim_put()*
...
+                  providers. Tuple `[first, last]`:
+                  • first: first line number of the range
+                  • last: last line number of the range
+                • cursor: Immediately update cursor position on the screen.
+                • status: Redraws (a combination of) the following status
+                  components (array of one or more strings):
+                  • "bar": |'winbar'|
+                  • "column": |'statuscolumn'|
+                  • "line": |'statusline'|
+                  • "tab": |'tabline'|
```
zeertzjq, line 1329: I don't think adding these new abbreviations is a good idea. I think it's better to use existing terminologies.

zeertzjq: I guess it's better to put them directly in `opts`, so that there will be fewer thing to decide when something new need to be added.

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,153 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+
+  if (status & kTabLine) {
+    // Just hid or showing tabline for the first time so flush later.
+    if (redraw_tabline && firstwin->w_lines_valid == 0) {
+      flush = true;
+    } else {
+      draw_tabline();
+    }
+  }
+  if (status != kTabLine) {
```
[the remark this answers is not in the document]
luukvbaal: No it's intentionally `!=`. It should still redraw the others if they are supplied.

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,153 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+///                 all windows.
+///               - winbar: Redraw the |'winbar'| in `buf`, `win` or all
+///                 windows.
+///               - tabline: Redraw the |'tabline'|.
+///               - win: Target a specific |window-ID| for `range`, `status`
+///                 `cursor`, and `clear` updates. Current window when 0, all
+///                 windows if omitted and applicable. Mutually exclusive
+///                 with `buf`.
+///               - buf: Target a specific buffer number for `range`, `status`,
+///                 and `clear` updates, current buffer when 0 or omitted.
```
zeertzjq, line 2453: Maybe it's better to just mention here that these two are used a target a specific window/buffer for other optional parameters (without mentioning their names), and make it clear in each optional parameter how it uses `buf`/`win`.

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,153 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+///               - buf: Target a specific buffer number for various
+///                 components.
+///               - flush: Update the screen with pending updates, even if
+///                 Nvim is waiting for input and would normally not do so.
+///               - clear: Forcefully redraw `win`, all windows containing
+///                 `buf`, or all windows.
+///               - range: Buffer range to redraw, useful for decoration
+///                 providers. Tuple `[first, last]`:
+///                   - first: First line number of the range.
+///                   - last: Last line number of the range.
```
zeertzjq, line 2442: There seems to be some line wrapping problems with docgen, plus there is no need to mention "line number of the range" twice
```suggestion
///                 providers. Tuple `[first, last]` are the first and last
///                 line numbers of the range, respectively.
```

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,151 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+    } else {
+      redraw_all_later(UPD_NOT_VALID);
+    }
+  }
+
+  if (HAS_KEY(opts, redraw, range)) {
+    VALIDATE(kv_size(opts->range) == 2
+             && kv_A(opts->range, 0).type == kObjectTypeInteger
+             && kv_A(opts->range, 1).type == kObjectTypeInteger,
+             "%s", "Invalid 'range': expected 2-tuple of Lua numbers", {
```
zeertzjq, line 2493:
```suggestion
             "%s", "Invalid 'range': expected 2-tuple of Integers", {
```

luukvbaal: Seemed more logical to me as well but then this should be updated first I think:
https://github.com/neovim/neovim/blob/8697f3274b240a5a00774dba90329c00c4e914ab/src/nvim/lua/converter.c#L829

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,151 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+
+  if (HAS_KEY(opts, redraw, win)) {
+    win = find_window_by_handle(opts->win, err);
+    if (ERROR_SET(err)) {
+      goto theend;
+    }
+  }
+
+  if (HAS_KEY(opts, redraw, buf)) {
+    VALIDATE(win == NULL, "%s", "Invalid 'buf': already specified 'win'.", {
```
zeertzjq, line 2470: API error messages usually don't end with a period
```suggestion
    VALIDATE(win == NULL, "%s", "Invalid 'buf': already specified 'win'", {
```

zeertzjq:
It seems that for some reason the Perl tests become very flaky on this PR. I guess that may have something to do with the added API metadata.

luukvbaal:
Is that our problem or should the provider handle it?

zeertzjq:
Not sure, but anyway it's out of the scope of this PR.

src/nvim/api/deprecated.c
```diff
@@ -697,3 +698,18 @@ static void set_option_to(uint64_t channel_id, void *to, OptReqScope req_scope,
     set_option_value_for(name.data, opt_idx, optval, opt_flags, req_scope, to, err);
   });
 }
+
+/// @nodoc
+/// @deprecated
+void nvim__buf_redraw_range(Buffer buffer, Integer first, Integer last, Error *err)
```
justinmk, line 704: this is a private api, can we just remove it now?

justinmk (approved):

src/nvim/api/deprecated.c
luukvbaal: Fine by me but considering that it is used in the wild a single release cycle with `@deprecated` could be good?

justinmk: In general we should be pretty strict about not making efforts (deprecation cycles) for APIs that were marked as private. I see some important plugins in there, though I think it will be easy for them to migrate...

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,140 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+             && kv_A(opts->range, 1).type == kObjectTypeInteger,
+             "%s", "Invalid 'range': Expected 2-tuple of Integers", {
+      return;
+    });
+    linenr_T first = (linenr_T)kv_A(opts->range, 0).data.integer;
+    linenr_T last = (linenr_T)kv_A(opts->range, 1).data.integer;
+    if (last < 0) {
+      last = buf->b_ml.ml_line_count;
+    }
+    redraw_buf_range_later(win ? win->w_buffer : (buf ? buf : curbuf), first + 1, last);
```
luukvbaal, line 2487: I think it should be reconsidered what `first` and `last` actually indicate, and how they are documented. It seems that in `nvim__buf_redraw_range` `first` was a 0-indexed row number, whereas `last` was a 1-indexed line number (why?). I'll go ahead and document it and update the internal calls like so:

```
- range: Redraw a range in `buf`, the buffer in `win` or
  the current buffer. Useful for decoration providers.
  Expects a tuple `[first, last]` with the first and last
  line number of the range, or [1, -1] to indicate the entire
  buffer.
```

luukvbaal: Ah yeah that makes more sense probably, so then they were both 0-indexed. But I think 1-indexed end-inclusive line numbers makes more sense for this API?

luukvbaal: I mean in a way it doesn't because the extmarks that necessitate this call also were placed 0-indexed (end inclusive). I just mean in general when talking about line numbers 1-indexed feels more natural but if this should be 0-indexed as well it just needs to be documented.

luukvbaal: It's also not entirely clear to me why this API is needed in the first place so that probably doesn't help. Don't the `nvim_buf_set_extmark()` calls already set `w_redraw_top/mod` through `decor_redraw()`? If this API is graduated the reasoning behind that should also probably be documented better (here)?

luukvbaal: I think all internal `nvim__buf_redraw_range(buf, 0, -1)` calls could/should be replaced by something like:

```diff
@@ -2425,8 +2425,9 @@ static void redraw_status(win_T *wp, Dict(redraw) *opts, bool *flush)
+///               - valid: When present and `true`, redraw the changed part of `win`,
+///                 all windows containing `buf`, or all windows. When `false`, redraw
+///                 the entire window.
 ///               - range: Redraw a range in `buf`, the buffer in `win` or the
 ///                 current buffer. Useful for decoration providers. Expects a
 ///                 tuple `[first, last]` with the first and last line number of
@@ -2462,13 +2463,14 @@ void nvim_redraw(Dict(redraw) *opts, Error *err)
     }
   }

-  if (opts->clear) {
+  if (HAS_KEY(opts, redraw, valid)) {
+    int type = opts->valid ? UPD_VALID : UPD_NOT_VALID;
     if (win != NULL) {
-      redraw_later(win, UPD_NOT_VALID);
+      redraw_later(win, type);
     } else if (buf != NULL) {
-      redraw_buf_later(buf, UPD_NOT_VALID);
+      redraw_buf_later(buf, type);
     } else {
-      redraw_all_later(UPD_NOT_VALID);
+      redraw_all_later(type);
     }
   }
```

Then, if placing ephemeral marks would call `decor_redraw()`, I don't think this `range` API would be needed.

luukvbaal: It also seems like the `nvim__buf_redraw_range(buf, 0, -1)` calls were "abusing" the private API. I will try to replace `clear` with `valid` in a way that I think should cover that use case.

src/nvim/api/vim.c
```diff
@@ -2389,3 +2389,144 @@ Dictionary nvim_complete_set(Integer index, Dict(complete_set) *opts, Arena *are
...
+    }
+    if (opts->statusline) {
+      win_redr_status(wp);
+    }
+    win_check_ns_hl(NULL);
+  }
+}
+
+/// Instruct Nvim to redraw various components.
+///
```
justinmk, line 2423:
```suggestion
///
/// @see |:redraw|
///
```

Let's also mention `nvim_redraw()` from the docs for :redraw/:redrawstatus/etc.

luukvbaal: Sure. Could add a line like this to each command

    See |nvim_redraw()| for a more granular alternative.

Or add some kind of header above the bunch of them as those commands are grouped.

justinmk: > Could add a line like this to each command

Just `See also |nvim_redraw()|.` on each command is the usual pattern. Headers are less effective for "discovery".

runtime/doc/api.txt
```diff
@@ -1307,6 +1307,33 @@ nvim_put({lines}, {type}, {after}, {follow})                      *nvim_put()*
                   |P|).
       • {follow}  If true place cursor at end of inserted text.
 
+nvim_redraw({opts})                                            *nvim_redraw()*
+    Instruct Nvim to redraw various components.
```
justinmk, line 1311: If {opts} is empty what does it do?
```suggestion
    Instruct Nvim to redraw various components. By default (i.e. if {opts} is empty), ...
```

src/nvim/api/deprecated.c
```diff
@@ -13,6 +13,7 @@
 #include "nvim/buffer_defs.h"
 #include "nvim/decoration.h"
 #include "nvim/decoration_defs.h"
+#include "nvim/drawscreen.h"
```
justinmk, line 16: is this still needed?

test/functional/api/vim_spec.lua
```diff
@@ -4995,4 +4995,212 @@ describe('API', function()
       eq(false, exec_lua('return _G.success'))
     end)
   end)
+
+  it('nvim_redraw', function()
```
justinmk, line 4999: is there a test for empty `opts` ? `nvim_redraw({})`

justinmk:
In order to unblock this and get it in the tree, shall we rename it to double-underscore `nvim__redraw`? That signals that it is experimental. Then we can finalize it in 0.11

luukvbaal:
Fine by me, updated. Not sure whether to keep the `various.txt` references. Referring to experimental API is a little weird but removing them risks forgetting them in the future.

clason:
Yeah, I'd avoid the links from `various.txt` for now.

justinmk:
It's marked as experimental, I don't see why we can't refer to it from other parts of the docs. If it were deprecated, that would be different, then we want to scrub the docs.

pidgeon777:
In my setup `nvim --version`:

```log
NVIM v0.10.0-dev-3057+g9e2f378b6
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info
```

and now it seems that `Neovim` is alot faster (writing in insert mode, copy pasting in normal mode and buffer switching/navigation). Could it be due to this merge?

I'm using Windows 11 Pro.

luukvbaal:
Uhm since this flushes the UI in addition to marking a buffer range for redraw, and it is used in the treesitter highlighter, an (unintended but wanted?) consequence could be that it made the editing experience feel more responsive (at the cost of more CPU cycles).

pidgeon777:
I have to say that it seems to have improved `Neovim` smoothness a lot, at least on Windows (I'm also using [Nvy](https://github.com/RMichelsen/Nvy) GUI).

Or, it could also be that this merge in conjunction with other recent ones (to master) improved the overall performance (I'm using the `Neovim` nightly version currently available).

src/nvim/api/vim.c
```diff
@@ -2305,3 +2306,158 @@ Dictionary nvim__complete_set(Integer index, Dict(complete_set) *opts, Arena *ar
...
+             && kv_A(opts->range, 1).type == kObjectTypeInteger
+             && kv_A(opts->range, 0).data.integer >= 0
+             && kv_A(opts->range, 1).data.integer >= -1,
+             "%s", "Invalid 'range': Expected 2-tuple of Integers", {
+      return;
+    });
+    linenr_T first = (linenr_T)kv_A(opts->range, 0).data.integer + 1;
+    linenr_T last = (linenr_T)kv_A(opts->range, 1).data.integer;
+    if (last < 0) {
+      last = buf->b_ml.ml_line_count;
```
zeertzjq, line 2419: This should use the same buffer as the buffer passed to the `redraw_buf_range_later()` below. And I don't think it should treat all negative value as the same.

luukvbaal: Good catch. Only -1 is a valid negative value, do you just mean that the if statement should be more explicit in that?

zeertzjq: Oops, I didn't see the `>= -1` check above. Then I think just changing this < 0` to `== -1` is enough.

topaxi:
Nightly started crashing for me when using a mouse click into neovim. After bisecting from yesterday to today, I found this PR/commit 7b14eb543d43344e2498335dc93a68d200469516 to be the cause.

I just get a segmentation fault and am not versed enough in C to be able to debug beyond that.

If I can provide anything that might help, let me know. I use quite a few plugins, and with a `--clean` run, the crash does not occur, so I assume this might be either a configuration issue or combined with a plugin.

I wasn't able to get anything out of any log.

zeertzjq:
It may be https://github.com/folke/flash.nvim/issues/333

topaxi:
Disabled `flash.nvim` and the crash still occurs, I might find time later today to enable/disable plugins one-by-one to find the offending one :)

zeertzjq:
Another possible cause is https://github.com/folke/noice.nvim/blob/0cbe3f88d038320bdbda3c4c5c95f43a13c3aa12/lua/noice/util/ffi.lua#L20

topaxi:
Good catch, your quick! That's the one!

clason:
PSA: If things crash, grep your plugin folder for `ffi`. If a plugin uses that (and many of Folke's do), then this is likely the fault and you should report it there. We have zero ABI compatibility guarantees, and _any_ commit may break FFI. Plugin authors should know that and react quickly.

justinmk:
coverity:

```
*** CID 493189:    (FORWARD_NULL)
/src/nvim/api/vim.c: 2419 in nvim__redraw()
2413                  "%s", "Invalid 'range': Expected 2-tuple of Integers", {
2414           return;
2415         });
2416         linenr_T first = (linenr_T)kv_A(opts->range, 0).data.integer + 1;
2417         linenr_T last = (linenr_T)kv_A(opts->range, 1).data.integer;
2418         if (last < 0) {
>>>     CID 493189:    (FORWARD_NULL)
>>>     Dereferencing null pointer "buf".
2419           last = buf->b_ml.ml_line_count;
2420         }
2421         redraw_buf_range_later(win ? win->w_buffer : (buf ? buf : curbuf), first, last);
2422       }
2423     
2424       if (opts->cursor) {
/src/nvim/api/vim.c: 2419 in nvim__redraw()
2413                  "%s", "Invalid 'range': Expected 2-tuple of Integers", {
2414           return;
2415         });
2416         linenr_T first = (linenr_T)kv_A(opts->range, 0).data.integer + 1;
2417         linenr_T last = (linenr_T)kv_A(opts->range, 1).data.integer;
2418         if (last < 0) {
>>>     CID 493189:    (FORWARD_NULL)
>>>     Dereferencing null pointer "buf".
2419           last = buf->b_ml.ml_line_count;
2420         }
2421         redraw_buf_range_later(win ? win->w_buffer : (buf ? buf : curbuf), first, last);
2422       }
2423     
2424       if (opts->cursor) {
```

luukvbaal:
Was fixed in https://github.com/neovim/neovim/pull/28614 @justinmk.

PR #28102, opened 2024-03-29
callHierarchy: gracefully handle when no item was resolved

coot:


coot:
Related to haskell/haskell-language-server#3831

runtime/lua/vim/lsp/buf.lua
```diff
@@ -462,6 +462,10 @@ local function call_hierarchy(method)
       return
     end
     local call_hierarchy_item = pick_call_hierarchy_item(result)
+    if not call_hierarchy_item then
+      vim.notify('No item resolved', vim.log.levels.WARN)
+      return
+    end
```
mfussenegger, line 468: I'd tend to move this up and add another check for the selected item:


```suggestion
    if not result then
      vim.notify('No item resolved', vim.log.levels.WARN)
      return
    end
    local call_hierarchy_item = pick_call_hierarchy_item(result)
    if not call_hierarchy_item then
      return
    end
```

If the user cancels the selection there's no need for a notification, because they made an active choice. If there is no result from the server some feedback seems justified.

coot: I applied.

PR #28131, opened 2024-03-31
vim-patch:9.1.{0234-0250}: lots of filetypes not being recognized

clason:
Problem:  filetype: support for Intel HEX files is lacking
Solution: Add more file extensions that are typical for Intel HEX files
          (Wu, Zhenyu)

Reference: https://en.wikipedia.org/wiki/Intel_HEX

closes: vim/vim#14355

https://github.com/vim/vim/commit/e523dd9803ed62ea0657af8c85ab7bdfe80f4c53

Co-authored-by: Wu, Zhenyu <wuzhenyu@ustc.edu>


clason:
first of a ~20 vim pr bomb; will add them to this PR as they're merged upstream.

runtime/lua/vim/filetype.lua
```diff
@@ -2049,6 +2051,7 @@ local pattern = {
   ['XF86Config%-4.*'] = starsetf(detect.xfree86_v4, { priority = -math.huge + 1 }),
   ['XF86Config.*'] = starsetf(detect.xfree86_v3),
   ['%.zcompdump.*'] = starsetf('zsh'),
+  ['%.zsh_history'] = starsetf('zsh'),
```
zeertzjq, line 2060: This should be in the `filename` table and should not use `starsetf()`.

clason: Why this one and not the one above?

clason: (`starsetf`, that is)

zeertzjq: 1. It's not used in Vim.
2.
```vim
" Function used for patterns that end in a star: don't set the filetype if the
" file name matches ft_ignore_pat.
" When using this, the entry should probably be further down below with the
" other StarSetf() calls.
func s:StarSetf(ft)
  if expand("<amatch>") !~ g:ft_ignore_pat
    exe 'setf ' . a:ft
  endif
endfunc
```

PR #28138, opened 2024-03-31
feat: x11 window manager

altermo:
Neovim has developed much over the past few years. But there was always one thing that I wished Neovim was which Neovim currently isn't, and that is being an X11 window manager. Now you might say that a text editor shouldn't be a window manager, but to that, I say Neovim already does other things that text editors shouldn't be (like `vim.system`).

**Why would anyone use it over other X11 window managers?**
There are multiple reasons to use Neovim as a window manager. One of the main reasons is that you can create your tiling layout (or other window management script) by positioning Vim windows using VimL or Nvim-Lua without needing to learn a single bit of X11.

**Why not Wayland?**
Because Wayland is hard.

**Why write it in Lua instead of C?**
Because I don't know a lot of the internal Neovim C functions. (If required, I can rewrite it in C)

**It uses LuaJit-FFI, and LuaJit is not a target**
Let's make an exception. And for those who can't have Neovim running with LuaJit because they can't compile LuaJit to that hardware, I am pretty sure that they also can't compile an X11 server on that hardware.

**TODO**
+ [ ] Floating Vim windows
+ [ ] Documentation
+ [ ] Testing


clason:
Yes, we would say that. Investigate one of the many GUIs for Neovim.

clason:
(But a nice April Fool's, if that's where you were going with this.)

PR #28143, opened 2024-04-01
fix(dialog): save unamed buffer correctly

### LINKED ISSUE #21682: `confirm` prompt is not helpful when working with dirty unnamed buffers and `nohidden`
### Describe the bug

The 'confirm' prompt is not useful / misleading when working with unnamed buffers and `nohidden`. In these circumstances, it asks whether to save the buffer, but then does nothing. This behaviour is confusing and not helpful. `confirm` should either not prompt in these cases or (preferred) it should prompt for the name of a file to save to, and then carry out the original action.

This bug is also present in Vim, so if you'd prefer I open it there first, I can.

### Steps to reproduce

* `nvim --clean`
* `:set nohidden`
* `:set confirm`
* Enter insert mode and enter anything into the scratch/unnnamed buffer
* `:e ~/any/existing/file.txt<CR>`
* At the confirm prompt, hit 'y'
* (This is the bug) Nothing happens - no switch to new buffer, no prompt for name of file.

### Expected behavior

NeoVim prompts for the name of the file to be saved, saves it, then switches to the new buffer.

### Neovim version (nvim -v)

NVIM v0.9.0-dev-622+g08ebf8d3a

Yes, VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Dec 11 2022 13:50:40) Included patches: 1-1046

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.26.5

xterm-kitty

### Installation

nvenv

glepnir:
Problem: when bufname is empty it can not save correctly

Solution: set bufname before save.

Fix #21682

### DIFF
--- a/src/nvim/ex_cmds2.c
+++ b/src/nvim/ex_cmds2.c
@@ -212,8 +212,12 @@ void dialog_changed(buf_T *buf, bool checkall)
   }
 
   if (ret == VIM_YES) {
-    if (buf->b_fname != NULL
-        && check_overwrite(&ea, buf, buf->b_fname, buf->b_ffname, false) == OK) {
+    if (buf->b_fname == NULL) {
+      buf->b_fname = "Untitled";
+      buf->b_ffname = xstrdup(NameBuff);
+    }
+
+    if (check_overwrite(&ea, buf, buf->b_fname, buf->b_ffname, false) == OK) {
       // didn't hit Cancel
       buf_write_all(buf, false);
     }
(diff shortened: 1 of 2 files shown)


glepnir:
vim in there has `browse_save_fname` we don't have browse functions ..  now just use default prompt string as fname  because there had trhough a prompt .  or we can use  `Save As` prompt and receive a file name ? this looks more better maybe.

Issue #27004, opened 2024-01-13
tests: improve test conventions (SUT, helpers.lua)


justinmk:
# Problem

Writing test code requires some "tribal knowledge" that requires the test author to understand which "helper" functions are dispatched to the `nvim` _test session_ (the "[SUT](https://en.wikipedia.org/wiki/System_under_test)") and which functions are handled in the context of the _test runner_.

Contributors must constantly be aware of the difference. Reading and writing tests can be confusing because of this, especially to non-experienced contributors.

# Expected behavior

Make the distinction clear by either extracting SUT functions to `test/sut.lua` or grouping them in a ~~sut~~ `nvim` container.

- [x] Rename `helpers.lua` to `testutil.lua` and change the import convention to (https://github.com/neovim/neovim/pull/28145):
   ```
   local t = require('testutil')
   ...
   t.foo()
   t.bar()
   ```
    - Thus there is no need anymore to do this:
      ```
      local helpers = require('testutil')
      local foo = helpers.foo
      local bar = helpers.bar
      ```
    - Test code can simply use `t.xx()` everywhere to call non-SUT test util functions.
- [ ] SUT functions live on `nvim`. When using SUT functions in test code, the usage looks like:
   ```lua
   nvim.api.nvim_get_commands()  -- instead of helpers.api.nvim_get_commands
   nvim.fn.foo()  -- instead of helpers.fn.foo()
   ```
    - See "Implementation" below.

# Implementation

Almost everything in `test/functional/testutil.lua` involves a "session". So in practice this would mean:

1. Lift any "pure" functionality (functions that don't require the nvim "test session") from `test/functional/testutil.lua` into `test/testutil.lua`.
    - `test/functional/testutil.lua` calls the `nvim` "test session" via RPC.
1. `test/testutil.lua` is no longer "merged" into `test/functional/testutil.lua`.
    - I.e. delete this line: https://github.com/neovim/neovim/blob/fd085d90820149caecf213b299f19e46305043ee/test/functional/testutil.lua#L19
1. Rename `test/functional/testutil.lua` to (e.g.) `…/nvimsession.lua`
1. All test imports now look like:
   ```lua
   -- Test runner functions/utils.
   local t = require('test.testutil')()
   -- Nvim RPC session.
   local nvim = require('test.functional.nvimsession')()
   ```
    - What should the test session be named?
        - ~~nvim? session? sut?~~ We decided on `n`.

# Future

Some questions to consider for the future (out of scope for this issue):

[...]

PR #28145, opened 2024-04-01
test: improve test conventions

dundargoc:
Work on https://github.com/neovim/neovim/issues/27004.


test/functional/vim.lua
```diff
@@ -0,0 +1,5 @@
+local helpers = require('test.functional.helpers')(after_each)
+
+vim.api = helpers.create_callindex(helpers.request_lua)
```
zeertzjq, line 3:
```suggestion
vim.api = helpers.create_callindex(helpers.request)
```

dundargoc:
@justinmk I dunno, was it something like this you had in mind? Unsure if this is useful tbh.

PR #28154, opened 2024-04-02
feat(filetype): Add typespec filetype

happenslol:
This adds a filetype for the [typespec language](https://typespec.io/), based on the `.tsp` extension.

This is my first time submitting a PR for neovim, please tell me if there's anything missing. There's no syntax/highlighting or anything else at the moment, but I'm writing a tree-sitter parser for the language, which would make use of this filetype.


clason:
Thank you, but we take the logic unchanged from Vim; please open your PR there!

PR #28155, opened 2024-04-02
test: reduce `exec_lua` calls

dundargoc:
`exec_lua` makes code slighly harder to read, so it's beneficial to
remove it in cases where it's possible or convenient.

Not all `exec_lua` calls should be removed even if the test passes as it
changes the semantics of the test even if it happens to pass.

From https://github.com/neovim/neovim/pull/28155#discussion_r1548185779:

"Note for tests like this, which fundamentally are about conversion, you
end up changing what conversion you are testing. Even if the result
happens to be same (as they often are, as we like the rules to be
consistent if possible), you are now testing the RPC conversion rules
instead of the vim script to in-process lua conversion rules."

From https://github.com/neovim/neovim/pull/28155#discussion_r1548190152:

"A test like this specifies that the cursor is valid immediately and not
after a separate cycle of normal (or an other input-processing) mode."

### DIFF
--- a/test/functional/api/buffer_spec.lua
+++ b/test/functional/api/buffer_spec.lua
@@ -1374,12 +1374,7 @@ describe('api/buf', function()
           -- immediate call to nvim_win_get_cursor should have returned the same position
           eq({ 2, 12 }, cursor)
           -- coladd should be 0
-          eq(
-            0,
-            exec_lua([[
-            return vim.fn.winsaveview().coladd
-          ]])
-          )
+          eq(0, fn.winsaveview().coladd)
         end)
 
         it('does not change cursor screen column when cursor >EOL and row got shorter', function()
@@ -1393,9 +1388,7 @@ describe('api/buf', function()
           -- turn on virtualedit
           command('set virtualedit=all')
           -- move cursor after eol
-          exec_lua([[
-            vim.fn.winrestview({ coladd = 5 })
-          ]])
+          fn.winrestview({ coladd = 5 })
 
           local cursor = exec_lua([[
             vim.api.nvim_buf_set_text(0, 0, 15, 2, 11, {
@@ -1414,12 +1407,7 @@ describe('api/buf', function()
           -- immediate call to nvim_win_get_cursor should have returned the same position
           eq({ 2, 26 }, cursor)
           -- coladd should be increased so that cursor stays in the same screen column
-          eq(
-            13,
-            exec_lua([[
-            return vim.fn.winsaveview().coladd
-          ]])
-          )
+          eq(13, fn.winsaveview().coladd)
         end)
 
         it(
@@ -1435,9 +1423,7 @@ describe('api/buf', function()
             -- turn on virtualedit
             command('set virtualedit=all')
             -- move cursor after eol
-            exec_lua([[
-            vim.fn.winrestview({ coladd = 21 })
-          ]])
+            fn.winrestview({ coladd = 21 })
 
             local cursor = exec_lua([[
             vim.api.nvim_buf_set_text(0, 0, 15, 2, 11, {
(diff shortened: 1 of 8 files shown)


luukvbaal:
Note that it is sometimes used to test specifically for redraw behavior that depends on multiple api calls in a single event loop cycle. Doing this will reduce the effectiveness of such tests.

test/functional/api/buffer_spec.lua
```diff
@@ -1076,10 +1075,8 @@ describe('api/buf', function()
         -- position the cursor on 't' in 'first'
         api.nvim_win_set_cursor(0, { 1, 19 })
 
-        local cursor = exec_lua([[
-          vim.api.nvim_buf_set_text(0, 0, 15, 2, 24, {'last'})
-          return vim.api.nvim_win_get_cursor(0)
```
bfredl, line 1081: it is not just about redraw, a test like this specifies that the cursor is valid _immediately_ and not after a separate cycle of normal (or an other input-processing) mode. The alternative, rather than blanked deleting `exec_lua`, would be to use `nvim_call_atomic` or an abstraction built out of it. not sure if that is better than just letting `exec_lua` live though.

test/functional/lua/vim_spec.lua
```diff
@@ -1348,66 +1343,67 @@ describe('lua stdlib', function()
   end)
 
   it('vim.empty_dict()', function()
+    api.nvim_set_var('listy', {})
+    api.nvim_set_var('dicty', vim.empty_dict())
+    local listy = fn.eval('listy')
+    local dicty = fn.eval('dicty')
     eq(
       { true, false, true, true },
-      exec_lua([[
```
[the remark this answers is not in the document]
dundargoc: OK, good to know. I've reverted all changes in this file.

zeertzjq (changes requested):
This will make tests more flaky because of splitting one RPC request into multiple.

test/functional/helpers.lua
```diff
@@ -112,6 +112,10 @@ function module.set_session(s)
   session = s
 end
 
+--- This nasty piece of shit does some sneaky stuff under the hood. So basically each use of this
+--- forces a new loop iteration, so you cannot use this to replace an `exec_lua` with multiple
+--- vim.api calls. Die.
+---
```
zeertzjq, line 118: No, consecutive requests may be accidentally executed in the same loop iteration (unless one of them is `nvim_input`), which is also another source of flakiness for code that assume they are executed in different loop iterations.

zeertzjq (approved):

lewis6991 (approved):

test/functional/api/buffer_spec.lua
justinmk: ^ These kinds of comments are useful for future reference, so I un-resolved them. Alternatively (better) is to add them to the commit message.

test/functional/core/remote_spec.lua
```diff
@@ -151,7 +151,7 @@ describe('Remote', function()
       -- our incoming --remote calls.
       clear()
       -- Call jobstart() and jobwait() in the same RPC request to reduce flakiness.
-      eq({ 2 }, exec_lua([[return vim.fn.jobwait({ vim.fn.jobstart(...) })]], bogus_argv))
+      eq({ 2 }, fn.jobwait({ fn.jobstart(bogus_argv) }))
```
zeertzjq, line 154: No

zeertzjq, line 154: Don't do this. The reason is literally in the comment above.

zeertzjq (changes requested):

test/functional/api/buffer_spec.lua
dundargoc: I've added them to the commit message with slight adjustments.

zeertzjq (approved):

PR #28156, opened 2024-04-02
refactor(tests): allow to extend the new base set of attrs

bfredl:
We start at 100 so we can make the base set larger if needed. (It might need to grow/shrink as a result of adopting the new default color scheme as the default for tests)

Usage best illustrataded by example.

Improving the workflow for making new tests with `screen:snapshot_util()` will be a follow up.

### DIFF
--- a/test/functional/ui/screen.lua
+++ b/test/functional/ui/screen.lua
@@ -255,6 +255,17 @@ function Screen:set_default_attr_ids(attr_ids)
   self._default_attr_ids = attr_ids
 end
 
+function Screen:add_extra_attr_ids(extra_attr_ids)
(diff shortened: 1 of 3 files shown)


test/functional/ui/screen.lua
```diff
@@ -255,6 +255,17 @@ function Screen:set_default_attr_ids(attr_ids)
   self._default_attr_ids = attr_ids
 end
 
+function Screen:add_extra_attr_ids(extra_attr_ids)
```
justinmk, line 258: Nit (only mentioning since you are planning a followup): "extra" is redundant since it is implied by "add"

    add_attr_ids

PR #11658, opened 2020-01-02
clipboard: do not close stderr together with stdout (fixup #11617)

bfredl:
stderr is needed to get error messages in case of failure, and job handler expects it to be open.

PR #28157, opened 2024-04-02
fix: prevent child processes from inheriting ui channel file descriptors

### LINKED ISSUE #26743: Regression: xclip leaves stderr open and causes GUI clients to hang on exit
### Problem

This PR caused a regression where Neovide and other Gui clients fails to exit until xclip is either killed or something else copied to the clipboard
* https://github.com/neovim/neovim/pull/21675
See:
* https://github.com/neovide/neovide/issues/2003

It can be fixed by the following patch:
```diff
diff --git a/runtime/autoload/provider/clipboard.vim b/runtime/autoload/provider/clipboard.vim
index 05f6bdb87..d478ba3a5 100644
--- a/runtime/autoload/provider/clipboard.vim
+++ b/runtime/autoload/provider/clipboard.vim
@@ -215,9 +215,10 @@ function! s:clipboard.set(lines, regtype, reg) abort
   if jobid > 0
     call jobsend(jobid, a:lines)
     call jobclose(jobid, 'stdin')
-    " xclip does not close stdout when receiving input via stdin
+    " xclip does not close stdout and stderr when receiving input via stdin
     if selection.argv[0] ==# 'xclip'
       call jobclose(jobid, 'stdout')
+      call jobclose(jobid, 'stderr')
     endif
     let selection.owner = jobid
     let ret = 1
```

But the code originally looked like that, and the closing of `stderr` was removed here:
* https://github.com/neovim/neovim/pull/11658

So clearly some other fix is needed, but I don't have time to learn that part of the code enough to do it myself right now.

NOTE: One thing that I don't understand in #21675, is the following line:
`dup2(STDERR_FILENO, STDIN_FILENO);` it seems like that would cause some kind of io loop within Neovim, and `stderr` is already duplicated to `stdout` earlier.

### Steps to reproduce

1. Start neovide 0.12.0  on an X11 system, which uses xclip
2. `set clipboard=unamedplus`
3. Yank something to the clipboard
4. Quit the application, notice that it hangs on exit until something else is copied or xclip is killed

### Expected behavior

The xclip stderr channel should be closed so that gui clients can perform a clean exit.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-1956+gec7453804

N/A

### Operating system/version

Arch Linux

### Terminal name/version

Alacritty/Neovide

alacritty

### Installation

build from repo

fredizzimo:
On Unix-like systems, all file descriptors are inherited by default when a new process is forked. This causes problems for UIs to reliably detect when nvim has exited and there's no more data to process.

The problem can most commonly be observed when a clipboard tool like `wl-copy` is used. Those tools are spawned and remain active after nvim exit. But because the channel file descriptors are inherited, they remain open until the tools have quite as well, preventing UI's like neovide from detecting the exit. The problem is not limited to these tools though, any process that is spawned in the background and remain active could cause the same issue.

This has been fixed by setting the flag `FD_CLOSEEXEC`, which ensures that the descriptors are closed in the new process.

This fixes #26743, which was caused by a regression in https://github.com/neovim/neovim/pull/11658

### DIFF
--- a/src/nvim/channel.c
+++ b/src/nvim/channel.c
@@ -544,8 +544,12 @@ uint64_t channel_from_stdio(bool rpc, CallbackReader on_output, const char **err
   }
 #else
   if (embedded_mode) {
-    stdin_dup_fd = dup(STDIN_FILENO);
-    stdout_dup_fd = dup(STDOUT_FILENO);
+    // In embedded mode redirect stdout and stdin to stderr, since they are used for the UI channel.
+    // NOTE: fnctl with F_DUPFD_CLOEXEC is used instead of dup to prevent child processes from
+    // inheriting the file descriptors, which make it impossible for UIs to detect when nvim exits
+    // while one or more of its child processes are still running.
+    stdin_dup_fd = fcntl(STDIN_FILENO, F_DUPFD_CLOEXEC, STDERR_FILENO + 1);
+    stdout_dup_fd = fcntl(STDOUT_FILENO, F_DUPFD_CLOEXEC, STDERR_FILENO + 1);
     dup2(STDERR_FILENO, STDOUT_FILENO);
     dup2(STDERR_FILENO, STDIN_FILENO);
   }


zeertzjq:
This will cause other problems, like #21756

justinmk:
> But because the channel file descriptors are inherited, they remain open until the tools have quite as well, preventing UI's like neovide from detecting the exit

I think we discussed a "quit" UI event.

But this change is good in any case.

PR #28160, opened 2024-04-02
vim-patch:9.1.{0258,0260,0280}

### LINKED ISSUE #28106: Broken scroll + zz mappings
### Problem

On master, a mapping like `<C-d>zz` prevents the cursor from moving, while this correctly scrolled and centered the cursor previously.

### Steps to reproduce

1. Write the following config file to `repro.lua`:
```lua
vim.keymap.set('n', '<C-d>', '<C-d>zz', { desc = 'Scroll downwards' })
vim.keymap.set('n', '<C-u>', '<C-u>zz', { desc = 'Scroll upwards' })
```

2. Open Neovim with `nvim --clean -u repro.lua file`, where `file` has enough content to be able to scroll through it.
3. Type `<C-d>` or `<C-u>`. Notice how the cursor stays fixed.

### Expected behavior

For the mappings to scroll and center the cursor.

### Neovim version (nvim -v)

v0.10.0-dev-38e38d1b4

No, vim seems to have the correct behaviour

### Operating system/version

Asahi Fedora Linux 39

### Terminal name/version

wezterm 20240325_093507_e5ac32f2

wezterm

### Installation

From source

### LINKED ISSUE #28115: Scrolling with `<C-d>` doesn't stop when EOF is visible at the bottom, only stopping when EOF is at the top
### Problem

When using `<C-d>` to scroll down, nvim would only stop once the last line of the file is at the top of the screen, not when the last line is visible at the bottom of the screen.

### Steps to reproduce

- no config needed
- Open a file (more than large enough for a full screen) & keep pressing `<C-d>` until it stops

### Expected behavior

When the last line has appeared on the screen (at the bottom), the cursor should move to that last line without letting that last line move up

### Neovim version (nvim -v)

v0.10.0-dev-2739+ge2224a793

no, vim 9.1

### Operating system/version

Arch Linux 6.8.2-arch2-1 x86_64 GNU/Linux

### Terminal name/version

alacritty 0.13.2 (bb8ea18e)

alacritty

### Installation

https://github.com/neovim/neovim/releases/download/nightly/nvim.appimage

### LINKED ISSUE #28211: <C-u> doesn't scroll half of the screen when there are closed folds
### Problem

When buffer contains closed folds (especially noticable when there are a lot of closed folds one after another). Previously `<C-u>` jumped/scrolled half of the screen no matter is line folded or not (so folds treated like 1 line), but currently `<C-u>` is stopped at each closed fold. `<C-d>` behave like it was previously (if we omit the situation about scrolling below the end of file).

### Steps to reproduce

You can test it with file like https://github.com/monkoose/neocodeium/blob/main/lua/neocodeium/server.lua or create some file with some folds one after another
- nvim --clean server.lua
- yank/paste few times folds in this file for greater visibility of the issue
- Close all folds
- From start of the file use `<C-d>` and then use `<C-u>` few times.

### Expected behavior

`<C-u>` should jump/scroll half of the window height even when closed folds are present.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2823+g7560aee595

no, vim 9.1.145 (some version before that notorious PR)

### Operating system/version

Linux

### Terminal name/version

st

st-256color

### Installation

AUR

luukvbaal:
**vim-patch:9.1.0258: half-page scrolling broke backward compatibility**

Problem:  Support for 'smoothscroll' in (half-)page scrolling
          broke backward compatibility and can be made to work better.
          (after v9.1.215)
Solution: Restore the previous cursor and end-of-buffer behavior for
          half-page scrolling and improve 'smoothscroll' support.
          (Luuk van Baal)

https://github.com/vim/vim/commit/cb204e688e5c9d56a78b621ef27c35d91860cb09

**vim-patch:9.1.0260: Problems with "zb" and scrolling to new topline with 'smoothscroll'**

Problem: "zb" does not reveal filler lines at the start of a buffer.
          Scrolled cursor position with 'smoothscroll' is unpredictable,
          and may reset skipcol later if it is not visible (after v9.1.258)
Solution: Replace confusing for loop that reaches final control value too
          early with while loop. Set "w_curswant" accordingly so cursor
          will be placed in visible part of topline.
          (Luuk van Baal)

https://github.com/vim/vim/commit/bd28cae1f1c21c0e3743e3427c98bbd848fad237

**vim-patch:9.1.0280: several issues with 'smoothscroll' support**

Problem:  Logic to make sure cursor is in visible part of the screen after
          scrolling the text with 'smoothscroll' is scattered, asymmetric
          and contains bugs.
Solution: Adjust and create helper function for 'smoothscroll' cursor logic.
          (Luuk van Baal)

https://github.com/vim/vim/commit/9148ba8a46baa3934c44164989cdcdec5d01d9e3

Fix #28106
Fix #28115
Fix #28211


src/nvim/move.c
```diff
@@ -2341,61 +2352,117 @@ static int get_scroll_overlap(Direction dir)
   }
 }
 
-/// Move screen "count" pages up ("dir" is BACKWARD) or down ("dir" is FORWARD)
-/// and update the screen.
+/// Scroll "count" lines with 'smoothscroll' in direction "dir". Adjust "count"
+/// when scrolling more than "count" and return TRUE when scrolling happened.
+static int scroll_with_sms(Direction dir, int *count)
```
zeertzjq, line 2357:
```suggestion
/// when scrolling more than "count" and return true when scrolling happened.
static bool scroll_with_sms(Direction dir, int *count)
```

BlankTiger:
I just compiled neovim straight from `luukvbaal:vim-9.1.0258` and while `<C-u>zz` and `<C-d>zz` work great, the same issue still persists with `<C-f>zz` and `<C-b>zz`. Essentially half-page scrolling seems to be fixed, but full-page scrolling still seems broken.

zeertzjq:
What is your expected behavior? Please create a separate issue.

BlankTiger:
It already has an issue: https://github.com/neovim/neovim/issues/28106. Basically it's broken scrolling but for full-page scroll mappings + zz (which is in scope of the mentioned issue because it doesn't specify that only half-page scrolling + zz is broken).

Sorry if it's the wrong place for this comment, I can add it in the mentioned issue if that would be more appropriate.

zeertzjq:
Please answer the question: What behavior do you expect?

BlankTiger:
Expected behavior would be the same as described in https://github.com/neovim/neovim/issues/28106: For the mappings to scroll and center the cursor.

Problematic mappings are the the same as in the mentioned issue with the difference being that `<C-d>` (half-page scroll downwards) becomes `<C-f>` (full-page scroll downwards), and `<C-u>` (half-page scroll upwards) becomes `<C-b>` (full-page scroll upwards):

```lua
vim.keymap.set('n', '<C-f>', '<C-f>zz', { desc = 'Scroll downwards' })
vim.keymap.set('n', '<C-b>', '<C-b>zz', { desc = 'Scroll upwards' })
```

zeertzjq:
That's a reasonable feature request, but I don't think they are not supposed to behave like that, so there is no bug.

zeertzjq:
I tested on older versions and it seems that they are not supposed to behave so, so there is no bug.

BlankTiger:
Weird, works for me on stable nvim 0.9.5 appimage with `--clean` flag and only those mapping set.

luukvbaal:
If you see difference in behavior please provide clear reproduction steps (buffer text, window size, options, keys pressed).

src/nvim/plines.c
```diff
@@ -885,6 +885,9 @@ int plines_m_win(win_T *wp, linenr_T first, linenr_T last, bool limit_winheight)
     count += plines_win_full(wp, first, &next, NULL, false, limit_winheight);
     first = next + 1;
   }
+  if (first == wp->w_buffer->b_ml.ml_line_count + 1) {
+    count += win_get_fill(wp, first);
+  }
```
luukvbaal, line 890: Over at vim/vim it was [suggested](https://github.com/vim/vim/pull/14410#discussion_r1554805999) to not add this to `plines_win()`, and decided to instead add it to `plines_m_win()`. Going by [this](https://github.com/vim/vim/pull/14410#discussion_r1554853125) logic, it might make more sense for us to add it to `plines_win_full()`? Or do we still have concerns of it causing problems there?

zeertzjq: But that'll lead to incorrect `w_cline_height`.

src/nvim/move.c
```diff
@@ -1185,6 +1189,122 @@ void f_virtcol2col(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
   rettv->vval.v_number = virtcol2col(wp, lnum, screencol);
 }
 
+/// Make sure the cursor is in the visible part of the topline after scrolling
+/// the screen with 'smoothscroll'.
+static void cursor_correct_sms(win_T *wp)
+{
+  if (!curwin->w_p_sms || !curwin->w_p_wrap
+      || curwin->w_cursor.lnum != curwin->w_topline) {
```
zeertzjq, line 1197:
```suggestion
  if (!wp->w_p_sms || !wp->w_p_wrap
      || wp->w_cursor.lnum != wp->w_topline) {
```

src/nvim/move.c
```diff
@@ -2341,61 +2404,120 @@ static int get_scroll_overlap(Direction dir)
...
+      }
+      if (n - count < curwin->w_height_inner) {
+        count = n - curwin->w_height_inner;
+      }
+    }
 
-    // Temporarily set 'smoothscroll' so that scrolling count lines
-    // does not skip over parts of the buffer with wrapped lines.
-    curwin->w_p_sms = true;
     if (count > 0) {
```
zeertzjq, line 2474:
```suggestion
    // Scroll the window and determine number of lines to move the cursor.
    if (count > 0) {
```

src/nvim/move.c
```diff
@@ -1185,6 +1189,121 @@ void f_virtcol2col(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
...
+    return;
+  }
+
+  int so = get_scrolloff_value(wp);
+  int width1 = wp->w_width_inner - win_col_off(wp);
+  int width2 = width1 + win_col_off2(wp);
+  int so_cols = so == 0 ? 0 : width1 + (so - 1) * width2;
+  int space_cols = (wp->w_height_inner - 1) * width2;
+  int size = so == 0 ? 0 : win_linetabsize(wp, wp->w_topline,
+                                           ml_get(wp->w_topline), (colnr_T)MAXCOL);
```
zeertzjq, line 1206:
```suggestion
                                           ml_get_buf(wp->w_buffer, wp->w_topline), (colnr_T)MAXCOL);
```

src/nvim/change.c
```diff
@@ -343,8 +343,7 @@ static void changed_common(buf_T *buf, linenr_T lnum, colnr_T col, linenr_T lnum
               || (wp->w_topline >= lnum
                   && wp->w_topline < lnume
                   && win_linetabsize(wp, wp->w_topline, ml_get(wp->w_topline), MAXCOL)
```
zeertzjq, line 345:
```suggestion
                  && win_linetabsize(wp, wp->w_topline, ml_get_buf(wp->w_buffer, wp->w_topline), MAXCOL)
```

src/nvim/move.c
zeertzjq: There is also another one in `scrolldown()`, maybe correct that one as well.

src/nvim/change.c
luukvbaal: Lol, should have done this tomorrow not just before bed. There was another offender in move.c from the recent decurwinification.

src/nvim/move.c
luukvbaal: Oh yeah, spotted that one as well.

zeertzjq (approved):

zeertzjq (approved):

PR #28166, opened 2024-04-03
add earthfile filetype

glehmann:
for https://earthly.dev/ file format


clason:
Thanks, but we take the filetype detection logic unchanged from Vim. Please make a PR to vim/vim.

PR #28169, opened 2024-04-03
feat(extmark): support end_col=-1 if strict=false

### LINKED ISSUE #27469: Cannot specify -1 for end_col in `nvim_buf_set_extmark()`
### Problem

As the title says, In `nvim_buf_add_highlight()`, since -1 can be specified as `{col}` (or `{col_end}` in `nvim_buf_add_highlight()`), it is lack coherence to not be able to use -1 in `{opts}.end_col`.
It should be fixed by tweaking this area.

https://github.com/neovim/neovim/blob/a376d979bda103fa9998d05c3cc4ba56d3c3cece/src/nvim/api/extmark.c#L556

### Steps to reproduce

```
nvim --clean
:call setline(1, 'foo')
:call nvim_buf_set_extmark(0, nvim_create_namespace('foo'), 0, 0, #{ end_row: 0, end_col: -1, hl_group: 'Error' })
```

### Expected behavior

Add highlight to the first line.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-a376d97

No

### Operating system/version

Ubuntu 20.04.5 LTS (WSL2)

### Terminal name/version

Windows terminal

tmux-256color

### Installation

build from repo

NStefan002:
As mentioned in #27469, there is an inconsistency between extmarks/highlights regarding the `end_col` option. Now the `end_col` option in the `nvim_buf_set_extmark` function can be set to `-1` - meaning to the end of the line.

Closes #27469.

### DIFF
--- a/src/nvim/api/extmark.c
+++ b/src/nvim/api/extmark.c
@@ -402,7 +402,7 @@ ArrayOf(DictAs(get_extmark_item)) nvim_buf_get_extmarks(Buffer buffer, Integer n
 /// @param opts  Optional parameters.
 ///               - id : id of the extmark to edit.
 ///               - end_row : ending line of the mark, 0-based inclusive.
-///               - end_col : ending col of the mark, 0-based exclusive.
+///               - end_col : ending col of the mark, 0-based exclusive, or -1 to extend the range to end of line.
 ///               - hl_group : highlight group used for the text range. This and below
 ///                   highlight groups can be supplied either as a string or as an integer,
 ///                   the latter of which can be obtained using |nvim_get_hl_id_by_name()|.
@@ -586,9 +586,13 @@ Integer nvim_buf_set_extmark(Buffer buffer, Integer ns_id, Integer line, Integer
   colnr_T col2 = -1;
   if (HAS_KEY(opts, set_extmark, end_col)) {
     Integer val = opts->end_col;
-    VALIDATE_RANGE((val >= 0 && val <= MAXCOL), "end_col", {
+    VALIDATE_RANGE((val >= -1 && val <= MAXCOL), "end_col", {
       goto error;
     });
+    if (val == -1) {
+      val = MAXCOL;
+    }
+
     col2 = (int)val;
   }
 
--- a/test/functional/api/extmark_spec.lua
+++ b/test/functional/api/extmark_spec.lua
@@ -153,7 +153,7 @@ describe('API/extmarks', function()
     )
   end)
 
-  it('can end extranges past final newline using end_col = 0', function()
+  it('can end extranges past final newline using end_col=0', function()
     set_extmark(ns, marks[1], 0, 0, {
       end_col = 0,
       end_row = 1,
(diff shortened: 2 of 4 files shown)


src/nvim/api/extmark.c
```diff
@@ -552,9 +552,9 @@ Integer nvim_buf_set_extmark(Buffer buffer, Integer ns_id, Integer line, Integer
   colnr_T col2 = -1;
   if (HAS_KEY(opts, set_extmark, end_col)) {
     Integer val = opts->end_col;
-    VALIDATE_RANGE((val >= 0 && val <= MAXCOL), "end_col", {
```
justinmk, line 555: why not validate it still?

    VALIDATE_RANGE((val >= -1 && val <= MAXCOL), "end_col", {

src/nvim/api/extmark.c
```diff
@@ -552,9 +552,9 @@ Integer nvim_buf_set_extmark(Buffer buffer, Integer ns_id, Integer line, Integer
...
-      goto error;
-    });
+    if (val < 0 || val > MAXCOL) {
```
justinmk, line 555: check for -1 explicitly, no need to allow other random values

src/nvim/api/extmark.c
NStefan002: Makes sense, it's done.

test/functional/api/extmark_spec.lua
```diff
@@ -173,6 +173,18 @@ describe('API/extmarks', function()
     })
   end)
 
+  it('col_end is -1', function()
```
zeertzjq, line 176:
```suggestion
  it('end_col is -1', function()
```

test/functional/api/extmark_spec.lua
```diff
@@ -173,6 +173,18 @@ describe('API/extmarks', function()
...
 
+  it('end_col is -1', function()
+    set_extmark(ns, marks[1], 0, 0, {
+      end_col = -1,
+      end_row = 0,
+      strict = false,
+    })
+    eq(
+      "Invalid 'end_col': out of range",
+      pcall_err(set_extmark, ns, marks[2], 0, 0, { end_col = -1 })
```
justinmk, line 184: what is this supposed to be testing? is `marks[2]` empty or something?

NStefan002: From `nvim_buf_set_extmark` docs:
```
- strict: boolean that indicates extmark should not be placed
  if the line or column value is past the end of the
  buffer or end of the line respectively. Defaults to true.
```
So I think the purpose of the test you're referring to is to test what happens when `strict = true`.

I guess the better way of doing this is explicitly setting `strict` to `true`, but I tried to keep it consistent with other tests.

justinmk: > I guess the better way of doing this is explicitly setting `strict` to `true`, but I tried to keep it consistent with other tests.

That would definitely clarify the test. Other tests can be updated if you want.

NStefan002: > That would definitely clarify the test. Other tests can be updated if you want.

I'll make sure to add it right away. Also, should I open a separate PR to clarify other tests?

PR #28171, opened 2024-04-04
refactor: remove fn_bool

dundargoc:
It's better to use vim.fn directly instead of creating minor
abstractions like fn_bool.

### DIFF
--- a/runtime/lua/nvim/health.lua
+++ b/runtime/lua/nvim/health.lua
@@ -1,18 +1,6 @@
 local M = {}
 local health = require('vim.health')
 
-local fn_bool = function(key)
-  return function(...)
-    return vim.fn[key](...) == 1
-  end
-end
-
-local has = fn_bool('has')
-local executable = fn_bool('executable')
-local empty = fn_bool('empty')
-local filereadable = fn_bool('filereadable')
-local filewritable = fn_bool('filewritable')
-
 local shell_error = function()
   return vim.v.shell_error ~= 0
 end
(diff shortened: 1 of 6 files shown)


runtime/lua/nvim/health.lua
```diff
@@ -121,30 +112,33 @@ local function check_config()
   local writeable = true
   local shadaopt = vim.fn.split(vim.o.shada, ',')
   local shadafile = (
-    empty(vim.o.shada) and vim.o.shada
+    vim.fn.empty(vim.o.shada) == 1 and vim.o.shada
```
bfredl, line 115: when `vim.fn.empty(str)` is used to check for an empty string, we should instead just compare `vim.o.shada == ''`

runtime/lua/nvim/health.lua
```diff
@@ -121,30 +109,33 @@ local function check_config()
...
-    empty(vim.o.shadafile)
-      and (empty(shadafile) and vim.fn.stdpath('state') .. '/shada/main.shada' or vim.fn.expand(
+    vim.o.shadafile == ''
+      and (shadafile == '' and vim.fn.stdpath('state') .. '/shada/main.shada' or vim.fn.expand(
         shadafile
       ))
     or (vim.o.shadafile == 'NONE' and '' or vim.o.shadafile)
   )
-  if not empty(shadafile) and empty(vim.fn.glob(shadafile)) then
+  if shadafile ~= '' and vim.fn.empty(vim.fn.glob(shadafile)) == 1 then
```
zeertzjq, line 122:
```suggestion
  if shadafile ~= '' and vim.fn.glob(shadafile) == '' then
```

runtime/lua/nvim/health.lua
```diff
@@ -160,7 +151,7 @@ local function check_performance()
 
   -- Check buildtype
   local buildtype = vim.fn.matchstr(vim.fn.execute('version'), [[\v\cbuild type:?\s*[^\n\r\t ]+]])
-  if empty(buildtype) then
+  if vim.fn.empty(buildtype) == 1 then
```
zeertzjq, line 154:
```suggestion
  if buildtype == '' then
```

runtime/lua/nvim/health.lua
```diff
@@ -196,12 +187,12 @@ local function check_rplugin_manifest()
   local require_update = false
   local handle_path = function(path)
     local python_glob = vim.fn.glob(path .. '/rplugin/python*', true, true)
-    if empty(python_glob) then
+    if vim.fn.empty(python_glob) == 1 then
```
zeertzjq, line 190:
```suggestion
    if python_glob == '' then
```

zeertzjq (commented):
`== 0` seems better than `~= 1` as it is more distinguishable from `== 1`. `== 0` is also more common in existing uses of `vim.fn`.

dundargoc:
> `== 0` seems better than `~= 1` as it is more distinguishable from `== 1`. `== 0` is also more common in existing uses of `vim.fn`.

Actually I don't think this works. It can return `-1` for e.g. `executable`.

zeertzjq (approved):

PR #28173, opened 2024-04-04
build: fix link error for `_BitScanForward64`

### LINKED ISSUE #28122: Windows build link error
### Problem

Get a link error,
While trying to build for windows, following instructions at:
https://github.com/neovim/neovim/blob/master/BUILD.md#building-on-windows

...
[385/421] Linking C executable bin\nvim.exe
FAILED: bin/nvim.exe lib/nvim.lib
C:\Windows\system32\cmd.exe /C "cd . && "C:\Program Files\CMake\bin\cmake.exe" -E vs_link_exe --intdir=src\nvim\CMakeFiles\nvim_bin.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100220~1.0\x86\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100220~1.0\x86\mt.exe --manifests D:\dev\neovim\src\nvim\os\nvim.manifest -- C:\PROGRA~1\MICROS~2\2022\COMMUN~1\VC\Tools\MSVC\1438~1.331\bin\Hostx86\x86\link.exe /nologo @CMakeFiles\nvim_bin.rsp  /out:bin\nvim.exe /implib:lib\nvim.lib /pdb:bin\nvim.pdb /version:0.0 /machine:X86 /INCREMENTAL:NO /subsystem:console   /INCREMENTAL:NO /LTCG && cd ."
LINK: command "C:\PROGRA~1\MICROS~2\2022\COMMUN~1\VC\Tools\MSVC\1438~1.331\bin\Hostx86\x86\link.exe /nologo @CMakeFiles\nvim_bin.rsp /out:bin\nvim.exe /implib:lib\nvim.lib /pdb:bin\nvim.pdb /version:0.0 /machine:X86 /INCREMENTAL:NO /subsystem:console /INCREMENTAL:NO /LTCG /MANIFEST:EMBED,ID=1 /MANIFESTINPUT:D:\dev\neovim\src\nvim\os\nvim.manifest" failed (exit code 1120) with the following output:
   Creating library lib\nvim.lib and object lib\nvim.exp
LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs; use /NODEFAULTLIB:library
math.c.obj : error LNK2001: unresolved external symbol __BitScanForward64
bin\nvim.exe : fatal error LNK1120: 1 unresolved externals
ninja: build stopped: subcommand failed.
...

full log is too long to paste here

### Steps to reproduce

VsDevCmd.bat -arch=x64
cmake -S cmake.deps -B .deps -G Ninja -D CMAKE_BUILD_TYPE=Release
cmake --build .deps --config Release
cmake -B build -G Ninja -D CMAKE_BUILD_TYPE=Release
cmake --build build --config Release

### Expected behavior

expected to build fully

### Neovim version (nvim -v)

source

not building vim

### Operating system/version

windows 10 pro

### Terminal name/version

powershell

not important

### Installation

not installed

famiu:
Problem: The usage of `_BitScanForward64` causes linking to fail on some systems.

Solution: Correctly check if it exists using `check_c_source_compiles`.

Fix #28122


famiu:
> We could use `check_c_source_compiles` and use that to conditionally decide?
> 
> https://github.com/neovim/neovim/blob/1b55ed58ecaba96f91a21685b7019362320ea156/cmake.config/CMakeLists.txt#L9
> 
> Was #27969 for performance reasons or just nice to have?

Primarily for consistency and performance reasons.

Good point on using `check_c_source_compiles`. Gonna do that now.

justinmk (approved):

dundargoc:
It's not a blocker for this PR, but still something that should be fixed in the future IMO.

bfredl:
does `check_c_source_compiles` check if things can _link_ ? because that was the failing step.

dundargoc:
> does `check_c_source_compiles` check if things can _link_ ? because that was the failing step.

Checked the docs, they claim that linking is also checked.

famiu:
> Wait, isn't ` _BitScanForward64` always available on 64-bit windows or was that an incorrect assumption?

@dundargoc  I thought so too but explicitly checking if `_WIN64` was defined did not prevent the link failure either.

dundargoc:
Hmm, weird. OK, I guess this is needed then until we know more

cmake.config/CMakeLists.txt
```diff
@@ -186,3 +186,15 @@ configure_file (
...
+#include <intrin.h>
+
+int main(void)
+{
+  unsigned long index;
+  unsigned char mask = 0x8000;
+  _BitScanForward64(&index, mask);
+  return 0;
+}
+" HAVE_BITSCANFORWARD64)
```
zeertzjq, line 200: Shouldn't there also be a change to `cmake.config/config.h.in`?

zeertzjq: And I think this should be place near other `check_c_source_compiles` checks instead of being at the end of the file.

src/nvim/math.c
```diff
@@ -56,7 +56,7 @@ int xctz(uint64_t x)
   // Use compiler builtin if possible.
 #if defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 4))
   return __builtin_ctzll(x);
-#elif defined(_MSC_VER)
+#elif HAVE_BITSCANFORWARD64
```
zeertzjq, line 59:
```suggestion
#elif defined(HAVE_BITSCANFORWARD64)
```

famiu: Oh gosh how did I miss that

PR #28179, opened 2024-04-04
vim-patch:408281e16a36

clason:
runtime: Remove fallback :CompilerSet definition from compiler plugins

The :CompilerSet command was added in version Vim 6.4 which was released
twenty years ago.  Other runtime files do not support versions of that
vintage so it is reasonable to remove this fallback command definition
now.

closes: vim/vim#14399

https://github.com/vim/vim/commit/408281e16a36c15eed10fbf0406fa8ab159fc4bf

Co-authored-by: Doug Kearns <dougkearns@gmail.com>

### DIFF
--- a/runtime/compiler/ant.vim
+++ b/runtime/compiler/ant.vim
@@ -2,15 +2,12 @@
 " Compiler:	ant
 " Maintainer:	Johannes Zellner <johannes@zellner.org>
 " Last Change:	Mi, 13 Apr 2005 22:50:07 CEST
+"		2024 Apr 03 by The Vim Project (removed :CompilerSet definition)
 
 if exists("current_compiler")
     finish
 endif
 let current_compiler = "ant"
-
-if exists(":CompilerSet") != 2		" older Vim always used :setlocal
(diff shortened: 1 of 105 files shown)


zeertzjq:
Omitting `compiler/context.vim` is very strange here, as that will make it the only remaining compiler file that contains such a block. So I think it's better to remove it:
```diff
diff --git a/runtime/compiler/context.vim b/runtime/compiler/context.vim
index cb78c96df..fa2626994 100644
--- a/runtime/compiler/context.vim
+++ b/runtime/compiler/context.vim
@@ -9,10 +9,6 @@ endif
 let s:keepcpo= &cpo
 set cpo&vim
 
-if exists(":CompilerSet") != 2    " older Vim always used :setlocal
-  command -nargs=* CompilerSet setlocal <args>
-endif
-
 " If makefile exists and we are not asked to ignore it, we use standard make
 " (do not redefine makeprg)
 if get(b:, 'context_ignore_makefile', get(g:, 'context_ignore_makefile', 0)) ||
```

zeertzjq (approved):

PR #21393, opened 2022-12-12
feat:  `vim.inspect_pos()`, `vim.show_pos()` and `:Inspect`

folke:
This PR adds `vim.inspect_pos()`, `vim.show_pos()` and `:Inspect` to get/show the items at a given buffer position.

This PR includes 2e328ba7d2b6f10fc9f495b836debbc3036507ca `vim.semantic_tokens.get_at_pos`

Right now, the function uses `nvim_echo`, but I can change it to use a floating window if that would be better suited.

# Todo

- [x] add special handling for semantic token types and modifiers. Would be great if #21373 could be re-opened and merged to implement this
- [x] put the `nvim_create_user_command` somewhere. What would be the best place for this? `_editor.lua`? `plugin/inspector.lua`?
- [x] improve the docs on the return values
- [x] add something to `news.txt`

# Included Items

- [x] Treesitter captures
- [x] Semantic Tokens
- [x] Syntax Highlight Groups  
- [x] Extmarks 

# Implementation details

- The implementation resides in `runtime/lua/vim/_inspector.lua`
- The submodule is lazy-loaded inside `_init_packages.lua`, so there's no impact on performance
- If you want to test, make sure to build Neovim so the lazy-loading is included
- any parameter can be `nil`, meaning the position under the cursor

# User Commands

- `:Inspect` is the same as `vim.show_pos()`
- `:Inspect!` is the same as `vim.pretty_print(vim.inspect_pos())`

# `vim.show_pos(bufnr, row, col, filter)`

image

image

# `vim.inspect_pos(bufnr, row, col, filter)`

```lua
{
  extmarks = { {
      col = 6,
      id = 190,
      ns = "illuminate.highlight",
      ns_id = 30,
      opts = {
        end_col = 7,
        end_right_gravity = false,
        end_row = 4,
        hl_eol = false,
        hl_group = "IlluminatedWordWrite",
        hl_group_link = "IlluminatedWordWrite",
        priority = 1000,
        right_gravity = true
      },
      row = 4
    } },
  semantic_tokens = { {
      client_id = 1,
      end_col = 7,
      extmark_added = true,
      hl_groups = {
        modifiers = { {
            hl_group = "@declaration",
            hl_group_link = "@variable"
          } },
        type = {
          hl_group = "@variable",
          hl_group_link = "@variable"
        }
      },
      line = 4,
      modifiers = { "declaration" },
      start_col = 6,
      type = "variable"
    } },
  syntax = {},
  treesitter = { {
      capture = "variable",
[...]

PR #28181, opened 2024-04-05
feat(defaults): add :Inspect to right-click menu

zeertzjq:
Ref #21393

- Move default user commands to _defaults.lua as that now contains all
  kinds of defaults rather than just default mappings and menus.
- Remove the :aunmenu as there are no menus when _defaults.lua is run.


gpanders (approved):
Great idea!

PR #28189, opened 2024-04-05
docs: don't mention executable() can return -1

dundargoc:
This cannot happen for neovim.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1268,7 +1268,6 @@ executable({expr})                                                *executable()*
 		The result is a Number:
 			1	exists
 			0	does not exist
-			-1	not implemented on this system
 		|exepath()| can be used to get the full path of an executable.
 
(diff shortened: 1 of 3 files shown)


src/nvim/CMakeLists.txt
```diff
@@ -951,6 +951,7 @@ add_custom_command(
     ${PROJECT_SOURCE_DIR}/src/nvim/eval.lua
     ${PROJECT_SOURCE_DIR}/src/nvim/options.lua
     ${PROJECT_SOURCE_DIR}/src/nvim/vvars.lua
+    ${NVIM_RUNTIME_DIR}/doc/builtin.txt
```
zeertzjq, line 954: Should other generated files also be added here then?

dundargoc: Nah, I'm deleting this line. The whole thing needs to be rewritten.

zeertzjq (approved):

PR #28205, opened 2024-04-06
docs: tags for commenting mappings without "-default" suffix

zeertzjq:
This actually won't cause "duplicate tag" errors if plugins have tags of
the same name, because E154 is only given for duplicate tags in the same
directory.

Without those tags, trying to use :h for these mappings jumps to other
places, because there are matches with higher score.


echasnovski:
Yeah, it was my mistake of not actually verifying that it would give duplicated tags error and instead relying on my experience with having that error (which indeed for me was for tags in the same directory).

PR #28208, opened 2024-04-06
docs: various fixes

dundargoc:


BUILD.md
```diff
@@ -5,6 +5,8 @@
 
 1. Install [build prerequisites](#build-prerequisites) on your system
 2. `git clone https://github.com/neovim/neovim`
+  - You may need to install git if it's not available by default on your
+    system.
```
justinmk, line 9: Not in favor of this. The point of a "quick start" is to give a very condensed starting point to cover the common case. Adding contingent language and trying to cover uncommon cases defeats the purpose of a quick-start.

People that don't have git and don't understand what `git clone` implies are explicitly not supported.

dundargoc: > Adding contingent language and trying to cover uncommon cases defeats the purpose of a quick-start.

Yeah, that's a fair point. Removing it.

PR #28215, opened 2024-04-07
fix(health): check unmatching python_glob as empty table

barrettruth:
The current health code checking for remote python plugins passes `true` to `vim.fn.glob` indicating a table return type, yet compares it to an empty string.

The following error results:
image

Comparing against an empty table fixes the issue, resulting in the following:
image

This output does seem quite spammy for that of a default health check - that may be another issue.

### DIFF
--- a/runtime/lua/nvim/health.lua
+++ b/runtime/lua/nvim/health.lua
@@ -187,7 +187,7 @@ local function check_rplugin_manifest()
   local require_update = false
   local handle_path = function(path)
     local python_glob = vim.fn.glob(path .. '/rplugin/python*', true, true)
-    if python_glob == '' then
+    if vim.tbl_isempty(python_glob)  then
       return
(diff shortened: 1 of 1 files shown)


zeertzjq (approved):

runtime/lua/nvim/health.lua
```diff
@@ -187,7 +187,7 @@ local function check_rplugin_manifest()
   local require_update = false
   local handle_path = function(path)
     local python_glob = vim.fn.glob(path .. '/rplugin/python*', true, true)
-    if python_glob == '' then
+    if vim.tbl_isempty(python_glob)  then
```
zeertzjq, line 190:
```suggestion
    if vim.tbl_isempty(python_glob) then
```

barrett-ruth:
TY

PR #28216, opened 2024-04-07
fix(lsp): empty commands should not be considered executable

ofseed:
1. `CodeLens.command` will exist only when it is resolved, according to https://github.com/neovim/neovim/blob/067f51e3aadfeb797b38331776f2a6447b565456/runtime/lua/vim/lsp/codelens.lua#L275-L278, there may be a very short time interval in which this field does not exist;
2. Once `resolve_lenses` is done, the `CodeLens.command` and all its subfields will exist, but not all of them will be executable, at this time, `CodeLens.command.command` will be `''`, an empty string. In this case, we should not add them to `options`.

### DIFF
--- a/runtime/lua/vim/lsp/codelens.lua
+++ b/runtime/lua/vim/lsp/codelens.lua
@@ -79,7 +79,7 @@ function M.run()
   local lenses_by_client = lens_cache_by_buf[bufnr] or {}
   for client, lenses in pairs(lenses_by_client) do
     for _, lens in pairs(lenses) do
-      if lens.range.start.line == (line - 1) then
+      if lens.range.start.line == (line - 1) and lens.command and lens.command.command ~= '' then
         table.insert(options, { client = client, lens = lens })
       end
     end


justinmk:
>  there may be a very short time interval in which this field does not exist;

Is that "race" because of our internal design or because of interactions with the LSP server?

IOW: should this be an error? Should we at least log it?

ofseed:
> Is that "race" because of our internal design or because of interactions with the LSP server?

To get code lenses to be shown on these buffers, there will be two-stage:

1. First request of `textDocument/codeLens` to get these lenses, in this stage, the LSP server returns `CodeLens[]`, but these `CodeLens[]` is incomplete, they have two fields: `data` and `range`, but no `command` field.
2. To get complete information on these `CodeLens[]` (specifically, it is the `command` field), the client must request `codeLense/resolve`, and the parameter of this request is a certain Codelens obtained in stage 1, that is, the full information of which codelens is to be obtained.

Now answer the question, after stage 1 we have no `command` fields but we saved these lenses, so this condition may occur in the middle of the two stages, although codeLens/resolve is generally very fast. So I think it should be the latter.

> IOW: should this be an error? Should we at least log it?

This should not be an error. We just should not treat unresolved codelens as executable commands, since they do not have a `command` field, and will be resolved soon.

ofseed:
By the way, let me explain the second situation in this PR description in detail.
According to the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command), the `CodeLens.command` is optional but the `CodeLens.command.command` is not optional, which means the correct representation of an executable is indeed a `title` to display and an empty string as `command`.

Issue #24972, opened 2023-09-01
error attempt to get length of local 'prev_line' 


eyalk11:
### Problem

Sometimes lsp client hangs and get 

```
Error executing lua callback: ...rs\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp/sync.lua:177: attempt to get length of local 'prev_line' (a nil value)
stack traceback:
	...rs\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp/sync.lua:177: in function 'compute_start_range'
	...rs\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp/sync.lua:408: in function 'compute_diff'
	C:\Users\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp.lua:446: in function 'incremental_changes'
	C:\Users\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp.lua:639: in function 'send_changes'
	C:\Users\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp.lua:1577: in function <C:\Users\ekarni\Neovim\share\nvim\runtime/lua/vim/lsp.lua:1564>


```

The file is python file. I use lsp_config and nvim-lsp-installer with copilot , null-ls , pyright, pylsp as lsp clients. I don't know which failed. 
I got this message constantly when writing a line (on every char). It got OK eventually. 

### Steps to reproduce

Unkown

### Expected behavior

This shouldn't happen

### Neovim version (nvim -v)

[...]

PR #28217, opened 2024-04-07
fix(prompt): emit change event for prompt newline

icholy:
See #24972

### DIFF
--- a/src/nvim/eval.c
+++ b/src/nvim/eval.c
@@ -17,6 +17,7 @@
 #include "nvim/autocmd.h"
 #include "nvim/buffer.h"
 #include "nvim/buffer_defs.h"
+#include "nvim/change.h"
 #include "nvim/channel.h"
 #include "nvim/charset.h"
 #include "nvim/cmdexpand_defs.h"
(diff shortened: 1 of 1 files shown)


zeertzjq:
Can you add a test in `test/functional/lua/buffer_updates_spec.lua`?

icholy:
I'm trying to reproduce this with the following code:

``` lua
-- create prompt buffer
local buf = vim.api.nvim_create_buf(true, true)

-- attach listener
vim.api.nvim_buf_attach(buf, true, {
  on_lines = function (_, _, _, firstline, lastline, _, _)
	vim.print("on_lines", { firstline = firstline, lastline = lastline })
  end
})

-- convert it to a prompt
vim.api.nvim_buf_set_name(buf, '[repl]')
vim.bo[buf].buftype = 'prompt'
vim.bo[buf].buflisted = false
vim.bo[buf].filetype = 'repl'
vim.fn.prompt_setprompt(buf, 'repl> ')
vim.fn.prompt_setcallback(buf, vim.print)

-- send some keys
vim.api.nvim_set_current_buf(buf)
vim.api.nvim_feedkeys("i\n", "n", true)
```

But it the `firstline` is coming out correctly.

zeertzjq:
I tried to add a test for this and it seems there is a missing call to `on_bytes`:
```
on_bytes: total bytecount of buffer is wrong
Expected objects to be the same.
Passed in:
(number) 5
Expected:
(number) 6
```
I think it should use `appended_lines_mark()` instead.

icholy:
Closing in favor of #28260

PR #28219, opened 2024-04-07
fix(drawline): don't invoke on_line for filler line

tomtomjhj:
Problem:
Decoration provider `on_line` handler is invoked for diff filler line below the last buffer line. This does not match the documentation: "called for each buffer line".

Solution:
Check `end_fill`.


zeertzjq:
Add a test?

tomtomjhj:
added a test

zeertzjq (approved):

PR #25512, opened 2023-10-05
feat(lsp)!: `vim.lsp.inlay_hint.get()`

llllvvuu:
resolves #25069

refactor!: `vim.lsp.inlay_hint()` -> `vim.lsp.inlay_hint.enable()`

Problem
---
The LSP specification allows inlay hints to include tooltips, clickable label parts, and code actions; but Neovim provides no API to query for these.

Solution
---
Add minimal viable extension point from which plugins can query for inlay hints in a range, in order to build functionality on top of.

Possible Next Steps
---

- Add `virt_text_idx` field to `vim.fn.getmousepos()` return value, for usage in mappings of `<LeftMouse>`, `<C-LeftMouse>`, etc

PR #28227, opened 2024-04-07
feat(diagnostic): is_enabled, enable(…, enable:boolean)

justinmk:
## Problem:
`vim.diagnostic.is_disabled` and `vim.diagnostic.disable` are unnecessary and inconsistent with the "toggle" pattern established by `vim.lsp.inlay_hint`, see https://github.com/neovim/neovim/pull/25512#pullrequestreview-1676750276

As a reminder, the rationale is:
- we always need `enable()`
- we always end up needing `is_enabled()`
- "toggle" can be achieved via `enable(not is_enabled())`
- therefore,
    - `toggle()` and `disable()` are redundant
    - `is_disabled()` is a needless inconsistency

## Solution:
- Introduce `vim.diagnostic.is_enabled`, and `vim.diagnostic.enable(…, enable:boolean)`
    - Deprecate `vim.diagnostic.is_disabled`, `vim.diagnostic.disable`
- Introduce overload `enable(buf:number, enable:boolean, opts: table)`.
    - Deprecate legacy signature  `enable(buf:number, namespace:number)`

cc @glepnir @gpanders

### DIFF
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -616,23 +616,19 @@ count({bufnr}, {opts})                                *vim.diagnostic.count()*
         (`table`) Table with actually present severity values as keys (see
         |diagnostic-severity|) and integer counts as values.
 
-disable({bufnr}, {namespace})                       *vim.diagnostic.disable()*
-    Disable diagnostics in the given buffer.
+enable({bufnr}, {namespace}, {enable})               *vim.diagnostic.enable()*
+    Enables or disables diagnostics.
+
+    To "toggle", pass the inverse of `is_enabled()`: >lua
+        vim.diagnostic.enable(0, not vim.diagnostic.is_enabled())
+<
 
     Parameters: ~
-      • {bufnr}      (`integer?`) Buffer number, or 0 for current buffer. When
-                     omitted, disable diagnostics in all buffers.
-      • {namespace}  (`integer?`) Only disable diagnostics for the given
-                     namespace.
-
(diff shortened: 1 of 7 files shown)


justinmk:
> Should take an optional table for the type of diagnostic (sign, virtual text, ...).

Do you mean for both `enable()` and `is_enabled()`? I added a `opts: { ns_id: number }` arg to `enable()`, so now its signature is in place and can be extended in the future. Adding support for other filters seems reasonably out of scope for this PR.

runtime/doc/diagnostic.txt
```diff
@@ -616,23 +623,20 @@ count({bufnr}, {opts})                                *vim.diagnostic.count()*
...
-
-    Parameters: ~
-      • {bufnr}      (`integer?`) Buffer number, or 0 for current buffer. When
-                     omitted, disable diagnostics in all buffers.
-      • {namespace}  (`integer?`) Only disable diagnostics for the given
-                     namespace.
+enable({bufnr}, {enable}, {opts})                    *vim.diagnostic.enable()*
+    Enables or disables diagnostics.
 
-enable({bufnr}, {namespace})                         *vim.diagnostic.enable()*
```
justinmk, line 628: `enable({bufnr}, {enable})` is the new signature. But the old `enable({bufnr}, {namespace})` is supported for back-compat, until deprecation cycle is finished.

justinmk:
Continued in https://github.com/neovim/neovim/pull/28374

PR #28230, opened 2024-04-08
fix(tests): use more global highlight definitions

bfredl:


test/functional/ui/output_spec.lua
```diff
@@ -234,15 +223,15 @@ describe('shell command :!', function()
       -- Note: only the first example of split composed char works
       screen:expect([[
                                                              |
-        {4:                                                     }|
-        :]] .. cmd .. [[                                 |
+        {3:                                                     }|
+        :!./shell-test UTF-8                                 |
```
zeertzjq, line 227:
```suggestion
        :]] .. cmd .. [[                                 |
```

PR #22866, opened 2023-04-02
feat(health): fold successful healthchecks

ateoi:
Fold successful health checks. This feature can be disable by setting the `g:checkhealth_folding_enable` to `v:false`.

Resolves #22796

Issue #28225, opened 2024-04-07
:checkhealth unclear output


mAmineChniti:
### Problem

:checkhealth has unclear output that can't be expanded by clicking + or space or leader or anything
![WindowsTerminal_i7M13NXpqJ](https://github.com/neovim/neovim/assets/124379229/bb0358ba-64db-44ab-bea5-de7b04ebae09)

### Steps to reproduce

:checkhealth

### Expected behavior

to have a clear :checkhealth output or one that expands to view the ouput when you click a specific key

### Neovim version (nvim -v)

v0.10.0-dev-2827+g067f51e3a

### Vim (not Nvim) behaves the same?

NA

### Operating system/version

Ubuntu 23.10 on Windows 10 x86_64 (WSL 2)

### Terminal name/version

Windows Terminal

### $TERM environment variable

xterm-256color

### Installation

built from repo

zeertzjq:
It works if you `:set foldopen=all`. However that's global option and cannot be added to the ftplugin.

mAmineChniti:
@zeertzjq isn't there a way to do that by default like how can i put that in my config?

zeertzjq:
`vim.o.foldopen = 'all'`

mAmineChniti:
@zeertzjq thanks a lot

justinmk:
> can't be expanded by clicking + or space or leader or anything

Is there a reason we shouldn't have left mouse click expand folds by default?

PR #23561, opened 2023-05-09
feat(health): show attached buffers in LSP report

krady21:
Problem:
Users using `vim.lsp.start` directly (instead of nvim-lspconfig) need
more visibility for troubleshooting. For example, troubleshooting
unnecesary servers or servers that aren't attaching to expected buffers.

Solution:
Mention attached buffers in the `:checkhealth lsp` report.

Example:

    vim.lsp: Active Clients ~
    - clangd (id=1, root_dir=~/dev/neovim, attached_to=[7])
    - lua_ls (id=2, root_dir=~/dev/neovim, attached_to=[10])

Issue #22796, opened 2023-03-27
checkhealth: fold "OK" (successful) sections by default


dundargoc:
### Problem

With more neovim modules and plugins using checkhealth popping up, `:checkhealth` is starting to feel like reading CI logs with the amount of information that is presented to the user.

### Proposed solution

Automatically fold modules/sections that have passed their tests, but not for failed tests. This way the user will be shown the most important parts they need to know and fix, while also being able to inspect the information from the passing tests.

austinwhite:
I'll work on this.

zeertzjq:
How about purely informational health checks that don't have OK/FAIL status? Hiding them goes against their purpose.

dundargoc:
Hmm, that's a good point. Those should probably be shown as well.

PR #28232, opened 2024-04-08
revert: "feat(health): fold successful healthchecks #22866"

dundargoc:
This reverts commit 4382d2ed564b80944345785d780cf1b19fb23ba8.

The story for this feature was left in an incomplete state. It was never
the intention to unilaterally fold all information, only the ones that
did not contain relevant information. This feature does more harm than
good in its incomplete state.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -393,9 +393,6 @@ The following changes to existing APIs or features add new behavior.
 • |:lua| with a |[range]| executes that range in the current buffer as Lua code
   regardless of its extension.
 
-• |:checkhealth| buffer now implements |folding|. The initial folding status is
-  defined by the 'foldenable' option.
-
 • |:Man| now respects 'wrapmargin'
 
 • |gx| now uses |vim.ui.open()| and not netrw. To customize, you can redefine
--- a/runtime/ftplugin/checkhealth.vim
+++ b/runtime/ftplugin/checkhealth.vim
@@ -9,9 +9,6 @@ endif
 runtime! ftplugin/help.vim
 
 setlocal wrap breakindent linebreak
-setlocal foldexpr=getline(v:lnum-1)=~'^=\\{78}$'?'>1':(getline(v:lnum)=~'^=\\{78}'?0:'=')
-setlocal foldmethod=expr
-setlocal foldtext=v:lua.require('vim.health').foldtext()
 let &l:iskeyword='!-~,^*,^|,^",192-255'
 
 if exists("b:undo_ftplugin")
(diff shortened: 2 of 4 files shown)


justinmk:
This PR doesn't explain why `foldlevel=99` is not an acceptable way to opt-out.

The original PR explains that part of the reason was to unblock https://github.com/neovim/neovim/pull/23561 which was blocked because of fear of checkhealth being too noisy.

> It was never the intention to unilaterally fold all information, only the ones that did not contain relevant information.

According to https://github.com/neovim/neovim/issues/22796 the intention was:

> Automatically fold modules/sections that have passed their tests, but not for failed tests

Wasn't that implemented?

clason:
> Wasn't that implemented?

No, that's exactly the point. People just disabled folding.

justinmk:
> > Wasn't that implemented?
> 
> No, that's exactly the point.  People just disabled folding.

The logic collects stuff in `vim.b.failedchecks`. I see that it doesn't expand those folds (I didn't notice until now), but why can't it be changed to expand those folds? Currently it only adds "WE" to the fold text.

dundargoc:
> Why isn't `foldlevel=99` an acceptable way to opt-out? Users can put that in their config. I have it in mine.

Unsure how that helps. The problem isn't folding, it's that the folding doesn't work correctly.

> It's also not clear why the current logic can't be improved, why are we reverting the entire thing.

We could, but I was thinking maybe rethinking our approach here and starting from scratch is the way to go. There's not a lot of work we're throwing away here.

> > It was never the intention to unilaterally fold all information, only the ones that did not contain relevant information.
> 
> According to #22796 the intention was:
> 
> > Automatically fold modules/sections that have passed their tests, but not for failed tests
> 
> Wasn't that implemented?

Not really. Since everything is folded it leads to situations like https://github.com/neovim/neovim/issues/28225 which is exactly the scenario I feared would happen.

justinmk:
> Unsure how that helps. The problem isn't folding, it's that the folding doesn't work correctly.

Originally I thought this was about opting out. `foldlevel=99` expands everything by default (which is my personal preference) while allowing folds to still be "active".  

> I was thinking maybe rethinking our approach here

Ok

justinmk (approved):

dundargoc:
> The logic collects stuff in vim.b.failedchecks. I see that it doesn't expand those folds (I didn't notice until now), but why can't it be changed to expand those folds? Currently it only adds "WE" to the fold text.

It can. Someone needs to do the work though, which could be in the near future or long time from now. I don't have the energy to learn whatever eldritch vimscript abomination needed to make this work at the moment. I thought it'd make more sense to just revert it until then.

> If we revert this are we going to reopen https://github.com/neovim/neovim/issues/22796?

Maybe. I'm not certain this is how we should approach it anymore.

PR #28235, opened 2024-04-08
fix(api): do not update grid position in nvim_win_set_cursor

### LINKED ISSUE #28023: Cursor is blinking when using `nvim_win_set_cursor` in another window
### Problem

I'm implementing autoscroll feature for Test Explorer which updates very often. Scrolling this window causes blinking of the cursor:

https://github.com/neovim/neovim/assets/3128467/6b7997cd-c79f-4771-a5c7-ce920eeca183

### Steps to reproduce

```
---Autoscrolls the cursor to the provided row.
---@param row number|nil
local function autoscroll_cursor(row)
  local winnr = M.bufnr and vim.fn.win_findbuf(M.bufnr)[1]

  if winnr and config.cursor_follows_tests and row and last_set_cursor_row ~= row then
    -- V1
    -- vim.api.nvim_win_call(winnr, function()
    --   vim.cmd(tostring(row))
    --   last_set_cursor_row = row
    -- end)

    -- V2
    vim.api.nvim_win_set_cursor(winnr, { row, 0 })
    last_set_cursor_row = row
  end
end
```

I tried both approaches `V1` and `V2` and it happens for both.

### Expected behavior

Cursor should not blink

### Neovim version (nvim -v)

v0.9.5 and v0.10.0

-

### Operating system/version

macOS 14.4

### Terminal name/version

iTerm 2

xterm-256

### Installation

Homebrew/source code

luukvbaal:
Revert commit c971f538ab87b537ae4c97bd44167661c5691a2d.
Forcing grid cursor position will need a new API like originally proposed in #27858.

### DIFF
--- a/test/functional/api/window_spec.lua
+++ b/test/functional/api/window_spec.lua
@@ -219,21 +219,6 @@ describe('API/win', function()
 
       -- curwin didn't change back
       neq(win, curwin())
-
-      -- shows updated position after getchar() #20793
-      feed(':call getchar()<CR>')
-      api.nvim_win_set_cursor(win, { 1, 5 })
-      screen:expect {
-        grid = [[
-                                      |
-        {1:~                             }|*2
-        {2:[No Name]                     }|
-        prolo^gue                      |
-                                      |*2
-        {3:[No Name] [+]                 }|
-        :call getchar()               |
-      ]],
-      }
     end)
 
     it('remembers what column it wants to be in', function()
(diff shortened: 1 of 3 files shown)


test/functional/api/window_spec.lua
```diff
@@ -220,8 +220,22 @@ describe('API/win', function()
       -- curwin didn't change back
       neq(win, curwin())
 
-      -- shows updated position after getchar() #20793
+      -- does not show updated position after getchar() in non-current window
```
zeertzjq, line 223: What does "getchar() in non-current window" mean?

luukvbaal: Wrong word ordering. It's the cursor that's set in a non-current window(after a getchar()).

echasnovski:
I can confirm that it fixes [this issue](https://github.com/neovim/neovim/pull/27858#issuecomment-2043317842) (at least for 'mini.clue').

Issue #6584, opened 2017-04-25
Extra character "ff" appeared after ':' when enter ":" in normal mode


douglasyin:
- `nvim --version`:
```
NVIM v0.2.0-1451-g7f6d3d3
Build type: Debug
Compilation: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_MSGPACK_HAS_FLOAT32 -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -
...
...
Optional features included (+) or not (-): +acl   +iconv    +jemalloc +tui
For differences from Vim, see :help vim-differences

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

```
- Vim (version: ) behaves differently?

 vim version 7.4.8056 is OK.

- Operating system/version:

OSX 10.12.4

- Terminal name/version:

iterm2 Build 3.0.15

- `$TERM`:

xterm-256color

### Actual behaviour
string "ff" displayed on cursor when I enter insert mode, or enter ":" in normal mode
### Steps to reproduce using `nvim -u NORC`

```
nvim -u NORC

```
NORC mode is OK

screenshots:

<img width="260" alt="2017-04-25 13 23 35" src="https://cloud.githubusercontent.com/assets/6890111/25370091/f1583a86-29ba-11e7-870a-74753b5e1396.png">

<img width="335" alt="2017-04-25 13 24 04" src="https://cloud.githubusercontent.com/assets/6890111/25370103/0a708c76-29bb-11e7-832b-2f858c794dfb.png">

**I found this happend after commit c703d0529b3c4f6485d6ce6f37033d4658356b05**

```
commit c703d0529b3c4f6485d6ce6f37033d4658356b05
Author: Justin M. Keyes <justinkz@gmail.com>
Date:   Sat Apr 22 18:19:37 2017 +0200

    'guicursor': iTerm: Set cursor color.

    iTerm uses proprietary escape codes to set cursor color.
    https://www.iterm2.com/documentation-escape-codes.html
```

kassio:
[...]

PR #28236, opened 2024-04-08
fix(tui): don't set cursor color when there is none

### LINKED ISSUE #27611: `vim.opt.guicursor = "a:HIDDEN"` produces extra characters "ff" when entering command mode
### Problem

When entering command mode `:` extra characters "ff" show if the following options are active:
```lua
vim.api.nvim_set_hl(0, "HIDDEN", { blend = 100, nocombine = true })
vim.opt.guicursor = "a:HIDDEN"
```
This issue was initially discussed [here](https://github.com/folke/noice.nvim/issues/671) and it seems it is to be addressed to neovim core itself.

### Steps to reproduce

nvim --clean
```lua
vim.api.nvim_set_hl(0, "HIDDEN", { blend = 100, nocombine = true })
vim.opt.guicursor = "a:HIDDEN"
```
produces the flickering effect `ff` as per recording below:

https://github.com/neovim/neovim/assets/15387611/ac41e8f6-e0f0-44b6-8844-6c067e028d6c

### Expected behavior

No extra "ff" appear.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2433+g9418381cc

no, only on nightly

### Operating system/version

MacOs Monterey 12.7

### Terminal name/version

iTerm2 (Version 3.4.23)

TERM=xterm-256color

### Installation

building from source

Sam-programs:
*Presumably* Fixes: https://github.com/neovim/neovim/issues/27611
I say presumably, because i guessed since last time (https://github.com/neovim/neovim/issues/6584) writing the color from an undefined highlight caused a similar issue, it might be the same problem.

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -1237,7 +1237,7 @@ static void tui_set_mode(TUIData *tui, ModeShape mode)
       // We interpret "inverse" as "default" (no termcode for "inverse"...).
       // Hopefully the user's default cursor color is inverse.
       unibi_out_ext(tui, tui->unibi_ext.reset_cursor_color);
-    } else {
+    } else if (!tui->want_invisible && aep.rgb_bg_color >= 0) {
       char hexbuf[8];
       if (tui->set_cursor_color_as_str) {
         snprintf(hexbuf, 7 + 1, "#%06x", aep.rgb_bg_color);


Sam-programs:
Hey @gennaro-tedesco, Could you please check if this works ?

gennaro-tedesco:
Building from this PR seems to indeed solve the problem (on `iTerm2` at least)!

src/nvim/tui/tui.c
```diff
@@ -1233,6 +1233,9 @@ static void tui_set_mode(TUIData *tui, ModeShape mode)
     HlAttrs aep = kv_A(tui->attrs, c.id);
 
     tui->want_invisible = aep.hl_blend == 100;
+    if (tui->want_invisible || aep.rgb_bg_color == -1) {
+      return;
+    }
     if (!tui->want_invisible && aep.rgb_ae_attr & HL_INVERSE) {
```
zeertzjq, line 1239:
```suggestion
    if (aep.rgb_ae_attr & HL_INVERSE) {
```

src/nvim/tui/tui.c
```diff
@@ -1233,6 +1233,9 @@ static void tui_set_mode(TUIData *tui, ModeShape mode)
     HlAttrs aep = kv_A(tui->attrs, c.id);
 
     tui->want_invisible = aep.hl_blend == 100;
+    if (tui->want_invisible || aep.rgb_bg_color == -1) {
+      return;
```
zeertzjq, line 1237: I don't think it should just return when the cursor background is invalid, because it still need to set cursor shape in the code below. Maybe the code to set cursor color should be moved into a separate function.

PR #28238, opened 2024-04-08
feat: add colorscheme fakedonalds.nvim

dundargoc:
Based on the popular colorscheme
https://github.com/dundargoc/fakedonalds.nvim.

Also remove all other colorthemes as they are now redundant.


ii14 (approved):

clason:
Thank you, but we don't maintain these colorschemes. Please open a PR at vim/colorschemes instead.

PR #28255, opened 2024-04-09
 feat(lua): add `vim.fs.mkdir`

famiu:
Problem: `vim.uv.fs_mkdir` does not allow recursively creating the directory structure. https://github.com/libuv/libuv/issues/923

Solution: Add `vim.fs.mkdir` which accepts a `parents` optional argument, which creates the directory structure recursively.

(Depends on #28203 and #28187)

TODO: Add tests

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2847,6 +2847,24 @@ vim.keymap.set({mode}, {lhs}, {rhs}, {opts})                *vim.keymap.set()*
 ==============================================================================
 Lua module: vim.fs                                                    *vim.fs*
 
+vim.fs.abspath({path}, {opts})                              *vim.fs.abspath()*
+    Convert path to an absolute path. Does not check if the path exists,
+    resolve symlinks or hardlinks (including `.` and `..`), or expand
+    environment variables. If the path is already absolute, it is returned
+    unchanged.
+
+    Parameters: ~
+      • {path}  (`string`) Path
+      • {opts}  (`table?`) Optional keyword arguments:
+                • {cwd}? (`string`, default: |current-directory|) Current
+                  working directory to use as the base for the relative path.
+                  This option is ignored for C:foo\bar style paths on Windows,
+                  as those paths use the current directory of the drive
+                  specified in the path.
+
+    Return: ~
+        (`string`) Absolute path
+
 vim.fs.basename({file})                                    *vim.fs.basename()*
     Return the basename of the given path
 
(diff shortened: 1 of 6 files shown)


justinmk:
Why is this needed vs `vim.fn.mkdir()` (which currently has more features than this)?

famiu:
> > Problem: `vim.uv.fs_mkdir` does not allow recursively creating the directory structure
> 
> This doesn't explain why the alternatives such as `vim.fn.mkdir()` are insufficient, thus it doesn't explain what problem is being solved. (`vim.fn.mkdir()` also currently has more features, see also [python](https://docs.python.org/3/library/os.html#os.mkdir).)
> 

Does that not also apply to most of the things already in `vim.fs`, though? Perhaps I'm missing something?

> Of course, we eventually want `vim.fs` to be "complete". But if there's not a strong need for the "completeness", it's better to wait. Otherwise we get stuck with interfaces that were rushed.
>

Fair enough, I don't want to rush this either, fwiw, I'm fine letting this closing this PR if this isn't deemed necessary yet. Completeness was the primary reason I wanted to add this.

> Also, `vim.fs` currently does little that _writes_ the filesystem. We might want to think about how error-handling works as a general design for `vim.fs`. So that's another reason not to add kind of thing yet.
>

That is a valid concern as well. Though, if I may reframe this a bit, this could be the trigger to spark that discussion, which we would likely eventually need to have anyway. At least, that's my perspective. Does that make sense?

famiu:
> see also [python](https://docs.python.org/3/library/os.html#os.mkdir).)
> 

@justinmk Regarding this, we should follow pathlib instead: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

According to @dundargoc, pathlib is basically the successor to `os.path` and if it implements something, that's the example we should follow over `os.path`

justinmk:
> pathlib is basically the successor to `os.path` and if it implements something, that's the example we should follow over `os.path`.

Sure, but IIRC pathlib was obsessed with OOP, so ignore that part.

famiu:
> > pathlib is basically the successor to `os.path` and if it implements something, that's the example we should follow over `os.path`.
> 
> Sure, but IIRC pathlib was obsessed with OOP, so ignore that part.

Even when we ignore OOP, the fact remains that `os.path`'s `mkdir` function does not have an option to create parents, and `pathlib`'s does. I don't think that has much (if anything) to do with OOP (do correct me if I'm wrong)

barrettruth:
hey @famiu are you still working on this?  Thanks for your work in `vim.fs` up to this point.

famiu:
> hey @famiu are you still working on this?  Thanks for your work in `vim.fs` up to this point.

I am not, feel free to take over if you wish

barrettruth:
superseded by #40599

PR #28260, opened 2024-04-10
fix(prompt): emit change event for prompt newline

zeertzjq:
Co-authored-by: Ilia Choly <ilia.choly@gmail.com>


icholy:
@zeertzjq thanks for helping me out with this. I've tested your branch against the repro from #24972 and confirmed it's fixed.

PR #28273, opened 2024-04-11
fix(diagnostic): filter diagnostic by using end_lnum

### LINKED ISSUE #21541: vim.diagnostic.get incorrectly filters out multi-line diagnostics when `lnum` filter is used
### Neovim version (nvim -v)

0.8.1

N/A

### Operating system/version

Linux

### Terminal name/version

xfce4-terminal

tmux-256color

### Installation

system package manager

---

Create this file in Lua with sumneko:
```
for
test({
})
```

Observe that `vim.diagnostic.get(0)` returns 3 diagnostics (zero-indexed lines and cols):
- "Missed symbol \`in\`." from (1,4) to (1,4)
- "Unexpected \<exp\> ." from (1,4) to (2,2)
- "Miss corresponding \`end\` ." from (0,0) to (0,3)

However, when the cursor is on the last line (2 with zero-indexing) and `vim.diagnostic.get(0, { lnum = 2 })` is called, no diagnostics are returned. We should expect to see "Unexpected <exp>." returned here.

I believe the issue is here: https://github.com/neovim/neovim/blob/bf9ad5db962509f3434726b7ad84d31d002fb8a3/runtime/lua/vim/diagnostic.lua#L410
That line uses an exact equality check for `lnum` and doesn't account for being in the middle of a range.

glepnir:
Problem: opts.lnum in vim.diagnostic.get doesn't support using end_lnum.

Solution: add end_lnum support.

fix #21541

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -706,7 +706,7 @@ local function get_diagnostics(bufnr, opts, clamp)
   ---@param b integer
   ---@param d vim.Diagnostic
   local function add(b, d)
-    if not opts.lnum or d.lnum == opts.lnum then
+    if not opts.lnum or (opts.lnum >= d.lnum and opts.lnum <= (d.end_lnum or d.lnum)) then
       if clamp and api.nvim_buf_is_loaded(b) then
         local line_count = buf_line_count[b] - 1
         if
@@ -1129,7 +1129,7 @@ end
 --- Limit diagnostics to one or more namespaces.
 --- @field namespace? integer[]|integer
 ---
---- Limit diagnostics to the given line number.
+--- Limit diagnostics to those spanning the specified line number.
 --- @field lnum? integer
 ---
 --- See |diagnostic-severity|.
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -373,7 +373,8 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
     Fields: ~
       • {namespace}?  (`integer[]|integer`) Limit diagnostics to one or more
                       namespaces.
-      • {lnum}?       (`integer`) Limit diagnostics to the given line number.
+      • {lnum}?       (`integer`) Limit diagnostics to those spanning the
+                      specified line number.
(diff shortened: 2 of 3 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -706,7 +706,7 @@ local function get_diagnostics(bufnr, opts, clamp)
   ---@param b integer
   ---@param d vim.Diagnostic
   local function add(b, d)
-    if not opts.lnum or d.lnum == opts.lnum then
+    if not opts.lnum or d.lnum == opts.lnum or d.end_lnum == opts.lnum then
```
gpanders, line 709: Rather than comparing directly to `lnum` or `end_lnum`, `opts.lnum` should get any diagnostic that spans that line number. So it should be (I think):

```lua
(opts.lnum >= d.lnum) and (opts.lnum <= (d.end_lnum or d.lnum))
``` 

So if a diagnostic has _only_ a `lnum` (no `end_lnum`) this works as before. If `lnum` and `end_lnum` are equal, this works as before. But if `end_lnum > lnum`, then any `opts.lnum` that lies within the range would be returned.

The docstring for `GetOpts` should be updated too.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1129,7 +1129,10 @@ end
 --- Limit diagnostics to one or more namespaces.
 --- @field namespace? integer[]|integer
 ---
---- Limit diagnostics to the given line number.
+--- Limit diagnostics to those spanning the specified line number.
+--- Diagnostics are included if the specified line number falls between
+--- the `lnum` and `end_lnum` of the diagnostic. If `end_lnum` is not provided,
+--- it is assumed to be equal to `lnum`.
```
gpanders, line 1135: "Provided" here implies that `end_lnum` is given in the options, rather than something that exists in the diagnostic. 

Personally I don't think we need to leak implementation details here, "spanning the specified line number" is sufficient to describe how this works.

```suggestion
```

gpanders (dismissed):

runtime/lua/vim/diagnostic.lua
glepnir: removed.

gpanders (approved):

PR #28279, opened 2024-04-11
refactor(argedit): check duplicate and insert in correct index

### LINKED ISSUE #11705: :argedit of file in argument list -- behaviour has changed
### Steps to reproduce using `nvim -u NONE`

```
nvim -u NONE a.txt b.txt
:echo argv()
:argedit b.txt
:echo argv()
```

### Actual behaviour
Argument list originally contained `['a.txt', 'b.txt']`
Now contains `['a.txt', 'b.txt', 'b.txt']`

### Expected behaviour
Should contain  `['a.txt', 'b.txt']` both times.

This would match the documentation of `argedit` that says:
`When {name} already exists in the argument list, this entry is edited.`
			
## Behaviour introduced in bbdb074 (vim-patch:8.0.0721)
Seems that commit introduced a new feature that `:argedit` can add multiple
files to the argument list at once.

This is done by passing the arguments to `do_arglist` with the `AL_ADD` flag.
That essentially does the same as `:argadd`.

`:argadd` documentation says `There is no check for duplicates, it is possible
to add a file to the argument list twice`, while `:argedit` documentation says
`When {name} already exists in the argument list, this entry is edited.`

Previously the behaviour of `:argedit` was to switch to the relevant argument
without adding a duplicate to the argument list.

(N.B -- mainly for extra cross-reference search terms -- the plugin `vimfindsme` relies on the previous behaviour of argedit for :VFMArgument)

glepnir:
Problem: 
1. when a filename already exists in the argument list, that entry should be edited. However, argedit does not check for duplicates and instead inserts a new one.  duplicate items not make sense in arglist.

2. it does work like argadd. For instance, in the test case Test_argedit,  after `argedit a` current file is `a`, then executing `argedit C D` should insert C and D after a. However, the expected result is `a b C D`, similar to the behavior observed with argadd in help doc where the current file is b: executing `:argadd x` results in the argument list `a b x c`.  `x` after current `b`.

Solution: Refactor argedit to avoid inserting duplicate items and to insert new items at the correct position.

Note breakchange. Not compatible with vim.  Maybe some discussion is needed. Does duplicate files items in arglist make any sense. i don't use arg relate command. but i see some issues both in vim/neovim. so i guess it make more confuse.

Fix #11705

### DIFF
--- a/runtime/doc/editing.txt
+++ b/runtime/doc/editing.txt
@@ -637,9 +637,6 @@ list of the current window.
 			This is like using |:argadd| and then |:edit|.
 			Spaces in filenames have to be escaped with "\".
 			[count] is used like with |:argadd|.
-			If the current file cannot be |abandon|ed {name}s will
-			still be added to the argument list, but won't be
-			edited. No check for duplicates is done.
 			Also see |++opt| and |+cmd|.
 
 :[count]arga[dd] {name} ..			*:arga* *:argadd* *E479*
--- a/src/nvim/arglist.c
+++ b/src/nvim/arglist.c
@@ -414,6 +414,21 @@ static void arglist_del_files(garray_T *alist_ga)
   ga_clear(alist_ga);
 }
 
+/// Expand the file names in the argument list based on wildcards.
+static bool expand_arglist_files(garray_T *ga, int *count, char ***files)
+{
+  int result;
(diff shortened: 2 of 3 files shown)


glepnir:
It is closed in vim. . If we also consider duplicate elements in arglist is makes sense also close this pr..

PR #28289, opened 2024-04-12
fix(defaults): only repeat macro for each selected line if linewise

### LINKED ISSUE #28287: When applying a macro in visual mode, the visual selection is lost
### Problem

When applying a macro while in visual mode, the macro _is_ applied but in doing so the visual selection is lost, almost as if applying the macro forces NeoVim into normal mode first. This makes it impossible to use macros for recording jumps (e.g. "jump to the next `)`") and then apply those to a visual selection.

### Steps to reproduce

This reproduces with `nvim --clean`. First, create the file `test.txt` with these contents:

```
foo(100, 200, 300)
foo(100000, 200000, 300000)
foo(1, 2, 3)
```

Then run `nvim --clean test.txt`. Perform the following steps:

1. Place the cursor on the first `f` (the default)
2. Press `v` to enter visual mode, followed by `e` to select the first `foo`
3. Press `qq` to record a macro in the `q` register
4. Press `t)` to jump to the last character before the `)`, note that the visual selection is extended
5. Press `q` to stop recording the macro
6. Exit visual mode
7. Move to `f` on line 2, press `v` to enter visual mode and `e` to select `foo`
8. Press `@q` and observe that while the jump _is_ performed, the visual selection is lost

This can be observed in this recording:

[Screencast from 2024-04-12 01-51-34.webm](https://github.com/neovim/neovim/assets/86065/2ff561d7-4877-4459-8afc-729da7b37c4a)

### Expected behavior

The visual selection is retained.

### Neovim version (nvim -v)

aa56370f

No, it (9.1.264) behaves as expected

### Operating system/version

Fedora 39

### Terminal name/version

Ghostty

xterm-ghostty

### Installation

Git

zeertzjq:
As mentioned in #28287, repeating a macro for each selected line doesn't
really make sense in non-linewise Visual mode.

Fix #28287

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -317,8 +317,8 @@ The following new APIs and features were added.
 
 • |nvim_input_mouse()| supports mouse buttons "x1" and "x2".
 
-• |v_Q-default| and |v_@-default| repeat a register for each line of a visual
-  selection.
+• |v_Q-default| and |v_@-default| repeat a register for each line of a linewise
+  visual selection.
 
 • |vim.diagnostic.count()| returns the number of diagnostics for a given
   buffer and/or namespace, by severity. This is a faster alternative to
--- a/runtime/doc/repeat.txt
+++ b/runtime/doc/repeat.txt
@@ -149,8 +149,8 @@ q			Stops recording.
 @@			Repeat the previous @{0-9a-z":*} [count] times.
 
 							*v_@-default*
-{Visual}@{0-9a-z".=*+}	In Visual mode, execute the contents of the register
-{Visual}@@		but for each selected line.
+{Visual}@{0-9a-z".=*+}	In linewise Visual mode, execute the contents of the
+{Visual}@@		register for each selected line.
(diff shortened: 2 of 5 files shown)


runtime/doc/news.txt
```diff
@@ -317,8 +317,8 @@ The following new APIs and features were added.
 
 • |nvim_input_mouse()| supports mouse buttons "x1" and "x2".
 
-• |v_Q-default| and |v_@-default| repeat a register for each line of a visual
-  selection.
+• |v_Q-default| and |v_@-default| repeat a register for each line of a linewise
+  visual selection.
```
justinmk, line 321: 👍  constraining to linewise is a useful, intuitive approach for various mappings. For example I have mapped linewise visual `D` to diff visual selections: https://github.com/justinmk/config/blob/94629007e99ad99b40166b58c7a32ba98576cb47/.config/nvim/init.vim#L257-L258

We should leverage this for more cases in the future.

justinmk (approved):

PR #28301, opened 2024-04-12
fix(diagnostic): open_float on multi-line diagnostics

### LINKED ISSUE #21949: vim.diagnostic.open_float does not work for lines after first one in multi-line diagnostics
### Describe the bug

If you have a multi-line diagnostic, `vim.diagnostic.open_float` works on the first line but fails on the subsequent lines.

### Steps to reproduce

1. In `nvim --clean`, open a new buffer and paste the following contents.
```
local ns = vim.api.nvim_create_namespace('diagnostics')
vim.diagnostic.reset(ns, 0)
vim.diagnostic.set(ns, 0, { {
  lnum = 0,
  col = 0,
  end_lnum = 2,
  end_col = 0,
  message = 'Two-line diagnostic'
} })
```
2. Save it
3. Source it
4. Observe that the first two lines are underlined
5. Observe that `:lua vim.diagnostic.open_float` works on the first line.
6. Observe that `:lua vim.diagnostic.open_float` does not work on the second line.

### Expected behavior

`vim.diagnostic.open_float` should work anywhere inside the diagnostic.

### Neovim version (nvim -v)

0.8.2

N/A

### Operating system/version

Arch Linux 5.19.9

### Terminal name/version

xfce4-terminal 1.0.4

tmux-256color

### Installation

system package manager

glepnir:
Problem: when diagnostic have a range of line, open_float not work.

Solution: filter diagnsotic by a line number range.

Fix #21949

### DIFF
--- a/test/functional/lua/diagnostic_spec.lua
+++ b/test/functional/lua/diagnostic_spec.lua
@@ -2444,6 +2444,47 @@ describe('vim.diagnostic', function()
       ]]
       )
     end)
+
+    it('works for multi-line diagnostics #21949', function()
+      -- open float failed non diagnostic lnum
+      eq(
+        vim.NIL,
+        exec_lua [[
+        local diagnostics = {
+          make_error("Error in two lines lnum is 1 and end_lnum is 2", 1, 1, 2, 3),
+        }
+        local winids = {}
+        vim.api.nvim_win_set_buf(0, diagnostic_bufnr)
+        vim.diagnostic.set(diagnostic_ns, diagnostic_bufnr, diagnostics)
+        local _, winnr = vim.diagnostic.open_float(0, { header = false })
+        return winnr
+      ]]
+      )
+
+      -- can open a float window on lnum 1
+      eq(
+        { '1. Error in two lines lnum is 1 and end_lnum is 2' },
+        exec_lua [[
+        vim.api.nvim_win_set_cursor(0, {2, 0})
+        local float_bufnr, winnr = vim.diagnostic.open_float(0, { header = false })
+        local lines = vim.api.nvim_buf_get_lines(float_bufnr, 0, -1, false)
+        vim.api.nvim_win_close(winnr, true)
+        return lines
+      ]]
+      )
+
+      -- can open a float window on end_lnum 2
+      eq(
+        { '1. Error in two lines lnum is 1 and end_lnum is 2' },
+        exec_lua [[
+        vim.api.nvim_win_set_cursor(0, {3, 0})
+        local float_bufnr, winnr = vim.diagnostic.open_float(0, { header = false })
+        local lines = vim.api.nvim_buf_get_lines(float_bufnr, 0, -1, false)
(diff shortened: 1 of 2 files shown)


test/functional/lua/diagnostic_spec.lua
```diff
@@ -2444,6 +2444,36 @@ describe('vim.diagnostic', function()
...
+        vim.api.nvim_win_set_cursor(0, {3, 0})
+        local _, w3 = vim.diagnostic.open_float(0, { header = false })
+        if w3 then
+          table.insert(winids, w3)
+          vim.api.nvim_win_close(w3, true)
+        end
+        return #winids
+      ]]
+      eq(2, result)
+    end)
```
gpanders, line 2476: The fix looks right, but I'm not sure I follow what this test is doing? For one thing, I don't see any multi-line diagnostics. Why is the test result expecting 2 elements in `winids`? Can you either rewrite the test to be a little clearer, or at least add some clarifying comments to help understand the flow of the test?

test/functional/lua/diagnostic_spec.lua
```diff
@@ -2444,6 +2444,36 @@ describe('vim.diagnostic', function()
       ]]
       )
     end)
+
+    it('works for multi-line diagnostics', function()
```
gpanders, line 2448:
```suggestion
    it('works for multi-line diagnostics #21949', function()
```

test/functional/lua/diagnostic_spec.lua
glepnir: make_error the line is 0 row and 1 row mean line 1 and line 2. put cursor in line 1 has w1 in line 2 has w2 in line 3 nothing happen.

glepnir: updated and more clear now ❓

gpanders: Thanks for clarifying, I still think this test is needlessly obscure though. What does the result being `2` mean? You can use `exec_lua` and `eq` more than once in a single test, perhaps break the large `exec_lua` up into smaller bits, each of which tests some behavior (so it's clear exactly what's being tested).

glepnir: yup and done.

test/functional/lua/diagnostic_spec.lua
```diff
@@ -2444,6 +2444,47 @@ describe('vim.diagnostic', function()
...
+        { '1. Error in two lines lnum is 1 and end_lnum is 2' },
+        exec_lua [[
+        vim.api.nvim_win_set_cursor(0, {3, 0})
+        local float_bufnr, winnr = vim.diagnostic.open_float(0, { header = false })
+        local lines = vim.api.nvim_buf_get_lines(float_bufnr, 0, -1, false)
+        vim.api.nvim_win_close(winnr, true)
+        return lines
+      ]]
+      )
+    end)
```
gpanders, line 2487: Much clearer, thanks!

gpanders (approved):

PR #28302, opened 2024-04-12
fix(lsp): check the markdown content is empty whic only wrapped by codeblock

### LINKED ISSUE #27956: vim.lsp.buf.hover() .  'width' key must be a positive Integer
### Problem

many release versions are affected, including latest stable v0.9.5 .
a mitigation is present on master, see below

`vim.lsp.buf.hover()` may trigger
```
Error executing vim.schedule lua callback: $HOME/.local/share/nvim/runtime/lua/vim/lsp/util.lua:1721: 'width' key must be a positive Integer        
stack traceback:
        [C]: in function 'nvim_open_win'
        $HOME/.local/share/nvim/runtime/lua/vim/lsp/util.lua:1721: in function 'handler'
        $HOME/.local/share/nvim/runtime/lua/vim/lsp.lua:1502: in function ''
```

related:
* https://github.com/neovim/neovim/discussions/27434
* https://github.com/neovim/nvim-lspconfig/issues/3036

`minimal.lua`

```
-- Run this file as `nvim --clean -u minimal.lua`

for name, url in pairs {
  -- ADD PLUGINS _NECESSARY_ TO REPRODUCE THE ISSUE, e.g:
  -- some_plugin = 'https://github.com/author/plugin.nvim'
  lspconfig = 'https://github.com/neovim/nvim-lspconfig.git'
} do
  end
end

-- ADD INIT.LUA SETTINGS _NECESSARY_ FOR REPRODUCING THE ISSUE
local lspconfig = require("lspconfig")
lspconfig.pylsp.setup{}

-- condensed from: https://github.com/neovim/nvim-lspconfig#suggested-configuration
  group = vim.api.nvim_create_augroup('UserLspConfig', {}),
  callback = function(ev)
    local opts = { buffer = ev.buf }
    vim.keymap.set('n', 'K', vim.lsp.buf.hover, opts)
  end
})
```

`repro.py`

```
from datetime import date

# REPRO
# * go to last line
# * try `vim.lsp.buf.hover()` on parts of right side. works
# * try it on `d`, fails

d = date.fromtimestamp(0)
```

### Expected behavior

empty box is shown or no box at all

the issue is not present on current master, it's fixed (by accident) by cfd4a9dfaf.
with that commit, an empty box is shown.

### Neovim version (nvim -v)

v0.10.0.-dev-763+gc01e624b07

python-language-server 1.10.1 / python3.12

### Operating system/version

linux/rolling

### Log file

_No response_

glepnir:
Problem: pylsp return  hover result be like only an empty line which wrapped by codeblock.

Solution: if it's only have an empty line wrapped by codeblock, does not show a float window

Fix #27956

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -857,6 +857,7 @@ end
 ---@see https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover
 function M.convert_input_to_markdown_lines(input, contents)
   contents = contents or {}
+  local code_block = '```'
   -- MarkedString variation 1
   if type(input) == 'string' then
     list_extend(contents, split_lines(input))
@@ -868,9 +869,9 @@ function M.convert_input_to_markdown_lines(input, contents)
       list_extend(contents, split_lines(value))
       -- MarkupString variation 2
     elseif input.language then
-      table.insert(contents, '```' .. input.language)
+      table.insert(contents, code_block .. input.language)
       list_extend(contents, split_lines(input.value or ''))
-      table.insert(contents, '```')
+      table.insert(contents, code_block)
       -- By deduction, this must be MarkedString[]
     else
       -- Use our existing logic to handle MarkedString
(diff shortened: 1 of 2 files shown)


clason:
We need to refactor/remove this codepath anyway based on the new markdown renderer and capabilities due to @MariaSolOs

runtime/lua/vim/lsp/util.lua
```diff
@@ -866,11 +867,21 @@ function M.convert_input_to_markdown_lines(input, contents)
     if input.kind then
       local value = input.value or ''
       list_extend(contents, split_lines(value))
+      -- Check that the content contains blank lines and is wrapped by codeblock
```
MariaSolOs, line 870: This feels like an extremely specific fix. Why not just sent contents to `{}` if it consists of a single empty line? Having the length check feels a bit too much.

justinmk:
> We need to refactor/remove this codepath anyway based on the new markdown renderer and capabilities due to

Let's do that instead

glepnir:
emm this is result check not relate ts markdown render ,  maybe can move logic  to handler

PR #28323, opened 2024-04-14
test: fail if .nvimlog is not empty

dundargoc:
This will prevent adding passing tests which logs errors in .nvimlog.

### DIFF
--- a/test/functional/lua/hl_spec.lua
+++ b/test/functional/lua/hl_spec.lua
@@ -9,12 +9,6 @@ local command = n.command
 local clear = n.clear
 local api = n.api
 
-local testlog = 'Xtest-highlight-log'
-
-after_each(function()
-  os.remove(testlog)
-end)
-
 describe('vim.hl.range', function()
   local screen
 
@@ -179,7 +173,6 @@ describe('vim.hl.on_yank', function()
   end)
 
   it('does not show in another window', function()
-    clear({ env = { NVIM_LOG_FILE = testlog } })
     command('vsplit')
     exec_lua(function()
       vim.api.nvim_buf_set_mark(0, '[', 1, 1, {})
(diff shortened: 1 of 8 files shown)


dundargoc:
Wait, we have a `assert_log` with an inverse flag as well as `assert_nolog`. Think we can just remove `assert_nolog` here.

dundargoc:
Progress report: there is something sussy wussy happening with tests where the logfile is created *after* the test is finished: https://asciinema.org/a/wewPiHmK6jirJBisAdyDmxJuj . I've asked for backup to take a look at this when they have time.

zeertzjq:
How about just canceling the timer instead: #32508

test/functional/lua/hl_spec.lua
```diff
@@ -9,6 +9,12 @@ local command = n.command
 local clear = n.clear
 local api = n.api
 
+local testlog = 'Xtest-highlight-log'
+
+after_each(function()
+  os.remove(testlog)
+end)
+
```
zeertzjq, line 17: remove

test/functional/lua/hl_spec.lua
```diff
@@ -173,6 +179,7 @@ describe('vim.hl.on_yank', function()
   end)
 
   it('does not show in another window', function()
+    clear({ env = { NVIM_LOG_FILE = testlog } })
```
zeertzjq, line 182: remove

test/functional/lua/hl_spec.lua
```diff
@@ -191,6 +198,7 @@ describe('vim.hl.on_yank', function()
   end)
 
   it('removes old highlight if new one is created before old one times out', function()
+    clear({ env = { NVIM_LOG_FILE = testlog } })
```
zeertzjq, line 201: remove

Issue #23235, opened 2023-04-21
release checklist 0.10


justinmk:
(previous: https://github.com/neovim/neovim/issues/20434)

- [Release guidelines](https://github.com/neovim/neovim/blob/master/MAINTAIN.md#release-policy)
- [Dependencies](https://github.com/neovim/neovim/blob/master/MAINTAIN.md#third-party-dependencies)

# Blocking issues

- https://github.com/neovim/neovim/pull/28227
- https://github.com/neovim/neovim/pull/26138
- https://github.com/neovim/neovim/pull/28450
- https://github.com/neovim/neovim/pull/28440
- https://github.com/neovim/neovim/issues/27953
- https://github.com/neovim/neovim/issues/28548
- https://github.com/neovim/neovim/pull/28432
- (partial) https://github.com/neovim/neovim/issues/24572

# Pre-release (early)

- [x] run `scripts/download-unicode-files.sh`
- [x] Bump deps in [cmake.deps/deps.txt](https://github.com/neovim/neovim/blob/master/cmake.deps/deps.txt)
    - [x] run `scripts/bump_deps.lua` (luajit, luv, libuv, treesitter)
    - [x] treesitter parsers (C, Lua, Viml, Vimdoc, Query)
        - [x] update queries from nvim-treesitter
    - [x] lpeg (also in `neovim/deps`)
    - [x] libvterm (also in `neovim/deps`)
    - [x] msgpack
    - [x] [wintools](https://github.com/neovim/neovim/blob/9319d97679053a219b1701d70abce414d0f6a8b0/cmake.deps/deps.txt#L28-L36)
- [x] check s930x CI logs for errors pertaining to big-endian systems
    - [run1](https://github.com/neovim/neovim/actions/runs/9062490238/job/24896468103?pr=28730) [run2](https://github.com/neovim/neovim/actions/runs/9062490238/job/24896468325?pr=28730) (result: no new regressions)

# Pre-release

- [x] Remove deprecated functions marked for this release
    - [x] `Iter:nextback() Iter:peekback() Iter:skipback() Iter:nthback(n)`
    - [x] https://github.com/neovim/neovim/pull/28725
- [x] Clean up `news.txt` (sorting, [phrasing](https://www.youtube.com/watch?v=Qs8Gohi1e04))
- [x] Merge `nvim.appdata.xml` changes from the maintenance branch. https://github.com/neovim/neovim/pull/28769
- [x] Run `scripts/release.sh`

# Post-release

[...]

PR #28324, opened 2024-04-14
feat(defaults): use ripgrep (rg) for 'grepprg' if available

LunarLambda:
Discussed as an idea in https://github.com/neovim/neovim/discussions/28296.

Changes the defaults for `'grepprg'` and `'grepformat'` to use ripgrep and its `--vimgrep` mode, if available on the system.

Points discussed in the thread:

- Zero configuration: This PR provides a better default for an existing vim feature (`:grep`), which the user is free to configure themselves as needed. It adds no additional configuration settings or hooks.
- Low to zero maintenance: The change consists of a single `executable()` check and two option changes. It is very unlikely that ripgrep will experience breaking changes in its CLI that would necessitate adjusting the code or adding version checking.
- Performance: A basic benchmark (running a recursive, fixed string search on the neovim repository) shows significant performance improvements over GNU grep (few seconds vs <1 second).
- Precedence: Many popular plugins already integrate with ripgrep for searching, and ripgrep itself provides features specifically for integrating with vim (`--vimgrep` format, `--smart-case` option). This makes it a promising candidate for a place where Neovim can match the pulse of the community.

This is my first PR to adjust core Neovim code, as such there's a few points I am seeking feedback on:

- Is this change wanted? The discussion thread concluded that the change is promising, but may not be wanted by Neovim users.
- Is the documentation appropriate? I Added a small note to the documentation of `'grepprg'` that points the user toward ripgrep and why they may want to use it. Is it clear enough that this default only takes effect if the *user makes sure* that the `rg` binary is found by Neovim?
- Does this need extra testing? @clason mentioned a `:checkhealth` test, I'm unsure what that should look like.

I'm happy to make any necessary adjustments.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -370,6 +370,7 @@ The following changes to existing APIs or features add new behavior.
     correctly without it. (Use |gF| for filepaths suffixed with ":line:col").
   • 'comments' includes "fb:•".
   • 'shortmess' includes the "C" flag.
+  • 'grepprg' defaults to using ripgrep if available.
   • Automatic linting of treesitter query files (see |ft-query-plugin|).
     Can be disabled via: >lua
       vim.g.query_lint_on = {}
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -2851,9 +2851,10 @@ A jump table for the options with a short description can be found at |Q_op|.
 	This is a scanf-like string that uses the same format as the
 	'errorformat' option: see |errorformat|.
 
+	If ripgrep ('grepprg') is available, this option defaults to `%f:%l:%c:%m`.
+
 						*'grepprg'* *'gp'*
-'grepprg' 'gp'		string	(default "grep -n ",
-                                 Unix: "grep -n $* /dev/null")
+'grepprg' 'gp'		string	(default see below)
 			global or local to buffer |global-local|
 	Program to use for the |:grep| command.  This option may contain '%'
 	and '#' characters, which are expanded like when used in a command-
@@ -2870,6 +2871,16 @@ A jump table for the options with a short description can be found at |Q_op|.
 	apply equally to 'grepprg'.
 	This option cannot be set from a |modeline| or in the |sandbox|, for
 	security reasons.
+	This option defaults to:
+	- `rg --vimgrep -uuu $* ...` if ripgrep is available (|:checkhealth|),
+	- `grep -n $* /dev/null` on Unix,
+	- `findstr /n $* nul` on Windows.
+	Ripgrep can perform additional filtering such as using .gitignore rules
+	and skipping hidden or binary files. This is disabled by default (see the -u option)
+	to more closely match the behaviour of standard grep.
+	You can make ripgrep match Vim's case handling using the
+	-i/--ignore-case and -S/--smart-case options.
+	An |OptionSet| autocmd can be used to set it up to match automatically.
 
 				*'guicursor'* *'gcr'* *E545* *E546* *E548* *E549*
 'guicursor' 'gcr'	string	(default "n-v-c-sm:block,i-ci-ve:ver25,r-cr-o:hor20")
--- a/runtime/lua/nvim/health.lua
+++ b/runtime/lua/nvim/health.lua
@@ -383,6 +383,19 @@ local function check_terminal()
   end
 end
 
+local function check_external_tools()
+  health.start('External Tools')
+
+  if vim.fn.executable('rg') == 1 then
+    local rg = vim.fn.exepath('rg')
+    local cmd = 'rg -V'
+    local out = vim.fn.system(vim.fn.split(cmd))
+    health.ok(('%s (%s)'):format(vim.trim(out), rg))
+  else
+    health.warn('ripgrep not available')
+  end
+end
+
 function M.check()
   check_config()
   check_runtime()
@@ -390,6 +403,7 @@ function M.check()
   check_rplugin_manifest()
   check_terminal()
   check_tmux()
+  check_external_tools()
 end
 
 return M
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -486,3 +486,11 @@ if tty then
     end
   end
 end
+
+--- Default 'grepprg' to ripgrep if available.
+if vim.fn.executable('rg') == 1 then
+  -- Match :grep default, otherwise rg searches cwd by default
+  -- Use -uuu to make ripgrep not do its default filtering
+  vim.o.grepprg = 'rg --vimgrep -uuu $* ' .. (vim.fn.has('unix') == 1 and '/dev/null' or 'nul')
+  vim.o.grepformat = '%f:%l:%c:%m'
+end
--- a/src/nvim/options.lua
+++ b/src/nvim/options.lua
@@ -3368,6 +3368,8 @@ return {
         Format to recognize for the ":grep" command output.
         This is a scanf-like string that uses the same format as the
         'errorformat' option: see |errorformat|.
+
+        If ripgrep ('grepprg') is available, this option defaults to `%f:%l:%c:%m`.
       ]=],
       full_name = 'grepformat',
       list = 'onecomma',
@@ -3382,8 +3384,7 @@ return {
         condition = 'MSWIN',
         if_false = 'grep -n $* /dev/null',
         if_true = 'findstr /n $* nul',
-        doc = [["grep -n ",
-           Unix: "grep -n $* /dev/null"]],
+        doc = [[see below]],
       },
       desc = [=[
         Program to use for the |:grep| command.  This option may contain '%'
@@ -3401,6 +3402,16 @@ return {
         apply equally to 'grepprg'.
         This option cannot be set from a |modeline| or in the |sandbox|, for
         security reasons.
+        This option defaults to:
+        - `rg --vimgrep -uuu $* ...` if ripgrep is available (|:checkhealth|),
+        - `grep -n $* /dev/null` on Unix,
+        - `findstr /n $* nul` on Windows.
+        Ripgrep can perform additional filtering such as using .gitignore rules
+        and skipping hidden or binary files. This is disabled by default (see the -u option)
+        to more closely match the behaviour of standard grep.
+        You can make ripgrep match Vim's case handling using the
+        -i/--ignore-case and -S/--smart-case options.
+        An |OptionSet| autocmd can be used to set it up to match automatically.
       ]=],
       expand = true,
       full_name = 'grepprg',
--- a/runtime/doc/vim_diff.txt
+++ b/runtime/doc/vim_diff.txt
@@ -52,6 +52,7 @@ Defaults					            *nvim-defaults*
 - 'encoding' is UTF-8 (cf. 'fileencoding' for file-content encoding)
 - 'fillchars' defaults (in effect) to "vert:│,fold:·,foldsep:│"
 - 'formatoptions' defaults to "tcqj"
+- 'grepprg' defaults to using ripgrep if available
 - 'hidden' is enabled
 - 'history' defaults to 10000 (the maximum)
 - 'hlsearch' is enabled
--- a/runtime/lua/vim/_meta/options.lua
+++ b/runtime/lua/vim/_meta/options.lua
@@ -2625,6 +2625,8 @@ vim.go.gd = vim.go.gdefault
 --- This is a scanf-like string that uses the same format as the
 --- 'errorformat' option: see `errorformat`.
 ---
+--- If ripgrep ('grepprg') is available, this option defaults to `%f:%l:%c:%m`.
+---
 --- @type string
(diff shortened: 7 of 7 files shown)


src/nvim/options.lua
```diff
@@ -3401,6 +3401,13 @@ return {
         apply equally to 'grepprg'.
         This option cannot be set from a |modeline| or in the |sandbox|, for
         security reasons.
+        If ripgrep is available (|executable()|) then this option and |'grepformat'|
+        default to "rg --vimgrep ..." and "%f:%l:%c:%m" respectively.
+        Ripgrep is a faster, recursive grep that respects gitignore rules
+        and skips hidden and binary files by default, and supports most of the common grep options.
```
clason, line 3407: No need to explain what `ripgrep` is -- but the different defaults (`.gitignore`) should be called out (and possibly disabled by default; some people -- myself included -- prefer the `grep` default).

src/nvim/options.lua
```diff
@@ -3401,6 +3401,13 @@ return {
         apply equally to 'grepprg'.
         This option cannot be set from a |modeline| or in the |sandbox|, for
         security reasons.
+        If ripgrep is available (|executable()|) then this option and |'grepformat'|
+        default to "rg --vimgrep ..." and "%f:%l:%c:%m" respectively.
```
clason, line 3405:
```suggestion
        default to `rg --vimgrep ...` and `%f:%l:%c:%m` respectively.
```

src/nvim/options.lua
```diff
@@ -3382,7 +3382,7 @@ return {
         condition = 'MSWIN',
         if_false = 'grep -n $* /dev/null',
         if_true = 'findstr /n $* nul',
-        doc = [["grep -n ",
+        doc = [[Windows: "findstr /n $* nul",
            Unix: "grep -n $* /dev/null"]],
```
clason, line 3386: I don't think that's correct now (since the default can be `ripgrep`). Maybe just "see below" if it's too much to explain here.

clason:
> a :checkhealth test, I'm unsure what that should look like.

I would recommend a new section on "optional tools" that checks `vim.fn.executable('rg') == 1` and prints `OK` if so and `WARN` if not. Bonus points for adding 
* `which rg` and
* `rg --version` 

output. (We can extend this later with `curl` and `git` and `tree-sitter`, once we use them.)

LunarLambda:
Have adjusted the documentation, turned off the default filtering (with a note in the documentation hinting the user they may want to re-enable it if they wish), and added a rudimentary healthcheck.

runtime/lua/nvim/health.lua
```diff
@@ -383,13 +383,27 @@ local function check_terminal()
   end
 end
 
+local function check_external_tools()
+  health.start('External Tools')
```
justinmk, line 387: Should be in "Performance" section unless @clason strongly objects

clason: Not strongly, but I would like to extend this section to other "optional tools" (like `curl`).

runtime/doc/options.txt
```diff
@@ -2870,6 +2869,16 @@ A jump table for the options with a short description can be found at |Q_op|.
 	apply equally to 'grepprg'.
 	This option cannot be set from a |modeline| or in the |sandbox|, for
 	security reasons.
+	This option defaults to `findstr /n $* nul` on Windows
+	and `grep -n $* /dev/null` elsewhere.
+	Additionally, if ripgrep is available (|executable()|) then this option
+	and |'grepformat'| default to `rg --vimgrep -uuu ...` and `%f:%l:%c:%m` respectively.
```
dundargoc, line 2875: I think listing the order of tools in reverse priority extremely unintuitive. I think it's better to present it like `:h clipboard-tool` to eliminate any ambiguities. Something like (adjust as needed):

  - ripgrep
  - findstr (Windows)
  - grep (Unix)

justinmk: some structure in the text would also be helpful.

LunarLambda:
I have overhauled the documentation as requested. I've kept the health check in the proposed 'External Tools' section for now.

runtime/lua/nvim/health.lua
```diff
@@ -383,13 +383,27 @@ local function check_terminal()
...
+
+  if vim.fn.executable('rg') == 1 then
+    local rg = vim.fn.exepath('rg')
+    local cmd = 'rg -V'
+    local out = vim.fn.system(vim.fn.split(cmd))
+    health.ok(('%s (%s)'):format(vim.trim(out), rg))
```
clason, line 393: Follow-up: the location of the used `rg` would be _really_ useful (`vim.fn.exepath()`) to debug `$PATH` issues.

runtime/lua/vim/_defaults.lua
```diff
@@ -486,3 +486,11 @@ if tty then
     end
   end
 end
+
+--- Default grepprg to ripgrep if available.
```
justinmk, line 490:
```suggestion
--- Default 'grepprg' to ripgrep if available.
```

src/nvim/options.lua
```diff
@@ -3368,6 +3368,8 @@ return {
         Format to recognize for the ":grep" command output.
         This is a scanf-like string that uses the same format as the
         'errorformat' option: see |errorformat|.
+
+        If ripgrep (|'grepprg'|) is available, this option defaults to `%f:%l:%c:%m`.
```
justinmk, line 3372: single-quotes automatically link options, so bars are redundant.
```suggestion
        If ripgrep ('grepprg') is available, this option defaults to "%f:%l:%c:%m".
```

LunarLambda: done

runtime/lua/nvim/health.lua
LunarLambda: It's already being printed? it prints ripgrep v1.14.0 (/usr/bin/rg) on my system.

or are you saying the value of `grepprg` should use the exepath too?

clason: That's fine, then! (Sorry, should have tested before commenting.)

clason:
Needs a `news.txt` (and probably also `vim-diff.txt`) entry.

LunarLambda:
> Needs a `news.txt` (and probably also `vim-diff.txt`) entry.

For vim-diff, which section? `nvim-defaults`? Is a single line calling out `grepprg` sufficient?

For news.txt, is this `news-breaking` or `news-changed`? Is there a specific ordering to those sections or should it simply go at the end?

clason:
> For vim-diff, which section? `nvim-defaults`? Is a single line calling out `grepprg` sufficient?

I'd say `nvim-defaults`

> For news.txt, is this `news-breaking` or `news-changed` (or `news-features`, even) ? Is there a specific ordering to those sections or should it simply go at the end?

I'd say to the "Defaults:" list under `news-changed` (at the end).

LunarLambda:
Added. For the sake of brevity I did not include the full default value, since both places link to the option where it's documented.

runtime/doc/news.txt
```diff
@@ -377,6 +377,7 @@ The following changes to existing APIs or features add new behavior.
   • Enabled treesitter highlighting for treesitter query files.
   • Enabled treesitter highlighting for help files.
   • Enabled treesitter highlighting for Lua files.
+  • 'grepprg' defaults to using ripgrep if available.
```
clason, line 380: Nit: in _this_ case I would move this line to after the `'shortmess'` entry (to keep treesitter stuff at the end).

clason (approved):

runtime/doc/news.txt
LunarLambda: done

justinmk (approved):

LunarLambda:
Happy to have landed this! Thank you for the feedback and advice.

dundargoc:
Thanks for this! The PR description and the context is useful, be sure to add that to the commit message as well next time :)

runtime/lua/vim/_defaults.lua
```diff
@@ -486,3 +486,11 @@ if tty then
     end
   end
 end
+
+--- Default 'grepprg' to ripgrep if available.
+if vim.fn.executable('rg') == 1 then
+  -- Match :grep default, otherwise rg searches cwd by default
+  -- Use -uuu to make ripgrep not do its default filtering
+  vim.o.grepprg = 'rg --vimgrep -uuu $* ' .. (vim.fn.has('unix') == 1 and '/dev/null' or 'nul')
```
gpanders, line 494: Apologies for the late review, but what is the rationale for `-uuu` rather than `-uu`? Is this strictly for compatibility with `grep` (which also searches binary files by default)?

Omitting binary file matches strikes me as a better default for a text editor. If the concern is consistent behavior for both `rg` and `grep`, I'd suggest adding `-I` to the default `grepprg` to ignore binary file matches in all cases.

LunarLambda: yes rg -uuu matches grep -r. I do agree. The documentation at least points out that the user may wish to adjust the flags used

gpanders: The use of `/dev/null` here is also unnecessary AFAIK. It's used for grep because some versions of grep (e.g. BSD grep) will try to read from stdin if no file argument is given, so a plain `grep foo` will block indefinitely and the `/dev/null` argument prevents that.

But ripgrep does not behave that way, and adding `/dev/null` here means you cannot use `:grep foo` to search for foo recursively (since that expands to `rg -uuu foo /dev/null`, which of course returns nothing).

LunarLambda: ripgrep searches the current directory by default. given it's also recursive by default, the /dev/null placeholder is to be both consistent with regular grep, and to prevent an accidental grep without files from potentially spinning up a lengthy process.

LunarLambda:
Given the feedback by @gpanders, is it worthwhile to revisit this before 0.10? Either when https://github.com/neovim/neovim/issues/23235 is closer to completion, or I could do it this weekend. The change is still very young, so I don't think many people have seen or given feedback on it yet, but I do think the suggestions are worth considering.

clason:
Sure, can still be changed (but also in a 0.10.x, so not a blocker either way).

akshayshedbalkar:
Hello @clason, @LunarLambda @justinmk,

thank you for your work on a great tool such as neovim. I appreciate the effort you guys do, I do not want to come across as thankless. I wanted to give feedback on this particular idea, and how it affected a simple workflow of me ( a normal user of vim and neovim). 

In my init.vim, I have some search and refactor expressions like so:

`nnoremap <Leader>s :grep -r --exclude-dir={cmake-build-debug,build,config,.git,tasking_build,delivery_build,tools,doc,cmocka,test,scripts,.vscode} --exclude={tags,*.swp,*.sqlite,*.obj,*.a,*.html,*.exe,*.rdump} <cword> .<CR>`

`command -nargs=1 Search :grep -r --exclude-dir={cmake-build-debug,build,config,.git,tasking_build,delivery_build,tools,doc,cmocka,test,scripts,.vscode} --exclude={tags,*.swp,*.sqlite,*.obj,*.a,*.html,*.exe,*.rdump} <args> .`

`command -nargs=+ Refactor cfdo :%s/<args>/g|update`

Yes I also use LSP, but these expressions are helpful for those times that LSP doesn't work. You might guess the problem: they use grep. These lines have been in my vimrc and later in init.vim for years and suddenly they stopped working in neovim. I had ripgrep installed on my system and so neovim automatically chose that and then couldn't understand the grep syntax anymore. 

I get that ripgrep is the new fancy tool and lots of people want to use it. I don't expect this change to be reverted or any action at all. I just wanted to ask you, if in the future it would be possible to think about such breaking changes more carefully :(

clason:
How should we know it is breaking something if nobody tells us? Taken seriously, that would mean never changing anything (which is what Vim does, so there's already a tool for people for whom that is important).

I'm sure someone can help you translate these commands to something compatible with ripgrep, if you ask nicely.

gpanders:
>I'm sure someone can help you translate these commands to something compatible with ripgrep

Or, if your commands are assuming that `'grepprg'` is set to `grep` then it is probably best to explicitly set that value anyway:

```vim
set grepprg=grep\ -HIn\ $*\ /dev/null
```

akshayshedbalkar:
Thank you for your responses. This is my first time ever interacting with developers of a large tool, I really thought my comment was respectful. If it wasn't, I am sorry. 

I did figure out a solution for me, it was: 

` vim.o.grepprg = 'grep'`

I was just excited about interacting with you.

gpanders:
> I did figure out a solution for me, it was:
> 
> ` lua <<EOF vim.o.grepprg = 'grep' EOF`


@akshayshedbalkar Presumably this is in a Vimscript (init.vim) file. You do not need to use Lua to set the `'grepprg'` option. See my [comment above](https://github.com/neovim/neovim/pull/28324#issuecomment-2256067266).

clason:
@akshayshedbalkar It was respectfully formulated, but the underlying request wasn't really. No harm done, but I like to be clear about these things (also for other people reading this now or later). As @gpanders wrote, a better approach would have been (and will be, for future interactions) to 
1. phrase this as a question on how best to adjust to the change and
2. do this in a new Discussion or a comment in Neovim Chat.

PR #28327, opened 2024-04-14
refactor: fix clang NonNullParamChecker warnings

zeertzjq:


src/nvim/regexp.c
```diff
@@ -5320,7 +5320,7 @@ static regprog_T *bt_regcomp(uint8_t *expr, int re_flags)
   }
   // Remember whether this pattern has any \z specials in it.
   r->reghasz = (uint8_t)re_has_z;
-  scan = r->program + 1;        // First BRANCH.
+  scan = &r->program[1];  // First BRANCH.
```
zeertzjq, line 5323: `&r->program[1]` are `r->program + 1` generate exactly the same assembly code. Not sure why clang analyzer thinks `r->program + 1` can be NULL.

justinmk (approved):

PR #28330, opened 2024-04-14
fix(inspect): use quoting style from stylua.toml

clason:
Problem: `vim.inspect` uses double quotes by default, which means that
copying and pasting the output into our sources will fail CI.

Solution: Flip quotes.

Note: The output still is not guaranteed to pass linting (trailing
commas, line lengths), but this reduces the number of changes required.

### DIFF
--- a/runtime/lua/vim/_meta/options.lua
+++ b/runtime/lua/vim/_meta/options.lua
@@ -53,7 +53,7 @@ vim.go.ari = vim.go.allowrevins
 --- (https://www.unicode.org/reports/tr11).
 ---
 --- @type string
-vim.o.ambiwidth = "single"
+vim.o.ambiwidth = 'single'
 vim.o.ambw = vim.o.ambiwidth
 vim.go.ambiwidth = vim.o.ambiwidth
 vim.go.ambw = vim.go.ambiwidth
@@ -209,7 +209,7 @@ vim.go.awa = vim.go.autowriteall
 --- the setting of the 'background' option.
 ---
 --- @type string
-vim.o.background = "dark"
+vim.o.background = 'dark'
 vim.o.bg = vim.o.background
 vim.go.background = vim.o.background
 vim.go.bg = vim.go.background
@@ -229,7 +229,7 @@ vim.go.bg = vim.go.background
 --- the ways mentioned for the items above are possible.
 ---
 --- @type string
-vim.o.backspace = "indent,eol,start"
+vim.o.backspace = 'indent,eol,start'
 vim.o.bs = vim.o.backspace
 vim.go.backspace = vim.o.backspace
 vim.go.bs = vim.go.backspace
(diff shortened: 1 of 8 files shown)


runtime/lua/vim/_meta/options.lua
```diff
@@ -1926,7 +1926,7 @@ vim.go.ef = vim.go.errorfile
...
+vim.o.errorformat = '%*[^"]"%f"%*\\D%l: %m,"%f"%*\\D%l: %m,%-Gg%\\?make[%*\\d]: *** [%f:%l:%m,%-Gg%\\?make: *** [%f:%l:%m,%-G%f:%l: (Each undeclared identifier is reported only once,%-G%f:%l: for each function it appears in.),%-GIn file included from %f:%l:%c:,%-GIn file included from %f:%l:%c\\,,%-GIn file included from %f:%l:%c,%-GIn file included from %f:%l,%-G%*[ ]from %f:%l:%c,%-G%*[ ]from %f:%l:,%-G%*[ ]from %f:%l\\,,%-G%*[ ]from %f:%l,%f:%l:%c:%m,%f(%l):%m,%f:%l:%m,"%f"\\, line %l%*\\D%c%*[^ ] %m,%D%*\\a[%*\\d]: Entering directory %*[`\']%f\',%X%*\\a[%*\\d]: Leaving directory %*[`\']%f\',%D%*\\a: Entering directory %*[`\']%f\',%X%*\\a: Leaving directory %*[`\']%f\',%DMaking %*\\a in %f,%f|%l| %m'
```
clason, line 1929: 👀 

(Note that `stylua` wants to format this differently due to the `AutoPrefer` option, which `inspect.lua` is not smart enough for in this case.)

clason:
Not sure if the meta and test churn is worth it?

justinmk (approved):
should `runtime/lua/vim/_meta/options.lua` be removed from an ignore list somewhere?

clason:
No, since the `vim.inspect` output still doesn't pass lualint.

clason:
Don't know why the `ui_spec` is failing (and what it expects).

zeertzjq:
Why should the behavior of an stdlib function depend on Nvim source code conventions?

test/functional/lua/ui_spec.lua
```diff
@@ -142,7 +142,7 @@ describe('vim.ui', function()
       end
       if not is_os('bsd') then
         matches(
-          'vim.ui.open: command failed %(%d%): { "[^"]+", .*"non%-existent%-file" }',
+          "vim.ui.open: command failed %(%d%): { '[^']+', 'non%-existent%-file' }",
```
zeertzjq, line 145:
```suggestion
          "vim.ui.open: command failed %(%d%): { '[^']+', .*'non%-existent%-file' }",
```

justinmk:
> Why should the behavior of an stdlib function depend on Nvim source code conventions?

We have to choose something, so why not align with the conventions we intentionally selected?

- Eliminates some contributor confusion and toil. 
- We chose the conventions because we think they are good. So it makes sense for our stdlib to produce things that are also good.

PR #6380, opened 2017-03-28
[RDY] hl-Syntax should override cursorline and cursorcolumn

zhou13:
This patch makes syntax highlight correctly overrides 'cursorline' and 'cursorcolumn'

Before this patch (default json syntax highlights every comment as error):

2017-03-27_18 53 53_826x446
2017-03-27_18 53 46_826x446

After this patch:
2017-03-27_18 47 36_826x446
2017-03-27_18 47 35_826x446

While `nvim -d` or "folding" is not affected (This may be good but I don't know exactly why):
2017-03-27_18 55 55_826x446

PR #28345, opened 2024-04-15
fix(highlight): combine line attributes before column attributes

### LINKED ISSUE #28118: Colorcolumn is not displayed above modified background
### Problem

When running neovim in diff mode, the color column is not displayed above the background indicating changes:
with termguicolors:
image
with the `vim` color scheme:
image

(the default color scheme without termguicolors works fine because it inverts 

### Steps to reproduce

`nvim --clean -d foo bar`
Then run
`:set textwidth=20 | set colorcolumn=+1 | set termguicolors`
or
`set textwidth=20 | set colorcolumn=+1 | colorscheme vim`

### Expected behavior

vimdiff:
image

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2754+g12240600f

no, VIM - Vi IMproved 9.0  Included patches: 1-841

### Operating system/version

Ubuntu 20.04

### Terminal name/version

st with tmux. Same behavour in xterm

screen-256color

### Installation

built from repo

jgouly:
Fixes: #28118

TODO: add test
 ```
    nvim --clean -u NONE -c 'color vim' -c 'norm Ia' -c 'norm ob'  -c 'vnew|norm Ia' -c 'norm oc' -c 'windo diffthis|set cc=8'
```
```
    nvim --clean -u NONE -c 'color vim' -c 'hi ColorColumn guibg=blue' -c 'norm Iab' -c 'norm obc'  -c 'vnew|norm Iab' -c 'norm ocd' -c 'windo diffthis|set cc=2'
```

I don't know what other things this could affect, so opening it now before writing a test, in case it's the wrong thing to do.

### DIFF
--- a/src/nvim/drawline.c
+++ b/src/nvim/drawline.c
@@ -2606,6 +2606,8 @@ int win_line(win_T *wp, linenr_T lnum, int startrow, int endrow, int col_rows, s
 
           int col_attr = base_attr;
 
+          col_attr = hl_combine_attr(col_attr, wlv.line_attr);
+
           if (wp->w_p_cuc && vcol_hlc(wlv) == wp->w_virtcol
               && lnum != wp->w_cursor.lnum) {
             col_attr = hl_combine_attr(col_attr, cuc_attr);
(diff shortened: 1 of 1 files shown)


zeertzjq:
This is makes the priority order at end-of-line inconsistent with buffer text. For buffer text, in Nvim it's `line_attr > char_attr > col_attr > line_attr_lowprio`, while in Vim it's `col_attr > line_attr > char_attr`.

jgouly:
Is there a reason nvim has the order changed?

zeertzjq:
The earliest one I can find is #6380

sergei-grechanik:
Any updates on the issue? I think there were some discussions about introducing priorities into highlight groups. It may be a better approach than changing the internal attribute combining order, since some highlight groups should be displayed above the column while others should be displayed below. Are there any plans to implement this?

justinmk:
https://github.com/neovim/neovim/pull/6380 has test coverage so if this doesn't regress it then I guess no problem?

zeertzjq:
The problem is inconsistency, not regression.

PR #28355, opened 2024-04-15
fix(msgpack): retain grid line event memory

### LINKED ISSUE #27859: Segfault (msgpack)
# Problems

1) exit with status 1 (logs: `chan_close_with_error:560: RPC: (null)`)
2) Crash with status 130:

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x00000003209cd220
Exception Codes:       0x0000000000000001, 0x00000003209cd220

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [11660]

VM Region Info: 0x3209cd220 is not in any region.  Bytes after previous region: 2385498657  Bytes before following region: 54213684704
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      unused __TEXT               2926cc000-2926d0000    [   16K] r--/r-- SM=COW  unused  unknown system shared lib __TEXT
--->  GAP OF 0xd2d930000 BYTES
      commpage (reserved)         fc0000000-1000000000   [  1.0G] ---/--- SM=NUL  reserved VM address space (unallocated)

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   nvim                          	       0x100d60ab4 tui_raw_line + 80
1   nvim                          	       0x100bf62a8 parse_msgpack + 348
2   nvim                          	       0x100bf3ff0 receive_msgpack + 304
3   nvim                          	       0x100af0dc8 read_cb + 404
4   libuv.1.dylib                 	       0x1014e35b4 uv__stream_io + 1004
5   libuv.1.dylib                 	       0x1014ea164 uv__io_poll + 1408
6   libuv.1.dylib                 	       0x1014da93c uv_run + 272
7   nvim                          	       0x100d689bc ui_client_run + 172
8   nvim                          	       0x100b963cc main + 17512
9   dyld                          	       0x198bc60e0 start + 2360
```

### Steps to reproduce

Scroll any file

### Expected behavior

No crash/errors

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2582+g2a8cef6bd-Homebrew
NVIM v0.10.0-dev-2593+ga6b6d036b-Homebrew
NVIM v0.10.0-dev-2536+g55c9e2c96 <-- broken commit

no

### Operating system/version

macOS 14

### Terminal name/version

Alacritty

xterm-256color

### Installation

brew

theofabilous:
Fix #27859. Still a draft because there are other solutions that might be worthwhile, and I haven't written any tests.

See my comments on that issue (https://github.com/neovim/neovim/issues/27859#issuecomment-2057556931 and https://github.com/neovim/neovim/issues/27859#issuecomment-2057916584) for details on the specific situation that causes the crash. But TLDR:
- there's a hand-rolled msgpack decoder specific to redraw events, implemented as a finite state machine with states 0-16
- the state machine can be interrupted/resumed
- memory for `GridLineEvent`s is allocated in state 11 only
- states 14-16 use the `GridLineEvent` assuming its memory is allocated
- if the state machine completes a cycle (reaches 16) and then executes the redraw event, the grid line event memory is freed ([`channel.c`:288-298](https://github.com/neovim/neovim/blob/5cfdaaaeac0f53a621696d8eb6b5a3ba90438c98/src/nvim/msgpack_rpc/channel.c#L288-L298)):
```c
static void parse_msgpack(Channel *channel) {
  Unpacker *p = channel->rpc.unpacker;
  while (unpacker_advance(p)) {
    if (p->type == kMessageTypeRedrawEvent) { /* ... */
      if (p->grid_line_event) {
        ui_client_event_raw_line(p->grid_line_event);
      } /* ... */
    }
    arena_mem_free(arena_finish(&p->arena)); // <-- freed here
  } /* ... */
}
```
- but if there are any remaining calls (`p->ncalls`) that can be read from the RPC stream, the FSM resumes *directly* at state 14 ([`unpacker.c`:320:326](https://github.com/neovim/neovim/blob/5cfdaaaeac0f53a621696d8eb6b5a3ba90438c98/src/nvim/msgpack_rpc/unpacker.c#L320-L326), [`unpacker.c`:359:368](https://github.com/neovim/neovim/blob/5cfdaaaeac0f53a621696d8eb6b5a3ba90438c98/src/nvim/msgpack_rpc/unpacker.c#L359-L368)):
```c
bool unpacker_advance(Unpacker *p) { /* ... */
  if (!unpacker_parse_redraw(p)) {
    return false;
  } /* ... */
  switch (p->state) { /* ... */
  case 16:
    p->ncalls--;
    if (p->ncalls > 0) {
      p->state = (p->state == 16) ? 14 : 12;
    } /* ... else ... */
    return true;
    /* ... */
  }
}
```
- ...bypassing the reallocation of the event and writing to released memory on the next call to `unpacker_parser_redraw` ([`unpacker.c`:428:437](https://github.com/neovim/neovim/blob/5cfdaaaeac0f53a621696d8eb6b5a3ba90438c98/src/nvim/msgpack_rpc/unpacker.c#L428-L437)):
```c
bool unpacker_parse_redraw(Unpacker *p) { /* ... */
  switch (p->state) { /* ... */
      case 11: // allocation happens here
      // we jump here directly, skipping allocation in state 11
      case 14: // write to p->grid_line_event
      /* ... */
    }
}
```

The bug is sneaky/tricky because it's only *apparent* in very specific cases, i.e. when the pending channel `rstream` is very large (> `8192` bytes), can't be read in a single step, and another event in between the pending redraw is sent and overwrites the `GridLineEvent` memory before the remaining redraw event data is read.

~~**Current approach**: just don't free the memory when returning successfully from `unpacker_advance` midway through a redraw event when the next FSM execution would expect the grid line event memory to be valid.~~

~~**Other approaches**~~ **Current approach**: Don't *ever* allocate/free the grid line event, i.e. store it as a value and not a pointer? It's not that big of a struct..

### DIFF
--- a/src/nvim/msgpack_rpc/channel.c
+++ b/src/nvim/msgpack_rpc/channel.c
@@ -289,8 +289,9 @@ static void parse_msgpack(Channel *channel)
     if (p->type == kMessageTypeRedrawEvent) {
       // When exiting, ui_client_stop() has already been called, so don't handle UI events.
       if (ui_client_channel_id && !exiting) {
-        if (p->grid_line_event) {
-          ui_client_event_raw_line(p->grid_line_event);
+        if (p->has_grid_line_event) {
+          ui_client_event_raw_line(&p->grid_line_event);
+          p->has_grid_line_event = false;
         } else if (p->ui_handler.fn != NULL && p->result.type == kObjectTypeArray) {
           p->ui_handler.fn(p->result.data.array);
         }
--- a/src/nvim/msgpack_rpc/unpacker.c
+++ b/src/nvim/msgpack_rpc/unpacker.c
@@ -187,6 +187,8 @@ void unpacker_init(Unpacker *p)
   p->unpack_error = ERROR_INIT;
 
   p->arena = (Arena)ARENA_EMPTY;
+
+  p->has_grid_line_event = false;
 }
 
 void unpacker_teardown(Unpacker *p)
@@ -303,6 +305,7 @@ error:
 bool unpacker_advance(Unpacker *p)
 {
   assert(p->state >= 0);
+  p->has_grid_line_event = false;
   if (p->state == 0) {
     if (!unpacker_parse_header(p)) {
       return false;
@@ -323,6 +326,7 @@ bool unpacker_advance(Unpacker *p)
 
     if (p->state == 16) {
       // grid_line event already unpacked
+      p->has_grid_line_event = true;
       goto done;
     } else {
       assert(p->state == 12);
@@ -378,7 +382,7 @@ bool unpacker_parse_redraw(Unpacker *p)
 
   const char *data = p->read_ptr;
   size_t size = p->read_size;
-  GridLineEvent *g = p->grid_line_event;
+  GridLineEvent *g = &p->grid_line_event;
 
 #define NEXT_TYPE(tok, typ) \
   result = mpack_rtoken(&data, &size, &tok); \
@@ -420,16 +424,10 @@ bool unpacker_parse_redraw(Unpacker *p)
     p->read_size = size;
     if (p->ui_handler.fn != ui_client_event_grid_line) {
       p->state = 12;
-      if (p->grid_line_event) {
-        arena_mem_free(arena_finish(&p->arena));
-        p->grid_line_event = NULL;
-      }
       return true;
     } else {
       p->state = 14;
       p->arena = (Arena)ARENA_EMPTY;
-      p->grid_line_event = arena_alloc(&p->arena, sizeof *p->grid_line_event, true);
-      g = p->grid_line_event;
     }
     FALLTHROUGH;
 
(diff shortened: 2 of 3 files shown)


bfredl:
Thanks for investigating.

> Making the grid_line_event field in Unpacker a value instead of a pointer (thus never allocating/freeing it, and never needing to do so) seems to fix the bug

I think this would be the proper approach. Need to use a `bool` or something instead for keeping track of there is a grid_line_event ready to be consumed.

bfredl:
looks alright to me. something missing to make it ready?

theofabilous:
It's good to go! Marking as ready for review

PR #28366, opened 2024-04-16
fix(spell): Moved Python @spell to (string_content)

### LINKED ISSUE #28365: Python Tree-sitter @spell highlights string prefixes
### Problem

image

Python strings can have prefixes such as r, f, u, etc. These shouldn't be spell checked but they are and are recognized as misspellings (see screenshot).

### Steps to reproduce

Load Neovim with Python tree-sitter parser enabled. Then `:set spell`

### Expected behavior

The "r" in the screenshot is not highlighted as a misspelled word.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2791+g99b3a068d

not sure

### Operating system/version

Ubuntu 20.02

### Terminal name/version

Wezterm

xterm-256color

### Installation

compiled from source

ColinKennedy:
Closes: https://github.com/neovim/neovim/issues/28365

## Before
image

## After
image

For those that use nvim-treesitter, their parser has the same problem (which I can also make a PR for).

### DIFF
--- a/runtime/queries/python/highlights.scm
+++ b/runtime/queries/python/highlights.scm
@@ -204,19 +204,19 @@
   (comment)*
   .
   (expression_statement
-    (string) @string.documentation @spell))
+    (string (string_content) @spell) @string.documentation))
 
 (class_definition
   body: (block
     .
     (expression_statement
-      (string) @string.documentation @spell)))
+      (string (string_content) @spell) @string.documentation)))
 
 (function_definition
   body: (block
     .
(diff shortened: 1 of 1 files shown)


justinmk:
Based on https://github.com/neovim/neovim/issues/28365#issuecomment-2058328962 it sounds like this can be closed?

clason:
Yes, superseded by https://github.com/nvim-treesitter/nvim-treesitter/pull/6449

PR #28369, opened 2024-04-16
fix(api): don't create or free luaref in arena object

### LINKED ISSUE #28353: Loss of Arbitrary Functions in Autocommands After Upgrading to Neovim nightly (v0.10.0-dev-2830)
### Problem

Updating Neovim from v0.9.5 to v0.10.0-dev-2830+g541c2d381 appears to disrupt the passing of arbitrary functions through autocommands. They become lost after the initial handler.

### Steps to reproduce

```lua
local fun = function()
  return 42
end

local group = vim.api.nvim_create_augroup("MyTest", { clear = false })

vim.api.nvim_create_autocmd("User", {
  group = group,
  callback = function(payload)
    print(vim.inspect(payload.data))
  end,
  pattern = "MyEvent",
})
vim.api.nvim_create_autocmd("User", {
  group = group,
  callback = function(payload)
    print(vim.inspect(payload.data))
  end,
  pattern = "MyEvent",
})

vim.api.nvim_exec_autocmds("User", {
  group = group,
  pattern = "MyEvent",
  data = { x = 1, fun = fun },
})
```

### Expected behavior

I expect the functions to be kept.

Expected output:

```
{
  fun = <function 1>,
  x = 1
}
{
  fun = <function 1>,
  x = 1
}
```

But it is:

```
{
  fun = <function 1>,
  x = 1
}
{
  x = 1
}
```

### Neovim version (nvim -v)

v0.10.0-dev-2830+g541c2d381

I don't know

### Operating system/version

Manjaro Linux 23.1.3

### Terminal name/version

alacritty 0.13.1

xterm-256color

### Installation

aur

zeertzjq:
Fix #28353

### DIFF
--- a/src/nvim/lua/converter.c
+++ b/src/nvim/lua/converter.c
@@ -766,8 +766,6 @@ void nlua_push_Object(lua_State *lstate, Object *obj, bool special)
     break;
   case kObjectTypeLuaRef: {
     nlua_pushref(lstate, obj->data.luaref);
-    api_free_luaref(obj->data.luaref);
-    obj->data.luaref = LUA_NOREF;
     break;
   }
 #define ADD_TYPE(type, data_key) \
--- a/src/nvim/api/autocmd.c
+++ b/src/nvim/api/autocmd.c
@@ -292,7 +292,7 @@ Array nvim_get_autocmds(Dict(get_autocmds) *opts, Arena *arena, Error *err)
         switch (cb->type) {
         case kCallbackLua:
           if (nlua_ref_is_function(cb->data.luaref)) {
-            PUT_C(autocmd_info, "callback", LUAREF_OBJ(api_new_luaref(cb->data.luaref)));
+            PUT_C(autocmd_info, "callback", LUAREF_OBJ(cb->data.luaref));
           }
           break;
         case kCallbackFuncref:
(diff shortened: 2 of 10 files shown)


src/nvim/lua/converter.c
```diff
@@ -766,8 +766,6 @@ void nlua_push_Object(lua_State *lstate, Object *obj, bool special)
     break;
   case kObjectTypeLuaRef: {
     nlua_pushref(lstate, obj->data.luaref);
-    api_free_luaref(obj->data.luaref);
-    obj->data.luaref = LUA_NOREF;
```
bfredl, line 770: hmm, this would cause leaks in cases a arena allocated value is returned with included luarefs which are unique. 

But perhaps autocmd was the only case where we did this? I thought it was more but I could remember wrong ( speaking about real useful functions, not the `nvim__id` clownshow).

zeertzjq: If I understand the meaning of "unique luarefs" correctly, after this PR an arena allocated value shouldn't include unique luarefs, because `api_nvim_luaref()` is no longer used in an arena-allocated value.

bfredl: Ye I understand the intention. As the tests don't catch any new leaks, it is probably correct.

PR #28370, opened 2024-04-16
fix: call do_autochdir inside ex_win_close

### LINKED ISSUE #15280: nvim_open_win with non-current buffer leads to wrong working directory with 'autochdir'
### Neovim Version

NVIM v0.5.0-dev+1115-gc1fbc2ddf

### Operating system/version:

WSL Ubuntu

### Terminal name/version:

mintty 3.4.5

### TERM environment variable

xterm-256color

### Installation

Built from source

### Steps to reproduce

vimrc:
```vim
set autochdir

autocmd CmdlineLeave * call Leave()

let s:buf = nvim_create_buf(0, 1)
call nvim_buf_set_name(s:buf, 'Foo')

function! Leave()
  let l:win = nvim_open_win(s:buf, 0, {
        \ 'relative': 'editor',
        \ 'height': 1,
        \ 'width': 1,
        \ 'row': 0,
        \ 'col': 0,
        \ })
  call nvim_win_close(l:win, 1)
endfunction
```

1. Open nvim: `nvim -u vimrc`.
2. Edit a file in a different directory: `:e ~/foo/bar.txt`
3. Check pwd: `pwd`

### Vim

No, Vim doesn't have floating windows

### Expected behavior

`pwd` should show `~/foo`.

### Actual behavior

`pwd` shows `~`.

glepnir:
Problem: after nvim_win_close called in event callback on cmdline the autodir not clear

Solution: call do_autochdir inside ex_win_close

Fix #15280

### DIFF
--- a/src/nvim/ex_docmd.c
+++ b/src/nvim/ex_docmd.c
@@ -4872,6 +4872,8 @@ void ex_win_close(int forceit, win_T *win, tabpage_T *tp)
   } else {
     win_close_othertab(win, !need_hide && !buf_hide(buf), tp);
   }
+
+  do_autochdir();
 }
 
 /// ":tabclose": close current tab page, unless it is the last one.
--- a/test/functional/api/window_spec.lua
+++ b/test/functional/api/window_spec.lua
@@ -680,6 +680,31 @@ describe('API/win', function()
       eq(prevwin, api.nvim_tabpage_get_win(tab))
       assert_alive()
     end)
+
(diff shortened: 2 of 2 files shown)


zeertzjq:
I don't think this makes sense. Closing non-current window shouldn't trigger directory change by 'autochdir', while closing current window already does that.

glepnir:
emm i do a new change ..

PR #28371, opened 2024-04-16
fix(float): reset pwd after win_set_buf in nvim_open_win

### LINKED ISSUE #15280: nvim_open_win with non-current buffer leads to wrong working directory with 'autochdir'

glepnir:
Problem:  win_set_buf will make buffer as current and do_autochdir when p_acd is true
Solution:  temporarily disable 'autochdir' when using win_set_buf()

Fix #15280

### DIFF
--- a/test/functional/api/window_spec.lua
+++ b/test/functional/api/window_spec.lua
@@ -1748,6 +1748,31 @@ describe('API/win', function()
         pcall_err(api.nvim_open_win, 0, true, { split = 'below', win = 0 })
       )
     end)
+
+    it('do not change dir when enter is false', function()
+      local expected = fn.getcwd() .. '/foo'
+      t.mkdir('foo')
+      exec_lua [[
+        vim.opt.autochdir = true
+        local buf = vim.api.nvim_create_buf(false, true)
+        vim.api.nvim_buf_set_name(buf, 'Foo')
+        vim.api.nvim_create_autocmd('CmdlineEnter', {
+          callback = function()
+            local winid = vim.api.nvim_open_win(buf, false, {
+              relative = 'editor',
+              height = 1,
+              width = 1,
+              row = 1,
+              col = 1,
+            })
+            vim.api.nvim_win_close(winid, true)
+          end,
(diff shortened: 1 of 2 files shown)


zeertzjq:
Another approach is to temporarily disable 'autochdir' when using `win_set_buf()` on non-current window. A similar one can be see in `aucmd_prepbuf()`.

glepnir:
oh thanks do change now.

test/functional/api/window_spec.lua
```diff
@@ -551,6 +551,33 @@ describe('API/win', function()
     end)
   end)
 
+  describe('reset dir', function()
+    it('reset dir in nvim_open_win', function()
```
justinmk, line 555: should this test live in the nvim_open_win describe() block? : https://github.com/neovim/neovim/blob/57adf8c6e01d9395eb52fe03571c535571efdc4b/test/functional/api/window_spec.lua#L1164

PR #28375, opened 2024-04-16
ci: cache build directory

dundargoc:
Testing around, unsure if this is worth it.


dundargoc:
This does not seem to work. Even when the cache key hits the entire project is rebuilt from scratch. This looks like it might be inherent to how cmake works with modification times and looks hard to circumvent.

PR #28376, opened 2024-04-16
fix(api): ignore 'autochdir' when renaming other buf

zeertzjq:
Problem:  Renaming non-current buffer changes working directory when
          'autochdir' is set.
Solution: Temporarily disable 'autochdir'.  Add more tests for the
          win_set_buf change.

### DIFF
--- a/src/nvim/api/buffer.c
+++ b/src/nvim/api/buffer.c
@@ -39,6 +39,7 @@
 #include "nvim/memory_defs.h"
 #include "nvim/move.h"
 #include "nvim/ops.h"
+#include "nvim/option_vars.h"
 #include "nvim/pos_defs.h"
 #include "nvim/state_defs.h"
 #include "nvim/types_defs.h"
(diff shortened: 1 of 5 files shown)


src/nvim/api/buffer.c
```diff
@@ -984,12 +985,23 @@ void nvim_buf_set_name(Buffer buffer, String name, Error *err)
...
+  const bool is_curbuf = buf == curbuf;
+  const int save_acd = p_acd;
+  if (!is_curbuf) {
+    // Temporarily disable 'autochdir' when setting file name for another buffer.
+    p_acd = false;
+  }
+
   // Using aucmd_*: autocommands will be executed by rename_buffer
   aco_save_T aco;
   aucmd_prepbuf(&aco, buf);
```
justinmk, line 997: should the `p_acd` save/restore be done by `aucmd_prepbuf` ?

zeertzjq: I think that'll break other APIs.

PR #28377, opened 2024-04-16
fix: vim.validate() order is not deterministic

justinmk:
# Problem:
The order of the validation performed by vim.validate() is unpredictable.
- harder to write reliable tests.
- confusing UX because validation result might return different errors randomly.

# Solution:
Iterate the input using `vim.spairs()`.

# Future:
Ideally, the caller could provide an "ordered dict".

### DIFF
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -578,7 +578,7 @@ end
 ---@return fun(table: table<K, V>, index?: K):K, V # |for-in| iterator over sorted keys and their values
 ---@return T
 function vim.spairs(t)
-  vim.validate({ t = { t, 't' } })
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/shared.lua
```diff
@@ -578,7 +578,7 @@ end
 ---@return fun(table: table<K, V>, index?: K):K, V # |for-in| iterator over sorted keys and their values
 ---@return T
 function vim.spairs(t)
-  vim.validate({ t = { t, 't' } })
+  assert(type(t) == 'table', ('expected table, got %s'):format(type(t)))
```
justinmk, line 581: Avoid infinite loop

PR #28379, opened 2024-04-16
fix(healthcheck): expand vimrc variable

sarmong:
`MYVIMRC` can have `~` or env variables in there.

In my specific case I am setting `VIMINIT` to move `vimrc` out of the home directory. However, because this variable will affect neovim, I have to set it for neovim as well.

```bash
export VIMINIT='let $MYVIMRC = !has("nvim") ? "$XDG_CONFIG_HOME/vim/vimrc" : exists("$NVIM_APPNAME") ? "$XDG_CONFIG_HOME/$NVIM_APPNAME/init.lua" : "$XDG_CONFIG_HOME/nvim/init.lua" | so $MYVIMRC'

```

### DIFF
--- a/runtime/lua/nvim/health.lua
+++ b/runtime/lua/nvim/health.lua
@@ -50,7 +50,7 @@ local function check_config()
 
   local init_lua = vim.fn.stdpath('config') .. '/init.lua'
   local init_vim = vim.fn.stdpath('config') .. '/init.vim'
-  local vimrc = vim.env.MYVIMRC or init_lua
+  local vimrc = vim.fn.expand(vim.env.MYVIMRC or init_lua)
 
   if vim.fn.filereadable(vimrc) == 0 and vim.fn.filereadable(init_vim) == 0 then
     ok = false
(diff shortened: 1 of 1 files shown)


justinmk (approved):

runtime/lua/nvim/health.lua
```diff
@@ -50,7 +50,7 @@ local function check_config()
 
   local init_lua = vim.fn.stdpath('config') .. '/init.lua'
   local init_vim = vim.fn.stdpath('config') .. '/init.vim'
-  local vimrc = vim.env.MYVIMRC or init_lua
+  local vimrc = vim.fn.expand(vim.env.MYVIMRC or init_lua)
```
zeertzjq, line 53: `init_lua` should not be expanded

sarmong: I did it, because `expand` on a non-existing variable returns `v:null` and so `or` isn't triggered.

Is having pseudo-ternary like this is fine?

```suggestion
  local vimrc = vim.env.MYVIMRC and vim.fn.expand(vim.env.MYVIMRC) or init_lua
```

zeertzjq: Yes

PR #28387, opened 2024-04-17
fix(float): tail truncate the long title

### LINKED ISSUE #23602: inconsist behavior: a over long float window title gets truncated from the right side
### Problem

hi folks, currently when the title of a float win is longer than the window width, it gets truncated from the right side, and this behavior suprises me, because from my past experience, &winbar and tabpage title in &tabline truncate the text from the left side.

my usecase is that: i have a file browser plugin that shows files under a directory in a float window, and i name the title of that window using the absolute path of that directory. that's said the right most part of the path is most important to me.

i surely can handle this by myself by cheking the length of title and window's width, but i think it's still worth to consider making nvim be consistent across its UI.

(i asked this in the matrix room weeks ago, and got answered from glepnir, but i failed to find an issue that tracks other thoughts on this, sorry if there is one already)

### Expected behavior

a over long float window title gets truncated from the left side

glepnir:
Problem: 
now we truncate the title from start,This is somewhat inconsistent with the way vim works. Usually we use < to truncate and keep more trailing characters.

Solution: 
keep more trailing characters. 

Fix #23602

### DIFF
--- a/src/nvim/grid.c
+++ b/src/nvim/grid.c
@@ -1085,8 +1085,14 @@ void grid_del_lines(ScreenGrid *grid, int row, int line_count, int end, int col,
 }
 
 static void grid_draw_bordertext(VirtText vt, int col, int winbl, const int *hl_attr,
-                                 BorderTextType bt)
+                                 BorderTextType bt, int over_flow)
 {
+  int default_attr = hl_attr[bt == kBorderTextTitle ? HLF_BTITLE : HLF_BFOOTER];
+  if (over_flow > 0) {
+    grid_line_puts(1, "<", -1,  hl_apply_winblend(winbl, default_attr));
+    col += 1;
+  }
+
   for (size_t i = 0; i < kv_size(vt);) {
     int attr = -1;
     char *text = next_virt_text_chunk(vt, &i, &attr);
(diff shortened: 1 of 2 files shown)


src/nvim/grid.c
```diff
@@ -1085,16 +1085,35 @@ void grid_del_lines(ScreenGrid *grid, int row, int line_count, int end, int col,
...
     int attr = -1;
     char *text = next_virt_text_chunk(vt, &i, &attr);
     if (text == NULL) {
       break;
     }
     if (attr == -1) {  // No highlight specified.
-      attr = hl_attr[bt == kBorderTextTitle ? HLF_BTITLE : HLF_BFOOTER];
+      attr = default_attr;
+    }
+    if (over_flow > 0) {
```
justinmk, line 1107: this code will be hard to understand in the future. now is the time to add a 1-line comment that says what it does. As always, a brief input/output example is 100x more helpful than a long comment:

    // Example: "..." => "..."

is there not a truncate function that can be used? `trunc_string` looks promising: https://github.com/neovim/neovim/blob/64cf63a88110459d70d140f6074fa014adaef331/src/nvim/message.c#L518

PR #28388, opened 2024-04-17
feat(lsp): add vim.lsp.buf.subtypes(), vim.lsp.buf.supertypes()

jiangyinzuo:
Implement https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareTypeHierarchy


jiangyinzuo:
image

runtime/doc/news.txt
```diff
@@ -209,6 +209,9 @@ The following new APIs and features were added.
     https://microsoft.github.io/language-server-protocol/specification/#textDocument_inlayHint
   • Implemented pull diagnostic textDocument/diagnostic: |vim.lsp.diagnostic.on_diagnostic()|
     https://microsoft.github.io/language-server-protocol/specification/#textDocument_diagnostic
+  • Implemented LSP type hierarchy: |vim.lsp.buf.supertypes()| and |vim.lsp.buf.subtypes()|
+  • |vim.lsp.status()| consumes the last progress messages as a string.
```
mfussenegger, line 213:
```suggestion
```

Looks like a copy error?

runtime/lua/vim/lsp/buf.lua
```diff
@@ -495,6 +495,69 @@ function M.outgoing_calls()
...
+  end
+  if #type_hierarchy_items == 1 then
+    return type_hierarchy_items[1]
+  end
+  local items = {}
+  for i, item in pairs(type_hierarchy_items) do
+    local entry = item.detail or item.name
+    table.insert(items, string.format('%d. %s', i, entry))
+  end
+  local choice = vim.fn.inputlist(items)
```
mfussenegger, line 511: This could use `vim.ui.select` instead. It's slightly easier to use and provides a more consistent interface for users, as the other lsp operations also use it.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -495,6 +495,69 @@ function M.outgoing_calls()
...
+    return
+  end
+  return choice
+end
+
+--- @param method string
+local function type_hierarchy(method)
+  local params = util.make_position_params()
+  --- @param result lsp.TypeHierarchyItem[]?
+  request(ms.textDocument_prepareTypeHierarchy, params, function(err, result, ctx)
```
mfussenegger, line 522: For proper multi-client handling this should use `buf_request_all`. The results from the servers can be merged/appended.

As it is right now, it could result in multiple prompts to select an item, and then the quickfix list would be populated once for each client/server.

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -565,6 +565,35 @@ M[ms.callHierarchy_incomingCalls] = make_call_hierarchy_handler('from')
...
+    if not result then
+      return
+    end
+    local items = {}
+    for _, type_hierarchy_item in pairs(result) do
+      table.insert(items, {
+        filename = assert(vim.uri_to_fname(type_hierarchy_item.uri)),
+        text = type_hierarchy_item.name,
+        lnum = type_hierarchy_item.range.start.line + 1,
+        col = type_hierarchy_item.range.start.character + 1,
```
mfussenegger, line 583: This needs to consider the offset encoding of the client/server, and convert the (likey UTF-16) offset into byte offsets. You can search for `offset_encoding` to see how this translation is done in other places.

mfussenegger (commented):
Left a few comments, but otherwise this look good to me.

jiangyinzuo:
> Left a few comments, but otherwise this look good to me.

In the first commit https://github.com/neovim/neovim/commit/dc279a6df5c1b300805224589639e223093711bd , I borrowed some code from `vim.lsp.buf.outgoing_calls` . Maybe we need another PR to refactor the following code.

https://github.com/neovim/neovim/blob/e1ca7a7bfc954fae6991fc86d5da1a4eda864903/runtime/lua/vim/lsp/buf.lua#L456-L460

https://github.com/neovim/neovim/blob/e1ca7a7bfc954fae6991fc86d5da1a4eda864903/runtime/lua/vim/lsp/buf.lua#L443-L448

https://github.com/neovim/neovim/blob/e1ca7a7bfc954fae6991fc86d5da1a4eda864903/runtime/lua/vim/lsp/handlers.lua#L538-L560

runtime/lua/vim/lsp/buf.lua
```diff
@@ -495,6 +495,62 @@ function M.outgoing_calls()
...
+      request(method, { item = merged_results[1] }, handler)
+    else
+      vim.ui.select(merged_results, {
+        prompt = 'Select a type hierarchy item:',
+        kind = 'typehierarchy',
+      }, function(item)
+        request(method, { item = item }, handler)
+      end)
+    end
+  end)
```
mfussenegger, line 537:
```suggestion
    local merged_results = {}
    for client_id, client_result in pairs(results) do
      if client_result.error then
        vim.notify(client_result.error.message, vim.log.levels.WARN)
      elseif client_result.result then
        for _, item in pairs(client_result.result) do
          table.insert(merged_results, {client_id, item})
        end
      end
    end
    return merged_results
  end

  local bufnr = api.nvim_get_current_buf()
  local params = util.make_position_params()
  --- @param results table<integer, {error: lsp.ResponseError, result: lsp.TypeHierarchyItem[]?}>
  vim.lsp.buf_request_all(bufnr, ms.textDocument_prepareTypeHierarchy, params, function(results)
    local merged_results = merge_results(results)
    if #merged_results == 0 then
      vim.notify('No items resolved', vim.log.levels.INFO)
      return
    end

    if #merged_results == 1 then
      local item = merged_results[1]
      local client = vim.lsp.get_client_by_id(item[1])
      client.request(method, { item = item[2] }, nil, bufnr)
    else
      local opts = {
        prompt = 'Select a type hierarchy item:',
        kind = 'typehierarchy',
        format_item = function(item)
          return item[2].name
        end
      }

      vim.ui.select(merged_results, opts, function(item)
        local client = vim.lsp.get_client_by_id(item[1])
        client.request(method, { item = item[2] }, nil, bufnr)
      end)
    end
  end)
```

Suggestions are untested, but they try to address a couple of things:

- Only one notification if there are no merged results (instead of 1 per client without results)
- bufnr is frozen, so that buffer switches don't impact the context of the async functions
- Includes client-id in the merged results, to be able to use the right client for the follow up request. Otherwise the follow up request is sent to all clients, and some of them wouldn't be able to handle the item. It would also then trigger the handler many times, instead of just once - causing the quickfix list to get populated and overriden for each client
- Adds `format_item` to the ui.select opts. Not sure if `name` is suffiction. Maybe including `detail` if available makes it easier to select the right option?
- Removes the `handler` lookup. The default behavior is to look for handlers in the client, followed by the global lookup.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -495,6 +495,62 @@ function M.outgoing_calls()
...
+end
+
+--- Lists all the subtypes of the symbol under the
+--- cursor in the |quickfix| window. If the symbol can resolve to
+--- multiple items, the user can pick one in the |inputlist()|.
```
mfussenegger, line 542:
```suggestion
--- multiple items, the user can pick one using |vim.ui.select()|.
```

runtime/lua/vim/lsp/buf.lua
```diff
@@ -495,6 +495,62 @@ function M.outgoing_calls()
...
+function M.subtypes()
+  type_hierarchy(ms.typeHierarchy_subtypes)
+end
+
+--- Lists all the supertypes of the symbol under the
+--- cursor in the |quickfix| window. If the symbol can resolve to
+--- multiple items, the user can pick one in the |inputlist()|.
```
mfussenegger, line 549:
```suggestion
--- multiple items, the user can pick one in the |vim.ui.select()|.
```

mfussenegger (commented):
Thanks for addressing my comments. I left a couple more suggestions. 

Should be good to merge after they're addressed.

mfussenegger (approved):

jiangyinzuo:
> Thanks for addressing my comments. I left a couple more suggestions.
> 
> Should be good to merge after they're addressed.

clangd does not returns `detail` field, but jdtls does. I have added `detail` field to the quickfix `text`.
In the following screenshot, `mylist` is the `detail` that jdtls returns.
image

runtime/lua/vim/lsp/buf.lua
```diff
@@ -495,6 +495,90 @@ function M.outgoing_calls()
...
+--- Lists all the subtypes of the symbol under the
+--- cursor in the |quickfix| window. If the symbol can resolve to
+--- multiple items, the user can pick one using |vim.ui.select()|.
+function M.subtypes()
+  type_hierarchy(ms.typeHierarchy_subtypes)
+end
+
+--- Lists all the supertypes of the symbol under the
+--- cursor in the |quickfix| window. If the symbol can resolve to
+--- multiple items, the user can pick one in the |vim.ui.select()|.
```
MariaSolOs, line 577: Nit, but just to remain consistent with the documentation above:

```suggestion
--- multiple items, the user can pick one using |vim.ui.select()|.
```

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -565,6 +565,47 @@ M[ms.callHierarchy_incomingCalls] = make_call_hierarchy_handler('from')
 --- @see # https://microsoft.github.io/language-server-protocol/specifications/specification-current/#callHierarchy_outgoingCalls
 M[ms.callHierarchy_outgoingCalls] = make_call_hierarchy_handler('to')
 
+--- @private
```
MariaSolOs, line 568: No need to mark local functions as `private` (they are already scoped to the module).

MariaSolOs (approved):

Issue #28384, opened 2024-04-17
Fixing `'shell'` on Windows, properly


LunarLambda:
### Problem

See https://github.com/neovim/neovim/issues/16957 for some of the existing issues with using a Unix shell in Neovim on Windows. 

I've dug into it now, seeing as no workaround really managed to fully take, and here's what I was able to find.

The root of evil is the fact that `CreateProcessA` takes some long string as the command line. The Microsoft C runtime takes that string, runs it through splitting logic described [here](https://learn.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments?view=msvc-170), then passes the processed result to `main`. The result of that processing is what `'shell'` ultimately receives.

Vim on the other hand, wants to be able to think about running programs in terms of regular C `argv` arrays.

Assuming `shell=bash`, `shellcmdflag=-c` and `shellxquote="`, running `!ls -l`, produces the argument vector `[bash, -c, "ls -l"]` (note the quotes around `ls -l`). The resulting argument vector is then [passed to libuv](https://github.com/neovim/neovim/blob/master/src/nvim/event/libuv_process.c#L22), which then [reconstitutes the entire thing back to a single string](https://github.com/libuv/libuv/blob/v1.x/src/win/process.c#L521) so it can pass it to `CreateProcessA`.

This is where the issue arises. libuv expects an *unquoted* argument vector (`[bash, -c, ls -l]`, no quotes), as you would pass to `execve` on Unix, but gets `[bash, -c, "ls -l"]` instead. This causes it to *quote the quoted string* on Windows, which ultimately causes bash to receive the command `"ls -l"` (one shell word, with quotes) instead of `ls -l` (two shell words, no quotes).

So, basically, there are two places in the chain that are trying to make the argument vector "windows-friendly": Vim's `shell_build_argv`, and libuv's `make_program_args`. Only one of the two should be done. To only use vim's behaviour, it is probably sufficient to change [libuv_process_spawn](https://github.com/neovim/neovim/blob/master/src/nvim/event/libuv_process.c#L32) to always pass `UV_PROCESS_WINDOWS_VERBATIM_ARGUMENTS`. Since `cmd.exe` already has to use this behaviour since its argument parsing does *not* follow that of the C runtime (i.e. all other programs), this seems like it will cause the least amount of breakage.

[...]

PR #28389, opened 2024-04-17
build(manifest): set longPathAware in Manifest on Windows

LunarLambda:
Was briefly mentioned as part of https://github.com/neovim/neovim/issues/28384

Allows Windows file APIs (and anything that uses them) to bypass the 260-character `MAX_PATH` limitation on Windows 10 1607 or later. This does not otherwise effect the behaviours of any File APIs, so this is a strictly backwards compatible change.

NOTE: This change by itself does not change the behaviour of running Neovim. The system must also have the Windows registry key `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled` set to a `REG_DWORD` with value 1.

See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell#enable-long-paths-in-windows-10-version-1607-and-later for more information.

### DIFF
--- a/src/nvim/os/nvim.manifest
+++ b/src/nvim/os/nvim.manifest
@@ -18,8 +18,9 @@
     </application>
   </compatibility>
   <asmv3:application>
-    <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">
-      <activeCodePage>UTF-8</activeCodePage>
+    <asmv3:windowsSettings>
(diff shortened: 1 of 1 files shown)


LunarLambda:
There are maybe 1 or 2 spots in Neovim's codebase that currently limit themselves to `MAX_PATH` on Windows (particularly, resolving `.lnk` shortcut files). The core editor I/O does not have such limits. Fixing those 1 or 2 places could be seen as a future, optional improvement.

(One would have to look if the OLE APIs used in that function are even affected by the long paths feature anyway...)

justinmk (approved):

PR #28305, opened 2024-04-12
fix(contrib): nix flake dependencies

lxsymington:
As mentioned in this discussion:

https://github.com/neovim/neovim/discussions/27294

The Nix flake build in the `contrib` folder is currently failing. The discussion mentions that this is likely due to an incompatible `tree-sitter` version. A `nix flake update` resolves this issue bringing in later versions of the dependencies.

For those with nix tooling this can be verified by running:

```
nix run 'github:neovim/neovim?dir=contrib'
```

where you will get the following err:

```
$ nix run 'github:neovim/neovim?dir=contrib'                                               
error: builder for '/nix/store/lzjc50adxxxl21kq9ysgy4sijf78r0sj-neovim-unwrapped-8cca787.drv' failed with exit code 2;
       last 10 log lines:
       > void ts_query_cursor_set_match_limit(TSQueryCursor *, uint32_t);
       >      ^
       > 1 error generated.
       > make[2]: *** [src/nvim/CMakeFiles/nvim_bin.dir/build.make:3433: src/nvim/CMakeFiles/nvim_bin.dir/lua/treesitter.c.o] Error 1
       > make[2]: *** Waiting for unfinished jobs....
       > [ 69%] Building C object src/nvim/CMakeFiles/nvim_bin.dir/map_glyph_cache.c.o
       > [ 69%] Building C object src/nvim/CMakeFiles/nvim_bin.dir/mapping.c.o
       > [ 69%] Building C object src/nvim/CMakeFiles/nvim_bin.dir/mark.c.o
       > make[1]: *** [CMakeFiles/Makefile2:624: src/nvim/CMakeFiles/nvim_bin.dir/all] Error 2
       > make: *** [Makefile:156: all] Error 2
       For full logs, run 'nix log /nix/store/lzjc50adxxxl21kq9ysgy4sijf78r0sj-neovim-unwrapped-8cca787.drv'.
```

Whereas running:

```
nix run 'github:lxsymington/neovim?dir=contrib&ref=fix/contrib/nix-flake'
```

will enter you into a build of neovim

PR #28394, opened 2024-04-18
build(nix): update flake to prevent build errors

ribru17:
- Updates nixpkgs to source a necessary Tree-sitter version
- Updates to a new llvm version as the old one was removed from nixpkgs
- Properly moves `doCheck` from a regular param to an attribute option

I think the documentation may still need an update, because `nix-shell '<nixpkgs>' -A neovim-unwrapped` remains broken, but now the main build works as does `nix develop` (which seems to give basically the same effect as the above command). Sadly I don't think I know enough about nix to provide proper documentation

### DIFF
--- a/contrib/flake.lock
+++ b/contrib/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1701680307,
-        "narHash": "sha256-kAuep2h5ajznlPMD9rnQyffWG8EM/C73lejGofXvdM8=",
+        "lastModified": 1710146030,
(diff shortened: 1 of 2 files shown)


clason:
I think the flake is meant to be moved anyway, so not sure changing it makes sense here? @teto

justinmk:
see https://github.com/neovim/neovim/pull/28305#issuecomment-2052066183 : 

> The plan is to remove the flake from the nix repository because of the overhead it creates to contributors and move it to https://github.com/nix-community/neovim-nightly-overlay/. I haven't had the time to do so. Is this something you would be interested in ?

ribru17:
I made a [PR](https://github.com/nix-community/neovim-nightly-overlay/pull/483) to copy the flake over, once/if it gets merged I think it will be fine to remove it here, and mention that in `BUILD.md` of course

PR #28395, opened 2024-04-18
refactor(options)!: use forward slash as path separator on Windows

famiu:
Problem: All modern Windows versions support using a forward slash as path separator on most applications, including the two most common shells, cmd.exe and PowerShell. Therefore, using backward slash as path separator is largely useless.

Solution: Use forward slash as path separator on every platform. Also remove `completeslash` and `shellslash` as they are no longer needed.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -6957,10 +6957,8 @@ sha256({string})                                                      *sha256()*
 shellescape({string} [, {special}])                              *shellescape()*
 		Escape {string} for use as a shell command argument.
 
-		On Windows when 'shellslash' is not set, encloses {string} in
-		double-quotes and doubles all double-quotes within {string}.
-		Otherwise encloses {string} in single-quotes and replaces all
-		"'" with "'\''".
+		On Windows, encloses {string} in single-quotes and replaces
+		all "'" with "'\''".
 
 		If {special} is a |non-zero-arg|:
 		- Special items such as "!", "%", "#" and "<cword>" will be
--- a/src/nvim/path.c
+++ b/src/nvim/path.c
@@ -1409,37 +1409,23 @@ static int expand_backtick(garray_T *gap, char *pat, int flags)
   return cnt;
 }
 
-#ifdef BACKSLASH_IN_FILENAME
-/// Replace all slashes by backslashes.
-/// This used to be the other way around, but MS-DOS sometimes has problems
-/// with slashes (e.g. in a command name).  We can't have mixed slashes and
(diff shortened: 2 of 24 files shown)


src/nvim/path.c
```diff
@@ -1409,38 +1409,6 @@ static int expand_backtick(garray_T *gap, char *pat, int flags)
...
-#ifdef BACKSLASH_IN_FILENAME
-/// Replace all slashes by backslashes.
-/// This used to be the other way around, but MS-DOS sometimes has problems
-/// with slashes (e.g. in a command name).  We can't have mixed slashes and
-/// backslashes, because comparing file names will not work correctly.  The
-/// commands that use a file name should try to avoid the need to type a
-/// backslash twice.
-/// When 'shellslash' set do it the other way around.
-/// When the path looks like a URL leave it unmodified.
-void slash_adjust(char *p)
```
zeertzjq, line 1421: Why is this removed? Won't that cause backslashes in paths return by libuv to remain backslashes?

zeertzjq (changes requested):
This should make `exists('+shellslash')` always return `false`, which isn't the case with `immutable = true`. There are two options:
- Use `enable_if = false` instead if `immutable = true`.
- Make `exists('+option')` return `false` for options with `immutable = true`.

src/nvim/path.c
clason: That assumes that libuv returns backslashes, which afaik is not the case.

zeertzjq: I just tested with both luv and FFI on Windows. It is the case.

runtime/doc/builtin.txt
```diff
@@ -6957,10 +6957,8 @@ sha256({string})                                                      *sha256()*
 shellescape({string} [, {special}])                              *shellescape()*
 		Escape {string} for use as a shell command argument.
 
-		On Windows when 'shellslash' is not set, encloses {string} in
-		double-quotes and doubles all double-quotes within {string}.
-		Otherwise encloses {string} in single-quotes and replaces all
-		"'" with "'\''".
+		On Windows, encloses {string} in single-quotes and replaces
+		all "'" with "'\''".
```
justinmk, line 6961: This is not correct. If the shell is cmd.exe, double quotes should be used (and "doubles all double-quotes"). shellescape() inherently depends on 'shell'.

zeertzjq:
vim-go uses 'noshellslash': https://github.com/fatih/vim-go/blob/ac1fb9611de435d80a6badcf8482cb1428449b10/plugin/go.vim#L105

famiu:
Superseded by #28400

PR #28397, opened 2024-04-18
fix(diagnostic): invalid col number  compare in next_diagnostic

### LINKED ISSUE #26079: When lsp diagnostics are on wrong line, goto_prev stops working
### Problem

I'm not sure if this is a problem only with pylsp.pylint diagnostic or a general problem.

From this state
```py
I      ■ [missing-module-docstring] Missing module docstring

I def foo():     ■ [disallowed-name] Disallowed name "foo"
      pass
``` 
By entering insert mode and adding a newline and then exit insert mode. the diagnostics will update to look like below:
```py
I      ■ [missing-module-docstring] Missing module docstring

I      ■ [disallowed-name] Disallowed name "foo"
W def foo():     ■ E303 too many blank lines (3)
      pass
```

As you can see the `[disallowed-name]` diagnostic from pylint is on the wrong line now. Also if the cursor is on this line and we try to execute `lua vim.diagnostic.goto_prev()`, nothing will happen.

### Steps to reproduce

config:

```vim
Plug 'neovim/nvim-lspconfig'
lua local lsp = require('lspconfig')
lua lsp.pylsp.setup { settings = { pylsp = { plugins = { pylint = { enabled = true }}}}}
```

### Expected behavior

goto_prev to not "hang" even if the diagnostic position is incorrect

### Neovim version (nvim -v)

0.9.4

N/A?

### Operating system/version

Ubuntu 22.04

### Terminal name/version

wezterm over ssh

xterm-256color

### Installation

download and unpack linux64 tar.gz release

glepnir:
Problem: when line is a blank line there will got an invalid column number in math.min compare.

Solution: make sure the min col number is 0 not an illegal number.

Fix #26079

### DIFF
--- a/test/functional/lua/diagnostic_spec.lua
+++ b/test/functional/lua/diagnostic_spec.lua
@@ -978,6 +978,29 @@ describe('vim.diagnostic', function()
       ]]
       )
     end)
+
+    it('works on blank line #28397', function()
+      eq(
+        { 0, 2 },
+        exec_lua [[
+        local test_bufnr = vim.api.nvim_create_buf(true, false)
+        vim.api.nvim_buf_set_lines(test_bufnr, 0, -1, false, {
+          'first line',
+          '',
+          '',
+          'end line',
+        })
+        vim.diagnostic.set(diagnostic_ns, test_bufnr, {
+          make_info('Diagnostic #1', 0, 2, 0, 2),
+          make_info('Diagnostic #2', 2, 0, 2, 0),
+          make_info('Diagnostic #3', 2, 0, 2, 0),
+        })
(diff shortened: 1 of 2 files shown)


glepnir:
cc @gpanders

test/functional/lua/diagnostic_spec.lua
```diff
@@ -978,6 +978,29 @@ describe('vim.diagnostic', function()
       ]]
       )
     end)
+
+    it('works on blank line #28397', function()
```
justinmk, line 982: I tried this locally, the test passes even if I revert the changes to `runtime/lua/vim/diagnostic.lua`

glepnir: fixed

PR #26850, opened 2024-01-02
refactor(lsp): rename domain_socket_connect -> connect_pipe

mfussenegger:

PR #28398, opened 2024-04-18
refactor(lsp): merge rpc.domain_socket_connect into rpc.connect

mfussenegger:
See discussion in https://github.com/neovim/neovim/pull/26850

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -2229,32 +2229,20 @@ Lua module: vim.lsp.rpc                                              *lsp-rpc*
       • {terminate}   (`fun()`)
 
 
-connect({host}, {port})                                *vim.lsp.rpc.connect()*
-    Create a LSP RPC client factory that connects via TCP to the given host
-    and port.
+connect({host_or_path}, {port})                        *vim.lsp.rpc.connect()*
+    Create a LSP RPC client factory that connects to either:
+    • a named pipe (windows)
+    • a domain socket (unix)
+    • a host and port via TCP
 
     Return a function that can be passed to the `cmd` field for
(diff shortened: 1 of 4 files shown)


justinmk (approved):
Nice!

mfussenegger:
@TheLeoP heads up as this will break https://github.com/TheLeoP/powershell.nvim

TheLeoP:
Hey. I just tried testing this and it seems like it introduced a regression. As I stated in [this comment](https://github.com/neovim/neovim/pull/26032#issue-1991183835):

> I'm using vim.schedule on the write transport as a workaround because, on my tests using Powershell Editor Services, Neovim sended the initialization requets before the pipe connection was stablished.

At the time I didn't look too much into it, I guess this may be some kind of race condition because of Windows pipes being slow to connect to(?). But, since `vim.schedule_wrap` was removed from the write handler, I'm no longer able to connect to the LSP using them.

mfussenegger:
> Hey. I just tried testing this and it seems like it introduced a regression. As I stated in [this comment](https://github.com/neovim/neovim/pull/26032#issue-1991183835):
> 
> > I'm using vim.schedule on the write transport as a workaround because, on my tests using Powershell Editor Services, Neovim sended the initialization requets before the pipe connection was stablished.
> 
> At the time I didn't look too much into it, I guess this may be some kind of race condition because of Windows pipes being slow to connect to(?). But, since `vim.schedule_wrap` was removed from the write handler, I'm no longer able to connect to the LSP using them.

Oh, I was wondering what the `vim.schedule_wrap` was for - because there was no fast-loop issue there, so I didn't see a need for it.

Could you try with https://github.com/neovim/neovim/pull/28507 ?

PR #28400, opened 2024-04-18
feat(options)!: disallow setting hidden options

famiu:
Problem: Currently, there are three different ways of marking an option as hidden, `enable_if = false`, `hidden = true` and `immutable = true`. These also have different behaviors. Options hidden with `enable_if = false` can't have their value fetched using Vim script or the API, but options hidden with `hidden = true` or `immutable = true` can. On the other hand, options with `hidden = true` do not error when trying to set their value, but options with `immutable = true` do.

Solution: Remove `enable_if = false`, remove the `hidden` property for options, and use `immutable = true` to mark an option as hidden instead. Also make hidden option variable pointers always point to the default value, which allows fetching the value of every hidden option using Vim script and the API.

BREAKING CHANGE: Trying to set a hidden option will now give an error instead of just being ignored.


test/functional/options/num_options_spec.lua
```diff
@@ -100,19 +101,20 @@ describe(':set validation', function()
...
 
   it('set maxcombine resets to 6', function()
     local function setto(value)
       feed_command('setglobal maxcombine=' .. value)
       feed_command('setlocal maxcombine=' .. value)
-      api.nvim_set_option_value('maxcombine', value, {})
-      eq(6, api.nvim_get_option_value('maxcombine', {}))
+      feed_command('set maxcombine=' .. value)
+      eq(6, eval('&maxcombine'))
+      matches('maxcombine=6', exec_capture('set maxcombine?'))
```
justinmk, line 117: why use matches for this instead of eq() with the actual option value? this would incorrectly pass if e.g. `maxcombine=666` was returned

src/nvim/option.h
```diff
@@ -49,8 +56,8 @@ typedef struct {
                             ///< buffer-local option: global value
   idopt_T indir;            ///< global option: PV_NONE;
                             ///< local option: indirect option index
-  bool hidden;              ///< option is hidden, any attempt to set its value will be ignored.
-  bool immutable;           ///< option is immutable, trying to set its value will give an error.
+  bool hidden;              ///< option is hidden, cannot get/set its value
+  OptImmutable immutable;   ///< option immutability, see OptImmutable
```
justinmk, line 60: if this is going to be an enum, can "hidden" also be expressed in the same enum? Perhaps this should just be part of `OptTypeFlags` ?

```suggestion
  OptType2 hidde;   ///< option hidden and/or immutable.
```

justinmk:
Seems like a nice simplification. Do we have tests for all the affected options?

famiu:
> Seems like a nice simplification. Do we have tests for all the affected options?

Yes, we have tests for `'maxcombine'`, which is now a hidden option, so that should cover everything

src/nvim/option.c
```diff
@@ -3345,7 +3332,7 @@ static char *option_get_valid_types(OptIndex opt_idx)
 /// @return  True if option is hidden, false otherwise. Returns false if option name is invalid.
 bool is_option_hidden(OptIndex opt_idx)
 {
-  return opt_idx == kOptInvalid ? false : get_varp(&options[opt_idx]) == NULL;
+  return opt_idx == kOptInvalid ? false : options[opt_idx].hidden;
```
zeertzjq, line 3335: `exists('+option')` should probably return `false` for `immutable` option as well, but may that can be done in another PR.

src/nvim/options.lua
```diff
@@ -91,11 +91,11 @@ return {
     {
       abbreviation = 'al',
       defaults = { if_true = 224 },
-      enable_if = false,
```
zeertzjq, line 94: Should `false` be removed from the type annotation of `enable_if` now?

src/nvim/option.c
famiu: Should it? The way I see it, `immutable` options are options that do exist, but just cannot have their value changed. The way I implemented it, options could be set to `immutable` temporarily as well, if we ever want them to not be changed.

zeertzjq: Hmm, I don't think this matters much for now.

famiu:
I made some new changes here with regards to `immutable` options. Specifically, I made it so that `immutable` options can have a value that's not its `def_val`. This was already the case in `gen_options.lua`, but was not reflected in the C code.

src/nvim/option.c
```diff
@@ -4620,9 +4603,9 @@ void *get_option_varp_scope_from(OptIndex opt_idx, int scope, buf_T *buf, win_T
 
 void *get_varp_from(vimoption_T *p, buf_T *buf, win_T *win)
 {
-  // hidden option, always return NULL
-  if (p->var == NULL) {
-    return NULL;
+  // hidden options always use the same var pointer
```
justinmk, line 4606: should we add an `p->var != NULL` here ?

famiu: > should we add an `p->var != NULL` here ? and the other functions where `p->var == NULL` was being checked.

`p->var` cannot be NULL anymore, unless something has gone seriously wrong

src/nvim/generators/gen_options.lua
```diff
@@ -167,8 +167,11 @@ local function dump_option(i, o)
 
   -- An option cannot be both hidden and immutable.
   assert(not o.hidden or not o.immutable)
+  -- Hidden options cannot have a varname.
+  assert(not o.hidden or not o.varname)
+  -- `enable_if` cannot be used with `hidden`.
+  assert(not o.enable_if or not o.hidden)
```
justinmk, line 173: the comments are helpful, but might as well pass them as messages too
```suggestion
  assert(not o.hidden or not o.varname, 'hidden options cannot have a varname')
  assert(not o.enable_if or not o.hidden, '`enable_if` cannot be used with `hidden`')
```

src/nvim/generators/gen_options.lua
```diff
@@ -211,7 +214,10 @@ local function dump_option(i, o)
   end
   if o.enable_if then
     w('#else')
-    w('    .var=NULL')
+    -- Hidden option directly points to default value.
+    w(('    .var=&options[%u].def_val'):format(i - 1))
+    -- Option is always hidden on the false branch of `enable_if`.
+    w('    .hidden=true')
```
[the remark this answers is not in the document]
famiu: False branch as in the `#else` preprocessor branch of the C code, which only exists if `o.enable_if` is provided and the condition is false

justinmk:
> allow fetching the value of a hidden option using Vim script and the API.

This is changing the semantics of hidden options. 

- that is the main change of this PR, it's totally different than the PR title (which just sounds like a refactor)
-  `:help hidden-options` would need to be updated.
- now what is the difference vs immutable options, exactly?

famiu:
> > allow fetching the value of a hidden option using Vim script and the API.
> 
> This is changing the semantics of hidden options.
> 
> * that is the main change of this PR, it's totally different than the PR title (which just sounds like a refactor)
> * `:help hidden-options` would need to be updated.
> * now what is the difference vs immutable options, exactly?

This change was suggested by @bfredl, so I think he can explain the reasoning better

As for the difference vs immutable options, immutable options are options that exist but can't be changed. This could also be set as a temporary flag for an option if we ever want to disallow changing the value of an option. Trying to change an immutable option will give an error, unlike hidden options where trying to change it is just ignored.

justinmk:
That means we are removing the main "feature" of hidden options. Now the only remaining difference is that hidden options don't error if set. The purpose of that was to avoid bothering users that might have some old lines in their configs.

```
pastetoggle
aleph
browsedir
guioptions
guitabtooltip
imcmdline
imdisable
mouseshape
opendevice
pastetoggle
termencoding
```

Rather than continuing to maintain what is now a very narrow difference for these 10 "hidden" options, we can just treat them as immutable.

And we can later decide whether immutable options should error or not. But that can be a separate PR, if at all.

famiu:
> That means we are removing the main "feature" of hidden options. Now the only remaining difference is that hidden options don't error if set. The purpose of that was to avoid bothering users that might have some old lines in their configs.
> 
> ```
> pastetoggle
> aleph
> browsedir
> guioptions
> guitabtooltip
> imcmdline
> imdisable
> mouseshape
> opendevice
> pastetoggle
> termencoding
> ```
> 
> Rather than continuing to maintain what is now a very narrow difference for these 10 "hidden" options, we can just treat them as immutable.
> 
> And we can later decide whether immutable options should error or not. But that can be a separate PR, if at all.

Alternatively, I suggest removing `immutable` for now and keeping `hidden`, as that's more useful

justinmk:
> Alternatively, I suggest removing `immutable` fo

more useful in what way? we want to show errors for the existing immutable options, that was an intentional decision to avoid silent, wrong behavior. That is entirely different than users having to remove a couple lines in their config.

famiu:
> > Alternatively, I suggest removing `immutable` fo
> 
> more useful in what way? we want to show errors for the existing immutable options, that was an intentional decision to avoid silent, wrong behavior. That is entirely different than users having to remove a couple lines in their config.

How do you think we should handle erroring for invalid options for now? Since if we remove one of `immutable` or `hidden`, we have to keep the behavior of the other. Should all unavailable options error when trying to set them, or should it just be ignored?

justinmk:
> Should all unavailable options error when trying to set them, or should it just be ignored?

review https://github.com/neovim/neovim/pull/28400#issuecomment-2067644808

famiu:
> > Should all unavailable options error when trying to set them, or should it just be ignored?
> 
> review [#28400 (comment)](https://github.com/neovim/neovim/pull/28400#issuecomment-2067644808)

If I'm not wrong, you're suggesting that all options are made immutable, this does mean all hidden options will now error when trying to set them **for now**, is that okay?

justinmk:
> all hidden options will now error when trying to set them **for now** (until we discuss and decide to change that), is that okay?

correct. for those 10 options, which are pretty uncommon, users will have to delete any lines from their configs that assign those options. We can choose to pay this cost, since we are already choosing to change the semantics of hidden options, we might as well go all the way and remove the remaining complexity of the last remaining distinction.

famiu:
> > all hidden options will now error when trying to set them **for now** (until we discuss and decide to change that), is that okay?
> 
> correct. for those 10 options, which are pretty uncommon, users configs/plugins will have to delete any lines that _set_ those options. We can choose to pay this cost, since we are already choosing to change the semantics of hidden options, we might as well go all the way and remove the remaining complexity of the last remaining distinction.

Also what about trying to get the option value? Should that be allowed? @bfredl suggested to me in Matrix that getting the value of immutable / hidden options should be allowed.

justinmk:
> Also what about trying to get the option value? Should that be allowed? @bfredl suggested to me in Matrix that getting the value of immutable / hidden options should be allowed.

yes, my previous comments were written with the assumption that that was already settled

zeertzjq:
With this PR `:set shellslash` will be an error on non-Windows, while `:set noshellslash` will not. However, `:set shellslash` is more commonly used, so I wonder if it's better to make 'shellslash' default to true on non-Windows to reduce breakage. That'll also avoid all the test changes for 'shellslash', but may break vim-go: https://github.com/fatih/vim-go/blob/ac1fb9611de435d80a6badcf8482cb1428449b10/autoload/go/util.vim#L335

justinmk:
> `set shellslash` will be an error on non-Windows

platform-specific options are an antipattern. can we allow 'shellslash' on non-Windows but make it have no effect? (I know that's the purpose of "hidden options", but this one option is not enough to create a new class of options; we already have platform-specific handling of various options without designating the option as a special kind)

zeertzjq:
`has('+shellslash')` is commonly used to check if setting 'shellslash' has an effect. If `has('+shellslash')` always returns 1, it will cause even more breakage.

famiu:
> > `set shellslash` will be an error on non-Windows
> 
> platform-specific options are an antipattern. can we allow 'shellslash' on non-Windows but make it have no effect? (I know that's the purpose of "hidden options", but this one option is not enough to justify a special kind of options; we already have platform-specific handling of various "normal" options)

This is the purpose of https://github.com/neovim/neovim/pull/28395 btw

famiu:
> `If `has('+shellslash')` always returns 1

Not after this PR (unless that's what you're talking about?)

zeertzjq:
Not before this PR either

test/old/testdir/test_compiler.vim
```diff
@@ -10,9 +10,12 @@ func Test_compiler()
   let save_LC_ALL = $LC_ALL
   let $LC_ALL= "C"
 
-  " %:S does not work properly with 'shellslash' set
-  let save_shellslash = &shellslash
-  set noshellslash
+  " Nvim doesn't allow setting value of a hidden option to non-default value
+  if exists('+shellslash')
+    let save_shellslash = &shellslash
```
zeertzjq, line 15:
```suggestion
  let save_shellslash = &shellslash
  " Nvim doesn't allow setting value of a hidden option to non-default value
  if exists('+shellslash')
```

test/old/testdir/test_compiler.vim
```diff
@@ -35,7 +38,9 @@ func Test_compiler()
   \ .               'requires explicit package name', a)
 
 
-  let &shellslash = save_shellslash
+  if exists('save_shellslash')
+    let &shellslash = save_shellslash
+  endif
```
zeertzjq, line 43:
```suggestion
  let &shellslash = save_shellslash
```

src/nvim/options.lua
```diff
@@ -1493,7 +1493,7 @@ return {
       cb = 'did_set_completeslash',
       defaults = { if_true = '' },
       desc = [=[
-        		only for MS-Windows
+        		only changeable in MS-Windows
```
zeertzjq, line 1496:
```suggestion
        		only modifiable in MS-Windows
```

src/nvim/options.lua
```diff
@@ -7210,9 +7210,14 @@ return {
...
-      defaults = { if_true = false },
+      defaults = {
+        condition = 'MSWIN',
+        if_true = false,
+        if_false = true,
+        doc = 'on, Windows: off',
+      },
       desc = [=[
-        		only for MS-Windows
+        		only changeable in MS-Windows
```
zeertzjq, line 7220:
```suggestion
        		only modifiable in MS-Windows
```

test/functional/options/num_options_spec.lua
```diff
@@ -108,18 +108,4 @@ describe(':set validation', function()
...
-      feed_command('setlocal maxcombine=' .. value)
-      api.nvim_set_option_value('maxcombine', value, {})
-      eq(6, api.nvim_get_option_value('maxcombine', {}))
-      eq('', eval('v:errmsg'))
-    end
-    setto(0)
-    setto(1)
-    setto(6)
-    setto(7)
-  end)
```
zeertzjq, line 124: cc @bfredl

runtime/doc/news.txt
```diff
@@ -110,6 +110,7 @@ OPTIONS
   of just string |global-local| options.
 • `:setlocal {option}<` copies the global value to the local value for number
   and boolean |global-local| options instead of removing the local value.
+• Setting |hidden-options| now gives an error.
```
zeertzjq, line 113:
```suggestion
• Setting |hidden-options| now gives an error. In particular, setting
  'noshellslash' is now only allowed on Windows.
```

zeertzjq (commented):
How about changing 'set noshellslash' to `set shellslash&` in tests (and `zip.vim`) instead?

runtime/doc/options.txt
```diff
@@ -641,11 +641,12 @@ Hidden options						*hidden-options*
 
 Not all options are supported in all versions.  This depends on the supported
 features and sometimes on the system.  A remark about this is in curly braces
-below.  When an option is not supported it may still be set without getting an
-error, this is called a hidden option.  You can't get the value of a hidden
-option though, it is not stored.
+below.  When an option is not supported, it is called a hidden option.  Trying
+to get the value of a hidden option will not give an error, it will return the
+default value for that option instead. You can't set the value of a hidden
```
zeertzjq, line 646:
```suggestion
default value for that option instead. You can't change the value of a hidden
```

src/nvim/options.lua
zeertzjq: You still haven't made this change.

famiu:
> How about changing 'set noshellslash' to `set shellslash&` in tests (and `zip.vim`) instead?

Imho it doesn't convey the intent properly. Plus MS-Windows specific tests still use `set noshellslash`, so it adds inconsistency as well

zeertzjq:
Yes, that makes sense.

src/nvim/options.lua
zeertzjq: @famiu Did you seem this comment?

famiu: > @famiu Did you seem this comment?

Oh sorry I missed it

runtime/doc/options.txt
```diff
@@ -641,11 +641,12 @@ Hidden options						*hidden-options*
 
 Not all options are supported in all versions.  This depends on the supported
 features and sometimes on the system.  A remark about this is in curly braces
-below.  When an option is not supported it may still be set without getting an
-error, this is called a hidden option.  You can't get the value of a hidden
-option though, it is not stored.
+below.  When an option is not supported, it is called a hidden option.  Trying
+to get the value of a hidden option will not give an error, it will return the
```
justinmk, line 645: I thought this concept is being removed entirely? So `hidden-options` should move to `nvim-removed`

src/nvim/option.h
```diff
@@ -49,7 +49,6 @@ typedef struct {
                             ///< buffer-local option: global value
   idopt_T indir;            ///< global option: PV_NONE;
                             ///< local option: indirect option index
-  bool hidden;              ///< option is hidden, any attempt to set its value will be ignored.
   bool immutable;           ///< option is immutable, trying to set its value will give an error.
```
justinmk, line 52:
```suggestion
  bool immutable;           ///< option is "hidden", trying to set it will give an error.
```

src/nvim/api/deprecated.c
```diff
@@ -647,10 +647,6 @@ static bool option_has_scope(OptIndex opt_idx, OptReqScope req_scope)
 
   vimoption_T *opt = get_option(opt_idx);
 
-  // Hidden option.
-  if (opt->var == NULL) {
-    return false;
-  }
```
justinmk, line 653: should we assert here , or do hidden options now have a scope?

src/nvim/option_vars.h
```diff
@@ -530,6 +530,8 @@ EXTERN char *p_mef;             ///< 'makeef'
 EXTERN char *p_mp;              ///< 'makeprg'
 EXTERN char *p_mps;             ///< 'matchpairs'
 EXTERN OptInt p_mat;            ///< 'matchtime'
+EXTERN OptInt p_mco;            ///< 'maxcombine'
+#define MAX_MCO  6  // fixed value for 'maxcombine'
```
justinmk, line 534: why is this defined in 2 different places?

famiu: > why is this defined in 2 different places?

That was an error on my part. Fixed.

justinmk (approved):

happy-dude:
Chimed in in the matrix channel, but wanted to mention that this seems to break https://github.com/fatih/vim-go in the latest neovim nightlies -- particularly the latest (no)shellslash changes.

ref: https://matrix.to/#/!HCjHPBLFfoFpYNgwdE:matrix.org/$S1LkCWRvKrgw_5gA35eyU6ERyr441lr6VKFg7Irjot0?

PR #28402, opened 2024-04-18
docs: update supported platforms

dundargoc:


clason:
> Tier 3: Not tested and no guarantees, but may work.

"and not all features may work". (Still don't think that qualifies as a "supported platform"; I'd just remove the line -- and the Tier, if you prefer -- but keep the note if you must.

justinmk (approved):

runtime/doc/support.txt
```diff
@@ -13,12 +13,13 @@ Supported platforms					 *supported-platforms*
 
 `System`          `Tier`      `Versions`                  `Tested versions`
 Linux            1      >= 2.6.32, glibc >= 2.12     Ubuntu 22.04
-macOS (Intel)    1      >= 10.15                     macOS 12
-Windows 64-bit   1      >= 8 (see note below)        Windows Server 2022
+macOS (Intel)    1      >= 11                        macOS 12
+macOS (M1)       2      >= 11                        macOS 14
```
zeertzjq, line 17: Why do macOS (Intel) and macOS (M1) have different support tiers?

clason: Because we only test(ed) on Intel before M1 runners were GA.

dundargoc: Yeah, probably just outdated. Send PR

clason: (and chocolate.)

PR #28408, opened 2024-04-18
filetype: map `flake.lock` to `json` filetype

ribru17:
This is the name used for Nix flake lock files.

Helix did it first! https://github.com/helix-editor/helix/pull/8304 :suspect:


clason:
And Vim needs to do it second!

PR #28412, opened 2024-04-19
feat(treesitter): get highlight from parser

altermo:
Sometimes you want to apply treesitter highlighting to non-buffer text (such as foldtext or the message area). The problem is that there is currently no easy way to get the treesitter highlighting except for implementing a function yourself. So let's add such a function.

It has been brought to my attention that this feature is also useful for applying a highlight to a specific region in a buffer, by getting the text in the region, parsing it, getting the highlights, and setting the highlights as extmarks.

### DIFF
--- a/runtime/lua/vim/treesitter.lua
+++ b/runtime/lua/vim/treesitter.lua
@@ -495,4 +495,26 @@ function M.foldexpr(lnum)
   return M._fold.foldexpr(lnum)
 end
 
+---@private
+--- Get highlight information from parser.
+---
+---@param parser vim.treesitter.LanguageTree
+---@return vim.treesitter.highlighter.Info[]
+function M._get_highlight(parser)
+  local highlights = {}
+  parser:for_each_tree(function(tree, ltree)
+    local source = ltree:source()
+    local hl_query = M.query.get(ltree:lang(), 'highlights')
+
+    if not hl_query then
+      return
+    end
+
+    for info in M.highlighter._iter_highlight_info(hl_query, tree:root(), source, 0, -1) do
+      table.insert(highlights, info)
+    end
+  end)
+  return highlights
+end
+
 return M
--- a/runtime/lua/vim/treesitter/highlighter.lua
+++ b/runtime/lua/vim/treesitter/highlighter.lua
@@ -4,12 +4,30 @@ local Range = require('vim.treesitter._range')
 
 local ns = api.nvim_create_namespace('treesitter/highlighter')
 
----@alias vim.treesitter.highlighter.Iter fun(end_line: integer|nil): integer, TSNode, vim.treesitter.query.TSMetadata, TSQueryMatch
+--- @class vim.treesitter.highlighter.Info
+--- @inlinedoc
+--- The range of the highlight
+--- @field range Range6
+--- Highlight |group-name|
+--- @field hl_group string
+--- Highlight priority |treesitter-highlight-priority|
+--- @field priority integer
+--- Highlight conceal |treesitter-highlight-conceal|
+--- @field conceal string?
+--- Highlight spell/nospell |treesitter-highlight-spell|
+---   - If `true` then `@spell` is set
+---   - If `false` then `@nospell` is set
+---   - If `nil` then neither `@spell` nor `@nospell` is set
+--- @field spell boolean?
+--- Extmark URL, see "url" option at |nvim_buf_set_extmark()|
+--- @field url string?
+
+---@alias vim.treesitter.highlighter.Iter fun(end_line:integer):vim.treesitter.highlighter.Info?,TSNode?
 
 ---@class (private) vim.treesitter.highlighter.Query
 ---@field private _query vim.treesitter.Query?
 ---@field private lang string
----@field private hl_cache table<integer,integer>
+---@field private hl_cache table<string,integer>
 local TSHighlighterQuery = {}
 TSHighlighterQuery.__index = TSHighlighterQuery
 
@@ -32,19 +50,18 @@ function TSHighlighterQuery.new(lang, query_string)
 end
 
 ---@package
----@param capture integer
+---@param name string
 ---@return integer?
-function TSHighlighterQuery:get_hl_from_capture(capture)
-  if not self.hl_cache[capture] then
-    local name = self._query.captures[capture]
+function TSHighlighterQuery:get_hl_from_capture(name)
+  if not self.hl_cache[name] then
     local id = 0
     if not vim.startswith(name, '_') then
-      id = api.nvim_get_hl_id_by_name('@' .. name .. '.' .. self.lang)
+      id = api.nvim_get_hl_id_by_name(name)
     end
-    self.hl_cache[capture] = id
+    self.hl_cache[name] = id
   end
 
-  return self.hl_cache[capture]
+  return self.hl_cache[name]
 end
 
 ---@nodoc
@@ -247,11 +264,11 @@ function TSHighlighter:get_query(lang)
 end
 
 --- @param match TSQueryMatch
---- @param bufnr integer
+--- @param source integer|string Buffer id, or text
 --- @param capture integer
 --- @param metadata vim.treesitter.query.TSMetadata
 --- @return string?
-local function get_url(match, bufnr, capture, metadata)
+local function get_url(match, source, capture, metadata)
   ---@type string|number|nil
   local url = metadata[capture] and metadata[capture].url
 
@@ -269,7 +286,7 @@ local function get_url(match, bufnr, capture, metadata)
   -- from the first.
   local other_node = captures[url][1]
 
-  return vim.treesitter.get_node_text(other_node, bufnr, {
+  return vim.treesitter.get_node_text(other_node, source, {
     metadata = metadata[url],
   })
 end
@@ -284,6 +301,57 @@ local function get_spell(capture_name)
     return false, 1
   end
   return nil, 0
+end
+
+---@param hl_query vim.treesitter.Query
+---@param root_node TSNode
+---@param source integer|string
+---@param start_row integer?
+---@param end_row integer?
+---@return vim.treesitter.highlighter.Iter
+function TSHighlighter._iter_highlight_info(hl_query, root_node, source, start_row, end_row)
+  local iter = hl_query:iter_captures(root_node, source, start_row, end_row)
+
+  return function(end_line)
+    local capture, node, metadata, match = iter(end_line)
+    if not node then
+      return
+    end
+
+    if not capture then
+      return nil, node
+    end
+
+    local range = vim.treesitter.get_range(node, source, metadata and metadata[capture])
+
+    local capture_name = hl_query.captures[capture]
+
+    local hl_group = capture_name
+    if not vim.startswith(hl_group, '_') then
+      hl_group = '@' .. hl_group .. '.' .. hl_query.lang
+    end
+
+    local spell, spell_pri_offset = get_spell(capture_name)
+
+    local priority = (
+      tonumber(metadata.priority or metadata[capture] and metadata[capture].priority)
+      or vim.highlight.priorities.treesitter
+    ) + spell_pri_offset
+
+    local conceal = metadata.conceal or metadata[capture] and metadata[capture].conceal
+
+    local url = get_url(match, source, capture, metadata)
+
+    return {
+      range = range,
+      hl_group = hl_group,
+      priority = priority,
+      conceal = conceal,
+      spell = spell,
+      url = url,
+    },
+      node
+  end
 end
 
 ---@param self vim.treesitter.highlighter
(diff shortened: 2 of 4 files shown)


clason:
Thank you, but what's the point of this function? (Needs to be explained in the PR description and commit message.) We don't add API functions "just because we can".

runtime/lua/vim/treesitter.lua
```diff
@@ -466,4 +466,74 @@ function M.foldexpr(lnum)
   return M._fold.foldexpr(lnum)
 end
 
+---@param parser vim.treesitter.LanguageTree
+---@return {group:string,range:integer[]}[]
+function M.get_text_highlight(parser)
```
justinmk, line 471: `parser` is a `vim.treesitter.LanguageTree`, which is already a complex object. Should this instead be a method on that object?

altermo: Ok

justinmk:
@chipsenkbeil is this what you were looking for in https://github.com/neovim/neovim/discussions/27940 ?

chipsenkbeil:
> @chipsenkbeil is this what you were looking for in https://github.com/neovim/neovim/discussions/27940 ?

I believe I could implement what I need with this feature, yeah.

lewis6991 (changes requested):
This can quickly fall out of sync with the highlighter. Would it be possible to pull out common code?

I believe `:Inspect` and `:Tohtml` also contain similar code duplication.

Also, I find this code very difficult to read. Can you please add linebreaks at appropriate places so different blocks can be easily distinguished.

runtime/lua/vim/treesitter/languagetree.lua
```diff
@@ -1184,4 +1184,104 @@ function LanguageTree:language_for_range(range)
...
+      or (
+        a.range[1] == b.range[1]
+        and a.range[2] == b.range[2]
+        and a.type == 'start'
+        and b.type == 'end'
+      )
+  end)
+
+  local result = {}
+  ---@type  {group:string,range:integer[],priority:integer,type:'start'|'end',start?:table,subpriority:integer}[]
```
lewis6991, line 1236: Since this has 6 fields, it would be much more readable as a `@class`.

`binary_find` can also be pulled out as a put function with `stack` as an argument.

runtime/lua/vim/treesitter/languagetree.lua
```diff
@@ -1184,4 +1184,104 @@ function LanguageTree:language_for_range(range)
...
+    if not hl_query then
+      return
+    end
+
+    for id, node, metadata in hl_query:iter_captures(tree:root(), source, 0, -1) do
+      local name = hl_query.captures[id]
+      local start_row, start_col, end_row, end_col = node:range()
+      local priority = tonumber(metadata.priority or vim.highlight.priorities.treesitter)
+
+      table.insert(positions, {
```
lewis6991, line 1206: Is there a reason we need to inject two elements into `positions` instead of just providing the full range?

altermo: That's how the algorithm works.

The algorithm resolves node priorities so that no node's range overlaps with another.

More about the algorithm:
A list of positions is prepared where each element is either a node's beginning or end position.
The positions are sorted by position (this separates the beginning and end positions of the same node).
Loop over the positions:
When encountering a beginning position its associated node gets appended to a priority sorted list (higher priority positions always come before lower priority positions).
When encountering an ending position its associated node gets removed from the priority sorted list.
On every cycle, if there is distance between the positions, the first node (e.g. the node with the highest priority) on the priority sorted list gets appended to the highlights list with its range set to the range between the previous position in the loop and the current position in the loop.

lewis6991: These level of explanations work really well as code documentation.

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -4,12 +4,12 @@ local Range = require('vim.treesitter._range')
 
 local ns = api.nvim_create_namespace('treesitter/highlighter')
 
----@alias vim.treesitter.highlighter.Iter fun(end_line: integer|nil): integer, TSNode, vim.treesitter.query.TSMetadata, TSQueryMatch
+---@alias vim.treesitter.highlighter.Iter fun(end_line:integer):{range:Range6, hl_group:string, priority:integer, conceal:string?, spell:boolean?, url:string?}?,TSNode?
```
lewis6991, line 7: Can you make the first return type a `@class`?

runtime/lua/vim/treesitter/languagetree.lua
```diff
@@ -1184,4 +1184,140 @@ function LanguageTree:language_for_range(range)
...
+---@field range integer[]
+---@field priority integer
+---@field type 'start'|'end'
+---@field start? table
+---@field subpriority integer
+---@field info table
+
+---@param stack vim.treesitter.LanguageTree.highlight_resolve_priority.obj[]
+---@param subpriority number
+---@return number
```
lewis6991, line 1221: Shouldn't this be(?):
```suggestion
---@return integer
```

runtime/doc/treesitter.txt
```diff
@@ -734,6 +734,33 @@ get_captures_at_pos({bufnr}, {row}, {col})
...
+    Get highlight information from parser.
+
+    Parameters: ~
+      • {parser}  (`vim.treesitter.LanguageTree`)
+
+    Return: ~
+        (`vim.treesitter.highlighter.Iter.ret[]`)
+
+                             *vim.treesitter.get_highlight_resolve_priority()*
+get_highlight_resolve_priority({parser})
```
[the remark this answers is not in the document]
altermo: That function is for when you want to apply treesitter highlighting on non-buffer texts.
These include foldtext, notifications, and cmdline(using `g:Nvim_color_cmdline`).

But if it's not desired then I won't include this function.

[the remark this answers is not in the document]
lewis6991: Should it be maybe named `get_highlight_virtual_text`?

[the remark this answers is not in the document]
altermo: What about `get_highlight_remove_overlap`?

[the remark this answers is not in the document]
luukvbaal: Or `get_highlight_no_overlap`?

> These include foldtext, notifications, and cmdline(using `g:Nvim_color_cmdline`).

(If all goes according to plan with replacing the message grid, the only remaining of those will be foldtext.)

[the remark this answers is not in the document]
altermo: If that's happening then the function becomes basically useless, so I won't include it.

justinmk (approved):

runtime/lua/vim/treesitter.lua
```diff
@@ -466,4 +466,27 @@ function M.foldexpr(lnum)
...
+    local source = ltree:source()
+    local hl_query = M.query.get(ltree:lang(), 'highlights')
+
+    if not hl_query then
+      return
+    end
+
+    local highlighter = require('vim.treesitter.highlighter')
+    for info in highlighter._iter_highlight_info(hl_query, tree:root(), source, 0, -1) do
+      info.range = { M._range.unpack4(info.range) }
```
wookayin, line 485: `range` declared as Range6 but the content is Range4. Also it would be nice to add some tests for the fields in `vim.treesitter.highlighter.Iter.ret`.

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -4,12 +4,30 @@ local Range = require('vim.treesitter._range')
 
 local ns = api.nvim_create_namespace('treesitter/highlighter')
 
----@alias vim.treesitter.highlighter.Iter fun(end_line: integer|nil): integer, TSNode, vim.treesitter.query.TSMetadata, TSQueryMatch
+--- @class vim.treesitter.highlighter.Iter.ret
```
wookayin, line 7: Not a blocker, but given this is a structure for the public API `|vim.treesitter.get_highlight()|` it would deserve a better class name rather than internal class name, say `vim.treesitter.highlighter.Info` or `vim.treesitter.HighlightInfo`, etc. Though I don't think these two are good enough.

runtime/lua/vim/treesitter.lua
```diff
@@ -466,4 +466,26 @@ function M.foldexpr(lnum)
...
+  parser:for_each_tree(function(tree, ltree)
+    local source = ltree:source()
+    local hl_query = M.query.get(ltree:lang(), 'highlights')
+
+    if not hl_query then
+      return
+    end
+
+    local highlighter = require('vim.treesitter.highlighter')
+    for info in highlighter._iter_highlight_info(hl_query, tree:root(), source, 0, -1) do
```
clason, line 484:
```suggestion
    for info in M.highlighter._iter_highlight_info(hl_query, tree:root(), source, 0, -1) do
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -244,11 +261,11 @@ function TSHighlighter:get_query(lang)
 end
 
 --- @param match TSQueryMatch
---- @param bufnr integer
+--- @param source integer|string
```
justinmk, line 264:
```suggestion
--- @param source integer|string Buffer name or number
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -4,12 +4,30 @@ local Range = require('vim.treesitter._range')
...
 local ns = api.nvim_create_namespace('treesitter/highlighter')
 
----@alias vim.treesitter.highlighter.Iter fun(end_line: integer|nil): integer, TSNode, vim.treesitter.query.TSMetadata, TSQueryMatch
+--- @class vim.treesitter.highlighter.Info
+--- @inlinedoc
+--- The range of the highlight
+--- @field range Range6
+--- The group of the highlight
+--- @field hl_group string
+--- The priority of the highlight
```
justinmk, line 13: are there relevant docs we can link to, for some of these fields? the current descriptions are almost useless

```suggestion
--- Highlight priority |treesitter-highlight-priority|
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -4,12 +4,30 @@ local Range = require('vim.treesitter._range')
 
 local ns = api.nvim_create_namespace('treesitter/highlighter')
 
----@alias vim.treesitter.highlighter.Iter fun(end_line: integer|nil): integer, TSNode, vim.treesitter.query.TSMetadata, TSQueryMatch
+--- @class vim.treesitter.highlighter.Info
+--- @inlinedoc
+--- The range of the highlight
+--- @field range Range6
+--- The group of the highlight
```
justinmk, line 11:
```suggestion
--- Highlight |group-name|
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -4,12 +4,30 @@ local Range = require('vim.treesitter._range')
...
+--- @field hl_group string
+--- The priority of the highlight
+--- @field priority integer
+--- The conceal of the highlight, nil for no conceal
+--- @field conceal string?
+--- Whether the highlight is spell or not
+---   - If `nil` then the highlight doesn't have a spell set
+---   - If `true` then the highlight is spell
+---   - If `false` then the highlight is nospell
```
justinmk, line 20: is this referring to 'nospell' option or something else (extmark-specific, or syntax specific...?)

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -4,12 +4,30 @@ local Range = require('vim.treesitter._range')
...
+--- @field spell boolean?
+--- The url of the highlight, nil for no url
```
justinmk, line 22:
```suggestion
--- Extmark URL, see "url" option at |nvim_buf_set_extmark()|
```

runtime/lua/vim/treesitter/highlighter.lua
clason: Note that `@spell` (and `@conceal`) is not technically a highlight. I would leave it out for now.

justinmk: > Note that `@spell` (and `@conceal`) is not technically a highlight. I would leave it out for now.

Is there some other way to get that info? Maybe we should rename the function if "highlight" is objectionable. What matters is that it gets useful info, the technical distinction seems minor?

clason: Well, the question then becomes (again) "what is this function (meant to be) used for?"

lewis6991: Could we consider making this internal only for now in order to progress this easier?

altermo: Sure

PR #28416, opened 2024-04-19
refactor: deprecate vim.region()

justinmk:
`vim.region()` is redundant with `getregionpos()`.

### DIFF
--- a/runtime/doc/deprecated.txt
+++ b/runtime/doc/deprecated.txt
@@ -19,6 +19,8 @@ API
 - nvim_subscribe()	Plugins must maintain their own "multicast" channels list.
 - nvim_unsubscribe()	Plugins must maintain their own "multicast" channels list.
 
+LUA
+- vim.region()		Use |getregionpos()| instead.
 
 ------------------------------------------------------------------------------
 DEPRECATED IN 0.10					*deprecated-0.10*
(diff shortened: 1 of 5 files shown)


dundargoc:
Should we remove the `vim.region` tests?

clason:
@zeertzjq

zeertzjq:
`vim.region()` is also used for adding highlight

justinmk:
This is very annoying. `getregion()` was added but it only returns text, without any way to map that to a region of the source buffer.

Shougo:
`getregionpos()` is added.

runtime/lua/vim/highlight.lua
```diff
@@ -40,22 +40,36 @@ M.priorities = {
...
   opts = opts or {}
   local regtype = opts.regtype or 'v'
-  local inclusive = opts.inclusive or false
+  local exclusive = opts.inclusive and 0 or 1
   local priority = opts.priority or M.priorities.user
   local scoped = opts._scoped or false
 
-  -- TODO: in case of 'v', 'V' (not block), this should calculate equivalent
-  -- bounds (row, col, end_row, end_col) as multiline regions are natively
-  -- supported now
```
zeertzjq, line 54: Why is this TODO removed? The extmarks still don't use multiline range.

PR #28417, opened 2024-04-19
perf(treesitter): incremental foldupdate

tomtomjhj:
Problem:
While the fold level computation is incremental, the evaluation of the foldexpr is done on the full buffer. Despite that the foldexpr reads from the cache, it can take tens of milliseconds for moderately big (10K lines) buffers.

Solution:
Track the range of lines on which the foldexpr should be evaluated.

### DIFF
--- a/runtime/lua/vim/treesitter/_fold.lua
+++ b/runtime/lua/vim/treesitter/_fold.lua
@@ -4,10 +4,21 @@ local Range = require('vim.treesitter._range')
 
 local api = vim.api
 
+---Treesitter folding is done in two steps:
+---(1) compute the fold levels with the syntax tree and cache the result (`compute_folds_levels`)
+---(2) evaluate foldexpr for each window, which reads from the cache (`foldupdate`)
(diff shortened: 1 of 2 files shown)


lewis6991 (approved):

justinmk:
coverity warning:

```
*** CID 492676:  Null pointer dereferences  (NULL_RETURNS)
/src/nvim/lua/stdlib.c: 564 in nlua_foldupdate()
558         return 0;
559       }
560     
561       linenr_T start = (linenr_T)luaL_checkinteger(lstate, 2);
562       linenr_T end = (linenr_T)luaL_checkinteger(lstate, 3);
563     
>>>     CID 492676:  Null pointer dereferences  (NULL_RETURNS)
>>>     Dereferencing a pointer that might be "NULL" "win" when calling "foldUpdate".
564       foldUpdate(win, start + 1, end);
565     
566       return 0;
567     }
568     
569     // Access to internal functions. For use in runtime/
```

Issue #859, opened 2014-06-17
string handling, SDS / bstrlib / ...


aktau:
Hey all, so one of the ongoing cleanup efforts in the neovim codebase, has been string handling. Recently, I proposed replacing `strncpy` and its variants (`STRNCPY` and `vim_strncpy`) with `(x)strlcpy` (#743, et al.) , because it's safer. *

While this didn't go entirely to plan (issue #858) because of slight differences in behaviour **, on the whole I think this is a positive evolution. What's clear is that these changes need to be a bit more carefully planned and reviewed, no one person can hope to catch all the subtleties.

That said, there's room for improvement on the string front in other ways as well:
1. We should create a wiki page detailing the best practices when dealing with strings. What functions do we use when? What are common idioms? (e.g.: `stpcpy` for repeat concatenating, ...)
2. Are there any new functions that we could introduce? In a few others issues I've mentioned this. For some cases neither `strncpy + NUL-termination` nor `strlcpy` nor `stp(n)cpy` are optimal. I'll describe below what I can think of.
## Example situations

Sometimes it's easiest to come up with better approaches by handling a few real life situations. Let's look at something @dougschneider [bumped into recently](https://github.com/dougschneider/neovim/commit/ded825dae7d04c0dda6b6917c62d8835742997ea#commitcomment-6681351).

The original code:

``` c
   STRCPY(IObuff, ":return ");
   STRNCPY(IObuff + 8, s, IOSIZE - 8);
    if (STRLEN(s) + 8 >= IOSIZE)
      STRCPY(IObuff + IOSIZE - 4, "...");
```

The rework as proposed [here](https://github.com/dougschneider/neovim/commit/955122a7cab6c716c402982d834d448e43bf37a3). There might be a small bug lurking in the fact that `sizeof(ret)` is 9 and thus `IObuff + sizeof(ret)` would be after the copied NUL byte.

``` c
  const char ret[] = ":return ";
  STRCPY(IObuff, ret);
  size_t slen = STRLCPY(IObuff + sizeof(ret), s, IOSIZE - sizeof(ret));
  if (slen + sizeof(ret) >= IOSIZE)
      STRCPY(IObuff + IOSIZE - 4, "...");
```

A few alternatives:

``` c
// compact and correct, but possibly a bit slower because of the first strlcpy call
size_t startlen = STRLCPY(IObuff,  ":return ", IOSIZE); 
[...]

PR #28422, opened 2024-04-20
refactor: add xmemcpyz() and use it in place of some xstrlcpy()

zeertzjq:
Problem:  Using xstrlcpy() when the exact length of the string to be
          copied is known is not ideal because it requires adding 1 to
          the length and an unnecessary strlen().
Solution: Add xmemcpyz() and use it in place of such xstrlcpy() calls.

Related #859

### DIFF
--- a/src/nvim/memory.c
+++ b/src/nvim/memory.c
@@ -202,7 +202,7 @@ void *xmallocz(size_t size)
   }
 
   void *ret = xmalloc(total_size);
-  ((char *)ret)[size] = 0;
+  ((char *)ret)[size] = '\0';
 
   return ret;
 }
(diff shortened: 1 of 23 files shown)


src/nvim/memory.c
```diff
@@ -222,6 +222,21 @@ void *xmemdupz(const void *data, size_t len)
   return memcpy(xmallocz(len), data, len);
 }
 
+/// Duplicates `len` bytes of `src` to `dst` and zero terminates it.
+/// and returns a pointer to the allocated memory. If the allocation fails,
+/// the program dies.
```
justinmk, line 227: Is this actually true? I see that this docstring was copied from `xmemdupz`, which calls `xmallocz` which will terminate the program. But this new function does not call anything that would kill if allocation fails.

zeertzjq: Oops, forgot to delete these two lines.

justinmk:
coverity warning:

```
*** CID 492655:  Memory - corruptions  (OVERRUN)
/src/nvim/cmdexpand.c: 3050 in expand_shellcmd()
3044         }
3045     
3046         size_t l = (size_t)(e - s);
3047         if (l > MAXPATHL - 5) {
3048           break;
3049         }
>>>     CID 492655:  Memory - corruptions  (OVERRUN)
>>>     Overrunning buffer pointed to by "s" of 1 bytes by passing it to a function which accesses it at byte offset 4090 using argument "l" (which evaluates to 4091).
3050         expand_shellcmd_onedir(buf, s, l, pat, matches, numMatches, flags, &found_ht, &ga);
3051         if (*e != NUL) {
3052           e++;
3053         }
3054       }
3055       *matches = ga.ga_data;
```

zeertzjq:
It's already fixed.

PR #28425, opened 2024-04-20
fix(man.vim): q exits after jumping to another manpage

### LINKED ISSUE #25943: man.vim: q does not exit, throws E444 instead, if you jump to another manpage
### Problem

As was suggested in #18443:

> Then the ftplugin can check `b:pager` and use `<C-W>c` or `<C-W>q`, respectively.

_Originally posted by @gpanders in https://github.com/neovim/neovim/issues/18443#issuecomment-1119938663_

Ftplugin now checks `b:pager`, but If you open another page inside "pager", mapping will be overridden, just like when you open a man page in a standard nvim instance.

My suggestion is to make `b:pager` a global variable. Then opening nvim with +Man! will globally define the behavior of that instance as a pager, not just for the first buffer.

### Steps to reproduce

1. set `MANPAGER` to `nvim +Man!`
2. `man git`
3. place the cursor to `gittutorial(7)`
4. press `<C-]>` to jump to a tag
5. now try to exit with `q`

### Expected behavior

nvim closed with `q`

### Neovim version (nvim -v)

NVIM v0.10.0-dev-a8fc94a

idk

### Operating system/version

NixOS unstable `ca012a02bf8327be9e488546faecae5e05d7d749`

### Terminal name/version

foot version: 1.16.1 +pgo +ime +graphemes -assertions +ucurl

foot

### Installation

Nixpkgs

RandomChugokujin:
Problem: When Neovim is set as the MANPAGER, pressing q should cause it to exit. However, after jumping to another tag in the manpager with `<C-]>`, E444 is thrown and Neovim does not exit.

Solution: Add `b:pager` as a global variable in man.vim


zeertzjq:
I don't think this makes sense. If Nvim is not invoked as MANPAGER, `b:pager` should not be set.

RandomChugokujin:
> I don't think this makes sense. If Nvim is not invoked as MANPAGER, `b:pager` should not be set.

Pretty sure if you invoke Nvim as MANPAGER and jumps to another manpage, Nvim does not consider that buffer to be one invoked as MANPAGER, and therefore pressing q will not cause that buffer to be closed.

gpanders:
The issue described in the PR description should be fixed, but this is not that fix. A user can use the `:Man` command in an already running Nvim instance, and in that case `b:pager` should not be set.

RandomChugokujin:
Ok, I see what you guys are saying. Would it be accept to change `b:pager` to `g:pager`?

RandomChugokujin:
I was finally able to find where `b:pager` was originally defined, now it is `g:pager`

PR #28427, opened 2024-04-20
test: improve test conventions

### LINKED ISSUE #27004: tests: improve test conventions (SUT, helpers.lua)

dundargoc:
Specifically, functions that are run in the context of the test runner
are put in module `test/testutil.lua` while the functions that are run
in the context of the test session are put in
`test/functional/testnvim.lua`.

Closes https://github.com/neovim/neovim/issues/27004.


dundargoc:
Todo: write in test readme wtf is happening

test/functional/lua/vim_spec.lua
```diff
@@ -1817,7 +1817,7 @@ describe('lua stdlib', function()
       local function get_counter() return counter end
       vim.t.AddCounter = add_counter
       vim.t.GetCounter = get_counter
-      vim.api.fn = {add = add_counter, get = get_counter}
+      vim.t.fn = {add = add_counter, get = get_counter}
```
justinmk, line 1820: ?

zeertzjq: What's the problem?

justinmk: I thought this was maybe an accidental search-replace. But now I see this might be *fixing* an accidental replace.

test/benchmark/preload.lua
```diff
@@ -1,4 +1,4 @@
 -- Modules loaded here will not be cleared and reloaded by Busted.
 -- Busted started doing this to help provide more isolation.  See issue #62
 -- for more information about this.
-local t = require('test.functional.testutil')
+local t = require('test.functional.testutil')()
```
dundargoc, line 4: Todo: sus

test/README.md
```diff
@@ -119,7 +122,7 @@ Debugging tests
   If `$VALGRIND` is also set it will pass `--vgdb=yes` to valgrind instead of
   starting gdbserver directly.
 
-  See [test/functional/testutil.lua](https://github.com/neovim/neovim/blob/9cadbf1d36b63f53f0de48c8c5ff6c752ff05d70/test/functional/testutil.lua#L52-L69) for details.
+  See [test/functional/nvimsession.lua](https://github.com/neovim/neovim/blob/9cadbf1d36b63f53f0de48c8c5ff6c752ff05d70/test/functional/nvimsession.lua#L52-L69) for details.
```
zeertzjq, line 125: This file doesn't not exist on this commit.

zeertzjq: And there is another instance in `runtime/doc/dev_tools.txt`

zeertzjq:
```diff
diff --git a/test/functional/provider/define_spec.lua b/test/functional/provider/define_spec.lua
index fc1b05e5b..49887cf72 100644
--- a/test/functional/provider/define_spec.lua
+++ b/test/functional/provider/define_spec.lua
@@ -30,7 +30,7 @@ local function runx(sync, handler, on_setup)
   local function setup_cb(...)
     on_setup(...)
     -- need to stop on setup callback because there's two session:request
-    -- calls in `request/testutil.lua`. The second call will always return
+    -- calls in `request/nvimsesion.lua`. The second call will always return
     -- after pending notification/request callbacks are processed
     stop()
   end
diff --git a/test/functional/ui/embed_spec.lua b/test/functional/ui/embed_spec.lua
index da5e21642..949165474 100644
--- a/test/functional/ui/embed_spec.lua
+++ b/test/functional/ui/embed_spec.lua
@@ -233,7 +233,7 @@ describe('--embed UI', function()
     }
     eq({ [16777215] = true }, seen)
 
-    -- NB: by accident how functional/testutil.lua currently handles the default color scheme, the
+    -- NB: by accident how functional/nvimsession.lua currently handles default color scheme, the
     -- above is sufficient to test the behavior. But in case that workaround is removed, we need
     -- a test with an explicit override like below, so do it to remain safe.
     startup('--cmd', 'hi NORMAL guibg=#FF00FF')
```

test/benchmark/preload.lua
```diff
@@ -1,4 +1,4 @@
 -- Modules loaded here will not be cleared and reloaded by Busted.
 -- Busted started doing this to help provide more isolation.  See issue #62
 -- for more information about this.
-local t = require('test.functional.testutil')
+local n = require('test.functional.nvimsession')()
```
zeertzjq, line 4: Calling the returned function here doesn't do anything.
```suggestion
local n = require('test.functional.nvimsession')
```

zeertzjq:
`skip()` and `skip_fragile()` should be in `test/testutil.lua`

test/README.md
dundargoc: I've removed them.

runtime/doc/dev_tools.txt
```diff
@@ -88,8 +88,6 @@ Then, in another terminal:
     gdb build/bin/nvim
     target remote localhost:7777
 <
-- See also test/functional/testutil.lua https://github.com/neovim/neovim/blob/3098b18a2b63a841351f6d5e3697cb69db3035ef/test/functional/testutil.lua#L38-L44.
```
justinmk, line 91: the URL was broken but we could fix it: https://github.com/neovim/neovim/blob/a4fc3bb0e68c8b078377fd9826e4cca3b4b3fdbf/test/functional/testutil.lua#L29-L51

or mention "see nvim_argv in ..."

test/README.md
```diff
@@ -37,18 +37,21 @@ Layout
...
   variables in lua tests.
 - `/test/includes` : include-files for use by luajit `ffi.cdef` C definitions
   parser: normally used to make macros not accessible via this mechanism
   accessible the other way.
 - `/test/*/preload.lua` : modules preloaded by busted `--helper` option
-- `/test/**/testutil.lua` : common utility functions for test code
+- `/test/**/testutil.lua` : common utility functions in the context of the test
+  runner
+- `/test/**/nvimsession.lua` : common utility functions in the context of the
+  test session
```
justinmk, line 51:
```suggestion
  test session (RPC channel to the Nvim child process created by clear() for each test)
```

runtime/doc/dev_tools.txt
dundargoc: We haven't merged this PR yet, how could we refer to a commit hash that hasn't been created?

test/functional/script/luacats_parser_spec.lua
```diff
@@ -1,4 +1,5 @@
-local t = require('test.functional.testutil')()
+local t = require('test.testutil')
```
justinmk, line 1: Since we no longer merge the modules, this makes it obvious that this test doesn't use a Nvim session.

test/old/testdir/load.vim
```diff
@@ -1,4 +1,4 @@
-" Also used by: test/functional/testutil.lua
+" Also used by: test/functional/nvimsessions.lua
```
justinmk, line 1: typo

runtime/doc/dev_tools.txt
dundargoc: I'll go ahead with the "see nvim_argv in ..." option

justinmk (approved):
This is great. Should help contributors quite a bit.

Future improvements to really help discoverability: 

- ✅  (already done?) See if we can convince LuaLS to show a docstring when LSP hover doc (K) is used on the `n` variable. 
- Other "Future" ideas in https://github.com/neovim/neovim/issues/27004

dundargoc:
> ✅ (already done?) See if we can convince LuaLS to show a docstring when LSP hover doc (K) is used on the n variable.

Technically I guess? When you press `K` the description is hidden since it's written last and there are so many functions in the module.

PR #28429, opened 2024-04-20
vim-patch:9.1.0357: Page scrolling should place cursor at window boundaries

luukvbaal:
Problem:  Page scrolling does not always place the cursor at the top or
          bottom of the window (Mathias Rav)
Solution: Place the cursor at the top or bottom of the window.
          (Luuk van Baal)

https://github.com/vim/vim/commit/4b6b0c4024df08dd8ce49dff3c76356ff81190c4

### DIFF
--- a/test/functional/ui/decorations_spec.lua
+++ b/test/functional/ui/decorations_spec.lua
@@ -4268,7 +4268,7 @@ if (h->n_buckets < new_n_buckets) { // expand
     screen:expect{grid=[[
       {16:refactor(khash): }take size of values as parameter |
       Author: Dev Devsson, {18:Tue Aug 31 10:13:37 2021}     |
-      ^if (h->n_buckets < new_n_buckets) { // expand     |
+      if (h->n_buckets < new_n_buckets) { // expand     |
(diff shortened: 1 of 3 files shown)


test/functional/ui/decorations_spec.lua
```diff
@@ -4268,15 +4268,15 @@ if (h->n_buckets < new_n_buckets) { // expand
...
+      if (h->n_buckets < new_n_buckets) { // expand     |
         khkey_t *new_keys = (khkey_t *)krealloc((void *)|
       h->keys, new_n_buckets * sizeof(khkey_t));        |
         h->keys = new_keys;                             |
         if (kh_is_map && val_size) {                    |
           char *new_vals = krealloc( h->vals_buf, new_n_|
       buckets * val_size);                              |
           h->vals_buf = new_vals;                       |
-        }                                               |
+        ^}                                               |
```
zeertzjq, line 4279: This seems to be a problem, as this suggests that the new behavior is different from the old behavior present since 2021. It may need more inspection to find what the old behavior comes from. Ref https://github.com/vim/vim/pull/14586#issuecomment-2067801396

Issue #2204, opened 2015-03-20
Number keys 


unphased:
Sorry if this has been addressed in docs, but I searched and couldn't find anything. 

in my vimrc I have 

```
set <k0>=^[Op
set <k1>=^[Oq
set <k2>=^[Or
set <k3>=^[Os
set <k4>=^[Ot
set <k5>=^[Ou
set <k6>=^[Ov
set <k7>=^[Ow
set <k8>=^[Ox
set <k9>=^[Oy
```

This is mainly being used with vim-mark now (which already has binds for the numpad `<k0-9>` keys). 

However I am finding that in neovim, the `set <k0>=` isn't actually successfully setting a keycode (`E846: Key code not set: <k0>`) like it does in Vim. 

It also appears that the numpad numbers are being interpreted the same as the actual numbers, so it adds to the ephemeral number count buffer. Which is fine, but not the behavior I am looking for.

So I am wondering what the proper way to do this is. Has setting keycodes with `set` been abolished?

justinmk:
> `set <k0>=` isn't actually successfully setting a keycode 

Did you try setting it to something other than a literal byte sequence? `^[Op` is (presumably) a terminal control sequence, this probably won't work in Neovim, and there's no point to it.

unphased:
What would I set it to? I don't know how else to have it actually function in vanilla Vim.

unphased:
Can you tell me what you mean by `set <k0>=foo`? What would such a thing accomplish? That is not a sane terminal escape code. I tried it, and afterwards neovim still gives E846 when I ask for what `<k0>` is set to.

justinmk:
> foo is not a sane terminal escape code

Right. There is no reason to set anything to a terminal escape code in neovim.

unphased:
Okay, so what's the verdict? neovim does not support arbitrary terminal codes?

Where does it fetch the escape codes to interpret then? terminfo?

justinmk:
What happens if you do this:

```
:map <k0> :echo 'hi'<cr>
:call feedkeys("\<k0>")
```

unphased:
It printed `hi`, just like it should. That doesn't have any bearing on me being able to drive `<k0>` with my computer keyboard (or remote computer keyboard over SSH through tmux, which is how I really use it most of the time, but which also isn't relevant)

justinmk:
Does `0` on your numpad echo "hi" or not?

unphased:
No, it continues to function identically to the 0 key.

[...]

PR #28431, opened 2024-04-20
Recognize <Find> and <Select> key codes

grawity:
These are the `ESC[1~` and `ESC[4~` keys that PuTTY sends for Home/End; they don't match what the 'xterm' terminfo has, so libtermkeys reports them as the original VT220 names.

I was hesitant to just permanently map them to K_HOME/K_END here, but having Neovim recognize them as their own keys will at the very least allow users to `:imap <Find> <Home>` or something along those lines.

(Of course, configuring PuTTY to set TERM=putty[-256color] would *in theory* be the preferred solution, but in practice it causes other problems due to other features missing from the 'putty' terminfo, as well as older systems that lack the terminfo entirely. Having Neovim recognize the keys is the least bad workaround.)


justinmk:
https://github.com/neovim/neovim/issues/2204 would be an alternative I guess, then you could `:set <F32>=^[...` and map to `<F32>` instead of e.g. `<Find>`. However the [latest comment](https://github.com/neovim/neovim/issues/2204#issuecomment-1890988865) there points out that this is not great for maintaining a config.

> I was hesitant to just permanently map them to K_HOME/K_END here

Is that because of the risk of "false positives"?

I can't think of any reason _not_ add more TERMCAP2KEY mappings, except that the `<Find>` and `<Select>` keycodes are a bit random, and Vim doesn't have them. Do you have a reference for the "DEC keys"?

Reference: last time we added keycodes: https://github.com/neovim/neovim/commit/d13803f64fc5607c6319087240e35a8b86082f64

If others don't have objections, this would need docs and a [test](https://github.com/neovim/neovim/blob/5e6240ffc24e55ecf7721fd9dc3f33c6f178be8c/test/functional/ui/input_spec.lua#L16).

src/nvim/keycodes.h
```diff
@@ -373,6 +373,9 @@ enum key_extra {
 #define K_KPAGEDOWN     TERMCAP2KEY('K', '5')   // keypad pagedown (lower R.)
 #define K_KORIGIN       TERMCAP2KEY('K', '2')   // keypad center
 
+#define K_KFIND         TERMCAP2KEY('@', '0')   // DEC keys ("Home" and "End"
+#define K_KSELECT       TERMCAP2KEY('*', '6')   // in PuTTY with TERM=xterm)
+
```
zeertzjq, line 378: Why are these placed in the middle of keypad keys? I think these should be placed around `K_HELP`. And why are they named `K_KFIND` and `K_KSELECT` instead of `K_FIND` and `K_SELECT`?

src/nvim/keycodes.c
```diff
@@ -283,6 +283,9 @@ static const struct key_name_entry {
   { K_KORIGIN,         "kOrigin" },
   { K_KORIGIN,         "KP5" },
 
+  { K_KFIND,           "Find" },        // "Home" and "End" in
+  { K_KSELECT,         "Select" },      // PuTTY with TERM=xterm
+
```
zeertzjq, line 288: ditto

grawity:
> except that the `<Find>` and `<Select>` keycodes are a bit random

I wasn't the one who made them up. Those are the existing key names that neovim *already* receives from libtermkey in this situation (as I've found out via strace between the two nvim processes – I literally saw the strings `<Find>` and `<Select>` being exchanged through the RPC).

But currently Neovim completely ignores them and doesn't allow mapping them to anything else; the patch just makes that possible.

> Do you have a reference for the "DEC keys"?

My assumption is that they're named after the DEC LK201 keyboard that the physical VT220 used. Looking at photos online, it had 'Find' as key 1 and 'Select' as key 4 in roughly the same area where we now have Ins/Home/Del/End, so it seems to fit the explanation. I didn't dig deeper into why libtermkey uses those names.

src/nvim/keycodes.h
grawity: > Why are these placed in the middle of keypad keys?

I placed them there because that's where K_HOME and K_END are also defined, even though they're not keypad keys either.

> And why are they named K_KFIND and K_KSELECT instead of K_FIND and K_SELECT?

`K_SELECT` is already in use (by mouse-related things it seems) and I could not decide what other name to choose. Renaming the existing K_SELECT seemed too invasive so I thought "I'll name it K_KSELECT until someone suggests better names".

zeertzjq: Oh, then I think `K_KFIND` and `K_SELECT` are fine.

justinmk:
@grawity thanks, that is great info to add to the commit message!

grawity:
> Last time we added keycodes: [d13803f](https://github.com/neovim/neovim/commit/d13803f64fc5607c6319087240e35a8b86082f64)

Okay, so looking at this, I assume I could add `case K_FIND: ... return VTERM_KEY_HOME` in terminal.c to permanently?

After thinking more about it, false positives are probably unlikely, as tmux also sends the same key sequences when translating Home/End (and PC keyboards do not have Find/Select keys anyway).

I moved the keycode definitions next to K_HELP and renamed K_KFIND to K_FIND, but still not sure what better name to use instead of K_KSELECT (as K_SELECT is already taken).

leonerd:
@grawity You're right - I picked the names to match what the original DEC keyboard and terminals would do there. A fair number of modern terminals specifically use "CSI H" and "CSI F" as encodings for the `<Home>` and `<End>` keys, rather than functions 1 and 4.
I'm not really familiar with how neovim uses libvterm or not, and therefore what the logic of the TERMCAP2KEY function is doing here, so I'm not sure how much further I could help here.

justinmk:
@grawity I missed this somehow. Please rebase then I'll merge it.

PR #25455, opened 2023-10-01
feat(ui): allow to get the highlight namespace (rebase of #24390)

bfredl:
closes #24390

PR #28432, opened 2024-04-20
refactor(api): nvim_win_xx_ns are EXPERIMENTAL

justinmk:
# Problem:

The `nvim_win_xx_ns` function family introduced in ba0370b1d718d473d0ef51c35d88b98ba220082b needs more bake-time. Currently it's narrowly defined for "window" scope only; but other scopes ("buffer") and features are likely in the future.

# Solution

Rename the API with double-underscore to mark it as EXPERIMENTAL.

# Open questions

- We could choose *either*:
    - "store scopes on namespaces", or
    - "store namespaces on scopes (w:/b:/…)"
    - "store namespaces as global-local option" (requires [some work](https://github.com/neovim/neovim/pull/25455#discussion_r1345851839))
- Refactor to this interface:
  ```
    nvim_ns_add(
      ns_id: number,
      opts: {
        win: number
      }
    ): { ok: boolean }

    nvim_ns_get(
      ns_id: number,
      opts: {
        win: number
      }
    ): { ids: number[] }

    nvim_ns_del(
      ns_id: number,
      opts: {
        win: number
      }
    ): { ok: boolean }
  ```

# Closed questions

- `nvim_get_hl_ns` was added in this cycle: https://github.com/neovim/neovim/blob/47dbda97d2f40729733b1c0d1d13d914065af23c/src/nvim/api/vim.c#L202-L203
    - is it still needed or could it be addressed by `nvim_ns_get` ?
        - There is much more to "unwind" (`nvim_set_hl_ns, nvim_set_hl_ns_fast, nvim_get_hl_ns, nvim_win_set_hl_ns`), so  can leave `nvim_get_hl_ns` for now.

# Reference

I read https://github.com/neovim/neovim/pull/27361#issuecomment-1956487066 : 

> As far as I can tell this PR by itself is mostly forward-compatible with this extended proposal. The only difference I think is if we want to "mark" each individual extmark as scoped, or as you say, mark the entire namespace as scoped and then need to opt-in to it in specific windows.
> 
> `nvim_ns_add_tabpage_scope` would be a convenient short-hand for marking each individual window in that tab. convenient, but can be done as an addon. buffer-scoped would be mostly relevant for highlight namespaces, which we can consider but is separate work from this (and needs design discussion on its own. I have some other ideas how to rework those)

- It sounds like it makes sense to model these functions similar to the `nvim_get_option_info2()` interface. In general, we should avoid adding new `nvim_win`/`nvim_buf`/etc functions, if the main difference between each of the functions is only what kind of scope (win/buf/etc) it has, because that creates 4x docs and functions for not much reason.
- Returning boolean is probably a mistake, it closes the door on returning any other info in the future. We should usually return Dict.

cc @altermo @bfredl


runtime/doc/develop.txt
```diff
@@ -430,10 +430,11 @@ Example: >
...
 
-If the function acts on an object then {noun} is the name of that object
-(e.g. "buf" or "win"). If the function operates in a "global" context then
-{noun} is usually omitted (but consider "namespacing" your global operations
-with a {noun} that groups functions under a common concept).
+Do not add new nvim_buf/nvim_win/nvim_tabpage APIs, unless you are certain the
+concept will NEVER be applied to more than one "scope". That is, {noun} should
+be the TOPIC ("ns", "extmark", "option", …) that acts on the scope(s)
+(buf/win/tabpage/global), it should NOT be the scope. The scope should be
+a parameter.
```
justinmk, line 437: Updated API guidance (see also below)

runtime/doc/develop.txt
```diff
@@ -443,6 +444,59 @@ with a {noun} that groups functions under a common concept).
...
+        handle: integer (buf/win/tabpage id)
+        scope: "global" | "win" | "buf" | "tabpage"
+      }
+    ): { ok: boolean }
+
+    nvim_ns_get(
+      ns_id: int,
+      filter: {
+        handle: integer (buf/win/tabpage id)
+        scope: "global" | "win" | "buf" | "tabpage"
```
luukvbaal, line 465: Where does this filter->handle/scope convention come from and is it really better than opts->buf/win/tabpage which is used all over the place already? I think omission of all three is unambiguous enough to mean global scope? (No strong opinion but mainly asking to know whether I should update nvim_redraw() with this pattern.)

luukvbaal: > Where does this `filter->handle/scope` convention

From `nvim_set_option_value()` I guess, like you mentioned. It might make sense for options but is scope` really necessary for most APIs that just need to target `win/buf/tabpage`?

justinmk: > It might make sense for options but is scope really necessary for most APIs that just need to target `win/buf/tabpage`

How else would they do it? Adding triple or quadruple functions (nvim_win/nvim_buf/nvim_tabpage) isn't really gaining anything, while adding a lot of noise.

luukvbaal: > How else would they do it?

`opts->win/buf/tabpage` as opposed to `opts->filter->global/win/buf/tabpage` which seems to be the proposal here. Wasn't suggesting separate functions but merely keeping with what I thought was an already established convention.

justinmk: Oh, yeah. Either way is fine. `opts->win/buf/tabpage` can be a bit awkward because of the "mutual exclusion". However, it's arguably more readable, less fussy, and pretty common already. So I'll update the doc.

runtime/doc/news.txt
```diff
@@ -183,7 +183,17 @@ The following new APIs and features were added.
...
+• API enhancements
+  • |nvim_buf_set_extmark()| supports inline virtual text.
+  • |nvim_win_text_height()| computes the number of screen lines occupied
+    by a range of text in a given window.
+  • New RPC client type `msgpack-rpc` is added for |nvim_set_client_info()| to
+    support fully MessagePack-RPC compliant clients.
+  • Floating windows can now be hidden by setting `hide` in |nvim_open_win()| or
+    |nvim_win_set_config()|.
+  • |nvim_input_mouse()| supports mouse buttons "x1" and "x2".
+  • Added |nvim_tabpage_set_win()| to set the current window of a tabpage.
```
zeertzjq, line 195:
```suggestion
  • |nvim_tabpage_set_win()| sets the current window of a tabpage.
```

PR #28433, opened 2024-04-20
refactor(api): deprecate nvim_call_atomic

justinmk:
Use `nvim_exec_lua()` instead


zeertzjq:
Some APIs are RPC-only (`FUNC_API_REMOTE_ONLY`), especially `nvim_ui_*` APIs, which cannot be used in `nvim_exec_lua`.

justinmk:
Good callout, thanks. But that isn't a blocker, since this only marks the API and won't break anything until Nvim 2.0. This will allow us to get the signal out and address any concerns.

For example, `nvim_buf_detach` is `FUNC_API_REMOTE_ONLY`, but why? We could just change that. Same for the `nvim_ui_*` functions.

Maybe we can change `FUNC_API_REMOTE_ONLY` to allow vimscript/lua to pass an explicit `channel_id`.

zeertzjq:
I think there is another problem that `nvim_exec_lua` has more overhead than `nvim_call_atomic`. `nvim_call_atomic` only involves a `RPC -> API` conversion, while `nvim_exec_lua` requires a `RPC -> API -> Lua -> API` conversion.

justinmk:
> `nvim_exec_lua` has more overhead than `nvim_call_atomic`

True, but we should focus on improving that generally instead of offering a half-baked workaround.

zeertzjq:
> > `nvim_exec_lua` has more overhead than `nvim_call_atomic`
> 
> True, but we should focus on improving that generally instead of offering a half-baked workaround.

But that may be infeasible, as the Lua C API is a fixed interface :(

justinmk:
Examples of general perf improvements:

- the work @bfredl did to avoid allocations
- cache RPC Lua scripts: https://github.com/neovim/neovim/issues/8092

PR #28434, opened 2024-04-21
fix(float): wrong position when bufpos is out of range

### LINKED ISSUE #23120: nvim_open_win behaves oddly when using bufpos outside the buffer text
### Problem

When `bufpos` is set to a value outside the text of the buffer, the window is positioned oddly.

### Steps to reproduce

Save the following lua script, then open a clean Neovim and source it.
```
vim.api.nvim_open_win(vim.api.nvim_create_buf(false, true), 0, {
  relative = "win",
  width = 10,
  height = 5,
  bufpos = { 3, 3 },
  row = 0,
  col = 0,
  style = "minimal",
  border = "single"
})
```

### Expected behavior

Either an error because `bufpos` is out of range, or the window to appear in the top-left corner. Instead, the window appears in the bottom-left corner.

### Neovim version (nvim -v)

v0.10.0-dev-75+gc2e47e7be

N/A

### Operating system/version

Rocky Linux 8.6

### Terminal name/version

xfce4-terminal 1.0.4

tmux-256color

### Installation

appimage

glepnir:
Problem: when lnum in bufpos is out of range the position of float is wired.

Solution: avoid the height value out of buffer line range.

Fix #23120

### DIFF
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -827,7 +827,8 @@ void ui_ext_win_position(win_T *wp, bool validate)
         row += row_off;
         col += col_off;
         if (c.bufpos.lnum >= 0) {
-          pos_T pos = { c.bufpos.lnum + 1, c.bufpos.col, 0 };
+          int lnum = MIN(c.bufpos.lnum + 1, win->w_buffer->b_ml.ml_line_count);
+          pos_T pos = { lnum, c.bufpos.col, 0 };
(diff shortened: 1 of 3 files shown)


bfredl:
For future reference why this is not an error: `buf_pos` is not (yet) synced to buffer changes, thus the same situation can occur by editing text after using `nvim_open_win` even with in-bounds values.

PR #28436, opened 2024-04-21
feat: namespace scoped autocmd

altermo:
Currently(as of 2024-04-21) the only namespace scope implemented is window scoped so this is only useful for setting window local autocmds. Later on, when more namespace scopes get implemented, this will get way more useful. For example, when buffer scoped namespace gets implemented, plugins could define all the autocmds once with a namespace scope and then apply set the namespace to the plugins buffer (rather than needing to create all the autocmds every time the plugin creates a new buffer).

### DIFF
--- a/runtime/lua/vim/_meta/api_keysets.lua
+++ b/runtime/lua/vim/_meta/api_keysets.lua
@@ -86,6 +86,7 @@ error('Cannot require a meta file')
 --- @field nested? boolean
 --- @field once? boolean
 --- @field pattern? any
+--- @field ns_id? integer
 
 --- @class vim.api.keyset.echo_opts
 --- @field verbose? boolean
--- a/src/nvim/api/autocmd.c
+++ b/src/nvim/api/autocmd.c
@@ -501,7 +501,8 @@ Integer nvim_create_autocmd(uint64_t channel_id, Object event, Dict(create_autoc
                                   opts->once,
                                   opts->nested,
                                   desc,
-                                  aucmd);
+                                  aucmd,
+                                  HAS_KEY(opts, create_autocmd, ns_id) ? (uint32_t)opts->ns_id : 0);
       });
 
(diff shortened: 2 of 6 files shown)


justinmk:
I think the first step here should be reworking the internals of autocmds to be backed by namespaces. Then "augroups" are simply namespaces. Are you planning to do that here?

altermo:
No, I'm not planning to do it here.

Though I'm fine with refactoring it in another PR. (give some response if you're okay with that)

justinmk:
> Though I'm fine with refactoring it in another PR. (Give some response if you're okay with that)

SGTM, but I see it as a blocker before we start adding other namespace-related features to autocmd.

> explain how augroups would work as namespaces (for example how would you list out all augroups (using `:augroup`)

namespaces have ids and names, just like "augroups". The proposed change is to remove the separation, i.e. augroups are now implemented as namespaces instead of their own special, separate address space.

Proposal:

- https://github.com/neovim/neovim/pull/28441#issuecomment-4453558348

ii14:
Not sure if I understand the use case for this. What makes it different from:
```lua
local win_set = {}

api.nvim_create_autocmd('...', {
  callback = function()
    if not win_set[api.nvim_get_current_win()] then
      return
    end
    -- ...
  end,
})

function add_win(win)
  if win == nil or win == 0 then
    win = api.nvim_get_current_win()
  end
  win_set[win] = true
end
```

altermo:
Nothing (well technically some autocmds may trigger for non-current window (e.g. `WinClosed`, `WinScrolled` and `WinResized`))

But the idea is that when more namespace scopes get added, this will become more useful. 

And if https://github.com/neovim/neovim/pull/28441 gets merged (which is a blocking PR), then augroups would be namespaces, and then it would be logical to apply namespace scopes to autocmds.

PR #28440, opened 2024-04-21
refactor(lua): rename tbl_islist, tbl_isarray

justinmk:
tbl_isarray was not released yet, so doesn't need a deprecation cycle.

tbl_islist will go through a deprecation cycle.

ref #24572


clason:
Will this apply to all `tbl_foo` functions? If not, what is the distinction?

justinmk:
> Will this apply to all `tbl_foo` functions? If not, what is the distinction?

The distinction for "is" functions is easy: they are not specific to tables, any arg is valid. So the tbl_ prefix is nonsense.

More generally, https://github.com/neovim/neovim/issues/24572 already discusses the plan and rationale for tbl_ functions.

clason:
> More generally, https://github.com/neovim/neovim/issues/24572 already discusses the plan and rationale for tbl_ functions.

That needs some updating after you changed the landscape, by the way.

PR #28441, opened 2024-04-21
refactor!: implement augroups as namespaces

altermo:
From https://github.com/neovim/neovim/pull/28436#issuecomment-2068036755:
> The proposed change is to remove the separation, i.e. augroups are now implemented as namespaces instead of their own special, separate address space.

##### Implementation details:

**Deleting augroups:**
*Background:*
Deleting named namespaces is not allowed, while there are 2 ways of deleting augroups (normal and legacy). Legacy deleted augroups change their id to `augroup_deleted` but do not change their name, which also is not allowed for named namespaces. augroups with the id `augroup_deleted` or which don't have a name are considered deleted, have the name `"--Deleted--"` and error when trying to access them (most of the time). There are some other important differences which will be noted later.
*Implementation:*
When an augroup gets deleted, its id is appended to the set `deleted_augroup_map`. Any augroup in the `deleted_augroup_map` or doesn't have an associated name (e.g. an anonymous namespace) is considered deleted. When legacy deleting augroups the same steps happen as with normal deletion but all the autocmds associated with the augroup change their group to the id returned by `get_deleted_augroup_id()`. This is the special deleted augroup and has the name `"--Deleted--"` and there is only one of these (This is a BREAKING-CHANGE, see *Example 1*).

**Creating augroups:**
*Implementation:*
Creating augroups will use `nvim_create_namespace()` to create augroups (This is a BREAKING-CHANGE as augroups now can't have the name `""`(empty string)). If the returned id is in `deleted_augroup_map` then it will be removed from there.

**Listing augroups:**
*Background:*
There are two (or more) ways of listing augroups: using `:augroup` (which I will refer to as normal listing) and using `getcompletion('','augroup')` (which I will refer to as completion listing). Whether a deleted augroup gets listed depends on what listing method you use (normal or completion) and whether the augroup was normal or legacy deleted. Completion listing lists all augroups, including normal deleted ones (all deleted/legacy deleted augroups have the name `"--Deleted--"`). Normal listing will only list legacy deleted augroups.
*Implementation:*
Listing augroups will also list all other named namespaces, as there is no difference between named namespaces and augroups (This might break some plugin tests). Normal deleted augroups will never be listed (not even in completion listings) and there will only be one `"--Deleted--"` listed after any number of augroups get legacy deleted (these are BREAKING-CHANGEs).

**Other:**
The first-ever namespaces created with `nvim_create_namespace()` after initializing the runtime files will now not be 1 as the runtime files create some augroups on initialization (This might break some plugin tests).

##### Examples (click to expand):
<details><summary>Example 1</summary>

```lua
local group=vim.api.nvim_create_augroup('test',{})
vim.api.nvim_create_autocmd('FileType',{group=group,command='',pattern='foo'})
vim.cmd'augroup! test'
assert(vim.api.nvim_get_autocmds({pattern='foo'})[1].group==group)
```

</details>

### DIFF
--- a/src/nvim/autocmd.c
+++ b/src/nvim/autocmd.c
@@ -8,6 +8,7 @@
 #include <string.h>
 
 #include "klib/kvec.h"
+#include "nvim/api/extmark.h"
 #include "nvim/api/private/helpers.h"
 #include "nvim/ascii_defs.h"
 #include "nvim/autocmd.h"
@@ -91,12 +92,11 @@ static const char e_autocommand_nesting_too_deep[]
 // Code for automatic commands.
 static AutoPatCmd *active_apc_list = NULL;  // stack of active autocommands
 
-// ID for associating autocmds created via nvim_create_autocmd
-// Used to delete autocmds from nvim_del_autocmd
-static int next_augroup_id = 1;
-
-// use get_deleted_augroup() to get this
-static const char *deleted_augroup = NULL;
+// The ID of the deleted group.
+// Zero until the group is initialized.
+static int deleted_augroup_id = 0;
+static const char *deleted_augroup_name = "--Deleted--";
+static Set(int) deleted_augroup_map = SET_INIT;
 
 // The ID of the current group.
 static int current_augroup = AUGROUP_DEFAULT;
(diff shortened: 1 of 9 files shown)


src/nvim/autocmd.c
```diff
@@ -407,7 +404,7 @@ int augroup_add(const char *name)
     augroup_map_del(existing_id, name);
   }
 
-  int next_id = next_augroup_id++;
+  int next_id = next_namespace_id++;
   String name_key = cstr_to_string(name);
   String name_val = cstr_to_string(name);
   map_put(String, int)(&map_augroup_name_to_id, name_key, next_id);
```
justinmk, line 410: probably want to replace all this with a call to `nvim_create_namespace`

src/nvim/autocmd.c
```diff
@@ -112,22 +113,11 @@ static bool autocmd_include_groups = false;
 static char *old_termresponse = NULL;
 
 // Map of autocmd group names and ids.
-//  name -> ID
 //  ID -> name
-static Map(String, int) map_augroup_name_to_id = MAP_INIT;
 static Map(int, String) map_augroup_id_to_name = MAP_INIT;
```
justinmk, line 117: I know this is WIP, but just mentioning: we should also be able to eliminate `map_augroup_id_to_name`.

justinmk:
Let's revisit this soon after releasing 0.10

justinmk:
Mind rebasing this? Is it ready for review or still WIP?

altermo:
It is ready for review

altermo:
After some thinking, I want to redesign how it works...

> <details><summary>Note to self</summary>
> Have a flag for each namespace, marking it as an augroup namespace (or also maybe a deleted augroup), this way we can possibly have maximum backwards compatibility....
> </details> 

> <details><summary>Note to self (2)</summary>
> Maybe a simpler solution would be to treat every namespace which has a linked augroup list as an augroup, so completion and other listings still work...
> </details>

justinmk:
Yes backwards compatibility is tricky here, because existing plugins might create conflicting names: `nvim_create_namespace('foo')` and `nvim_create_augroup('foo')`.

However I'm hoping that instead of a flag or a "flavor", we can do the following approach:

1. `:augroup foo` (`nvim_create_augroup('foo')`) automatically prepends a `ev.` prefix.
   Thus augroup `foo` occupies namespace `ev.foo`.
2. Throw an error if user code tries `nvim_create_namespace('ev.foo')`.
    - (If someone, somewhere was already using the `ev.` prefix for namespace, that's just too bad.)
3. Legacy things like `:au` and `nvim_get_autocmds()` can show both the "group' and "ns":
   ```
   group = 'foo'
   ns = 'ev.foo'
   ```
4. This allows us to move forward with a "single address space" instead of "flavors".
    - And we can deprecate the augroup parameters and nudge users towards namespaces.

PR #28453, opened 2024-04-22
fix(tui): disable sync output and extended underline check for Apple Terminal

niw:
**Problems**

When launching Neovim on Terminal.app on macOS (Apple Terminal,) it briefly shows like
`p$qm+q5463;524742;73657472676266;73657472676262$qm` in orange background color partially on the screen.

**Solution**

Since Apple Terminal seems not supporting synchronized output and extended underline, calling `tui_request_term_mode` and `tui_query_extended_underline` caused this unexpected output.

Therefore, if we know it's Apple Terminal (when `nsterm` is `true`,) do not call these checks.

Tested on Terminal.app (2.14, 453) on macOS 14.4.1.

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -410,11 +410,15 @@ static void terminfo_start(TUIData *tui)
 
   // Query support for mode 2026 (Synchronized Output). Some terminals also
   // support an older DCS sequence for synchronized output, but we will only use
-  // mode 2026
-  tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  // mode 2026.
+  // Some terminals (such as Terminal.app) do not support DECRQM, so skip the query.
+  if (!nsterm) {
+    tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  }
 
   // Don't use DECRQSS in screen or tmux, as they behave strangely when receiving it.
-  if (tui->unibi_ext.set_underline_style == -1 && !(screen || tmux)) {
+  // Terminal.app also doesn't support DECRQSS.
+  if (tui->unibi_ext.set_underline_style == -1 && !(screen || tmux || nsterm)) {
     // Query the terminal to see if it supports extended underline.
     tui_query_extended_underline(tui);
   }


niw:
Here is actual visual issue on Terminal.app without this patch appearing when Neovim launched.
Example

src/nvim/tui/tui.c
```diff
@@ -411,10 +411,12 @@ static void terminfo_start(TUIData *tui)
   // Query support for mode 2026 (Synchronized Output). Some terminals also
   // support an older DCS sequence for synchronized output, but we will only use
   // mode 2026
-  tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  if (!nsterm) {
+    tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  }
 
   // Don't use DECRQSS in screen or tmux, as they behave strangely when receiving it.
```
zeertzjq, line 418:
```suggestion
  // Don't use DECRQSS in screen or tmux, as they behave strangely when receiving it.
  // Apple Terminal also doesn't support extended underline.
```

niw:
It's strange that unit test on macOS 12.0 failed on the unrelated part of code
```
RUN       typval.c dict extend() works: 45.00 ms ERR
```
Likely... it's flaky test as like other tests skipped...?

gpanders:
What is the problem with `tui_request_term_mode`? The output you're seeing is only from the DECRQSS sequence, not DECRQM.

src/nvim/tui/tui.c
```diff
@@ -410,11 +410,15 @@ static void terminfo_start(TUIData *tui)
 
   // Query support for mode 2026 (Synchronized Output). Some terminals also
   // support an older DCS sequence for synchronized output, but we will only use
-  // mode 2026
-  tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  // mode 2026.
+  // Apple Terminal doesn't support it.
```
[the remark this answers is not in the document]
justinmk: Yes, in general we expect terminals not to barf if we request things that are unknown to them. But in some narrow cases, we may choose to "stop the bleeding" by special casing. But that should only be for egregious cases. This one looks fairly painful, and the fix appears to be cheap, so we can special case it.

niw:
> What is the problem with `tui_request_term_mode`? The output you're seeing is only from the DECRQSS sequence, not DECRQM.
> 
> Ah ok I just tried it in Apple Terminal myself, it indeed does not support DECRQM.

Yeah, the first `p` in the screen short I attached is coming from `tui_request_term_mode`.

src/nvim/tui/tui.c
```diff
@@ -410,11 +410,15 @@ static void terminfo_start(TUIData *tui)
...
-  tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  // mode 2026.
+  // Some terminals (such as Terminal.app) do not support DECRQM, so skip the query.
+  if (!nsterm) {
+    tui_request_term_mode(tui, kTermModeSynchronizedOutput);
+  }
 
   // Don't use DECRQSS in screen or tmux, as they behave strangely when receiving it.
-  if (tui->unibi_ext.set_underline_style == -1 && !(screen || tmux)) {
+  // Apple Terminal also doesn't support extended underline.
```
zeertzjq, line 420: Oops, it should say DECRQSS here
```suggestion
  // Terminal.app also doesn't support DECRQSS.
```

niw: It's better to say consistenly "Terminal.app" in the comment as well, thanks!

PR #28460, opened 2024-04-22
fix(treesitter): clip end row early 

tomtomjhj:
Also add some more argument checks.

Ref: https://github.com/neovim/neovim/pull/28417#issuecomment-2068319504
cc @justinmk

### DIFF
--- a/src/nvim/lua/stdlib.c
+++ b/src/nvim/lua/stdlib.c
@@ -549,19 +549,21 @@ static int nlua_iconv(lua_State *lstate)
 static int nlua_foldupdate(lua_State *lstate)
 {
   handle_T window = (handle_T)luaL_checkinteger(lstate, 1);
-  Error err = ERROR_INIT;
-  win_T *win = find_window_by_handle(window, &err);
-  if (ERROR_SET(&err)) {
-    nlua_push_errstr(lstate, err.msg);
-    api_clear_error(&err);
-    lua_error(lstate);
-    return 0;
+  win_T *win = handle_get_window(window);
+  if (!win) {
+    return luaL_error(lstate, "invalid window");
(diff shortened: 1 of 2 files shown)


src/nvim/lua/stdlib.c
```diff
@@ -549,19 +549,21 @@ static int nlua_iconv(lua_State *lstate)
...
-  Error err = ERROR_INIT;
-  win_T *win = find_window_by_handle(window, &err);
-  if (ERROR_SET(&err)) {
-    nlua_push_errstr(lstate, err.msg);
-    api_clear_error(&err);
-    lua_error(lstate);
-    return 0;
+  win_T *win = handle_get_window(window);
+  if (!win) {
+    return luaL_error(lstate, "invalid window");
```
justinmk, line 554: if it's a false positive (i.e. "can never happen"), assert() or abort() is fine too

tomtomjhj: `find_window_by_handle` in the old version uses `handle_get_window` and does null check internally, and sets `err` in that case. So `ERROR_SET(&err)` in the old version implies `win` is non-null. But adding assert to that is somewhat redundant, so I replaced those stuff to `handle_get_window` and direct null check.

tomtomjhj:
With the additional argument checks, I noticed a regression from #28417. cc @lewis6991

src/nvim/lua/stdlib.c
ii14: It's redundant either way, because that warning was a false positive.

tomtomjhj: Right, but I just wanted to reduce the noise from the warning

PR #28470, opened 2024-04-23
fix(lua): refactor `vim.deprecate()` for performance in "Nvim" plugin

### LINKED ISSUE #28459: `vim.deprecate()` can be significantly slower than deprecated function
### Problem

Adding `vim.deprecate()` to a deprecated function can lead to its relatively significant performance decrease. In particular, this is noticeable after #28440.

### Steps to reproduce

1. Create the following 'init.lua':
    ```lua
    _G.bench_tbl_islist = function()
      local durations, arg = {}, { 'a' }
      for _ = 1, 5 do
        local start_time = vim.loop.hrtime()
        vim.tbl_islist(arg)
        -- Insert duration in milliseconds
        table.insert(durations, 0.000001 * (vim.loop.hrtime() - start_time))
      end
      return durations
    end
    ```
2. `nvim --clean -u init.lua`
3. `:=bench_tbl_islist()`. It gives me the following results:
    -  On current Nightly (`NVIM v0.10.0-dev-2976+g208852126`) it is something like `{ 3.72243, 0.918169, 0.968143, 0.763256, 0.783424 }`.
    - On 0.9.5: `{ 0.002955, 0.000361, 0.000281, 0.000251, 0.00019 }`.

The main slowdown seems to be due to [usage of `vim.version()`](https://github.com/neovim/neovim/blob/ea1c9f60e04429e8472370b59616d069ad7fbc8c/runtime/lua/vim/_editor.lua#L1058-L1059), which in turn seems to be due to [usage of `vim.fn.api_info()`](https://github.com/neovim/neovim/blob/ea1c9f60e04429e8472370b59616d069ad7fbc8c/runtime/lua/vim/version.lua#L455).

### Expected behavior

Deprecating (except hard deprecating) should not significantly impact performance of a deprecated function.

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2976+g208852126

### Operating system/version

EndeavourOS Linux x86_64 (6.8.7-arch1-1)

### Terminal name/version

st-0.9

st-256color

### Installation

appimage

echasnovski:
Problem: `vim.deprecate()` can be relatively significantly slower than
  the deprecated function in "Nvim" plugin.
Solution: Optimize checks for "Nvim" plugin. This also results into not
  distinguishing "xxx-dev" and "xxx" versions when doing checks, which
  is essentially covered by the deprecation logic itself.

Resolve #28459.

------

With this rewrite I get the times from #28459: `{ 0.024827, 0.003797, 0.002024, 0.001774, 0.001703 }`.
For quicker reference:
    -  On current Nightly it is something like `{ 3.72243, 0.918169, 0.968143, 0.763256, 0.783424 }`.
    - On 0.9.5: `{ 0.002955, 0.000361, 0.000281, 0.000251, 0.00019 }`.

------

If these changes are too much, another approach might be to treat `vim.tbl_islist()` as a special case and use custom logic there instead of `vim.deprecate()`. Something along the lines of `if vim.fn.has('nvim-0.11') == 1 then ...`, etc.

### DIFF
--- a/runtime/lua/vim/_editor.lua
+++ b/runtime/lua/vim/_editor.lua
@@ -1042,43 +1042,32 @@ end
 ---
 ---@return string|nil # Deprecated message, or nil if no message was shown.
 function vim.deprecate(name, alternative, version, plugin, backtrace)
-  vim.validate {
-    name = { name, 'string' },
-    alternative = { alternative, 'string', true },
-    version = { version, 'string', true },
-    plugin = { plugin, 'string', true },
-  }
   plugin = plugin or 'Nvim'
   local will_be_removed = 'will be removed'
 
   -- Only issue warning if feature is hard-deprecated as specified by MAINTAIN.md.
   -- Example: if removal_version is 0.12 (soft-deprecated since 0.10-dev), show warnings starting at
-  -- 0.11, including 0.11-dev (hard_deprecated_since = 0.11-dev).
+  -- 0.11, including 0.11-dev
   if plugin == 'Nvim' then
-    local current_version = vim.version() ---@type vim.Version
-    local removal_version = assert(vim.version.parse(version))
-    local is_hard_deprecated ---@type boolean
+    local major, minor = version:match('(%d+)%.(%d+)')
+    major, minor = tonumber(major), tonumber(minor)
 
-    if removal_version.minor > 0 then
-      local hard_deprecated_since = assert(vim.version._version({
-        major = removal_version.major,
-        minor = removal_version.minor - 1,
-        patch = 0,
-        prerelease = 'dev', -- Show deprecation warnings in devel (nightly) version as well
-      }))
-      is_hard_deprecated = (current_version >= hard_deprecated_since)
-    else
-      -- Assume there will be no next minor version before bumping up the major version;
-      -- therefore we can always show a warning.
-      assert(removal_version.minor == 0, vim.inspect(removal_version))
-      is_hard_deprecated = true
-    end
-
+    local hard_deprecated_since = string.format('nvim-%d.%d', major, minor - 1)
+    -- Assume there will be no next minor version before bumping up the major version
+    local is_hard_deprecated = minor == 0 or vim.fn.has(hard_deprecated_since) == 1
     if not is_hard_deprecated then
       return
-    elseif current_version >= removal_version then
-      will_be_removed = 'was removed'
     end
+
+    local removal_version = string.format('nvim-%d.%d', major, minor)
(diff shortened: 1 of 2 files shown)


justinmk (approved):

dundargoc (approved):

miduddin:
This seems to break diffview.nvim for me (not sure if its on neovim or diffview). Related issue: https://github.com/sindrets/diffview.nvim/issues/488

zeertzjq:
Hmm, that already happens for `vim.notify()` as well.

echasnovski:
> This seems to break diffview.nvim for me (not sure if its on neovim or diffview). Related issue: [sindrets/diffview.nvim#488](https://github.com/sindrets/diffview.nvim/issues/488)

*If* it is this PR that broke it, then it will be due to `vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback` and this one introduced usage of `vim.fn.has()`.

My first instinct is to suggest plugins to conditionally move to `vim.islist()` right away in such cases.

Otherwise, reverting this change and **highly** optimizing `vim.version()` might be the route here. But I think reaching the 0.002 ms mark will be tough here.

bfredl:
We could mark `fn.has` as async-safe (just need an exception for `has("python")` unless the recently provider refactor made it safe as well)

sid-6581:
This has broken a lot of plugins for a lot of people, so I suggest reverting it if a fix can't be implemented quickly. It's unrealistic to expect multiple plugins to update immediately. As it stands, I can't use neovim at all unless I revert this.

justinmk:
@sid-6581 which change broke plugins exactly, and how?

`tbl_istlist` only calls vim.deprecate(). That isn't going to be reverted. It has no effect (except a millisecond change in run-time) until 0.11 anyway.

echasnovski:
> This has broken a lot of plugins for a lot of people, so I suggest reverting it if a fix can't be implemented quickly. It's unrealistic to expect multiple plugins to update immediately. As it stands, I can't use neovim at all unless I revert this.

Reverting this will make a lot of those plugins unusable on Nightly anyway, because of `vim.tbl_islist()` will be about 1000x times slower.

The only solution I see is [suggested by @bfredl](https://github.com/neovim/neovim/pull/28470#issuecomment-2074650012).

clason:
@sid-6581 polite reminder: we make zero guarantees about HEAD -- you're welcome to use it, but shit _will_ break. We appreciate a heads-up about possible effects on plugins, but we _don't_ appreciate the tone of "fix it nao _or else_".

justinmk:
https://github.com/neovim/neovim/pull/28492 will fix the `E5560: Vimscript function must not be called in a lua loop callback` issue.

PR #28471, opened 2024-04-23
fix(type): allow nil severity field in diagnostic.GotoOpts

guilhas07:
## Problem

A user when using the functions : 
- `vim.diagnostic.goto_next` 
- `vim.diagnostic.goto_prev`

And trying to pass a table e.g., `vim.diagnostic.goto_next({ float = false })` will see a warning reporting missing fields `severity`, despite the field not being needed.

## Solution
Set `severity` field as optional.


gpanders (approved):

dundargoc:
Thank you for your contribution! We collect all typo fixes and similar small fixes into a single pull request and merge it once it gets big enough: https://github.com/neovim/neovim/pull/28208

PR #28480, opened 2024-04-24
feat(api): allow floats to be opened in non-current tabpage

### LINKED ISSUE #28234: Relative floating window rendered on wrong tab page
### Problem

Creating a relative floating window when on another tab page, creates the floating window for the current tab page, instead of the actual tab page for the one where the actual window is located. Further more going to the first tab page where the target window is actually located, has no floating window.

Another issue is that for some reason the tab page is obscured when the floating window is rendered
image
 
If you go to the first tab page, then go back to the second the tab page line is correctly rendered above and not overlapping with the floating window.
image

### Steps to reproduce

```lua

vim.cmd("set showtabline=2")

local winid = vim.api.nvim_get_current_win()
vim.print('current window on page 1 ' .. winid)
vim.cmd.tabnew()

local lines = { "Test float text" }
vim.api.nvim_buf_set_option(bufnr, 'modifiable', true)
vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, lines)
vim.api.nvim_buf_set_option(bufnr, 'modifiable', false)

local winid2 = vim.api.nvim_get_current_win()
vim.print('current window on page 2 ' .. winid2)

vim.api.nvim_open_win(bufnr, false, { win = winid, width = 50, col = 0, row = 0, height = 1, relative = 'win' })

```

### Expected behavior

Floating relative windows should be tab page aware. 

### Neovim version (nvim -v)

0.9.5

no

### Operating system/version

ubuntu 22

### Terminal name/version

alacritty

alacritty

### Installation

github

willothy:
Allows floating windows to be opened on non-current tabpages by using `relative = "win"` and assigning `win` to a window in a non-current tabpage. Also prevents `win_set_config` from moving the autocmd win to another tabpage.

fixes #28234

### DIFF
--- a/src/nvim/api/win_config.c
+++ b/src/nvim/api/win_config.c
@@ -236,19 +236,19 @@ Window nvim_open_win(Buffer buffer, Boolean enter, Dict(win_config) *config, Err
 
   win_T *wp = NULL;
   tabpage_T *tp = curtab;
-  if (is_split) {
-    win_T *parent = NULL;
-    if (config->win != -1) {
-      parent = find_window_by_handle(fconfig.window, err);
-      if (!parent) {
-        // find_window_by_handle has already set the error
-        goto cleanup;
-      } else if (parent->w_floating) {
-        api_set_error(err, kErrorTypeException, "Cannot split a floating window");
-        goto cleanup;
-      }
+  win_T *parent = NULL;
+  if (config->win != -1) {
+    parent = find_window_by_handle(fconfig.window, err);
+    if (!parent) {
+      // find_window_by_handle has already set the error
+      goto cleanup;
+    } else if (is_split && parent->w_floating) {
+      api_set_error(err, kErrorTypeException, "Cannot split a floating window");
+      goto cleanup;
     }
-
+    tp = win_find_tabpage(parent);
+  }
+  if (is_split) {
     if (!check_split_disallowed_err(parent ? parent : curwin, err)) {
       goto cleanup;  // error already set
     }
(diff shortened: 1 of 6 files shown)


src/nvim/api/win_config.c
```diff
@@ -409,23 +409,28 @@ void nvim_win_set_config(Window window, Dict(win_config) *config, Error *err)
...
+    parent = find_window_by_handle(fconfig.window, err);
+    if (!parent) {
+      return;
+    } else if (to_split && parent->w_floating) {
+      api_set_error(err, kErrorTypeException, "Cannot split a floating window");
+      return;
+    }
+
+    // Prevent autocmd window from being moved into another tabpage
+    if (is_aucmd_win(win) && win_find_tabpage(win) != win_find_tabpage(parent)) {
```
willothy, line 423: @zeertzjq wasn't totally sure about the autocmd win (didn't know about it at all before this tbh). Is just checking and throwing an error a reasonable way to ensure it's not moved to another tabpage by `set_config`?

zeertzjq: Yes

willothy: Awesome, thanks!

src/nvim/winfloat.c
```diff
@@ -41,7 +41,23 @@
 win_T *win_new_float(win_T *wp, bool last, WinConfig fconfig, Error *err)
 {
   if (wp == NULL) {
-    wp = win_alloc(last ? lastwin : lastwin_nofloating(), false);
+    tabpage_T *tp = NULL;
+    win_T *tp_last = last ? lastwin : lastwin_nofloating();
+    if (fconfig.window != 0) {
```
zeertzjq, line 46:
```suggestion
    if (fconfig.window != 0) {
      assert(!last);
```

willothy:
Awesome, glad this will be in 0.10 :)

PR #28483, opened 2024-04-24
feat(lsp): support vim.lsp.ListOpts for typehierarchy()

tom-anders:


runtime/lua/vim/lsp/handlers.lua
```diff
@@ -593,8 +593,19 @@ local function make_type_hierarchy_handler()
...
+    if config.loclist then
+      vim.fn.setloclist(0, {}, ' ', { title = title, items = items, context = ctx })
+      api.nvim_command('lopen')
+    elseif config.on_list then
+      assert(type(config.on_list) == 'function', 'on_list is not a function')
+      config.on_list({ title = title, items = items, context = ctx })
+    else
+      vim.fn.setqflist({}, ' ', { title = title, items = items, context = ctx })
+      api.nvim_command('botright copen')
+    end
```
MariaSolOs, line 608: You can even make the entire table a shared variable:
```suggestion
    config = config or {}
    local list = { title = 'LSP type hierarchy', items = items, context = ctx }

    if config.loclist then
      vim.fn.setloclist(0, {}, ' ', list)
      api.nvim_command('lopen')
    elseif config.on_list then
      assert(type(config.on_list) == 'function', 'on_list is not a function')
      config.on_list(list)
    else
      vim.fn.setqflist({}, ' ', list)
      api.nvim_command('botright copen')
    end
```

runtime/doc/lsp.txt
```diff
@@ -1437,13 +1437,14 @@ type_definition({options})                     *vim.lsp.buf.type_definition()*
     Parameters: ~
       • {options}  (`vim.lsp.LocationOpts?`) See |vim.lsp.LocationOpts|.
 
-typehierarchy({kind})                            *vim.lsp.buf.typehierarchy()*
+typehierarchy({kind}, {options})                 *vim.lsp.buf.typehierarchy()*
```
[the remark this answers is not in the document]
tom-anders: Ah right, I basically `vim.lsp.buf.references()` also calls this `options` instead of `opts`, we should probably change that as well then?

tom-anders:
Thanks for the review!

runtime/doc/lsp.txt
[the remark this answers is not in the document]
justinmk: > I basically copied this from `vim.lsp.buf.references()` which also calls this `options` instead of `opts`, we should probably change that as well then?

yes. shorter is better. and "options" means something very different in vim/nvim.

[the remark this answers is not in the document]
tom-anders: Right, I added an additional commit that fixes this for all other functions in buf.lua

runtime/doc/lsp.txt
```diff
@@ -1245,30 +1245,30 @@ add_workspace_folder({workspace_folder})
 clear_references()                            *vim.lsp.buf.clear_references()*
     Removes document highlights from current buffer.
 
-code_action({options})                             *vim.lsp.buf.code_action()*
+code_action({opts})                                *vim.lsp.buf.code_action()*
```
MariaSolOs, line 1248: Maybe its something with my browser, but there seems to be some trailing whitespace here (?). Assuming you didn't modify these files manually and instead ran `make doc`, this might be a bug in the doc generator (cc @lewis6991)

zeertzjq: More whitespace is needed to align the tags.

MariaSolOs: Oh duh. In that case ignore my comment lol.

tom-anders:
Latest push fixes the lua warning about a shadowed parameter. Is there any way to disable the next.txt check? Or do we just ignore it?

glepnir:
news for *feat* change scope will skip it.

tom-anders:
> news for _feat_ change scope will skip it.

Right, but I think `feat` is appropiate here, isn't it? Or should this be a `fix` instead?

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -593,8 +593,19 @@ local function make_type_hierarchy_handler()
...
+      vim.fn.setloclist(0, {}, ' ', list)
+      api.nvim_command('lopen')
+    elseif config.on_list then
+      assert(type(config.on_list) == 'function', 'on_list is not a function')
+      config.on_list(list)
+    else
+      vim.fn.setqflist({}, ' ', list)
+      api.nvim_command('botright copen')
+    end
   end
```
justinmk, line 609: is there a test that at least exercises most of this?

runtime/lua/vim/lsp/buf.lua
```diff
@@ -47,6 +47,16 @@ local function request_with_options(name, params, options)
   request(name, params, req_handler)
 end
 
+local function client_request_with_options(client, name, params, options, bufnr)
```
justinmk, line 50: don't repeat parameters in the name, unless it is truly necessary to disambiguate from some other local function. the docstring should be enough (which this currently lacks)

```suggestion
--- Calls client.request with the given `opts`.
local function client_request(client, name, params, opts, bufnr)
```

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -593,8 +593,19 @@ local function make_type_hierarchy_handler()
...
       })
     end
-    vim.fn.setqflist({}, ' ', { title = 'LSP type hierarchy', items = items })
-    api.nvim_command('botright copen')
+    config = config or {}
+    local list = { title = 'LSP type hierarchy', items = items, context = ctx }
+
+    if config.loclist then
+      vim.fn.setloclist(0, {}, ' ', list)
+      api.nvim_command('lopen')
```
lewis6991, line 601:
```suggestion
      vim.cmd.lopen()
```

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -593,8 +593,19 @@ local function make_type_hierarchy_handler()
...
+    elseif config.on_list then
+      assert(type(config.on_list) == 'function', 'on_list is not a function')
+      config.on_list(list)
+    else
+      vim.fn.setqflist({}, ' ', list)
+      api.nvim_command('botright copen')
```
lewis6991, line 607:
```suggestion
      vim.cmd('botright copen')
```

tom-anders:
Thanks for the review again!

I noticed that the `loclist` key to `vim.lsp.ListOpts` is not actually documented, but most LSP handlers still support it (e.g. references, documentSymbols, workspaceSymbol). There are some LSP handlers though that do not support it (declaration, definition, typeDefinition, implementation). Should I fix this in this PR as well, or maybe open a new one instead?

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -593,8 +593,19 @@ local function make_type_hierarchy_handler()
...
-    vim.fn.setqflist({}, ' ', { title = 'LSP type hierarchy', items = items })
-    api.nvim_command('botright copen')
+    config = config or {}
+    local list = { title = 'LSP type hierarchy', items = items, context = ctx }
+
+    if config.loclist then
+      vim.fn.setloclist(0, {}, ' ', list)
+      vim.cmd.lopen()
+    elseif config.on_list then
+      assert(type(config.on_list) == 'function', 'on_list is not a function')
```
wookayin, line 603: Please consider using `vim.is_callable()` instead.

tom-anders:
All suggestions should be applied now, I moved the cleanup to a new PR: #28637

test/functional/plugin/lsp_spec.lua
```diff
@@ -3597,6 +3597,94 @@ describe('LSP', function()
       eq(expected, qflist)
     end)
 
+    it('Opens the loclist if config contains `loclist`', function()
+      clear()
```
MariaSolOs, line 3601: I don't think this `clear` is needed since the `before_each` test function already calls it.

runtime/lua/vim/lsp/handlers.lua
famiu: `vim.cmd.botright('copen')` would be better, no?

zeertzjq: `vim.cmd.copen({ mods = { split = 'botright' } })`

tom-anders: @zeertzjq done :)

tom-anders:
(rebased to fix conflicts caused by #28637)

runtime/lua/vim/lsp/buf.lua
```diff
@@ -47,6 +47,17 @@ local function request_with_opts(name, params, opts)
   request(name, params, req_handler)
 end
 
+--- Calls client.request with the given `opts`.
+local function client_request_with_options(client, name, params, opts, bufnr)
```
justinmk, line 51: is this different from `request_with_opts` which was merged in https://github.com/neovim/neovim/pull/28637 ?

PR #28491, opened 2024-04-24
perf(diagnostic): avoid table copies to filter by severity

mfussenegger:
Instead of adding all diagnostics matching lnum filters to a table, and
then copying that table to another table while applying the severity
filter, this changes the flow to only add diagnostics matching both
filters in the first pass.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -372,43 +372,46 @@ local function to_severity(severity)
 end
 
 --- @param severity vim.diagnostic.SeverityFilter
+--- @return fun(vim.Diagnostic):boolean
+local function severity_predicate(severity)
+  if type(severity) ~= 'table' then
+    severity = assert(to_severity(severity))
+    ---@param d vim.Diagnostic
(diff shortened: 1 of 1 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -372,43 +372,46 @@ local function to_severity(severity)
...
   --- @cast severity vim.diagnostic.Severity[]
-
   local severities = {} --- @type table<vim.diagnostic.Severity,true>
   for _, s in ipairs(severity) do
     severities[assert(to_severity(s))] = true
   end
 
   --- @param t vim.Diagnostic
-  return vim.tbl_filter(function(t)
+  return function(t)
```
gpanders, line 402: Any reason to stick with `t` here instead of `d` like the others above?

gpanders (approved):

runtime/lua/vim/diagnostic.lua
mfussenegger: No real reason, I just missed it

PR #28492, opened 2024-04-24
fix(lua): vim.fn.has('nvim-0.10') in fast context, used by vim.deprecate

bfredl:


src/nvim/eval.c
```diff
@@ -8840,7 +8840,7 @@ typval_T eval_call_provider(char *provider, char *method, list_T *arguments, boo
 }
 
 /// Checks if provider for feature `feat` is enabled.
-bool eval_has_provider(const char *feat)
+bool eval_has_provider(const char *feat, bool throw_if_fast)
```
justinmk, line 8843: it's hard to choose a good name for this arg. but i think `is_fast` would be less confusing. because the caller is telling eval_has_provider if it's in a fast context, right? we always want to throw if this is called in a fast context.
```suggestion
bool eval_has_provider(const char *feat, bool is_fast)
```

justinmk (approved):

src/nvim/eval.c
bfredl: no, it is telling `eval_has_provider` that it should throw if we are in a fast context. as the current name suggests.

Ideally this should only have been a change in `f_has` but that would require repeating the check four times or restructure a lot of things. thus this little hotfix for the moment.

justinmk: > it is telling `eval_has_provider` that it should throw if we are in a fast context. as the current name suggests.

when would we _not_ want to throw if we're in a fast context? e.g. I see that `false` is passed in `command_line_handle_key`. Why not pass `true` there?

echasnovski:
Thanks, @bfredl!

Issue #25272, opened 2023-09-20
The culling of vim.lsp.util


gpanders:
### Problem

vim.lsp.util exists

### Expected behavior

vim.lsp.util does not exist.

MariaSolOs:
[`try_trim_markdown_code_blocks`](https://github.com/MariaSolOs/neovim/blob/23c21e763074d401e8b36a91e6568bebc1655ce9/runtime/lua/vim/lsp/util.lua#L1996) also feels like an utility function that we should get rid of.

mfussenegger:
> _get_completion_item_kind_name
> This is a one line function. Why is this needed? Can it be inlined/deleted? @mfussenegger

Looks like it was due to testing: https://github.com/neovim/neovim/commit/9a67b030d9a054648296b45b615684dee768582d 
But inlining sounds good to me - or a candidate for ⬇️

> text_document_completion_list_to_complete_items
> The name alone suggests this is a special purpose function and doesn't belong in "util". Is there a more appropriate location?

There is some discussion around introducing a `lsp.completion` module. My preference would be to then make it private within that module. See https://github.com/neovim/neovim/pull/24661#issuecomment-1691150950


> _str_utfindex_enc and _str_byteindex_enc
> These are advertised as "convenience wrappers" around vim.str_utfindex and vim.str_byteindex, respectively. Consider > updating those functions to accept an encoding instead.

Another potential alternative would be to mix this into some kind of position or range abstraction. I have some vague idea about that, but won't be able to try it anytime soon.


+1 for all the others.

justinmk:
> `_normalize_markdown` should be made public, but not under `vim.lsp.util`. TBD on the correct namespace

- ✅ We need a [`vim.text` module](https://github.com/neovim/neovim/pull/26069) for various text manipulation things (like `vim.text.indent()`, `vim.text.dedent()`).
    - Counterpoint: `normalize_markdown` is markdown specific so it should live in `vim.filetype.markdown` ?

> Another potential alternative would be to mix this into some kind of position or range abstraction.

Reference: https://github.com/neovim/neovim/issues/25509

💯  for the other suggestions in this issue, with these comments:

- ✅  Should the `open_float` features be owned by `nvim_open_win` instead of adding `vim.ui.open_float` ?
    - done: https://github.com/neovim/neovim/pull/25550
[...]

PR #28494, opened 2024-04-24
refactor(lsp): extract `resolve_bufnr` to `vim.lsp.util`

ofseed:
When trying to do some stuff for codelens, I found that the local function `resolve_bufnr` is slightly different from the others which have the same name but in other files.

It might be better if we put it into `vim.lsp.utils` so that we can use the name as expected.

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -68,18 +68,6 @@ lsp._request_name_to_capability = {
 
 -- TODO improve handling of scratch buffers with LSP attached.
 
---- Returns the buffer number for the given {bufnr}.
----
----@param bufnr (integer|nil) Buffer number to resolve. Defaults to current buffer
----@return integer bufnr
-local function resolve_bufnr(bufnr)
-  validate({ bufnr = { bufnr, 'n', true } })
-  if bufnr == nil or bufnr == 0 then
-    return api.nvim_get_current_buf()
-  end
-  return bufnr
-end
-
 ---@private
 --- Called by the client when trying to call a method that's not
 --- supported in any of the servers registered for the current buffer.
@@ -242,7 +230,7 @@ end
 function lsp.start(config, opts)
   opts = opts or {}
   local reuse_client = opts.reuse_client or reuse_client_default
-  local bufnr = resolve_bufnr(opts.bufnr)
+  local bufnr = util.resolve_bufnr(opts.bufnr)
 
   for _, client in pairs(all_clients) do
     if reuse_client(client, config) then
@@ -467,7 +455,7 @@ end
 ---Buffer lifecycle handler for textDocument/didSave
 --- @param bufnr integer
 local function text_document_did_save_handler(bufnr)
-  bufnr = resolve_bufnr(bufnr)
+  bufnr = util.resolve_bufnr(bufnr)
   local uri = vim.uri_from_bufnr(bufnr)
   local text = once(lsp._buf_get_full_text)
   for _, client in ipairs(lsp.get_clients({ bufnr = bufnr })) do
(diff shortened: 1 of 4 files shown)


ofseed:
Not sure if it should be private.

MariaSolOs:
I'm unsure about this change. Although I admit that we repeat this logic in multiple places, some LSP functions interpret the meaning of `nil` differently and so there's no global agreement on what resolving the buffer number is (e.g. `vim.lsp.codelens.clear` will treat a nil buffer parameter refreshing all buffers, while 0 means the current buffer).

ofseed:
@MariaSolOs  Thanks for pointing this out, I hadn't noticed that `vim.lsp.codelens.clear` also uses this function. But in my opinion, if the functions have the same name (even if they are defined in different places), it's best if they do the same thing. So far, it seems that the most useful use of this function is to treat `nil` as the current buffer, not all buffers.

I think the problem you mentioned is essentially due to the design of this API. The current codelens API is designed to be easily triggered manually. In this case, we may need a method to indicate that it is effective for all buffers. However, for other LSP-related APIs, such as semantic token or inlay hint, `nil` represents the current buffer, not all buffers. So for the consistency of the API, I even suggest changing this behavior, but this will be another breaking change, so it is better not to do it.

By the way, the stuff I am trying to work on is rewriting codelens to provide something like `vim.lsp.codelens.enable` like `vim.lsp.inlay_hint.enable`, I see `vim.diagnostic` made a similar change recently, so it should be acceptable, and the we have no more special case like this.

mfussenegger:
Nothing of `resolve_bufnr` is specific to lsp, so it imho doesn't really fit into vim.lsp.util (which we kinda want to get rid of https://github.com/neovim/neovim/issues/25272).

If useful enough, it should be in a general purpose module - but given how trivial it is and the aspects @MariaSolOs mentioned I'm not convinced it's worth having as public API (but open to more inputs, together with suggestions where it would fit)

mfussenegger:
I should probably also point out that it's current implementation contradicts the dev guidelines: 

https://github.com/neovim/neovim/blob/a1550dbf0ac252052a7a71780708d3ec05aa3686/runtime/doc/develop.txt#L315

justinmk:
> ome LSP functions interpret the meaning of `nil` differently and so there's no global agreement on what resolving the buffer number is (e.g. `vim.lsp.codelens.clear` will treat a nil buffer parameter refreshing all buffers, while 0 means the current buffer).

we should fix that though. having common functions helps with that. `vim._util` module for internal use would be better than ad-hoc duplication everywhere, to start with. and will help avoid the inconsistencies.

gpanders:
>the preferred convention for bufnr/winnr/etc is:
>
>* nil means "all"
>* 0 means "current"

Which means we really shouldn't need these "resolve" helper functions at all, since most (all?) API functions take `0` as "current buffer" anyway, and if bufnr is nil then the function has to handle that independently anyway.

ofseed:
> the preferred convention for bufnr/winnr/etc is:
>
> nil means "all"
> 0 means "current"

If it needs to be implemented in this way, it will be convenient for users. But it's a bit hard to implement. There are a few problems we may encounter.

1. Sometimes the `bufnr` is not a direct parameter for these API functions, it might be wrapped in a table, e.g. `opts.bufnr` in `vim.lsp.codelens.refresh` and `filter.bufnr` in `vim.lsp.inlay_hint.enable`, should we do this in these places?
2. The current implementation of many of these functions, like `vim.lsp.inlay_hint.enable`, needs a lot of work to make it take effect on all buffers, because 
    1. `0` or any specific number is for one buffer, this is how these functions are designed;
    2. Intuitively, "all buffers" should not only refer to "all current buffers," but also include all future buffers. For `vim.lsp.inlay_hint.enable`, all current buffers should be enabled, and the same for the future buffers.

Then like @gpanders said, we do not need these "resolve" helper functions at all. So the best way I can think of at the moment is,  to delete all resolve_bufnr functions. If a function can handle the "all buffer" situation, then `bufnr` is optional; otherwise at least `0` must be passed as `bufnr`. Since we have lua_ls and a lot of annotations, this should not be a big problem. Then we can gradually add support for processing "all buffers" to the functions in need, or think that this function should only handle one buffer, and then mark `bufnr` as required instead of optional. 

However, this approach maintains API consistency.

justinmk:
> So the best way I can think of at the moment is, to delete all resolve_bufnr functions. If a function can handle the "all buffer" situation, then `bufnr` is optional; otherwise at least `0` must be passed as `bufnr`

sounds like a reasonable cleanup task. could be risky so should wait until 0.10 is released in a couple weeks.

PR #28495, opened 2024-04-24
fix(man.vim): q quits after jump to different tag in MANPAGER modified

RandomChugokujin:
Problem: In MANPAGER mode, pressing q after jumping to a different tag does not quit, instead E444 is thrown (#25943)

Solution: 
1. Change `b:pager` into `g:pager`, a global
2. Functions in man.lua now will not change the status of `g:pager` if it has already been set.

### DIFF
--- a/runtime/lua/man.lua
+++ b/runtime/lua/man.lua
@@ -411,15 +411,13 @@ local function find_man()
   return false
 end
 
----@param pager boolean
-local function set_options(pager)
+local function set_options()
   vim.bo.swapfile = false
   vim.bo.buftype = 'nofile'
   vim.bo.bufhidden = 'unload'
   vim.bo.modified = false
   vim.bo.readonly = true
   vim.bo.modifiable = false
-  vim.b.pager = pager
   vim.bo.filetype = 'man'
 end
 
@@ -475,7 +473,7 @@ local function put_page(page)
   vim.cmd([[silent! keeppatterns keepjumps %s/\s\{199,}/\=repeat(' ', 10)/g]])
   vim.cmd('1') -- Move cursor to first line
   highlight_man_page()
-  set_options(false)
+  set_options()
 end
 
 local function format_candidate(path, psect)
@@ -662,7 +660,8 @@ function M.init_pager()
     vim.cmd.file({ 'man://' .. fn.fnameescape(ref):lower(), mods = { silent = true } })
   end
 
-  set_options(true)
+  vim.g.pager = true
+  set_options()
 end
 
 ---@param count integer
@@ -730,7 +729,7 @@ function M.open_page(count, smods, args)
   if not ok then
     error(ret)
   else
-    set_options(false)
+    set_options()
   end
 
   vim.b.man_sect = sect
--- a/test/functional/plugin/man_spec.lua
+++ b/test/functional/plugin/man_spec.lua
@@ -192,6 +192,7 @@ describe(':Man', function()
       '--headless',
       '+autocmd VimLeave * echo "quit works!!"',
       '+Man!',
+      '+tag ls',
       '+call nvim_input("q")',
     }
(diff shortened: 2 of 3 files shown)


runtime/lua/man.lua
```diff
@@ -729,6 +733,8 @@ function M.open_page(count, smods, args)
 
   if not ok then
     error(ret)
+  elseif vim.fn.exists('g:pager') == 1 then
+    set_options(vim.g.pager)
```
gpanders, line 737: If `g:pager` is a global it doesn't need to be set in `set_options`, so you can remove the argument from `set_options` and just change this to `set_options()`.

runtime/lua/man.lua
```diff
@@ -475,7 +475,11 @@ local function put_page(page)
   vim.cmd([[silent! keeppatterns keepjumps %s/\s\{199,}/\=repeat(' ', 10)/g]])
   vim.cmd('1') -- Move cursor to first line
   highlight_man_page()
-  set_options(false)
+  if vim.fn.exists('g:pager') == 1 then
+    set_options(vim.g.pager)
+  else
+    set_options(false)
+  end
```
gpanders, line 482:
```suggestion
  set_options()
```

runtime/lua/man.lua
```diff
@@ -419,7 +419,7 @@ local function set_options(pager)
   vim.bo.modified = false
   vim.bo.readonly = true
   vim.bo.modifiable = false
-  vim.b.pager = pager
+  vim.g.pager = pager
```
gpanders, line 422:
```suggestion
```

runtime/lua/man.lua
gpanders: GitHub does not let me comment on unchanged lines, but also change `set_options(true)` on line 669 to

```lua
vim.g.pager = true
set_options()
```

RandomChugokujin: That does seem like a more clean solution, but `g:pager` now does not exist in normal Man buffers. I don't know whether this is an ok behavior. Previously it is set to false.

gpanders: It's fine if `g:pager` doesn't exist in normal Man buffers.

runtime/lua/man.lua
```diff
@@ -729,6 +728,8 @@ function M.open_page(count, smods, args)
 
   if not ok then
     error(ret)
+  elseif vim.fn.exists('g:pager') == 1 then
+    set_options(vim.g.pager)
```
gpanders, line 732:
```suggestion
```

gpanders: And change the line below from `set_options(false)` to `set_options()`

RandomChugokujin: Sorry forgot this one

runtime/lua/man.lua
```diff
@@ -662,7 +660,8 @@ function M.init_pager()
     vim.cmd.file({ 'man://' .. fn.fnameescape(ref):lower(), mods = { silent = true } })
   end
 
-  set_options(true)
+  vim.g.pager = true
```
justinmk, line 663: Although this is a "private" (undocumented) variable, since it's now global we probably should prefix it, as we already do for `g:man_hardwrap` and others.

```suggestion
  vim.g.man_pager = true
```

test/functional/plugin/man_spec.lua
```diff
@@ -197,6 +197,19 @@ describe(':Man', function()
     matches('quit works!!', fn.system(args, { 'manpage contents' }))
   end)
 
+  it('q quits in "$MANPAGER mode" after jumping to a different tag (:Man!) #25943', function()
```
justinmk, line 200: this test is nearly a duplicate of the above. can you just add `+tag ls` to the above test , I don't think we need a new, separate test.

justinmk (approved):

PR #28497, opened 2024-04-24
build: do not use GIT_REPOSITORY for local dependencies

dundargoc:
This reverts a large portion of
2c1e8f7e96926f70151d737ea32f1e6ff3589263.

The conclusion from that commit is incorrect: local builds are not
used/updated correctly so much as it's not used at all. Instead the
build will always use `master` branch rather than the current files.


dundargoc:
Always running `make distclean` sucks but at least it's correct 🤷‍♂️

justinmk (approved):

dundargoc:
We can investigate how to make iterating local development faster in a future PR. This should suffice for now though.

Issue #24252, opened 2023-07-04
default mappings for LSP, diagnostics, etc


justinmk:
# Problem

LSP and diagnostics don't have core mappings. In the case of LSP some of its natural mappings "conflict" with ftplugins or the builtin normal commands.

Related:

- https://github.com/neovim/neovim/pull/24331
- https://github.com/neovim/neovim/pull/16230
- https://github.com/neovim/neovim/pull/28500
- https://github.com/neovim/neovim/pull/26745
- https://github.com/neovim/neovim/issues/24054
    - https://github.com/neovim/neovim/pull/18506

# Expected behavior

We could introduce a way to _opt-in_ to more aggressive, more risky defaults (essentially the antithesis of https://github.com/neovim/neovim/issues/6293 , which proposed a way to enable "nineties mode" for slow devices/networks). But that kind of sucks, it's not really a long-term solution, it creates an "inner platform" that then gets its own inertia and baggage.

Instead:

- ✅ revisit [#16230](https://github.com/neovim/neovim/pull/16230) (was there objection to `]g` and friends? potentially better alternatives, borrowing from vim-unimpaired, vim-scriptease, something else?)
    - ✅ we can probably claim `]d` and friends, the default behavior is rarely useful.
- ✅ resolve https://github.com/neovim/neovim/issues/24252#issuecomment-1620459930
- finish [#18506](https://github.com/neovim/neovim/pull/18506) to deliver the full `vim.lsp.config` interface
- ~~add `vim.lsp.buf.set_defaults({buf})`, which sets recommended default mappings and other settings.~~

## LSP defaults

The defaults should answer these questions:

> - What about all the lsp functions that one can call?
> - What do I need to do to trigger auto-formatting?
> - ✅ Does gd call the LSP by default when one is attached?
>   - ✅ What about K?
> - ✅ How do I run code actions?
> - ✅ What if I want to see the documentation of the completion entry while I browse them?
> - ✅ What if I need the entries to show more than just plain white text without context?
> - ✅ Inlay hints are cool but do I really need a custom on_attach function to enable them globally?

YodaEmbedding:
I'll start things off with a small selection from my [personal LSP mappings](https://github.com/YodaEmbedding/dotfiles/blob/master/nvim/.config/nvim/lua/mappings/_nvim_lsp.lua):

```lua
nnoremap("gd", vim.lsp.buf.definition)
nnoremap("gr", vim.lsp.buf.references)
[...]

PR #24977, opened 2023-09-02
docs(lsp): use protocol.Methods instead of raw strings

MariaSolOs:
IMO examples should encourage the use of `vim.lsp.protocol.Methods` instead of string literals.

Issue #24229, opened 2023-07-02
LSP events for dynamic capabilities


stasjok:
### Problem

Usually we define mappings or something similar in on_attach callback or in a `LspAttach` callback. With the introduction of dynamic capabilities it's not enough, because `client.supports_method()` can return false for something that would be registered lately.

### Expected behavior

I think it would be convenient if there was an event like `LspRegisterCapability` and `LspUnregisterCapability`. User, plugins (maybe even core for setting default settings) can register callbacks for these events. There could be some de-duplication logic so that events are fired only when there are changes in effective capabilities (is it possible that the same server registers same capability twice?). Changed capability should be in the data table.

If two new autocommands are too much maybe it can be only one for any changes in capabilities (`LspUpdateCapability`?). It should include what really changed in the data table. Should it also fire when the client first attached to a buffer?

As alternative maybe `client/registerCapability` and `client/unregisterCapability` can run a callback from config parameter. Then user can do something like:

```lua
vim.lsp.handlers["client/registerCapability"] = vim.lsp.with(
  vim.lsp.handlers["client/registerCapability"], {
    callback = function()
      -- do something
    end,
  }
)
```

It already can be done with something like:

```lua
local orig_handler = vim.lsp.handlers["client/registerCapability"]
vim.lsp.handlers["client/registerCapability"] = function(err, result, ctx)
  local result = orig_handler(err, result, ctx)
  -- do something
  return result
)
```

But it complicates things when multiple places want to add a callback. With autocommands completely separate callbacks can be defined or removed.

justinmk:
> It already can be done with something like:
> 
> ```lua
> local orig_handler = vim.lsp.handlers["client/registerCapability"]
> vim.lsp.handlers["client/registerCapability"] = function(err, result, ctx) ...
> ```
> 
> But it complicates things when multiple places want to add a callback.

https://github.com/neovim/neovim/pull/22598 should address this. 

In general the Nvim [events model](https://github.com/neovim/neovim/issues/19156) is:

- functions (which can be hooked into by `on_func()`)
- events (autocmds)

mfussenegger:
> #22598 should address this.
> 
> In general the Nvim [events model](https://github.com/neovim/neovim/issues/19156) is:
> 
> * functions (which can be hooked into by [on_func()](https://github.com/neovim/neovim/pull/22598))
> * events (autocmds)


[...]

PR #28500, opened 2024-04-25
feat(lsp): add more LSP defaults

gpanders:
Ref: https://github.com/neovim/neovim/issues/24252

- `crn` for rename
- `crr` for code actions
- `gr` for references
- `<C-S>` (in Insert mode) for signature help

`cr` is a generic "refactor" prefix (fits nicely with the mappings from [vim-abolish](https://github.com/tpope/vim-abolish/blob/master/doc/abolish.txt#L146)) too.

`crn` mnemonic: reName
`crr` is code actions for now, but can eventually be extended to provide a generic "menu" of refactor options

`gr` overrides a builtin mapping, but this is deemed acceptable since the default behavior is rarely useful. Users can delete the new default `gr` mapping if they don't want it.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -61,47 +61,41 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
+                                                  *crr* *v_crr* *crn* *i_CTRL-S*
+Some keymaps are created unconditionally when Nvim starts:
+- "crn" is mapped in Normal mode to |vim.lsp.buf.rename()|
+- "crr" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
+- "gr" is mapped in Normal mode to |vim.lsp.buf.references()| |gr-default|
+- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
+
+If not wanted, these keymaps can be removed at any time using
+|vim.keymap.del()| or |:unmap|.
+
                                                         *lsp-defaults-disable*
 To override the above defaults, set or unset the options on |LspAttach|: >lua
+
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(ev)
         vim.bo[ev.buf].formatexpr = nil
         vim.bo[ev.buf].omnifunc = nil
-        vim.keymap.del("n", "K", { buffer = ev.buf })
+        vim.keymap.del('n', 'K', { buffer = ev.buf })
       end,
     })
 
-To use other LSP features like hover, rename, etc. you can set other keymaps
-on |LspAttach|. Example: >lua
-    vim.api.nvim_create_autocmd('LspAttach', {
-      callback = function(args)
-        vim.keymap.set('n', 'K', vim.lsp.buf.hover, { buffer = args.buf })
-      end,
-    })
+To use other LSP features, set keymaps on |LspAttach|. Not all language
+servers provide the same capabilities. To ensure you only set keymaps if the
+language server supports a feature, guard keymaps behind capability checks.
+Example: >lua
 
-The most common functions are:
-
-- |vim.lsp.buf.hover()|
-- |vim.lsp.buf.format()|
-- |vim.lsp.buf.references()|
-- |vim.lsp.buf.implementation()|
-- |vim.lsp.buf.code_action()|
-
-
-Not all language servers provide the same capabilities. To ensure you only set
-keymaps if the language server supports a feature, you can guard the keymap
-calls behind capability checks:
->lua
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(args)
         local client = vim.lsp.get_client_by_id(args.data.client_id)
-        if client.server_capabilities.hoverProvider then
-          vim.keymap.set('n', 'K', vim.lsp.buf.hover, { buffer = args.buf })
+        if client.supports_method('textDocument/implementation') then
+          vim.keymap.set('n', 'g<C-I>', vim.lsp.buf.implementation, { buffer = args.buf })
         end
       end,
     })
 <
-
 To learn what capabilities are available you can run the following command in
 a buffer with a started LSP client: >vim
 
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -400,6 +400,10 @@ The following changes to existing APIs or features add new behavior.
   • 'comments' includes "fb:•".
   • 'shortmess' includes the "C" flag.
   • 'grepprg' defaults to using ripgrep if available.
+  • |crn| in Normal mode maps to |vim.lsp.buf.rename()|.
+  • |crr| in Normal and Visual mode maps to |vim.lsp.buf.code_action()|.
+  • "gr" in Normal mode maps to |vim.lsp.buf.references()| |gr-default|
+  • |i_CTRL-S| in Insert mode maps to |vim.lsp.buf.signature_help()|
   • Automatic linting of treesitter query files (see |ft-query-plugin|).
     Can be disabled via: >lua
       vim.g.query_lint_on = {}
@@ -430,9 +434,10 @@ The following changes to existing APIs or features add new behavior.
 
   :call netrw#BrowseX(expand(exists("g:netrw_gx")? g:netrw_gx : '<cfile>'), netrw#CheckIfRemote())<CR>
 
-• |vim.lsp.start()| now maps |K| to use |vim.lsp.buf.hover()| if the server
-  supports it, unless |'keywordprg'| was customized before calling
-  |vim.lsp.start()|.
+• |vim.lsp.start()| now creates the following default keymaps (assuming the
+  server supports the feature):
+  - |K| in Normal mode maps to |vim.lsp.buf.hover()|, unless |'keywordprg'|
+    was customized before calling |vim.lsp.start()|.
 
 • Terminal buffers started with no arguments (and use 'shell') close
   automatically if the job exited without error, eliminating the (often
--- a/runtime/doc/vim_diff.txt
+++ b/runtime/doc/vim_diff.txt
@@ -137,6 +137,10 @@ of these in your config by simply removing the mapping, e.g. ":unmap Y".
 - * |v_star-default|
 - gc |gc-default| |v_gc-default| |o_gc-default|
 - gcc |gcc-default|
+- |crn|
+- |crr|
+- gr |gr-default|
+- <C-S> |i_CTRL-S|
 - Nvim LSP client defaults |lsp-defaults|
   - K |K-lsp-default|
 
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -144,6 +144,31 @@ do
     end
     vim.keymap.set({ 'o' }, 'gc', textobject_rhs, { desc = 'Comment textobject' })
   end
+
+  --- Default maps for LSP functions.
+  ---
+  --- These are mapped unconditionally to avoid confusion. If no server is attached, or if a server
+  --- does not support a capability, an error message is displayed rather than exhibiting different
+  --- behavior.
+  ---
+  --- See |gr-default|, |crn|, |crr|, |i_CTRL-S|.
+  do
+    vim.keymap.set('n', 'crn', function()
+      vim.lsp.buf.rename()
+    end, { desc = 'vim.lsp.buf.rename()' })
+
+    vim.keymap.set({ 'n', 'v' }, 'crr', function()
+      vim.lsp.buf.code_action()
+    end, { desc = 'vim.lsp.buf.code_action()' })
+
+    vim.keymap.set('n', 'gr', function()
+      vim.lsp.buf.references()
+    end, { desc = 'vim.lsp.buf.references()' })
+
+    vim.keymap.set('i', '<C-S>', function()
+      vim.lsp.buf.signature_help()
+    end, { desc = 'vim.lsp.buf.signature_help()' })
+  end
 end
 
 --- Default menus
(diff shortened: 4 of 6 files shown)


runtime/doc/lsp.txt
```diff
@@ -60,6 +60,14 @@ options are not restored when the LSP client is stopped or detached.
...
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
+- *crr* is mapped in Normal mode to |vim.lsp.buf.rename()| if no existing
+  keymap exists
+- *cra* is mapped in Normal mode to |vim.lsp.buf.code_action()| if no existing
+  keymap exists
+- |gr| is mapped in Normal mode to |vim.lsp.buf.references()| if no existing
+  keymap exists
+- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()| if no
+  existing keymap exists *i_CTRL-S*
```
zeertzjq, line 70:
```suggestion
  mapping exists
- *cra* is mapped in Normal mode to |vim.lsp.buf.code_action()| if no existing
  mapping exists
- |gr| is mapped in Normal mode to |vim.lsp.buf.references()| if no existing
  mapping exists
- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()| if no
  existing mapping exists *i_CTRL-S*
```

MariaSolOs:
@folke @siduck @tjdevries (and everyone else that has an initial nvim configuration/distro, sorry you're a lot so I'm tagging the 1st that come to mind): You might want to reconsider your default keybinds to match these newly added ones.

runtime/doc/lsp.txt
```diff
@@ -60,6 +60,14 @@ options are not restored when the LSP client is stopped or detached.
   - To opt out of this use |gw| instead of gq, or set 'formatexpr' on LspAttach.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
+- *crr* is mapped in Normal mode to |vim.lsp.buf.rename()| if no existing
```
MariaSolOs, line 63: I think we should also document on how to modify/remove these if desired.

gpanders: Already mentioned just below this, but I updated the example to include the new default mappings too.

runtime/doc/lsp.txt
```diff
@@ -60,48 +60,43 @@ options are not restored when the LSP client is stopped or detached.
...
-Not all language servers provide the same capabilities. To ensure you only set
-keymaps if the language server supports a feature, you can guard the keymap
-calls behind capability checks:
->lua
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(args)
-        local client = vim.lsp.get_client_by_id(args.data.client_id)
-        if client.server_capabilities.hoverProvider then
-          vim.keymap.set('n', 'K', vim.lsp.buf.hover, { buffer = args.buf })
+        if client.supports_method(ms.textDocument_implementation) then
```
MariaSolOs, line 94: I love the use of the `Methods` module here; however I think there's still some discussion going on (see https://github.com/neovim/neovim/pull/24977) about its usage in examples.

mfussenegger: I think some of the arguments made in that PR don't apply to the example here, but it looks like the `ms` definition is missing and that would definitely trip up people.

mfussenegger (approved):
3/4 match my current mappings, good enough for me :)

~We may want to look into https://github.com/neovim/neovim/issues/24229 as a follow up at some point, because code-action has `dynamicRegistration = true`, which means at the time `_set_default` runs, `client.supports_method(ms.textDocument_codeAction)` can return false and later on change to true.~

*Update*: Looks like we account for that already: https://github.com/neovim/neovim/blob/2b9df3f79541507bfc3ba2645e6b9740e7c922ed/runtime/lua/vim/lsp/handlers.lua#L117

mfussenegger:
After thinking about these some more I wonder if we should actually map commands like `gr` unconditionally, independent of a language server running and supporting the actions. 

With K, gq, and tagfunc mappings there are reasonable fallbacks sharing the semantics of the LSP versions, so a conditional mapping to enhance them makes sense. 

With `gr` you get entirely different behaviour if no server is running, which might be more confusing than getting an error stating that no language server is running or supports getting references.

No strong opinion on it, just wanted to bring it up

justinmk:
>  `crr` for rename. `cra` for code actions

`crn` is a nice mnemonic for "ReName". And that allows us to use `crr` for the general "code actions" entrypoint, which is analogous to the way that `dd`, `yy`, etc have traditionally been used as "convenience aliases".

>  `<C-S>` (in Insert mode) for signature help

would `ctrl-]` be worth claiming instead? somewhat analogous to normal-mode `ctrl-]`.

I have plans for `ctrl-s` in normal mode, to go to the "default [S]hell". But that doesn't necessarily conflict with this insert-mode purpose.

gpanders:
> would `ctrl-]` be worth claiming instead? somewhat analogous to normal-mode `ctrl-]`.

`C-]` already exists in insert mode to expand an abbreviation. I also use that as my “expand snippet” trigger (expanding a snippet is similar in spirit to expanding an abbreviation).

gpanders:
> > `crr` for rename. `cra` for code actions
> 
> `crn` is a nice mnemonic for "ReName". And that allows us to use `crr` for the general "code actions" entrypoint, which is analogous to the way that `dd`, `yy`, etc have traditionally been used as "convenience aliases".


So you’re suggesting `crn` for rename and `crr` for code actions? That seems a lot more confusing than `r` for Rename and `a` for Action, but if it gets this out the door then sure.

justinmk:
> So you’re suggesting `crn` for rename and `crr` for code actions? That seems a lot more confusing than `r` for Rename and `a` for Action,

My thinking is the "cr" prefix is for Refactoring-related things. `crr` can show a "code changes" menu even for non-LSP contexts (similar to `gO` which has different "backends"). So the LSP "actions" jargon shouldn't necessary drive this.

`crr` is also easier to type for this somewhat common use-case.

mfussenegger:
One more point comes to mind: The code action mappings should probably exist for both normal and visual mode.
E.g. in eclipse.jdt.ls there's support to select a range, and then use a "extract method" code action.

No strong opinion on the `crn`/`crr` discussion, but I'd prefer `crr` for rename and `cra` for code action.
(But I also use `<a-CR>` for code actions due to intellij habits, and have an additional `<leader>r` mapping for `vim.lsp.buf.code_action { context = { only = {'refactor'}}}`)

gpanders:
> After thinking about these some more I wonder if we should actually map commands like `gr` unconditionally, independent of a language server running and supporting the actions.
> 
> With K, gq, and tagfunc mappings there are reasonable fallbacks sharing the semantics of the LSP versions, so a conditional mapping to enhance them makes sense.
> 
> With `gr` you get entirely different behaviour if no server is running, which might be more confusing than getting an error stating that no language server is running or supports getting references.
> 
> No strong opinion on it, just wanted to bring it up

I think this makes sense and I did so in 25b7b1b06f69e3c42f05ac43f9f26c19bbf33786. Note that some of these mappings like `vim.lsp.buf.references()` and `vim.lsp.buf.code_action()` don't show any error or message of any kind if no LSP client is attached, but we can fix that later.

>My thinking is the "cr" prefix is for Refactoring-related things. crr can show a "code changes" menu even for non-LSP contexts (similar to gO which has different "backends"). So the LSP "actions" jargon shouldn't necessary drive this.

Agree on the `cr` prefix. For the "menu" concept tho, why not use `crO` and continue the `O` = "table of contents" paradigm?

>And that allows us to use crr for the general "code actions" entrypoint, which is analogous to the way that dd, yy, etc have traditionally been used as "convenience aliases".

`dd` and `yy` and `gcc` and `g??` etc. have the semantics of "operate on the current line". `crr` for rename doesn't follow this, but neither does `crr` for a menu. So I don't think `dd` and friends provide this precedent.

Although, looking at what [actual people use today](https://sourcegraph.com/search?q=context:global+lang:Lua+vim.lsp.buf.rename&patternType=keyword&sm=0), the `rn` mnemonic for "rename" is more common than I would have expected. So that's a point in `crn`'s favor.

gpanders:
Changed rename to `crn`.

siduck:
> @folke @siduck @tjdevries (and everyone else that has an initial nvim configuration/distro, sorry you're a lot so I'm tagging the 1st that come to mind): You might want to reconsider your default keybinds to match these newly added ones.

will this get added in v0.10?

runtime/doc/lsp.txt
gpanders: Updated in ac59371ef9aa2c979043bee574e576bdf62fb664

justinmk:
> Agree on the `cr` prefix. For the "menu" concept tho, why not use `crO` and continue the `O` = "table of contents" paradigm?

"O" in `gO` means "outline", not "menu". And uppercase mappings are both hard to type and remember, so should be avoided, especially when we have a "family" that is prefixed with something like `cr`.

asmodeus812:
`_defaults` has become a hodgepodge of mappinngs related to lsp, treesitter (gcc), quite a few autocmds, then we have others which enhance the defaults (gx). I think it is time to think about splitting this. And allowing an opt out of certain ones. The original content of this file, with things like `*` and `#` in visual mode, `gx`, and the terminal and other `autocmds` which generally added nice value, but this is getting a bit too opinionated. Not to mention the current mappings for lsp are set regardless of lsp client being attached.

justinmk:
> `_defaults` has become a hodgepodge of mappinngs (vastly differing contexts) such as lsp, treesitter (gcc), quite a few autocmds, then we have others which enhance the defaults (gx). I think it is time to think about splitting this.

That's an implementation detail , and solving it can happen after some thought. Doesn't need to be rushed.

> And allowing an opt out of certain ones.

Opting out of LSP mappings is described at `:h lsp-defaults-disable`.

 @gpanders should we define these mappings in the same place where we define 'K':
https://github.com/neovim/neovim/blob/47dbda97d2f40729733b1c0d1d13d914065af23c/runtime/lua/vim/lsp.lua#L318

gpanders:
> Ok. Do we need to update `:h lsp-defaults-disable` ?

Done so

runtime/doc/lsp.txt
```diff
@@ -61,47 +61,41 @@ options are not restored when the LSP client is stopped or detached.
...
+                                                *gr-default*
+                                                *i_CTRL-S*
+Some keymaps are created unconditionally when Nvim starts:
+- *crn* is mapped in Normal mode to |vim.lsp.buf.rename()|
+- *cra* is mapped in Normal mode to |vim.lsp.buf.code_action()|
+- |gr| is mapped in Normal mode to |vim.lsp.buf.references()|
+- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
+
+If not wanted, these keymaps can be removed at any time using
+|vim.keymap.del()|.
```
justinmk, line 73:
```suggestion
|vim.keymap.del()| or |:unmap|.
```

runtime/doc/lsp.txt
```diff
@@ -61,47 +61,41 @@ options are not restored when the LSP client is stopped or detached.
...
->lua
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(args)
-        local client = vim.lsp.get_client_by_id(args.data.client_id)
-        if client.server_capabilities.hoverProvider then
-          vim.keymap.set('n', 'K', vim.lsp.buf.hover, { buffer = args.buf })
+        if client.supports_method('textDocument/implementation') then
+          vim.keymap.set('n', 'g<C-I>', vim.lsp.buf.implementation, { buffer = args.buf })
         end
       end,
```
justinmk, line 96: Nice cleanup! I think `local client = vim.lsp.get_client_by_id(args.data.client_id)` is still needed though?

runtime/doc/lsp.txt
```diff
@@ -61,47 +61,41 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
+                                                *gr-default*
+                                                *i_CTRL-S*
+Some keymaps are created unconditionally when Nvim starts:
+- *crn* is mapped in Normal mode to |vim.lsp.buf.rename()|
+- *cra* is mapped in Normal mode to |vim.lsp.buf.code_action()|
```
mfussenegger, line 68:
```suggestion
- *crr* is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
```

(ignore the `crr` parts if the conclusion was to stick with `cra` - I don't mind which)

runtime/doc/news.txt
```diff
@@ -388,6 +388,10 @@ The following changes to existing APIs or features add new behavior.
   • 'comments' includes "fb:•".
   • 'shortmess' includes the "C" flag.
   • 'grepprg' defaults to using ripgrep if available.
+  • |crn| in Normal mode maps to |vim.lsp.buf.rename()|.
+  • |cra| in Normal mode maps to |vim.lsp.buf.code_action()|.
```
mfussenegger, line 392:
```suggestion
  • |crr| in Normal and Visual mode maps to |vim.lsp.buf.code_action()|.
```

runtime/doc/vim_diff.txt
```diff
@@ -137,6 +137,10 @@ of these in your config by simply removing the mapping, e.g. ":unmap Y".
 - * |v_star-default|
 - gc |gc-default| |v_gc-default| |o_gc-default|
 - gcc |gcc-default|
+- |crn|
+- |cra|
```
mfussenegger, line 141:
```suggestion
- |crr|
```

runtime/lua/vim/_defaults.lua
```diff
@@ -143,6 +143,29 @@ do
...
+  --- does not support a capability, an error message is displayed rather than exhibiting different
+  --- behavior.
+  do
+    vim.keymap.set('n', 'crn', function()
+      vim.lsp.buf.rename()
+    end, { desc = 'vim.lsp.buf.rename()' })
+
+    vim.keymap.set('n', 'cra', function()
+      vim.lsp.buf.code_action()
+    end, { desc = 'vim.lsp.buf.code_action()' })
```
mfussenegger, line 159:
```suggestion
    vim.keymap.set({'n', 'v'}, 'crr', vim.lsp.buf.code_action, { desc = 'vim.lsp.buf.code_action()' })
```

runtime/lua/vim/_defaults.lua
```diff
@@ -143,6 +143,29 @@ do
...
+
+    vim.keymap.set('n', 'gr', function()
+      vim.lsp.buf.references()
+    end, { desc = 'vim.lsp.buf.references()' })
+
+    vim.keymap.set('i', '<C-S>', function()
```
justinmk, line 165: `i_CTRL-G_CTRL-P` (mnemonic: "parameters") is available, opens the door for future `i_CTRL-G` mappings.

runtime/doc/lsp.txt
gpanders: Thanks, removed that by accident.

runtime/doc/lsp.txt
gpanders: I think the conclusion is to leave `crr` open for now, and use `crn` for rename and `cra` for code actions. @justinmk?

justinmk: I'm in favor of `crr` for this PR. But I'm not going to block on that.

dundargoc:
> but this is getting a bit too opinionated.

I never understood this objection. What is the problem of changing the mapping (or unmapping) the mappings you dislike?

runtime/doc/lsp.txt
gpanders: Ok misunderstanding on my part, updated.

justinmk (approved):

runtime/lua/vim/_defaults.lua
gpanders: Not crazy about the double mapping, but we can revisit. We have a little time before this makes it to a release. If we did do the `<C-G>` prefix, I have a preference for `<C-G><C-S>` (mnemonic: "signature") since that can be typed with one hand (at least on QWERTY).

`<C-H>` might also be worth considering, though some people might use the default "backspace" behavior.

gpanders:
Reverted in #28649 to postpone until the 0.11 release cycle (since the 0.10 release is so close). See discussion in that PR and in #28634 for rationale. #28650 will re-introduce defaults in 0.11 so follow that PR for discussion. cc @siduck @folke @tjdevries.

PR #28501, opened 2024-04-25
syntax spec: add missing items to scripts section

gotmax23:
- %generate_buildrequires — added in RPM 4.15
- %conf — added in RPM 4.18

Ref: https://rpm-software-management.github.io/rpm/manual/spec.html#build-scriptlets


clason:
Thank you, but we don't maintain these files. Please contact the maintainer listed at the top of the file, who will push updates to vim/vim first.

PR #28502, opened 2024-04-25
fix(vim.ui)!: change open() to return pcall-like values

justinmk:
## Problem:
`vim.ui.open` unnecessarily invents a different success/failure convention. Its return type was changed in 57adf8c6e01d, so we might as well change it to have a more conventional form.

## Solution:
Change the signature to use the `pcall` convention of `status, result`.

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -98,18 +98,19 @@ do
   --- Map |gx| to call |vim.ui.open| on the <cfile> at cursor.
   do
     local function do_open(uri)
-      local cmd, err = vim.ui.open(uri)
-      local rv = cmd and cmd:wait(1000) or nil
-      if cmd and rv and rv.code ~= 0 then
-        err = ('vim.ui.open: command %s (%d): %s'):format(
+      local ok, cmd_or_err = vim.ui.open(uri)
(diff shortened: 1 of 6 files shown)


runtime/lua/vim/_defaults.lua
```diff
@@ -98,18 +98,19 @@ do
   --- Map |gx| to call |vim.ui.open| on the <cfile> at cursor.
   do
     local function do_open(uri)
-      local cmd, err = vim.ui.open(uri)
-      local rv = cmd and cmd:wait(1000) or nil
-      if cmd and rv and rv.code ~= 0 then
-        err = ('vim.ui.open: command %s (%d): %s'):format(
+      local ok, cmd_or_err = vim.ui.open(uri)
+      local rv = ok and (cmd_or_err --[[@as vim.SystemObj]]):wait(1000) or nil
```
justinmk, line 102: typescript type system can model this so that `ok=true` would be understood to always have a valid `vim.SystemObj`, then this `--[[@as vim.SystemObj]]` wouldn't be needed. I wonder if LuaLS can do that

justinmk:
May need to revert or revisit this, see https://github.com/neovim/neovim/pull/28478#discussion_r1579576304

PR #28505, opened 2024-04-25
refactor(lsp): rename `foos_by_bar` to `bar_foos` for better consistency

ofseed:
`client_hints` is an ambiguous name, because it can refer to two different meanings
* Lenses of one single client `lsp.InlayHint[]`, like in https://github.com/neovim/neovim/blob/b72ee822df933e6711ebcb0d0c8cb3a282b4890d/runtime/lua/vim/lsp/codelens.lua#L67
* Index lenses by client `table<integer, lsp.InlayHint[]>` (what it really means in this case)

It even names `hints_by_client` when assigned to a new variable
https://github.com/neovim/neovim/blob/a1c2da56ecef9c7a0e17be02f587d7c7f9eee170/runtime/lua/vim/lsp/inlay_hint.lua#L44
And there is a similar name with this naming style
https://github.com/neovim/neovim/blob/a1c2da56ecef9c7a0e17be02f587d7c7f9eee170/runtime/lua/vim/lsp/inlay_hint.lua#L180

### DIFF
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -210,8 +210,8 @@ local function clear(bufnr)
     return
   end
   local bufstate = bufstates[bufnr]
-  local client_lens = (bufstate or {}).client_hints or {}
-  local client_ids = vim.tbl_keys(client_lens) --- @type integer[]
+  local client_hints = (bufstate or {}).client_hints or {}
+  local client_ids = vim.tbl_keys(client_hints) --- @type integer[]
   for _, iter_client_id in ipairs(client_ids) do
     if bufstate then
       bufstate.client_hints[iter_client_id] = {}
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -41,13 +41,13 @@ function M.on_inlayhint(err, result, ctx, _)
     bufstate.client_hints = vim.defaulttable()
     bufstate.version = ctx.version
   end
-  local hints_by_client = bufstate.client_hints
+  local client_hints = bufstate.client_hints
   local client = assert(vim.lsp.get_client_by_id(client_id))
 
-  local new_hints_by_lnum = vim.defaulttable()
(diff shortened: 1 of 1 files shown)


justinmk (changes requested):
> `client_hints` is an ambiguous name, because it can refer to two different meanings

but that would be the case even if we renamed everything to `foo_by_client`.

`client_hints` is a better convention because the "key" is the client, so it's more intuitive to see that come first in the name.

If you want to rename things, rename the `foos_by_bar` stuff to `bar_foos`.

ofseed:
Thanks, I don't mind the naming, `bar_foos` does look better, it's just that both usages exist at the same time.

> If you want to rename things, rename the foos_by_bar stuff to bar_foos.

Then I think I should close this PR...or implement this? Feel free to close it.

ofseed:
I renamed all `foos_by_bar` to `bar_foos`, It’s up to you whether to close this PR or not.

justinmk (approved):
LGTM but I'm not intimate with this code, so I'll wait for others

MariaSolOs (approved):

ofseed:
Summary for these naming changes:

For a `table<integer, table<integer, Item[]>>`, name it to `foo_items`, then name its subitem to `bar_items`, which is a `table<integer, Item[]>`. Finally, name `Item[]` to `items`. Avoid using `items_by_bar` or `items_by_foo`.

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -41,13 +41,13 @@ function M.on_inlayhint(err, result, ctx, _)
     bufstate.client_hints = vim.defaulttable()
     bufstate.version = ctx.version
   end
-  local hints_by_client = bufstate.client_hints
+  local client_hints = bufstate.client_hints
   local client = assert(vim.lsp.get_client_by_id(client_id))
 
-  local new_hints_by_lnum = vim.defaulttable()
+  local new_lnum_hints = vim.defaulttable()
```
justinmk, line 47: what is "new", the lnum or the hints? if the hints are new, should this be named `lnum_new_hints` ?

Issue #28485, opened 2024-04-24
vim.lsp.rpc.connect doesn't works on windows


pogyomo:
### Problem

When I try to start to write gdscript code which is used in Godot Engine,  I found the engine provide a default language server and can communicate it on port `6005`, so I wrote config for it using `vim.lsp.start` and `vim.lsp.rpc.connect` as follow.
```lua
vim.api.nvim_create_autocmd("FileType", {
    pattern = { "gd", "gdscript", "gdscript3" },
    group = vim.api.nvim_create_augroup("gdscript-lsp-start", {}),
    callback = function()
        vim.lsp.start {
            name = "gdscript",
            cmd = vim.lsp.rpc.connect("127.0.0.1", 6005),
            capabilities = capabilities,
            root_dir = vim.fs.dirname(vim.fs.find({ "project.godot", ".git" }, { upward = true })[1]),
        }
    end,
})
```
But the neovim doesn't attach client and so that `vim.lsp.get_clients` returns empty list. Afterthat, I found `cmd` accept a command and so that rewrite config using `ncat` command.
```lua
vim.api.nvim_create_autocmd("FileType", {
    pattern = { "gd", "gdscript", "gdscript3" },
    group = vim.api.nvim_create_augroup("gdscript-lsp-start", {}),
    callback = function()
        vim.lsp.start {
            name = "gdscript",
            cmd = { "ncat", "127.0.0.1", "6005" },
            capabilities = capabilities,
            root_dir = vim.fs.dirname(vim.fs.find({ "project.godot", ".git" }, { upward = true })[1]),
        }
    end,
})
```
Unlike when using `vim.lsp.rpc.connect`, this works well and `vim.lsp.get_clients` returns the client infomation.

### Steps to reproduce

I don't know any other lsp which can communicate with tcp, so I just introduce the procedure using Godot Engine.
1. create a `init.lua` with following content
```lua
vim.api.nvim_create_autocmd("FileType", {
    pattern = { "gd", "gdscript", "gdscript3" },
    group = vim.api.nvim_create_augroup("gdscript-lsp-start", {}),
    callback = function()
        vim.lsp.start {
            name = "gdscript",
            cmd = vim.lsp.rpc.connect("127.0.0.1", 6005),
            capabilities = capabilities,
            root_dir = vim.fs.dirname(vim.fs.find({ "project.godot", ".git" }, { upward = true })[1]),
        }
    end,
})
```
[...]

PR #28507, opened 2024-04-25
fix(lsp): buffer messages until connected to server

mfussenegger:
`handle:write(msg)` can fail if the socket is not yet connected to the
server.

Should address https://github.com/neovim/neovim/pull/28398#issuecomment-2078152491


mfussenegger:
It would probably be cleaner to solve this via interface changes that make the full chain async, but I'd rather avoid that this close to 0.10

TheLeoP:
Yup, this solves my problem. Thanks @mfussenegger ! This should also solve #28485

PR #28511, opened 2024-04-26
docs: update ubuntu dependencies

divanvisagie:
On a fresh install of Ubuntu 24.04 LTS, make install will not run without this extra lib included

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -291,7 +291,7 @@ Platform-specific requirements are listed below.
 ### Ubuntu / Debian
 
 ```sh
-sudo apt-get install ninja-build gettext cmake unzip curl build-essential
+sudo apt-get install ninja-build gettext cmake unzip curl build-essential libnsl-dev
 ```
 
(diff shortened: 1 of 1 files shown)


dundargoc:
I cannot recreate this with docker. What error are you getting?

divanvisagie:
> I cannot recreate this with docker. What error are you getting?

Hmm, interesting. So the way I discovered this was that I have been doing a lot of ubuntu desktop installs lately and ran into this each time. 

I think there might be a difference here between ubuntu server and ubuntu desktop where the one may have this package pre-installed for whatever reason.

Another issue might be that I am doing the 'minimal' install of ubuntu desktop, which means it might not be installing packages that may depend on this one and thus the package is possibly available on the 'full' installation.

I will set up some VMs this weekend and see if I can get some more hard evidence.

divanvisagie:
Alright, tried it on a clean install and it works without my suggestion, welp, that was weird. Sorry for the inconvenience.

PR #28523, opened 2024-04-26
feat(lsp): use a parameter strategy similar to `enable`

ofseed:
As we discussed in https://github.com/neovim/neovim/pull/28521#discussion_r1581156758

Before this, `vim.diagnostic.enable` and `vim.diagnostic.is_enabled()` use the same strategy.

### DIFF
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -364,12 +364,29 @@ api.nvim_set_decoration_provider(namespace, {
   end,
 })
 
---- @param bufnr (integer|nil) Buffer handle, or 0 for current
+--- @param filter vim.lsp.inlay_hint.enable.Filter
 --- @return boolean
 --- @since 12
-function M.is_enabled(bufnr)
+function M.is_enabled(filter)
+  ---@type integer
+  local bufnr
+  if type(filter) == 'number' then
+    vim.deprecate(
+      'vim.lsp.inlay_hint.is_enabled(bufnr:number)',
+      'vim.lsp.inlay_hint.is_enabled(filter:table)',
+      '0.10-dev'
+    )
+    bufnr = filter
+  else
+    vim.validate({ filter = { filter, 'table', true } })
+    filter = filter or {}
+    bufnr = filter.bufnr
+  end
+
   vim.validate({ bufnr = { bufnr, 'number', true } })
-  if bufnr == nil or bufnr == 0 then
+  if bufnr == nil then
+    return globalstate.enabled
+  elseif bufnr == 0 then
     bufnr = api.nvim_get_current_buf()
   end
   return bufstates[bufnr].enabled
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1645,13 +1645,15 @@ get({filter})                                       *vim.lsp.inlay_hint.get()*
         • {client_id} (`integer`)
         • {inlay_hint} (`lsp.InlayHint`)
 
-is_enabled({bufnr})                          *vim.lsp.inlay_hint.is_enabled()*
+is_enabled({filter})                         *vim.lsp.inlay_hint.is_enabled()*
 
     Note: ~
       • This API is pre-release (unstable).
 
     Parameters: ~
-      • {bufnr}  (`integer?`) Buffer handle, or 0 for current
+      • {filter}  (`table`) Optional filters |kwargs|, or `nil` for all.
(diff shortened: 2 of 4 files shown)


runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -353,11 +353,25 @@ api.nvim_set_decoration_provider(namespace, {
...
+    )
+    bufnr = filter
+  else
+    vim.validate({ filter = { filter, 'table', true } })
+    filter = filter or {}
+    bufnr = filter.bufnr
+  end
+
   if bufnr == nil or bufnr == 0 then
     bufnr = api.nvim_get_current_buf()
```
justinmk, line 376: This contradicts the docs which say `is_enabled(nil)` means "all". I side-stepped this in the previous round of changes, because it's not clear to me if that can be meaningfully supported. If you can take a closer look and see if we can support that, that will be very helpful :)

ofseed: Currently `vim.lsp.inlay_hint.enable` does not support directly enabling for all buffers. Even if the `bufnr` passed to it is `nil`, it is only effective for the current buffer. This means that `is_enable(nil)` cannot mean "all" also.

Do you mean to make `vim.lsp.inlay_hint(true, { bufnr = nil})` effective for all buffers? Although I am willing to implement this feature, as we discussed in https://github.com/neovim/neovim/pull/28494#issuecomment-2076723683, it is now close to the release of 0.10, so it is better not to make more risky new features now. Simply put, the current code relies on creating a series of autocmds when calling `enable` on a `bufnr`, so the changes required to implement this feature are no less than https://github.com/neovim/neovim/pull/28521, and of course, we need new tests for this.

The reason I want to merge this PR before the 0.10 release is to reduce the impact on stable version users if https://github.com/neovim/neovim/pull/28521 is merged after 0.10, and no matter what changes will be made in the future, I think the parameter of `is_enable` will also change to accept `filter ` instead of receiving `bufnr` directly. 

Before this PR, `is_enable(nil)` didn't mean `all` either, so I think the question is a bit beyond the scope of this PR. For now, I tend to change the documentation to what it is now, rather than implement what the current documentation says it does.

justinmk: > Do you mean to make `vim.lsp.inlay_hint(true, { bufnr = nil})` effective for all buffers?

ideally, yes.



> as we discussed in [#28494 (comment)](https://github.com/neovim/neovim/pull/28494#issuecomment-2076723683), it is now close to the release of 0.10, so it is better not to make more risky new features now

this feature is unreleased. there is no better time than now.



> The reason I want to merge this PR before the 0.10 release is to reduce the impact on stable version users if #28521 is merged after 0.10, and no matter what changes will be made in the future, I think the parameter of `is_enable` will also change to accept `filter ` instead of receiving `bufnr` directly.

agree, filter param is probably the right approach for future-proofing.

ofseed: OK, I created https://github.com/neovim/neovim/pull/28543.

ofseed:
In case I forget, although I introduced a `globalstate.enabled` in #28543, I'm now sure whether we should return it when calling `is_enabled(nil)`. In most cases, people care about whether inlay hints are enabled on the current buffer instead of globally, which may contradict their intuition.

Most importantly, I found even `vim.diagnostic.enable(nil)` returns the current value.
https://github.com/neovim/neovim/blob/a41546d4ac60cecc48b83f85ea9d53ee2fb7dfbb/runtime/lua/vim/diagnostic.lua#L1576-L1576
https://github.com/neovim/neovim/blob/a41546d4ac60cecc48b83f85ea9d53ee2fb7dfbb/runtime/lua/vim/diagnostic.lua#L549-L554
Logic similar to `resolve_bufnr` is still used here.

ofseed:
Now the PR should be ready to be merged.

justinmk (approved):

justinmk:
@ofseed something seems wrong here. With latest master, if I visit a C file with clangd LSP:

```lua
vim.lsp.inlay_hint.is_enabled()  -- false (this is expected ✅)
vim.lsp.inlay_hint.enable()  -- enable globally
vim.lsp.inlay_hint.is_enabled()  -- true (ok ✅)
vim.lsp.inlay_hint.is_enabled({bufnr=0}) -- true (ok ✅), but inlay hints are not shown (wrong ❌)
vim.lsp.inlay_hint.enable(true, {bufnr=0}) -- now inline hints are shown
```

PR #28527, opened 2024-04-26
feat: add "default" field to maparg() return value

gpanders:
With so many default mappings, plugins are not able to detect if an existing mapping is one created by the user or one created by Nvim.

Add a "default" field to the dict return value of maparg() that plugins can use to determine if a mapping is a default mapping provided by Nvim (which they can safely override).

### DIFF
--- a/src/nvim/mapping.c
+++ b/src/nvim/mapping.c
@@ -132,6 +132,16 @@ static const char e_entries_missing_in_mapset_dict_argument[]
 static const char e_illegal_map_mode_string_str[]
   = N_("E1276: Illegal map mode string: '%s'");
 
+/// Whether new mappings should be marked as "default". This is modified only in map_set_default().
+static bool create_default_maps;
+
+/// Enable or disable creationg of default mappings.
+///
+/// @param value True if new mappings should be marked as default mappings.
+void map_set_default(bool value) {
+  create_default_maps = value;
+}
+
 /// Get the start of the hashed map list for "state" and first character "c".
 mapblock_T *get_maphash_list(int state, int c)
 {
@@ -540,6 +550,8 @@ static void map_add(buf_T *buf, mapblock_T **map_table, mapblock_T **abbr_table,
   if (args->desc != NULL) {
     mp->m_desc = xstrdup(args->desc);
   }
+
+  mp->m_default = create_default_maps;
 
   // add the new entry in front of the abbrlist or maphash[] list
   if (is_abbr) {
@@ -2118,6 +2130,7 @@ static Dictionary mapblock_fill_dict(const mapblock_T *const mp, const char *lhs
   PUT_C(dict, "mode", CSTR_AS_OBJ(mapmode));
   PUT_C(dict, "abbr", INTEGER_OBJ(abbr ? 1 : 0));
   PUT_C(dict, "mode_bits", INTEGER_OBJ(mp->m_mode));
+  PUT_C(dict, "default", BOOLEAN_OBJ(mp->m_default));
 
   return dict;
 }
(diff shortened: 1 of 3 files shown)


dundargoc (approved):

src/nvim/mapping.c
```diff
@@ -132,6 +132,16 @@ static const char e_entries_missing_in_mapset_dict_argument[]
 static const char e_illegal_map_mode_string_str[]
   = N_("E1276: Illegal map mode string: '%s'");
 
+/// Whether new mappings should be marked as "default". This is modified only in map_set_default().
+static bool create_default_maps;
+
+/// Enable or disable creationg of default mappings.
```
zeertzjq, line 138: Is this comment accurate?

justinmk:
> plugins are not able to detect if an existing mapping is one created by the user or one created by Nvim.

The reason for choosing to set default mappings instead of creating builtins was because this shouldn't matter. Builtin default mappings are just like any other plugin default mappings.

> plugins are not able to detect if an existing mapping is one created by the user or one created by Nvim.

And what would the plugin do with that information? Users can `:mapclear` if they don't want default mappings. Then plugins won't see any non-user-defined mappings.

gpanders:
> > plugins are not able to detect if an existing mapping is one created by the user or one created by Nvim.
> 
> The reason for choosing to set default mappings instead of creating builtins was because this shouldn't matter.

How could it _not_ matter? A mapping created by the user and a mapping that comes bundled with Neovim are different. Well-behaved plugins use `maparg()` to ensure they don't clobber a user's mapping, but they're free to override over builtins/defaults.

> > plugins are not able to detect if an existing mapping is one created by the user or one created by Nvim.
> 
> And what would the plugin do with that information? Users can `:mapclear` if they don't want default mappings. Then plugins won't see any non-user-defined mappings.

Plugins use this information to determine if a mapping was set by the user or not. A plugin can override a default Nvim mapping harmlessly, but a plugin should not override a user's own mapping.

src/nvim/mapping.c
```diff
@@ -2118,6 +2130,7 @@ static Dictionary mapblock_fill_dict(const mapblock_T *const mp, const char *lhs
   PUT_C(dict, "mode", CSTR_AS_OBJ(mapmode));
   PUT_C(dict, "abbr", INTEGER_OBJ(abbr ? 1 : 0));
   PUT_C(dict, "mode_bits", INTEGER_OBJ(mp->m_mode));
+  PUT_C(dict, "default", BOOLEAN_OBJ(mp->m_default));
```
zeertzjq, line 2133: It may be a bit inconsistent for this to be a boolean as there are no other boolean fields. For consistency this should be an integer, but that's inconvenient for Lua plugins. Since this is mostly likely to be used by Lua plugins, maybe it should be like this:
```suggestion
  if (mp->m_default) {
    PUT_C(dict, "default", INTEGER_OBJ(1));
  }
```
In future it may also be possible for the `default` field to have a different type (e.g. string) to provide more information, in which case a missing `default` field indicating a non-default mapping makes more sense.

zeertzjq (commented):
This may not be a good idea. For default autocommands their augroups have a common `nvim_` prefix. For mappings there is a proposal for namespaced mappings. If namespaced mappings are implemented, default mappings can be in namespaces with specific names, and this field will become redundant again.

gpanders:
> This may not be a good idea. For default autocommands their augroups have a common `nvim_` prefix. For mappings there is a proposal for namespaced mappings. If namespaced mappings are implemented, default mappings can be in namespaces with specific names, and this field will become redundant again.

Even with namespaces, we would need a way to designate that a mapping is "bundled" with Nvim, since anybody (including a plugin) could just add a new mapping to the "defaults" namespace.

src/nvim/mapping.c
gpanders: Changed as suggested d72a4beebea0c6d658d03bdcd8352ccdfcde2611.

justinmk:
what about other default mappings, e.g. `%` ?

gpanders:
Will close this since there doesn't seem to be an acute need anymore. Can re-open in the future or make a new PR if this comes up again.

Issue #25514, opened 2023-10-05
vim.ui enhancements: proposal and tracking issue


gpanders:
This ticket specifically tracks enhancements and modifications to `vim.ui` proposed in #25272 and related issues.

# Problem

Creating floating windows with all of the visual goodies that most users expect and want requires a lot of boilerplate. This boilerplate has coalesced into `vim.lsp.util.open_floating_preview`, but such functionality should be both (1) pruned down and curated into a more thoughtful and coherent interface, and (2) moved out of `vim.lsp.util` and into `vim.ui`.

# Solution

Create `vim.ui.open_win` that handles the use case of creating windows of all kinds in an ergonomic way.

## Requirements

* `vim.ui.open_win` should be able to open floating windows as well as splits (horizontal and vertical) and tabs.
* `vim.ui.open_win` should substantially reduce boilerplate for creating and opening floating windows.

## Proposed interface

```lua
--- @param contents string[]|int An array of strings containing the lines
---                              of text to display in the window OR an
---                              integer representing a buffer number to
---                              display in the window
--- @param opts table Options table (see details below)
--- @return number |winid| of new window
vim.ui.open_win(contents, {
  -- Opts table is a "sum type" / "union", the type of window it
  -- creates is based on the top-level key provided. It as an error
  -- to provide more than one top-level key
  float = {
    -- Create a floating window. All keys from |nvim_open_win()| are valid in this table.
    -- Omitted keys use default values from |vim.ui.config()|.
    height = 100,
    width = 50,
  },

  split = {
    -- Create a split
    vertical = true | false, -- defaults to false
  },

  tab = {
    -- Create a new tab
  }
})
```

In addition, create `vim.ui.config()` that sets default values for some options. This should work similarly to `vim.diagnostic.config()`. Plugin authors can use `vim.ui.open_win()` and omit values to use the defaults supplied by the user in `vim.ui.config()`:

```lua
vim.ui.config({
  float = {
    border = "rounded",
  },
  split = {
    -- Prefer vertical splits by default
    vertical = true,
  },
})
```

`vim.ui.config()` is not limited to configuring options for `vim.ui.open_win()`, but covers any user-definable default/preference for anything under `vim.ui`.

mfussenegger:
> > Any reason for a vim.ui.config table instead of going with new regular vim options?
> 
> Because we try to keep things in Lua if possible (and do you really want to touch Vim options? especially window options? especially global-local window options?)
> 
[...]

PR #28531, opened 2024-04-27
fix(diagnostic): get border from config

glepnir:
Problem: border value is hardcoded

Solution: get it from config interface

### DIFF
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -499,6 +499,7 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
                           diagnostic instead of prepending it. Overrides the
                           setting from |vim.diagnostic.config()|.
       • {focus_id}?       (`string`)
+      • {border}?         (`string`) see |vim.api.nvim_open_win|.
 
 *vim.diagnostic.Opts.Signs*
 
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -499,7 +499,7 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
                           diagnostic instead of prepending it. Overrides the
                           setting from |vim.diagnostic.config()|.
       • {focus_id}?       (`string`)
-      • {border}?         (`string`) see |vim.api.nvim_open_win|.
+      • {border}?         (`string`) see |vim.api.nvim_open_win()|.
 
 *vim.diagnostic.Opts.Signs*
 
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -190,7 +190,7 @@ do
     })
 
     vim.keymap.set('n', '<C-W>d', function()
-      vim.diagnostic.open_float({ border = 'rounded' })
+      vim.diagnostic.open_float()
     end, {
       desc = 'Open a floating window showing diagnostics under the cursor',
     })
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -153,7 +153,7 @@ local M = {}
 ---
 --- @field focus_id? string
 ---
---- @field border? string see |vim.api.nvim_open_win|.
+--- @field border? string see |vim.api.nvim_open_win()|.
 
 --- @class vim.diagnostic.Opts.Underline
 ---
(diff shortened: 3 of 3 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -152,6 +152,8 @@ local M = {}
 --- @field suffix? string|table|(fun(diagnostic:vim.Diagnostic,i:integer,total:integer): string, string)
 ---
 --- @field focus_id? string
+---
+--- @field border? string
```
justinmk, line 156: is there a generic float class that `vim.diagnostic.Opts.Float` can inherit from?

runtime/lua/vim/_defaults.lua
```diff
@@ -190,7 +190,10 @@ do
     })
 
     vim.keymap.set('n', '<C-W>d', function()
-      vim.diagnostic.open_float({ border = 'rounded' })
+      local config = vim.diagnostic.config()
+      local border = type(config.float) == 'table' and (config.float.border or 'rounded')
+        or 'rounded'
```
justinmk, line 195: could use `vim.tbl_get`.  

side-note:  should `vim.tbl_get` allow non-table input? that would make it more flexible, but less strict.

```suggestion
      local border = vim.tbl_get(config, 'float', 'border') or 'rounded'
```

glepnir: forget some util functions someitme .......

runtime/lua/vim/diagnostic.lua
glepnir: just search float in help doc seems like we don't have a float class  ?

justinmk: yeah i don't see one in `runtime/lua/vim/_meta/api.lua`

mfussenegger:
Would it be an option to have a global `float_border` option? The border is also a frequent config option for plugins and in some LSP methods. It would be great if a user would only have to configure it once instead of per module/functionality to have consistent borders

justinmk:
> Would it be an option to have a global `float_border` option?

a 'winoptions' or 'winstyle' option is probably needed at some point. 

but this PR seems good meanwhile, because this mapping is for diagnostics specifically.

gpanders:
> Would it be an option to have a global `float_border` option? The border is also a frequent config option for plugins and in some LSP methods. It would be great if a user would only have to configure it once instead of per module/functionality to have consistent borders

Related: `vim.ui.config()` https://github.com/neovim/neovim/issues/25514

runtime/lua/vim/_defaults.lua
```diff
@@ -190,7 +190,9 @@ do
     })
 
     vim.keymap.set('n', '<C-W>d', function()
-      vim.diagnostic.open_float({ border = 'rounded' })
+      local config = vim.diagnostic.config()
+      local border = vim.tbl_get(config, 'float', 'border') or 'rounded'
+      vim.diagnostic.open_float({ border = border })
```
gpanders, line 195:
```suggestion
      vim.diagnostic.open_float()
```

justinmk (approved):

runtime/doc/diagnostic.txt
```diff
@@ -499,6 +499,7 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
                           diagnostic instead of prepending it. Overrides the
                           setting from |vim.diagnostic.config()|.
       • {focus_id}?       (`string`)
+      • {border}?         (`string`)
```
gpanders, line 502: Can we link to nvim_open_win so that users know what valid values for this option are?

glepnir: before i tried add a see section. but i don't saw it in hover idk.

runtime/doc/diagnostic.txt
```diff
@@ -499,6 +499,7 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
                           diagnostic instead of prepending it. Overrides the
                           setting from |vim.diagnostic.config()|.
       • {focus_id}?       (`string`)
+      • {border}?         (`string`) see |vim.api.nvim_open_win|.
```
justinmk, line 502: not a valid tag. that's why docs ci is failing
```suggestion
      • {border}?         (`string`) see |vim.api.nvim_open_win()|.
```

gpanders (approved):

PR #28532, opened 2024-04-27
refactor(build): make all generated c files headers

bfredl:
There's no "rule" or bad practice or whatever that says we cannot generate c files. it is is just that we have ~20 generated headers and ~2 generated sources and there is nothing in these two generated source files which sets them apart. Lua bindings are not different from rpc bindings, and pathdef is not different from versiondef.

So to simplify build logic and ease the future port to build.zig, streamline the build to only have generated headers, no direct generated .c files.

Also "nlua_add_api_functions" had its prototype duplicated twice which defeated the point of having mandatory prototypes (one source of truth).


src/nvim/generators/gen_api_dispatch.lua
```diff
@@ -717,21 +717,6 @@ end
 output = assert(io.open(lua_c_bindings_outputf, 'wb'))
 
 output:write([[
```
dundargoc, line 719: Just sanity checking, this empty `write` is intentional I assume?

dundargoc (approved):

justinmk (approved):

PR #28537, opened 2024-04-27
feat(defaults): visual CTRL-R for LSP mappings

### LINKED ISSUE #28528: New lsp keybindings: `crr` in visual mode breaks visual selection -> `c` workflow
### Problem

The new lsp keybindings include `crr` in visual mode for code actions. This breaks the visual mode -> `c` workflow because the user now has to wait for `timeoutlen`.

### Steps to reproduce

After a server has attached, select some text, then press `c`

### Expected behavior

Selected text is gone and nvim is in insert mode immediately

### Neovim version (nvim -v)

0.10.0 commit 9b8a0755390b7eb3ad369f3a0a42eb9aecb8cbe2

no

### Operating system/version

N/A

### Terminal name/version

N/A

N/A

### Installation

N/A

justinmk:
fix #28528

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -61,10 +61,12 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                                  *crr* *v_crr* *crn* *i_CTRL-S*
+                 *crr* *v_crr* *crn* *i_CTRL-S* *v_CTRL-R_CTRL-R* *v_CTRL-R_r*
 Some keymaps are created unconditionally when Nvim starts:
 - "crn" is mapped in Normal mode to |vim.lsp.buf.rename()|
-- "crr" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
+- "crr" is mapped in Normal mode to |vim.lsp.buf.code_action()|
+- CTRL-R CTRL-R (also "CTRL-R r") is mapped in Visual mode to
+  |vim.lsp.buf.code_action()|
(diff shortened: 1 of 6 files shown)


runtime/lua/vim/_defaults.lua
```diff
@@ -159,7 +159,16 @@ do
       vim.lsp.buf.rename()
     end, { desc = 'vim.lsp.buf.rename()' })
 
-    vim.keymap.set({ 'n', 'v' }, 'crr', function()
```
justinmk, line 162: was 'v' instead of 'x' intentional?

gpanders: No

mfussenegger (approved):
Personally I'd prefer a symmetric mapping, but the rationale behind this makes sense, and I don't have a better sugestion either, so :+1:

runtime/lua/vim/_defaults.lua
echasnovski: Besides, `<C-r>` in Select mode is a thing, so using `'x'` mode is probably safer.

gpanders (approved):

runtime/doc/lsp.txt
```diff
@@ -61,10 +61,12 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                                  *crr* *v_crr* *crn* *i_CTRL-S*
+                 *crr* *v_crr* *crn* *i_CTRL-S* *v_CTRL-R_CTRL-R* *v_CTRL-R_r*
```
kang8, line 64: It seems that **v_crr** needs to be removed?

```suggestion
                 *crr* *crn* *i_CTRL-S* *v_CTRL-R_CTRL-R* *v_CTRL-R_r*
```

PR #28538, opened 2024-04-27
fix(snippet): do not add extra indent on newlines

mfussenegger:
Reverts parts of https://github.com/neovim/neovim/pull/27674

LSP snippets typically do include tabs or spaces to add extra
indentation and don't rely on the client using `autoindent`
functionality.

For example:

    public static void main(String[] args) {\n\t${0}\n}

Notice the `\t` after `{\n`

Adding spaces or tabs independent of that breaks snippets for languages
like Haskell where you can have snippets like:

    ${1:name} :: ${2}\n${1:name} ${3}= ${0:undefined}

To generate:

    name ::
    name = undefined

### DIFF
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -459,8 +459,7 @@ function M.expand(input)
       end
       -- Add the base indentation.
       if i > 1 then
-        line = #line ~= 0 and base_indent .. line
-          or (expandtab and (' '):rep(shiftwidth) or '\t'):rep(vim.fn.indent('.') / shiftwidth + 1)
+        line = base_indent .. line
       end
       lines[#lines + 1] = line
     end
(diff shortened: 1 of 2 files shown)


MariaSolOs (approved):

MariaSolOs:
I guess this will undo whatever bug (if any?) @glepnir tried to fix in #27674, so just a heads up.

glepnir:
maybe try to fix not revert.

mfussenegger:
> maybe try to fix not revert. newline should add indent this is all the snippet plugins do. no indent will very confuse me on there ...

Can you explain what's broken that would need fixing?

Luasnip also expands `function($2)\n$3\nend` without indentation in the second line.

glepnir:
oh yup i am missing that mode is 2 is what i want and clangd not return textmode. thanks for info

PR #28539, opened 2024-04-27
docs(lsp): update override mappings example

MariaSolOs:
Updating the docs to include all default keybindings in the "override" example.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -79,6 +79,10 @@ To override the above defaults, set or unset the options on |LspAttach|: >lua
         vim.bo[ev.buf].formatexpr = nil
         vim.bo[ev.buf].omnifunc = nil
         vim.keymap.del('n', 'K', { buffer = ev.buf })
+        vim.keymap.del('n', 'crn', { buffer = ev.buf })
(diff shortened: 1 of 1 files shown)


runtime/doc/lsp.txt
```diff
@@ -79,6 +79,10 @@ To override the above defaults, set or unset the options on |LspAttach|: >lua
         vim.bo[ev.buf].formatexpr = nil
         vim.bo[ev.buf].omnifunc = nil
         vim.keymap.del('n', 'K', { buffer = ev.buf })
+        vim.keymap.del('n', 'crn', { buffer = ev.buf })
```
gpanders, line 82: These aren’t buffer local mappings, and they’re not created when the LSP client is attached. They’re created unconditionally in _defaults.lua, so it doesn’t make sense to delete them in LspAttach.

There is already a blurb just above this mentioning to use vim.keymap.del() or :unmap to remove these. Any reason that is insufficient? Can that text be clarified more?

MariaSolOs: I see. That's okay, from your comment in https://github.com/neovim/neovim/pull/28500#discussion_r1578790094 I thought that you had forgotten to update the example or that the change got lost with later force pushes. I'm fine with this as is.

PR #28540, opened 2024-04-27
build: simplify policy-setting for dependencies

dundargoc:
Passing `CMAKE_POLICY_DEFAULT_CMP0092=NEW` to all dependencies is
simpler than setting it manually in each file.


dundargoc:
TODO: this make the results somehow worse from the looks of gettext logs?

PR #16230, opened 2021-11-03
feat(diagnostic): add default mappings for diagnostics

gpanders:
Add mappings for `goto_next` and `goto_prev`.

By default, disables the floating window. Considering `virtual_text` is enabled by default, having the float feels very noisy to me. Happy to be convinced otherwise. I'd also be happy to disable `virtual_text` by default and enable float popups for these maps.

PR #28542, opened 2024-04-27
fix(defaults): do not hardcode borders in diagnostic.open_float

antonk52:
As per [discussion](https://github.com/neovim/neovim/pull/16230#discussion_r1581643975) in #16230


justinmk:
https://github.com/neovim/neovim/pull/28531

PR #28543, opened 2024-04-28
feat(lsp): vim.lsp.inlay_hint.enable(nil) applies to all buffers

ofseed:
**Problem**:

The current implementation of inlay hints contradicts our dev guidelines.
https://github.com/neovim/neovim/blob/a41546d4ac60cecc48b83f85ea9d53ee2fb7dfbb/runtime/doc/develop.txt#L313-L317

**Solution**:

Implement globally enabling inlay hints.


ofseed:
I keep the commit history as clean as possible to make it easier to review. Thereby reducing the risk of merging this PR.

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -4,13 +4,32 @@ local ms = require('vim.lsp.protocol').Methods
 local api = vim.api
 local M = {}
 
----@class (private) vim.lsp.inlay_hint.bufstate
+---@class (private) vim.lsp.inlay_hint.globalstate Global state for inlay hints
+---@field enabled boolean Whether inlay hints are enabled for this scope
```
MariaSolOs, line 8: Is the scope always the buffer?

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -34,7 +53,7 @@ function M.on_inlayhint(err, result, ctx, _)
     return
   end
   local bufstate = bufstates[bufnr]
-  if not bufstate or not bufstate.enabled then
+  if not bufstate.enabled then
```
MariaSolOs, line 56: Why did you remove the `not bufstate` condition here?

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -154,7 +169,7 @@ function M.get(filter)
   end
 
   local bufstate = bufstates[bufnr]
-  if not (bufstate and bufstate.client_hints) then
+  if not bufstate.client_hints then
```
MariaSolOs, line 172: Again: Why not check that `bufstate` isn't nil first?

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -206,9 +221,6 @@ local function clear(bufnr)
   if bufnr == nil or bufnr == 0 then
```
MariaSolOs, line 221: We're not allowing `bufnr` to be nil anymore though right?

ofseed: Thanks, just forgot to remove the judgment here

runtime/lua/vim/lsp/inlay_hint.lua
ofseed: Because I changed `bufstates` to a `vim.defaulttable`, which makes every time the indexed `bufstates[bufnr]` does not exist, it will create a new `bufstate` according to the factory method, which makes `bufstates[bufnr] ~= nil` will no longer make sense because it will always be `true`.

runtime/lua/vim/lsp/inlay_hint.lua
ofseed: It could be enabled globally because we do not only expect to be enabled on buffers that exist now but also for the buffers that will be created in the future.

ofseed:
@MariaSolOs More explanations for why make `bufstates` as a `vim.defaulttable`:

1. To make `bufstate`s inherit values from the `globalstate`, we need `setmetatable` for them. Instead of defining a factory function for `bufstate`, I think creating them when indexing `bufstates` will be simpler and more effective, this approach also avoids thinking about when this factory function should be called.
2. As you can see we can avoid a lot of judgment for whether the `bufstates[bufnr]` exists.

But now I admit that this idea is not perfect. The `on_win` callback of the `nvim_set_decoration_provider` is called too trivially, if the `vim.defaulttable` takes effect here, there will be many tables in the `bufstates`. Even a temporary window like a completion candidate window of `nvim-cmp`, will have its own `bufstate` in `bufstate`. Although this won't produce a fatal error, I think it's better to avoid it. In fact, all my changes since the opening of this PR have been to optimize this problem.

Currently, to avoid this situation, I made `on_win` cannot trigger the factory function when `bufstates[bufnr]` does not exist by utilizing `rawget` instead of indexing directly inside it.

MariaSolOs:
@ofseed thanks for explaining! I must confess that that level of manipulation of metatables is beyond my expertise, so I'll let others comment on the correctness of the implementation.

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -4,13 +4,32 @@ local ms = require('vim.lsp.protocol').Methods
...
+    __index = globalstate,
+    __newindex = function(state, key, value)
+      if globalstate[key] == value then
+        rawset(state, key, nil)
+      else
+        rawset(state, key, value)
+      end
+    end,
+  })
+  return bufstate
```
famiu, line 31:
```suggestion
  return setmetatable({ applied = {} }, {
    __index = globalstate,
    __newindex = function(state, key, value)
      if globalstate[key] == value then
        rawset(state, key, nil)
      else
        rawset(state, key, value)
      end
    end,
  })
```

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -4,13 +4,32 @@ local ms = require('vim.lsp.protocol').Methods
...
----@field enabled boolean Whether inlay hints are enabled for this buffer
 ---@type table<integer, vim.lsp.inlay_hint.bufstate>
-local bufstates = {}
+local bufstates = vim.defaulttable(function(_)
+  local bufstate = { applied = {} }
+  setmetatable(bufstate, {
+    __index = globalstate,
+    __newindex = function(state, key, value)
+      if globalstate[key] == value then
+        rawset(state, key, nil)
```
famiu, line 25: Is this not problematic? If you explicitly bufstate to a value equal to globalstate, and then change the value, it will change the value for the buffer even though it already has a local value

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -392,11 +409,28 @@ function M.enable(enable, filter)
...
   vim.validate({ enable = { enable, 'boolean', true }, filter = { filter, 'table', true } })
+  enable = enable == nil or enable
   filter = filter or {}
-  if enable == false then
-    _disable(filter.bufnr)
+
+  if filter.bufnr == nil then
+    globalstate.enabled = enable
+    for bufnr, _ in pairs(bufstates) do
+      if api.nvim_buf_is_valid(bufnr) then
```
famiu, line 418: Are we sure this should use valid buffers instead of loaded buffers?

runtime/lua/vim/lsp/inlay_hint.lua
ofseed: There doesn't seem to be a clear convention for what to do in this situation, so I deliberately designed it like this.

>If you explicitly bufstate to a value equal to globalstate, and then change the value of globalstate, it will change the value for the buffer even though it already is supposed to have a local value

So, this is what I want to achieve.

If we expose variables to users, the situation is simple, as long as the local variable exists, we use the local variable; otherwise we use the global variable. This is very similar to what you expected.

But if the API we exposed is functions like `enable`, It’s hard to say whether the above approach is correct or not. Because there is no way to remove "local variables" by using `enable`. Once the user manages a buffer's `bufstate`, it means that they must manage it all the time, there's no way to make them follow the `globalstate`.

The logic behind this design is that if it can be globally `enable` or locally `enable`, when the user selects local `enable`, we think the user's purpose is to distinguish the local state from the global state. I think this design is good for interaction,

You can think about whether what you said is similar to what I said. When the user explicitly sets the previously unequal `bufstate` to be equal to `globalstate`, he is probably thinking "this buffer is no longer special" rather than "no matter what the `globalstate` is, this `bufstate` should be like this."

famiu: I'm still not fully convinced. But I'll refer to @justinmk as he's the best person to ask for this kind of thing

runtime/lua/vim/lsp/inlay_hint.lua
glepnir: simply ternary
```lua
return setmetatable({applied = {}}, {
    __index = globalstate,
    __newindex = function(state, key, value)
        rawset(state, key, (globalstate[key] == value) and nil or value)
    end
})
```

famiu: > simply ternary
> 
> ```lua
> return setmetatable({applied = {}}, {
>     __index = globalstate,
>     __newindex = function(state, key, value)
>         rawset(state, key, (globalstate[key] == value) and nil or value)
>     end
> })
> ```

This is incorrect. `nil or value` will always return `value`, the `nil` value will never be chosen

ofseed: I'd say I write this in such a verbose way is I want to emphasize that the `bufstates` is an array of `bufstate`s, and we can only get `bufstate` from it, not anything else. Maybe the effect is not very obvious? I might change it.

runtime/doc/lsp.txt
```diff
@@ -1611,7 +1611,7 @@ enable({enable}, {filter})                       *vim.lsp.inlay_hint.enable()*
     Parameters: ~
       • {enable}  (`boolean?`) true/nil to enable, false to disable
       • {filter}  (`table?`) Optional filters |kwargs|, or `nil` for all.
-                  • {bufnr} (`integer?`) Buffer number, or 0/nil for current
+                  • {bufnr} (`integer?`) Buffer number, or 0 for current
                     buffer.
```
justinmk, line 1615:
```suggestion
                    buffer, or nil for all.
```

runtime/lua/vim/lsp/inlay_hint.lua
justinmk: buffer-local should always override global. the global should act as the "default" in the absence of a buffer-local setting. this is well-established by e.g. buffer-local options.

it sounds like that's what's implemented here, correct?

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -4,13 +4,32 @@ local ms = require('vim.lsp.protocol').Methods
...
+}
+
+---@class (private) vim.lsp.inlay_hint.bufstate: vim.lsp.inlay_hint.globalstate Buffer local state for inlay hints
 ---@field version? integer
 ---@field client_hints? table<integer, table<integer, lsp.InlayHint[]>> client_id -> (lnum -> hints)
 ---@field applied table<integer, integer> Last version of hints applied to this line
----@field enabled boolean Whether inlay hints are enabled for this buffer
 ---@type table<integer, vim.lsp.inlay_hint.bufstate>
-local bufstates = {}
+local bufstates = vim.defaulttable(function(_)
```
justinmk, line 19: off topic for this PR: 

Usually the correct and strongly preferred way to store buffer-local state is to use a buffer-local variable. Then if a buffer is deleted, its state is also deleted, without needing any lifecycle management. So this `bufstates` pattern should be avoided.

Since you're looking at this code, do you see any reason why these flags can't be stored as a `vim.b` (`b:`) variable? (That change should be a separate PR, if we decide to explore it.)

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -392,11 +409,28 @@ function M.enable(enable, filter)
...
+
+  if filter.bufnr == nil then
+    globalstate.enabled = enable
+    for bufnr, _ in pairs(bufstates) do
+      if api.nvim_buf_is_loaded(bufnr) then
+        if enable == false then
+          _disable(bufnr)
+        else
+          _enable(bufnr)
+        end
```
justinmk, line 423: Looks like this erases all the buffer-local states. That seems to match `vim.diagnostic.enable()`: https://github.com/neovim/neovim/blob/e778e0116198470ba037b9426f4ff7fa5cb7f880/runtime/lua/vim/diagnostic.lua#L2046-L2053

That seems a bit "destructive". I tend to think that global enable() should act like a gate: 

- if globally enabled, then defer to the buffer-local flag
- if globally disabled, don't check the buffer-local flag

But since this matches `vim.diagnostic.enable()` I think we can go with this for now.

justinmk (approved):
This is an enhancement to the current behavior. Thank you for carefully thinking about this!

Long-term, we will need to decide a clear pattern: https://github.com/neovim/neovim/issues/28603
That doesn't need to block this though.

runtime/lua/vim/lsp/inlay_hint.lua
famiu: > I'd say I write this in such a verbose way is I want to emphasize that the `bufstates` is an array of `bufstate`s, and we can only get `bufstate` from it, not anything else. Maybe the effect is not very obvious? I might change it.

I understand, but it's already very clear. It's pretty obvious that an item in a list named `bufstates` will be a single bufstate, so the compact form should be strongly preferred

runtime/lua/vim/lsp/inlay_hint.lua
ofseed: I think it should be possible to store `bufstate` as a `vim.b` variable. Formerly there was no requirement to prepare `bufstate` for every buffer, so we only create them when `enable` inlay hints on a buffer. Maybe it's just because this method is the most direct, and there doesn't seem to be any other LSP-related places using `vim.b`.

ofseed:
Since you mentioned this life cycle management issue, I wonder if I should change the way `bufstates` are managed here before this PR is merged.

I explained [above](https://github.com/neovim/neovim/pull/28543#issuecomment-2081686918) why using `vim.defaulttable` manage `bufstates`.
> (1) To make `bufstate`s inherit values from the `globalstate`, we need `setmetatable` for them. Instead of defining a factory function for `bufstate`, I think creating them when indexing `bufstates` will be simpler and more effective, this approach also avoids thinking about when this factory function should be called.
> (2) As you can see we can avoid a lot of judgment for whether the `bufstates[bufnr]` exists.

If we changed the implementation to `vim.b...` in the future, we still need judgments for nil check, and then add them back, this makes (2) moot. For (1), It’s not too troublesome to use factory functions or manage life cycles, we can create `bufstate` by a factor function when `LspAttach` and delete them when `LspDettach` or on buffer detach. Also avoiding potential performance issues.

famiu (approved):

runtime/lua/vim/lsp/inlay_hint.lua
justinmk: > Maybe it's just because this method is the most direct, 

Direct how? I would expect that leveraging a buffer-local variable avoids the need for logic to "cleanup" old state in this long-lived container.

> and there doesn't seem to be any other LSP-related places using `vim.b`.

I wouldn't take that as precedent, since (1) it might be driven by the performance consideration mentioned above, and (2) I think LSP features are typically per-client instead of per-buffer.

PR #28569, opened 2024-04-29
feat(lsp): redrawstatus on progress notifications

mfussenegger:
Problem:

Currently users can include `vim.lsp.status()` in their statusline, but
they won't see status changes if the statusline isn't redrawn on progress updates.

The simple solution for that is:

    autocmd LspProgress * redrawstatus

But that has the problem that the last message remains stale until
something else triggers a redraw. To fix the stale messages, users would
have to add something like this:

    local timer = vim.loop.new_timer()
    api.nvim_create_autocmd("LspProgress", {
      group = lsp_group,
      callback = function()
        vim.cmd.redrawstatus()
        if timer then
          timer:stop()
          timer:start(500, 0, vim.schedule_wrap(function()
            timer:stop()
            vim.cmd.redrawstatus()
          end))
        end
      end
    })

That's quite complex.

Solution:

Call redrawstatus by default. To avoid `redrawstatus()` burst the calls
are debounced to 60 FPS.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -562,9 +562,8 @@ LspProgress                                                       *LspProgress*
     `result` properties. `result` will contain the request params sent by the
     server.
 
-    Example: >vim
-        autocmd LspProgress * redrawstatus
-<
+    |:redrawstatus| happens implicitly with LspProgress events.
+
 
 LspRequest                                                        *LspRequest*
     For each request sent to an LSP server, this event is triggered for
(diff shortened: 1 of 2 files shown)


justinmk (approved):
Do we need a debounce util?

gpanders:
Is `redrawstatus` "cheap" if the statusline hasn't changed? Otherwise, this could be wasteful if the user doesn't use LSP status messages in their statusline (unless we include _that_ by default too?).

linrongbin16:
Is `redrawstatus` a blocking/sync IO on whole statusline?

My plugin [lsp-progress.nvim](https://github.com/linrongbin16/lsp-progress.nvim) will first format and cache each token received from `LspProgress`, and flush these string (status info) on a scheduled job running on background with a fixed interval (say 200 milliseconds).

The (async/non-blocking) produce & consume pattern seems resolve your issue?

I don't think sending `LspProgress` should always force the editor to redraw statusline.

mfussenegger:
I also noticed that this causes cursor flickering in insert mode. So this isn't viable as default.

PR #28573, opened 2024-04-29
vim-patch:9.1.0381

### LINKED ISSUE #28544: cbuffer doesn't take visual selection as range '<,'>cad
### Problem

Error "Invalid range" if you visual select something and do cbuffer. Passing a single number like 1cb works.
Bug also in upstream https://github.com/vim/vim/issues/14638 but the suggested patch is for a file that no longer exists in neovim (and I couldn't make it work by modyfing ex_cmds.lua addr_type to ADDR_LINES for cgetbuffer

### Steps to reproduce

nvim --clean
write something
select it
:'<,'>cbuffer

### Expected behavior

Cbuffer should take visual range (at least that's my expectation reading that cbuffer can take ranges)

### Neovim version (nvim -v)

 v0.10.0-dev-2898+g4c31a1b80 (same on v0.9.5)

yes 8.2.212 reproductible on master too

### Operating system/version

windows 11 (wsl2 ubuntu 22.04)

### Terminal name/version

windows terminal

xterm-256color

### Installation

appimage for 0.9.5 compilation for master

ilan-schemoul:
Fixes #28544 (problem which actually is caused by upstream)

Porting patch 9.1.0381 https://github.com/vim/vim/commit/652c821366691a8bd5474766581090df0c742fa1
```
Problem:  cbuffer and similar quickfix and locationlist commands don't
          accept a range, even so it is documented they should
          (ilan-schemoul, after 8.1.1241)
Solution: Define ex commands with ADDR_LINES instead of ADDR_OTHER

fixes: vim/vim#14638
closes: vim/vim#14657
```

### DIFF
--- a/runtime/doc/quickfix.txt
+++ b/runtime/doc/quickfix.txt
@@ -275,7 +275,8 @@ processing a quickfix or location list command, it will be aborted.
 			current window is used instead of the quickfix list.
 
 						*:cb* *:cbuffer* *E681*
-:cb[uffer][!] [bufnr]	Read the error list from the current buffer.
+:[range]cb[uffer][!] [bufnr]
+			Read the error list from the current buffer.
 			When [bufnr] is given it must be the number of a
 			loaded buffer.  That buffer will then be used instead
 			of the current buffer.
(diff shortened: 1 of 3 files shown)


zeertzjq:
Merged in #28571

ilan-schemoul:
Oh :( I was a bit late I didn't notice
Thanks for the PR anyway I am glad we can finally use visual range in buffers (e.g : term buffers)

PR #28576, opened 2024-04-30
feat(completion): add `pum_matches` key to `complete_info()`

famiu:
Problem: Currently there is no way of getting the list of completion items that are actually put in the popupmenu. Which means completion providers have no way of knowing if a completion was exhausted.

Solution: Add a `pum_matches` key to `complete_info()` which contains the indices of items that are put in the popupmenu.

Ref: https://github.com/vim/vim/issues/10007

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -796,6 +796,8 @@ complete_info([{what}])                                        *complete_info()*
 				dictionary containing the entries "word",
 				"abbr", "menu", "kind", "info" and "user_data".
 				See |complete-items|.
+		   pum_matches	Indices of completion matches that are put
+				in the |ins-completion-menu|.  First index is zero.
 		   selected	Selected item index.  First index is zero.
 				Index is -1 if no item is selected (showing
 				typed text only, or the last completion after
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -385,6 +385,9 @@ The following new APIs and features were added.
 • |vim.fs.root()| finds project root directories from a list of "root
   markers".
 
+• |complete_info()| returns a `pum_matches` item to indicate which items are put
+  in the |ins-completion-menu|.
+
 ==============================================================================
 CHANGED FEATURES                                                 *news-changed*
 
--- a/runtime/lua/vim/_meta/vimfn.lua
+++ b/runtime/lua/vim/_meta/vimfn.lua
@@ -1018,6 +1018,8 @@ function vim.fn.complete_check() end
 ---     dictionary containing the entries "word",
 ---     "abbr", "menu", "kind", "info" and "user_data".
 ---     See |complete-items|.
+---    pum_matches  Indices of completion matches that are put
+---     in the |ins-completion-menu|.  First index is zero.
 ---    selected  Selected item index.  First index is zero.
 ---     Index is -1 if no item is selected (showing
 ---     typed text only, or the last completion after
--- a/src/nvim/eval.lua
+++ b/src/nvim/eval.lua
@@ -1367,6 +1367,8 @@ M.funcs = {
       		dictionary containing the entries "word",
       		"abbr", "menu", "kind", "info" and "user_data".
       		See |complete-items|.
+         pum_matches	Indices of completion matches that are put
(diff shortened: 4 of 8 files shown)


glepnir:
don't get point. When there are no candidate words, that is when match_array is empty, the completechanged event will still be triggered ? idk why there need a new vvar ..complete_info also provide some infomations..

famiu:
> don't get point. When there are no candidate words, that is when match_array is empty, the completechanged event will still be triggered ? idk why there need a new vvar ..complete_info also provide some infomations..

Yes but you still need to know if completion is exhausted or not, and what the completion items are. Currently there is no way of getting that. Also, look at https://github.com/vim/vim/issues/10007 to know why `complete_info()` is not enough

glepnir:
It seems that because the items of complete_info are all items, it cannot distinguish the candidate words displayed in the current pum. I implemented similar behavior locally in my completion plugin. Now the order of items will not be disrupted so you can easily infer the candidate words displayed in the current pum.

runtime/doc/builtin.txt
```diff
@@ -796,6 +796,8 @@ complete_info([{what}])                                        *complete_info()*
 				dictionary containing the entries "word",
 				"abbr", "menu", "kind", "info" and "user_data".
 				See |complete-items|.
+		   pum_items	Indices of completion matches that are shown
```
echasnovski, line 799: There is already an `items` field which is a "list of completion matches". The field `pum_items` then suggests that the value has the same structure as the `items`, which it does not (and rightfully so).

As the new field is more aligned with `selected` field, maybe `shown` is a better name? Using suggested `shown_idx` is also good, just does not fit as well here.

famiu: > There is already an `items` field which is a "list of completion matches". The field `pum_items` then suggests that the value has the same structure as the `items`, which it does not (and rightfully so).
> 
> As the new field is more aligned with `selected` field, maybe `shown` is a better name? Using suggested `shown_idx` is also good, just does not fit as well here.

How about `pum_shown`?

echasnovski: > How about `pum_shown`?

Yep, also seems fine.

famiu: oops, accidentally deleted my comment

famiu: `shown` is probably a better name since `selected` is also related to PUM but doesn't have a `pum_` prefix

runtime/doc/builtin.txt
```diff
@@ -796,6 +796,8 @@ complete_info([{what}])                                        *complete_info()*
 				dictionary containing the entries "word",
 				"abbr", "menu", "kind", "info" and "user_data".
 				See |complete-items|.
+		   shown	Indices of completion matches that are shown
```
justinmk, line 799: in parallel with the existing `pum_visible` and `pumvisible()`, seems like "visible" is the right name for this.

```suggestion
		   visible	Indices of completion matches that are shown
```

zeertzjq:
The name "visible" may also be ambiguous. If there are many matches, there will be a scrollbar and some matches are only shown after scrolling. Should those be considered "visible" or not?

justinmk:
> If there are many matches, there will be a scrollbar and some matches are only shown after scrolling. Should those be considered "visible" or not?

And that should be spelled out in the docs. I thought for sure that the purpose of this PR was to indicate which items are visible on the screen. What, then, is the purpose? (Say it in the docs, not to me)

echasnovski:
> The name "visible" may also be ambiguous. If there are many matches, there will be a scrollbar and some matches are only shown after scrolling. Should those be considered "visible" or not?

True. Maybe "matched" then?

famiu:
> >

This makes the most amount of sense, considering how that's what it's called in the code as well.

famiu:
> > The name "visible" may also be ambiguous. If there are many matches, there will be a scrollbar and some matches are only shown after scrolling. Should those be considered "visible" or not?
> 
> True. Maybe "matched" then?

This doesn't work because completion items are called completion matches in several places in the documentation

zeertzjq:
Superseded by #31796

PR #28578, opened 2024-04-30
fix(ui): avoid recursiveness and invalid memory access

### LINKED ISSUE #21604: Crash with :redraw from vim.ui_attach() callback
### Neovim version (nvim -v)

6ba34e21fee2a81677e8261dfeaf24c8cd320500

no

### Operating system/version

Debian Sid

### Terminal name/version

Kitty 0.26.5-2, XTerm 377-1

xterm-kitty for kitty, xterm-256color for xterm

### Installation

from source

Having `~/tmp/noice-min.lua` (from https://github.com/folke/noice.nvim/wiki/Minimal-%60init.lua%60-to-Reproduce-an-Issue + modifications) with
```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  {
    "folke/noice.nvim",
    dependencies = {
      "MunifTanjim/nui.nvim",
      "rcarriga/nvim-notify",
    },
  },
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

-- add anything else here
vim.opt.termguicolors = true
vim.cmd([[colorscheme tokyonight]])
require("noice").setup({
  messages = {
    view = 'mini',
    view_search = false,
    view_error = 'mini',
    view_warn = 'mini',
  },
  notify = {
    enabled = true,
    view = 'mini',
  },
  lsp = {
    signature = {
      enabled = false,
    },
    hover = {
      enabled = false,
    }
  },
  views = {
    mini = {
      timeout = 5000,
    },
  },
})

vim.keymap.set('c', '<C-x>', function()
  require('noice').redirect(vim.fn.getcmdline())
  vim.api.nvim_input('<esc>')
end, { desc = 'Redirect Cmdline' })
```
`local.mk` used to build with 
```makefile
CMAKE_BUILD_TYPE=RelWithDebInfo
CMAKE_EXTRA_FLAGS="-DCMAKE_INSTALL_PREFIX=${HOME}/tmp/nvim-from-src"
CMAKE_EXTRA_FLAGS += -DENABLE_LIBINTL=OFF
```
run 
```
~/tmp/nvim-from-src/bin/nvim -u ~/tmp/noice-min.lua
```
type `:version` and submit with `C-x`. 

### Expected behavior

Nvim does not crash.

### Actual behavior

Nvim crashes without message on xterm and kitty, works as expected with nvim-qt.

luukvbaal:
Problem:  Calling :redraw from vim.ui_attach() callback results in
          recursive cmdline/message events.
Solution: Avoid recursiveness where possible and replace global "call_buf"
          with separate, temporary buffers for each event so that when a Lua
          callback for one event fires another event, that does not result
          in invalid memory access.

Fix #21604

### DIFF
--- a/src/nvim/ui.c
+++ b/src/nvim/ui.c
@@ -71,8 +71,6 @@ static bool has_mouse = false;
 static int pending_has_mouse = -1;
 static bool pending_default_colors = false;
 
-static Array call_buf = ARRAY_DICT_INIT;
-
 #ifdef NVIM_LOG_DEBUG
 static size_t uilog_seen = 0;
 static const char *uilog_last_event = NULL;
@@ -128,14 +126,11 @@ void ui_init(void)
   default_grid.handle = 1;
   msg_grid_adj.target = &default_grid;
   ui_comp_init();
-  kv_ensure_space(call_buf, 16);
 }
 
 #ifdef EXITFREE
 void ui_free_all_mem(void)
 {
-  kv_destroy(call_buf);
-
   UIEventCallback *event_cb;
   map_foreach_value(&ui_event_cbs, event_cb, {
     free_ui_event_callback(event_cb);
@@ -197,11 +192,9 @@ void ui_refresh(void)
   int width = INT_MAX;
   int height = INT_MAX;
   bool ext_widgets[kUIExtCount];
-  for (UIExtension i = 0; (int)i < kUIExtCount; i++) {
-    ext_widgets[i] = ui_active();
-  }
-
   bool inclusive = ui_override();
+  memset(ext_widgets, ui_active(), ARRAY_SIZE(ext_widgets));
+
   for (size_t i = 0; i < ui_count; i++) {
     RemoteUI *ui = uis[i];
     width = MIN(ui->width, width);
@@ -218,7 +211,7 @@ void ui_refresh(void)
     if (i < kUIGlobalCount) {
       ext_widgets[i] |= ui_cb_ext[i];
     }
-    // Set 'cmdheight' to zero when ext_messages becomes active for all tabpages.
+    // Set 'cmdheight' to zero for all tabpages when ext_messages becomes active.
     if (i == kUIMessages && !ui_ext[i] && ext_widgets[i]) {
       set_option_value(kOptCmdheight, NUMBER_OPTVAL(0), 0);
       command_height();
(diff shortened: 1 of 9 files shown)


przepompownia:
Does not crash for me on minimal repro from
https://github.com/neovim/neovim/issues/21604#issuecomment-1372918925

luukvbaal:
> stop emitting event when "call_buf" was changed by Lua handler.

IDK if this is OK. Alternatively the memory can be copied of course but that would presumably introduce a considerable performance hit. Perhaps the generator can be modified to just give each event it's own `call_buf`?

src/nvim/ui.c
```diff
@@ -722,6 +715,9 @@ void ui_call_event(char *name, Array args)
   map_foreach_value(&ui_event_cbs, event_cb, {
     Error err = ERROR_INIT;
     Object res = nlua_call_ref(event_cb->cb, name, args, kRetNilBool, NULL, &err);
+    // TODO(bfredl/luukvbaal): should this be documented or reconsidered?
+    // Why does truthy return from Lua callback mean remote UI should not receive
+    // the event.
     if (LUARET_TRUTHY(res)) {
```
justinmk, line 721: I assume this was in analogy w/ `nvim_create_autocmd` https://github.com/neovim/neovim/blob/0b8a72b73934d33a05e20c255298e88cd921df32/runtime/doc/api.txt#L3455-L3456

though I don't know what the actual use-case is, here.

justinmk (approved):
Very nice!

folke:
Great!

I need to revisit some of the hacks I added in Noice, since you've been very active in fixing a bunch of related underlying issues :)

PR #26739, opened 2023-12-26
refactor(completion): improve popup window position 

glepnir:
Problem: 
1. when new leader add and selected is -1 the pum window width changed like (be small) the info floating window position not changed.
2.  correct selected check avoid memory leak.
3. The info string does not need to be stored in compl, it is usually a very long byte string. So a large amount of memory needs to be allocated. We just need to write them into the info buffer.

Solutioin: when new leader add and selected is -1 we still need adjust info window width. And don't store info value. 

Fix https://github.com/neovim/neovim/issues/27768

Issue #16488, opened 2021-12-01
Completion API: completion handlers and calling lua callbacks on completion events


chentau:
### Feature already in Vim?

No

### Feature description

#13854 adds support for general-purpose fuzzy filtering of completion candidates, but I've been recently thinking about ways to make this more general, and allow plugins to have more control over the completion process beyond just filtering. More specifically, I'm proposing an API such as the following:

```lua
vim.api.nvim_completion_attach(mode, {
  on_init = function(line, index)
    -- returns the index where completion should start, like omnifunc
  end,
  on_fetch = function(prefix)
    -- returns a list of completion candidates, overriding the current completion list.
  end,
  on_fetch_chain = function(prefix)
    -- returns a list of completion candidates, which are then appended to the current completion list.
  end
  on_filter = function(prefix, match)
    -- returns a float representing the match between prefix and match.
    -- higher numbers are sorted first in the completion list.
  end,
  on_select = function(match)
    -- called when a completion candidate is selected.
  end,
  on_accept = function(match)
    -- called when a completion candidate is accepted.
  end,
  on_cancel = function() end,
})
```
where `mode` is one of the builtin completion modes, such as `CTRL_X_{NORMAL, FILES, TAGS}`, etc.

As follow up work, we can also allow for custom user completion modes via a new api function `nvim_complete`, which is also implemented in #13854:

```
vim.api.nvim_completion_attach("user:git", ...)
vim.api.nvim_complete(1, {"git", "files", "here"}, {completion_mode = "git"})
```

I'm posting this issue to try and get feedback on whether this is even a good idea or not, and general API design, before I start writing anything.

mjlbach:
First off, I think this is a great idea. Built-in completion is extremely underutilized. I'm curious how completion_mode intersects with the chain completion, can you provide an array of completion_modes?

clason:
Quick question: should this be a general API (`nvim_XXX`, which comes with a rigid contract and is also potentially available through vim script or rpc), or is a Lua API (`vim.completion.XXX`) enough?

muniter:
> Quick question: should this be a general API (`nvim_XXX`, which comes with a rigid contract and is also potentially available through vim script or rpc), or is a Lua API (`vim.completion.XXX`) enough?

[...]

PR #28579, opened 2024-04-30
fix(api): mark nvim__complete_set as experimental

justinmk:
Problem:
nvim_complete_set was added in 5ed55ff14c8b7e346811cb6228bf63fb5106bae9 but needs more bake time.

Solution:
Rename it, mark it as experimental.

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -584,6 +584,23 @@ created for extmark changes.
 ==============================================================================
 Global Functions                                                  *api-global*
 
+nvim__complete_set({index}, {opts})                     *nvim__complete_set()*
+    EXPERIMENTAL: this api may change in the future.
+
+    Sets info for the completion item at the given index. If the info text was
+    shown in a window, returns the window and buffer ids, or empty dict if not
+    shown.
+
+    Parameters: ~
+      • {index}  Completion candidate index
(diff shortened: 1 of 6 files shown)


clason:
I checked and don't see any current use of it on Github. However, that may be due to low visibility; if plugins want to make use of `completeopt`, they may need a function like that (e.g., in an async context where completion items are fetched on demand).

clason:
Collecting here for the purpose of recording things that should be improved:
* `nvim_complete_set` should be able to set multiple candidates at once
* `nvim_complete_set` should be able to set multiple properties
* float management is unrelated and should be kept out of this (if plugins implement their custom pum float, they are perfectly capable of keeping track of it themselves)
* better and documented handling of the state of the completion (robust handling of calling without active completion, or invalid completion index) via explicit return values.

First two points point toward `{ id1 = { info = ..., kind = ....}, id2 = { ... }, ... }` signature.

glepnir:
> but needs more bake time.

after https://github.com/neovim/neovim/pull/26739  i used popup for a while and it woks no problem . fair enough. Because we can integrate it into the new `nvim_complete` ref https://github.com/neovim/neovim/issues/16488. nvim_complete_set will be redundant then. as experimental  can make sure we remove it safely in future.

PR #28580, opened 2024-04-30
docs(api): sort unreleased nvim__ functions last

justinmk:
cc @lewis6991

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -584,99 +584,6 @@ created for extmark changes.
 ==============================================================================
 Global Functions                                                  *api-global*
 
-nvim__complete_set({index}, {opts})                     *nvim__complete_set()*
-    EXPERIMENTAL: this api may change in the future.
-
-    Sets info for the completion item at the given index. If the info text was
-    shown in a window, returns the window and buffer ids, or empty dict if not
-    shown.
-
-    Parameters: ~
-      • {index}  Completion candidate index
-      • {opts}   Optional parameters.
-                 • info: (string) info text.
-
-    Return: ~
-        Dictionary containing these keys:
-        • winid: (number) floating window id
-        • bufnr: (number) buffer id in floating window
-
-nvim__get_runtime({pat}, {all}, {opts})                  *nvim__get_runtime()*
-    Find files in runtime directories
-
-    Attributes: ~
-        |api-fast|
-
-    Parameters: ~
-      • {pat}   pattern of files to search for
-      • {all}   whether to return all matches or only the first
-      • {opts}  is_lua: only search Lua subdirs
-
-    Return: ~
-        list of absolute paths to the found files
-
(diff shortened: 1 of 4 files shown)


runtime/doc/api.txt
```diff
@@ -1572,6 +1479,99 @@ nvim_unsubscribe({event})                                 *nvim_unsubscribe()*
     Parameters: ~
       • {event}  Event type string
 
+nvim__complete_set({index}, {opts})                     *nvim__complete_set()*
+    EXPERIMENTAL: this api may change in the future.
```
clason, line 1483:
```suggestion
    EXPERIMENTAL: this API may change in the future.
```
This actually applies to all these functions. Can we move them to their own section, with this caveat at the head?

runtime/doc/api.txt
```diff
@@ -1572,6 +1479,99 @@ nvim_unsubscribe({event})                                 *nvim_unsubscribe()*
...
+      • {opts}   Optional parameters.
+                 • info: (string) info text.
+
+    Return: ~
+        Dictionary containing these keys:
+        • winid: (number) floating window id
+        • bufnr: (number) buffer id in floating window
+
+nvim__get_runtime({pat}, {all}, {opts})                  *nvim__get_runtime()*
+    Find files in runtime directories
```
clason, line 1500: I hope we can graduate this soon... 😞

runtime/doc/api.txt
```diff
@@ -1572,6 +1479,99 @@ nvim_unsubscribe({event})                                 *nvim_unsubscribe()*
...
+      • {opts}  is_lua: only search Lua subdirs
+
+    Return: ~
+        list of absolute paths to the found files
+
+nvim__id({obj})                                                   *nvim__id()*
+    Returns object given as argument.
+
+    This API function is used for testing. One should not rely on its presence
+    in plugins.
```
clason, line 1517: There's a few other experimental APIs that are not documented at all (e.g., `nvim__get_lib_dir()`) -- should those also be documented, or -- vice versa -- all internal functions just be dropped from the documentation?

runtime/doc/api.txt
justinmk: that can be a followup. this is just an incremental improvement compared to the current situation where these are prominently at the top of the section

runtime/doc/api.txt
clason: Personally, I'd suggest
* documenting _experimental_ API (as experimental): `complete_set`,  `get_runtime`, `get_lib_dir`, `redraw_range`
* not documenting _internal_ API (as `@nodoc`): the rest

justinmk: > Should those also be documented, or -- vice versa -- all internal functions just be dropped from the documentation?

I thought about that, but then that makes this less useful for "experiments", since people can't find the docs.

Out of scope for this PR though

lewis6991 (approved):

PR #28581, opened 2024-04-30
feat(completion): add new highlight PmenuMatch

### LINKED ISSUE #8780: new highlight group for Pmenu text-typed-so-far
It would be nice if there was a way to have standard Pmenu have a highlight group for the text typed so far.

Something like [ncm2-match-highlight](https://github.com/ncm2/ncm2-match-highlight) but with a highlight group instead of replacing characters with their [math symbols](https://en.wikipedia.org/wiki/Mathematical_Alphanumeric_Symbols).

glepnir:
Problem: missing highlight for matched leader text within candidate words.

Solutin: add highlight for the text pos which matched leader in candidate words.

Fix #8780

draft as 0.11


clason:
Such changes should go through Vim first, if possible.

glepnir:
yes. but vim doesn't have  grid. but i will try to port it, so just keep it as a draft for now 😸

clason:
It's about the interface and documentation more than about the implementation.

zeertzjq:
Superseded by #29311

PR #28591, opened 2024-05-01
ci: update vim_patches.yml to follow conventional commits

dundargoc:
This will change the commits messages from

```
version.c: update [skip ci]
```

to

```
docs: update version.c [skip ci]
```


justinmk (approved):
This is fine, but normal `vim-patch:` commits should not change. We decide the conventions, after all.

PR #28594, opened 2024-05-01
fix(:set): clear the message grid after showing an option

Sam-programs:
Fixes #28593

### DIFF
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -3879,7 +3879,10 @@ void cursorcmd(void)
 
 void gotocmdline(bool clr)
 {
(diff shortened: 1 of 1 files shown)


zeertzjq:
I think the correct solution is to use something else instead of the `gotocmdline()` in `do_one_set_option()` when `ext_cmdline` is active.

zeertzjq:
Or maybe `gotocmdline()` should still call `msg_clr_eos()` when `clr` is `true` even if `ext_cmdline` is active.

zeertzjq:
Or maybe the change should be in `gotocmdline()`, making it do something different when `ext_cmdline` or `ext_messages` is active and `clr` is `true`. That may fix #11558 as well.

PR #28597, opened 2024-05-02
Add file type detection for Inko

yorickpeterse:
This is so that when one uses Tree-sitter (see https://github.com/nvim-treesitter/nvim-treesitter/pull/6549 for that) instead of the Vim plugin (https://github.com/inko-lang/inko.vim), NeoVim automatically picks up on the file type.

### DIFF
--- a/test/old/testdir/test_filetype.vim
+++ b/test/old/testdir/test_filetype.vim
@@ -348,6 +348,7 @@ func s:GetFilenameChecks() abort
     \ 'ipfilter': ['ipf.conf', 'ipf6.conf', 'ipf.rules'],
     \ 'iss': ['file.iss'],
     \ 'ist': ['file.ist', 'file.mst'],
+    \ 'inko': ['file.inko'],
     \ 'j': ['file.ijs'],
     \ 'jal': ['file.jal', 'file.JAL'],
     \ 'jam': ['file.jpl', 'file.jpr', 'JAM-file.file', 'JAM.file', 'Prl-file.file', 'Prl.file'],
--- a/runtime/lua/vim/filetype.lua
+++ b/runtime/lua/vim/filetype.lua
@@ -358,6 +358,7 @@ local extension = {
   bat = 'dosbatch',
   wrap = 'dosini',
   ini = 'dosini',
+  inko = 'inko',
   INI = 'dosini',
   vbp = 'dosini',
   dot = 'dot',


test/old/testdir/test_filetype.vim
```diff
@@ -348,6 +348,7 @@ func s:GetFilenameChecks() abort
     \ 'ipfilter': ['ipf.conf', 'ipf6.conf', 'ipf.rules'],
     \ 'iss': ['file.iss'],
     \ 'ist': ['file.ist', 'file.mst'],
+    \ 'inko': ['file.inko'],
```
zeertzjq, line 351: This should be between `inittab` and `ipfilter`

zeertzjq:
Search previously closed filetype PRs for how to add a new filetype.

test/old/testdir/test_filetype.vim
yorickpeterse: @zeertzjq The list itself is fairly arbitrary (i.e. it's only partially alphabetical, the rest is seemingly a mess). Is the need for this documented somewhere?

yorickpeterse:
@zeertzjq 

> Search previously closed filetype PRs for how to add a new filetype.

While I appreciate the desire to be efficient in writing, I'd also appreciate _a little more_ than the equivalent of "Read the unwritten manual or go away".

When adding these changes I looked at existing commits, which seem to mostly port changes from Vim. Vim itself doesn't have the detection rules, and since it doesn't include e.g. syntax rules and what not (that's what https://github.com/inko-lang/inko.vim is for), I'm not sure there's any benefit to adding _just_ filetype detection to Vim and then port it over.

If there's any documentation on this I'd love to read it, but "Go search through thousands of PRs to find something that _might_ be relevant" isn't helpful.

gpanders:
>When adding these changes I looked at existing commits, which seem to mostly port changes from Vim. Vim itself doesn't have the detection rules, and since it doesn't include e.g. syntax rules and what not (that's what https://github.com/inko-lang/inko.vim is for), I'm not sure there's any benefit to adding just filetype detection to Vim and then port it over.

We do typically require filetype changes to go through Vim first, even if it is just for the filetype detection. Since you are the owner of the inko.vim plugin, you could upstream your plugin (syntax and ftplugin and all) to Vim, and that would get ported to Neovim.

yorickpeterse:
@gpanders In that case, where can I find more details on porting over the file type changes in the desired format? Looking at `CONTRIBUTING.md` I'm not able to find any mention of a script of sorts, or whether this is simply a manual process.

gpanders:
>Or is there simply no documentation on this and is it a manual process?

I think this is unfortunately the case, the vast majority of these ports are done by a small group of people (@zeertzjq and @clason typically) who are familiar with their process already.

If the port is simple it's usually not too involved though. In your case, since it's a simple extension based detection, you would add `inko` to the `extension` list, as this PR does, but it should be sorted alphabetically based on _filetype_ (not extension).

zeertzjq:
I think this should be documented in `CONTRIBUTING.md` too.

PR #28600, opened 2024-05-02
fix(colorscheme): add missing LSP groups

echasnovski:
After #28500 I realized that some LSP highlight groups did not have any default definitions (for example, `LspSignatureActiveParameter` for the now built-in `<C-S>` mapping). This PR fixes the issue.

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -5054,6 +5054,8 @@ QuickFixLine	Current |quickfix| item in the quickfix window. Combined with
 							*hl-Search*
 Search		Last search pattern highlighting (see 'hlsearch').
 		Also used for similar items that need to stand out.
+							*hl-SnippetTabstop*
+SnippetTabstop	Tabstops in snippets. |vim.snippet|
 							*hl-SpecialKey*
 SpecialKey	Unprintable characters: Text displayed differently from what
 		it really is. But not 'listchars' whitespace. |hl-Whitespace|
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -4338,7 +4338,7 @@ vim.snippet.expand({input})                             *vim.snippet.expand()*
     https://microsoft.github.io/language-server-protocol/specification/#snippet_syntax
     for the specification of valid input.
 
-    Tabstops are highlighted with hl-SnippetTabstop.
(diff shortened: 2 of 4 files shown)


runtime/doc/lsp.txt
```diff
@@ -406,6 +406,10 @@ LspSignatureActiveParameter
     Used to highlight the active parameter in the signature help. See
     |vim.lsp.handlers.signature_help()|.
 
+                                                           *hl-SnippetTabstop*
+SnippetTabstop
+    Used to color tabstops in snippets.
```
clason, line 411: Snippets are not really LSP specific (they live under `vim.snippet`, after all), so I don't think they should be documented here.

clason (changes requested):

runtime/doc/lsp.txt
echasnovski: Snippets specifically are designed to follow LSP specification and are added (only) to provide its full coverage. I understand that they are implemented in separate module, yet in spirit they are closer to `hl-Lsp***` highlight groups than `:h highlight-groups` (which do not contain any highlight groups coming from built-in Lua modules).

clason: Yes, because the LSP specification has become a lingua franca for snippets. They are not _intended_ to be LSP-only (even if that's all they're _currently_ used for _in core_). I'm afraid I'll have to insist.

clason (approved):

runtime/doc/syntax.txt
```diff
@@ -5054,6 +5054,8 @@ QuickFixLine	Current |quickfix| item in the quickfix window. Combined with
 							*hl-Search*
 Search		Last search pattern highlighting (see 'hlsearch').
 		Also used for similar items that need to stand out.
+							*hl-SnippetTabstop*
+SnippetTabstop	Tabstops in snippets.
```
clason, line 5058:
```suggestion
SnippetTabstop	Tabstops in snippets. |vim.snippet|
```

echasnovski: Maybe link directly to `vim.snippet.expand()` instead?

clason: I thought about that, but thought the main module would be simpler here. But either would work.

PR #28606, opened 2024-05-02
docs: fix filetype.lua example for matching on contents

gpanders:
When using a table, the first value is the filetype string or function,
and the second value is the options table. The current example "works"
because the function still has index 1, but the priority field is
ignored.


clason (approved):

dundargoc:

PR #28560, opened 2024-04-29
fix(lsp): remove redundant vim.snippet.jumpable

MariaSolOs:
Closes #28548

PR #28607, opened 2024-05-02
docs: fix snippets keymap examples

jonathan-raines:
The `vim.snippet.jumpable` function was removed in
[28560](https://github.com/neovim/neovim/pull/28560) but the docs still refer to it in the examples
leading to errors when trying to jump between snippets.


zeertzjq:
I think it should link to the snippet at `vim.snippet.active()` instead, or maybe that snippet should be moved here.

jonathan-raines:
Looks like there's two ways these jumps are defined. The way that checks if the snippet is active and without. The other definition is in the lua/snippet_spec.lua tests ( which is still slightly different than what I modified )

dundargoc:

PR #28609, opened 2024-05-02
docs: misc

### LINKED ISSUE #28484: Docs: describe priority between editorconfig and autocmd like FileType
### Problem

`:h editorconfig` does not explain when and how the settings are applied, jus that it is automatic (in nvim 0.9.4).

It seems like it applies FileType autocmd first and then editorconfig after that - so that editorconfig has precedence. To override the editorconfig one can explicitly do `:set ft=python` for example.

### Expected behavior

The behavior should be described in the help page.

### LINKED ISSUE #28719: Incorrect type annotation for vim.lsp.buf_request_sync
### Problem

The type annotation for [vim.lsp.buf_request_sync](https://github.com/neovim/neovim/blob/4e5c633ed4871a948aff7338b793ac5f93484153/runtime/lua/vim/lsp.lua#L941C1-L941C106) has the following return type annotation:
```lua
---@return table<integer, {err: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
---@return string? err On timeout, cancel, or error, `err` is a string describing the failure reason, and `result` is nil.
```
The `err: lsp.ResponseError` part of the type annotation is incorrect and actually returns `error: lsp.ResponseError`.

The real return type of the function is:
```lua
---@return table<integer, {error: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
---@return string? err On timeout, cancel, or error, `err` is a string describing the failure reason, and `result` is nil.
```

I believe the type annotation should be changed to match what the function returns.

minimal_init.lua
```lua
local pattern = 'lua'
local cmd = { 'lua-language-server' }

    pattern = pattern,
    callback = function(args)
        vim.lsp.start({
            name = 'lua_ls',
            cmd = cmd,
            root_dir = root_dir,
            settings = settings,
        })
    end,
})

vim.api.nvim_create_user_command('Test', function(_)
    local bufnr = 0
    local params = {} -- making this empty will force lsp to return the error response we want to test
    local timeout = 1000
    local results, _ = vim.lsp.buf_request_sync(bufnr, 'textDocument/codeAction', params, timeout)
    if results[1].error == nil and result[1].err ~= nil then
        vim.print('return type matched original type annotation')
    elseif results[1].error ~= nil and results[1].err == nil then
        vim.print(
            'return type has the `error` field, but the original type annotation incorrectly says it should be `err`'
        )
    else
        vim.print("something was returned that I didn't expect when making this test")
    end
    vim.print(results)
end, {
    desc = 'Tests error response',
})
```

1. Run `nvim ./ -u minimal_init.lua` on a lua project 
2. Open a lua file with netrw
3. Wait for lsp to start
4. Run the command `:Test`
5. Inspect the messages `:messages` printed by the test command 

### Expected behavior

The expected messages should look something like:
```
return type has the `error` field, but the original type annotation incorrectly says it should be `err`
{ {
    error = {
      code = -32603,
      message = "[08:10:01.606][error][#0:script\\provider\\provider.lua:864]: script\\provider\\provider.lua:864: attempt to index a nil value (field 'textDocument')\nstack traceback:\n\tscript\\provider\\provider.lua:864: in function <script\\provider\
\provider.lua:862>\n\t[C]: in function 'xpcall'\n\tscript\\proto\\proto.lua:203: in function <script\\proto\\proto.lua:178>\n",
      <metatable> = {
        __tostring = <function 1>
      }
    }
  } }
```

The field `error` is returned which does not match the type annotation.

### Neovim version (nvim -v)

v0.10.0-dev-2889+gf064e72b9

lua-language-server 3.7.4

### Operating system/version

Windows 11 Pro/23H2

### Log file

_No response_

dundargoc:
Closes https://github.com/neovim/neovim/issues/28484.
Closes https://github.com/neovim/neovim/issues/28719.

Co-authored-by: Chris <crwebb85@gmail.com>
Co-authored-by: Gregory Anders <greg@gpanders.com>
Co-authored-by: Jake B <16889000+jakethedev@users.noreply.github.com>
Co-authored-by: Jonathan Raines <jonathan.s.raines@gmail.com>
Co-authored-by: Yi Ming <ofseed@foxmail.com>
Co-authored-by: Zane Dufour <zane@znd4.me>
Co-authored-by: zeertzjq <zeertzjq@outlook.com>

### DIFF
--- a/MAINTAIN.md
+++ b/MAINTAIN.md
@@ -22,7 +22,7 @@ In practice we haven't found a way to forecast more precisely than "next" and
 * Next feature-release (1.x.0)
 
 The forecasting problem might be solved with an explicit priority system (like
-Bram's todo.txt). Meanwhile the Neovim priority system is defined by:
+Vim's todo.txt). Meanwhile the Neovim priority system is defined by:
 
 * PRs nearing completion.
 * Issue labels. E.g. the `has:plan` label increases the ticket's priority merely
@@ -57,8 +57,8 @@ has a major bug:
 
 ### Release automation
 
-Neovim automation includes a [backport bot](https://github.com/zeebe-io/backport-action).
-Trigger the action by labeling a PR with `backport release-X.Y`. See `.github/workflows/backport.yml`.
+Neovim automation includes a [backport bot](https://github.com/korthout/backport-action).
+Trigger the action by labeling a PR with `ci:backport release-x.y`. See `.github/workflows/backport.yml`.
 
 Deprecating and removing features
 ---------------------------------
@@ -74,7 +74,7 @@ When a (non-experimental) feature is slated to be removed it should:
 1. Be _soft_ deprecated in the _next_ release
   - Use of the deprecated feature will still work.
   - This means deprecating via documentation and annotation (`@deprecated`).
-  - Include a note in `news.txt` under `DEPRECATIONS`.
+  - Include a note in `deprecated.txt`.
   - For Lua features, use `vim.deprecate()`. The specified version is the
     current minor version + 2. For example, if the current version is
     `v0.10.0-dev-1957+gd676746c33` then use `0.12`.
@@ -132,7 +132,7 @@ Some can be auto-bumped by `scripts/bump_deps.lua`.
 * [gettext](https://ftp.gnu.org/pub/gnu/gettext/)
 * [libiconv](https://ftp.gnu.org/pub/gnu/libiconv)
 * [libuv](https://github.com/libuv/libuv)
-* [libvterm](http://www.leonerd.org.uk/code/libvterm/)
+* [libvterm](https://www.leonerd.org.uk/code/libvterm/)
   * Downloading from the original source is unreliable, so we use our [mirror](https://github.com/neovim/libvterm) instead.
 * [lua-compat](https://github.com/keplerproject/lua-compat-5.3)
 * [tree-sitter](https://github.com/tree-sitter/tree-sitter)
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -897,12 +897,12 @@ end
 ---@param bufnr (integer) Buffer handle, or 0 for current.
 ---@param method (string) LSP method name
 ---@param params (table|nil) Parameters to send to the server
----@param handler fun(results: table<integer, {error: lsp.ResponseError, result: any}>) (function)
+---@param handler fun(results: table<integer, {error: lsp.ResponseError?, result: any}>) (function)
 --- Handler called after all requests are completed. Server results are passed as
 --- a `client_id:result` map.
 ---@return function cancel Function that cancels all requests.
 function lsp.buf_request_all(bufnr, method, params, handler)
-  local results = {} --- @type table<integer,{error:lsp.ResponseError, result:any}>
+  local results = {} --- @type table<integer,{error: lsp.ResponseError?, result: any}>
   local result_count = 0
   local expected_result_count = 0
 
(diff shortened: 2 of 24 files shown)


MAINTAIN.md
```diff
@@ -22,7 +22,7 @@ In practice we haven't found a way to forecast more precisely than "next" and
 * Next feature-release (1.x.0)
 
 The forecasting problem might be solved with an explicit priority system (like
-Bram's todo.txt). Meanwhile the Neovim priority system is defined by:
+vim's todo.txt). Meanwhile the Neovim priority system is defined by:
```
gpanders, line 25:
```suggestion
Vim's todo.txt). Meanwhile the Neovim priority system is defined by:
```

runtime/doc/options.txt
```diff
@@ -6886,8 +6886,6 @@ A jump table for the options with a short description can be found at |Q_op|.
 	   options	options and mappings local to a window or buffer (not
 			global values for local options)
 	   localoptions same as "options"
-	   slash	|deprecated| Always enabled. Uses "/" in filenames.
-	   unix		|deprecated| Always enabled. Uses "\n" line endings.
```
justinmk, line 6890: I don't see why we need to remove these. They are useful for explicitness.

justinmk (approved):

runtime/doc/options.txt
dundargoc: We've removed all instances of one-liner descriptions of deprecated options in `options.txt` before. Unsure why these are special tbh.

dundargoc: I still don't get it. If we *remove* a flag of an option they should not be mentioned in `options.txt`, but if we *deprecate* a flag they should be? This seems inconsistent.

I will revert the change in any case.

dundargoc: Also hard disagree, deprecation texts like this are obnoxious and unhelpful and I hate it. Dunno if that's a technical reason enough, but 99/10 times my instinctive response to reading these is "Cool, this information is completely useless. Why are you telling me about it? I want my 5 seconds of time back."

justinmk: Alternatively we could just mention the "/" and `\n` in the option description?

dundargoc: This is fine for now, let's evaluate it on a future PR 👍

src/nvim/options.lua
```diff
@@ -9388,8 +9388,8 @@ return {
            options	options and mappings local to a window or buffer (not
         		global values for local options)
            localoptions same as "options"
-           slash	|deprecated| Always enabled. Uses "/" in filenames.
-           unix		|deprecated| Always enabled. Uses "\n" line endings.
+           slash       |deprecated| Always enabled. Uses "/" in filenames.
+           unix                |deprecated| Always enabled. Uses "\n" line endings.
```
zeertzjq, line 9392: What is this whitespace change?

runtime/lua/vim/lsp.lua
```diff
@@ -938,7 +938,7 @@ end
 ---@param params table? Parameters to send to the server
 ---@param timeout_ms integer? Maximum time in milliseconds to wait for a result.
 ---                           (default: `1000`)
----@return table<integer, {err: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
+---@return table<integer, {error: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
```
wookayin, line 941: https://github.com/neovim/neovim/pull/28720#pullrequestreview-2051491767:

```suggestion
---@return table<integer, {error?: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
```

zeertzjq: This change seems also needed for `vim.lsp.buf_request_all()` and the two `results` variables in `vim.lsp.buf.typehierarchy()`.

PR #28613, opened 2024-05-02
fix(treesitter): escape `\` in `:InspectTree`

ribru17:
Some parsers for, e.g., LaTeX or PHP have anonymous nodes like `"\"` or `"\text"` that behave wonkily (especially the first example) in the `InspectTree` window, so this PR escapes them by adding another backslash in front of them

### DIFF
--- a/runtime/lua/vim/treesitter/dev.lua
+++ b/runtime/lua/vim/treesitter/dev.lua
@@ -226,7 +226,7 @@ function TSTreeView:draw(bufnr)
         text = string.format('(%s', item.node:type())
       end
     else
-      text = string.format('"%s"', item.node:type():gsub('\n', '\\n'):gsub('"', '\\"'))
+      text = string.format('%q', item.node:type()):gsub('\n', 'n')
     end
 
     local next = self:get(i + 1)


runtime/lua/vim/treesitter/dev.lua
```diff
@@ -226,7 +226,10 @@ function TSTreeView:draw(bufnr)
         text = string.format('(%s', item.node:type())
       end
     else
-      text = string.format('"%s"', item.node:type():gsub('\n', '\\n'):gsub('"', '\\"'))
+      text = string.format(
+        '"%s"',
+        item.node:type():gsub('\\', '\\\\'):gsub('\n', '\\n'):gsub('"', '\\"')
+      )
```
justinmk, line 232: it's probably time to use an actual pattern,

```suggestion
      text = string.format('"%s"', item.node:type()::gsub('["\\\n]', '\\%1'))
```

ribru17: Sadly it doesn't work, when it hits a newline instead of spitting out `\\n` it gives
```
\

```
(a backslash concatenated with a new line)

justinmk: ok then worse case:

    text = string.format('"%s"', item.node:type():gsub('\n', '\\n'):gsub('["\\]', '\\%1'))

zeertzjq:
Can it use `%q` instead `"%s"`?

ribru17:
Done; I tested and it works with `"\n"`, `"\""`, and `"\\.*"`. The final `gsub` is there because `%q` turns `hel\nlo` into
```
hel\
lo
```

justinmk (approved):

PR #28615, opened 2024-05-03
perf(extmarks): better track whether namespace has extmarks

zeertzjq:
This avoids redraw when adding/removing an empty namespace for a window.

This also avoids marktree traversal when clearing a namespace that has already been cleared, which is added as a benchmark.

### DIFF
--- a/runtime/lua/vim/highlight.lua
+++ b/runtime/lua/vim/highlight.lua
@@ -129,13 +129,13 @@ function M.on_yank(opts)
     yank_cancel()
   end
 
+  vim.api.nvim_win_add_ns(winid, yank_ns)
   M.range(bufnr, yank_ns, higroup, "'[", "']", {
     regtype = event.regtype,
     inclusive = event.inclusive,
     priority = opts.priority or M.priorities.user,
     _scoped = true,
   })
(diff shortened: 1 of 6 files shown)


zeertzjq:
This also avoids marktree traversal when clearing a namespace that has already been cleared. The following script takes 8.65s without this PR, while it only takes 1.07s with this PR:
```lua
vim.api.nvim_buf_set_lines(0, 0, -1, true, { "foo", "bar" })
local ns0 = vim.api.nvim_create_namespace("ns0")
local ns1 = vim.api.nvim_create_namespace("ns1")

for _ = 1, 10000 do
  vim.api.nvim_buf_set_extmark(0, ns0, 0, 0, {})
end

vim.api.nvim_buf_set_extmark(0, ns1, 1, 0, {})
for _ = 1, 10000 do
  vim.api.nvim_buf_clear_namespace(0, ns1, 0, -1)
end
```

zeertzjq:
The tracking isn't 100% accurate. If the first line to clear in `nvim_buf_clear_namespace` isn't 0, then the namespace may not be considered empty even if all extmarks in the namespace has been cleared.

justinmk (approved):

PR #20176, opened 2022-09-13
feat(term): pipe osc and dcs commands to underlying terminal emulator

sassanh:
Cursor shape changes are working but `imgcat` of iTerm2 still doesn't work, I guess Neovim/libuv/ncurses library/something is changing terminal mode/settings. I don't have much knowledge about that but if I write the exact same bytes to a file and `tail -f` that file in another tab of iTerm2, image shows correctly, but in Neovim it shows like this:

image

Issue #4349, opened 2016-02-25
:term should forward unknown sequences to host


chemzqm:
iTerm2 version 3 beta support show image inside terminal with command like `imgcat`, but the image could not be seen with that command inside terminal of neovim.
That would also be cool if neovim support show image in buffer with an api .

justinmk:
Some people are already doing this, see https://github.com/neovim/neovim/issues/4056

chemzqm:
@justinmk I'm using latest neovim, but I can't get any image ouput with command `imgcat` inside the terminal of neovim

justinmk:
It looks like [imgcat is using](https://raw.githubusercontent.com/gnachman/iTerm2/master/tests/imgcat) a custom DCS `1337`. I am guessing that is not a code that is recognized by libvterm (@leonerd can you confirm), but on the other hand we should forward it to iTerm2, but maybe we aren't?

@chemzqm Clearly there is _some_ seequence that is recognized by libvterm to display images as demonstrated by https://github.com/neovim/neovim/issues/4349. imgcat could be modified to emit those codes instead?

DarkDefender:
@justinmk IIRC the method in #4056 does not use terminal codes. It instead calls a image display program that displays the image at the requested location.

That's probably why it doesn't output the image at the correct location when inside neovim. Because it assumes that it is using the full terminal window. So it will not work with splits and so on.

I guess that the iTerm2 codes will eliminate this problem because the terminal will then know where to put the image depending on where it reads the DCS 1337 code

sassanh:
+1

rzpilgrim:
+1

sassanh:
hi @justinmk 
sorry to ping you here. I just got a notif from github that my comment here has been deleted. Do you have any idea why? Deletion of my comment is not important by itself, I just found it suspicious and thought I should inform you, specially considering another comment has been deleted.
I don't know about contents of the comments but I'm sure mine aged at least 6 months.

sassanh:
Checking my email I found the other comment was a "+1", unfortunately I don't have my own comment's content in my email but I guess it was also a "+1" and maybe "neovim" is a bot that deletes +1s which is good.

liubog2008:
imgcat can work well in vim but in neovim it prints chars @justinmk

liubog2008:
It looks that `vterm_state_set_unrecognised_fallbacks` is not called though libvterm is updated?

kdheepak:
[...]

PR #28621, opened 2024-05-03
fix(terminal): report terminal window size in pixels

### LINKED ISSUE #8259: "Terminal does not support reporting screen size s via the TIOCGWINSZ ioctl"
- `nvim --version`: NVIM v0.2.3-975-g84bac9f50
- Vim (version: ) behaves differently?  Different error but doesn't work either
- Operating system/version: macos 10.13
- Terminal name/version: kitty
- `$TERM`: xterm-256color

### Steps to reproduce using `nvim -u NORC`

Requires kitty terminal emulator and [its `icat` script](https://gist.github.com/abd37a3d7a0bb3f6f4d8ca15c6b1dfa3)

```
nvim -u NORC
:term
icat (some img file)
```

### Actual behavior

`Terminal does not support reporting screen sizes via the TIOCGWINSZ ioctl`

### Expected behavior

An image should be displayed.

lumynou5:
Window size in pixels is required by some programs, reporting it can help those programs work inside NeoVim terminals.

Closes #8259.

However, this patch doesn't make [`icat`](https://sw.kovidgoyal.net/kitty/kittens/icat/) work. Instead, it just makes the program show another error, because the problem is not only related to window sizes in pixels but how escape sequences (specifically, APC) are handled.  See #4349, #20176.

### DIFF
--- a/src/nvim/os/pty_process_unix.c
+++ b/src/nvim/os/pty_process_unix.c
@@ -171,13 +171,17 @@ int pty_process_spawn(PtyProcess *ptyproc)
   Process *proc = (Process *)ptyproc;
   assert(proc->err.closed);
   uv_signal_start(&proc->loop->children_watcher, chld_handler, SIGCHLD);
-  ioctl(1, TIOCGWINSZ, &ptyproc->winsize);
-  ptyproc->winsize = (struct winsize){
-    ptyproc->height,
-    ptyproc->width,
-    (ptyproc->winsize.ws_xpixel / ptyproc->winsize.ws_col) * ptyproc->width,
-    (ptyproc->winsize.ws_ypixel / ptyproc->winsize.ws_row) * ptyproc->height,
-  };
+  if (isatty(1)) {
+    ioctl(1, TIOCGWINSZ, &ptyproc->winsize);
+    ptyproc->winsize = (struct winsize){
+      ptyproc->height,
+      ptyproc->width,
+      (ptyproc->winsize.ws_xpixel / ptyproc->winsize.ws_col) * ptyproc->width,
+      (ptyproc->winsize.ws_ypixel / ptyproc->winsize.ws_row) * ptyproc->height,
+    };
+  } else {
+    ptyproc->winsize = (struct winsize){ ptyproc->height, ptyproc->width, 0, 0 };
+  }
   uv_disable_stdio_inheritance();
   int master;
   int pid = forkpty(&master, NULL, &termios_default, &ptyproc->winsize);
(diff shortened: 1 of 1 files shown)


src/nvim/os/pty_process_unix.c
```diff
@@ -171,7 +171,13 @@ int pty_process_spawn(PtyProcess *ptyproc)
   Process *proc = (Process *)ptyproc;
   assert(proc->err.closed);
   uv_signal_start(&proc->loop->children_watcher, chld_handler, SIGCHLD);
-  ptyproc->winsize = (struct winsize){ ptyproc->height, ptyproc->width, 0, 0 };
+  ioctl(1, TIOCGWINSZ, &ptyproc->winsize);
+  ptyproc->winsize = (struct winsize){
+    ptyproc->height,
+    ptyproc->width,
+    ptyproc->winsize.ws_xpixel / ptyproc->winsize.ws_col * ptyproc->width,
+    ptyproc->winsize.ws_ypixel / ptyproc->winsize.ws_row * ptyproc->height,
```
justinmk, line 179: use parens to make the order of operations obvious
```suggestion
    (ptyproc->winsize.ws_xpixel / ptyproc->winsize.ws_col) * ptyproc->width,
    (ptyproc->winsize.ws_ypixel / ptyproc->winsize.ws_row) * ptyproc->height,
```

justinmk (approved):
Nice, thanks. If no one else has comments, ping me if this isn't merged in a couple days

justinmk:
I wonder if this fixes any of the log messages that were showing in Windows CI https://github.com/neovim/neovim/blob/cf9f002f31c8b4d9d42912a3f45f5d3db4462fd9/test/testutil.lua#L95

lumynou5:
Currently the test is failed.  When stdout is redirected, `ioctl` calls fail with errno 25, and the program crashes due to invalid division.
Adding a check to set pixel sizes to `0` in case of failed calls can help, but it seems not a desirable solution?

```c
  int res = ioctl(1, TIOCGWINSZ, &ptyproc->winsize);
  ptyproc->winsize = (struct winsize){
    ptyproc->height,
    ptyproc->width,
    res ? 0 : (ptyproc->winsize.ws_xpixel / ptyproc->winsize.ws_col) * ptyproc->width,
    res ? 0 : (ptyproc->winsize.ws_ypixel / ptyproc->winsize.ws_row) * ptyproc->height,
  };
```

zeertzjq:
It seems this doesn't actually solve anything, and may not be correct. For example, when running Nvim with `--headless` and attaching another TUI with `--remote-ui`, this will use the pixel size of the terminal running the `--headless` instance instead of the one running the `--remote-ui` instance.

lumynou5:
> It seems this doesn't actually solve anything, and may be incorrect. For example, when running Nvim with `--headless` and attaching another TUI with `--remote-ui`, this will use the pixel size of the terminal running the `--headless` instance instead of the one running the `--remote-ui` instance.

Thanks. I didn't think about remote. I'll try to solve it.

PR #28623, opened 2024-05-03
build: remove `lintdoc` from `lint` target

dundargoc:
`lintdoc` takes too long to be part of `lint`. It may be reintroduced
once it's possible to only run lintdoc on files that have been changed.


justinmk:
> It may be reintroduced once it's possible to only run lintdoc on files that have been changed.

that's reasonable I guess. `lint` doesn't seem to slow to me, but I only run it when it's time to create a PR.

and the CI job still runs `lintdoc`, which is most important.

PR #28628, opened 2024-05-03
refactor(snippet): rename exit() => stop()

justinmk:
Reduce "vocabulary creep".

### DIFF
--- a/runtime/doc/develop.txt
+++ b/runtime/doc/develop.txt
@@ -373,6 +373,9 @@ Use existing common {verb} names (actions) if possible:
     - add:          Appends or inserts into a collection
     - attach:       Listens to something to get events from it (TODO: rename to "on"?)
     - call:         Calls a function
+    - cancel:       Cancels or dismisses an event or interaction, typically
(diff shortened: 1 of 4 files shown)


clason:
I'm not sure that has the right connotation here; "cancel" implies "do _not_ do this, revert", but `exit` _will_ have expanded the snippet.

justinmk:
> I'm not sure that has the right connotation here; "cancel" implies "do _not_ do this, revert",

It cancels the current operation, which indeed this function does, i.e. it doesn't proceed with the remaining snippet activities. That is pretty analogous to canceling a mapping, insert mode, etc: none of them *revert* any changes or edits.

clason:
I don't think we talk anywhere about "canceling insert mode" -- we "leave" a mode (which is also a valid semantic cousin). And "cancelling a mapping" _does_ prevent it from being executed. If it must be `stop` instead of `exit`, so be it. (As to "why", remember that this feature was provided by a new contributor that had not yet internalized such guidance.)

justinmk:
> This wasn't me making arbitrary naming decisions.

I know.

PR #28633, opened 2024-05-04
docs: fix lua type warnings

MariaSolOs:
Minor type improvements.

### DIFF
--- a/runtime/lua/vim/_meta/builtin_types.lua
+++ b/runtime/lua/vim/_meta/builtin_types.lua
@@ -127,3 +127,11 @@
 --- @field skipcol integer
 --- @field topfill integer
 --- @field topline integer
+
+--- @class vim.fn.getscriptinfo.ret
+--- @field autoload false
+--- @field functions? string[]
+--- @field name string
+--- @field sid string
+--- @field variables? table<string, any>
+--- @field version 1
--- a/src/nvim/eval.lua
+++ b/src/nvim/eval.lua
@@ -4470,11 +4470,12 @@ M.funcs = {
 
       Examples: >vim
       	echo getscriptinfo({'name': 'myscript'})
-      	echo getscriptinfo({'sid': 15}).variables
+      	echo getscriptinfo({'sid': 15})[0].variables
       <
     ]=],
     name = 'getscriptinfo',
     params = { { 'opts', 'table' } },
+    returns = 'vim.fn.getscriptinfo.ret[]',
     signature = 'getscriptinfo([{opts}])',
   },
   gettabinfo = {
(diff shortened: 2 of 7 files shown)


runtime/lua/vim/_meta/builtin_types.lua
```diff
@@ -127,3 +127,11 @@
...
 --- @field topfill integer
 --- @field topline integer
+
+--- @class vim.fn.getscriptinfo.ret
+--- @field autoload false
+--- @field functions? string[]
+--- @field name string
+--- @field sid string
+--- @field variables? table<string, any>
+--- @field vim 1
```
zeertzjq, line 137:
```suggestion
--- @field version 1
```

src/nvim/eval.lua
```diff
@@ -4475,6 +4475,7 @@ M.funcs = {
     ]=],
     name = 'getscriptinfo',
     params = { { 'opts', 'table' } },
+    returns = 'vim.fn.getscriptinfo.ret|vim.fn.getscriptinfo.ret[]',
```
zeertzjq, line 4478: This function also returns a list

zeertzjq, line 4478: This function always returns a list

runtime/lua/vim/_meta/builtin_types.lua
wookayin, line 137:
```suggestion
--- @field version 1
```

src/nvim/eval.lua
MariaSolOs: Nope, the docs say that is `sid` is specified, only information about the script with ID `"sid"` is returned. There is even this example: `echo getscriptinfo({'sid': 15}).variables`

zeertzjq: Hmm, the docs is wrong. When `sid` specified, it returns a list with a single item.

MariaSolOs: Oh yes you're right, I just checked the tests and it always returns a list. Fixing it.

src/nvim/eval.lua
```diff
@@ -4470,11 +4470,12 @@ M.funcs = {
 
       Examples: >vim
       	echo getscriptinfo({'name': 'myscript'})
-      	echo getscriptinfo({'sid': 15}).variables
+      	echo getscriptinfo({'sid': 15})[1].variables
```
zeertzjq, line 4473:
```suggestion
      	echo getscriptinfo({'sid': 15})[0].variables
```

MariaSolOs: uh duh Lua's 1-based indexing is getting into me.

PR #28638, opened 2024-05-04
feat/docs(editorconfig): add support for spelling_language

marcg03:
This fixes #28626.

For the other **editorconfig** properties it's easier to check for validity since they are either a number or part of an enum with few elements. The problem for `spelling_language` is that it is part of a greater enum since it's a concatenation of a [ISO 639 language code](https://www.loc.gov/standards/iso639-2/php/code_list.php) and an optional [ISO 3166 territory identifier](https://www.iso.org/obp/ui/#search).

For now the code only checks that the value provided in `spelling_language` is of the format [a-z][a-z] or [a-z][a-z]-[a-z][a-z] (regex), but it doesn't check whether it is a valid ISO639/ISO3166 code and thus when inputting an invalid value the `spelllang` option is also set to an unknown language (this results in a prompt asking you to download the unknown language and then failing and continuing if accepted).

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -128,9 +128,7 @@ PERFORMANCE
 
 PLUGINS
 
-• TODO
-
-• Editorconfig
+• EditorConfig
   • spelling_language property is now supported.
 
 STARTUP
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -130,6 +130,9 @@ PLUGINS
 
 • TODO
 
+• Editorconfig
+  • spelling_language property is now supported.
+
 STARTUP
 
 • TODO
--- a/runtime/lua/editorconfig.lua
+++ b/runtime/lua/editorconfig.lua
@@ -203,6 +203,8 @@ function properties.spelling_language(bufnr, val)
   if val:len() == 2 then
     vim.bo[bufnr].spelllang = language_code
   else
+    assert(val:sub(3, 3) == '-', error_msg)
+
     local territory_code = val:sub(4, 5):lower()
     assert(language_code:match('%l%l'), error_msg)
     vim.bo[bufnr].spelllang = language_code .. '_' .. territory_code
--- a/runtime/lua/editorconfig.lua
+++ b/runtime/lua/editorconfig.lua
@@ -206,7 +206,7 @@ function properties.spelling_language(bufnr, val)
     assert(val:sub(3, 3) == '-', error_msg)
 
     local territory_code = val:sub(4, 5):lower()
-    assert(language_code:match('%l%l'), error_msg)
+    assert(territory_code:match('%l%l'), error_msg)
     vim.bo[bufnr].spelllang = language_code .. '_' .. territory_code
   end
 
--- a/runtime/lua/editorconfig.lua
+++ b/runtime/lua/editorconfig.lua
@@ -193,19 +193,19 @@ end
 --- A code of the format ss or ss-TT, where ss is an ISO 639 language code and TT is an ISO 3166 territory identifier.
 --- Sets the 'spell' and 'spelllang' options.
 function properties.spelling_language(bufnr, val)
-  local errorMsg =
+  local error_msg =
     'spelling_language must be of the format ss or ss-TT, where ss is an ISO 639 language code and TT is an ISO 3166 territory identifier.'
 
-  assert(val:len() == 2 or val:len() == 5, errorMsg)
+  assert(val:len() == 2 or val:len() == 5, error_msg)
 
-  local languageCode = val:sub(1, 2):lower()
-  assert(languageCode:match('%l%l'), errorMsg)
+  local language_code = val:sub(1, 2):lower()
+  assert(language_code:match('%l%l'), error_msg)
   if val:len() == 2 then
-    vim.bo[bufnr].spelllang = languageCode
+    vim.bo[bufnr].spelllang = language_code
   else
-    local territoryCode = val:sub(4, 5):lower()
-    assert(languageCode:match('%l%l'), errorMsg)
-    vim.bo[bufnr].spelllang = languageCode .. '_' .. territoryCode
+    local territory_code = val:sub(4, 5):lower()
+    assert(language_code:match('%l%l'), error_msg)
+    vim.bo[bufnr].spelllang = language_code .. '_' .. territory_code
   end
 
   vim.o.spell = true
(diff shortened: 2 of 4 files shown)


runtime/lua/editorconfig.lua
```diff
@@ -191,6 +191,27 @@ function properties.insert_final_newline(bufnr, val)
   end
 end
 
+--- A code of the format ss or ss-TT, where ss is an ISO 639 language code and TT is an ISO 3166 territory identifier.
+--- Sets the 'spell' and 'spelllang' options.
+function properties.spelling_language(bufnr, val)
+  local errorMsg =
```
gpanders, line 197: Nit: use snake case

```suggestion
  local error_msg =
```

runtime/lua/editorconfig.lua
```diff
@@ -191,6 +191,27 @@ function properties.insert_final_newline(bufnr, val)
...
+    'spelling_language must be of the format ss or ss-TT, where ss is an ISO 639 language code and TT is an ISO 3166 territory identifier.'
+
+  assert(val:len() == 2 or val:len() == 5, errorMsg)
+
+  local languageCode = val:sub(1, 2):lower()
```
gpanders, line 202:
```suggestion
  local language_code = val:sub(1, 2):lower()
```

runtime/lua/editorconfig.lua
```diff
@@ -191,6 +191,27 @@ function properties.insert_final_newline(bufnr, val)
...
+  assert(languageCode:match('%l%l'), errorMsg)
+  if val:len() == 2 then
+    vim.bo[bufnr].spelllang = languageCode
+  else
+    local territoryCode = val:sub(4, 5):lower()
```
gpanders, line 207:
```suggestion
    local territory_code = val:sub(4, 5):lower()
```

gpanders (commented):
Looks good on first glance, thanks! Will give it a test whenever you mark it ready for review.

gpanders:
Can you please also update `news.txt` and add a test case to `editorconfig_spec.lua`?

sus-domesticus:
I encountered problems with the `test_case` method (in `editorconfig_spec.lua`).
The error is from passing `{ buf = 0 }` to `nvim_get_option_value` when getting the value of `spell` because it's **window-local** not **buffer-local**.
error

Should I adapt `test_case` for the **window-local** `spell` option?


https://github.com/sus-domesticus/neovim/blob/8372a48a3564564577ca5096a0c4584f0cdcd9a8/test/functional/plugin/editorconfig_spec.lua#L14-L22

test/functional/plugin/editorconfig_spec.lua
```diff
@@ -16,8 +16,18 @@ local testdir = 'Xtest-editorconfig'
...
   local filename = testdir .. pathsep .. name
   command('edit ' .. filename)
+
   for opt, val in pairs(expected) do
-    eq(val, api.nvim_get_option_value(opt, { buf = 0 }), name)
+    local opt_info = api.nvim_get_option_info2(opt)
+    if opt_info.scope == 'win' then
+      eq(val, api.nvim_get_option_value(opt, { win = 0 }), name)
+    else
+      if opt_info.scope == 'buf' then
```
zeertzjq, line 25: This can use `elseif`

sus-domesticus:
Looks like I forgot to add some parameters. It's weird because when I ran `make test` locally I didn't get an error.

test/functional/plugin/editorconfig_spec.lua
```diff
@@ -16,8 +16,16 @@ local testdir = 'Xtest-editorconfig'
 local function test_case(name, expected)
   local filename = testdir .. pathsep .. name
   command('edit ' .. filename)
+
   for opt, val in pairs(expected) do
-    eq(val, api.nvim_get_option_value(opt, { buf = 0 }), name)
+    local opt_info = api.nvim_get_option_info2(opt)
```
zeertzjq, line 21:
```suggestion
    local opt_info = api.nvim_get_option_info2(opt, {})
```

runtime/lua/editorconfig.lua
```diff
@@ -191,6 +191,27 @@ function properties.insert_final_newline(bufnr, val)
...
+  assert(val:len() == 2 or val:len() == 5, error_msg)
+
+  local language_code = val:sub(1, 2):lower()
+  assert(language_code:match('%l%l'), error_msg)
+  if val:len() == 2 then
+    vim.bo[bufnr].spelllang = language_code
+  else
+    local territory_code = val:sub(4, 5):lower()
+    assert(language_code:match('%l%l'), error_msg)
+    vim.bo[bufnr].spelllang = language_code .. '_' .. territory_code
```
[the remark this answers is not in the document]
sus-domesticus: Yes you are right! And also on line **208** instead of checking `territory_code` I am checking `language_code` again so I'll fix that as well.

test/functional/plugin/editorconfig_spec.lua
```diff
@@ -93,6 +101,12 @@ setup(function()
 
     [max_line_length.txt]
     max_line_length = 42
+
+    [short_spelling_language.txt]
+    spelling_language = en
+
+    [long_spelling_language.txt]
+    spelling_language = en-US
```
justinmk, line 109: probably should test something other than the default

gpanders (approved):
Sorry for the delay. Just needs a rebase onto the new `news.txt` file and then should be good.

runtime/doc/news.txt
```diff
@@ -130,6 +130,9 @@ PLUGINS
 
 • TODO
 
+• Editorconfig
```
gpanders, line 133: Nit
```suggestion
• EditorConfig
```

Can also delete the `TODO` just above

runtime/lua/editorconfig.lua
```diff
@@ -190,6 +190,29 @@ function properties.insert_final_newline(bufnr, val)
   end
 end
 
+--- A code of the format ss or ss-TT, where ss is an ISO 639 language code and TT is an ISO 3166 territory identifier.
+--- Sets the 'spell' and 'spelllang' options.
```
gpanders, line 194: I am not sure how I feel about this also setting `'spell'`. Someone may want to apply `spelling_language` globally (using a `[*]` glob) so that _when they do_ enable spell checking, it will use the specified language. But that does not imply they necessarily want spell checking always enabled.

I think we should remove `'spell'` for now and set only `'spelllang'`. This is reversible if we need to re-enable it later.

sus-domesticus: The spec also doesn't say anything about enabling spell checking.
> Sets the natural language that should be used...

sus-domesticus:
I could also try to tidy the commit history a bit if you think it's needed.

gpanders (approved):

gpanders:
Thanks!

Issue #19649, opened 2022-08-05
Related information of diagnostics is not parsed & displayed.


BigRedEye:
### Neovim version (nvim -v)

NVIM v0.8.0-dev-758-g711ef4eac9

### Language server name/version

clangd 14.0

### Operating system/version

Arch Linux

### Steps to reproduce using "nvim -u minimal_init.lua"

Many servers support relatedInformation; I'm personally struggling with clagnd.

```bash
$ cat <<EOF >repro.cpp
void foo(char);
void foo(double);
void bar() {
    foo(1);
}
EOF
$ nvim -u minimal_init.lua repro.cpp
:lua require('lspconfig').clangd.setup{}
:lua vim.diagnostic.setqflist()
```

### Expected behavior

Clangd shows three diagnostics (error + 2x info), just like standalone compiler:
```
repro.cpp:4:5: error: call to 'foo' is ambiguous
    foo(1);
    ^~~
repro.cpp:1:6: note: candidate function
void foo(char);
     ^
repro.cpp:2:6: note: candidate function
void foo(double);
```

VSCode handles this correctly:
![img](https://user-images.githubusercontent.com/17150781/183085819-919ca79b-ce48-4225-b224-19678647d458.png)

### Actual behavior

Clangd reports one diagnostic with two items in the `relatedInformation` list:
https://gist.github.com/BigRedEye/253629638e1d71f4df667a3e8e05a1d3.

Neovim does not parse relatedInformation and displays just one main diagnostic:
![img](https://user-images.githubusercontent.com/17150781/183086488-b136360c-7e23-4fd8-b4b8-0a01f7418fed.png)

tom-anders:
@glepnir Are you still working on this? I also really would like to have this feature and would like to help out with the implementation if needed!

tom-anders:
In the meantime, for anyone who wants this, here's a quick and dirty implementation I added to my config:

```lua
local function getlines(location)
  local uri = location.targetUri or location.uri
  if uri == nil then
    return
  end
  local bufnr = vim.uri_to_bufnr(uri)
  if not vim.api.nvim_buf_is_loaded(bufnr) then
    vim.fn.bufload(bufnr)
  end
  local range = location.targetRange or location.range

  local lines = vim.api.nvim_buf_get_lines(bufnr, range.start.line, range['end'].line+1, false)
  return table.concat(lines, '\n')
end

vim.diagnostic.config({float = {format = function(diag) 
    local message = diag.message
[...]

PR #28640, opened 2024-05-04
feat(lsp): add related_information to vim.Diagnostic

tom-anders:
Basically a simplified version of https://github.com/neovim/neovim/issues/19649#issuecomment-1750272564

Looking to get some initial feedback here. Instead of just appending to the message, we might instead want to store the related info in the `vim.Diagnostic` struct, and then instead extend the logic in `vim.diagnostic.open_float()`...?

This PR:
Screenshot_2024-05-04_21-20-17
vscode:
Screenshot_2024-05-04_21-22-00

EDIT: PR now only adds the related_information field to the toplevel of vim.Diagnostic (see discussion below). We could then decide in a follow-up PR if and how we should display this to the user.

### DIFF
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -350,21 +350,25 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
     0-based rows and columns). |api-indexing|
 
     Fields: ~
-      • {bufnr}?      (`integer`) Buffer number
-      • {lnum}        (`integer`) The starting line of the diagnostic
-                      (0-indexed)
-      • {end_lnum}?   (`integer`) The final line of the diagnostic (0-indexed)
-      • {col}         (`integer`) The starting column of the diagnostic
-                      (0-indexed)
-      • {end_col}?    (`integer`) The final column of the diagnostic
-                      (0-indexed)
-      • {severity}?   (`vim.diagnostic.Severity`) The severity of the
-                      diagnostic |vim.diagnostic.severity|
-      • {message}     (`string`) The diagnostic text
-      • {source}?     (`string`) The source of the diagnostic
-      • {code}?       (`string|integer`) The diagnostic code
-      • {user_data}?  (`any`) arbitrary data plugins can add
-      • {namespace}?  (`integer`)
+      • {bufnr}?                (`integer`) Buffer number
+      • {lnum}                  (`integer`) The starting line of the
+                                diagnostic (0-indexed)
+      • {end_lnum}?             (`integer`) The final line of the diagnostic
+                                (0-indexed)
+      • {col}                   (`integer`) The starting column of the
+                                diagnostic (0-indexed)
+      • {end_col}?              (`integer`) The final column of the diagnostic
+                                (0-indexed)
+      • {severity}?             (`vim.diagnostic.Severity`) The severity of
+                                the diagnostic |vim.diagnostic.severity|
+      • {message}               (`string`) The diagnostic text
+      • {related_information}?  (`vim.diagnostic.RelatedInformation[]`) An
+                                array of related diagnostic information. See
+                                |vim.diagnostic.RelatedInformation|.
+      • {source}?               (`string`) The source of the diagnostic
+      • {code}?                 (`string|integer`) The diagnostic code
+      • {user_data}?            (`any`) arbitrary data plugins can add
+      • {namespace}?            (`integer`)
 
 *vim.diagnostic.GetOpts*
     A table with the following keys:
@@ -596,6 +600,23 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
                               |nvim_buf_set_extmark()|.
       • {virt_text_win_col}?  (`integer`) See |nvim_buf_set_extmark()|.
       • {virt_text_hide}?     (`boolean`) See |nvim_buf_set_extmark()|.
+
+*vim.diagnostic.RelatedInformation*
+    Related message and location for a diagnostic. This can be used to point
+    to code locations that cause or are related to a diagnostic, e.g when
+    duplicating a symbol in a scope.
+
+    Fields: ~
+      • {bufnr}?     (`integer`) Buffer number
+      • {lnum}       (`integer`) The starting line of the related information
+                     (0-indexed)
+      • {end_lnum}?  (`integer`) The final line of the related information
+                     (0-indexed)
+      • {col}        (`integer`) The starting column of the related
+                     information (0-indexed)
+      • {end_col}?   (`integer`) The final column of the related information
+                     (0-indexed)
+      • {message}    (`string`) The related information text
 
 
 config({opts}, {namespace})                          *vim.diagnostic.config()*
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -114,6 +114,9 @@ LSP
 • Completion side effects (including snippet expansion, execution of commands
   and application of additional text edits) is now built-in.
 • |vim.lsp.util.locations_to_items()| sets `end_col` and `end_lnum` fields.
+• Add a new `related_information` field to |vim.Diagnostic| in favor of
+  `vim.Diagnostic.user_data.lsp.relatedInformation` (now deprecated). See
+  |vim.diagnostic.RelatedInformation|.
 
 LUA
 
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -29,6 +29,9 @@ local M = {}
 --- The diagnostic text
 --- @field message string
 ---
+--- An array of related diagnostic information
+--- @field related_information? vim.diagnostic.RelatedInformation[]
+---
 --- The source of the diagnostic
 --- @field source? string
 ---
@@ -41,6 +44,29 @@ local M = {}
 --- @field user_data? any arbitrary data plugins can add
 ---
 --- @field namespace? integer
+
+--- Related message and location for a diagnostic.
+--- This can be used to point to code locations that cause or are related to
+--- a diagnostic, e.g when duplicating a symbol in a scope.
+--- @class vim.diagnostic.RelatedInformation
+---
+--- Buffer number
+--- @field bufnr? integer
+---
+--- The starting line of the related information (0-indexed)
+--- @field lnum integer
+---
+--- The final line of the related information (0-indexed)
+--- @field end_lnum? integer
+---
+--- The starting column of the related information (0-indexed)
+--- @field col integer
+---
+--- The final column of the related information (0-indexed)
+--- @field end_col? integer
+---
+--- The related information text
+--- @field message string
 
 --- Many of the configuration options below accept one of the following:
 --- - `false`: Disable this feature
--- a/runtime/lua/vim/lsp/diagnostic.lua
+++ b/runtime/lua/vim/lsp/diagnostic.lua
@@ -97,6 +97,38 @@ local function tags_lsp_to_vim(diagnostic, client_id)
   return tags
 end
 
+---@param related_informations lsp.DiagnosticRelatedInformation[]
+---@param offset_encoding string
+---@return vim.diagnostic.RelatedInformation[]
+local function related_information_lsp_to_vim(related_informations, offset_encoding)
+  return vim
+    .iter(related_informations or {})
+    :map(
+      --- @param related_information lsp.DiagnosticRelatedInformation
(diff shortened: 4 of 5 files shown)


runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -110,6 +110,14 @@ local function diagnostic_lsp_to_vim(diagnostics, bufnr, client_id)
   return vim.tbl_map(function(diagnostic)
     local start = diagnostic.range.start
     local _end = diagnostic.range['end']
+    for _, relatedInfo in pairs(diagnostic.relatedInformation or {}) do
```
wookayin, line 113: `relatedInformation` is an array, so it looks like `ipairs` instead of `pairs` would be better?

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -110,6 +110,14 @@ local function diagnostic_lsp_to_vim(diagnostics, bufnr, client_id)
...
+      if relatedInfo.location then
```
wookayin, line 114: `location` seems to be a non-nil (required) field of the `lsp.DiagnosticRelatedInformation` class.

mfussenegger:
I think instead of extending the message this should go into the structure, otherwise users can't customise this. (And I'd wager there's a bunch who don't want this in the message).

There are also actions where we send the diagnostic back to the server and I think for the server to be able to re-associate the payloads with their internal state the diagnostic mustn't change.

tom-anders:
> I think instead of extending the message this should go into the structure, otherwise users can't customise this. (And I'd wager there's a bunch who don't want this in the message).
> 
> There are also actions where we send the diagnostic back to the server and I think for the server to be able to re-associate the payloads with their internal state the diagnostic mustn't change.

Sounds fair - I changed the PR so that now it only adds the `relatedInformation` field to the toplevel of `vim.Diagnostic`. We could then decide in a follow-up PR if and how we should display this to the user.

If this looks good, I'll add a test to `diagnostic_spec.lua` that covers this new field.

tom-anders:
> `vim.Diagnostic` is not an "LSP diagnostic" though, and we try very hard to avoid imposing LSP specific concepts into the (generic) diagnostic structure. In this case it is probably fine since the "related information" seems pretty generic (could be used by non-LSP diagnostic producers).
> 
> If this does need to be in the top level structure it should follow the standard naming conventions (`related_information`, not `relatedInformation`). The latter style indicates that it is a field that maps directly to some LSP concept.

Yeah to me this seems general enough (I could imagine something like overseer.nvim make use of this). Thanks for the hint related to naming conventions, fixed!

runtime/lua/vim/diagnostic.lua
```diff
@@ -42,6 +45,29 @@ local M = {}
 ---
 --- @field namespace? integer
 
+--- Related message and location for a diagnostic.
+--- This can be used to point to code locations that cause or are related to
+--- a diagnostics, e.g when duplicating a symbol in a scope.
```
gpanders, line 50:
```suggestion
--- a diagnostic, e.g when duplicating a symbol in a scope.
```

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -97,6 +97,33 @@ local function tags_lsp_to_vim(diagnostic, client_id)
...
+      local _end = related_information.location.range['end']
+      --- @type vim.DiagnosticRelatedInformation
+      return {
+        lnum = start.line,
+        col = line_byte_from_position(buf_lines, start.line, start.character, offset_encoding),
+        end_lnum = _end.line,
+        end_col = line_byte_from_position(buf_lines, _end.line, _end.character, offset_encoding),
+        message = related_information.message,
+      }
+    end, related_informations))
```
gpanders, line 124: Chaining a map with a filter is a good use case for `vim.iter`. `vim.iter(related_informations):map()` gives you filter+map semantics combined (if the callback function returns `nil` then the item is filtered out).

Although, is the original filter for `item ~= nil` even necessary? Presumably `related_informations` is an array, so it should not have any `nil` items in it.

tom-anders: The intention was to handle the `if not bufnr then return end` path - But thanks for the hint about using `vim.iter(..):map` (the rustacean in me would've expected this to be called `filter_map` though)

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -97,6 +97,33 @@ local function tags_lsp_to_vim(diagnostic, client_id)
   return tags
 end
 
+---@param related_informations lsp.DiagnosticRelatedInformation[]
+---@param offset_encoding string
+---@return vim.DiagnosticRelatedInformation[]
+local function related_information_lsp_to_vim(related_informations, offset_encoding)
+  return vim.iter(related_informations):map(
+  --- @param related_information lsp.DiagnosticRelatedInformation
+  --- @return vim.DiagnosticRelatedInformation
```
gpanders, line 106:
```suggestion
    --- @param related_information lsp.DiagnosticRelatedInformation
    --- @return vim.DiagnosticRelatedInformation?
```

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -97,6 +97,33 @@ local function tags_lsp_to_vim(diagnostic, client_id)
...
+    function(related_information)
+      local fname = vim.uri_to_fname(related_information.location.uri)
+      local bufnr = vim.fn.bufadd(fname)
+      if not bufnr then
+        return
+      end
```
gpanders, line 112: It's functionally equivalent in this case but imo is slightly clearer (we explicitly return either `nil` or a `vim.DiagnosticRelatedInformation` on all code paths).

```suggestion
      local bufnr = vim.fn.bufadd(fname)
      if not bufnr then
        return nil
      end
```

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -97,6 +97,33 @@ local function tags_lsp_to_vim(diagnostic, client_id)
...
+      local _end = related_information.location.range['end']
+      --- @type vim.DiagnosticRelatedInformation
+      return {
+        lnum = start.line,
+        col = line_byte_from_position(buf_lines, start.line, start.character, offset_encoding),
+        end_lnum = _end.line,
+        end_col = line_byte_from_position(buf_lines, _end.line, _end.character, offset_encoding),
+        message = related_information.message,
+      }
+    end)
```
gpanders, line 124: You also need `:totable()` to collapse the iterator back into a table (analagous to `collect()` in Rust)

```suggestion
    end):totable()
```

tom-anders:
Thanks for the review! I extended the tests in `lsp/diagnostic_spec.lua` now.

gpanders (approved):

runtime/lua/vim/diagnostic.lua
```diff
@@ -42,6 +45,29 @@ local M = {}
 ---
 --- @field namespace? integer
 
+--- Related message and location for a diagnostic.
+--- This can be used to point to code locations that cause or are related to
+--- a diagnostic, e.g when duplicating a symbol in a scope.
+--- @class vim.DiagnosticRelatedInformation
```
wookayin, line 51: How about having a namescope, like `vim.diagnostic.RelatedInformation`?

tom-anders: Good point, done

runtime/doc/news.txt
```diff
@@ -271,6 +271,10 @@ The following new APIs and features were added.
   • |vim.lsp.buf.definition()|, |vim.lsp.buf.declaration()|,
     |vim.lsp.buf.type_definition()|, and |vim.lsp.buf.implementation()| now
     support the `loclist` field of |vim.lsp.ListOpts|.
+  • Add `vim.Diagnostic.related_information` field and fill it with
+    `lsp.DiagnosticRelatedInformation[]`. Deprecate
```
gpanders, line 275:
```suggestion
  • Add `vim.Diagnostic.related_information` field and fill it with
    `vim.diagnostic.RelatedInformation[]`. Deprecate
```

tom-anders: I intentionally wrote `lsp.DiagnosticRelatedInformation` here though, to highlight the new logic I added in `lsp/diagnostic.lua`

tom-anders:
(rebased to fix conflicts)

runtime/doc/news.txt
gpanders: `lsp.DiagnosticRelatedInformation[]` does not link to anything in Nvim's help docs so I'm not sure it's valuable to have here in the user facing news. `vim.diagnostic.RelatedInformation` does have a help tag though. Maybe we keep this simple and just say:

```suggestion
• Add a new `related_information` field to |vim.Diagnostic| in favor of
  `vim.Diagnostic.user_data.lsp.relatedInformation` (now deprecated). See
  |vim.diagnostic.RelatedInformation|.
```

gpanders:
Needs another rebase.

runtime/doc/news.txt
tom-anders: sounds good, done

mfussenegger (commented):
Looks like this is missing the reverse mapping from vim to lsp. (And I'm starting to think we could just keep the original payload in a _lsp field on the vim.Diagnostic table, to ensure we send the original to the server and avoid any conversion errors)

tom-anders:
> Looks like this is missing the reverse mapping from vim to lsp. (And I'm starting to think we could just keep the original payload in a _lsp field on the vim.Diagnostic table, to ensure we send the original to the server and avoid any conversion errors)

Thanks for the hint - We already have `vim.diagnostic.user_data.lsp` which stored a subset of the original LSP diagnostic, I now simply made it store the entire table instead. That should also take care of the conversion back to LSP (Since `diagnostic_vim_to_lsp` already merges the `user_data.lsp` table

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -118,17 +150,15 @@ local function diagnostic_lsp_to_vim(diagnostics, bufnr, client_id)
...
       _tags = tags_lsp_to_vim(diagnostic, client_id),
       user_data = {
-        lsp = {
-          -- usage of user_data.lsp.code is deprecated in favor of the top-level code field
-          code = diagnostic.code,
-          codeDescription = diagnostic.codeDescription,
-          relatedInformation = diagnostic.relatedInformation,
-          data = diagnostic.data,
-        },
+        lsp = diagnostic,
```
mfussenegger, line 161: Can you also update `diagnostic_vim_to_lsp` to make use of this?

tom-anders: See my top level comment: if I understand the code correctly, `diagnostic_vim_to_lsp` already makes use of this, since it already merges the contents of `user_data.lsp` into the returned table. (This is also why the unit tests currently fails, I fixed this now in the latest push)

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -97,6 +97,38 @@ local function tags_lsp_to_vim(diagnostic, client_id)
...
+---@param offset_encoding string
+---@return vim.diagnostic.RelatedInformation[]
+local function related_information_lsp_to_vim(related_informations, offset_encoding)
+  return vim
+    .iter(related_informations or {})
+    :map(
+      --- @param related_information lsp.DiagnosticRelatedInformation
+      --- @return vim.diagnostic.RelatedInformation?
+      function(related_information)
+        local fname = vim.uri_to_fname(related_information.location.uri)
```
mfussenegger, line 110: @gpanders what do you think of making the related_information property a iterable or table?

It would allow us to make this lazy, so that users not using the information wouldn't have to pay for the conversion costs here

runtime/doc/diagnostic.txt
```diff
@@ -597,6 +601,23 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
       • {virt_text_win_col}?  (`integer`) See |nvim_buf_set_extmark()|.
       • {virt_text_hide}?     (`boolean`) See |nvim_buf_set_extmark()|.
 
+*vim.diagnostic.RelatedInformation*
```
justinmk, line 604: any problem with using a less verbose name?
```suggestion
*vim.diagnostic.Related*
```

tom-anders:
Hi everyone, I shifted focus to other projects and currently don't have time to work on this PR anymore. If somebody wants to take over, feel free to do so

PR #28646, opened 2024-05-05
fix(defaults): make descriptions of diagnostic mappings up to date

echasnovski:
Also:

- Make code formatted more like the other file (options table is on single line if possible).
- Omit "Open a floating window " in descriptions as it feels a bit redundant.

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -184,26 +184,22 @@ do
   do
     vim.keymap.set('n', ']d', function()
       vim.diagnostic.goto_next({ float = false })
-    end, {
-      desc = 'Jump to the next diagnostic with the highest severity',
-    })
+    end, { desc = 'Jump to the next diagnostic' })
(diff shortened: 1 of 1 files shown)


gpanders (approved):

runtime/lua/vim/_defaults.lua
```diff
@@ -184,26 +184,22 @@ do
...
       vim.diagnostic.open_float()
-    end, {
-      desc = 'Open a floating window showing diagnostics under the cursor',
-    })
-
-    vim.keymap.set('n', '<C-W><C-D>', '<C-W>d', {
-      remap = true,
-      desc = 'Open a floating window showing diagnostics under the cursor',
-    })
+    end, { desc = 'Show diagnostics under the cursor' })
```
justinmk, line 195: This is correct. "Show" is a standard verb and doesn't need hyper-explanation

justinmk (approved):

Issue #28476, opened 2024-04-23
Defaults: map gd to CTRL-]


gpanders:
The default mapping for "jump to tag" is `<C-]>`, but this is awkward or difficult to type on some keyboards.

Nvim's LSP client sets `tagfunc` by default which enables jumping to a definition with `<C-]>`. `gd` is a very popular custom mapping for jumping to a definition, so mapping `gd` to `<C-]>` would essentially provide that mapping for free.

We could either do this universally, or only when an LSP buffer is attached. One downside of doing it universally is that if an LSP buffer is _not_ attached, and there are no valid tags files, then we lose the builtin `gd` behavior (which is, frankly, not very good, but some people may rely on it).

Note that we do not simply map `gd` to `vim.lsp.buf.definition()` by default because using `'tagfunc'` provides a lot of other useful functionality (opening a definition in a split, searching for identifiers with `:ts`, etc.).

**EDIT:** This is really only useful if mapping `gd` to `<C-]>` also gets us `<C-W>gd` "for free", which I'm not sure if it does.

altermo:
We could also do `<C-]>` when either `'tagfunc'` is defined or when `len(tagfiles())` is over 0, and otherwise `gd`.
<details><summary>Example keymap</summary>

```lua
vim.keymap.set('n', 'gd', function()
    if vim.o.tagfunc ~= '' or #vim.fn.tagfiles() > 0 then
        return '<C-]>'
    end
    return 'gd'
end, { expr = true })
```

</details> 

Or we also could do that if `<C-]>` errors with `E426: Tag not found` or `E433: No tags file` the fallback to `gd`, though this may require som C code changes.

mfussenegger:
Note that we currently also have the unmapped:

- `vim.lsp.buf.declaration()`
- `vim.lsp.buf.implementation()`
- `vim.lsp.buf.type_definition()`

all of which also fit into the `gd` / `gD` area. My vote would go towards using the `gd` and `gD` to cover these by default instead of having two ways for go-to-definition.

bluss:
Consider doing this for help and its impact on help? This would increase the usability of help even though that kind of major change will take a long time to establish itself. It's a breaking change there too, and fundamental docs (the help landing page and `|bars|`) would be updated to mention it in that case.

clason:
> Consider doing this for help and its impact on help?

Help already has `K` mapped to it.

gpanders:
[...]

PR #28650, opened 2024-05-06
feat(defaults): add LSP default mappings (again)

gpanders:
This is the PR to re-introduce default LSP mappings after being reverted in #28649. The goal is to merge this early in the 0.11 release cycle.

I've changed the defaults to use the `gl` prefix rather than `cr`. This avoids all of the messiness with operator-pending mode that was discussed in #28634. We had originally wanted to "reserve" `gl` for a future "align" feature, but using it as the general purpose "LSP prefix" might be more high leverage (and we can always use `gla` or `ga` for "align", falling back to `:ascii` for the default `ga` behavior).

We can still revisit these mappings (I think in particular the `<C-S>` mapping in Insert mode is a bit controversial), but at _some point_ an executive decision will need to be made. We can't please everyone with defaults, so some people are going to be unhappy, unfortunately, but the _hope_ is that providing reasonable defaults makes the onboarding experience easier for more users.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -61,6 +61,16 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
+                                                *grr* *gra* *grn* *i_CTRL-S*
+Some keymaps are created unconditionally when Nvim starts:
+- "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
+- "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
+- "grr" is mapped in Normal mode to |vim.lsp.buf.references()| |gr-default|
+- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
+
+If not wanted, these keymaps can be removed at any time using
+|vim.keymap.del()| or |:unmap|.
+
                                                         *lsp-defaults-disable*
 To override the above defaults, set or unset the options on |LspAttach|: >lua
 
@@ -80,9 +90,6 @@ Example: >lua
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(args)
         local client = vim.lsp.get_client_by_id(args.data.client_id)
-        if client.supports_method('textDocument/rename') then
-          -- Create a keymap for vim.lsp.buf.rename()
-        end
         if client.supports_method('textDocument/implementation') then
           -- Create a keymap for vim.lsp.buf.implementation
         end
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -65,7 +65,7 @@ options are not restored when the LSP client is stopped or detached.
 Some keymaps are created unconditionally when Nvim starts:
 - "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
-- "grr" is mapped in Normal mode to |vim.lsp.buf.references()| |gr-default|
+- "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
 - CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
 
 If not wanted, these keymaps can be removed at any time using
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -69,7 +69,7 @@ Some keymaps are created unconditionally when Nvim starts:
 - CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
 
 If not wanted, these keymaps can be removed at any time using
-|vim.keymap.del()| or |:unmap|.
+|vim.keymap.del()| or |:unmap| (see also |gr-default|).
 
                                                         *lsp-defaults-disable*
 To override the above defaults, set or unset the options on |LspAttach|: >lua
--- a/runtime/doc/change.txt
+++ b/runtime/doc/change.txt
@@ -281,10 +281,11 @@ gr{char}		Replace the virtual characters under the cursor with
 			that have a special meaning in Insert mode, such as
 			most CTRL-keys, cannot be used.
 
+							*gr-default*
 			Nvim creates default mappings with "gr" as a prefix,
-			which may inhibit the behavior of |gr|. Use
-			|vim.keymap.del()| or |:unmap| to delete the default
-			mappings.
+			which may inhibit the behavior of |gr|. Use the
+			following to restore the builtin behavior: >
+				nnoremap <nowait> gr gr
 
 						*digraph-arg*
 The argument for Normal mode commands like |r| and |t| is a single character.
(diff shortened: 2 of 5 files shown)


runtime/doc/lsp.txt
```diff
@@ -61,6 +61,16 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
+                                                *gll* *glr* *gln* *i_CTRL-S*
```
justinmk, line 64:
```suggestion
                                                *grl* *grr* *grn* *i_CTRL-S*
```

gpanders: Oops, I actually _did_ mean for the mappings to be `gl` prefixed, not `gr`, but I forgot to change them in `_defaults.lua`.

gpanders:
Argument against `gr` prefix is that if a user does want to restore the builtin `gr` behavior, they would have to delete _all_ of the mappings under the `gr` prefix (although I suppose the same argument applies to anyone using a plugin that currently maps `gl`, so maybe that's actually not a very strong argument).

zeertzjq:
I don't think the same argument applies to plugins that map `gl`. The builtin `gr` can be following by any character, but that's not likely to apply to plugins that map `gl`.

gpanders:
> I don't think the same argument applies to plugins that map `gl`. The builtin `gr` can be following by any character, but that's not likely to apply to plugins that map `gl`.

Ah I see what you mean. I am referring to the fact that if we create new `gl` prefixed keys, then pressing `gl` will require a delay to use the `gl` behavior provided by the plugin. If a user wants the mapping to work as before, they have to delete all other keymaps that are prefixed with `gl` (or change the mapping to `gla` or something, which is probably what I will do if this goes through).

justinmk (changes requested):
gr

> Argument against `gr` prefix is that if a user does want to restore the builtin `gr` behavior, they would have to delete _all_ of the mappings under the `gr` prefix

That needs to be solved some other way. It is a temporary problem.

- "r" ("refactor") is a more appropriate mnemonic than "l" ("language"). 
- "gl" will be useful for other purposes. Whereas there is no alternative, planned use-case for "gr".

> `gl` is easier to type

"grr" and "gra" can be typed with one hand on querty, which is relevant for refactoring-related mappings which may involve the mouse.

famiu:
> Whereas there is no alternative, planned use-case for "gr".

"gr" could be used for LSP references

gpanders:
> > Whereas there is no alternative, planned use-case for "gr".
> 
> "gr" could be used for LSP references

`grr` is LSP references. `gr` becomes the prefix for "code refactor" type things (including "find references").

Current proposal is:

`grr` -> `vim.lsp.buf.references()`
`gra` -> `vim.lsp.buf.code_action()` (works in Visual mode too)
`grn` -> `vim.lsp.buf.rename()`

In Insert mode:

`<C-S>` -> `vim.lsp.buf.signature_help()`

justinmk (approved):

justinmk:
> > Whereas there is no alternative, planned use-case for "gr".
> 
> "gr" could be used for LSP references

That is completely missing the point.

deathbeam:
`gr` is very commonly used for LSP references (see for example https://www.lazyvim.org/keymaps#lsp, https://github.com/nvim-lua/kickstart.nvim/blob/master/init.lua#L475 its also recommended mapping by bunch of other stuff). and `g` as prefix is often used as goto (see native gg, gd, gD) so stuff like code actions, rename (and possibly) format imo do not rly makes sense behind it.

Imo the lsp refactor stuff and lsp goto stuff should have separate prefixes as its 2 very different operations but thats just my 2 cents.

gpanders:
>and g as prefix is often used as goto (see native gg, gd, gD)

The `g` prefix does not mean goto. `gs`, `gp`, `ga`, `g8`, `g?`, `gJ`, `gu`, `gU`, `gv`, `gq`, `gw`, and `g~` are all counter examples.

deathbeam:
> > Imo the lsp refactor stuff and lsp goto stuff
> 
> "LSP goto stuff" is just `'tagfunc'`, which uses the builtin mappings from `:h tag`.
> 

Well LSP goto stuff can also include implementation, type definiton, defition, declaration, which could possibly also have mappings.

> > gr is very commonly used for LSP references
> 
> The proposal here is to use `grr` for references. Is moving from `gr` to `grr` that much of a problem?

Well is there any reason to not use the mappings that ppl are already used to? The mappings are unused and LSP being integral part of neovim I think justifies not putting the whole thing behind single prefix (especially when `gr` doesnt even mean anything here that could be used as mnemonic)

gpanders:
>Well is there any reason to not use the mappings that ppl are already used to?

Yes, because default mappings have much greater constraints than arbitrary user mappings. If we make `gr` map to references, we cannot use `gr` as a prefix for anything else. So instead of `grr`, `gra`, and `grn` (3 default mappings), we have `gr` (1 mapping) and have to find separate mappings for the other two. Finding unused default mappings is _extremely_ difficult (if you've been following the "new LSP defaults" saga then you will have observed this).

I will be 100% clear about this (and I will probably end up repeating myself): it is not possible, nor is there any expectation, that we will find default mappings that make everyone happy. The goal is to provide a better out of the box experience for new and less experienced users by providing access to builtin Neovim capabilities without needing much (or any) additional configuration. Experienced and power users will likely have separate preferences, but they already know how to create and delete mappings, so they are free to ignore the defaults.

justinmk (approved):

runtime/doc/lsp.txt
```diff
@@ -61,6 +61,16 @@ options are not restored when the LSP client is stopped or detached.
...
 
+                                                *grr* *gra* *grn* *i_CTRL-S*
+Some keymaps are created unconditionally when Nvim starts:
+- "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
+- "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
+- "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
+- CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
+
+If not wanted, these keymaps can be removed at any time using
+|vim.keymap.del()| or |:unmap|.
```
justinmk, line 72: for `gr` we may want to mention `nnoremap <nowait> gr gr`

gpanders: That doesn't seem to work. If I use that mapping then I'm not able to use any of the new defaults. It reverts to the builtin `gr` behavior.

gpanders: If I use `vim.keymap.set('n', 'gr', 'gr', { nowait = true })` _before_ creating the other `gr` mappings, then I can still use the new mappings, but it doesn't seem to behave any differently than it does without the `<nowait>` mapping. So I'm not sure what it's meant to be doing.

zeertzjq: > That doesn't seem to work. If I use that mapping then I'm not able to use any of the new defaults. It reverts to the builtin `gr` behavior.

That's exactly what is supposed to do: conveniently disable all these mappings.

gpanders: Ah that makes sense, I misunderstood the original suggestion. I’ll include that

gpanders: 78f2c0749c9f4b588b4154e585233e6c5aa76e1e

justinmk (approved):

bluebrown:
This is not setting `definition` Is this already covered by something else? Otherwise, imo, `definition` and `references` are the most important ones.

What determines the selection of these mappings?

arcxio:
> Is this already covered by something else

it's C-] by default. there is a proposal to map it to gd in #28476

PR #28659, opened 2024-05-07
refactor(options): generate available options values from lua

glepnir:
Problem: currently the available values of options is split from options meta lua file.

Solution: generate it from lua file.

cc @justinmk


src/nvim/options.lua
```diff
@@ -30,6 +30,7 @@
 --- @field cb? string
 --- @field expand_cb? string
 --- @field tags? string[]
+--- @field available_values? table
```
justinmk, line 33: Is there a more conventional name that can be used? There are many such examples in the codebase.

src/nvim/generators/gen_options.lua
```diff
@@ -228,6 +231,48 @@ local function dump_option(i, o)
       w('#endif')
     end
   end
+
+  if gen_values and o.available_values then
```
justinmk, line 235: probably worth extracting this logic to a function

src/nvim/options.lua
famiu: I would prefer `valid_values`

src/nvim/generators/gen_options.lua
```diff
@@ -1,12 +1,15 @@
 local options_file = arg[1]
 
+local gen_values = options_file:find('options_values') and true or false
+
```
famiu, line 4: I don't like this pattern of using the same generator for 2 completely different purposes. Use a separate generator file for this, similar to what `gen_options_enum` does

justinmk: These are not "complete different purposes". And why is `gen_options_enum.lua` in a separate file?

famiu: > These are not "complete different purposes".

The script acts does two very different things depending on which file it's being called on, that seems different to me. A script's behavior depending on the filename is certainly not ideal.

> And why is `gen_options_enum.lua` in a separate file?

Precisely because it does a completely different thing. `gen_options` generates the `options[]` array, `gen_options_enum` generates the option index enum (using information from the `options[]` array, so order matters here) and the hash table for `findoption()`

glepnir: this PR has not been processed for a long time, some code may be outdated. the conflict is resolved but seems like test failed. If you have new ideas, you can continue this PR.

famiu:
@glepnir I have ideas on how we can expand this PR to also auto-generate option variables (and the option flag macros). Since you have given permission, I am working on a PR that supersedes this one.

Thank you for your work on this PR, though, as it has been immensely useful and has given me a good starting point to work with.

PR #28663, opened 2024-05-07
spec.vim: Add keywords from newer rpm builds

berolinux:
Add the BuildSystem and BuildOption headers (rpm 4.20), conf, generate_buildrequires, sourcelist and patchlist sections (rpm 4.18), and the macros for working with cmake, meson and ninja build systems


clason:
Thank you, but we don't maintain these files. Please contact the maintainer listed at the top of the file.

PR #28665, opened 2024-05-08
Update snippet.lua

chiendo97:
Update the runtime doc to use new `active` instead of removed `jumpable`


zeertzjq:
Duplicate of #28607

Issue #28667, opened 2024-05-08
0.10 colorscheme flickering


sarmong:
### Problem

There are a few conditions when colorscheme flickers when launching neovim. By flickering I mean that it between changin background from terminal background to neovim background, there is a period of time when the background is black.

I think it is somewhat related to #19362 . These only reproduce on nightly, 0.9.5 is fine.

### Steps to reproduce

For any of this, run `nvim --clean -u minimal.lua` with the contents of `minimal.lua` described below.

#### termguicolors

After setting `termguicolor`, background switches to black before switching to the user-defined one. Run with `sleep` to pause on black background, or with some synchronous code to see flickering.

I know that `termguicolors` is enabled automatically in 0.10, however this causes 0.9 configs to flicker and people waste time debugging. Also, I don't know how this works under the hood, but automatic setting of `termguicolors` seems to happen after `init.lua` is loaded, because `vim.o.termguicolors` returns `false` if printed inside user config (when not set manually).

```lua
vim.opt.termguicolors = true
-- vim.cmd.sleep(1)
vim.fn.system("notify-send loading")

vim.cmd.colorscheme("blue")
```

#### VimEnter

If you have a VimEnter autocommand with any synchronous action, the same thing happens. There are 3 cases. 

If you run (1) - you will see the same effect as with `termguicolors` - background switches to black. 
If you run (2) - you may see very subtle flickering of the screen. I choose `fn.system` as a generic synchronous side-effect. The same happens with `system():wait`.
If you run (3) - no flickering. The same is described in https://github.com/neovim/neovim/issues/19362#issuecomment-1924993842

```lua
vim.api.nvim_create_autocmd("VimEnter", {
	callback = function()
		--[[ (1) ]] vim.cmd.sleep(1)
		--[[ (2) ]] vim.fn.system("notify-send vimenter")
		--[[ (3) ]] vim.print("asdf")
	end,
})
vim.cmd.colorscheme("blue")
```

#### unnamedplus

Same flickering, however you cannot pause on it by running `sleep`, maybe it happens at different point in time. But, if you run synchronous commands, then you can see the flickering.

```lua
vim.opt.clipboard = "unnamedplus"
vim.fn.system("notify-send loading")
vim.cmd.colorscheme("blue")
```

### Expected behavior

No flickering like on 0.9.5. 

### Neovim version (nvim -v)

0.10.0-dev-3097+g064f3e42e

### Operating system/version

Linux/X11

### Terminal name/version

[...]

Issue #21018, opened 2022-11-10
setting guicursor too quickly leads to unexpected results


folke:
### Neovim version (nvim -v)

nightly

### Vim (not Nvim) behaves the same?

not relevant

### Operating system/version

linux

### Terminal name/version

kitty

### $TERM environment variable

xterm-kitty

### Installation

bob

### How to reproduce the issue

```lua
vim.go.guicursor = "" -- uses a block cursor for all modes

-- this will effectively hide the cursor
vim.api.nvim_set_hl(0, "NoiceHiddenCursor", { blend = 100, nocombine = true })
vim.go.guicursor = "a:NoiceHiddenCursor/NoiceHiddenCursor"

-- reset the properties for all modes
vim.go.guicursor = "a:"

-- restore the original cursor behavior
vim.go.guicursor = ""
```

### Expected behavior

Cursor should be restored

### Actual behavior

The cursor remains hidden.

When restoring the cursor with defer, it does work as expected:

```lua
vim.go.guicursor = ""
vim.api.nvim_set_hl(0, "NoiceHiddenCursor", { blend = 100, nocombine = true })
vim.go.guicursor = "a:NoiceHiddenCursor/NoiceHiddenCursor"

vim.defer_fn(function()
  vim.go.guicursor = "a:"
  vim.defer_fn(function()
    vim.go.guicursor = ""
  end, 100)
end, 100)
```

folke:
Related https://github.com/folke/noice.nvim/issues/114

zeertzjq:
I cannot reproduce this. This works as expected for me.

folke:
@zeertzjq I've updated the code above. You can test it with `nvim -u NONE`.

zeertzjq:
You can add a redraw between the two:
```lua
vim.cmd[[set termguicolors]]

-- make the cursor red
vim.api.nvim_set_hl(0, "NoiceHiddenCursor", { bg = "red" })
vim.go.guicursor = "a:NoiceHiddenCursor/NoiceHiddenCursor"

-- restore the cursor after 1 second
vim.defer_fn(function()
  -- reset all mode attribute
  vim.go.guicursor = "a:"
  vim.cmd.redraw()
  -- set cursor to block cursor in all modes
  vim.go.guicursor = ""
end, 1000)
```

folke:
ah perfect. Definitely better than using defer. Thanks!

zeertzjq:
Related: #6432

[...]

PR #28669, opened 2024-05-08
fix(redraw): only flush screen state when "flush" is true

### LINKED ISSUE #28668: nvim__redraw causes some visible screen tearing
### Problem

Since https://github.com/neovim/neovim/commit/037ea6e786b5d05f4a8965e4c2ba6aa60ec7c01a, Nvim occasionally exhibits some black "tearing"/flickering on startup.

It is difficult to capture so I recorded a slow motion video on my phone:

**Before (no black tearing)**:
 
https://github.com/neovim/neovim/assets/8965202/ffd619d5-7714-43be-8387-409b80962ff7

**After (with black tearing)**:

https://github.com/neovim/neovim/assets/8965202/468c8800-2d34-4003-b7bb-975ddb288905

Note that in both cases we can see the screen state being drawn incrementally, but in the latter video the background of the top half of the screen becomes black first. In real time this has a noticeable "flicker" effect.

### Steps to reproduce

In my case this seems to be caused by the `nvim__redraw` call in the Tree-sitter highlighter, so the following is sufficient to reproduce the issue:

```
cat >test.vim <<EOF
colorscheme blue
au FileType c lua vim.treesitter.start()
EOF
nvim --clean -u test.vim src/nvim/main.c
```

### Expected behavior

The background should not become black when redrawing.

### Neovim version (nvim -v)

064f3e42e8d33b0a9e560dfb7c9a42b2fc1ed868

No

### Operating system/version

macOS

### Terminal name/version

Ghostty

xterm-ghostty

### Installation

Source

gpanders:
Fixes https://github.com/neovim/neovim/issues/28668.

### DIFF
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -2456,8 +2456,8 @@ void nvim__redraw(Dict(redraw) *opts, Error *err)
   // redrawing a status component may have changed the grid dimensions.
   if (flush && !cmdpreview) {
     update_screen();
+    ui_flush();
   }
-  ui_flush();
 
   RedrawingDisabled = save_rd;
   p_lz = save_lz;


luukvbaal:
I agree it makes more sense inside the `flush` guard. Don't remember why I placed it outside tbh, can reconsider the specific use case if this was ext_ui driven and I run into it again. I do remember wanting to reconsider this anyways but it was merged as experimental and I forgot about it.

I do wonder if the treesitter highlighter wants to set, `flush == true` then. I think this was an improvement: https://github.com/neovim/neovim/pull/28101#issuecomment-2091057965.

luukvbaal:
Altough with this change status components may not be flushed to the screen then with `flush == false` which may be unexpected? `:redrawstatus` also calls `ui_flush()`. Perhaps it's fine though when `flush` is documented in proximity.

gpanders:
> I do wonder if the treesitter highlighter wants to set, `flush == true` then. I think this was an improvement: [#28101 (comment)](https://github.com/neovim/neovim/pull/28101#issuecomment-2091057965).

Hm if I set `flush = true` in the highlighter, then I still get the black flickering from https://github.com/neovim/neovim/issues/28668 even with this PR.

Perhaps it is possible to redraw without setting the background to black? That would maybe solve #28667 as well.

luukvbaal:
Yeah fixing the underlying issue seems necessary regardless. The following patch seems to prevent all flickering from #28667 and #28668. Whether it can also cause problems I refer to @bfredl, although it seems unlikely to me.

```diff
diff --git a/src/nvim/tui/tui.c b/src/nvim/tui/tui.c
index e1bb1e55e3..ad359f4c53 100644
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -1654,6 +1654,10 @@ void tui_raw_line(TUIData *tui, Integer g, Integer linerow, Integer startcol, In
 
 static void invalidate(TUIData *tui, int top, int bot, int left, int right)
 {
+  if (starting != 0) {
+    return;
+  }
+
   Rect *intersects = NULL;
 
   for (size_t i = 0; i < kv_size(tui->invalid_regions); i++) {
```

I'm fine with this PR if this needs to be merged quick for 0.10 but I would like to hear @bfredl @zeertzjq thought's on what to do with the `ui_flush()` in `nvim__redraw()` considering my comments above.

zeertzjq:
Hmm, I also noticed that sometimes there is need to call `ui_flush()` alone, e.g., to flush `mode_info_set` (#21018) or to only flush pending messages.

zeertzjq (approved):
I think this should be merged for v0.10. This is most frequently used to redraw a range in a buffer (to replace `nvim___buf_redraw_range`) or to redraw cursor (to replace an FFI call to `setcursor_mayforce()`), neither of which expect an extra `ui_flush()`.

luukvbaal:
Fine with me, but can anyone acknowledge why the treesitter highlighter should or shouldn't `flush`? As I see it this adds in a `ui_flush()` that only improves responsiveness as soon as treesitter wants to add highlighting.

luukvbaal:
Upon reconsideration I'm fine with it if it's necessary for 0.10(although I think #28676 also fixes the problem this PR aims to fix), but I still believe the user doesn't necessarily need to know the distinction between flushing pending updates to the UI (`ui_flush()`) and pushing pending updates to the grid (`update_screen()`). `flush` was intended as the latter, i.e. pending window updates.

I also think the above statements aren't actually correct, nothing `nvim__redraw()` does actually becomes visible on the screen without calling `ui_flush()`, so why shouldn't it be called implicitly? `setcursor_mayforce()` just sets a pending cursor update, `ui_call_grid_cursor_goto()` actually moves the cursor. `range` without `ui_flush()` just marks the range for redraw, so that wouldn't match the current description either.

Under the premise "the user called `nvim__redraw()` and Nvim should call any code necessary to reflect the requested changes" I'm not sure implicitly calling `ui_flush()` is actually harmful. I think the fact that `:redraw/status/tabline` all call `ui_flush()` is also a good indication that `nvim__redraw()` should. So I think keeping it as is or adding an explicit `flush` for `ui_flush()`(renaming the current flush to `update`) makes more sense than this PR.

gpanders:
#28676 does indeed fix the problem this was intended to solve, so I leave it up to others to decide the fate of this PR (I am fine with either closing or merging).

zeertzjq:
This may indeed not be the best approach. Let's close this.

PR #28673, opened 2024-05-08
build: allow sccache as compiler cache

### LINKED ISSUE #28670: Use `sccache` for build process if avalible
### Problem

Neovim doesn't use [sccache](https://github.com/mozilla/sccache) as a possible caching tool in the compilation step.

### Expected behavior

Modify `CMakeLists.txt` so that the build process uses `sccache` if the binary exists.

dundargoc:
Also enable caching for dependencies.

Closes https://github.com/neovim/neovim/issues/28670


dundargoc:
TODO: write in my TODO.txt:

> I advise you to remove all channels (they are not great), clone nixpkgs and do `export NIX_PATH=nixpkgs=/home/dundargoc/nixpkgs`  then you know what version of nixpkgs your commands are running against. Or use flakes.

dundargoc:
TODO: test if this works with the environment variable.

clason:
Also, sccache claims to support Rust, so this might help with the wasmtime PR. Investigate?

(Also has GHA.)

dundargoc:
Running `hyperfine --prepare="make distclean" 'make DEPS_CMAKE_FLAGS=-DENABLE_WASMTIME=ON deps'`

```
Benchmark 1: no caching
  Time (mean ± σ):     90.766 s ±  0.274 s    [User: 353.839 s, System: 20.009 s]
  Range (min … max):   90.268 s … 91.192 s    10 runs

Benchmark 2: sccache
  Time (mean ± σ):     87.828 s ±  0.798 s    [User: 324.237 s, System: 18.397 s]
  Range (min … max):   86.917 s … 89.486 s    10 runs
```

clason:
(I think the Rust support is also quite restricted and requires specific flags.)

dundargoc:
Unsure if I need to pass some specific flags for `sccache` to work better, but right now it's so ineffective that I don't think we need to consider it for rust compilation.

dundargoc:
Followup PR: enable ccache/sccache on GHA.

PR #28676, opened 2024-05-09
fix(tui): initialize clear attrs with current terminal background

### LINKED ISSUE #28667: 0.10 colorscheme flickering

luukvbaal:
Problem:  Invalidated regions that are flushed during startup are
          cleared with unitialized "clear_attrs", which is perceived as
          flickering.
Solution: Initialize "clear_attrs" with current terminal background color.

Fix #28667

### DIFF
--- a/test/functional/terminal/tui_spec.lua
+++ b/test/functional/terminal/tui_spec.lua
@@ -2003,6 +2003,39 @@ describe('TUI', function()
     ]])
   end)
 
+  it('invalidated regions are cleared with terminal background attr', function()
+    local screen = Screen.new(50, 10)
+    screen:set_default_attr_ids({ [1] = { foreground = Screen.colors.Black } })
+    screen:attach()
+    fn.termopen({
+      nvim_prog,
+      '--clean',
+      '--cmd',
+      'set termguicolors',
+      '--cmd',
+      'sleep 10',
+    }, {
+      env = {
+        VIMRUNTIME = os.getenv('VIMRUNTIME'),
+      },
+    })
+    screen:expect({
+      grid = [[
+        {1:^                                                  }|
+        {1:                                                  }|*8
+                                                          |
+      ]],
+    })
+    screen:try_resize(51, 11)
+    screen:expect({
+      grid = [[
+        {1:^                                                   }|
+        {1:                                                   }|*9
+                                                           |
+      ]],
+    })
+  end)
+
   it('argv[0] can be overridden #23953', function()
     if not exec_lua('return pcall(require, "ffi")') then
       pending('missing LuaJIT FFI')
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -23,6 +23,7 @@
 #include "nvim/globals.h"
 #include "nvim/grid.h"
 #include "nvim/grid_defs.h"
+#include "nvim/highlight.h"
 #include "nvim/highlight_defs.h"
 #include "nvim/log.h"
 #include "nvim/macros_defs.h"
@@ -165,6 +166,15 @@ void tui_start(TUIData **tui_p, int *width, int *height, char **term, bool *rgb)
   tui->seen_error_exit = 0;
   tui->loop = &main_loop;
   tui->url = -1;
+  // Because setting the default colors is delayed until after startup to avoid
+  // flickering with the default colorscheme background, any flush that happens
+  // during startup in turn would result in clearing invalidated regions with
+  // uninitialized attrs(black). Instead initialize clear_attrs with current
+  // terminal background so that it is at least not perceived as flickering, even
+  // though it may be different from the colorscheme that is set during startup.
+  tui->clear_attrs.rgb_bg_color = normal_bg;
+  tui->clear_attrs.cterm_bg_color = (int16_t)cterm_normal_bg_color;
+
   kv_init(tui->invalid_regions);
   kv_init(tui->urlbuf);
   signal_watcher_init(tui->loop, &tui->winch_handle, tui);


luukvbaal:
Was unsuccessful in adding a test for this so far, can try again tomorrow if this fix seems right to you @bfredl.

gpanders:
I am not sure this fixes #28667. Using the init script mentioned in that file:

```vim
colorscheme blue
au VimEnter * sleep 1
```

with `VIMRUNTIME=runtime build/bin/nvim --clean -u init.vim`, I still see the black screen.

luukvbaal:
Oh? I don't see the black screen. Are you actually seeing a black screen or is your terminal background black and is that what you're seeing? That's what I see and what I would expect after this change.

luukvbaal:
> Hmm seems terminal dependent, I do still get a black screen in wezterm, not in st or kitty. Not sure about the fix then.

The following change also gets rid of the black screen in wezterm, guess I'll change it to that for now.

```diff
diff --git a/src/nvim/tui/tui.c b/src/nvim/tui/tui.c
index 4c87612e87..7f8a92b3c0 100644
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -1438,7 +1438,7 @@ void tui_flush(TUIData *tui)
   while (kv_size(tui->invalid_regions)) {
     // We don't want to flush invalid regions during startup to avoid clearing
     // with attributes different from the visible background after startup.
-    if (tui->is_starting) {
+    if (starting != 0) {
       kv_size(tui->invalid_regions) = 0;
       break;
     }
```

gpanders:
>EDIT: Hmm seems terminal dependent, I do still get a black screen in wezterm, not in st or kitty. Not sure about the fix then

Weird, I still see the black screen on every terminal I've tried (Ghostty, Alacritty, WezTerm, and Kitty).

>Are you actually seeing a black screen or is your terminal background black and is that what you're seeing?

My terminal background color is #2e3440, which is not black. The "black" is, I believe, the default colorscheme background, which is #14161b.

luukvbaal:
> Weird, I still see the black screen on every terminal I've tried (Ghostty, Alacritty, WezTerm, and Kitty).

Probably timing related instead then(for me the results are consistent depending on the terminalon a relatively old laptop)...

> My terminal background color is #2e3440, which is not black. The "black" is, I believe, the default colorscheme background, which is #14161b.

No I don't think it is the default colorscheme background in this case(the ones from #28667 and #28668). What we're seeing is actually black(#000000) because the `clear_attr` becomes `0` when the grid is completely undrawn during startup.

luukvbaal:
Updated to avoid flushing in a way that shouldn't be inter-process timing dependent, while still avoiding any flickering without breaking any existing expected results. Still not sure how to add a test for this...

test/functional/terminal/tui_spec.lua
```diff
@@ -2003,6 +2003,39 @@ describe('TUI', function()
     ]])
   end)
 
+  it('invalidated regions are cleared with terminal background attr', function()
+    local screen = Screen.new(50, 10)
+    screen:set_default_attr_ids({ [1] = { foreground = Screen.colors.Black } })
```
luukvbaal, line 2008: Added a test that at least fails on master, although I'm not sure if it tests for the behavior correctly. Presumably `Screen.colors.Black` **is** `normal_bg`, in which case it does. Could also mean that the test still produces an unwanted black screen here? I can no longer reproduce that locally.

gpanders:
Confirmed this fixes both #28667 and #28668 for me.

gpanders (approved):

PR #28678, opened 2024-05-09
docs(gen_help_html.lua): wrap legacy help at word-boundary

gusain71:
Problem:

On the page: https://neovim.io/doc/user/dev_vimpatch.html
The links extend beyond the container and thus end up behind the navigation to the right.

Solution: 

Adding these lines to get_help_html.lua solves the problem:
white-space: normal;
word-wrap: break-word;

Before:
Skärmavbild 2024-05-10 kl  13 33 46

After:
Skärmavbild 2024-05-10 kl  13 35 21

### DIFF
--- a/scripts/gen_help_html.lua
+++ b/scripts/gen_help_html.lua
@@ -1094,8 +1094,7 @@ local function gen_css(fname)
       word-wrap: break-word;
     }
     .old-help-para pre {
-      /* All text in .old-help-para is formatted as "white-space:pre" so text following <pre> is
-         already visually separated by the linebreak. */
+      /* Text following <pre> is already visually separated by the linebreak. */
       margin-bottom: 0;
     }
 
--- a/scripts/gen_help_html.lua
+++ b/scripts/gen_help_html.lua
@@ -1088,9 +1088,10 @@ local function gen_css(fname)
       padding-bottom: 10px;
       /* Tabs are used for alignment in old docs, so we must match Vim's 8-char expectation. */
       tab-size: 8;
-      white-space: pre;
+      white-space: normal;
       font-size: 16px;
       font-family: ui-monospace,SFMono-Regular,SF Mono,Menlo,Consolas,Liberation Mono,monospace;
+      word-wrap: break-word;
     }
     .old-help-para pre {
       /* All text in .old-help-para is formatted as "white-space:pre" so text following <pre> is


wookayin:
Thanks for your contribution on improving docs! It would be appreciated if you could [describe the problem] this PR is going to fix, and/or [describe why this change would be needed](https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md#commit-messages). Screenshots (before v.s. after) would also work in this particular case.

gusain71:
> Thanks for your contribution on improving docs! It would be appreciated if you could [describe the problem](https://lamport.azurewebsites.net/pubs/state-the-problem.pdf) this PR is trying to fix, and/or [provide more explanation about why this change would be needed](https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md#commit-messages). Screenshots (before v.s. after) would also work in this particular case.

Description together with before and after images have been added.

scripts/gen_help_html.lua
```diff
@@ -1088,9 +1088,10 @@ local function gen_css(fname)
...
       /* Tabs are used for alignment in old docs, so we must match Vim's 8-char expectation. */
       tab-size: 8;
-      white-space: pre;
+      white-space: normal;
       font-size: 16px;
       font-family: ui-monospace,SFMono-Regular,SF Mono,Menlo,Consolas,Liberation Mono,monospace;
+      word-wrap: break-word;
     }
     .old-help-para pre {
       /* All text in .old-help-para is formatted as "white-space:pre" so text following <pre> is
```
wookayin, line 1097: One minor thing: The comment here doesn't apply here (see #24151), because we no longer have `white-space: pre`, though apparently we'll still need `margin-bottom: 0` in both cases.

justinmk: yes this needs to be fixed.
```suggestion
      /* Text following .old-help-para <pre> is
```

gusain71: I believe that I have updated the pull request with these changes. If I have misinterpreted anything, please let me know.

justinmk:
We can try this. But the purpose of not wrapping was to avoid wrapping codeblocks and other lines that should not be wrapped. The fact that the right-column menu occasionally gets in the way is not a big deal, but people seem to care about it... (why does it matter if you see the last bits of a gigantic URL?)

dundargoc:
> why does it matter if you see the last bits of a gigantic URL?

I don't understand how to answer this question. It's not possible to stack text on top of each other and read both texts at the same time. The documentation on mobile is impossible to read because the stupid bar overlaps all text.

clason:
And, more concretely, as the screenshot shows, it interferes with the navigation (TOC).

justinmk:
That could be solved by making the ToC opaque. But on big screens (or short ToC), there is no readability issue and no "looks like ass" issue.

clason:
Well, yes, if there is no issue, there is no issue.

PR #28679, opened 2024-05-09
Typo: INSTALL.md -> Updating listed 'major' version to match current choco version

asyncjake:
I was curious if 0.7 was really the latest on choco, but `choco info neovim` shows 0.9.5 at the time of writing. Checked the contributing guidelines and there wasn't much about readme typos, so apologies if this is too informal

Given that choco is fully up to date, is choco updated as regularly as the other distribution channels? If so that's awesome news for windows users 😄

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -36,7 +36,7 @@ Windows 8+ is required. Windows 7 or older is not supported.
 
 ### [Chocolatey](https://chocolatey.org)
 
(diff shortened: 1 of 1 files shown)


clason:
I would just drop the version number entirely; that'll bitrot the second you change it.

jakethedev:
Updated - I noticed debian is back at 0.4.x, but I wasn't sure what style of writing would be preferred for adding some kind of notice about that, so I'm just keeping it simple haha

dundargoc:

PR #28681, opened 2024-05-09
Add repology badge in right corner in INSTALL.md

Drsheppard01:


clason:
Thank you, but we already have this badge on the main README (where it fits better).

PR #28689, opened 2024-05-10
fix: remove vim.snippet.jumpable reference

mike325:


clason:
Thanks, but already fixed as part of https://github.com/neovim/neovim/pull/28609/files#diff-ab4560698af581a27397dea978b95c9654abe017147e4355941f50871a5685b3

PR #28693, opened 2024-05-10
vim-patch:9.1.{0401,0402} zsh module files are not recognized

clason:
- **vim-patch:9.1.0401: filetype: zsh module files are not recognized**
- **vim-patch:9.1.0402: filetype: mdd files detected as zsh filetype**


runtime/lua/vim/filetype/detect.lua
```diff
@@ -1146,7 +1146,9 @@ function M.proto(_, bufnr)
   -- comments, thus we can't match comments to see the difference.
   -- IDL files can have a single ';' in the second line, require at least one
   -- character before the ';'.
-  if getline(bufnr, 2):find('.;$') then
+  if getline(bufnr, 2):find('/%* Generated automatically %*/') then
```
zeertzjq, line 1149: The comments above should be moved below this condition or removed

clason: Oops, missed the comment change.

clason: I have to admit I don't understand this patch; the new comment and link don't seem to have anything to do with the condition...

zeertzjq (approved):

PR #28696, opened 2024-05-11
docs(lsp): improve API documentation for inlay hint

ofseed:
The current documentation is slightly inaccurate or missing.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1589,7 +1589,7 @@ save({lenses}, {bufnr}, {client_id})                 *vim.lsp.codelens.save()*
 Lua module: vim.lsp.inlay_hint                                *lsp-inlay_hint*
 
 enable({enable}, {filter})                       *vim.lsp.inlay_hint.enable()*
-    Enables or disables inlay hints for a buffer.
+    Enables or disables inlay hints for the {filter}ed scope.
 
     To "toggle", pass the inverse of `is_enabled()`: >lua
         vim.lsp.inlay_hint.enable(not vim.lsp.inlay_hint.is_enabled())
@@ -1636,6 +1636,7 @@ get({filter})                                       *vim.lsp.inlay_hint.get()*
         • {inlay_hint} (`lsp.InlayHint`)
 
 is_enabled({filter})                         *vim.lsp.inlay_hint.is_enabled()*
+    Query whether inlay hint is enabled in the {filter}ed scope
 
     Note: ~
       • This API is pre-release (unstable).
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -98,7 +98,7 @@ function M.on_inlayhint(err, result, ctx, _)
   api.nvim__redraw({ buf = bufnr, valid = true })
 end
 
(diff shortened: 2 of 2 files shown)


MariaSolOs:
I don't think this makes the documentation clearer since it's not directly specified what "scope" here is referring to.

ofseed:
@MariaSolOs Would this be better?

MariaSolOs:
> it is mainly to fix bugs rather than provide more detailed descriptions.

? I thought that the purpose of this PR was to improve the documentation, but now you're saying it is a bugfix...?

ofseed:
improve while fixing :) these are wrong

https://github.com/neovim/neovim/blob/8c7a8be27437d8f2d640daf1c2598d4c18933e04/runtime/lua/vim/lsp/inlay_hint.lua#L101-L101

https://github.com/neovim/neovim/blob/8c7a8be27437d8f2d640daf1c2598d4c18933e04/runtime/lua/vim/lsp/inlay_hint.lua#L405-L405

runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -402,14 +403,20 @@ end
 --- Buffer number, or 0 for current buffer, or nil for all.
 --- @field bufnr integer?
 
---- Enables or disables inlay hints for a buffer.
+--- Enables or disables inlay hints for the |filter|ed scope.
```
wookayin, line 406: What does `|filter|ed` scope precisely mean? The help entry `|filter|` doesn't seem relevant.

ofseed: Oh yes, it's a misuse exactly. Would the new way of writing this be better?

runtime/doc/lsp.txt
```diff
@@ -1589,12 +1589,16 @@ save({lenses}, {bufnr}, {client_id})                 *vim.lsp.codelens.save()*
 Lua module: vim.lsp.inlay_hint                                *lsp-inlay_hint*
 
 enable({enable}, {filter})                       *vim.lsp.inlay_hint.enable()*
-    Enables or disables inlay hints for a buffer.
+    Enables or disables inlay hints for the {filter}ed scope.
 
-    To "toggle", pass the inverse of `is_enabled()`: >lua
+    To "toggle" globally, pass the inverse of `is_enabled()`: >lua
```
justinmk, line 1594: this sentence isn't about scope, it's about toggling. toggling is always done with is_enabled, regardless of scope.

the example happens to set global scope, but that's irrelevant.
```suggestion
    To "toggle", pass the inverse of `is_enabled()`: >lua
```

runtime/doc/lsp.txt
```diff
@@ -1589,12 +1589,16 @@ save({lenses}, {bufnr}, {client_id})                 *vim.lsp.codelens.save()*
...
         vim.lsp.inlay_hint.enable(not vim.lsp.inlay_hint.is_enabled())
 <
 
+    Or you can disable it individually for the current buffer: >lua
+        vim.lsp.inlay_hint.enable(false, { bufnr = 0 })
```
justinmk, line 1599: this is over-explaining. examples show mechanics of usage. they generally don't need to demonstrate usage of each parameter, unless the parameter is unusual.

dundargoc:

PR #28706, opened 2024-05-11
ci(test): bump ubuntu images to ubuntu-24.04

dundargoc:


dundargoc:
This doesn't actually work. I think the "Rollout Status of Latest Image Release" in https://github.com/actions/runner-images needs to be green first. Closing for the time being.

PR #28710, opened 2024-05-11
fix(lsp): disable watchfiles health check when not supported

MariaSolOs:
Updating the health check when the OS has no support for `didChangeWatchedFiles`.

I tried to clean things up a bit by moving the logic to `_watchfiles`, but lmk if you don't like this.

### DIFF
--- a/runtime/lua/vim/lsp/health.lua
+++ b/runtime/lua/vim/lsp/health.lua
@@ -51,6 +51,26 @@ end
 
 local function check_watcher()
   vim.health.start('vim.lsp: File watcher')
+
+  -- Only run the check if file watching has been enabled by a client.
+  local clients = vim.lsp.get_clients()
+  if
+    --- @param client vim.lsp.Client
+    vim.iter(clients):all(function(client)
+      local has_capability = vim.tbl_get(
+        client.capabilities,
+        'workspace',
+        'didChangeWatchedFiles',
+        'dynamicRegistration'
+      )
+      local has_dynamic_capability = client.dynamic_capabilities:get(vim.lsp.protocol.Methods.workspace_didChangeWatchedFiles)
+      return has_capability == nil or has_dynamic_capability == nil or client.workspace_folders == nil
+    end)
+  then
+    report_info('file watching "(workspace/didChangeWatchedFiles)" disabled on all clients')
+    return
+  end
+
   local watchfunc = vim.lsp._watchfiles._watchfunc
   assert(watchfunc)
   local watchfunc_name --- @type string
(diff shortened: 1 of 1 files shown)


runtime/lua/vim/lsp/health.lua
```diff
@@ -66,7 +66,7 @@ local function check_watcher()
   end
 
   report_info('File watch backend: ' .. watchfunc_name)
-  if watchfunc_name == 'libuv-watchdirs' then
+  if vim.lsp._watchfiles._is_os_supported() and watchfunc_name == 'libuv-watchdirs' then
```
mfussenegger, line 69: watchdirs isn't used on windows or Mac, see https://github.com/neovim/neovim/blob/080cd39d39539d53d5450ab585e5d9324c660157/runtime/lua/vim/lsp/_watchfiles.lua#L10

And users can still opt in to file watching. If we want to hide it if it's disabled we'd have to check active clients and see if it's disabled on all

MariaSolOs: Lmk if this is what you meant.

MariaSolOs: Wait I think I got that wrong since `capabilities` refers to Neovim's capabilities and not the server's. However I cannot find the respective server capability for `didChangeWatchedFiles`, so how exactly can we check if it's enabled?

runtime/lua/vim/lsp/health.lua
```diff
@@ -51,6 +51,24 @@ end
...
+      local has_capability = vim.tbl_get(
+        client.capabilities,
+        'workspace',
+        'didChangeWatchedFiles',
+        'dynamicRegistration'
+      )
+      return has_capability == nil or client.workspace_folders == nil
+    end)
+  then
+    report_info('No clients support file watching')
```
justinmk, line 68: (not a blocker) should we tend to mention the literal LSP name?
```suggestion
    report_info('No clients support file watching (workspace/didChangeWatchedFiles)')
```

runtime/lua/vim/lsp/health.lua
mfussenegger: Using the client capabilities is fine here. Users can change them.

E.g. I have:

```lua
  local capabilities = vim.tbl_deep_extend(
    "force",
    lsp.protocol.make_client_capabilities(),
    lsp_compl.capabilities(),
    {
      workspace = {
        didChangeWatchedFiles = {
          dynamicRegistration = true
        }
      },
    }
  )
```

and provide them to `vim.lsp.start`.
If we also wanted to check if a server makes use of the capability, we'd have to check `client.dynamic_capabilities:get("workspace/didChangeWatchedFiles")`

runtime/lua/vim/lsp/health.lua
mfussenegger: Or `file watching (workspace/didChangeWatchedFiles) disabled on all clients`

(To imply that users can opt-in, and enable it)

gpanders (approved):

PR #28712, opened 2024-05-11
vim-patch:86f6e2c2eed7

clason:
ftplugin(python): E16 fix, async keyword support for define (vim/vim#14751)

This change includes the following changes:
- Fix "E16: Invalid range" when using a count with jump to start/end of class/method
- Update define with optional async keyword
- Update maintainer email

https://github.com/vim/vim/commit/86f6e2c2eed7df2bf5c60cc74d08d7a8d3c75f45

Co-authored-by: Tom Picton <tom@tompicton.com>


clason:
Not sure why this change should make editorconfig tests fail? @gpanders

zeertzjq:
This runtime file is broken.

clason:
That would explain things :)

zeertzjq (approved):

PR #28713, opened 2024-05-12
fix(tui): reset clear_region attributes during startup

luukvbaal:
Problem:  Fix added in #28676 worked accidentally(used variables were
          themselves uninitialized at this point during startup) and
          does not always work.
Solution: Reset attributes when clearing regions during startup.

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -115,6 +115,7 @@ struct TUIData {
   kvec_t(HlAttrs) attrs;
   int print_attr_id;
   bool default_attr;
+  bool set_default_colors;
   bool can_clear_attr;
   ModeShape showing_mode;
   Integer verbose;
@@ -166,14 +167,6 @@ void tui_start(TUIData **tui_p, int *width, int *height, char **term, bool *rgb)
   tui->seen_error_exit = 0;
   tui->loop = &main_loop;
   tui->url = -1;
-  // Because setting the default colors is delayed until after startup to avoid
-  // flickering with the default colorscheme background, any flush that happens
-  // during startup in turn would result in clearing invalidated regions with
-  // uninitialized attrs(black). Instead initialize clear_attrs with current
-  // terminal background so that it is at least not perceived as flickering, even
-  // though it may be different from the colorscheme that is set during startup.
-  tui->clear_attrs.rgb_bg_color = normal_bg;
-  tui->clear_attrs.cterm_bg_color = (int16_t)cterm_normal_bg_color;
 
   kv_init(tui->invalid_regions);
   kv_init(tui->urlbuf);
@@ -1020,7 +1013,16 @@ static void clear_region(TUIData *tui, int top, int bot, int left, int right, in
 {
   UGrid *grid = &tui->grid;
 
-  update_attrs(tui, attr_id);
+  // Setting the default colors is delayed until after startup to avoid flickering
+  // with the default colorscheme background. Consequently, any flush that happens
+  // during startup would result in clearing invalidated regions with zeroed
+  // clear_attrs, perceived as a black flicker. Reset attributes to clear with
+  // current terminal background instead(#28667, #28668).
+  if (tui->set_default_colors) {
+    update_attrs(tui, attr_id);
+  } else {
+    unibi_out(tui, unibi_exit_attribute_mode);
+  }
 
   // Background is set to the default color and the right edge matches the
   // screen end, try to use terminal codes for clearing the requested area.
@@ -1423,6 +1425,7 @@ void tui_default_colors_set(TUIData *tui, Integer rgb_fg, Integer rgb_bg, Intege
   tui->clear_attrs.cterm_bg_color = (int16_t)cterm_bg;
 
   tui->print_attr_id = -1;
+  tui->set_default_colors = true;
   invalidate(tui, 0, tui->grid.height, 0, tui->grid.width);
 }
 
--- a/test/functional/terminal/tui_spec.lua
+++ b/test/functional/terminal/tui_spec.lua
@@ -2003,38 +2003,39 @@ describe('TUI', function()
     ]])
   end)
 
-  it('invalidated regions are cleared with terminal background attr', function()
-    local screen = Screen.new(50, 10)
-    screen:set_default_attr_ids({ [1] = { foreground = Screen.colors.Black } })
-    screen:attach()
-    fn.termopen({
-      nvim_prog,
-      '--clean',
-      '--cmd',
-      'set termguicolors',
-      '--cmd',
-      'sleep 10',
-    }, {
-      env = {
-        VIMRUNTIME = os.getenv('VIMRUNTIME'),
-      },
-    })
-    screen:expect({
-      grid = [[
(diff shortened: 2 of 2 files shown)


test/functional/terminal/tui_spec.lua
```diff
@@ -2003,38 +2003,49 @@ describe('TUI', function()
...
+        '--cmd',
+        'sleep 2',
+      }, {
+        env = {
+          VIMRUNTIME = os.getenv('VIMRUNTIME'),
+        },
+      })
+      screen:expect({
+        grid = i == 1 and [[
+          ^m                                                 |
```
luukvbaal, line 2028: Looked at this one more time because I still felt uneasy with some inexplicable debugging I did before. #28676 only worked accidentally for 'termguicolors' because it `normal_bg` is initialized to -1, whereas its cterm counterpart is initialized to 0 and still produced a black screen for the added test here. Again not sure about the validity/usefulness of this test as IDK what ^m means, but it does fail on master where that m dissapears and there is no further difference in the test output.

src/nvim/tui/tui.c
```diff
@@ -166,14 +166,13 @@ void tui_start(TUIData **tui_p, int *width, int *height, char **term, bool *rgb)
...
-  // terminal background so that it is at least not perceived as flickering, even
-  // though it may be different from the colorscheme that is set during startup.
-  tui->clear_attrs.rgb_bg_color = normal_bg;
-  tui->clear_attrs.cterm_bg_color = (int16_t)cterm_normal_bg_color;
+  // Setting the default colors is delayed until after startup to avoid flickering
+  // with the default colorscheme background. Consequently, any flush that happens
+  // during startup would result in clearing invalidated regions with zeroed
+  // clear_attrs, perceived as a black flicker. Initialize clear_attrs with -1
+  // instead to make clear_region()->update_attrs() use the current terminal background.
+  tui->clear_attrs.cterm_bg_color = -1;
```
gpanders, line 174: Should this be `0` instead? In `update_attrs` we have:

```c
    bg = (attrs.cterm_bg_color
          ? attrs.cterm_bg_color - 1 : (tui->clear_attrs.cterm_bg_color - 1));
    if (bg != -1) {
      UNIBI_SET_NUM_VAR(tui->params[0], bg);
      unibi_out(tui, unibi_set_a_background);
    }
```

But if `clear_attrs.cterm_bg_color` is `-1,` then `bg` ends up being `-2`, the `if` condition fails, and we still write a sequence to the terminal. It's a nonsense sequence, but it explains the `m` character (if I change `cterm_bg_color` to `0`, the test you wrote fails because the `m` is missing).

luukvbaal: Yeah this seemed suspect to me as well, and although what you say seems to make sense, my changes were result driven. If I set it to `0` I can still reproduce a black screen with `set notermguicolors` during startup. But I guess the nonsense sequence and `^m` result means we should for a different solution still...

gpanders: >If I set it to 0 I can still reproduce a black screen with set notermguicolors during startup.

Interesting, I cannot, and have never been able to, even before #28676. What are you doing to reproduce it? I am using

```vim
set notgc
colorscheme blue
au VimEnter * sleep 1
```

with `nvim --clean -u init.vim` and see no black screen (before and after #28676). If I remove `set notgc` then I see the black screen before #28676 but not after.

luukvbaal: Hmm just realized it may not be possible to reproduce on master. I first noticed it on my ext_ui branch and debugged it there. You can reproduce it by applying this diff:
```diff
diff --git a/runtime/lua/vim/_defaults.lua b/runtime/lua/vim/_defaults.lua
index 4684902410..fd47efbbe8 100644
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -1,3 +1,4 @@
+vim.ui_attach(vim.api.nvim_create_namespace(''), { ext_cmdline = true }, function() end)
 --- Default user commands
 do
   vim.api.nvim_create_user_command('Inspect', function(cmd)
```

and running

    build/bin/nvim --clean +"se notermguicolors" +"sleep 2"

The timing of `nlua_init_defaults()`(which is earlier and where I run `vim.ui_attach` currently on that branch) vs startup cmd/scripts parsing seems significant here.

Not sure what that means, just noticed that `-1` gets rid of the black screen also in that case, but as it turns out for the wrong reasons. Still would like to look for a way that also prevents black screen in that case if possible.

gpanders: I still cannot reproduce the black screen with your patch using your command. The screen maintains my terminal's background color the whole time.

luukvbaal: Hmm no looking at `main.c` the fact that this is executed in `_defaults.lua` seems irrelevant. Should be able to reproduce this with `--cmd`/`+` as well.

luukvbaal: Still unsure about what was going on here but I changed the fix yet again to one that works in all cases and I hope makes sense.

test/functional/terminal/tui_spec.lua
```diff
@@ -2003,38 +2003,43 @@ describe('TUI', function()
...
+        env = {
+          VIMRUNTIME = os.getenv('VIMRUNTIME'),
+        },
+      })
+      screen:expect({
+        grid = [[
+          ^                                                  |
+                                                            |*9
+        ]],
+        intermediate = true
```
zeertzjq, line 2032:
```suggestion
        intermediate = true,
```

luukvbaal: (Not sure why this is needed here btw)

gpanders:
@luukvbaal IIUC the current behavior on master is fine as-is and this PR fixes an issue when a redraw is performed in `_defaults.lua`? So kicking this to the 0.11 milestone should be ok?

luukvbaal:
Flush, not redraw. But yeah for reference: I noticed the test repro still producing a black screen on my ext_ui branch, which calls `vim.ui_attach()->ui_flush()` at the top of `_defaults.lua`. Adding any flush before the tty shenannigans in `_defaults.lua` may still produce a black screen(don't ask me why).

That isn't to say that I know for sure that that is the only way to still produce a black screen during start up. So in that sense this fix seems safer but if no one dares press merge before 0.10, after works too.

gpanders:
>(don't ask me why)

This is really the root of my hesitation to merge this pre 0.10. If it doesn't fix any user facing bugs (that we know of) in master then it should be safe to slow down and find a root cause. I'd like to understand this too, but I don't have the bandwidth to chase it down in the next day or two, so my preference is to kick the can to the 0.11 release cycle.

luukvbaal:
Yeah fair. I kind of lost interest in debugging the behavior on master further once I found this alternative solution and established that #28676 made incorrect assumptions.

Figuring out why trying to force `clear_region()->update_attrs()` to use the default background by initializing to 0/-1(which it turned #28676 actually amounted to), may still be interesting but I'm not sure how useful that actually still is give it's accidental nature. The updated test output(actual blank screen vs some attr) also seems more reliable.

PR #28720, opened 2024-05-12
fix(lsp): vim.lsp.buf_request_sync type annotation

crwebb85:
Problem:

The vim.lsp.buf_request_sync return type did not match the actual response.

Solution:

Changed the `err` field annotion to `error`.

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -938,7 +938,7 @@ end
 ---@param params table? Parameters to send to the server
 ---@param timeout_ms integer? Maximum time in milliseconds to wait for a result.
 ---                           (default: `1000`)
----@return table<integer, {err: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
+---@return table<integer, {error: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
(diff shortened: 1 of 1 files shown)


dundargoc:

runtime/lua/vim/lsp.lua
```diff
@@ -938,7 +938,7 @@ end
 ---@param params table? Parameters to send to the server
 ---@param timeout_ms integer? Maximum time in milliseconds to wait for a result.
 ---                           (default: `1000`)
----@return table<integer, {err: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
+---@return table<integer, {error: lsp.ResponseError, result: any}>? result Map of client_id:request_result.
```
wookayin, line 941: Technically speaking, `{ error: lsp.ResponseError?, ... }` is a correct typing here because when there is no error, the `error` field won't be set (i.e. `nil`).

crwebb85: Yeah, I thought about fixing that as well but there are other functions with the same type signature that need to be changed to optional, and I would rather handle those in a PR specific to that. I will probably see about fixing those later.

wookayin: It is already fixed on master.

PR #28721, opened 2024-05-12
fix(lsp): update the changetracking buffer state on reload

### LINKED ISSUE #24972: error attempt to get length of local 'prev_line'

icholy:
Fixes #24972

When a buffer is reloaded, each client is detached from that buffer and then re-attached:

https://github.com/neovim/neovim/blob/4e5c633ed4871a948aff7338b793ac5f93484153/runtime/lua/vim/lsp.lua#L557-L561

When there’s only 1 LSP, the change tracking buffer state is deleted and re-created when the file is externally modified. This updates the `buf_state.lines` with the new buffer contents.

https://github.com/neovim/neovim/blob/4e5c633ed4871a948aff7338b793ac5f93484153/runtime/lua/vim/lsp/_changetracking.lua#L205

If there are 2 LSPs attached, the buffer state is never deleted because the `refs` count never hits zero. So the buf state is never updated with the new buffer contents.

https://github.com/neovim/neovim/blob/4e5c633ed4871a948aff7338b793ac5f93484153/runtime/lua/vim/lsp/_changetracking.lua#L202-L204

### DIFF
--- a/runtime/lua/vim/lsp/_changetracking.lua
+++ b/runtime/lua/vim/lsp/_changetracking.lua
@@ -38,6 +38,7 @@ local M = {}
 --- @field refs integer how many clients are using this group
 ---
 --- @class vim.lsp.CTGroupState
+--- @field group vim.lsp.CTGroup
 --- @field buffers table<integer,vim.lsp.CTBufferState>
 --- @field debounce integer debounce duration in ms
 --- @field clients table<integer, table> clients using this state. {client_id, client}
@@ -138,6 +139,7 @@ function M.init(client, bufnr)
     state.clients[client.id] = client
   else
     state = {
+      group = group,
       buffers = {},
       debounce = client.flags.debounce_text_changes or 150,
       clients = {
@@ -219,6 +221,19 @@ function M.reset(client)
       reset_timer(buf_state)
     end
     state.buffers = {}
+  end
+end
+
+--- @param bufnr integer
+function M.reload(bufnr)
+  for _, state in pairs(state_by_group) do
+    assert(state.buffers, 'CTGroupState must have buffers')
+    if state.group.sync_kind == protocol.TextDocumentSyncKind.Incremental then
+      local buf_state = state.buffers[bufnr]
+      if buf_state then
+        buf_state.lines = api.nvim_buf_get_lines(bufnr, 0, -1, true)
+      end
+    end
   end
 end
 
--- a/runtime/lua/vim/lsp/_changetracking.lua
+++ b/runtime/lua/vim/lsp/_changetracking.lua
@@ -231,6 +231,8 @@ function M.reload(bufnr)
     if state.group.sync_kind == protocol.TextDocumentSyncKind.Incremental then
       local buf_state = state.buffers[bufnr]
       if buf_state then
+        buf_state.needs_flush = false
+        buf_state.pending_changes = {}
         buf_state.lines = api.nvim_buf_get_lines(bufnr, 0, -1, true)
       end
     end
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -554,6 +554,7 @@ local function buf_attach(bufnr)
     end,
 
     on_reload = function()
+      changetracking.reload(bufnr)
       local params = { textDocument = { uri = uri } }
       for _, client in ipairs(lsp.get_clients({ bufnr = bufnr })) do
         changetracking.reset_buf(client, bufnr)


icholy:
Alternatively, we could do something like this:

``` lua
--- @param bufnr integer
function M.dirty_buf(bufnr)
  for _, state in pairs(state_by_group) do
    local buf_state = state.buffers[bufnr]
    if buf_state then
      buf_state.needs_reload = true
    end
  end
end
```

icholy:
I'm not sure how to test this.

icholy:
I'm testing some more scenarios and the lazy reloading with the `needs_reload` flag is still failing. Switching to eager reloading for now.

icholy:
I'm not attached to this implementation. If someone wants to suggest an alternate implementation or make a spin-off PR I'd be more than happy to test it out.

icholy:
@mfussenegger I was looking through the history and it seems like you've done the most work in this area of the code. Any feedback would be appreciated.

icholy:
Another way to fix this would be do change the `on_reload` callback to detach all the clients and then re-attach them all in a second pass.

``` lua
on_reload = function()
  local params = { textDocument = { uri = uri } }
  local clients = lsp.get_clients({ bufnr = bufnr })
  for _, client in ipairs(clients) do
    changetracking.reset_buf(client, bufnr)
    if vim.tbl_get(client.server_capabilities, 'textDocumentSync', 'openClose') then
      client.notify(ms.textDocument_didClose, params)
    end
  end
  for _, client in ipairs(clients) do
    client:_text_document_did_open_handler(bufnr)
  end
end,
```

runtime/lua/vim/lsp/_changetracking.lua
```diff
@@ -222,6 +224,19 @@ function M.reset(client)
...
 end
 
+--- @param bufnr integer
+function M.reload(bufnr)
+  for _, state in pairs(state_by_group) do
+    assert(state.buffers, 'CTGroupState must have buffers')
+    if state.group.sync_kind == protocol.TextDocumentSyncKind.Incremental then
+      local buf_state = state.buffers[bufnr]
+      if buf_state then
+        buf_state.lines = api.nvim_buf_get_lines(bufnr, 0, -1, true)
```
icholy, line 234: It might make sense to also clear any pending changes too.

icholy:
Closing in favor of https://github.com/neovim/neovim/pull/28875

PR #28723, opened 2024-05-12
perf(lsp): only joinpath for dirs in watchdirs

mfussenegger:
Doesn't have a huge impact, but showed up in profile output using
`require("jit.p").start("i1", "/tmp/profile")`

before:

    31%  joinpath
    25%  fs.lua:0
    13%  normalize
    13%  skip
     8%  _watchfunc
     5%  gsplit
     3%  spairs

after:

    34%  skip
    29%  fs.lua:0
    12%  joinpath
     7%  normalize
     5%  _watchfunc
     5%  spairs

### DIFF
--- a/runtime/lua/vim/_watch.lua
+++ b/runtime/lua/vim/_watch.lua
@@ -200,11 +200,13 @@ function M.watchdirs(path, opts, callback)
   local max_depth = 100
 
   for name, type in vim.fs.dir(path, { depth = max_depth }) do
-    local filepath = vim.fs.joinpath(path, name)
-    if type == 'directory' and not skip(filepath, opts) then
-      local handle = assert(uv.new_fs_event())
-      handles[filepath] = handle
-      handle:start(filepath, {}, create_on_change(filepath))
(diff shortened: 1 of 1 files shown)


justinmk:
What do the absolute times look like? It's possible that this change merely shuffles the proportions around. E.g. now `skip` dominates, but just like `joinpath`, that is presumably trivial.

wookayin:
FYI, the `r` option (in addition to `i1` will show sample counts instead of percentage (so it's like milliseconds): e.g. `jit.p.start("ri1m0", ...)`

mfussenegger:
> What do the absolute times look like? It's possible that this change merely shuffles the proportions around. E.g. now `skip` dominates, but (like `joinpath`) that is presumably trivial, and doesn't necessarily mean that much work was saved.

The absolute numbers will vary a lot depending on the project size, and ratio of files vs. directory. The project I used has 25826 files of which 4367 are directories:

Before:

94.894235ms
91.499979ms

After:

62.750427ms
67.906174ms

PR #28724, opened 2024-05-12
feat(lsp): add the end location information of lsp to quickfix item

ofseed:
There's the support of `end_lnum` and `end_col` for quickfix already, which was originally designed to support LSP locations, see https://github.com/vim/vim/pull/8393, and it has already been patched to nvim.

After making these distinctions, this code looks ugly. I'm not sure if I need to refactor

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -41,7 +41,10 @@ EVENTS
 
 LSP
 
-• TODO
+• |vim.lsp.buf.declaration()|, |vim.lsp.buf.definition()|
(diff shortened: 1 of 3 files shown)


ofseed:
Sample usage of nvim-bqf, which utilizes the `end_lnum` and `end_col`

Before this PR:
image

After:

image

ofseed:
The information will also be shown in the default quickfix window
image

rockyzhang24:
I think the implementation https://github.com/neovim/neovim/pull/29164 is much better and more consistent.

PR #28727, opened 2024-05-13
fix(treesitter): indent size for inspect_tree

wookayin:
Problem: For :InspectTree, indent size (`&shiftwidth`) for the tree
viewer may be incorrect.

This is because the tree viewer buffer with the filetype `query` does
not explicitly configures the tab size, which can mismatch with the
default indent size (2) assumed by TSTreeView's implementation.

Solution: Set shiftwidth to be the same as TSTreeViewOpts specifies,
which defaults to 2.

### DIFF
--- a/runtime/lua/vim/treesitter/dev.lua
+++ b/runtime/lua/vim/treesitter/dev.lua
@@ -158,7 +158,8 @@ end
 
 ---@param w integer
 ---@param b integer
-local function set_dev_options(w, b)
+---@param opts nil|{ indent?: integer }
+local function set_dev_options(w, b, opts)
   vim.wo[w].scrolloff = 5
   vim.wo[w].wrap = false
   vim.wo[w].foldmethod = 'expr'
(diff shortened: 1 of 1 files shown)


wookayin:
Before (with some indent-line plugin) when user's global `&shiftwidth = 4`:

Before

After: `vim.bo.shiftwidth` is now 2.

After


Remark: we could instead set the default indent size for `&ft = query` (`runtime/query.lua`), but I chose to configure buffer-local settings because `treeview.opts` *may* have a different indent size other than 2 in the current implementation.

zeertzjq:
I don't think 'tabstop' should be changed, as there are no Tabs in the text. But maybe 'expandtabs' needs to be set?

wookayin:
Forgot to mention that tabstop, etc. are for the query editor rather than InspectTree. For InspectTree right, we don't have tabs --- but the default options for query editors will be noexpandtab and tabsize=8. I thus have added the `expandtab` option in `set_dev_properties`.

clason:
Let's leave the query editor for now; the plan is to include the formatter from nvim-treesitter for query files (including the editor).

justinmk:
I don't think this needs to block 0.10

clason:
@wookayin During feature freeze, we (only) use the milestone for actual blockers in order to manage the release process. Obviously this can and will be backported after the 0.10 release (and we can re-use the milestone for 0.10.x then).

justinmk:
@wookayin would you mind rebasing this

PR #28728, opened 2024-05-13
feat: set properties on namespaces

altermo:
Use a singular function that can specify properties on the namespace.

Removed `nvim__win_xx_ns` functions as they are no longer needed.

### DIFF
--- a/src/nvim/api/extmark.c
+++ b/src/nvim/api/extmark.c
@@ -18,6 +18,7 @@
 #include "nvim/decoration_provider.h"
 #include "nvim/drawscreen.h"
 #include "nvim/extmark.h"
+#include "nvim/globals.h"
 #include "nvim/grid.h"
 #include "nvim/highlight_group.h"
 #include "nvim/map_defs.h"
@@ -41,6 +42,7 @@ void api_extmark_free_all_mem(void)
     xfree(name.data);
   })
   map_destroy(String, &namespace_ids);
+  set_destroy(uint32_t, &namespace_localscope);
 }
 
 /// Creates a new namespace or gets an existing one. [namespace]()
@@ -177,10 +179,6 @@ static Array extmark_to_array(MTPair extmark, bool id, bool add_dict, bool hl_na
     }
     if (mt_invalid(start)) {
       PUT_C(dict, "invalid", BOOLEAN_OBJ(true));
-    }
-
-    if (mt_scoped(start)) {
-      PUT_C(dict, "scoped", BOOLEAN_OBJ(true));
     }
 
     decor_to_dict_legacy(&dict, mt_decor(start), hl_name, arena);
@@ -489,8 +487,6 @@ Array nvim_buf_get_extmarks(Buffer buffer, Integer ns_id, Object start, Object e
 ///                   used together with virt_text.
 ///               - url: A URL to associate with this extmark. In the TUI, the OSC 8 control
 ///                   sequence is used to generate a clickable hyperlink to this URL.
-///               - scoped: boolean (EXPERIMENTAL) enables "scoping" for the extmark. See
-///                 |nvim__win_add_ns()|
 ///
 /// @param[out]  err   Error details, if any
 /// @return Id of the created/updated extmark
@@ -749,11 +745,6 @@ Integer nvim_buf_set_extmark(Buffer buffer, Integer ns_id, Integer line, Integer
   }
 
   if (opts->ephemeral && decor_state.win && decor_state.win->w_buffer == buf) {
-    if (opts->scoped) {
-      api_set_error(err, kErrorTypeException, "not yet implemented");
-      goto error;
-    }
-
     int r = (int)line;
     int c = (int)col;
     if (line2 == -1) {
@@ -834,7 +825,7 @@ Integer nvim_buf_set_extmark(Buffer buffer, Integer ns_id, Integer line, Integer
     extmark_set(buf, (uint32_t)ns_id, &id, (int)line, (colnr_T)col, line2, col2,
                 decor, decor_flags, right_gravity, opts->end_right_gravity,
                 !GET_BOOL_OR_TRUE(opts, set_extmark, undo_restore),
-                opts->invalidate, opts->scoped, err);
+                opts->invalidate, err);
     if (ERROR_SET(err)) {
       decor_free(decor);
       return 0;
@@ -960,7 +951,7 @@ Integer nvim_buf_add_highlight(Buffer buffer, Integer ns_id, String hl_group, In
   decor.data.hl.hl_id = hl_id;
 
   extmark_set(buf, ns, NULL, (int)line, (colnr_T)col_start, end_line, (colnr_T)col_end,
-              decor, MT_FLAG_DECOR_HL, true, false, false, false, false, NULL);
+              decor, MT_FLAG_DECOR_HL, true, false, false, false, NULL);
   return ns_id;
 }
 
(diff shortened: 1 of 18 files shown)


zeertzjq:
It may be better to be able to change a namespace from global scope to window scope in one API call. The API will be like `nvim_ns_set_scope` or some variation.

justinmk:
> The API will be like `nvim_ns_set_scope`

Can we have a generic `nvim_ns_set()` function which sets various properties of the namespace?

src/nvim/plines.c
```diff
@@ -158,7 +159,7 @@ CharSize charsize_regular(CharsizeArg *csarg, char *const cur, colnr_T const vco
         break;
       } else if (mark.pos.col == col) {
         if (!mt_end(mark) && (mark.flags & MT_FLAG_DECOR_VIRT_TEXT_INLINE)
-            && mt_scoped_in_win(mark, wp)) {
+            && ns_scoped_in_win(mark.ns, wp)) {
```
justinmk, line 162:
```suggestion
            && ns_in_win(mark.ns, wp)) {
```

src/nvim/api/extmark.h
```diff
@@ -4,14 +4,25 @@
...
 #include "nvim/decoration_defs.h"  // IWYU pragma: keep
 #include "nvim/macros_defs.h"
 #include "nvim/map_defs.h"
 #include "nvim/types_defs.h"
 
 EXTERN Map(String, int) namespace_ids INIT( = MAP_INIT);
+EXTERN Set(uint32_t) namespace_scoped INIT( = SET_INIT);
 EXTERN handle_T next_namespace_id INIT( = 1);
 
+static inline bool ns_scoped_in_win(uint32_t ns_id, win_T *wp)
```
justinmk, line 17:
```suggestion
/// Returns true if the namespace is scoped in the given window.
static inline bool ns_scoped_in_win(uint32_t ns_id, win_T *wp)
```

src/nvim/api/extmark.c
```diff
@@ -1291,3 +1281,47 @@ Boolean nvim__win_del_ns(Window window, Integer ns_id, Error *err)
...
+
+/// EXPERIMENTAL: this API will change in the future.
+///
+/// Set some properties for namespace
+///
+/// @param ns_id Namespace
+/// @param opts Optional parameters to set:
+///           - scoped: (boolean) Make the namespace a scoped namespace
+///
+void nvim__ns_set(Integer ns_id, Dict(ns_opts) *opts, Error *err)
```
justinmk, line 1293: this should accept `opts.win` (or maybe `opts.wins: number[]` if we really care about avoiding multiple calls) to set the scope.

That allows us to remove the `nvim__win_xx_ns` functions, right?

src/nvim/api/extmark.c
```diff
@@ -1291,3 +1281,47 @@ Boolean nvim__win_del_ns(Window window, Integer ns_id, Error *err)
...
   return true;
 }
+
+/// EXPERIMENTAL: this API will change in the future.
+///
+/// Set some properties for namespace
+///
+/// @param ns_id Namespace
+/// @param opts Optional parameters to set:
+///           - scoped: (boolean) Make the namespace a scoped namespace
```
justinmk, line 1291: I'm still [wondering](https://github.com/neovim/neovim/pull/27361#discussion_r1597710114) why the `opts.scoped` parameter is needed. If a namespace was assigned to windows/buffers/etc, it's scoped. Else it's not scoped.

justinmk:
This looks like the right direction to me. 

Would it be more ideal to [set namespaces as global-local options](https://github.com/neovim/neovim/pull/25455#discussion_r1345851839)? That might still be blocked by difficult work. @bfredl

src/nvim/api/extmark.c
altermo: Then I'll just remote the option

src/nvim/window.c
```diff
@@ -5195,8 +5192,6 @@ void win_free(win_T *wp, tabpage_T *tp)
   // Don't execute autocommands while the window is halfway being deleted.
   block_autocmds();
 
-  set_destroy(uint32_t, &wp->w_ns_set);
```
justinmk, line 5198: Internally, storing the namespaces on the window (`w_ns_set `) might still be the right approach. Does that make cleanup easier (when a window is destroyed)? Otherwise `namespace_scope_map[..]->windows` will have stale ids in it? It also seems a bit more natural.

So the interface is `nvim__ns_set(..., {wins:...})`, but internally the namespaces are stored on the window/buffer/tabpage objects. This is a compromise because we don't have the [ideal(?) interface (global-local options)](https://github.com/neovim/neovim/pull/28728#issuecomment-2117553866).

src/nvim/api/extmark.h
```diff
@@ -4,14 +4,26 @@
...
 EXTERN handle_T next_namespace_id INIT( = 1);
 
+/// Returns true if the namespace is scoped in the given window.
+static inline bool ns_in_win(uint32_t ns_id, win_T *wp)
+{
+  if (!set_has(uint32_t, &namespace_scoped, ns_id)) {
+    return true;
+  }
+
+  return set_has(uint32_t, &wp->w_ns_set, ns_id);
```
justinmk, line 24: This seems strange. Why is `namespace_scoped` necessary? If ns_id is *not* in `namespace_scoped`, this function returns `true`? Why isn't the decision fully decided by `w_ns_set`, why is `namespace_scoped` needed at all?

altermo: `namespace_scoped` is needed because it currently is equivalent to "namespace is scoped in at least one window" and if it doesn't contain the namespace, then that namespace is "global scoped" (which is what you asked for).

Though `namespace_scoped` can sometimes contain namespaces that are not scoped in any windows, because if a window is deleted, and that window was the only window where the namespace was scoped, then `namespace_scoped` isn't updated. (I could change this so that it's always synced, but that would entail that when a window is deleted, a namespace can suddenly become global scope, which is probably not what one wants.)

src/nvim/api/extmark.h
```diff
@@ -4,14 +4,26 @@
...
 #include "nvim/decoration_defs.h"  // IWYU pragma: keep
 #include "nvim/macros_defs.h"
 #include "nvim/map_defs.h"
 #include "nvim/types_defs.h"
 
 EXTERN Map(String, int) namespace_ids INIT( = MAP_INIT);
+EXTERN Set(uint32_t) namespace_scoped INIT( = SET_INIT);
 EXTERN handle_T next_namespace_id INIT( = 1);
 
+/// Returns true if the namespace is scoped in the given window.
```
justinmk, line 17:
```suggestion
/// Returns true if the namespace is global or scoped in the given window.
```

src/nvim/api/extmark.h
justinmk: right, the docstring and names were confusing.



> `namespace_scoped` is needed because it currently is equivalent to "namespace is scoped in at least one window"

Sounds like an unnecessary optimization. We don't need `namespace_scoped`, we can iterate through all windows and check `w_ns_set`. That is less confusing and requires less bookkeeping.

src/nvim/api/extmark.h
```diff
@@ -4,14 +4,26 @@
...
 #include "nvim/api/keysets_defs.h"  // IWYU pragma: keep
 #include "nvim/api/private/defs.h"  // IWYU pragma: keep
+#include "nvim/buffer_defs.h"
 #include "nvim/decoration_defs.h"  // IWYU pragma: keep
 #include "nvim/macros_defs.h"
 #include "nvim/map_defs.h"
 #include "nvim/types_defs.h"
 
 EXTERN Map(String, int) namespace_ids INIT( = MAP_INIT);
+EXTERN Set(uint32_t) namespace_scoped INIT( = SET_INIT);
```
justinmk, line 14:
```suggestion
/// Indicates whether a namespace has non-global scope.
EXTERN Set(uint32_t) namespace_localscope INIT( = SET_INIT);
```

justinmk, line 14: Is it too early to introduce (pseudocode):

    namespaces: Map<id, NamespaceProps>

where `NamespaceProps` is a struct like:

```
{ scope: NsScope; ... }
```

test/functional/lua/highlight_spec.lua
```diff
@@ -126,9 +126,10 @@ describe('vim.highlight.on_yank', function()
       vim.api.nvim_buf_set_mark(0,"]",1,1,{})
       vim.highlight.on_yank({timeout = math.huge, on_macro = true, event = {operator = "y"}})
     ]])
-    neq({}, api.nvim__win_get_ns(0))
+    local ns = api.nvim_create_namespace('hlyank')
+    eq(api.nvim_get_current_win(), api.nvim__ns_get(ns).wins[1])
     command('wincmd w')
-    eq({}, api.nvim__win_get_ns(0))
+    neq(api.nvim_get_current_win(), api.nvim__ns_get(ns).wins[1])
```
justinmk, line 132: can we turn this into an eq() test somehow (same for the other case below). that is a much strong test.

```lua
    eq(win, api.nvim__ns_get(ns).wins[1])
    command('wincmd w')
    eq(win, api.nvim__ns_get(ns).wins[1])
```

src/nvim/api/extmark.h
```diff
@@ -4,14 +4,27 @@
...
 #include "nvim/api/keysets_defs.h"  // IWYU pragma: keep
 #include "nvim/api/private/defs.h"  // IWYU pragma: keep
+#include "nvim/buffer_defs.h"
 #include "nvim/decoration_defs.h"  // IWYU pragma: keep
 #include "nvim/macros_defs.h"
 #include "nvim/map_defs.h"
 #include "nvim/types_defs.h"
 
 EXTERN Map(String, int) namespace_ids INIT( = MAP_INIT);
+/// Non-global namespaces.
```
justinmk, line 14: Let's explicitly document this quirk, until we decide what to do about it:

```suggestion
/// Non-global namespaces. A locally-scoped namespace may be "orphaned" if all
/// window(s) it was scoped to, are destroyed.
```

justinmk (approved):
After the last small comments, this is good enough to merge and can build on it later.

PR #28734, opened 2024-05-13
perf(fs): normalize path only once in fs.dir

mfussenegger:
Re-normalizing a path after a `joinpath` isn't necessary. Calling
`normalize` on each child directory had quite a bit of impact when
traversing a large directory.

A simple test showed:

Before: ~144ms
After: ~80ms

running the following logic against a dir with 4367 child folders and
25826 files:

    local files = {}
    local start = uv.hrtime()
    for name, type in vim.fs.dir(path, { depth = max_depth }) do
      table.insert(files, { name, type })
    end
    local duration = uv.hrtime() - start

Relates to https://github.com/neovim/neovim/issues/23291

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -120,8 +120,9 @@ function M.dir(path, opts)
     skip = { opts.skip, { 'function' }, true },
   })
 
+  path = M.normalize(path)
   if not opts.depth or opts.depth == 1 then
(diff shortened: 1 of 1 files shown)


dundargoc:
Not against this PR per se, but does this indicate a root problem with `vim.fs.normalize()` that we need to fix?

gpanders (approved):

Issue #2434, opened 2015-04-15
ZB: command to delete buffer without closing the window


norcalli:
All other solutions I know of are hacks that kinda work (usually by creating a new split and then deleting the buffer in that window, and attempting to cover all possible cases for this), but intrinsic support for this would be greatly appreciated.

justinmk:
> usually by creating a new split and then deleting the buffer in that window

That isn't necessary at all. The hack I use switches to the next valid buffer (or creates a new empty one), then deletes the original.

norcalli:
@justinmk I think you'll find [it's](http://vim.wikia.com/wiki/Deleting_a_buffer_without_closing_the_window) [been](https://github.com/moll/vim-bbye) [done](https://github.com/rbgrouleff/bclose.vim) [already](http://stackoverflow.com/questions/4465095/vim-delete-buffer-without-losing-the-split-window) [in many different ways](http://stackoverflow.com/questions/4298910/vim-close-buffer-but-not-split-window).

The fact that you call it a hack would seem to support this even more. It doesn't seem like it should be difficult to implement.

justinmk:
Yes, I know. But the idea of adding a new core command for this doesn't sound appealing. Details would be welcome though.

splinterofchaos:
`:enew!`?

> Edit a new, unnamed buffer.  Discard any changes to the current buffer.

norcalli:
A whole terminal was added along with autocommands, a command, and a Terminal class. This is incredibly light weight in comparison. And surprisingly, `bclose` is not taken AFAIK, so that would be a valid name, and upon using it, it would destroy the buffer and move to the next available buffer while maintaining the split window or moving to a scratch buffer if it is the last buffer. It is meant to preserve window layout while still closing a buffer.

@splinterofchaos `enew!` doesn't close the buffer with `set hidden`

tarruda:
Why not simply implement this as a buffer option?

justinmk:
> A whole terminal was added along with autocommands, a command, and a Terminal class. This is incredibly light weight in comparison.

I'm talking about the UI, not the technical cost, and in *proportion* to the benefit. The technical and UI cost of `:terminal` in relation to its benefit is worth it. Adding `:bdelete_but_leave_my_window_alone` may not be.

> bclose is not taken AFAIK,

So what about someone who wants to `:bwipe` or `:bunload` without closing the window? 

> Why not simply implement this as a buffer option?

[...]

PR #28742, opened 2024-05-14
store w_last_used timestamp on each window

### LINKED ISSUE #28745: Adding lastused property to windows
### Problem

I'm trying to add a `lastused` property to `vim.fn.getwininfo`, like in `vim.fn.getbufinfo`.

It's just a timestamp for when a windown was last entered into

I've made a start: https://github.com/neovim/neovim/pull/28742 and for the most part it works, however, there's one case where it doesn't work as expected;

If a buffer is opened in multiple windows, the timestamp (`w_last_used`) won't get updated when you enter any of the other windows, apart from the first window that the buffer was opened in and all the windows display the same timestamp as the first window.

**Please note** I'm completly new to this codebase and any help to pointing me to the right direction, or just any tips/guidance in general would be greatly appreciated

## Background
In case anyone is curious as to why I want this;

It's to do with the qf window, specifically what happens when you select an item in the qf window that's open in multiple windows.

When I'm developing, I may have multiple windows open of the same buffer (due to me jumping around the code).

If there's an item in the qf window that's open in multiple windows, I would like that item to open in the windown that was most reccently accessed, see; https://github.com/kevinhwang91/nvim-bqf/pull/140/files

### Expected behavior

I expect each window to have it's timestamp updated each time I enter it

kkrime:
fixes https://github.com/neovim/neovim/issues/28745

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -2363,9 +2363,7 @@ static void patch_terminfo_bugs(TUIData *tui, const char *term, const char *colo
   // DECSCUSR (cursor shape) is widely supported.
   // https://github.com/gnachman/iTerm2/pull/92
   if (!bsdvt
-      && (xterm         // anything claiming xterm compat
-          // per MinTTY 0.4.3-1 release notes from 2009
-          || putty
+      && (putty         // per MinTTY 0.4.3-1 release notes from 2009
           // per https://chromium.googlesource.com/apps/libapps/+/a5fb83c190aa9d74f4a9bca233dac6be2664e9e9/hterm/doc/ControlSequences.md
           || hterm
           // per https://bugzilla.gnome.org/show_bug.cgi?id=720821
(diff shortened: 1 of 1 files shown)


clason:
What problem does this solve?

bfredl:
This can be done with a WinEnter autocmd.

justinmk:
This could be interesting, because it could be an alternative to providing a "history" of last-used objects. I think we clearly want "history" for last-used, so this `last_used: datetime` could be a way to avoid an extra ringbuffer for each object type.

That is, instead of:

```
histget('win')
histget('buf')
histget('tabpage')
```

users/plugins could get the "history" of windows/buffers/tabpages by sorting by `last_used`. And even if we provide `histget('win')`, that could internally use the same approach.

Having a history of last-used bufs/wins/tabs would help improve the behavior of `g<tab>`, [ZB](https://github.com/neovim/neovim/issues/2434) (delete buffer without losing split), etc.

Caveat: multple windows could have the same `last_used` timestamp, then the "history" is not 100% reliable. But storing microsecond-granularity should be good enough.

kkrime:
Hi All,

Thank you for looking into this PR, if you want more context for this chance, please take a look at https://github.com/neovim/neovim/issues/28745

PR #28743, opened 2024-05-14
feat(fs): follow symlinks in find()

sarmong:
Problem: 
`find` doesn't look for files in symlinked directories.

Solution:
add `opts.follow_symlinks` to `find()` method to treat links that point to directories as directories.


sarmong:
Need assistance to write a test. Current tests for `.find()` test against `build` directories and there are no symlinks there.

dundargoc:
Fixed in https://github.com/neovim/neovim/pull/31551

PR #28748, opened 2024-05-14
feat(lua): add buflocal

mfussenegger:
This is currently more of a proof of concept to get some feedback on whether
this is a horrible idea, or somewhat reasonable.

I was mostly interested if this is possible at all and Justin mentioned using
`b:` elsewhere, which afaik has some vimscript marshalling limitations for some
types(?) and you cannot keep state completely hidden as implementation detail.

I'm also not yet sure if this can cover most cases. If there's positive feedback I'd try and convert some more spots, otherwise 🚮

---

There are a few modules, each of which is managing state per buffer and
duplicating logic like cleanup.
Properly cleaning up can be tricky - see:

https://github.com/neovim/neovim/commit/f32fd19f1eedbd75e6a37b73f28cf8761e0e875c

This introduces a `Buflocal` inspired by thread-local storage:

https://en.wikipedia.org/wiki/Thread-local_storage

The `Buflocal` can be used to access state for a buffer, which gets
cleaned up if the buffer is deleted.

(Although the name here may be a bit misleading - as the buffer number is parameterized and you can get the state of other buffers, not just current. )

### DIFF
--- a/runtime/lua/vim/lsp/_changetracking.lua
+++ b/runtime/lua/vim/lsp/_changetracking.lua
@@ -38,7 +38,7 @@ local M = {}
 --- @field refs integer how many clients are using this group
 ---
 --- @class vim.lsp.CTGroupState
---- @field buffers table<integer,vim.lsp.CTBufferState>
+--- @field buffers vim.Buflocal
 --- @field debounce integer debounce duration in ms
 --- @field clients table<integer, table> clients using this state. {client_id, client}
 
@@ -138,7 +138,20 @@ function M.init(client, bufnr)
     state.clients[client.id] = client
   else
     state = {
-      buffers = {},
+      buffers = vim.buflocal(function(b)
+        local bufstate = {
+          name = api.nvim_buf_get_name(b),
+          lines = {},
+          lines_tmp = {},
+          pending_changes = {},
+          needs_flush = false,
+          refs = 0,
+        }
+        if group.sync_kind == protocol.TextDocumentSyncKind.Incremental then
+          bufstate.lines = api.nvim_buf_get_lines(b, 0, -1, true)
+        end
+        return bufstate
+      end),
       debounce = client.flags.debounce_text_changes or 150,
       clients = {
         [client.id] = client,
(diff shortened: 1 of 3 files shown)


justinmk:
> Justin mentioned using `b:` elsewhere, which afaik has some vimscript marshalling limitations for some types(?) and you cannot keep state completely hidden as implementation detail.

See https://github.com/neovim/neovim/issues/12544#issuecomment-1116794687 for my proposal. If possible, I think it's ideal that we keep `vim.b`/`vim.w`/etc as the interface for "state that gets destroy when the object gets destroyed". I don't see marshalling limitations as a problem, because that only matters if Vimscript requests the state, and it can simply get a degraded form of the state.

mfussenegger:
> I don't see marshalling limitations as a problem

👍


> If possible, I think it's ideal that we keep vim.b/vim.w/etc as the interface for "state that is bound to the lifecycle of the object".

That would also imply that the state is always public, no? I imagine `diagnostic_cache`  in `vim.diagnostic` would then be exposed as `vim.b.vim_diagnostic_cache` or something?

---


On second thought, the `buflocal` might actually be closer to a `vim.defaulttable()`, so it could also be some kind of `vim.buftable()`, that behaves more like a table with `tbl[bufnr]` index access.

ii14:
`on_detach` runs on just rereading the file with `:e` (btw shouldn't `nvim_buf_attach` run `on_reload` in this case, or am I misremembering how it works?), so variables can be removed too soon.

Events on the other hand can be silenced (explicitly with `:noau`, `eventignore` or most often just because they were emitted inside some other autocmd and someone forgot to mark it `nested`). And it does happen in practice - I similarly maintain my own buffer list and it occasionally goes out of sync precisely because of nested autocmds being silenced by default somewhere.

As far as I know, there is no mechanism that lets you do this robustly right now?

altermo:
What about setting a buffer-local(`b:`) unique object and then doing cleanup when that object gets garbage collected because the buffer no longer exists?
Here is an example script:
```lua
local values=setmetatable({},{__mode='k'})
function Buf_set_var(id,key,value)
    if not vim.b[id].nvim_buflocal_id then
        local new_id_object=function () end
        vim.b[id].nvim_buflocal_id=new_id_object
        values[new_id_object]={}
    end
    local id_object=vim.b[id].nvim_buflocal_id
    values[id_object][key]=value
end
function Buf_get_var(id,key)
    if not vim.b[id].nvim_buflocal_id then
        return nil
    end
    local id_object=vim.b[id].nvim_buflocal_id
    return values[id_object][key]
end
```

justinmk:
> What about setting a buffer-local(`b:`) unique object and then doing cleanup when that object gets garbage collected because the buffer no longer exists?

Isn't that covered by [my suggestion](https://github.com/neovim/neovim/issues/12544#issuecomment-1116794687), which would be automatic and implicit for any use of `vim.b` in Lua.

altermo:
Yes

PR #28755, opened 2024-05-15
docs(gen_help_html.lua): fix broken pre text, and handle linewrap

### LINKED ISSUE #28754: Online documentation contains strange wrapping
### Problem

Online documentation  is wrapped strangely.  For example on this page: 

https://neovim.io/doc/user/

you see text jumbled together in a paragraph.  For instance, each section should be formatted like a table.  In the 'Basic Editing' section 'starting' should be in the first column and its corresponding description is in the second column.  In the next line you should see 'edit-files' and its corresponding description. Instead all the subsections are joined together in a paragraph.

Pages like https://neovim.io/doc/user/insert.html#insert.txt are also hard to read.

I didn't take the time to examine the script that generates the online manuals.  However, I did see a recent commit to it in regards to word wrapping.

### Steps to reproduce

View the reference manual online.

### Expected behavior

I expect the the contents of the online user manual to be formatted/wrapped similarly to the how neovim shows the same manual.

### Neovim version (nvim -v)

Not applicable

Not applicable

### Operating system/version

Not applicable

### Terminal name/version

Not applicable

Not applicable

### Installation

Not applicable

wookayin:
Problem:

- Since #28678, pre-formatted text in the online documentation do not
  render whitespaces correctly: should be pre-like text, but shown like
  normal paragraph (see #28754).

- Code blocks with long lines should not be wrapped (e.g. see
  |dev-vimpatch-list-management|).

Solution:

- Use `white-space: pre-wrap`. Compared to `white-space: pre`, this
  option will make long lines including a very long URL wrapped.
  This properly fixes #28754 as well as #28678.

- Use horizontal scrollbar for the code blocks that are horizontally too
  long, instead of wrapping text. This will make the code easy to read
  while the pre-text block not interfering with the navigation bar.

### DIFF
--- a/scripts/gen_help_html.lua
+++ b/scripts/gen_help_html.lua
@@ -1088,14 +1088,19 @@ local function gen_css(fname)
       padding-bottom: 10px;
       /* Tabs are used for alignment in old docs, so we must match Vim's 8-char expectation. */
       tab-size: 8;
-      white-space: normal;
+      white-space: pre-wrap;
       font-size: 16px;
       font-family: ui-monospace,SFMono-Regular,SF Mono,Menlo,Consolas,Liberation Mono,monospace;
       word-wrap: break-word;
     }
-    .old-help-para pre {
+    .old-help-para pre, .old-help-para pre:hover {
       /* Text following <pre> is already visually separated by the linebreak. */
       margin-bottom: 0;
+      /* Long lines that exceed the textwidth should not be wrapped (no "pre-wrap").
+         Since text may overflow horizontally, we make the contents to be scrollable
+         (only if necessary) to prevent overlapping with the navigation bar at the right. */
+      white-space: pre;
+      overflow-x: auto;
     }
 
     /* TODO: should this rule be deleted? help tags are rendered as <code> or <span>, not <a> */


wookayin:
Codeblocks will look like this 
now looks like this

instead of, for example: https://neovim.io/doc/user/dev_vimpatch.html#dev-vimpatch-list-management

clason:
Apropos: while you're at it, can you add handling for the nodes
* `modeline` (do nothing, i.e., `return ''`)
* `note` (maybe make bold?)

scripts/gen_help_html.lua
```diff
@@ -1088,14 +1088,19 @@ local function gen_css(fname)
...
+      white-space: pre-wrap;
       font-size: 16px;
       font-family: ui-monospace,SFMono-Regular,SF Mono,Menlo,Consolas,Liberation Mono,monospace;
       word-wrap: break-word;
     }
-    .old-help-para pre {
+    .old-help-para pre, .old-help-para pre:hover {
       /* Text following <pre> is already visually separated by the linebreak. */
       margin-bottom: 0;
+      /* Long lines that does not fit within the textwidth should not be wrapped (no "pre-wrap").
```
justinmk, line 1099:
```suggestion
      /* Long lines that exceed the textwidth should not be wrapped (no "pre-wrap").
```

justinmk (approved):
This is definitely a needed fix since the previous PR, thanks.

BTW (off topic): `dev_vimpatch` should use "flow" layout. Fixed: https://github.com/neovim/neovim/pull/28747

wookayin:
Anything missing before 0.10?

clason:
(The 0.10 release is unrelated because the docs are always generated from HEAD.)

PR #28766, opened 2024-05-16
docs(gen_help_html.lua): handle modeline and note nodes

wookayin:
https://github.com/neovim/neovim/pull/28755#issuecomment-2111725105

Problem:

'modeline' and 'note' are unhandled in the online HTML documentation.

Some (not all) modelines are parsed by the vimdoc parser as a node of
type 'modeline'.

Solution:

- Ignore 'modeline' in HTML rendering.
- Render 'note' text in boldface.

Remark:

- Words like 'Deprecated' and 'Note' are parsed as 'note' rather than 'word'.
- I checked that there are no other unhandled nodes in the generated HTML files.

### DIFF
--- a/scripts/gen_help_html.lua
+++ b/scripts/gen_help_html.lua
@@ -530,6 +530,8 @@ local function visit_node(root, level, lang_tree, headings, opt, stats)
     return ('%s<a href="%s">%s</a>%s'):format(ws(), fixed_url, fixed_url, removed_chars)
   elseif node_name == 'word' or node_name == 'uppercase_name' then
     return text
+  elseif node_name == 'note' then
+    return ('<b>%s</b>'):format(text)
   elseif node_name == 'h1' or node_name == 'h2' or node_name == 'h3' then
     if is_noise(text, stats.noise_lines) then
       return '' -- Discard common "noise" lines.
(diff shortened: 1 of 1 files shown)


clason:
If some modelines aren't parsed, please open an issue (or PR ;)) at neovim/tree-sitter-vimdoc.

justinmk:
> Some (not all) modelines are parsed by the vimdoc parser as a node

does that mean we still need this hack: https://github.com/neovim/neovim/blob/cdd87222c86c5b2274a13d36f23de0637462e317/scripts/gen_help_html.lua#L195-L198

justinmk (approved):

clason:
Let's try without and see how it goes.

wookayin:
Yes, I agree that having a "proper" fix rather than a bandaid solution would be better: fixing the vimdoc parser to correctly handle all the modeline instances. I will take a look at the vimdoc parser.

clason:
@wookayin I meant this in the sense of "this PR is fine and can be merged as-is" -- it's still an improvement, and any improvement in the vimdoc parser will be picked up automatically.

clason:
The legacy modeline hack can be removed in a follow-up PR.

PR #28767, opened 2024-05-16
build(vim-patch.sh): include commit subject

wookayin:
Problem: vim-patch commits lack an informative title and summary in the
very first line of the commit message when the vim-revision is a Git SHA
hash, unlike when is a Vim version. This makes it difficult to discern
at a glance what changes are introduced by such vim-patch commits (in
git log, PR title, changelog generated by git-cliff, etc.).

e.g. Before:

    vim-patch:abcdef123456

    runtime(vim): improve performance

    <some details>
    ...

Solution: Repeat the title of the upstream commit message, to improve
the clarity and visibility of the commit message.

e.g. After:

    vim-patch:abcdef123456: runtime(vim): improve performance

    <some details>
    ...

Note: the `vim-patch:<hash>` token is still needed by `vim-patch.sh`
(but not necessarily in the very first line of the commit message) to
determine which vim patches have been applied.

### DIFF
--- a/scripts/vim-patch.sh
+++ b/scripts/vim-patch.sh
@@ -122,7 +122,7 @@ commit_message() {
   if [[ "${vim_message}" == "vim-patch:${vim_version}:"* ]]; then
     printf '%s\n\n%s\n\n%s' "${vim_message}" "${vim_commit_url}" "${vim_coauthors}"
   else
-    printf 'vim-patch:%s\n\n%s\n\n%s\n\n%s' "$vim_version" "$vim_message" "$vim_commit_url" "$vim_coauthors"
+    printf 'vim-patch:%s: %s\n\n%s\n\n%s' "${vim_version:0:7}" "${vim_message}" "${vim_commit_url}" "${vim_coauthors}"
   fi
 }
 


wookayin:
We consider the 'new' commit message style starting from 11.0-dev cycle. Although we might discuss first to agree which format would be the best: since the prefix might be a bit too lengthy, we could also consider some alternative options like:

1. More abbreviated SHA hash: e.g. `vim-patch:abcdef01: runtime(vim): improve performance`
  - con: we might want lengthy enough SHA hashes (8+ or 12 chars), normalization might be an issue
  
2. Or do not include the SHA in the first-line title: e.g. `vim-patch: runtime(vim): improve performance`
  - Since the SHA hash itself is not very useful for human other than the internal purposes for `vim-patch.sh`
  - Note: the tokens like `vim-patch: abcdef012345` is still needed in the commit message body.

zeertzjq:
Actually the token only needs to have 7 hex digits:
```sh
  # Use sed…{7,7} to normalize (internal) Git hashes (for tokens caches).
  git -C "${NVIM_SOURCE_DIR}" log -E --grep='vim-patch:[^ ,{]{7,}' \
    | grep -oE 'vim-patch:[^ ,{:]{7,}' \
    | sort \
    | uniq \
    | sed -nEe 's/^(vim-patch:([0-9]+\.[^ ]+|[0-9a-z]{7,7})).*/\1/p'
```
```sh
    # Check for vim-patch:<commit_hash> (usually runtime updates).
    token="vim-patch:${line:0:7}"
    if [[ "${tokens[$token]-}" ]]; then
      continue
    fi
```

wookayin:
For this PR I've drafted with not changing the number of hex digits, but I will leave to other core dev's decision whether to use 12 hex digits (current, since 03a04172ec911a4334d814b87f4c9ecc805b817e), 7 digits, or other number of digits (say 8).

dundargoc (approved):

justinmk:
> will leave to other core dev's decision whether to use 12 hex digits (current, since [03a0417](https://github.com/neovim/neovim/commit/03a04172ec911a4334d814b87f4c9ecc805b817e)), 7 digits, or other

I generally recommend 12 digits, but for vim-patch it's low risk.

PR #28775, opened 2024-05-16
fix(man.vim): signcolumn causes broken wrap

belkka:
This is how man page looks like when using neovim as a man pager and having `signcolumn=yes`. Note the line wraps in Description:

man man
Note: I did **not** resize the window after running the command.

Steps to reproduce:
1. `set signcolumn=yes` in `~/.config/nvim/init.vim`
2. `MANPAGER='nvim +Man!' man man`

The suggested solution is to `set signcolumn=auto` in `ftplugin/man.vim`, because man pages are especially sensitive to terminal width. `man` hard-wraps content (using terminal width as a default value for MANWIDTH), not taking pager's UI into account (the default MANPAGER is `less`, which displays content at full width of the terminal)

`ftplugin/man.vim` already contains similar options (e.g. `set nonumber`) to address the issue. The benefit of those tweaks is lost if ANY of UI columns appear.

### DIFF
--- a/runtime/ftplugin/man.vim
+++ b/runtime/ftplugin/man.vim
@@ -8,14 +8,16 @@ let b:did_ftplugin = 1
 
 setlocal noexpandtab tabstop=8 softtabstop=8 shiftwidth=8
 setlocal wrap breakindent linebreak
+setlocal colorcolumn=0 nolist
 
 " Parentheses and '-' for references like `git-ls-files(1)`; '@' for systemd
 " pages; ':' for Perl and C++ pages.  Here, I intentionally omit the locale
 " specific characters matched by `@`.
 setlocal iskeyword=@-@,:,a-z,A-Z,48-57,_,.,-,(,)
 
-setlocal nonumber norelativenumber
-setlocal foldcolumn=0 colorcolumn=0 nolist nofoldenable
+" man page content is likely preformatted for the terminal width, so
+" narrowing display by any additional columns leads to Embarrassing Line Wrap
+setlocal nonumber norelativenumber foldcolumn=0
 
 setlocal tagfunc=v:lua.require'man'.goto_tag
 
@@ -35,6 +37,8 @@ if get(g:, 'ft_man_folding_enable', 0)
   setlocal foldenable
   setlocal foldmethod=indent
   setlocal foldnestmax=1
+else
+  setlocal nofoldenable
 endif
 
 let b:undo_ftplugin = ''
--- a/runtime/ftplugin/man.vim
+++ b/runtime/ftplugin/man.vim
@@ -17,7 +17,7 @@ setlocal iskeyword=@-@,:,a-z,A-Z,48-57,_,.,-,(,)
 
 " man page content is likely preformatted for the terminal width, so
 " narrowing display by any additional columns leads to Embarrassing Line Wrap
-setlocal nonumber norelativenumber foldcolumn=0
+setlocal nonumber norelativenumber foldcolumn=0 signcolumn=auto
 
 setlocal tagfunc=v:lua.require'man'.goto_tag
 


runtime/ftplugin/man.vim
```diff
@@ -8,14 +8,16 @@ let b:did_ftplugin = 1
...
 " Parentheses and '-' for references like `git-ls-files(1)`; '@' for systemd
 " pages; ':' for Perl and C++ pages.  Here, I intentionally omit the locale
 " specific characters matched by `@`.
 setlocal iskeyword=@-@,:,a-z,A-Z,48-57,_,.,-,(,)
 
-setlocal nonumber norelativenumber
-setlocal foldcolumn=0 colorcolumn=0 nolist nofoldenable
+" man page content is likely preformatted for the terminal width, so
+" narrowing display by any additional columns leads to Embarrassing Line Wrap
+setlocal nonumber norelativenumber foldcolumn=0 signcolumn=auto
```
justinmk, line 20: disabling random things is not ideal. instead, find where the width (MANWIDTH, iirc) is calculated in this plugin, and adjust it

justinmk (changes requested):
disabling random things is not ideal. instead, find where the width (MANWIDTH, iirc) is calculated in this plugin, and adjust it

runtime/ftplugin/man.vim
belkka: I don't think `MANWIDTH` is controlled by plugin when calling `man` from CLI. Personally, I have `MANPAGER` set in my environment as described in `:help man`, and just call `man echo` in bash (see steps to reproduce).

Note, that this PR only adds `signcolumn=auto` (may be easier to see when viewing separate commits). Other "random things" (visual columns occupied by editor's UI, to be precise) are already disabled. Suggested changes follow the existing approach and improve it (I believe), not aiming to seek for ideal.

If someone decide to spend time finding an "ideal" solution, they could still benefit from these changes by using them as a reference. E. g. just like they will see "nonumber", "norelativenumber", "foldcolumn" and include the width of these columns into their calculations of MANWIDTH (or whatever is relevant), they will also see the "signcolumn" and remember to take it into account as well.

Also, IMHO, benefits of what you call an "ideal" approach are questionable. ftplugins are specifically intended to improve user experience with specific `filetype`s by taking into account the meaning of the file type. Disabling/enabling specific things for specific filetypes feels totally reasonable for me. Do you think that other `setlocal` options <s>are not ideal</s> should be avoided in this ftplugin (e. g. `nolist`)?

belkka:
@justinmk how did you make your message appear in both "review comment" and PR conversation? Duplicating my response for clarity:

> I don't think MANWIDTH is controlled by plugin when calling man from CLI. Personally, I have MANPAGER set in my environment as described in :help man, and just call man echo in bash (see steps to reproduce).
> 
> Note, that this PR only adds signcolumn=auto (may be easier to see when viewing separate commits). Other "random things" (visual columns occupied by editor's UI, to be precise) are already disabled. Suggested changes follow the existing approach and improve it (I believe), not aiming to seek for ideal.
> 
> If someone decide to spend time finding an "ideal" solution, they could still benefit from these changes by using them as a reference. E. g. just like they will see "nonumber", "norelativenumber", "foldcolumn" and include the width of these columns into their calculations of MANWIDTH (or whatever is relevant), they will also see the "signcolumn" and remember to take it into account as well.
> 
> Also, IMHO, benefits of what you call an "ideal" approach are questionable. ftplugins are specifically intended to improve user experience with specific filetypes by taking into account the meaning of the file type. Disabling/enabling specific things for specific filetypes feels totally reasonable for me. Do you think that other setlocal options are not ideal should be avoided in this ftplugin (e. g. nolist)?

Marked that thread as resolved because suggested changes are both
- impossible. I believe `man` (invoked from shell) pipes its pre-formatted output to stdin of `$MANPAGER`; so no way for `$MANPAGER` — `nvim` in this case — to affect the environment (`$MANWIDTH`) of `man`
- out of scope of this PR (even if it was possible, e.g. when `:Man` is called from `nvim`), because current changes are only aimed to improve a minor flaw of existing approach (a single missing option), not to find completely another solution. Also, they don't make it harder to change the approach later (which should be a separate issue); in fact, the suggested refactoring (first commit) makes it easier.

I've finally updated commit messages to follow contribution guidelines (https://conventionalcommits.org), so removing "draft" status from this PR.

runtime/ftplugin/man.vim
justinmk: This is fine, since we already were doing `nonumber norelativenumber`, and :Man always opens a new window anyway (otherwise restoring window-local options is tricky).

why signcolumn=auto instead of signcolumn=no ?

justinmk (approved):

runtime/ftplugin/man.vim
belkka: @justinmk 
> why auto instead of no

my reasoning was: as far as I know, sings are not used in Vim *by default*, and therefore they only appear when user **explicitly configures them** (e. g. installs a plugin for that purpose). I only know about plugins that use signs for editable files (integration of git repos) or specific filetypes (integration of linters, compilers, build systems). So:
- in all situations that I can imagine, signcolumn is empty anyway in ft=man, so no difference between `auto` and `no`
- if by any chance there ARE some signs in singcolumn for ft=man, then user must have made an explicit **effort** to add them, and they could be annoyed by the fact that signs are still not visible.

On the other side, the bad scenario for this decision (using auto instead of no) is when user (1) has signcolumn=no in their global config AND (2) still creates some signs when browsing man pages (maybe unintentionally, with a plugin). It may be annoying for them to see signcolumn appear given that they have explicitly disabled it in `init.vim`. Regarding this case:
1. I'm fine with signcolumn=no if anyone believes it is better.
2. A more clever solution could be to check the current value of signcolumn and only downgrade from "yes" to "auto", but keep the original value otherwise. This is more in line with the motivation of the proposed change — to prevent signcolumn from being "yes" (which may be good as a global value, but bad for ft=man).

I'm equally ready to implement any of two approaches above or keep the current one.

belkka:
image
For the sake of clarity: "embarrassing line wrap" idiom that I've used refers to https://datatracker.ietf.org/doc/html/rfc3676#section-3.2
I didn't mean to attack anyone with the title :)

Also note, I've used this term in a comment: 
https://github.com/neovim/neovim/pull/28775/commits/3cf1c70c702786d7921438113818be3a1a2f8bd5#diff-da5cc7c2d752545d38ac134c454d1d71987c9a5df051fa2e5c34fba6bf768402R19

If anyone thinks it's inappropriate, I should edit the commit as well (change to "broken wrap"?)

justinmk:
https://github.com/neovim/neovim/pull/29326 is a reminder about handling window-local options carefully. In this case we are using `:setlocal` which is equivalent.

> refers to https://datatracker.ietf.org/doc/html/rfc3676#section-3.2 I didn't mean to attack anyone with the title :)

No worries, the wording was just a bit strange. I didn't know about that reference, thanks.

PR #28778, opened 2024-05-16
fix: change deprecation presentation

dundargoc:
Deprecation with vim.deprecate is currently too noisy. Show the
following warning instead:

[function] is deprecated. Run ":checkhealth vim.deprecated" for more information.

The important part is that the full message needs to be short enough to
fit in one line in order to not trigger the "Press ENTER or type command
to continue" prompt.

The full information and stack trace for the deprecated functions will
be shown in the new healthcheck `vim.deprecated`.

### DIFF
--- a/runtime/lua/vim/_editor.lua
+++ b/runtime/lua/vim/_editor.lua
@@ -1031,6 +1031,42 @@ function vim._cs_remote(rcid, server_addr, connect_error, args)
   }
 end
 
+do
+  local function truncated_echo(msg)
+    -- Truncate message to avoid hit-enter-prompt
+    local max_width = vim.o.columns * math.max(vim.o.cmdheight - 1, 0) + vim.v.echospace
+    local msg_truncated = string.sub(msg, 1, max_width)
+    vim.api.nvim_echo({ { msg_truncated, 'WarningMsg' } }, true, {})
+  end
+
+  local notified = false
+
+  function vim._truncated_echo_once(msg)
+    if not notified then
+      truncated_echo(msg)
+      notified = true
+      return true
+    end
+    return false
+  end
+end
+
+--- This is basically the same as debug.traceback(), except the full paths are shown.
+local function traceback()
+  local level = 4
+  local backtrace = { 'stack traceback:' }
+  while true do
+    local info = debug.getinfo(level, 'Sl')
+    if not info then
+      break
+    end
+    local msg = ('  %s:%s'):format(info.source:sub(2), info.currentline)
+    table.insert(backtrace, msg)
+    level = level + 1
+  end
+  return table.concat(backtrace, '\n')
+end
+
 --- Shows a deprecation message to the user.
 ---
 ---@param name        string     Deprecated feature (function, API, etc.).
@@ -1043,12 +1079,12 @@ end
 ---@return string|nil # Deprecated message, or nil if no message was shown.
 function vim.deprecate(name, alternative, version, plugin, backtrace)
   plugin = plugin or 'Nvim'
-  local will_be_removed = 'will be removed'
-
-  -- Only issue warning if feature is hard-deprecated as specified by MAINTAIN.md.
-  -- Example: if removal_version is 0.12 (soft-deprecated since 0.10-dev), show warnings starting at
-  -- 0.11, including 0.11-dev
   if plugin == 'Nvim' then
+    require('vim.deprecated.health').add(name, version, traceback(), alternative)
+
+    -- Only issue warning if feature is hard-deprecated as specified by MAINTAIN.md.
+    -- Example: if removal_version is 0.12 (soft-deprecated since 0.10-dev), show warnings starting at
+    -- 0.11, including 0.11-dev
     local major, minor = version:match('(%d+)%.(%d+)')
     major, minor = tonumber(major), tonumber(minor)
 
@@ -1059,8 +1095,12 @@ function vim.deprecate(name, alternative, version, plugin, backtrace)
       return
     end
 
-    local removal_version = string.format('nvim-%d.%d', major, minor)
-    will_be_removed = vim.fn.has(removal_version) == 1 and 'was removed' or will_be_removed
+    local msg = ('%s is deprecated. Run ":checkhealth vim.deprecated" for more information'):format(
+      name
+    )
+
+    local displayed = vim._truncated_echo_once(msg)
+    return displayed and msg or nil
   else
     vim.validate {
       name = { name, 'string' },
@@ -1068,22 +1108,16 @@ function vim.deprecate(name, alternative, version, plugin, backtrace)
       version = { version, 'string', true },
       plugin = { plugin, 'string', true },
     }
-  end
 
-  local msg = ('%s is deprecated'):format(name)
-  msg = alternative and ('%s, use %s instead.'):format(msg, alternative) or (msg .. '.')
-  msg = ('%s%s\nFeature %s in %s %s'):format(
-    msg,
-    (plugin == 'Nvim' and ' :help deprecated' or ''),
-    will_be_removed,
-    plugin,
-    version
-  )
-  local displayed = vim.notify_once(msg, vim.log.levels.WARN)
-  if displayed and backtrace ~= false then
-    vim.notify(debug.traceback('', 2):sub(2), vim.log.levels.WARN)
+    local msg = ('%s is deprecated'):format(name)
+    msg = alternative and ('%s, use %s instead.'):format(msg, alternative) or (msg .. '.')
+    msg = ('%s\nFeature will be removed in %s %s'):format(msg, plugin, version)
+    local displayed = vim.notify_once(msg, vim.log.levels.WARN)
+    if displayed and backtrace ~= false then
+      vim.notify(debug.traceback('', 2):sub(2), vim.log.levels.WARN)
+    end
+    return displayed and msg or nil
   end
-  return displayed and msg or nil
 end
 
 require('vim._options')
--- a/runtime/lua/vim/deprecated/health.lua
+++ b/runtime/lua/vim/deprecated/health.lua
@@ -1,0 +1,42 @@
+local M = {}
+local health = vim.health
+
+local deprecated = {}
+
+function M.check()
+  if next(deprecated) == nil then
+    health.ok('No deprecated functions detected')
+    return
+  end
+
+  for name, v in vim.spairs(deprecated) do
+    health.start('')
+
+    local version, backtraces, alternative = v[1], v[2], v[3]
+    local major, minor = version:match('(%d+)%.(%d+)')
+    major, minor = tonumber(major), tonumber(minor)
+    local removal_version = string.format('nvim-%d.%d', major, minor)
+    local will_be_removed = vim.fn.has(removal_version) == 1 and 'was removed' or 'will be removed'
+
+    local msg = ('%s is deprecated. Feature %s in Nvim %s'):format(name, will_be_removed, version)
+    local msg_alternative = alternative and ('use %s instead.'):format(alternative)
+    local advice = { msg_alternative }
+    table.insert(advice, backtraces)
+    advice = vim.iter(advice):flatten():totable()
+    health.warn(msg, advice)
+  end
+end
+
+function M.add(name, version, backtrace, alternative)
+  if deprecated[name] == nil then
+    deprecated[name] = { version, { backtrace }, alternative }
+    return
+  end
+
+  local it = vim.iter(deprecated[name][2])
+  if it:find(backtrace) == nil then
+    table.insert(deprecated[name][2], backtrace)
+  end
+end
+
+return M
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -1279,9 +1279,7 @@ M.handlers.signs = {
           vim.deprecate(
             'Defining diagnostic signs with :sign-define or sign_define()',
             'vim.diagnostic.config()',
-            '0.12',
-            nil,
-            false -- suppress backtrace
+            '0.12'
           )
 
           if not opts.signs.text then
--- a/test/functional/lua/vim_spec.lua
+++ b/test/functional/lua/vim_spec.lua
@@ -135,42 +135,46 @@ describe('lua stdlib', function()
     -- See MAINTAIN.md for the soft/hard deprecation policy
 
     describe(('vim.deprecate prerel=%s,'):format(prerel or 'nil'), function()
-      it('plugin=nil', function()
-        local curver = exec_lua('return vim.version()') --[[@as {major:number, minor:number}]]
-        -- "0.10" or "0.10-dev+xxx"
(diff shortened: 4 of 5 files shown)


clason:
The message can still contain specifics (like the name of the function, and possibly the file?) -- the important thing is that it is guaranteed to be a single line so it doesn't trigger the dreaded "press enter" prompt. In particular, the full stack trace should be punted to the health check.

dundargoc:
TODO:

- Check: https://github.com/echasnovski/mini.nvim/blob/c333187fcc76d7e772dac32c2a440a949fe34be4/lua%2Fmini%2Fai.lua#L1928-L1944

Apparently it's nonblocking or something

- Review wasmtime PR

justinmk:
> the important thing is that it is guaranteed to be a single line so it doesn't trigger the dreaded "press enter"

To be clear: it can occupy 'cmdheight' * `v:echospace` , which may be more than one line. Main goal is to avoid hit-enter prompt. `strcharpart()` may be useful.

dundargoc:
A problem is that even if we truncate, if there are two deprecation messages they will together pass the limit and trigger the hit enter prompt

dundargoc:
Prerequisite: https://github.com/neovim/neovim/pull/28836

dundargoc:
I need to fix the tests, but the behavior itself should be done. Review away.

runtime/lua/vim/shared.lua
```diff
@@ -553,7 +553,7 @@ end
 ---@param t table List-like table
 ---@return table Flattened copy of the given list-like table
 function vim.tbl_flatten(t)
-  vim.deprecate('vim.tbl_flatten', 'vim.iter(…):flatten():totable()', '0.13')
+  vim._deprecate('vim.tbl_flatten', 'vim.iter(…):flatten():totable()', '0.12')
```
wookayin, line 556: I think we wanted to extend its life until 0.13? #28780

runtime/lua/vim/_editor.lua
```diff
@@ -1042,50 +1042,75 @@ end
...
+    -- Truncate message to avoid hit-enter-prompt
+    local max_width = vim.o.columns * math.max(vim.o.cmdheight - 1, 0) + vim.v.echospace
+    local msg_truncated = string.sub(msg, 1, max_width)
+    vim.api.nvim_echo({ { msg_truncated, 'WarningMsg' } }, true, {})
+  end
+
+  local notified = false
+
+  function vim._notify_once_truncated(msg)
+    if not notified then
```
wookayin, line 1073: Unlike the semantic of `vim.notify_once` where only one notification is made per `msg`, this will suppress echoing for *any* further message. I can get why you chose this way --- to prevent multiple messages printed during startup and hence the 'hit-enter-prompt' (even if there are 2+ different messages). (Also, nitpicking on name: it's not using "vim.notify").

However, there is one problem about this. `vim._deprecate` warnings (for Nvim Lua APIs) *can* still happen again after vim startup and a warning message is shown once, when a deprecated function is used in later calls by plugin or user config upon some events, scheduled asynchronously, etc. Those warnings will be silenced, although collected in `:checkhealth vim.deprecate`.

runtime/lua/vim/_editor.lua
```diff
@@ -1042,50 +1042,75 @@ end
 ---
 ---@return string|nil # Deprecated message, or nil if no message was shown.
 function vim.deprecate(name, alternative, version, plugin, backtrace)
```
wookayin, line 1044: Although it's not very important, the behavior of `vim._deprecate` (plugin = 'Nvim') and `vim.deprecate` (for plugins?) are different. I wonder if we can make it more consistent while making deprecation warnings less annoying as intended.

wookayin (changes requested):
I left some comments on the behavior of `_notify_once_truncated` and `vim._deprecate`.

runtime/lua/vim/deprecated/health.lua
```diff
@@ -0,0 +1,28 @@
...
+    local msg = ('%s is deprecated. Feature %s in Nvim %s'):format(name, will_be_removed, version)
+    local msg_alternative = alternative and ('use %s instead.'):format(alternative)
+    local advice = { msg_alternative }
+    table.insert(advice, backtrace)
+    health.warn(msg, advice)
+  end
+end
+
+function M.add(name, alternative, version, backtrace)
+  deprecated[name] = { alternative, version, backtrace }
```
wookayin, line 25: Does it mean that we only collect one backtrace for each deprecated API? I think it would be good to collect multiple invocations. For instance, `vim.tbl_islist` is used in many (20+) places, and one might want to see them all at once rather than `:checkhealth` as many times.

MariaSolOs (approved):

runtime/lua/vim/_editor.lua
```diff
@@ -1042,50 +1042,75 @@ end
...
     vim.notify(debug.traceback('', 2):sub(2), vim.log.levels.WARN)
   end
   return displayed and msg or nil
 end
 
+do
+  local function truncated_echo(msg)
+    -- Truncate message to avoid hit-enter-prompt
+    local max_width = vim.o.columns * math.max(vim.o.cmdheight - 1, 0) + vim.v.echospace
+    local msg_truncated = string.sub(msg, 1, max_width)
```
echasnovski, line 1066:
```suggestion
    local msg_truncated = vim.fn.strcharpart(msg, 0, max_width)
```

Using `strcharpart()` makes it will work with multibyte characters. Unless `truncatead_echo()` will be used in a "fast" context while `strcharpart()` can not. But it seems like `strcharpart()` should be usable in fast context.

runtime/lua/vim/shared.lua
dundargoc: Oops, thanks.

runtime/lua/vim/_editor.lua
dundargoc: I don't get it. vim.notify triggers the press enter prompt. How does using that help?

runtime/lua/vim/_editor.lua
dundargoc: We could change vim.deprecate so plugins use the new deprecation functionality as well.. Do you think that's the way to go?

runtime/lua/vim/_editor.lua
wookayin: Hmm actually you're right. I was somehow assuming users would be overriding `vim.notify` with [some plugins](https://github.com/rcarriga/nvim-notify) so that it doesn't write into `:message`. But if such plugins are not used, i.e. with the factory default setting of `vim.notify`, it will trigger the hit-enter-prompt (no matter vim has entered or not). Never mind about the potential solution suggested. If we'd want to completely eliminate the hit-enter-prompt, we shouldn't use `vim.notify` at all.

runtime/lua/vim/_editor.lua
wookayin: If we go with this PR, then plugins should use the same way of showing deprecation messages as Nvim. Though `vim.deprecate` no longer benefits from `vim.notify` and `vim.notify_once`.

runtime/lua/vim/_editor.lua
dundargoc: On the other hand a non-blocking notification likr the plugin could also work, although I suspect it'll be more ambitious.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1276,12 +1276,10 @@ M.handlers.signs = {
...
         if sign then
           local severity = M.severity[v:upper()]
-          vim.deprecate(
+          vim._deprecate(
             'Defining diagnostic signs with :sign-define or sign_define()',
             'vim.diagnostic.config()',
-            '0.12',
-            nil,
-            false
+            '0.12'
```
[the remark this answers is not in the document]
dundargoc: Let's just merge your PR, I don't wanna think about this

runtime/lua/vim/_editor.lua
zeertzjq: The `defaults to "Nvim"` in the comment above is no longer true.

clason:
> Though vim.deprecate no longer benefits from vim.notify and vim.notify_once.

I've been public about `vim.notify` being a bad design (or at least premature without rethinking the underlying message subsystem). So I think that would be ok for the time being; obviously when the redesign lands, we'll have a (better) API for capturing and routing such messages (e.g., via `vim.ui_attach`).

dundargoc:
TODO: add opts table at end to dogfood a useful potential public API

runtime/lua/vim/_editor.lua
dundargoc: > If we go with this PR, then plugins should use the same way of showing deprecation messages as Nvim. Though `vim.deprecate` no longer benefits from `vim.notify` and `vim.notify_once`.

I'm not sure they should. I think a problem with vim.deprecate is that we made it public too soon and it just grew without much consideration. I think giving the internal deprecation mechanism to mature a bit would help us more in the long run.

dundargoc:
deprecated

runtime/lua/vim/_editor.lua
```diff
@@ -1036,56 +1036,97 @@ end
...
+    vim.api.nvim_echo({ { msg_truncated, 'WarningMsg' } }, true, {})
+  end
+
+  local notified = false
+
+  function vim._truncated_echo_once(msg)
+    if not notified then
+      truncated_echo(msg)
+      notified = true
```
wookayin, line 1074: I think https://github.com/neovim/neovim/pull/28778/files#r1606057221 is still applicable. `vim._deprecate` should give warning messages for any subsequent calls after the first invocation.

Kind of re-inventing the wheel like `vim.notify_once`, but:
```suggestion
    if not notified[msg] then
      truncated_echo(msg)
      notified[msg] = true
```

runtime/lua/vim/_editor.lua
```diff
@@ -1036,56 +1036,97 @@ end
...
+
+--- Shows a deprecation message to the user. Internal function customized to fit neovim's needs.
+---
+---@param name        string     Deprecated feature (function, API, etc.).
+---@param version     string     Version when the deprecated function will be removed.
+---@param opts        table|nil  Optional parameters.
+---                              - alternative: suggested replacement function
+---
+---@return string|nil # Deprecated message, or nil if no message was shown.
+function vim._deprecate(name, version, opts)
```
wookayin, line 1104: Changing signature to avoid `nil` positional arguments would be a good idea, one thing to think about the future: do we plan to eventually change the API of `vim.deprecate` match this (or supporting both signatures for a while ~~with deprecating `vim.deprecate`~~)?

If not for any reason, we should use a similar signature and parameters.

dundargoc: > Changing signature to avoid `nil` positional arguments would be a good idea, one thing to think about the future: do we plan to eventually change the API of `vim.deprecate` match this (or supporting both signatures for a while ~~with deprecating `vim.deprecate`~~)?

Not a definitive plan, but a strong possibility.

runtime/lua/vim/_editor.lua
dundargoc: Uh, you suggested I use notify which doesn’t work. What am I supposed to do?

runtime/lua/vim/diagnostic.lua
```diff
@@ -1276,12 +1276,10 @@ M.handlers.signs = {
...
         if sign then
           local severity = M.severity[v:upper()]
-          vim.deprecate(
+          vim._deprecate(
             'Defining diagnostic signs with :sign-define or sign_define()',
-            'vim.diagnostic.config()',
             '0.12',
-            nil,
-            false -- suppress backtrace
+            { alternative = 'vim.diagnostic.config()' }
```
wookayin, line 1282: Note: For now `vim._deprecate` does not have `backtrace` in opts, so this will no longer suppress backtrace.

Though it'd be even better if we can allow suppressing backtraces (not mandatory), adding stacktrace back here for this particular instance would be also fine, as the stacktrace information (which we think is useless) will be shown only in `:checkhealth` rather than as a message.

dundargoc: Do you anticipate suppressing backtracea in checkhealth being a real usecase for us?

dundargoc:
TODO: make function name opts arg. I can get it with some debug function like with t.skip

runtime/lua/vim/_editor.lua
justinmk, line 1104: can we avoid the `_deprecate` variant, and have just one function

clason: How else are we going to deprecate `vim.deprecate`?

clason: On a more serious note, this will allow us to get the interface right _first_ before replacing the old `vim.deprecate`.

justinmk: In that case we should support the old parameters/arity, so that plugins don't need to "migrate" once we swap over the new version. We can support kwargs as an alternative, for new callers, if needed.

dundargoc: We can but there'd be no point. It would essentially be two functions baked into one with a phat if/else check. The traceback argument would always be ignored for Nvim. It's just a mess in general. If you absolutely insist then I don't mind, but I really don't see the benefit.

And as clason said the current function signature is bad and not easily extensible as we don't have an opts table. I think we need to cut our losses with the current deprecation function and work on something new.

dundargoc: > In that case we should support the old parameters/arity, so that plugins don't need to "migrate" once we swap over the new version. We can support kwargs as an alternative, for new callers, if needed.

Not a fan. Variable number of arguments is not a good design becausenit is not extensible.

justinmk: > I think we need to cut our losses with the current deprecation function and work on something new.

how are we going to cut our losses if we have to support both?

dundargoc: I meant until we remove `vim.deprecate`.

dundargoc: OK, enough lazy backseat reviewing, I need actual guidance: do we add plugin deprecations into this mechanism or not? That completely changes the solution. It would technically be breaking though if we do include plugins into this behavior.

dundargoc: Fuck it, I'm reverting it then.

dundargoc: I'll revert this change then.

dundargoc:
@justinmk review

runtime/lua/vim/_editor.lua
dundargoc: I've reverted it to use old function signature and to only use a single function. Are you open to "fixing" the function signature in the future to comply with our guidelines (meaning, mandatory arguments first and optional arguments in an `opts` table)? I'd hate to solidify the current function signature in all eternity, it's just not good unfortunately.

dundargoc:
@justinmk review

PR #28781, opened 2024-05-16
fix(vim.iter): enable optimizations for arrays (lists with holes)

### LINKED ISSUE #28777: `vim.iter` with non-list tables (as a replacement of `vim.tbl_flatten`, etc.)
This issue would be more like a discussion, only stating potential issues from the perspective of practical user experience without a clear solution yet.

TL;DR) `vim.iter` have some *practical* problems when dealing with non-list tables (or *pseudo*-lists that contains `nil` as element), we might want to improve for better usability.

## Problem

`vim.tbl_flatten` has been deprecated since nvim 0.10+ in favor of `vim.iter`, but what is suggested --- `vim.iter(...):flatten():totable()` is NOT a exact drop-in replacement of `vim.tbl_flatten`.

A corner case is when the table is "not a list", e.g. containing a `nil` values:

```lua
-- The table {1, nil, 3} is not a valid list (i.e. `vim.islist` returns false).
-- Also equivalent to {[1] = 1, [3] = 3}

vim.iter({ 1, nil, 3 }):flatten():totable()
-- ERROR: flatten() requires a list-like table

vim.tbl_flatten({ 1, nil, 3 })
-- { 1, 3 }
```

... because `vim.iter` treats list-like tables and non-list-like tables (i.e. maps) differently:

```lua
vim.iter({ 1, nil, 3 }):totable()
-- { {1, 1}, {3, 3} }

vim.iter({ 1, 2, 3 }):totable()
-- { 1, 2, 3 }
```

This is due to the limitation of the Lua language that Lua table does not distiniguish map and list(array), so a valid list contains. While this is actually an intended, correct behavior, given how Lua works, for users and laypeople who are not very familiar with Lua the behavior can be confusing: because `totable()` returns a list of (key, value) tuple for non-list-like tables.

### Remark: it was an undefined behavior of `tbl_flatten()` 

Actually, the use of `vim.tbl_flatten()` for tables that are **not** list-like (e.g. including `nil` as a member) was never specified:

```
tbl_flatten({t})                                           *vim.tbl_flatten()*
    Creates a copy of a list-like table such that any nested tables are
    "unrolled" and appended to the result.

    Parameters:  
      • {t}  (table) List-like table

    Return:  
        (table) Flattened copy of the given list-like table

    See also:  
      • From https://github.com/premake/premake-core/blob/master/src/base/table.lua

```

However, it is worth noting that `vim.tbl_flatten()` is often slightly abused in the wild to filter `nil` values while doing flattening:

* https://github.com/nvim-telescope/telescope.nvim/blob/52f500110bcf9190b44b4d8640162adc86772ec4/lua/telescope/builtin/__git.lua#L54-L58
* https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/617
* and many more: https://grep.app/search?q=vim.tbl_flatten&filter[lang][0]=Lua

Note: the behavior of `vim.tbl_flatten()` on such 'lists with holes (nil)' is not very accurate (as of nvim 0.10.0); it's not a 100% correct replacement of "filter nil values and then flatten":

```lua
> vim.tbl_flatten { nil, 1, nil, 3 }
< { 1, 3 }
> vim.tbl_flatten { 1, nil, 3 }
< { 1 }   -- ?!??!, should be { 1, 3 }

> vim.tbl_filter(function(v) return v end, { 1, nil, 3 })
< { 1, 3 }  -- correct
```

### Potential impact:

Starting from the 0.11-dev (nightly) versions users will experience a looooot of deprecation warnings on the deprecated APIs, `vim.tbl_flatten`, `vim.tbl_islist`, etc. There are actually a lot of plugins that are still using `vim.tbl_flatten` which was deprecated only very recently. So plugin authors should migrate to `vim.iter`-based solution (unless we make additional changes) and remove the use of deprecated APIs sooner than later, but the subtle difference can lead to (minor) bugs.

## Possible solutions

Either:

- We don't change anything. Using `nil` in a list is an anti-pattern ?!
- Implement `flatten()` for non-list-like (dict-like) tables.
- Add some new options to `vim.iter` API to make `vim.iter(table):totable()` return a "map" instead of a list of (key, value) tuples.
- Allow `vim.iter()` to handle lists with holes as well
  #28781
- ???

## References

- #28448 
- #28780 
- #28781

gpanders:
The optimizations that vim.iter uses for array-like tables don't strictly require that the underlying table has no holes. The only thing that needs to change is the determination if a table is "list-like": rather than requiring consecutive, integer keys, we can simply test for integer keys only, and remove any holes in the original array when we make a copy for the iterator.

This allows `vim.iter(t):flatten():totable()` to work as a proper replacement for `vim.tbl_flatten`.

Fixes #28777.

### DIFF
--- a/runtime/lua/vim/iter.lua
+++ b/runtime/lua/vim/iter.lua
@@ -111,8 +111,9 @@ end
 
 local function sanitize(t)
   if type(t) == 'table' and getmetatable(t) == packedmt then
-    -- Remove length tag
+    -- Remove length tag and metatable
     t.n = nil
+    setmetatable(t, nil)
   end
   return t
 end
@@ -414,7 +415,7 @@ function ArrayIter:totable()
     return Iter.totable(self)
   end
 
-  local needs_sanitize = getmetatable(self._table[1]) == packedmt
+  local needs_sanitize = getmetatable(self._table[self._head]) == packedmt
 
   -- Reindex and sanitize.
   local len = self._tail - self._head
--- a/runtime/lua/vim/iter.lua
+++ b/runtime/lua/vim/iter.lua
@@ -1031,7 +1031,7 @@ function Iter.new(src, ...)
 
     local t = {}
 
-    -- O(n): scan the source table to decide if it is an array (only integer indices).
+    -- O(n): scan the source table to decide if it is an array (only positive integer indices).
     local length = 0
     for k, v in pairs(src) do
       if type(k) ~= 'number' then
@@ -1042,7 +1042,6 @@ function Iter.new(src, ...)
       end
       t[k] = v
     end
-    -- print("creating new ArrayLiter with length", length)
     return ArrayIter.new(t, length)
   end
 
@@ -1073,7 +1072,9 @@ end
 
 --- Create a new ArrayIter
 ---
----@param t table Array-like table. Caller guarantees that this table is a valid array.
+---@param t table Array-like table. Caller guarantees that this table is a valid array. Can have
+---               holes (nil values).
+---@param length integer Length of the table (largest positive integer index).
 ---@return Iter
 ---@private
 function ArrayIter.new(t, length)
--- a/runtime/lua/vim/iter.lua
+++ b/runtime/lua/vim/iter.lua
@@ -1033,17 +1033,13 @@ function Iter.new(src, ...)
     local t = {}
 
     -- O(n): scan the source table to decide if it is an array (only positive integer indices).
-    local length = 0
     for k, v in pairs(src) do
-      if type(k) ~= 'number' or math.floor(k) ~= k then
+      if type(k) ~= 'number' or k <= 0 or math.floor(k) ~= k then
         return Iter.new(pairs(src))
       end
-      if k > length then
-        length = k
-      end
-      t[k] = v
+      t[#t + 1] = v
     end
-    return ArrayIter.new(t, length)
+    return ArrayIter.new(t)
   end
 
   if type(src) == 'function' then
@@ -1075,14 +1071,13 @@ end
 ---
 ---@param t table Array-like table. Caller guarantees that this table is a valid array. Can have
 ---               holes (nil values).
----@param length integer Length of the table (largest positive integer index).
 ---@return Iter
 ---@private
-function ArrayIter.new(t, length)
+function ArrayIter.new(t)
   local it = {}
   it._table = t
   it._head = 1
-  it._tail = length + 1
+  it._tail = #t + 1
   setmetatable(it, ArrayIter)
   return it
 end
--- a/runtime/lua/vim/iter.lua
+++ b/runtime/lua/vim/iter.lua
@@ -1034,7 +1034,7 @@ function Iter.new(src, ...)
     -- O(n): scan the source table to decide if it is an array (only positive integer indices).
     local length = 0
     for k, v in pairs(src) do
-      if type(k) ~= 'number' then
+      if type(k) ~= 'number' or math.floor(k) ~= k then
         return Iter.new(pairs(src))
       end
       if k > length then
--- a/runtime/lua/vim/iter.lua
+++ b/runtime/lua/vim/iter.lua
@@ -120,28 +120,25 @@ end
 
 --- Flattens a single array-like table. Errors if it attempts to flatten a
 --- dict-like table
----@param v table table which should be flattened
+---@param t table table which should be flattened
 ---@param max_depth number depth to which the table should be flattened
 ---@param depth number current iteration depth
 ---@param result table output table that contains flattened result
 ---@return table|nil flattened table if it can be flattened, otherwise nil
-local function flatten(v, max_depth, depth, result)
-  if depth < max_depth and type(v) == 'table' then
-    local i = 0
-    for _ in pairs(v) do
-      i = i + 1
-
-      if v[i] == nil then
+local function flatten(t, max_depth, depth, result)
+  if depth < max_depth and type(t) == 'table' then
+    for k, v in pairs(t) do
+      if type(k) ~= 'number' or k <= 0 or math.floor(k) ~= k then
         -- short-circuit: this is not a list like table
         return nil
       end
 
-      if flatten(v[i], max_depth, depth + 1, result) == nil then
+      if flatten(v, max_depth, depth + 1, result) == nil then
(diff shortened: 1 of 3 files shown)


clason:
> The only thing that needs to change is the determination if a table is "list-like": rather than requiring consecutive, integer keys, we can simply test for integer keys only.

Luckily, we have a function for that as well: `vim.isarray()` :)

test/functional/lua/iter_spec.lua
```diff
@@ -482,18 +485,18 @@ describe('vim.iter', function()
...
 
-    local not_list_like = vim.iter({ [2] = 2 })
-    matches(flat_err, pcall_err(not_list_like.flatten, not_list_like))
+    local not_list_like = { [2] = 2 }
+    eq({ 2 }, vim.iter(not_list_like):flatten():totable())
 
-    local also_not_list_like = vim.iter({ nil, 2 })
-    matches(flat_err, pcall_err(not_list_like.flatten, also_not_list_like))
+    local also_not_list_like = { nil, 2 }
+    eq({ 2 }, vim.iter(also_not_list_like):flatten():totable())
```
wookayin, line 499: A corner case:
```
vim.iter({1, { { { nil } } } }):flatten():totable()
```

gpanders: No I think that's right, the default depth is `1` so tables are not recursively flattened.

justinmk (approved):
Fantastic! 

Does this potentially (in the future) expand the capabilities of pipelines to work on array-like stuff? For reference, currently we have defined `:help list-iterator` and various pipeline operations say that they require that, in their docstrings.

gpanders:
> Does this potentially (in the future) expand the capabilities of pipelines to work on array-like stuff? For reference, currently we have defined `:help list-iterator` and various pipeline operations say that they require that, in their docstrings.

This enables the whole vim.iter API to work with "array like" tables (i.e. lists with holes). If everything eventually uses vim.iter under the hood we could probably even relax the distinction between "list" and "array" (since both would be treated the same by the vim.iter API). It's still an important distinction though because Lua itself treats the two differently.

runtime/lua/vim/iter.lua
```diff
@@ -1030,17 +1031,18 @@ function Iter.new(src, ...)
...
         return Iter.new(pairs(src))
       end
-      t[count] = v
+      if k > length then
+        length = k
+      end
+      t[k] = v
     end
-    return ListIter.new(t)
+    return ArrayIter.new(t, length)
```
justinmk, line 1045: IIUC, the `length` parameter is the main change. But it looks like we're also changing `t[k]` to use the (non-contiguous) number keys; will that cause any problems for `ArrayIter:next()`? 

Do we actually need the original number keys, or would it make sense to store `t` as a 1..n continguous "list" (as we did before)? Possibly I'm missing something, and maybe we need the number keys for `get()`-like functionality, or maybe we want to pass the number keys as `k` arg to the pipeline functions.

justinmk: > If we keep the non-contiguous keys in the new array, then we have the property that `vim.iter(t):totable() == t` (holes and all)

That seems desirable. OTOH, to preserve the non-contiguous keys the caller also has the option of passing `vim.iter(pairs(t))`, right?

gpanders: That's right. Another benefit of stripping holes is that any functions used in the iterator can behave consistently for both arrays and function iterators. A function iterator will never return `nil` (`nil` is the signal that the function iterator is done), so if we strip holes in arrays then we preserve that property.

```lua
vim.iter(s):filter(function(v)
  -- v can never be nil, regardless of the value of s
end)
```

I think it also makes the mental model more consistent. Right now if a pipeline stage returns `nil` (e.g. `:map()`) then that acts as removing an item from the iterator (future pipeline stages won't see that value). But if arrays were allowed to have nil values that would be inconsistent.

The more I think about it the more I do think that stripping `nil` is the right call, but we can always change it later if I'm wrong.

runtime/lua/vim/iter.lua
```diff
@@ -247,7 +249,7 @@ function ListIter:flatten(depth)
 
     -- exit early if we try to flatten a dict-like table
     if flattened == nil then
-      error('flatten() requires a list-like table')
+      error('flatten() requires an array-like table')
```
wookayin, line 252: Due to L135:
```
-- short-circuit: this is not a list like table
```

`flatten()` on lists with hole still might fail:

```lua
vim.iter({nil, {1, nil, 2}, 3}):flatten():totable()
-- ERROR: flatten() requires an array-like table
```

For the following, an expected output could be `{{1, nil, 2}, 3}` instead of an error according to the array-like semantics:
```
vim.iter({nil, {1, nil, 2}, 3}):flatten():totable()
-- ERROR: flatten() requires an array-like table
```

gpanders: Expected output of `vim.iter({nil, {1, nil, 2}, 3}):flatten():totable()` should be `{1, 2, 3}`.

`vim.iter({nil, {1, nil, 2}, 3}):totable()` (no flatten) returns `{{1, nil, 2}, 3}`. Calling flatten on that would result in `{1, 2, 3}`.

wookayin: Oh that's right. My mistake; I was actually thinking about `vim.iter(**{** {nil, {1, nil, 2}, 3} **}** ):flatten(1):totable()` (one more depth) == `{ {1, nil, 3}, 3 }`. So the current implementation after the update seems correct: no longer errors.

runtime/lua/vim/iter.lua
wookayin: As of now, the behavior is like the latter (i.e. `vim.iter(t)` strips holes, NOT `vim.iter(t):totable() == t`; e.g. `vim.iter({nil, 1, 2, nil, 3}):totable() == {1, 2, 3}`.

While I agree that what's proposed makes sense and has many benefits, I'm not 100% convinced whether this is the best choice, because: (1) `vim.iter(t):totable() == t` also sounds sensible; (2) what if one wants to keep `nil` for any reason (use `filter` explicitly to filter `nil`)?

lewis6991 (commented):
Opened an issue https://github.com/neovim/neovim/issues/28941 to discuss splitting out `ArrayIter`.

PR #28785, opened 2024-05-16
fix(version): fix vim.version().prerelease

### LINKED ISSUE #28782: `vim.version` not returning the correct version
### Problem

In the current release of Neovim v0.10 it looks like the version is incorrectly given by `vim.version()`. Could this be that a `dev` build was made accidentally rather than a production release, or if there is a bug in `vim.version()`. But it currently outputs as if it is a `dev` version.

### Steps to reproduce

nvim --clean
:lua =vim.version()
:lua =vim.version():__tostring()

### Expected behavior

It should return a version table that isn't set with `prerelease = "dev"` and the output of the version string shouldn't be `0.10.0-dev...`

### Neovim version (nvim -v)

NVIM v0.10.0 (github release)

n/a

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.33.1

xterm-kitty

### Installation

official GitHub release, https://github.com/neovim/neovim/releases/download/v0.10.0/nvim-linux64.tar.gz

bfredl:
fixes #28782 (when backported)

### DIFF
--- a/src/nvim/generators/gen_api_dispatch.lua
+++ b/src/nvim/generators/gen_api_dispatch.lua
@@ -260,7 +260,11 @@ fixdict(1 + #version)
 for _, item in ipairs(version) do
   -- NB: all items are mandatory. But any error will be less confusing
   -- with placeholder vim.NIL (than invalid mpack data)
-  put(item[1], item[2] or vim.NIL)
+  local val = item[2]
+  if val == nil then
+    val = vim.NIL
+  end
+  put(item[1], val)
 end
 put('build', version_build)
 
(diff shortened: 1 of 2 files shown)


justinmk (approved):

wookayin:
Thought not mandatory, would it be possible to detect whether CI is building a "release" and check `prerelease` in `version_spec` so that #28782 can be prevented in the future?

bfredl:
well, the release flow just doesn't run any tests at all.

dundargoc:
It might be possible to apply different git tags to the repo and test expected version output, that way it won't matter if a commit was "release or not".

So basically in a test run "git tag 1.0" and then check stuff works on release tags. Unsure if it's worth it tho.

PR #28790, opened 2024-05-16
vim-patch:9.1.{0414,0416}

luukvbaal:
**vim-patch:9.1.0414: Unable to leave long line with 'smoothscroll' and 'scrolloff'**

Problem:  Unable to leave long line with 'smoothscroll' and 'scrolloff'.
          Corrupted screen near the end of a long line with 'scrolloff'.
          (Ernie Rael, after 9.1.0280)
Solution: Only correct cursor in case scroll_cursor_bot() was not itself
          called to make the cursor visible. Avoid adjusting for
          'scrolloff' beyond the text line height (Luuk van Baal)

https://github.com/vim/vim/commit/b32055e504ebd4f6183a93b92b08d61dad61c841

**vim-patch:9.1.0416: some screen dump tests can be improved**

Problem:  some screen dump tests can be improved (after 9.1.0414)
Solution: Make sure screen state changes properly and is captured in the
          screen dumps (Luuk van Baal)

https://github.com/vim/vim/commit/2e642734f4be506483315b8881748a7ef45854f4

### DIFF
--- a/src/nvim/move.c
+++ b/src/nvim/move.c
@@ -1579,8 +1579,20 @@ void adjust_skipcol(void)
     redraw_later(curwin, UPD_NOT_VALID);
     return;  // don't scroll in the other direction now
   }
-  colnr_T col = curwin->w_virtcol - curwin->w_skipcol + scrolloff_cols;
   int row = 0;
+  colnr_T col = curwin->w_virtcol + scrolloff_cols;
+
+  // Avoid adjusting for 'scrolloff' beyond the text line height.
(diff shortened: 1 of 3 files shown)


luukvbaal:
Pending https://github.com/vim/vim/pull/14788. Also, for the FIXME's, instead of the blank screen Vim produces, Nvim keeps an (also incorrect) `skipcol == 0` screen state.

zeertzjq (approved):

PR #28795, opened 2024-05-16
fix: trigger LspDetach on buffer delete

atoerien:
Currently, the LspDetach autocmd is triggered when the attached client quits, or if the `vim.lsp.buf_detach_client` function is called, but not if the buffer is deleted.

I've basically just updated the `on_detach` callback to be like `buf_detach_client`. Would it perhaps be better to just call `buf_detach_client` directly? I'm also not sure if the `vim.diagnostic.reset` bit is necessary.

There should probably be a test for this, but I have no idea how any of that works... this PR is mostly because it was easier than writing up a bug report :) Please feel free to make/suggest any necessary changes.

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -576,10 +576,19 @@ local function buf_attach(bufnr)
     on_detach = function()
       local params = { textDocument = { uri = uri } }
       for _, client in ipairs(lsp.get_clients({ bufnr = bufnr })) do
+        api.nvim_exec_autocmds('LspDetach', {
+          buffer = bufnr,
+          modeline = false,
+          data = { client_id = client.id },
+        })
+
         changetracking.reset_buf(client, bufnr)
         if client.supports_method(ms.textDocument_didClose) then
           client.notify(ms.textDocument_didClose, params)
         end
+
+        local namespace = lsp.diagnostic.get_namespace(client.id)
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/lsp.lua
```diff
@@ -567,10 +567,20 @@ local function buf_attach(bufnr)
     on_detach = function()
       local params = { textDocument = { uri = uri } }
       for _, client in ipairs(lsp.get_clients({ bufnr = bufnr })) do
+        api.nvim_exec_autocmds('LspDetach', {
+          buffer = bufnr,
+          modeline = false,
+          data = { client_id = client.id },
+        })
```
wookayin, line 574: Q: what happens if the autocmd handler throws an error?

AThePeanut4: Are errors propagated through nvim_exec_autocmds?

AThePeanut4: Tested with a simple handler that just unconditionally errors, and it doesn't seem to break anything for a normal LSP. I say normal because it was breaking things with null-ls (none-ls), but the error was coming from the `client.notify` call, which null-ls monkey-patches afaik. Not sure if this is something that should be accounted for.

mfussenegger:
Thanks. I extended a test case to cover the LspDetach event.

PR #28799, opened 2024-05-17
fix(path): avoid chdir() when resolving path

### LINKED ISSUE #28786: Strange `$MYVIMRC` behavior if `~/.config/nvim` is a symlink and current directory does not exist
### Problem

Sure, the title sounds like the setup to a bad joke. But it's an issue I actually did encounter for real.

If you start nvim while inside a directory that has been deleted (for example by a /tmp-cleaning daemon), then it seems `$MYVIMRC` changes its logic for resolving symlinks. If init.lua itself is a symlink it still works fine, but if `~/.config/nvim` is, then that is not resolved into its real path. This can have adverse effects on configs that look for files relative to `$MYVIMRC`.

I only talk about `init.lua` here, but `init.vim` seems to have the same issue.

### Steps to reproduce

(Does not work with `-u`, needs to be using the default vimrc resolution.)

```sh
❯ mv ~/.config/nvim ~/nvim-config # move existing config out of the way
❯ mkdir ~/.config/nvim-real
❯ ln -s nvim-real ~/.config/nvim
❯ echo 'print(vim.env.MYVIMRC)' > ~/.config/nvim-real/init.lua
❯ mkdir /tmp/dir
❯ cd /tmp/dir
❯ nvim
# prints `~/.config/nvim-real/init.lua` at startup
❯ rmdir ../dir
❯ nvim
# prints `~/.config/nvim/init.lua`
```

### Expected behavior

I'd understand if things behave a bit strangely when executing in a directory that does not exist, but I'd expect vimrc resolution to be unaffected. Considering the behavior when init.vim itself is a symlink, and vim's behavior, the expected behavior is probably to show the file's real path in both cases.

### Neovim version (nvim -v)

NVIM v0.9.5

No, it prints `~/.config/vim-real/vimrc` in both cases. (Vim 9.1, compiled May 11 2024 19:59:14)

### Operating system/version

Arch Linux

### Terminal name/version

MATE Terminal 1.28.1

xterm-256color

### Installation

pacman (system package manager)

zeertzjq:
Use `uv_fs_realpath()` instead.

Fix #28786

It seems that `uv_fs_realpath()` has some problems on non-Linux platforms:
```rst
        This function has certain platform-specific caveats that were discovered when used in Node.

        * macOS and other BSDs: this function will fail with UV_ELOOP if more than 32 symlinks are
          found while resolving the given path.  This limit is hardcoded and cannot be sidestepped.
        * Windows: while this function works in the common case, there are a number of corner cases
          where it doesn't:

          - Paths in ramdisk volumes created by tools which sidestep the Volume Manager (such as ImDisk)
            cannot be resolved.
          - Inconsistent casing when using drive letters.
          - Resolved path bypasses subst'd drives.

        While this function can still be used, it's not recommended if scenarios such as the
        above need to be supported.

        The background story and some more details on these issues can be checked
        `here <https://github.com/nodejs/node/issues/7726>`_.
```
I don't know if the old implementation that uses `uv_chdir()` and `uv_cwd()` also suffers from the same problems.
- For the ELOOP case, `chdir()` seems to have the same limitations as `realpath()`.
- On Windows, Vim doesn't use anything like `chdir()` either. It uses `_wfullpath()`, while libuv uses `GetFinalPathNameByHandleW()`.

### DIFF
--- a/src/nvim/eval/funcs.c
+++ b/src/nvim/eval/funcs.c
@@ -6479,7 +6479,7 @@ static void f_resolve(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
   char *v = os_resolve_shortcut(fname);
   if (v == NULL) {
     if (os_is_reparse_point_include(fname)) {
-      v = os_realpath(fname, v);
+      v = os_realpath(fname, NULL, MAXPATHL + 1);
     }
   }
   rettv->vval.v_string = (v == NULL ? xstrdup(fname) : v);
@@ -6631,7 +6631,7 @@ static void f_resolve(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
     xfree(buf);
   }
 # else
-  char *v = os_realpath(fname, NULL);
+  char *v = os_realpath(fname, NULL, MAXPATHL + 1);
   rettv->vval.v_string = v == NULL ? xstrdup(fname) : v;
 # endif
 #endif
--- a/src/nvim/os/env.c
+++ b/src/nvim/os/env.c
@@ -474,17 +474,9 @@ void init_homedir(void)
     var = os_homedir();
   }
 
-  if (var != NULL) {
-    // Change to the directory and get the actual path.  This resolves
-    // links.  Don't do it when we can't return.
-    if (os_dirname(os_buf, MAXPATHL) == OK && os_chdir(os_buf) == 0) {
-      if (!os_chdir(var) && os_dirname(IObuff, IOSIZE) == OK) {
-        var = IObuff;
-      }
-      if (os_chdir(os_buf) != 0) {
-        emsg(_(e_prev_dir));
-      }
-    }
+  // Get the actual path.  This resolves links.
+  if (var != NULL && os_realpath(var, IObuff, IOSIZE) != NULL) {
+    var = IObuff;
   }
 
   // Fall back to current working directory if home is not found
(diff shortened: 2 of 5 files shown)


zeertzjq:
It seems that `uv_fs_realpath()` has some problems on non-Linux platforms:
```rst
        This function has certain platform-specific caveats that were discovered when used in Node.

        * macOS and other BSDs: this function will fail with UV_ELOOP if more than 32 symlinks are
          found while resolving the given path.  This limit is hardcoded and cannot be sidestepped.
        * Windows: while this function works in the common case, there are a number of corner cases
          where it doesn't:

          - Paths in ramdisk volumes created by tools which sidestep the Volume Manager (such as ImDisk)
            cannot be resolved.
          - Inconsistent casing when using drive letters.
          - Resolved path bypasses subst'd drives.

        While this function can still be used, it's not recommended if scenarios such as the
        above need to be supported.

        The background story and some more details on these issues can be checked
        `here <https://github.com/nodejs/node/issues/7726>`_.
```
I don't know if the old implementation that uses `uv_chdir()` and `uv_cwd()` also suffers from the same problems.

justinmk (approved):
Great solution. The os_chdir approach is [from vim](https://github.com/neovim/neovim/commit/5762c4e528cbd6629319bf3958d1f6554b399b20) but they must have changed their approach at some point. Maybe we can mark some patches as N/A or merged, after this.

zeertzjq:
I think Vim has always uses the `chdir()` approach on Unix, but not on Windows.

PR #26618, opened 2023-12-17
refactor: soft-deprecate diagnostic signs configured with :sign-define

gpanders:
Diagnostic signs should now be configured with vim.diagnostic.config(), but "legacy" sign definitions are still supported for two more release cycles to help minimize breaking changes.

Ref: https://github.com/neovim/neovim/pull/26193#issuecomment-1859163716

PR #28824, opened 2024-05-17
fix(diagnostic): show backtrace for deprecation warnings

wookayin:
Problem: On nvim 11.0-dev, deprecation warnings due to an use of hard-deprecated APIs such as:
- `vim.diagnostic.disable()`
- `vim.diagnostic.is_disabled()`

etc. are not accompanied by backtrace information.

It makes difficult for users to figure out which lines or which plugins are still using deprecated APIs.

Solution: use `backtrace = true` in vim.deprecate() call.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -1281,7 +1281,7 @@ M.handlers.signs = {
             'vim.diagnostic.config()',
             '0.12',
             nil,
-            false
+            false -- suppress backtrace
           )
 
           if not opts.signs.text then
@@ -1596,7 +1596,7 @@ end
 
 --- @deprecated use `vim.diagnostic.is_enabled()`
 function M.is_disabled(bufnr, namespace)
-  vim.deprecate('vim.diagnostic.is_disabled()', 'vim.diagnostic.is_enabled()', '0.12', nil, false)
+  vim.deprecate('vim.diagnostic.is_disabled()', 'vim.diagnostic.is_enabled()', '0.12')
   return not M.is_enabled { bufnr = bufnr or 0, ns_id = namespace }
 end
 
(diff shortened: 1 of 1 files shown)


wookayin:
Note: The argument `deprecate = false` was first introduced in #26618, and I don't see any reason to suppress backtrace here as well.

gpanders:
> EDIT: seems it was intentional? https://github.com/neovim/neovim/pull/26618/files#r1429248469

That particular instance was intentional, because the backtrace there doesn't provide any useful information. It's not a call site of a specific function, it's warning about the use of `sign_define` for diagnostic signs, which could be happening anywhere in the user's config. The backtrace in this case would just be misleading.

wookayin:
I see, thanks for the explanation. This might be too niche, but in my specific configs it was a plugin (`neo-tree`) rather than user config:

```
vim.diagnostic.is_disabled() is deprecated, use vim.diagnostic.is_enabled() instead. :help deprecated
Feature will be removed in Nvim 0.12
stack traceback:
        ...neovim/runtime/lua/vim/diagnostic.lua:1597: in function 'is_disabled'
        .../neo-tree.nvim/lua/neo-tree/utils/init.lua:244: in function 'get_diagnostic_counts'
        .../neo-tree.nvim/lua/neo-tree/setup/init.lua:45: in function <.../neo-tree.nvim/lua/neo-tree/setup/init.lua:44>
        [C]: in function 'pcall'
        .../neo-tree.nvim/lua/neo-tree/events/queue.lua:65: in function 'fire_event_internal'
        .../neo-tree.nvim/lua/neo-tree/events/queue.lua:100: in function <....neo-tree.nvim/lua/neo-tree/events/queue.lua:99>
        [C]: in function 'pcall'
        .../neo-tree.nvim/lua/neo-tree/utils/init.lua:126: in function 'debounce'
        .../neo-tree.nvim/lua/neo-tree/utils/init.lua:71: in function ''
        vim/_editor.lua: in function ''
```

So I think backtrace information could be still useful for this particular case as well as for user configs?

gpanders:
> So I think backtrace information could be still useful for this particular case as well as for user configs?

For `vim.diagnostic.disable()` and `vim.diagnostic.is_disabled()` I agree. I don't know why the backtrace was omitted from those.

wookayin:
> That particular instance was intentional, because the backtrace there doesn't provide any useful information. It's not a call site of a specific function, it's warning about the use of `sign_define` for diagnostic signs, which could be happening anywhere in the user's config. The backtrace in this case would just be misleading.

This makes sense -- backtrace should be omitted for this. I've updated the PR accordingly.

PR #28826, opened 2024-05-17
docs: better explain how core dump files can be obtained

wookayin:
Problem: Docs about how to obtain backtraces on Linux is not very
beginner-friendly; some users used to have difficulties in getting
stacktrace against Nvim crash.

For instance, the `core` dump file might not appear in the current
directory on Ubuntu systems with apport, and the current docs do not
fully cover such cases.

Solution: Add more hints about where core dump files can be found. For
example, on Ubuntu where apport is managing core dump files, users would
want to find them in `/var/lib/apport/coredump`.

Ref: https://stackoverflow.com/questions/2065912/core-dumped-but-core-file-is-not-in-the-current-directory (also in the vimdoc)

Also, polish the docs a bit more.

### DIFF
--- a/runtime/doc/dev_tools.txt
+++ b/runtime/doc/dev_tools.txt
@@ -14,10 +14,10 @@ itself. See also |debug.txt| for advice that applies to Vim.
 ==============================================================================
 Backtraces                                            *dev-tools-backtrace*
 
-LINUX ~
+LINUX
 
-Core dumps are disabled by default on Ubuntu
-https://stackoverflow.com/a/18368068, CentOS and others. To enable core dumps:
+Core dumps are disabled by default on Ubuntu, CentOS and others.
+To enable core dumps:
 >bash
     ulimit -c unlimited
 <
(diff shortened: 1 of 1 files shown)


runtime/doc/dev_tools.txt
```diff
@@ -86,19 +94,21 @@ Use `TEST_TAG` to run tests matching busted tags (of the form `#foo` e.g.
 Then, in another terminal:
 >bash
     gdb build/bin/nvim
-    target remote localhost:7777
+    (gdb) target remote localhost:7777
 
 -- See `nvim_argv` in https://github.com/neovim/neovim/blob/master/test/functional/testnvim.lua.
 
-USING LLDB TO STEP THROUGH UNIT TESTS ~
+
```
zeertzjq, line 101: Is this extra blank line intentional?

wookayin: No :) thanks, fixed.

zeertzjq: You deleted the wrong blank line :)

runtime/doc/dev_tools.txt
```diff
@@ -14,32 +14,40 @@ itself. See also |debug.txt| for advice that applies to Vim.
...
+To enable core dumps:
 >bash
     ulimit -c unlimited
 <
 On systemd-based systems getting a backtrace is as easy as:
 >bash
     coredumpctl -1 gdb
 <
-It's an optional tool, so you may need to install it:
+`corecumpctl` is an optional tool, so you may need to install it:
```
lucario387, line 28: typo :D

runtime/doc/dev_tools.txt
```diff
@@ -14,32 +14,40 @@ itself. See also |debug.txt| for advice that applies to Vim.
...
 >bash
     sudo apt install systemd-coredump
 <
 
-The full backtrace is most useful, send us the `bt.txt` file:
```
dundargoc, line 33: I think I like the old formulation more, there's less risk of misunderstanding. "Send us bt.txt" vs "send us backtrace information".

dundargoc (approved):

runtime/doc/dev_tools.txt
wookayin: I don't mind with either, so I switched back to the previous one. :)

PR #28835, opened 2024-05-18
docs(undo): fix spelling error

alphashuro:
Found this while reading through the docs


clason:
Thank you, but that is actually not a typo -- Vim's(!) documentation refers here to the [Nvi editor](https://en.wikipedia.org/wiki/Nvi).

PR #28836, opened 2024-05-18
refactor: split up `vim.deprecate` into two functions

dundargoc:
`vim.deprecate` has become too complicated, and the usecase for a
general deprecation function that plugins can use vs a deprecation
function that fits neovim's growing needs is becoming too different.


dundargoc:
I'll incorporate these changes directly to https://github.com/neovim/neovim/pull/28778, that way it'll be easier to see if this is needed.

PR #28837, opened 2024-05-18
docs: misc

dundargoc:


camoz:
Hope it's okay to hijack this / post suggestions:

In the first codebox in [`:help lua-guide-vim-functions`](https://neovim.io/doc/user/lua-guide.html#lua-guide-vim-functions), I think the first line is redundant with the last, no? Should the last be removed?

dundargoc:
@camoz I'm having trouble understanding exactly what you mean. Mind sending a PR with your suggested changes?

dundargoc:
TODO: yeet docs for overridable functions

src/man/nvim.1
```diff
@@ -387,10 +387,10 @@ features like
 .El
 .Sh FILES
 .Bl -tag -width "~/.config/nvim/init.vim"
```
zeertzjq, line 389: Does this line need change?

JohnVillalovos: It does not need to change as it is the same length.  As in "vim" is the same length/width as "lua"

https://linux.die.net/man/7/groff_mdoc  Documents the `-width` parameter.

But it would not hurt to change it and may make it less confusing to change it.

dundargoc:
TODO: nvim vs nvim vs nvim

runtime/lua/vim/ui.lua
```diff
@@ -29,9 +29,6 @@ local M = {}
 ---               individual item from `items`. Defaults to `tostring`.
 ---     - kind (string|nil)
 ---               Arbitrary hint string indicating the item shape.
----               Plugins reimplementing `vim.ui.select` may wish to
----               use this to infer the structure or semantics of
----               `items`, or the context in which select() was called.
```
justinmk, line 34: why remove this note?

clason: Because we no longer wish to promote "monkey-patching" API function as an official customization method.

justinmk: where was that discussed? isn't that the purpose of `vim.ui.*` (and was the reason `kind` was introduced)?

dundargoc: Pretty much. There was rough consensus in devchat that overloading functions was a bad design choice we shouldn't perpetuate further. I suggested we remove the few instances we document this in order to prevent it from spreading in that case. No one protested so I went ahead with it 🤷.

Want me to keep this?

justinmk: > There was rough consensus in devchat that overloading functions was a bad design choice

Bad choice compare to what? Let's write down the reasoning in an issue. Making these decisions silently in devchat leads to confusion.

clason: Bad choice compared to offering an API that allows registering callbacks (since that -- in contrast to monkey-patching -- allows validation).

justinmk: This is totally premature and it's really unsettling that these decisions are not written in an issue. Propose an interface in an issue, don't make these random decisions based on devchat.

dundargoc: You're almost making it sound intentionally malicious or something. I didn't think it warranted mentioning as majority of people seemed to seemed to dislike it and it seemed "obvious" at the time that it was a bad idea. I'll remove these changes from there then.

PR #28838, opened 2024-05-18
fix(lsp): add textDocument/documentLink to capability map

icholy:
The `supports_method` method returns `true` if it doesn't recognize the LSP method. As a result, the following request is sent to clients that don't support it:

``` lua
local params = { textDocument = util.make_text_document_params() }
vim.lsp.buf_request(0, "textDocument/documentLink", params, function(err, result, ctx)
    // ...
end)
```


MariaSolOs (approved):

MariaSolOs:
I think a more appropriate PR title/commit message would be `fix(lsp): add textDocument/documentLink to capability map`.

icholy:
updated

PR #28839, opened 2024-05-18
refactor!: remove `nvim` and `provider` module for checkhealth

dundargoc:
The namespacing for healthchecks for neovim modules is inconsistent and
confusing. The completion for `:checkhealth` with `--clean` gives

```
nvim
provider.clipboard
provider.node
provider.perl
provider.python
provider.ruby
vim.lsp
vim.treesitter
```

There are now three top-level module names for nvim: `nvim`, `provider`
and `vim` with no signs of stopping. The `nvim` name is especially
confusing as it does not contain all neovim checkhealths, which makes it
almost a decoy healthcheck.

The confusion only worsens if you add plugins to the mix:

```
lazy
mason
nvim
nvim-treesitter
provider.clipboard
provider.node
provider.perl
provider.python
provider.ruby
telescope
vim.lsp
vim.treesitter
```

Another problem with the current approach is that it's not easy to run
nvim-only healthchecks since they don't share the same namespace. The
current approach would be to run `:che nvim vim.* provider.*` and would
also require the user to know these are the neovim modules.

Instead, use this alternative structure:

```
vim.health
vim.lsp
vim.provider.clipboard
vim.provider.node
vim.provider.perl
vim.provider.python
vim.provider.ruby
vim.treesitter
```

and

```
lazy
mason
nvim-treesitter
telescope
vim.health
vim.lsp
vim.provider.clipboard
vim.provider.node
vim.provider.perl
vim.provider.python
vim.provider.ruby
vim.treesitter
```

Now, the entries are properly sorted and running nvim-only healthchecks
requires running only `:che vim.*`.


dundargoc:
Future improvements:

- Reduce the five provider checks into a single checks. Triple nested checkhealths is overkill in granularity.
- `:checkhealth vim` should run all healthchecks under the `vim` module IMO. Right now it does nothing and requires it to be run as a glob with `:checkhealth vim.*`.
- (Optional) Maybe remove the globbing functionality entirely and focus solely on module based healthchecks (meaning `:che vi*` would not work). I suspect this functionality isn't well known tbh. On the other hand it's not expensive, but on the other other hand with module-based approach and improved completion we wouldn't need to rely on globbing.

dundargoc:
Tbh does anyone care about this?

justinmk:
"nvim" was for non-provider checks, especially configuration-related checks. Is `:che vim.health` now the way to do that? The provider checks can be slow.

dundargoc:
> "nvim" was for non-provider checks, especially configuration-related checks.

I know, but neovim is the only thing that has multiple top-level checkhealth modules. Remember, there was no real way to run neovim only healthchecks before.

> Is `:che vim.health` now the way to do that?

Ye, because `runtime/vim/health.vim` which provides `vim.health` is blocking the `:checkhealth vim` space. Not much that can be done. I'm up for suggestions if you have something else in mind. Something like `:checkhealth vim.config`?

justinmk:
As long as there's a way to run non-provider Nvim checks, I think it's fine. Not worried about the name too much, in this case.

PR #28841, opened 2024-05-18
fix(treesitter): `get_node_text()` newline handling consistency

### LINKED ISSUE #28677: `get_node_text()` behaves differently between buffer and string sources on EOL char
### Problem

If a treesitter node range's `end_col` is `0`, `vim.treesitter.get_node_text()` will not return the trailing new line char for buffer source.

### Steps to reproduce

test.lua:
```lua
local ts = vim.treesitter

local query = ts.query.parse("python", "(string_content) @s")

local function dbg(root, source)
  for _, nodes in query:iter_matches(root, source, 0, -1) do
    local node = nodes[#nodes]
    print("Range =", vim.inspect({ node:range() }))
    print("Text = |" .. ts.get_node_text(node, source) .. "|")
  end
end

local source = [[
"""test
"""
]]

print("====== string")
local root = ts.get_string_parser(source, "python"):parse()[1]:root()
dbg(root, source)

print("====== buffer")
local buf = vim.api.nvim_create_buf(true, true)
vim.api.nvim_buf_set_lines(buf, 0, -1, true, vim.split(source, "\n", { plain = true }))
root = ts.get_parser(buf, "python"):parse()[1]:root()
dbg(root, buf)
```

```console
> nvim -l test.lua 
====== string
Range = { 0, 3, 1, 0 }
Text = |test
|
====== buffer
Range = { 0, 3, 1, 0 }
Text = |test|
```

Node ranges are same, but the result of get_node_text() is different.

### Expected behavior

`get_node_text()` on buffers returns the trailing new line char.

### Neovim version (nvim -v)

NVIM v0.9.5 & NVIM v0.10.0-dev-d3fa88b

no

### Operating system/version

Linux

### Terminal name/version

kitty 0.34.1

xterm-kitty

### Installation

nixpkgs

dmehala:
##  Description
This pull request addresses the issue where `get_node_text()` behaves inconsistently between buffer and string sources.

Resolves #28677 

### Changes
- Modified `buf_range_get_text` to insert the trailing newline char.

### Rationale
My understanding is taht the range represents the part of the source to be captured. When set on a grid:
```
  0 1 2 3 4 5 6 7
0 " " " t e s t \n
1 " " "
```

With the range being `Range = { 0, 3, 1, 0 }`. I would then expect `\n` to be part of the output. Hence, I modified the behavior of the buffer source.

`nvim_buf_get_text` replaces `\n` with `\0`, and they are artificially recreated with `table.concat(lines, "\n")`. This function does not add the newline if there's only one input or if the last input is a newline. I added logic to remember if the part of the buffer to release ends with a newline to add it later.

### Note to reviewers
Please feel free to correct me if my understanding is incorrect. I would be happy to discuss it further.

### DIFF
--- a/runtime/lua/vim/treesitter.lua
+++ b/runtime/lua/vim/treesitter.lua
@@ -187,6 +187,7 @@ end
 ---@returns string
 local function buf_range_get_text(buf, range)
   local start_row, start_col, end_row, end_col = M._range.unpack4(range)
+  local insert_nl = false
   if end_col == 0 then
     if start_row == end_row then
       start_col = -1
@@ -194,8 +195,12 @@ local function buf_range_get_text(buf, range)
     end
     end_col = -1
     end_row = end_row - 1
+    insert_nl = true
   end
   local lines = api.nvim_buf_get_text(buf, start_row, start_col, end_row, end_col, {})
+  if insert_nl then
+    table.insert(lines, "")
+  end
   return table.concat(lines, '\n')
 end
 


wookayin:
I think the fix makes sense, but please review failing tests and fix accordingly.

cshuaimin:
I don't understand what the `if end_col == 0` condition is doing. Seems it's the same logic as in [`ts_utils.get_vim_range()`](https://github.com/nvim-treesitter/nvim-treesitter/blob/14db0433e145c8ea63efc42ddbdf517d6036d372/lua/nvim-treesitter/ts_utils.lua#L239-L266) function? Can anyone explain it to me?

runtime/lua/vim/treesitter.lua
```diff
@@ -187,15 +187,20 @@ end
...
+  local insert_nl = false
   if end_col == 0 then
     if start_row == end_row then
       start_col = -1
       start_row = start_row - 1
     end
     end_col = -1
     end_row = end_row - 1
+    insert_nl = true
   end
```
wookayin, line 199: It looks like we could also get rid of this `if end_col == 0 then ... end` block completely to achieve the same behavior, and use directly return what `nvim_buf_get_get_text()` returns.

That said, one might wonder why `get_node_text` was designed in the current way. It was introduced 58bbc2ea0b3dfed13471e8cc0447d7598be24276 --- not sure why. Maybe to match the behavior as the initial implementation 1f3c0593eb1d4e54ce1edf35da67d184807a9280 where only rows are specified.

dmehala: I ack the message. I will have a deeper look this week. Thanks for the review.

runtime/lua/vim/treesitter.lua
```diff
@@ -187,15 +187,20 @@ end
...
   local lines = api.nvim_buf_get_text(buf, start_row, start_col, end_row, end_col, {})
+  if insert_nl then
+    table.insert(lines, "")
```
Hamza12700, line 202: The double quotes are failing the linter test. Make it single quotes.

https://github.com/neovim/neovim/actions/runs/9142308887/job/25137718542?pr=28841

image

kristijanhusak:
I found another issue with `get_node_text` when using a string source that could probably be fixed as part of this PR.

When providing a custom range through `metadata`, it is not respected for string source, only for buffer source. 

For example, with this source:
```lua
print('test')
```

Opening it up as a buffer and doing this:

```
:lua= vim.treesitter.get_node_text(vim.treesitter.get_node():root(), 0, { metadata = { range = { 0, 0, 0, 10 } }})
```

Returns `print('tes` as expected.

Doing it with a string source:
```
:lua= vim.treesitter.get_node_text(vim.treesitter.get_node():root(), "print('test')", { metadata = { range = { 0, 0, 0, 10 } }})
```

Returns the whole line: `print('test')`

This patch would fix the issue:

```diff
diff --git a/runtime/lua/vim/treesitter.lua b/runtime/lua/vim/treesitter.lua
index 4629203138..ae0840f78c 100644
--- a/runtime/lua/vim/treesitter.lua
+++ b/runtime/lua/vim/treesitter.lua
@@ -213,13 +213,16 @@ function M.get_node_text(node, source, opts)
 
   if metadata.text then
     return metadata.text
-  elseif type(source) == 'number' then
-    local range = M.get_range(node, source, metadata)
+  end
+
+  local range = M.get_range(node, source, metadata)
+
+  if type(source) == 'number' then
     return buf_range_get_text(source, range)
   end
 
   ---@cast source string
-  return source:sub(select(3, node:start()) + 1, select(3, node:end_()))
+  return source:sub(range[3] + 1, range[6])
 end
 
 --- Determines whether (line, col) position is in node range

```

runtime/lua/vim/treesitter.lua
ribru17: I think this check prevents `Index out of bounds` errors when the node as at the end of the file. I just encountered this myself trying to get text using only `nvim_buf_get_text` without any special magic. So it looks like the check is necessary

justinmk:
@dmehala People spent time reviewing this. Why are you closing it? Add a test and fix the lint issues.

dmehala:
That's an interesting perspective @justinmk. I didn't mean to offend anyone or suggest their time isn't valuable—my apologies. Now that I finally have time to contribute, I'll work on fixing the remaining issues and adding tests.

justinmk:
still unaddressed, so closing and there are 2 other PRs addressing this.

@kristijanhusak pr #39455 #39409 has a patch in that area

PR #28852, opened 2024-05-19
feat(comment): commentstring hook for more control

### LINKED ISSUE #28830: Inbuilt comment feature not working properly for JSX files
### Problem

when i type 'gcc; this is how it gets commented in a JSX file
image

### Steps to reproduce

image

### Expected behavior

this is how it should be for JSX when commented.
image

### Neovim version (nvim -v)

0.10

no

### Operating system/version

ubuntu 22.04

### Terminal name/version

xterm-256color

### Installation

appimage

ribru17:
Problem: it's difficult for users to customize the inbuilt commenting behavior (changing or extending)

Solution: expose a hook that can override the selected commentstring, to allow for finer control

Closes #28830

### DIFF
--- a/runtime/doc/various.txt
+++ b/runtime/doc/various.txt
@@ -585,6 +585,28 @@ to look up the value of 'commentstring' corresponding to the cursor position.
 (This can be different from the buffer's 'commentstring' in case of
 |treesitter-language-injections|.)
 
+The comment module can take a custom hook to allow for more control when
+determining the 'commentstring'. For example, the following can be used to get
+more accurate comments in JSX-family files (assuming the appropriate
+|treesitter| parser has been installed): >lua
+
+    require('vim._comment').cs_hook = function(row, col)
+      local ft = vim.bo.filetype
+      if ft == 'typescriptreact' or ft == 'javascriptreact' then
+        local current_node = vim.treesitter.get_node({ pos = { row, col } })
+        while current_node do
+          if current_node:type() == 'jsx_element' then
+            return '{/*%s*/}'
+          end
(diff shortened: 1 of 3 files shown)


clason:
The limitations were deliberate, for the sake of maintenance and documentation. There are enough plugins out there for people who want more control.

(Not shutting this down, but there's a high bar to clear for adding functionality.)

ribru17:
Ah, damn. That's fair; I'll wait for someone else to close if they choose to

runtime/doc/various.txt
```diff
@@ -585,6 +585,28 @@ to look up the value of 'commentstring' corresponding to the cursor position.
 (This can be different from the buffer's 'commentstring' in case of
 |treesitter-language-injections|.)
 
+The comment module can take a custom hook to allow for more control when
+determining the 'commentstring'. For example, the following can be used to get
+more accurate comments in JSX-family files (assuming the appropriate
+|treesitter| parser has been installed): >lua
+
+    require('vim._comment').cs_hook = function(row, col)
```
justinmk, line 593: "hook" is not a name used anywhere else in vim/nvim. Use existing names/conventions, to help discoverability.

This would typically be modeled as an "event" (autocmd) or a callback (`on_foo`).

justinmk:
https://github.com/neovim/neovim/issues/28830#issuecomment-2119690661 gives a more likely alternative. Adding random hooks on top of the already numerous extension points is too much. 'commentstring' should be fixed instead. This PR doesn't explain why 'commentstring' isn't sufficient.

Issue #24802, opened 2023-08-20
inccommand=split does not work with mulitgrid


fredizzimo:
### Problem

For more details see https://github.com/neovide/neovide/issues/1947

When using `inccommand=split` together with Neovide, the window is blank This is caused by a new window being created and then being destroyed before it has a chance to show. This repeats for each character typed. See the following log from Neovide

```
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(WindowPosition { grid: 2, start_row: 0, start_column: 0, width: 104, height: 38 })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(GridLine { grid: 2, row: 0, column_start: 1, cells: [GridLineCell { text: "a", highlight_id: Some(10), repeat: None }] })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(GridLine { grid: 2, row: 1, column_start: 1, cells: [GridLineCell { text: "a", highlight_id: Some(10), repeat: None }] })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(GridLine { grid: 2, row: 2, column_start: 1, cells: [GridLineCell { text: "a", highlight_id: Some(10), repeat: None }] })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(GridLine { grid: 2, row: 3, column_start: 1, cells: [GridLineCell { text: "a", highlight_id: Some(10), repeat: None }] })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(GridLine { grid: 2, row: 4, column_start: 1, cells: [GridLineCell { text: "a", highlight_id: Some(10), repeat: None }] })
TRACE [neovide::channel_utils] neovide::editor::EditorCommand NeovimRedrawEvent(Resize { grid: 10, width: 104, height: 7 })
[...]

Issue #20463, opened 2022-10-03
redraw doesn't update the UI during substitute


folke:
When using `vim.ui_attach` with `ext_cmdline`, `:redraw` doesn't update the ui during `substitute`

folke:
The following hack seems to work:

```lua
  if vim.api.nvim_get_mode().mode == "c" and vim.fn.getcmdline():find("s/") then
    -- HACK: this will trigger redraw during substitue
    vim.api.nvim_input("<space><bs>")
  end
```

zeertzjq:
Hmm, this seem to be intentional from #9783

folke:
The `cmdline_show` events also do come in **before** the redraw lock, but it seems the `cmdline_pos` events come in **after** the redraw lock.

Maybe it can be fixes by also sending the `cmdline_pos` events before the redraw lock?

folke:
So, that's why my hack works. I simply get a new `cmdline_show` event

folke:
I was able to fix this for Noice by calling `update_screen()` with **ffi**

```lua
  local ffi = require("noice.util.ffi")
  if ffi.cmdpreview then
    -- HACK: this will trigger redraw during substitute and cmdpreview,
    -- but when moving the cursor, the screen will be cleared until
    -- a new character is entered
    ffi.update_screen()
  end
```

AndreM222:
has there been any news about this issue?

PR #9783, opened 2019-03-25
inccommand: disable :redraw, events

justinmk:
closes #9777

Issue #9777, opened 2019-03-23
inccommand: redraw from timer clears preview window and substitutions


gelguy:
- `nvim --version`: `v0.4.0-401-g5c836d2ef`
- Operating system/version: `Windows 10 WSL`
- Terminal name/version: `wsltty`
- `$TERM`: `xterm`

### Steps to reproduce using `nvim -u NORC`

```
nvim -u NORC

set inccommand=split

autocmd CmdlineChanged * call timer_start(1000, {-> execute('redraw')})
```

Open a buffer with text and type `%s/.`
### Actual behaviour
Preview window and substitutions disappear after a while.

### Expected behaviour
Preview window and substitutions do not disappear.

Also happens for `inccommand=nosplit`, with `redrawstatus` and when `redraw` is called from async actions.

Use case: Async action finishes and calls `redraw`/`redrawstatus` when `inccommand` preview is active.

justinmk:
Forcing a redraw from a timer is undefined and makes no sense.

justinmk:
I'm not sure why we even allow timers to fire during 'inccommand'. I think that should be prevented.

bfredl:
For the "redraw" part of the issue, an alternative could be to keep the preview window around longer. Right now we split, redraw, and close the window immediately on every valid keypress, so we rely on stale screen state to be kept to be displayed to the user. The alternative would be to not close the preview window until the user either makes the cmdline invalid (not a substitution) or exiting cmdline mode. 

`nosplit` mode would be more tricky though, one option could be to disable redraw for that one window for the same duration, unless the window was moved/resized or something. Though "eventually" maybe the preview can be re-implemented as virtual text, at least for changes that doesn't change line splits.

justinmk:
> an alternative could be to keep the preview window around longer. 

+1, unfinished todo from https://github.com/neovim/neovim/pull/5561 .

PR #28856, opened 2024-05-20
feat(inccommand): allow redrawing during cmdpreview

### LINKED ISSUE #28510: redraw a specific floating window ignoring other pending screen updates
### Problem

There is no way to redraw and update a specific window without updating other windows that "need redrawing". Individual windows can be marked for redraw, but updates are only ever flushed to the UI via `update_screen()` which will go through *all* windows with pending updates. _In most cases_ this is either necessary (e.g. for splits, resizing, etc) or it's not a problem.

However, there is much work being done in PR #27855 (tracking issue #27811) to externalize messages and the command line for the TUI via floating windows (also [noice.nvim](https://github.com/folke/noice.nvim)). The current high-level redrawing logic for the command line/message grid is orthogonal to that of windows, which has the following advantages:
1. Visual selection is not cleared when redrawing the command line. For `:'<,'>[command]...` this is great because you can see what the ex-cmd is (presumably) operating on. Any call to `:redraw`, opening a new float, etc. will clear the visual selection highlight.
2. no redrawing clash with `inccommand`. command line can be redrawn independently, without being blocked ([#9777](https://github.com/neovim/neovim/issues/9777), [#9783](https://github.com/neovim/neovim/pull/9783), [#20463](https://github.com/neovim/neovim/issues/20463)) or clearing the preview.

Neither can be achieved via regular window redrawing, so if the cmdline/msg grid is externalized with a floating window we lose both of these benefits (I think #27950 somewhat addresses point (2) but still wouldn't allow properly re-updating the cmdline afterwards). Afaik this also applies to the native popup wildmenu which has its own seperate redrawing codepath.

In brief, externalized ui component providers in the TUI (specifically `ext_cmdline` and `ext_messages`, also likely `ext_popumenu` for `wildmenu`) have to use hacks and workarounds to maintain ephemeral screen state (like visual selection in cmdline, or cmdpreview) because there is no publicly exposed mechanism to redraw a window without screwing up the dirty screen state (which should be kept dirty). Some PRs have addressed this problem but in an ad-hoc/hacky manner.

### Expected behavior

Either:
1. a function in the api/stdlib that allows redrawing a specific float without touching any other pending screen updates. I think(hope?) its reasonable to expect that the (re)drawing of ***floating*** windows can be done in isolation of other windows/the "main" window grid. Somewhat related: #28101 but in its current state/scope it does not fix the issue
2. internal logic for keeping/restoring desirable ephemeral screen states: specifically the visual selection and command preview. I suppose the redrawing logic could check and handle each of theses cases individually, e.g. *not* removing the visual highlight, re-triggering the cmdpreview callback if needed, etc. There may be other similar "dirty screen states" that we want to keep though, I just can't think of any others.

theofabilous:
**Problem**: The current `inccommand` implementation works by forcing a redraw right after applying changes and immediately restoring everything without a subsequent redraw, thereby relying on ephemeral screen state (the "preview" visible to the user does not represent the actual state of buffers/windows). This requires disabling redraws while command preview is active (#9777, #9783), which suppresses unrelated UI updates (#20463, #27950), complicates/prevents redrawing floating windows (#28510), and causes rendering issues for GUIs (#24802). In cases where a redraw is forced (e.g. editor resize), the command preview is cleared. Similarly, `<C-g>` and `<C-t>` during `:substitute` clear the preview.

**Solution**: Remove reliance on temporary screen state in command preview, actually keep the buffer changes after applying them. Only restore state when needed (e.g. before re-updating the preview, when leaving cmdline, ...). Allow redrawing during preview, and refresh command preview when the editor is resized or a scroll occurs.

---

> Helpful background: #28510. Notably, it highlights why refreshing command preview is required when scrolls/resizes occur (even when the preview state is persistent). In brief, commands may choose to apply previews only within the visible viewport, so when more lines are revealed the preview may need to be executed again. (If this weren't the case, the editor would basically brick on each keypress after `:%substitue/` for a sufficiently large file).

<details><summary><b>NOTE</b> about commit history</summary>

The commit history is kind of wonky, I tried to organize changes in a way that would make the implementation changes easier to follow (incrementally/chronologically) at the expense of a clean/logical commit sequence in its own right. When this PR gets closer to being ready and others have taken a look at it, I'd like to squash and rearrange some of these changes.

</details>

---

# Implementation

> **NOTE**: In this section, I often use the term "preview callback" or something similar, by which I mean the "per-command" previewing function. So for user-commands thats the `preview = ...` field in the `opts` arg of `nvim_create_user_command` (`:help :command-preview`), and for builtin functions that's `CommandDefinition.cmd_preview_func` (currently only exists for `:substitute`).

The changes in this PR can be roughly organized in 4 steps/categories, which loosely follow the commits chronologically:

1. Refactor the existing cmdpreview logic to ease the new implementation, without yet changing observable behaviour (i.e. still use the `apply + redraw + teardown/restore + disable redrawing` approach). The original approach only ever executes the command preview routines entirely and atomically, all in one go. The new approach requires executing different parts of this routine independently. Additionally, the preview state now needs to be kept alive across calls to update functions and may also be called externally (like, not within the "regular" command line event handling functions). This requires hoisting the state object to a global.

<details>

- split `cmdpreview_prepare()` in two: `cmdpreview_save()` and `cmdpreview_prepare()`. `save()` handles saving buffer information (undo, changedtick, ...), while `prepare()` sets things up only needed for the duration of the next preview execution (ensures undolevels are such that we can subsequently rollback changes, temporarily disable cursorline and incsearch, ...). Reason: later, we'll often want to only call `prepare()` without saving again.
- split `cmdpreview_restore_state()` in two, moving out memory deallocations to `cmdpreview_destroy()`. We'll eventually want to keep saved information around, but we still need to restore buffers before re-executing preview changes, so need to split restoration and deallocation.
- Use a global `CpInfo` object instead of one local to `cmdpreview_may_show()`. Move some of the local variables in that function to fields in `CpInfo` if we'll eventually need to keep track of them.
- extract part of the initial logic from `cmdpreview_may_show` to `cmdpreview_may_init`/`cmdpreview_check_cmdline`, which handle checking if the command is previewable and, if applicable, preparation for showing preview.
- reorganize where/when autocommands are blocked and messages are disabled. Before this wasn't a concern because all the subroutines where only ever called by `cmdpreview_may_show()`, so that function could simply block autocmds at the beginning and unblock them at the end. Eventually, we'll be calling some of those functions individually but still want to block events and error reporting.

</details>

At this point, the high level logic is the same: `cmdpreview_may_show()` handles the full init->show->destroy cycle at once.

2. Don't immediately tear down preview state. Now, `cmdpreview_may_show()` no longer unconditionally completes the full preview cycle. On preview success, the buffer changes are no longer reverted immediately. The `CpInfo` object is not destroyed and thus still holds the saved buffer information. On the next call, the buffer information doesn't need to be saved: instead, the buffers are just restored using the existing saved information before re-applying the preview changes.

<details>

- add `cmdpreview_transition_begin`/`_end` functions. These handle maintaining invariants that I found hard to keep track of manually. When entering the preview handling functions, we may or may not be already previewing, and when exiting the cmdpreview handling functions, we may or may not need to restore buffers, free memory, and/or close windows (for example, depending on whether the preview callback succeeded). The transition functions handle a lot these things.
- we previously didn't always call `cmdpreview_may_show` when the command line changes, this no longer works because the preview may still be active; skipping the cmdpreview updates would then keep those changes active. So, if the cmdline was updated but cmdpreview shouldn't be shown (e.g. when not interactive or input chars are available), call `cmdpreview_did_not_show` which handles cleanup, if applicable.

</details>

3. Handle buffer and window changes/invalidations between preview updates. Now that the changes are kept around, re-executing preview routines when preview is already active isn't as simple: buffers and windows may have been closed, so some of the saved buffer/window information may no longer be valid, and there may be new windows/buffers that now need to be saved.

<details>

- Use `bufref_T` instead of bufs for saved buffer information, so that we can check if the information refers to a invalid buffer. Free invalid buffer information without restoring them in these cases
- Store saved information in hashmaps instead of vecs: when preparing for updates, we want to go through open windows/buffers and check if we've already saved them. See commit message of 7d54e8b
    - **NOTE**: this is possibly less efficient in general unless you have a lot of buffers/windows open, but I'm going with naive/"obvious" solutions for the time being

</details>

4. Finally, allow redrawing while command preview is active. Also update cmdpreview when a resize or scroll occurs, if needed (see note above and/or #28510 for why the latter is required)

---

# Considerations/open questions

Now that preview changes are persistent, a lot more bookkeeping is required. Specifically, buffer and window information can arbitrarily change in between preview updates, and I'm not sure how to handle all such changes.
I have a few specific concerns/questions that I'll add in the comments when I get around to properly wording them.

Here's a rough, quick, likely non-exhaustive list of these concerns:

- What exactly is a preview function allowed to do? I think this should be documented, both for users' and devs' sake.

- In the spirit of playing nice with external UIs, I think there needs to be a way to mark windows as ignored by command preview (if there isn't already), or maybe just have a class of window/buffer option combinations that should be ignored by command preview (e.g. `nomodifiable`+`bufhidden=hide`+`buftype=nofile`+is a floating window+...)?
    - For now, nothing is ignored, but to allow UI widget modifications while cmdpreview is active, the following best effort "heuristic" approach is used (see https://github.com/neovim/neovim/pull/28856/commits/d5ea89cf300fdcd13bc0868cc7202ca168731241):
        - after the preview callback, mark all unchanged buffers (reason: we may "assume" that buffers not modified by the preview callback are more likely to be UI widget-ish)
        - on the next preview update, if a marked buffer has changed, don't attempt to restore it (reason: ui widgets may have been modified in the meantime, restoring them makes no sense and would break e.g. externalized cmdlines during preview)

- What/how do we handle the "in between preview updates" part? Preview changes are actually persistent now, and things can potentially be "messed up" in between consecutive calls to a preview callback (e.g. autocommands called after cmdline changed right before cmdpreview is executed, or a timer callback, ...). Specifically:
    - what if a buffer was changed during this time (possibly *after* being modified by cmdpreview, but *before* those changes were reverted)?

### DIFF
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -2351,7 +2351,7 @@ void nvim__redraw(Dict(redraw) *opts, Error *err)
 
   // Redraw pending screen updates when explicitly requested or when determined
   // that it is necessary to properly draw other requested components.
-  if (opts->flush && !cmdpreview) {
+  if (opts->flush) {
     update_screen();
   }
 
--- a/src/nvim/buffer.c
+++ b/src/nvim/buffer.c
@@ -461,6 +461,11 @@ bool buf_valid(buf_T *buf)
 static bool can_unload_buffer(buf_T *buf)
 {
   bool can_unload = !buf->b_locked;
+#ifdef EXITFREE
+  can_unload &= buf->b_cp_locked <= BCP_HALF_LOCKED || (exiting && entered_free_all_mem);
+#else
+  can_unload &= buf->b_cp_locked <= BCP_HALF_LOCKED;
+#endif
 
   if (can_unload && updating_screen) {
     FOR_ALL_WINDOWS_IN_TAB(wp, curtab) {
@@ -480,7 +485,7 @@ static bool can_unload_buffer(buf_T *buf)
 
 bool buf_locked(buf_T *buf)
 {
-  return buf->b_locked || buf->b_locked_split;
+  return buf->b_locked || buf->b_locked_split || buf->b_cp_locked > BCP_NONE;
 }
 
 /// Close the link to a buffer.
@@ -1302,6 +1307,14 @@ static int do_buffer_ext(int action, int start, int dir, int count, int flags)
       && buf != curbuf
       && !check_can_set_curbuf_forceit((flags & DOBUF_FORCEIT) ? true : false)) {
     // disallow navigating to another buffer when 'winfixbuf' is applied
+    return FAIL;
+  }
+
+  if (buf->b_cp_locked > BCP_NONE
+      && (action == DOBUF_SPLIT
+          || (action == DOBUF_GOTO && buf != curbuf && !is_aucmd_win(curwin)))) {
+    // disallow navigating to a buffer locked by cmdpreview
+    emsg(_(e_nav_cplocked));
     return FAIL;
   }
 
--- a/src/nvim/buffer_defs.h
+++ b/src/nvim/buffer_defs.h
@@ -70,6 +70,16 @@ typedef struct {
 // Mask to check for flags that prevent normal writing
 #define BF_WRITE_MASK   (BF_NOTEDITED + BF_NEW + BF_READERR)
 
+// values for b_cp_locked
+#define BCP_NONE        0       // Not locked by command preview.
+#define BCP_HALF_LOCKED 1       // Locked, but not currently in use
+                                // by command preview. Like BCP_FULL_LOCKED,
+                                // but can still be unloaded.
+#define BCP_FULL_LOCKED 2       // Locked and currently in use. Cannot be
+                                // unloaded or entered, and windows containing
+                                // it cannot be closed, split, or change
+                                // buffers.
+
 typedef struct wininfo_S WinInfo;
 typedef struct frame_S frame_T;
 typedef uint64_t disptick_T;  // display tick type
@@ -367,6 +377,9 @@ struct file_buffer {
                                 // a new window with it.
   int b_ro_locked;              // Non-zero when the buffer can't be changed.
                                 // Used for FileChangedRO
+  int b_cp_locked;              // Buffer locked by cmdpreview, disallowing
+                                // closing it and closing/splitting windows
+                                // that contain it
 
   // b_ffname   has the full path of the file (NULL for no name).
   // b_sfname   is the name as the user typed it (or NULL).
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -398,6 +398,9 @@ void screen_resize(int width, int height)
         }
       }
     }
+    if (cmdpreview_may_refresh()) {
+      update_screen();
+    }
     ui_flush();
   }
   resizing_screen = false;
--- a/src/nvim/errors.h
+++ b/src/nvim/errors.h
@@ -149,6 +149,8 @@ EXTERN const char e_using_number_as_bool_nr[] INIT(= N_("E1023: Using a Number a
 EXTERN const char e_not_callable_type_str[] INIT(= N_("E1085: Not a callable type: %s"));
 EXTERN const char e_auabort[] INIT(= N_("E855: Autocommands caused command to abort"));
 
+EXTERN const char e_nav_cplocked[] INIT(= N_("E1161: Can't navigate to buffer locked by 'inccomand'"));
+
 EXTERN const char e_api_error[] INIT(= N_("E5555: API call: %s"));
 
 EXTERN const char e_fast_api_disabled[] INIT(= N_("E5560: %s must not be called in a fast event context"));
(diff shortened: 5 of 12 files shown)


luukvbaal (commented):
Thanks for working on this! It seems quite strange to me to cram this into the redraw routines. If there is a need to refresh the inccommand window(s) dynamically, perhaps it makes sense to add a dedicated API for that?

EDIT: Oops, meant to add this to the change in `drawscreen.c`. Ah I remember now the changed viewport thing(scroll/resize), is this mainly for that? Could you add a docstring to `cmdpreview_may_refresh()` that explains why it is needed?

justinmk:
> Now that preview changes are persistent, a lot more bookkeeping is required.

Yeah, 500 lines of new C code isn't a feasible way to approach this. Think about leveraging UI events or at least shelling out to Lua.

Feel free to rebase and reopen if a lower-cost approach can be found.

Issue #28868, opened 2024-05-20
With ext_cmdline activated the buffer used for incsearch and command-preview is not the last buffer before entering cmdline_mode


smilhey:
### Problem

When using activating ext_cmdline, if I put the cursor in a float window during a substitution command, command-previews uses the buffer in the float instead of the current buffer. You see the effect when setting the enter field in nvim_open_win to true and then entering :%s/ ...

### Steps to reproduce

```lua
local function on_cmdline_show()
	local buf = vim.api.nvim_create_buf(false, true)
	local win = vim.api.nvim_open_win(buf, false, {
		relative = "editor",
		width = 50,
		height = 10,
		row = 1,
		col = 10,
		style = "minimal",
	})
	vim.api.nvim__redraw({ flush = true, cursor = true })
end

local function handler(event, ...)
	if event == "cmdline_show" then
		on_cmdline_show()
	else
		return
	end
end

local namespace = vim.api.nvim_create_namespace("cmdline")
local function attach()
	vim.ui_attach(namespace, { ext_cmdline = true }, function(event, ...)
		handler(event, ...)
		if event:match("cmd") ~= nil then
			return true
		end
		return false
	end)
end

attach()
```

### Expected behavior

I would expect the current buffer to be the last buffer before entering cmdline mode. 

### Neovim version (nvim -v)

0.11.0-dev-19+g0f4f7d32c

### Vim (not Nvim) behaves the same?

No

### Operating system/version

Fedora Linux 40

### Terminal name/version

kitty 0.34.1

### $TERM environment variable

xterm-256color

### Installation

build from repo

luukvbaal:
Why are are you providing a minimal init.lua and then giving another condition in the problem description. Is there a problem with this init.lua or is there just a problem when setting `enter == true`?

The point of `nvim__redraw->cursor == true` is that it allows moving the cursor to a window that is **not** the actual current window. Entering the cmdline window seems ill advised and changing the current window/buffer should probably be disallowed during cmdline callbacks.

smilhey:
Just when setting enter == true. 
That part was not clear to me in the doc but i guess it explains the distinction that you make "win or the current window". Changing to this does not draw the cursor in the float. Is that expected ? 

```lua
local function on_cmdline_show()
	local buf = vim.api.nvim_create_buf(false, true)
	local win = vim.api.nvim_open_win(buf, false, {
		relative = "editor",
		width = 50,
		height = 10,
		row = 1,
		col = 10,
[...]

PR #28873, opened 2024-05-20
fix(api): alloc and draw cursor window in nvim__redraw

luukvbaal:
Problem:  Unable to move cursor to recently opened window.
Solution: Make sure uninitialized window is drawn before trying to move
          the cursor to it.

Ref #28868

### DIFF
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -2393,10 +2393,6 @@ void nvim__redraw(Dict(redraw) *opts, Error *err)
     redraw_buf_range_later(rbuf, first, last);
   }
 
-  if (opts->cursor) {
-    setcursor_mayforce(win ? win : curwin, true);
-  }
-
   bool flush = opts->flush;
   if (opts->tabline) {
     // Flush later in case tabline was just hidden or shown for the first time.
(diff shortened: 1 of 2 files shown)


test/functional/api/vim_spec.lua
```diff
@@ -5020,12 +5020,29 @@ describe('API', function()
...
+      width = 1,
+      height = 1,
+      row = 1,
+      col = 10,
+    })
+    api.nvim__redraw({ win = newwin, cursor = true })
+    screen:expect({
+      grid = [[
+                                                                    |
+        {1:~         ^                                                  }|
```
luukvbaal, line 5041: It's probably more user friendly/expected to just draw the window in this case, so flush implicitly.

src/nvim/api/vim.c
```diff
@@ -2422,11 +2418,22 @@ void nvim__redraw(Dict(redraw) *opts, Error *err)
     }
   }
 
-  // Flush pending screen updates if "flush" or "clear" is true, or when
-  // redrawing a status component may have changed the grid dimensions.
+  win_T *cwin = win ? win : curwin;
+  // Allow moving cursor to recently opened window and make sure it is drawn(#28868).
```
zeertzjq, line 2422:
```suggestion
  // Allow moving cursor to recently opened window and make sure it is drawn #28868.
```

test/functional/api/vim_spec.lua
```diff
@@ -5020,12 +5020,29 @@ describe('API', function()
   it('nvim__redraw', function()
     local screen = Screen.new(60, 5)
     screen:attach()
-    local win = api.nvim_get_current_win()
     eq('at least one action required', pcall_err(api.nvim__redraw, {}))
     eq('at least one action required', pcall_err(api.nvim__redraw, { buf = 0 }))
     eq('at least one action required', pcall_err(api.nvim__redraw, { win = 0 }))
     eq("cannot use both 'buf' and 'win'", pcall_err(api.nvim__redraw, { buf = 0, win = 0 }))
+    local win = api.nvim_get_current_win()
+    -- Can move cursor to recently opened window and window is flushed(#28868)
```
zeertzjq, line 5028:
```suggestion
    -- Can move cursor to recently opened window and window is flushed #28868
```

clason:
@luukvbaal rebase?

PR #28897, opened 2024-05-21
refactor(lsp): use supports_method where applicable

icholy:


icholy:
That failure is a flake

MariaSolOs (approved):

PR #28914, opened 2024-05-22
fix(lsp): check if buffer was detached in on_init callback

### LINKED ISSUE #28575: LSP: _changetracking.lua:154: Invalid buffer id: (race condition in 'initialize' RPC?)
## Problem
When using a system to automatically attach buffers to a language server that reinitializes itself, or responds to the `initialize` RPC query slowly, or something similar, if one of those buffers becomes invalid then the `_on_attach` method of LSP Clients produces an invalid buffer-number error. 

This problem is compounded if for one reason or another the initialize query is repeatedly responded to or multiple buffers attached to a client become invalid over time (for example, if significant numbers of ephemeral preview buffers are attached to a language server and then removed in short succession), as all of these invalid buffers are provided to `_on_attach` on what I assume are repeated `initialize` responses from the LSP server, every time, creating a large error log spam which greatly interferes with typing >.<

While this is most acute when using a system that happens to attach and invalidate ephemeral buffers, it is (again, as far as I can tell) a race condition that should apply in any case where an LSP server reinitializes itself after any of it's attached buffers are rendered invalid. However, in practice it seems extremely difficult to reproduce....

```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/_changetracking.lua:154: Invalid buffer id: 44                                                                     
stack traceback:                                                                                                                                                                                  
        [C]: in function 'nvim_buf_get_name'                                                                                                                                                      
        /usr/share/nvim/runtime/lua/vim/lsp/_changetracking.lua:154: in function 'init'                                                                                                           
        /usr/share/nvim/runtime/lua/vim/lsp/client.lua:915: in function '_text_document_did_open_handler'                                                                                         
        /usr/share/nvim/runtime/lua/vim/lsp/client.lua:950: in function '_on_attach'                                                                                                              
        /usr/share/nvim/runtime/lua/vim/lsp/client.lua:623: in function ''                                                                                                                        
```

## Originating Code Path
`runtime/lua/vim/lsp/client.lua:608` - this is inside a callback passed to `rpc.request` with method `initialize`
```lua
    -- If server is being restarted, make sure to re-attach to any previously attached buffers.
    -- Save which buffers before on_init in case new buffers are attached.
    local reattach_bufs = vim.deepcopy(self.attached_buffers) -- <=== This is the source of the invalid bufnrs 

    self:_run_callbacks(self._on_init_cbs, lsp.client_errors.ON_INIT_CALLBACK_ERROR, self, result)

    for buf in pairs(reattach_bufs) do
      self:_on_attach(buf) -- <==== this is the source of the error(s) when the bufnr is invalid, which we will traverse into
    end
```

`runtime/lua/vim/lsp/client.lua:938` - this is the actual `_on_attach` method. The error comes from inside `_text_document_did_open_handler` method
```lua
--- @package
--- Runs the on_attach function from the client's config if it was defined.
--- @param bufnr integer Buffer number
function Client:_on_attach(bufnr)
  self:_text_document_did_open_handler(bufnr) -- <=== Source of error. 

  lsp._set_defaults(self, bufnr)

  api.nvim_exec_autocmds('LspAttach', {
    buffer = bufnr,
    modeline = false,
    data = { client_id = self.id },
  })
```

`runtime/lua/vim/lsp/client.lua:902`
```lua
--- @package
--- Default handler for the 'textDocument/didOpen' LSP notification.
---
--- @param bufnr integer Number of the buffer, or 0 for current
function Client:_text_document_did_open_handler(bufnr)
  changetracking.init(self, bufnr)  -- <=== Error comes from here. Note that further down there is some kind of bufnr validation
  if not vim.tbl_get(self.server_capabilities, 'textDocumentSync', 'openClose') then
    return
  end
  if not api.nvim_buf_is_loaded(bufnr) then
    return
  end
  local filetype = vim.bo[bufnr].filetype
```

`runtime/lua/vim/lsp/_changetracking.lua:149`, inside the `M.init` module function. 
```lua
  local buf_state = state.buffers[bufnr]
  if buf_state then
    buf_state.refs = buf_state.refs + 1
  else
    buf_state = {
      name = api.nvim_buf_get_name(bufnr), -- <=== The real source of the error ^.^
      lines = {},
      lines_tmp = {},
      pending_changes = {},
      needs_flush = false,
      refs = 1,
    }
    state.buffers[bufnr] = buf_state
```

I could not get a minimal reproduction, unfortunately. I spent many many hours trying. However, minimal reproductions of race conditions are rather difficult. In particular, I've failed to get a minimal config to get an invalid bufnr into the `attached_buffers` of any LSP server, even though I know it is possible because my custom LSP setup (which is a bit of an abomination at this point...) has a status command which I modified to report on that (among other things like listing files with each language server client, filtering by language, etc.)

## My final failed attempt at `minimal_init.lua`
```lua
local pattern = 'lua'
local cmd = { 'lua-language-server' }

-- LSP that we want to directly attach to
local client_id = nil

-- Files to open
--
-- First is the one put in a buffer to immediately wipe it out, second is to induce a re-initialize
local bufhidewipe_file = "dummy-lua.lua"
local reinit_trigger_file = "dummy-lua-two.lua"

-- Start the LSP with the bufnr, using the filename as the location...
--
-- Bufnr may not actually be a standard buffer constructed from something like `:edit` ^.^
local function do_start(bufnr, fname)
	local match = vim.fs.find(root_markers, { path = fname, upward = true })[1]
	return vim.lsp.start({
		name = 'lua-language-server',
		cmd = cmd,
		root_dir = root_dir,
		settings = settings,	
	}, { bufnr = bufnr })
end

--[[
	pattern = pattern,
	callback = function(args)
		client_id = do_start(0, args.file)
	end
})]]--



-- Open a file and heuristically locate it's buffer number
--
-- Note that this will totally break with duplicate files in different current working directories :p
-- It's just intended to be a very quick method of getting buffer numbers to work with :)
--
-- Optionally, specify `preview = true` to open in the preview window....
local function open_buffer_via_name(filename, preview)
	-- Open file in split.
	if not preview then
		vim.cmd.vsplit(filename)
	else
		vim.cmd.pedit(filename)
	end
	-- We need to find the bufnr for this buffer... can't find a way to actually directly create/edit a buffer in a way that gets the bufnr >.<
	local test_bufnr = nil
	for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
		local name = vim.api.nvim_buf_get_name(bufnr)
		-- Plain search
		if string.find(name, filename, 1, true) ~= nil then
			test_bufnr = bufnr
		end
	end
	return test_bufnr
end

-- Load a file into a scratch buffer. Will still have the filename as the name. Returns the bufnr
-- also allows you to set a filetype. Also reads contents from the file in question into the buffer.
--
-- This sets the scratch buffer's `bufhidden` option to `wipe`, which will wipe the buffer when it's not in a window
-- Importantly, this doesn't trigger autocommands! :p (i think this might be related).
--
-- It also creates a window (to prevent immediate wipe).
--
-- Returns the scratch buffer id as well as the scratch window id. The important thing is to destroy the buffer by hiding
-- the window and not deleting it directly ^.^

-- is in fzf-lua, which suggests this may be related to lsp??
local function open_scratch_buffer_via_name(filename, set_filetype)
	-- Read data into the buffer.
	--
	-- We read before creating the buffer so if there are errors no extranuous buffers
	-- are left hanging around :p
	local filelines = {}
	for l in io.lines(filename) do
		-- one-based indices 🤢
		filelines[#filelines + 1] = l
	end
	local scratch_buffer = vim.api.nvim_create_buf(false, true)
	vim.api.nvim_buf_set_lines(scratch_buffer, 0, 1, false, filelines)
	vim.api.nvim_buf_set_name(scratch_buffer, filename)
	-- This is probably the important bit. FZF sets `bufhidden=wipe` which does NOT CALL AUTOCMDS
	vim.api.nvim_set_option_value("bufhidden", "wipe", { buf = scratch_buffer })
	vim.api.nvim_set_option_value("filetype", set_filetype, { buf = scratch_buffer })

	-- Create the window. Doesn't need to be float or anything :p
	-- Automatically enter it.
	local scratch_window = vim.api.nvim_open_win(scratch_buffer, true, {
		relative = "cursor",
		row = 0,
		col = 0,
		height = 40,
		width = 30,
		noautocmd = true,
		border = "rounded"
	})
	return scratch_buffer, scratch_window
end


vim.api.nvim_create_user_command("ReproduceRaceCondition", function(args) 
	local bufhidewipe_bufnr, bufhidewipe_win_id = open_scratch_buffer_via_name(bufhidewipe_file, "lua")
	local ephemeral_client_id = do_start(bufhidewipe_bufnr, bufhidewipe_file)
	vim.defer_fn(function()
		vim.api.nvim_win_close(bufhidewipe_win_id, true)
		--[[vim.defer_fn(function() 
			local client = vim.lsp.get_client_by_id(ephemeral_client_id)
			client:stop()
		end)]]

		vim.defer_fn(function()
			local nonephemeral_bufnr = open_buffer_via_name(bufhidewipe_file, false)
			vim.lsp.buf_attach_client(nonephemeral_bufnr, ephemeral_client_id)
		end, 100)
	end, 50)
end, {
	nargs = 0
})
```
## Folder Arrangement
Just placing `minimal_init.lua` in a folder with an empty `.editorfile`, and `dummy-lua.lua` (the `.editorfile` is for lsp sharing) will get the configuration i was using while writing this failed minimal config.

## Actual Setup Behaviour
It's probably worth noting that the actual setup I have is more complex than this in certain ways that may be relevant. Importantly, the main LSP servers that I see in my listing (`:LspStatus`) command with invalid bufnrs is `marksman` (the markdown LSP I use). 

My suspicion is that somehow the autocmd I have that automatically produces a popup when typing (which is constantly being refreshed) manages to trigger this race condition. But it also seems to occur more after I use preview windows with `fzf-lua` if they are scrolled through very rapidly (I use no plugins for the popup/documentation aspect, so it's not an fzf-lua thing). However, this is not `marksman` stuff, even if I havent seen it in any kind of `:LspStatus` listing yet :/

## Logs
I have looked in logs, but there doesn't seem to be much useful there (though there are a signifcant number of errors from `marksman` and a few from my `lua-language-server` in some of the other earlier loglines. Still:
```
[START][2024-04-30 00:08:27] LSP logging initiated
[ERROR][2024-04-30 00:08:27] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:08:26 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:26:00] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:26:00 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:26:36] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:26:35 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:27:09] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:27:09 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:27:17] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:27:17 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:28:13] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:28:13 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:52] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:51 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:52] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:52 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:54] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:54 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:55] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:54 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:55] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:54 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:55] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:54 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:57] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:56 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:35:58] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:35:57 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:24] ...m/lsp/client.lua:980	"LSP[marksman]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:40] ...m/lsp/client.lua:980	"LSP[rust-analyzer]"	"on_error"	{  code = "ON_EXIT_CALLBACK_ERROR",  err = "vim/_editor.lua:0: E5560: Vimscript function must not be called in a lua loop callback"}
[ERROR][2024-04-30 00:36:59] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:36:59 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-04-30 00:36:59] .../vim/lsp/rpc.lua:752	"rpc"	"/usr/bin/marksman"	"stderr"	"[00:36:59 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
```
(I don't know what exactly is causing the loop callback error and it seems to sometimes not appear... I think it is not related since it seems to only occur after the first epoch reload I do, whereas the actual issue I'm reporting occurs all of the time - and yes, I have an epoch system for reloading my config... it's about 10x as horrifying as it sounds especially because I was rushing at the time :p)

## `:LspStatus` (my version) - probably not useful, but putting it here anyway in case it is potentially useful in figuring out the underlying causes
```
Client ID [1] did not have a client!                                                                                                                                                              
marksman [ID: 112]                                                                                                                                                                                
 Status:                                                                                                                                                                                          
  - Running                                                                                                                                                                                       
 Root Directory (config):                                                                                                                                                                         
  - (none)                                                                                                                                                                                        
 Workspace Directories (initial config):                                                                                                                                                          
  - (none)                                                                                                                                                                                        
 Workspace Directories (live):                                                                                                                                                                    
  - (none)                                                                                                                                                                                        
 Attached Buffers                                                                                                                                                                                 
  - [228] [INVALID BUFFER]                                                                                                                                                                        
rust-analyzer [ID: 111]                                                                                                                                                                           
 Status:                                                                                                                                                                                          
  - Running                                                                                                                                                                                       
 Root Directory (config):                                                                                                                                                                         
  - [anon]/sapient/self/limg                                                                                                                                                           
 Workspace Directories (initial config):                                                                                                                                                          
  - (none)                                                                                                                                                                                        
 Workspace Directories (live):                                                                                                                                                                    
  - file://[anon]/sapient/self/limg                                                                                                                                                    
 Attached Buffers                                                                                                                                                                                 
  - [1] [anon]/sapient/self/limg/src/lib.rs                                                                                                                                            
  - [27] [anon]/sapient/self/limg/src/bin/limg.rs                                                                                                                                      
  - [118] [anon]/sapient/self/limg/src/semantic_region.rs (current)                                                                                                                    
  - [72] [anon]/sapient/self/limg/src/locator.rs                                                                                                                                       
marksman [ID: 113]                                                                                                                                                                                
 Status:                                                                                                                                                                                          
  - Running                                                                                                                                                                                       
 Root Directory (config):                                                                                                                                                                         
  - (none)                                                                                                                                                                                        
 Workspace Directories (initial config):                                                                                                                                                          
  - (none)                                                                                                                                                                                        
 Workspace Directories (live):                                                                                                                                                                    
  - (none)                                                                                                                                                                                        
 Attached Buffers                                                                                                                                                                                 
  - [229] [INVALID BUFFER]                                                                                                                                                                        
```

### Expected behavior

Handle invalid buffers gracefully in whatever way is most conceptually in line with how the LSP infrastructure is meant to work - I don't know the best place to put the fix, even if it is extremely simple (just a check with `vim.api.nvim_buf_is_valid` ^.^). It may also not be worth doing as it seems to be quite hard to produce without some pretty weird config stuff, although it might just be that I didn't put the language servers under enough pressure in my many attempts at `minimal_init.lua`? 

It might be related to the todo in `runtime/vim/lsp.lua:70` relating to `resolve_bufnr` on scratch buffers, but I really just don't know 

From a practical perspective, fixing this issue could probably occur in two locations - the `rpc.request` call, and deep in the `_changetracker` init method - and it would be a very simple check. 

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2995+ga1550dbf0a (but applies in all versions with relevant code)

lua-language-server (3.8.3-1), marksman (2023_12_09-1), rust-analyzer (1.79.0-nightly), presumably others

### Operating system/version

Arch Linux

### Log file

_No response_

icholy:
Fixes: https://github.com/neovim/neovim/issues/28575

### DIFF
--- a/test/functional/plugin/lsp/testutil.lua
+++ b/test/functional/plugin/lsp/testutil.lua
@@ -39,8 +39,7 @@ M.create_server_definition = [[
         })
         local handler = handlers[method]
         if handler then
-          local response, err = handler(method, params)
-          callback(err, response)
+          handler(method, params, callback)
         elseif method == 'initialize' then
           callback(nil, {
             capabilities = opts.capabilities or {}
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -493,6 +493,35 @@ describe('LSP', function()
       eq(0, result.num_attached_after)
     end)
 
+    it('should not re-attach buffer if it was deleted in on_init #28575', function()
+      clear()
+      exec_lua(create_server_definition)
+      exec_lua([[
+        local server = _create_server({
+          handlers = {
+            initialize = function(method, params, callback)
+              vim.schedule(function()
+                callback(nil, { capabilities = {} })
+              end)
+            end
+          }
+        })
+        local bufnr = vim.api.nvim_create_buf(false, true)
+        local on_init_called = false
+        local client_id = vim.lsp.start({
+          name = 'detach-dummy',
+          cmd = server.cmd,
+          on_init = function()
+            vim.api.nvim_buf_delete(bufnr, {})
+            on_init_called = true
+          end
+        })
+        vim.lsp.buf_attach_client(bufnr, client_id)
+        local ok = vim.wait(1000, function() return on_init_called end)
+        assert(ok, "on_init was not called")
+      ]])
+    end)
+
     it('client should return settings via workspace/configuration handler', function()
       local expected_handlers = {
         { NIL, {}, { method = 'shutdown', client_id = 1 } },
@@ -662,7 +691,7 @@ describe('LSP', function()
             }
           },
           handlers = {
-            ['textDocument/willSaveWaitUntil'] = function()
+            ['textDocument/willSaveWaitUntil'] = function(_, _, callback)
               local text_edit = {
                 range = {
                   start = { line = 0, character = 0 },
@@ -670,7 +699,7 @@ describe('LSP', function()
                 },
                 newText = 'Hello'
               }
-              return { text_edit, }
+              callback(nil, { text_edit, })
             end
           },
         })
@@ -4136,8 +4165,8 @@ describe('LSP', function()
             }
           },
           handlers = {
-            ["textDocument/codeAction"] = function()
-              return {
+            ["textDocument/codeAction"] = function(_, _, callback)
+              callback(nil, {
                 {
                   title = "Code Action 1",
                   command = {
@@ -4145,10 +4174,10 @@ describe('LSP', function()
                     command = "command:1",
                   }
                 }
-              }
+              })
             end,
-            ["codeAction/resolve"] = function()
-              return nil, "resolve failed"
+            ["codeAction/resolve"] = function(_, _, callback)
+              callback("resolve failed", nil)
             end,
           }
         })
@@ -4336,7 +4365,7 @@ describe('LSP', function()
               },
             },
             handlers = {
-              ["textDocument/codeLens"] = function(method, params)
+              ["textDocument/codeLens"] = function(method, params, callback)
                 local lenses = {
                   {
                     range = {
@@ -4349,7 +4378,7 @@ describe('LSP', function()
                     },
                   },
                 }
-                return lenses
+                callback(nil, lenses)
               end,
             }
           })
@@ -4657,13 +4686,13 @@ describe('LSP', function()
           },
           handlers = {
             ---@return lsp.Location[]
-            ['textDocument/definition'] = function()
-              return { _G.mock_locations[1] }
+            ['textDocument/definition'] = function(_, _, callback)
+              callback(nil, { _G.mock_locations[1] })
             end,
             ---@return lsp.WorkspaceSymbol[]
-            ['workspace/symbol'] = function(_, request)
+            ['workspace/symbol'] = function(_, request, callback)
               assert(request.query == 'foobar')
-              return {
+              callback(nil, {
                 {
                   name = 'foobar',
                   kind = 13, ---@type lsp.SymbolKind
@@ -4674,7 +4703,7 @@ describe('LSP', function()
                   kind = 12, ---@type lsp.SymbolKind
                   location = _G.mock_locations[2],
                 }
-              }
+              })
             end,
           },
         })
--- a/runtime/lua/vim/lsp/client.lua
+++ b/runtime/lua/vim/lsp/client.lua
@@ -612,7 +612,10 @@ function Client:initialize()
     self:_run_callbacks(self._on_init_cbs, lsp.client_errors.ON_INIT_CALLBACK_ERROR, self, result)
 
     for buf in pairs(reattach_bufs) do
-      self:_on_attach(buf)
+      -- The buffer may have been detached in the on_init callback.
+      if self.attached_buffers[buf] then
+        self:_on_attach(buf)
+      end
     end
 
     log.info(
--- a/test/functional/plugin/lsp/inlay_hint_spec.lua
+++ b/test/functional/plugin/lsp/inlay_hint_spec.lua
@@ -70,8 +70,8 @@ before_each(function()
         inlayHintProvider = true,
       },
       handlers = {
-        ['textDocument/inlayHint'] = function()
-          return vim.json.decode(response)
+        ['textDocument/inlayHint'] = function(_, _, callback)
+          callback(nil, vim.json.decode(response))
         end,
       }
     })
@@ -106,8 +106,8 @@ describe('vim.lsp.inlay_hint', function()
           inlayHintProvider = true,
         },
         handlers = {
-          ['textDocument/inlayHint'] = function()
-            return {}
+          ['textDocument/inlayHint'] = function(_, _, callback)
+            callback(nil, {})
           end,
         }
       })
@@ -206,8 +206,8 @@ describe('vim.lsp.inlay_hint', function()
             inlayHintProvider = true,
           },
           handlers = {
-            ['textDocument/inlayHint'] = function()
-              return { expected2 }
+            ['textDocument/inlayHint'] = function(_, _, callback)
+              callback(nil, { expected2 })
             end,
           }
         })
--- a/test/functional/plugin/lsp/semantic_tokens_spec.lua
+++ b/test/functional/plugin/lsp/semantic_tokens_spec.lua
@@ -95,11 +95,11 @@ describe('semantic token highlighting', function()
             },
           },
           handlers = {
-            ['textDocument/semanticTokens/full'] = function()
-              return vim.fn.json_decode(response)
+            ['textDocument/semanticTokens/full'] = function(_, _, callback)
+              callback(nil, vim.fn.json_decode(response))
             end,
-            ['textDocument/semanticTokens/full/delta'] = function()
-              return vim.fn.json_decode(edit_response)
+            ['textDocument/semanticTokens/full/delta'] = function(_, _, callback)
+              callback(nil, vim.fn.json_decode(edit_response))
             end,
           }
         })
@@ -560,11 +560,11 @@ describe('semantic token highlighting', function()
             },
           },
           handlers = {
-            ['textDocument/semanticTokens/full'] = function()
-              return nil
+            ['textDocument/semanticTokens/full'] = function(_, _, callback)
+              callback(nil, nil)
             end,
             ['textDocument/semanticTokens/full/delta'] = function()
-              return nil
+              callback(nil, nil)
             end,
           }
         })
@@ -608,11 +608,11 @@ describe('semantic token highlighting', function()
             },
           },
           handlers = {
-            ['textDocument/semanticTokens/full'] = function()
-              return vim.fn.json_decode(response)
+            ['textDocument/semanticTokens/full'] = function(_, _, callback)
+              callback(nil, vim.fn.json_decode(response))
             end,
-            ['textDocument/semanticTokens/full/delta'] = function()
-              return vim.fn.json_decode(edit_response)
+            ['textDocument/semanticTokens/full/delta'] = function(_, _, callback)
+              callback(nil, vim.fn.json_decode(edit_response))
             end,
           }
         })
@@ -1075,8 +1075,8 @@ b = "as"]],
               },
             },
             handlers = {
-              ['textDocument/semanticTokens/full'] = function()
-                return vim.fn.json_decode(resp)
+              ['textDocument/semanticTokens/full'] = function(_, _, callback)
+                callback(nil, vim.fn.json_decode(resp))
               end,
             }
           })
@@ -1461,11 +1461,11 @@ int main()
               },
             },
             handlers = {
-              ['textDocument/semanticTokens/full'] = function()
-                return vim.fn.json_decode(resp1)
+              ['textDocument/semanticTokens/full'] = function(_, _, callback)
+                callback(nil, vim.fn.json_decode(resp1))
               end,
-              ['textDocument/semanticTokens/full/delta'] = function()
-                return vim.fn.json_decode(resp2)
+              ['textDocument/semanticTokens/full/delta'] = function(_, _, callback)
+                callback(nil, vim.fn.json_decode(resp2))
               end,
             }
           })


mfussenegger:
Looks good to me. Any chance you could add a test? With `_create_server`  in the lsp_spec I think it should be possible to simulate the scenario.

icholy:
@mfussenegger done

test/functional/plugin/lsp_spec.lua
```diff
@@ -493,6 +493,31 @@ describe('LSP', function()
       eq(0, result.num_attached_after)
     end)
 
+    it('should not re-attach buffer if it was deleted in on_init', function()
```
wookayin, line 496:
```suggestion
    it('should not re-attach buffer if it was deleted in on_init #28575', function()
```

However, I find the test does not fail without the patch on `vim.lsp.client`.

wookayin, line 496: I find the test does not fail without the patch on `vim.lsp.client`.

icholy: I doubled checked this and it's failing without the client change for me.

wookayin: I see. I also double checked, in my setup it does *not* fail. Is the execution order non-deterministic by any chance?

Actually I don't find the logic of this test and the anticipated order of execution quite clear, we probably meant buffer was gone during `on_init` and expect fail somewhere (but where exactly? some comments will help for elaboration) It looks like `on_init` can be called asynchronously, not necessarily *before* re-attaching. It'd be also a good idea to ensure a certain execution order by assertions.

icholy: The test assumes that the `on_init` callback will called twice:

1. Before the `vim.lsp.start` returns (`bufnr` should be `nil`)
2. After `client:initialise()` is called. (`bufnr` should be a valid buffer).

wookayin: Thanks, the tests look cleaner after the simplification. A bit off topic, but I was wondering: in which case and why `client:initialize` can be called twice? (it's not a public method, so not supposed to be called by user) `client:initialize` would have been called inside `vim.lsp.start()`. `on_init` is meant to be called only once.

While I think this is a good (though contrived) repro for the bug #28575, what's more impacting is the race condition when the client responds too late, isn't it? Can we possibly come up with a more realistic scenario where #28575 happens?

icholy: @wookayin `:LspRestart`

wookayin: I meant: calling `client:initialize()` twice in a real-world, not for the tests. If we want to simulate a delay we will have to write a similar code so that the asynchronous callback can be scheduled andexecute, rather than simply calling it from the main thread.

icholy: I'm not sure what you're suggestion. Do you have an idea for a better way to test this?

mfussenegger: > I meant: calling client:initialize() twice in a real-world, not for the tests.

I understood the real world scenario as:

- Client starts
- rpc.request("initialize") fires
- Client attaches one or more other buffers
- Server responds to initialize
- on_init fires and deletes a buffer


To mirror that scenario more closely in the test it would be necessary to override the `srv.request` of the `_create_server` and defer the `initialize` response until a second buffer is attached. But I think the current solution is also good enough.

icholy: I've updated the test, let me know what you think.

icholy: > or anytime until LSP initialization is done

I don't think this is correct. You can delete the buffer in that window and it should be removed from the `client.attached_buffers` table. See:

https://github.com/neovim/neovim/blob/f79e906442b77cfa4e5a02d51f5dcb2fb186afbc/runtime/lua/vim/lsp.lua#L588
https://github.com/neovim/neovim/blob/f79e906442b77cfa4e5a02d51f5dcb2fb186afbc/runtime/lua/vim/lsp.lua#L678

test/functional/plugin/lsp/testutil.lua
```diff
@@ -39,8 +39,12 @@ M.create_server_definition = [[
         })
         local handler = handlers[method]
         if handler then
-          local response, err = handler(method, params)
-          callback(err, response)
+          if debug.getinfo(handler, 'u').nparams == 3 then
```
icholy, line 42: Seems like this isn't working correctly in puc-lua.

icholy: I updated all the handlers to be async.

wookayin: Not using such a hacky solution sounds great. But changing all the handlers to be async might be too much. I haven't tested this much in my environments so it might not be accurate yet, but how about something like:

```diff
        local handler = handlers[method]
+       local delay = (opts.delay or {})[method] ---@type integer? (ms)
+       if delay then -- simulate 'slow', asynchronous response from LSP
+         local original_callback = callback
+         callback = function(err, response)
+           vim.defer_fn(function(),
+             original_callback(err, response)
+           end, delay)
+         end
+       end

        if handler then
          local response, err = handler(method, params)
          callback(err, response)
```

so one can write like and avoid touching other handlers:
```lua
        local server = _create_server {
          delay = {
            ['initialize'] = 100, -- simulate LSP server is responding 100ms late
          }
        }
```

test/functional/plugin/lsp_spec.lua
wookayin: Question: I get you are saying removal of a buffer will invalidate `attached_buffers[bufnr]` (either synchronously or asynchronously, not so sure). But then, to clarify, what's so special about buffer gets wiped out *during* `on_init`? Is this a realistic scenario? I was thinking that the use of `on_init` is an artificial scenario you chose to reproduce the bug, and the race condition could happen as buffer will be removed somewhere else than `on_init`.

icholy: I'm not sure _why_ someone would be deleting buffers in `on_init`, but based on my read-through of the code, that's the most obvious way the `Invalid buffer id: 44` error from #28575 could happen. It's definitely weird, but it's still a valid usage of the API and therefore should work. This might not end up being the root cause of #28575 but it's still worth fixing if only to narrow the search. The lack of a good repro doesn't leave us with a lot of good options.

icholy:
@mfussenegger this is ready for review again.

test/functional/plugin/lsp_spec.lua
```diff
@@ -493,6 +493,34 @@ describe('LSP', function()
...
+        local client_id = vim.lsp.start({
+          name = 'detach-dummy',
+          cmd = server.cmd,
+          on_init = function()
+            vim.api.nvim_buf_delete(bufnr, {})
+            on_init_called = true
+          end
+        })
+        vim.lsp.buf_attach_client(bufnr, client_id)
+        vim.fn.wait(1000, on_init_called)
```
wookayin, line 520:
```suggestion
        local ok = vim.wait(1000, function() return on_init_called end)
        assert(ok, "on_init was not called")
```

test/functional/plugin/lsp/testutil.lua
```diff
@@ -39,8 +39,7 @@ M.create_server_definition = [[
         })
         local handler = handlers[method]
         if handler then
-          local response, err = handler(method, params)
-          callback(err, response)
+          handler(method, params, callback)
```
wookayin, line 42: Is this a really necessary change (using `callback` for all the handlers) for this PR? Seems complex. This doesn't work like asynchronously, and doesn't seem quite relevant with `initialize` being deferred.

test/functional/plugin/lsp_spec.lua
wookayin: Hmm actually this won't be enough either, because `on_attach` will be called after `on_init` so we will have to wait a bit more. (btw, the previous implementation is not evaluating `on_init_called` properly) It should be something like: until `on_attach` is called.

test/functional/plugin/lsp/testutil.lua
icholy: You requested that the test simulate the more realistic async scenario. So I needed to change the test server to support async responses.

icholy: If the tests seem too verbose, we could add a `handler_stub` testutil like:

``` lua
function M.handler_stub(response, err)
  return function (_, _, callback)
    callback(err, response)
  end
end
```

The most handler would look like this:

```
local handlers = {
  ["textDocument/codeAction"] = handler_stub({
    title = "Code Action 1",
    command = {
      title = "Command 1",
      command = "command:1",
    }
  })
}
```

test/functional/plugin/lsp_spec.lua
icholy: > Hmm actually this won't be enough either, because on_attach will be called after on_init so we will have to wait a bit more.

I think it's fine because `on_init` and `_on_attach` are both called synchronously and should both run in the same event loop iteration.

test/functional/plugin/lsp/testutil.lua
icholy: I actually did exactly this at first. But passing a callback is less code, provides more control, and is easier to understand.

test/functional/plugin/lsp/testutil.lua
mfussenegger: Could also support either a function with synchronous result or a table that has something like: `{type: async|sync, handler: fun}`

Let me know what you think. Otherwise I can also merge the PR as is. Changing them all to async with callback is fine by me and would also match what came up as part of https://github.com/neovim/neovim/pull/24338 (which I should get back to at some point)

icholy: I'm happy with the current state of the PR. The callback signature didn't increase the existing test LOC so I don't think adding overloads is worth it. I've squashed the changes.

PR #28915, opened 2024-05-22
docs: move vim.health documentation to lua.txt

dundargoc:
`vim.health` is not a "plugin" but part of our Lua API and the
documentation should reflect that. This also helps make the
documentation maintenance easier as it is now generated.


clason (approved):
💯

runtime/doc/lua.txt
```diff
@@ -4430,4 +4430,129 @@ tohtml.tohtml({winid}, {opt})                         *tohtml.tohtml.tohtml()*
...
+with healthchecks for configuration, performance, python support, ruby
+support, clipboard support, and more.
+
+To run all healthchecks, use: >vim
+
+        :checkhealth
+<
+Plugin authors are encouraged to write new healthchecks. |health-dev|
+
+Commands                                                     *health-commands*
```
wookayin, line 4448: nit: the helptag *health* is gone, it might be a good idea to add here.

dundargoc: Added to https://github.com/neovim/neovim/pull/28837

justinmk:
I don't have a complete answer for when modules deserve their own help file, but this one could stay as its own file. There is no requirement that all Lua modules live in lua.txt (though that's good as a _starting point_ until we see a clear need to extract to a dedicated help file).

PR #28937, opened 2024-05-23
docs: extract health to its own file

dundargoc:


clason:
I think I'd prefer this to be a coordinated "Problem: lua.txt is too large" split-up into more dedicated files. I don't think `vim.health` is any larger or more different than, say, `vim.iter` or `vim.fs`.

My biggest gripe with `lua.txt` is the inconsistent headings (e.g., the `Ringbuf` module being part of `vim.inspector`). Try `gO`.

dundargoc:
> iff we decide that this is in fact a problem, which I don't believe we have?

I interpreted @justinmk's response in https://github.com/neovim/neovim/pull/28915#issuecomment-2126641258 to indicate that he was not satisfied with merging the health docs to lua.txt which is why I extracted it. I'm not sure what the problem we're solving here either.

clason:
That is not my interpretation (but he will say what he meant himself), only a general comment that it's not _necessary_ for Lua modules to live in `lua.txt` (but that it's a good default, so your PR is good).

runtime/doc/health.txt
```diff
@@ -0,0 +1,127 @@
+
```
justinmk, line 1: needs the usual headers, note about "gO", etc

dundargoc:
> That is not my interpretation (but he will say what he meant himself); it looked to me like only a general comment that it's not _necessary_ for Lua modules to live in `lua.txt` (but that it's a good default, so your PR is good).

For note keeping and reference: spoke to justin in private and I did indeed interpret him correctly and that he prefers health.txt being a separate file.

I will try documenting his reasoning in a way that makes sense in this PR.

scripts/gen_help_html.lua
```diff
@@ -1440,7 +1440,6 @@ function M.test_gen(help_dir)
     -- Because gen() is slow (~30s), this test is limited to a few files.
     { 'help.txt', 'index.txt', 'nvim.txt' }
   )
-  eq(3, #rv.helpfiles)
```
justinmk, line 1443: was this intentionally removed? the purpose is to positively confirm that something is actually being tested. otherwise it's possible that `eq(0, err_count)` (below) could be a false negative.

dundargoc: Yes, intentional. It seems useless.

> the purpose is to positively confirm that something is actually being tested.

Don't know what you mean. We're asserting that the length of the table is the length of the table.

dundargoc: We can rewrite this as `eq(vim.tbl_count(rv.helpfiles), #rv.helpfiles)` if you like that more?

clason: That's a _different_ table than the one on line 1441, though (output, not input).

dundargoc: In that case we should be comparing the actual length of the input table as to not need to manually update this value every time we remove or add an element rather than hardcode it. There is no way to know the intention of what `3` is supposed to represent without prior knowledge.

clason: That makes sense, yes, and would be a useful change.

dundargoc: https://github.com/neovim/neovim/pull/28976

PR #28938, opened 2024-05-23
fix(comment): fall back to using trimmed comment markers

### LINKED ISSUE #28872: Commenting is too restrictive for toggling uncomment
### Problem

Currently the built in comment module is using the commentstring option as is, and does not assume anything about the format of it. I understand that this is intended, and I am all for this when _adding_ a comment. However, IMO, it is a bit restrictive when toggling a comment (uncommenting).

Say I have the following file. I am working together with someone, and they have added a comment here and does so without a trailing space. Maybe they have done this without a comment plugin or something

```lua
--local foo = 10
```

My commentstring is the default one `-- %s`.
Toggling the comment will result in this:
```lua
-- --local foo = 10
```
This can easily be fixed with using visual block mode, or adding a space and then uncommenting.

However, say I have a project in C, and someone is commenting with the default commentstring of `/*%s*/`.

```c
int main() {
    /*int foo = 10;*/
}
```

Then I have the autocmd from the PR that always add a space, or maybe I have changed the commentstring to have a space: `/* %s */`. This is a surrounding comment, and will result in:

```c
int main() {
    /* /*int foo = 10;*/ */
}
```

This is even more annoying,, and I have to delete it from both sides of the line(s)

### Expected behavior

I would expect it to respect the commentstring 100% when commenting, and adding however many spaces there is in the commentstring option. However, I would also expect it to properly uncomment both of these examples, since they actually are comments, and I technically have the correct commentstring.

I am open to looking into this, and sending a PR, but I would like to open an issue first to see what the general opinion is for this.

echasnovski:
Problem: Currently comment detection, addition, and removal are done
  by matching 'commentstring' exactly. This has the downside when users
  want to add comment markers with space (like with `-- %s`
  commentstring) but also be able to uncomment lines that do not contain
  space (like `--aaa`).

Solution: Use the following approach:
  - Line is commented if it matches 'commentstring' with trimmed parts.
  - Adding comment is 100% relying on 'commentstring' parts (as is now).
  - Removing comment is first trying exact 'commentstring' parts with fallback on trying its trimmed parts.

------

Resolve #28872

### DIFF
--- a/runtime/doc/various.txt
+++ b/runtime/doc/various.txt
@@ -579,6 +579,10 @@ Acting on multiple lines behaves as follows:
   transformed to empty comments (e.g. `/**/`). Comment markers are aligned to
   the least indented line.
 
+Matching 'commentstring' does not account for whitespace in comment markers.
+Removing comment markers is first attempted exactly, with fallback to using
+markers trimmed from whitespace.
+
 If the filetype of the buffer is associated with a language for which a
 |treesitter| parser is installed, then |vim.filetype.get_option()| is called
 to look up the value of 'commentstring' corresponding to the cursor position.
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -77,14 +77,11 @@ local function make_comment_check(parts)
   local l_esc, r_esc = vim.pesc(parts.left), vim.pesc(parts.right)
 
   -- Commented line has the following structure:
-  -- <possible whitespace> <left> <anything> <right> <possible whitespace>
-  local nonblank_regex = '^%s-' .. l_esc .. '.*' .. r_esc .. '%s-$'
-
-  -- Commented blank line can have any amount of whitespace around parts
-  local blank_regex = '^%s-' .. vim.trim(l_esc) .. '%s*' .. vim.trim(r_esc) .. '%s-$'
+  -- <whitespace> <trimmed left> <anything> <trimmed right> <whitespace>
+  local regex = '^%s-' .. vim.trim(l_esc) .. '.*' .. vim.trim(r_esc) .. '%s-$'
 
   return function(line)
-    return line:find(nonblank_regex) ~= nil or line:find(blank_regex) ~= nil
+    return line:find(regex) ~= nil
   end
 end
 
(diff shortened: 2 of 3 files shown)


clason:
How does this work with _leading_ whitespace in `'cms'` (e.g., COBOL)?

runtime/doc/various.txt
```diff
@@ -579,6 +579,10 @@ Acting on multiple lines behaves as follows:
   transformed to empty comments (e.g. `/**/`). Comment markers are aligned to
   the least indented line.
 
+Matching 'commentstring' does not account for whitespace in comment markers.
```
clason, line 582:
```suggestion
Matching 'commentstring' does not account for whitespace in comment markers when uncommenting.
```

echasnovski:
> How does this work with _leading_ whitespace in `'cms'` (e.g., COBOL)?

I thought it was a joke, because there is no 'commentstring' in 'cobol.vim' and I did not understand that it was about leading spaces.

I believe the reason it has no set 'commentstring' is because `setlocal commentstring=\ \ \ \ \ \ *%s` will match any `*` in at least column 7, while the proper way seems to be `*` *exactly* in column 7. This also does not work with adding comment, as it will insert `      *` to any line (and again, not make `*` appear in column 7).

Right now it will correctly uncomment the correctly commented line. For lines like `  *` this PR will remove only asterisk while current behavior will add `      *`.

So all in all, I don't think this will work for COBOL.

runtime/doc/various.txt
echasnovski: Matching 'commentstring' is about deciding what action should be done: commenting or uncommenting. So I don't think it is a good idea to add this detail.

clason:
If the previous behavior didn't work (reliably) for that case either, then this is no objection.

justinmk (approved):

echasnovski:
> If the previous behavior didn't work (reliably) for that case either, then this is no objection.

Both current and this PR's behaviors work with 'commentstring' which has whitespace "outside" of comment parts (to the left of left part and to the right of right part). It is [tested](https://github.com/neovim/neovim/blob/5ac8db10f0428c976dfc9a4935a74d0fe160995c/test/functional/lua/comment_spec.lua#L182-L229).

It works in a sense that that whitespace is included in comment markers: adding comment adds markers with that whitespace, matching is done with trimmed markers (after this PR), and uncommenting (after this PR) first tries with that whitespace and then falls back without that whitespace. The logic is that any whitespace in commentstring is trimmed in fallback: line `--aaa` should be uncommented even with `  --  %s` commentstring. This is the core of the issue.

Now, when it comes to comment structure of COBOL specifically ("A comment line is any line with an asterisk (*) or slash (/) in the indicator area (column 7) of the line"), then it indeed doesn't work now for 100% of cases with `      *%s` commentstring:
- It works if line is properly commented (starts with `      *`) or not commented (starts at column 1). After this PR it will continue to work.
- It does not work otherwise. Adding comment does not ensure that line starts with `      *` but only adds this prefix at first non-blank column. After this PR it will continue to not work.

folke:
Great!

Maybe for a follow-up PR, could we also use `vim.opt.comments` for uncommenting?

clason:
If you can understand it, be our guest!

echasnovski:
> Great!
> 
> Maybe for a follow-up PR, could we also use `vim.opt.comments` for uncommenting?

I am more inclined to not have it used for uncommenting. Here is the copy of my reasoning from target issue:

> The most proper solution would be to have a separate options for "comment structure for adding comment" and "comment regex to match commented line", but there is no such thing. The closest is 'comments' option but it is more about how to place proper prefixes during editing (like after `<CR>`, `O`, `J`, etc.). It is also quite popularly used to define bulleted lists and has somewhat complicated set of flags to implement.

My guess would be that this subfeature alone would increase module LoC count by at least 50%. Together with its "not exactly comments" usage, I'd not rely on it.

A separate dedicated option ('uncommentlist' maybe?) might be a better solution.

echasnovski:
> So this should give a list of comment strings we can use for uncommenting

Which would itself increase complexity of current code designed only for a single source of comment markers.

PR #28939, opened 2024-05-23
refactor(lsp): reuse buf_detach_client logic in on_detach

mfussenegger:


justinmk (approved):
Nice

runtime/lua/vim/lsp.lua
```diff
@@ -501,6 +501,30 @@ local function text_document_did_save_handler(bufnr)
...
+  changetracking.reset_buf(client, bufnr)
+
+  if client.supports_method(ms.textDocument_didClose) then
+    local uri = vim.uri_from_bufnr(bufnr)
+    local params = { textDocument = { uri = uri } }
+    client.notify(ms.textDocument_didClose, params)
+  end
+
+  client.attached_buffers[bufnr] = nil
+  util.buf_versions[bufnr] = nil
```
icholy, line 522: This should only happen in `on_detach` no?

mfussenegger: Yes this seemed odd to me too but it matches the previous behavior and I didn't want to change that in the same commit.

icholy: Fair

icholy:
@justinmk @wookayin @mfussenegger this somehow fixed https://github.com/neovim/neovim/issues/28987 and should be backported.

mfussenegger:
> @justinmk @wookayin @mfussenegger this somehow fixed #28987 and should be backported.

Thanks for pointing out. I suspect it's due to the `_initialized` filter change in the get_clients call.

PR #28943, opened 2024-05-23
refactor(lsp): replace util.buf_versions with changedtick

mfussenegger:
`lsp.util.buf_versions` was already derived from changedtick (`on_lines`
from `buf_attach` synced the version)

As far as I can tell there is no need to keep track of the state in a
separate table.

### DIFF
--- a/runtime/doc/deprecated.txt
+++ b/runtime/doc/deprecated.txt
@@ -22,6 +22,9 @@ API
 LUA
 - vim.region()		Use |getregionpos()| instead.
 
+LSP
+- *vim.lsp.util.buf_versions*	Use |b:changedtick| instead.
+
 DIAGNOSTICS
 - *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count = 1}` instead.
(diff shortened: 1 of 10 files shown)


icholy:
https://github.com/neovim/nvim-lspconfig/blob/2c1877081b237a643e52ebdebaf36c84a2695639/lua/lspconfig/server_configurations/eslint.lua#L29

icholy:
Might make sense to hide the `vim.b[bufnr].changedtick` inside a `util.buf_version(bufnr)` function:

``` lua
function M.buf_version(bufnr)
  return vim.b[bufnr].changedtick
end
```

This way it can be changed in the future without breaking anything.

mfussenegger:
> Might make sense to hide the `vim.b[bufnr].changedtick` inside a `util.buf_version(bufnr)` function:
> 
> ```lua
> function M.buf_version(bufnr)
>   return vim.b[bufnr].changedtick
> end
> ```
> 
> This way it can be changed in the future without breaking anything.

I don't think that's necessary and it would defeat the purpose of this PR (getting rid of unnecessary extra API).
`util.buf_versions` wasn't even intended as public API (hence the `@nodoc`) and I was even inclined to skip the deprecation step because of that.

PR #28949, opened 2024-05-23
feat(snippet): add missing variables from VSCode snippet spec

chrisgrieser:
`vim.snippet.expand()` currently only resolves some variables [VSCode snippet spec](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables), namely the `TM_*` variables. 

This PR adds the remaining variables, so that nvim fully supports the VSCode snippet standard. This is also relevant for users of [friendly-snippets](https://github.com/rafamadriz/friendly-snippets), which does make use of the currently unsupported variables.

The only ones still missing are the ones not applicable to nvim, such as `$CURSOR_INDEX` due to vim lacking multi-cursor capability. A potential point of discussion is whether `$CLIPBOARD` should use the `+` or the `"` register. Luasnip expands `$CLIPBOARD` to the `"` register; however, I presume that `+` makes more sense, since snippets are often used to insert things copypasted from outside vim (e.g., a URL copied from the browser). I settled on using the `+` register, but can change that if there are arguments against it.

I did not update `news.txt`, since I assume this PR is not significant enough for the news.

### DIFF
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -29,6 +29,12 @@ local function resolve_variable(var, default)
     return expansion == '' and default or expansion
   end
 
+  --- @param str string
+  --- @return string
+  local function expand_date(str)
+    return tostring(os.date(str))
+  end
+
   if var == 'TM_SELECTED_TEXT' then
     -- Snippets are expanded in insert mode only, so there's no selection.
     return default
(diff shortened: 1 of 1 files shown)


clason:
> This PR adds the remaining variables, so that nvim fully supports the VSCode snippet standard.

We don't care about VS Code, though; we only care about the LSP specification.

mfussenegger:
Thanks for the PR, but as @clason mentioned anything vscode specific is out of scope for neovim core.
See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax  for what's in scope.

chrisgrieser:
Ugh, was not aware that there is an LSP snippet spec and an VSCode snippet spec, which are not the same. 

Well, makes sense to limit nvim core to the LSP spec, yeah.

folke:
What about a PR that exposes `vim.snippet.variables`? So that users/plugins can define their own variables?

clason:
Out of scope; we want to keep these functions minimal (for now). If you want more variables, lobby to include them in the spec.

clason:
(And there's no "VS Code spec". It's just VS Code acting like the 800 pound gorilla and doing whatever it wants without regard for "specs".)

folke:
I understand, but NEovim should also be extensible.
Lots of people use friendly-snippets and right now it's not really possible to configure custom variables when parsing snippets.

The only way to hook into the vim.snippets code is by getting/setting upvalues twice using lua debug functions.

That's far from ideal

clason:
That is very much a non-goal **right now**. We first want to implement the LSP spec (which includes snippet expansion for completion candidates). Once that is done, we will consider exposing the building blocks for plugins to build on in a way that makes sense given our API policy.

The idea is to offer a low-level API (expansion, jumping, etc.) we use ourselves but _not_ a "bundled plugin" that others can hook into or monkey patch.

telemachus:
> (Let's be clear here: "the specs, and nothing but the specs, ma'am" is a principled project policy. We can't let Microsoft get away with treating their own spec like a pirates' code forever.)

I understand wanting a minimal API, but I don't understand this. What are you trying to prevent? What do you mean by "a pirates' code" (cove?)?

telemachus:
@MariaSolOs, thanks for clarifying.

mfussenegger:
> Lots of people use friendly-snippets and right now it's not really possible to configure custom variables when parsing snippets.

For most of the variables this PR would add it's possible to pre-evaluate them on the snippet provider side and include the resulting text in the snippet instead of the variable. That's also how language servers deal with the limitation. 

Just need to escape `$`, `}` and `\` in case they can occur in the text.

E.g. I have stuff like:

```lua
local snippets = {
  ["*"] = {
    date = function()
      return os.date("%Y-%m-%d") --[[@as string]]
    end,
    uuid = function()
      return vim.trim(vim.fn.system("uuidgen"))
    end,
  },
  lua = {
    f = "local function ${1:name}(${2})\n  $0\nend",
    m = "function ${1:M}.${2:name}(${3})\n  $0\nend",
    ...
```


And use it with:

```lua
function M.expand()
  local lnum, col = unpack(api.nvim_win_get_cursor(0))
  local line = api.nvim_get_current_line()
  local word = vim.fn.matchstr(line, "\\w\\+\\%.c")
  local ft_snippets = snippets[vim.bo.filetype] or {}
  local snippet = ft_snippets[word]
  if not snippet then
    snippet = snippets["*"][word]
  end
  if snippet then
    api.nvim_buf_set_text(0, lnum - 1, col - #word, lnum - 1, col, {})
    if type(snippet) == "function" then
      snippet = snippet()
    end
    vim.snippet.expand(snippet)
    return
  end
  ...
```

folke:
@mfussenegger great! That's indeed a simple way to support snippet variables. Thanks for sharing.

PR #28954, opened 2024-05-23
docs: update LSP quickstart

gpanders:
The LSP quickstart can act as our true "entrypoint" for answering the question "How do I use LSP in Neovim?" As such, it can be a little more beginniner-friendly than other sections of our help docs by including explanatory comments and a more fleshed out example (including a `FileType` autocommand).

This also includes some other minor wording updates and points users toward `:checkhealth lsp`.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -36,17 +36,14 @@ Follow these steps to get LSP features:
       -- This handler will fire when the buffer's 'filetype' is "python"
       pattern = 'python',
       callback = function(ev)
-        -- The buffer number this event is associated with
-        local bufnr = ev.buf
-
         vim.lsp.start({
           name = 'my-server-name',
           cmd = {'name-of-language-server-executable', '--option', 'arg1', 'arg2'},
 
-          -- The "root directory" of the project is the parent directory,
-          -- relative to the buffer's file, containing a "setup.py" or
-          -- "pyproject.toml" file
-          root_dir = vim.fs.root(bufnr, {'setup.py', 'pyproject.toml'}),
+          -- Set the "root directory" to the parent directory of the
+          -- file in the current buffer (`ev.buf`) that contains either
+          -- a "setup.py" or a "pyproject.toml" file.
+          root_dir = vim.fs.root(ev.buf, {'setup.py', 'pyproject.toml'}),
         })
       end,
     })
@@ -54,7 +51,7 @@ Follow these steps to get LSP features:
 3. Check that the server attached to the buffer: >
     :checkhealth lsp
 
-4. (Optional) Configure keymaps and autocmds to use LSP features. |lsp-config|
+4. (Optional) Configure keymaps and autocommands to use LSP features. |lsp-config|
 
                                                         *lsp-defaults*
 When the LSP client starts it enables diagnostics |vim.diagnostic| (see
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -12,7 +12,7 @@ enhanced LSP tools.
 
     https://microsoft.github.io/language-server-protocol/
 
-LSP facilitates features like go-to-definition, find-references, hover,
+LSP facilitates features like go-to-definition, find references, hover,
 completion, rename, format, refactor, etc., using semantic whole-project
 analysis (unlike |ctags|).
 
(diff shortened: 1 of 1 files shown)


runtime/doc/lsp.txt
```diff
@@ -25,24 +25,37 @@ Nvim provides an LSP client, but the servers are provided by third parties.
 Follow these steps to get LSP features:
 
 1. Install language servers using your package manager or by following the
-   upstream installation instruction. You can find language servers here:
+   upstream installation instructions. You can find language servers here:
    https://microsoft.github.io/language-server-protocol/implementors/servers/
```
MariaSolOs, line 29: Would it make sense to link to `nvim-lspconfig` instead?

runtime/doc/lsp.txt
```diff
@@ -25,24 +25,37 @@ Nvim provides an LSP client, but the servers are provided by third parties.
...
 
 2. Configure the LSP client per language server. See |vim.lsp.start()| or use
```
MariaSolOs, line 31: I'm not a fan of this first sentence. If this is meant to be an introductory explanation, we should remain consistent of who's the client and who's the server.

runtime/doc/lsp.txt
```diff
@@ -25,24 +25,37 @@ Nvim provides an LSP client, but the servers are provided by third parties.
...
+        })
+      end,
     })
 <
 3. Check that the server attached to the buffer: >
-    :lua =vim.lsp.get_clients()
+    :checkhealth lsp
 
-4. Configure keymaps and autocmds to use LSP features. See |lsp-config|.
+4. (Optional) Configure keymaps and autocmds to use LSP features. |lsp-config|
```
MariaSolOs, line 57: For consistency with line 34:
```suggestion
4. (Optional) Configure keymaps and autocommands to use LSP features. |lsp-config|
```

runtime/doc/lsp.txt
gpanders: Not all users will have nvim-lspconfig installed, and in any case the hope here is that we can get more users to use LSP without it.

runtime/doc/lsp.txt
```diff
@@ -25,24 +25,37 @@ Nvim provides an LSP client, but the servers are provided by third parties.
...
    https://microsoft.github.io/language-server-protocol/implementors/servers/
 
 2. Configure the LSP client per language server. See |vim.lsp.start()| or use
    this minimal example as a guide: >lua
 
-    vim.lsp.start({
-      name = 'my-server-name',
-      cmd = {'name-of-language-server-executable'},
-      root_dir = vim.fs.root(0, {'setup.py', 'pyproject.toml'}),
+    -- Create an event handler for the FileType autocommand
```
clason, line 34: Nitpick: the "event handler" _is_ the "autocommand" (in Vimspeak) for the `FileType` "event".

Of course, the question is how the audience is here -- whether we should "push" Vimspeak or generic protocol language.

runtime/doc/lsp.txt
```diff
@@ -25,24 +25,37 @@ Nvim provides an LSP client, but the servers are provided by third parties.
...
+    vim.api.nvim_create_autocmd('FileType', {
+      -- This handler will fire when the buffer's 'filetype' is "python"
+      pattern = 'python',
+      callback = function(ev)
+        -- The buffer number this event is associated with
```
clason, line 39: I would recommend structuring these comments around the _intention_ rather than a technical description. In this case, I would actually _not_ define this here but use the event structure where it is needed (see below).

runtime/doc/lsp.txt
```diff
@@ -25,24 +25,37 @@ Nvim provides an LSP client, but the servers are provided by third parties.
...
+        local bufnr = ev.buf
+
+        vim.lsp.start({
+          name = 'my-server-name',
+          cmd = {'name-of-language-server-executable', '--option', 'arg1', 'arg2'},
+
+          -- The "root directory" of the project is the parent directory,
+          -- relative to the buffer's file, containing a "setup.py" or
+          -- "pyproject.toml" file
+          root_dir = vim.fs.root(bufnr, {'setup.py', 'pyproject.toml'}),
```
clason, line 49:
```suggestion
          -- Set the "root directory" to the parent directory of the
          -- file in the current buffer (`ev.buf`) that contains either
          --  a "setup.py" or a "pyproject.toml" file
          root_dir = vim.fs.root(ev.buf, {'setup.py', 'pyproject.toml'}),
```

runtime/doc/lsp.txt
gpanders: Trying to avoid Vim specific jargon here. We can leave lots of breadcrumbs (tags) to other help docs to guide the user to learn more, but at this point I want to assume as little a priori knowledge as possible.

runtime/doc/lsp.txt
gpanders: The intent here was to (hopefully) avoid confusion between "current buffer" and "buffer associated with the autocommand". Pulling the `bufnr` declaration out allows us to add a separate comment explaining this distinction. But perhaps this is not an important distinction for new users.

runtime/doc/lsp.txt
gpanders: Which sentence in particular? Do you have a suggestion on how to reword it?

mfussenegger (approved):

justinmk (approved):

clason (approved):

runtime/doc/lsp.txt
MariaSolOs: I was not suggesting replacing this by "use `nvim-lspconfig`", but to use their list of available servers instead since I think it will better reflect which LSPs currently have neovim support.
I don't have strong feelings about it though so it's fine.

MariaSolOs (approved):

PR #28961, opened 2024-05-24
fix(extmarks): issues with revalidating marks

### LINKED ISSUE #28906: Undo sometimes triggers assert when restoring extmarks
### Problem

On NEovim nightly, but **not** on **0.10.0**, I sometimes get segfaults when undoing.

stack trace:

```c
#0  0x0000746da4456e44 in ?? () from /usr/lib/libc.so.6
#1  0x0000746da43fea30 in raise () from /usr/lib/libc.so.6
#2  0x0000746da43e64c3 in abort () from /usr/lib/libc.so.6
#3  0x0000746da43e63df in ?? () from /usr/lib/libc.so.6
#4  0x0000746da43f6c67 in __assert_fail () from /usr/lib/libc.so.6
#5  0x00005db5c84d43bb in buf_signcols_count_range (buf=0x5db5c8aff9b0, row1=4, row2=5, add=add@entry=1, clear=clear@entry=kFalse)
    at /home/folke/projects/neovim/src/nvim/decoration.c:839
#6  0x00005db5c84d44b5 in buf_put_decor_sh (buf=buf@entry=0x5db5c8aff9b0, sh=sh@entry=0x5db5c8c55fe8, row1=row1@entry=4, row2=row2@entry=5)
    at /home/folke/projects/neovim/src/nvim/decoration.c:193
#7  0x00005db5c84d4507 in buf_put_decor (buf=0x5db5c8aff9b0, decor=..., row=4, row2=5) at /home/folke/projects/neovim/src/nvim/decoration.c:181
#8  0x00005db5c855ad8e in extmark_setraw (buf=0x5db5c8aff9b0, mark=<optimized out>, row=4, col=0, invalid=true) at /home/folke/projects/neovim/src/nvim/extmark.c:144
#9  0x00005db5c855e118 in extmark_apply_undo (undo_info=..., undo=<optimized out>) at /home/folke/projects/neovim/src/nvim/extmark.c:440
#10 0x00005db5c86ce286 in u_undoredo (undo=undo@entry=true, do_buf_event=do_buf_event@entry=true) at /home/folke/projects/neovim/src/nvim/undo.c:2431
#11 0x00005db5c86cebab in u_doit (startcount=startcount@entry=1, quiet=quiet@entry=false, do_buf_event=do_buf_event@entry=true)
    at /home/folke/projects/neovim/src/nvim/undo.c:1900
#12 0x00005db5c86cec42 in u_undo (count=1) at /home/folke/projects/neovim/src/nvim/undo.c:1797
#13 0x00005db5c85f8e46 in nv_kundo (cap=0x7fffa8e11140) at /home/folke/projects/neovim/src/nvim/normal.c:4432
#14 0x00005db5c85fc410 in nv_undo (cap=<optimized out>) at /home/folke/projects/neovim/src/nvim/normal.c:4417
#15 0x00005db5c85f6328 in normal_execute (state=0x7fffa8e110d0, key=<optimized out>) at /home/folke/projects/neovim/src/nvim/normal.c:1229
#16 0x00005db5c8697f8e in state_enter (s=0x7fffa8e110d0) at /home/folke/projects/neovim/src/nvim/state.c:101
#17 0x00005db5c85f350a in normal_enter (cmdwin=<optimized out>, noexmode=<optimized out>) at /home/folke/projects/neovim/src/nvim/normal.c:518
#18 0x00005db5c85b5bed in main (argc=2, argv=<optimized out>) at /home/folke/projects/neovim/src/nvim/main.c:664
```

### Steps to reproduce

I can reliably reproduce the issue in my config, but haven't been able to create a minimal repro.

I did find however that it seems to be related to setting `vim.opt.satuscolumn`. Even when setting to a lua function that returns an empty string, the segafult happens.

When not setting the option, I dont get the segfaults.

### Expected behavior

Should not segfault

### Neovim version (nvim -v)

NVIM v0.11.0-dev-45+g879d17ea8

na

### Operating system/version

linux

### Terminal name/version

wezterm

xterm-256-color

### Installation

bob

luukvbaal:
Problem:  Invalid marks appear to be revalidated multiple times, and
          decor is added at the old position for unpaired marks.
Solution: Avoid revalidating already valid marks, and don't use old
          position to add to decor for unpaired marks.

Fix #28906

### DIFF
--- a/src/nvim/extmark.c
+++ b/src/nvim/extmark.c
@@ -117,7 +117,14 @@ static void extmark_setraw(buf_T *buf, uint64_t mark, int row, colnr_T col, bool
   MarkTreeIter itr[1] = { 0 };
   MTKey key = marktree_lookup(buf->b_marktree, mark, itr);
   if (key.pos.row < 0 || (key.pos.row == row && key.pos.col == col)) {
+    // Does this hold? If it doesn't, we should still revalidate.
+    assert(!invalid || !mt_invalid(key));
     return;
+  }
+
+  // Key already revalidated(how?) Avoid adding to decor again.
+  if (invalid && !mt_invalid(key)) {
+    invalid = false;
   }
 
   // Only the position before undo needs to be redrawn here,
@@ -140,8 +147,8 @@ static void extmark_setraw(buf_T *buf, uint64_t mark, int row, colnr_T col, bool
   marktree_move(buf->b_marktree, itr, row, col);
 
   if (invalid) {
-    MTPos end = marktree_get_altpos(buf->b_marktree, key, NULL);
-    buf_put_decor(buf, mt_decor(key), row, end.row);
+    row2 = mt_paired(key) ? marktree_get_altpos(buf->b_marktree, key, NULL).row : row;
+    buf_put_decor(buf, mt_decor(key), row, row2);
   } else if (key.flags & MT_FLAG_DECOR_SIGNTEXT && buf->b_signcols.autom) {
     buf_signcols_count_range(buf, row1, row2, 0, kNone);
   }


src/nvim/extmark.c
```diff
@@ -117,9 +117,16 @@ static void extmark_setraw(buf_T *buf, uint64_t mark, int row, colnr_T col, bool
   MarkTreeIter itr[1] = { 0 };
   MTKey key = marktree_lookup(buf->b_marktree, mark, itr);
   if (key.pos.row < 0 || (key.pos.row == row && key.pos.col == col)) {
+    // Does this hold? If it doesn't, we should still revalidate.
+    assert(!invalid || !mt_invalid(key));
```
luukvbaal, line 121: `extmark_setraw()` seems to allow returning early when the mark doesn't need to be moved. I don't know if it can happen but maybe the mark may still be invalid at that point? Added an assert; could also just add the right condition to continue instead.

src/nvim/extmark.c
```diff
@@ -117,9 +117,16 @@ static void extmark_setraw(buf_T *buf, uint64_t mark, int row, colnr_T col, bool
...
     return;
   }
 
+  // Key already revalidated(how?) Avoid adding to decor again.
+  if (invalid && !mt_invalid(key)) {
```
luukvbaal, line 126: My mental model for this seems to have been wrong. I'm still not actually sure how this happens, it's very hard to follow the undo tracking in relevant repros. But since it does happen this seems like the right fix.

luukvbaal (commented):
Capturing this in a test is difficult, probably won't be able to do it anytime soon.
The repro from #28906 and some others I used locally now properly track the signcolumn.

WillEhrendreich:
Thanks!

PR #28964, opened 2024-05-24
fix(fs): make vim.fs.root work for relative paths and unnamed buffers

gpanders:
If a buffer does not have a backing file then fall back to the current working directory.

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -328,8 +328,11 @@ function M.find(names, opts)
   return matches
 end
 
---- Find the first parent directory containing a specific "marker", relative to a buffer's
---- directory.
+--- Find the first parent directory containing a specific "marker", relative to a file path or
+--- buffer.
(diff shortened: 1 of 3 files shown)


mfussenegger (approved):
:+1:

runtime/lua/vim/fs.lua
```diff
@@ -368,7 +371,7 @@ function M.root(source, marker)
 
   local paths = M.find(marker, {
     upward = true,
-    path = path,
+    path = vim.fn.fnamemodify(path, ':p:h'),
```
clason, line 374: Can't hurt to document this (that a string `{source}` can both be relative and absolute).

Same for file vs. directory (if both are supported now).

gpanders: 7f140398b9ec028a5396777d29c1e7f25d262ebe

mfussenegger (approved):

PR #28971, opened 2024-05-24
fix: show swapfile warning as a warning

gpanders:
The new default SwapExists autocommand displays warning text (W325) but does not use the WarningMsg highlight group as other warnings do. Use the WARN log level when displaying this warning.


justinmk (approved):

gpanders:
macos-14 clang unittest failure is a flake

Issue #28921, opened 2024-05-22
performance: vim.validate, vim.startswith, vim.endswith are slow


hoofcushion:
### Problem

I have identified a performance issue in the Neovim Lua API, specifically with the `vim.validate` function. The function is used to validate the types and values of function arguments, which is crucial for stability and correctness. However, it appears to be significantly slower than alternative approaches, which could impact the overall performance of Lua scripts and plugins in Neovim. And the plugin developers may need to create a no validation version of that function again.

### Conclusion
The performance of `vim.validate` is a critical concern for Lua script developers and users of Neovim. By implementing the proposed solution, we can maintain the robustness of our code while significantly improving performance. This will benefit the entire Neovim community by enabling faster and more efficient Lua scripting.

### Steps to reproduce

1. Run the attached Lua script.
2. Observe the execution times for each function.

### Script for Benchmarking
Below is the Lua script used for benchmarking the performance of different validation approaches:

```lua
---@diagnostic disable
local function test_speed(title, t, fn)
  local start_time = os.clock()
  local i = 0
  while i < t do
    fn()
    i = i + 1
  end
  local end_time = os.clock()
  local time_taken = end_time - start_time
  print(title .. ": " .. t .. " times took " .. time_taken .. " seconds")
end

local function empty() end

function normal_one(s, suffix)
  vim.validate({ s = { s, 's' }, suffix = { suffix, 's' } })
  return #suffix == 0 or s:sub(-#suffix) == suffix
end

local function no_validation(s, suffix)
  return #suffix == 0 or s:sub(-#suffix) == suffix
end

local function only_create_table(s, suffix)
  empty({ s = { s, "s" }, suffix = { suffix, "s" } })
end

local function pre_created_table(s, suffix)
  return #suffix == 0 or s:sub(-#suffix) == suffix
end

local function _validate(fn, vali, length)
  return function(...)
    for i = 1, length do
      vali[i][1] = select(i, ...)
    end
    vim.validate(vali)
    fn(...)
  end
end

pre_created_table = _validate(pre_created_table, {{ nil, "s" }, { nil, "s" }}, 2)

local t = 10000000
test_speed("vim.endswith", t, function() normal_one("hello world", "world") end)
test_speed("no validation", t, function() no_validation("hello world", "world") end)
test_speed("only create table", t, function() only_create_table("hello world", "world") end)
test_speed("pre-created table", t, function() pre_created_table("hello world", "world") end)
```

### Expected behavior

[...]

PR #28977, opened 2024-05-24
perf: add fast path to vim.validate

gpanders:
For many small/simple functions (like those found in shared.lua), the runtime of vim.validate can far exceed the runtime of the function itself. Add an "overload" to vim.validate that uses a simple assertion pattern, rather than parsing a full "validation spec".

Ref: https://github.com/neovim/neovim/issues/28921

Using a modified version of the script in the linked issue, the runtime performance compared to `vim.validate` is as follows:

```
$ VIMRUNTIME=runtime build/bin/nvim --headless -l scratch/test.lua
vim.validate (old): 10000000 times took 1.24831 seconds
no validation: 10000000 times took 0.0033800000000002 seconds
vim.validate (new): 10000000 times took 0.003174 seconds
```

The `vim.validate (new)` test case is this PR. We can see that this is just as fast as not using any validation at all (within measurement noise).


runtime/lua/vim/shared.lua
```diff
@@ -10,6 +10,21 @@
...
+--- @generic T
+--- @param v T
+--- @param expected string Expected type
+local function expecttype(v, expected)
+  local actual = type(v)
+
+  -- Use an if statement with error() instead of assert() to avoid calling string.format if we don't
+  -- need it.
+  if actual ~= expected then
+    error(('expected %s, got %s'):format(expected, actual))
```
mfussenegger, line 24: Maybe worth including the actual value in addition to the types?

runtime/lua/vim/shared.lua
```diff
@@ -10,6 +10,21 @@
...
+--- Assert that a value has the expected type.
+---
+--- @generic T
+--- @param v T
+--- @param expected string Expected type
+local function expecttype(v, expected)
+  local actual = type(v)
+
+  -- Use an if statement with error() instead of assert() to avoid calling string.format if we don't
+  -- need it.
```
gpanders, line 22: Reference: https://chrisfls.github.io/luajit-wiki/Numerical-Computing-Performance-Guide/

> * Do not intersperse expensive or uncompiled operations.
>   - E.g. avoid assert(type(x) == "number", "x is a "..mytype(x)")
>     *  The problem is not the assert() or the condition (basically free). The problem is the string concatenation, which has to be executed every time, even if the assertion never fails!

runtime/lua/vim/shared.lua
gpanders: For `nil` this shows `"expected string, got nil (nil)"` which is a little weird, but probably not weird enough to add a special case?

mfussenegger (approved):

runtime/lua/vim/shared.lua
clason: Since this is the "fast path" (and the fail state), I'd agree it's not worth adding complexity to pretty this up.

runtime/lua/vim/shared.lua
clason: Put the reference in the comment? (May linkrot though.)

clason:
Are there other functions where this would make sense (possibly in a follow-up), e.g., in `editor.lua`, `fs.lua` (`vim.fs.normalize` instead of `_fast = true`), or `treesitter/*` (`language.register` for some reason is fairly slow in nvim-treesitter startup)?

runtime/lua/vim/shared.lua
```diff
@@ -214,7 +228,7 @@ end
 ---@param t table<T, any> (table) Table
 ---@return T[] : List of keys
 function vim.tbl_keys(t)
-  vim.validate({ t = { t, 't' } })
+  expecttype(t, 'table')
```
justinmk, line 231: this could be an overload on `vim.validate`. so for the "fast path", one can call `vim.validate(val: any, type: string)` instead of the kwargs form.

test/functional/lua/vim_spec.lua
```diff
@@ -226,11 +226,8 @@ describe('lua stdlib', function()
     eq(false, fn.luaeval('vim.startswith("123", "2")'))
     eq(false, fn.luaeval('vim.startswith("123", "1234")'))
 
-    matches(
-      'prefix: expected string, got nil',
-      pcall_err(exec_lua, 'return vim.startswith("123", nil)')
-    )
-    matches('s: expected string, got nil', pcall_err(exec_lua, 'return vim.startswith(nil, "123")'))
+    matches('expected string, got nil', pcall_err(exec_lua, 'return vim.startswith("123", nil)'))
+    matches('expected string, got nil', pcall_err(exec_lua, 'return vim.startswith(nil, "123")'))
```
justinmk, line 230: since the string concatenation is guarded by a condition, is there any reason we can't mention the arg name in the message?
```suggestion
    matches('s: expected string, got nil', pcall_err(exec_lua, 'return vim.startswith(nil, "123")'))
```

gpanders: No reason other than we would have to pass the argument name and the argument itself to the function (an extra arg). Could be a bit tedious. vim.validate already does this though so it’s probably fine

justinmk: As long as it's still fast, a bit of tedium seems worth it for the UX improvement.

gpanders: Done here https://github.com/neovim/neovim/commit/8b88d99b2198cba251eb9b28717da523f7fb411f

runtime/doc/lua.txt
```diff
@@ -2366,10 +2366,29 @@ vim.trim({s})                                                     *vim.trim()*
...
+    Validate function arguments.
 
-    Usage example: >lua
+    This function has two valid forms:
+    1. vim.validate(name: str, value: any, type: string, optional: bool)
+    2. vim.validate(spec: table)
+
+    Form 1 validates that argument {name} with value {value} has the type
+    {type}. If {optional} is true, then {value} may be null. This form should
+    be preferred for simple cases.
```
wookayin, line 2377: Would we want to mention that it won't work with "callable" because this path does not use `vim.is_callable`?

gpanders: Mentioned in https://github.com/neovim/neovim/pull/28977/commits/d5d7079557607d124b1d51d90b2e851c1c35104c

runtime/doc/lua.txt
```diff
@@ -2366,10 +2366,29 @@ vim.trim({s})                                                     *vim.trim()*
       • https://www.lua.org/pil/20.2.html
 
 vim.validate({opt})                                           *vim.validate()*
-    Validates a parameter specification (types and values). Specs are
-    evaluated in alphanumeric order, until the first failure.
+    Validate function arguments.
 
-    Usage example: >lua
+    This function has two valid forms:
+    1. vim.validate(name: str, value: any, type: string, optional: bool)
```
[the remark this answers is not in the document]
gpanders: https://github.com/neovim/neovim/pull/28977/commits/d5d7079557607d124b1d51d90b2e851c1c35104c

runtime/doc/lua.txt
gpanders: (and slightly reworded in https://github.com/neovim/neovim/pull/28977/commits/717d170ae0dc753cbb5fac18be0e535c7489d5ae)

runtime/doc/lua.txt
```diff
@@ -2366,7 +2366,26 @@ vim.trim({s})                                                     *vim.trim()*
...
+    Validate function arguments.
+
+    This function has two valid forms:
+    1. vim.validate(name: str, value: any, type: string, optional?: bool)
+    2. vim.validate(spec: table)
+
+    Form 1 validates that argument {name} with value {value} has the type
+    {type}. {type} must be a value returned by |lua-type()|. If {optional} is
+    true, then {value} may be null. This form should be preferred for simple
+    cases.
```
justinmk, line 2378: should we hint that form 1 is faster?
```suggestion
    true, then {value} may be nil. Performance is comparable to assert().
```

runtime/lua/vim/shared.lua
```diff
@@ -930,7 +972,15 @@ do
...
   ---               informative error message as the second returned value.
   ---             - msg: (optional) error string if validation fails
-  function vim.validate(opt)
+  --- @overload fun(name: string, val: any, expected: string, optional?: boolean)
+  function vim.validate(opt, ...)
+    if select('#', ...) >= 2 then
+      -- Overloaded signature for fast/simple cases
+      local name = opt --[[@as string]]
+      local v, expected, optional = ... ---@type string, string, boolean?
+      return expecttype(name, v, expected, optional)
```
justinmk, line 981: is there a reason not to inline `expecttype` ? the extra names feel like needless indirection

justinmk (approved):

runtime/lua/vim/shared.lua
gpanders: c9e0debe156fa0551d53c45f66d60d1ca3546400

runtime/doc/lua.txt
gpanders: feec3da5dcc742ba42a552c3549ccaf02f569ba8

wookayin (approved):
Looks great overall. If not difficult, please consider adding simple test cases for the new signature in the `vim_spec` functionaltest.

justinmk (approved):

PR #28980, opened 2024-05-24
feat(lsp): update LSP healthcheck format

gpanders:
This is mostly an aesthetic change, although there are a few new pieces of information included. Originally I wanted to investigate including server capabilities in the healthcheck, but until we have the ability to fold/unfold text in health checks that would be too much information.

This is a small step toward https://github.com/neovim/neovim/issues/28479 by making `:checkhealth lsp` the go-to source for debugging common LSP issues.

**Before:**

```
vim.lsp: require("vim.lsp.health").check()

- LSP log level : WARN
- Log path: /Users/ganders/.local/state/nvim/lsp.log
- Log size: 1692 KB

vim.lsp: Active Clients ~
- clangd (id=1, root_dir=~/Developer/neovim, attached_to=[1])
- lua-language-server (id=2, root_dir=~/Developer/neovim, attached_to=[2])

vim.lsp: File watcher ~
- file watching "(workspace/didChangeWatchedFiles)" disabled on all clients
```

**After:**

```
vim.lsp: require("vim.lsp.health").check()

- LSP log level : WARN
- Log path: /Users/ganders/.local/state/nvim/lsp.log
- Log size: 1687 KB

vim.lsp: Active Clients ~
- clangd (id: 1)
    Root directory: ~/Developer/neovim
    Command: clangd --background-index
    Settings: {}
    Attached buffers: 1
- lua-language-server (id: 2)
    Root directory: ~/Developer/neovim
    Command: lua-language-server
    Settings: {
      Lua = {
        telemetry = {
          enable = false
        }
      }
    }
    Attached buffers: 2

vim.lsp: File Watcher ~
- file watching "(workspace/didChangeWatchedFiles)" disabled on all clients

vim.lsp: Position Encodings ~
- No buffers contain mixed position encodings
```

### DIFF
--- a/runtime/lua/vim/lsp/health.lua
+++ b/runtime/lua/vim/lsp/health.lua
@@ -40,13 +40,11 @@ local function check_active_clients()
         cmd = tostring(client.config.cmd)
       end
       report_info(table.concat({
-        client.name,
-        string.format('  ID: %d', client.id),
+        string.format('%s (id: %d)', client.name, client.id),
         string.format('  Root directory: %s', vim.fn.fnamemodify(client.root_dir, ':~')),
         string.format('  Command: %s', cmd),
         string.format('  Offset encoding: %s', string.upper(client.offset_encoding)),
         string.format('  Settings: %s', vim.inspect(client.settings, { newline = '\n  ' })),
-        string.format('  Flags: %s', vim.inspect(client.flags, { newline = '\n  ' })),
         string.format(
           '  Attached buffers: %s',
           vim.iter(pairs(client.attached_buffers)):map(tostring):join(', ')
--- a/runtime/lua/vim/lsp/health.lua
+++ b/runtime/lua/vim/lsp/health.lua
@@ -33,16 +33,25 @@ local function check_active_clients()
   local clients = vim.lsp.get_clients()
   if next(clients) then
     for _, client in pairs(clients) do
-      local attached_to = table.concat(vim.tbl_keys(client.attached_buffers or {}), ',')
-      report_info(
+      local cmd ---@type string
+      if type(client.config.cmd) == 'table' then
+        cmd = table.concat(client.config.cmd --[[@as table]], ' ')
+      elseif type(client.config.cmd) == 'function' then
+        cmd = tostring(client.config.cmd)
+      end
+      report_info(table.concat({
+        client.name,
+        string.format('  ID: %d', client.id),
+        string.format('  Root directory: %s', vim.fn.fnamemodify(client.root_dir, ':~')),
+        string.format('  Command: %s', cmd),
+        string.format('  Offset encoding: %s', string.upper(client.offset_encoding)),
+        string.format('  Settings: %s', vim.inspect(client.settings, { newline = '\n  ' })),
+        string.format('  Flags: %s', vim.inspect(client.flags, { newline = '\n  ' })),
         string.format(
-          '%s (id=%s, root_dir=%s, attached_to=[%s])',
-          client.name,
-          client.id,
-          vim.fn.fnamemodify(client.root_dir, ':~'),
-          attached_to
-        )
-      )
+          '  Attached buffers: %s',
+          vim.iter(pairs(client.attached_buffers)):map(tostring):join(', ')
+        ),
+      }, '\n'))
     end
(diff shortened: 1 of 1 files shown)


MariaSolOs (approved):

runtime/lua/vim/lsp/health.lua
```diff
@@ -33,16 +33,25 @@ local function check_active_clients()
...
+        cmd = table.concat(client.config.cmd --[[@as table]], ' ')
+      elseif type(client.config.cmd) == 'function' then
+        cmd = tostring(client.config.cmd)
+      end
+      report_info(table.concat({
+        client.name,
+        string.format('  ID: %d', client.id),
+        string.format('  Root directory: %s', vim.fn.fnamemodify(client.root_dir, ':~')),
+        string.format('  Command: %s', cmd),
+        string.format('  Offset encoding: %s', string.upper(client.offset_encoding)),
```
mfussenegger, line 47: To make the representation a bit more compact I'd tend to:

- Remove the flags - they're usually not very important
- Move the id the into the same line as the name `clang (id: 1)`
- Only show offset encoding if it is either diverging from the default, or if there are mixed encodings across clients.

But also okay as is.

mfussenegger (approved):

runtime/lua/vim/lsp/health.lua
gpanders: >Only show offset encoding if it is either diverging from the default, or if there are mixed encodings across clients.

I think this would be more trouble than it's worth. You and I know that the default offset encoding is UTF-16, but the "average" user may not. So if one client says `Offset encoding: UTF-8` and another client doesn't say anything, I suspect the reaction will be "why is offset encoding not listed here?" rather than "oh this must just be using the default value".

gpanders: But maybe we can just remove it completely. Not sure how valuable it is as far as debugging information goes. I think it could be useful for someone to verify that their configuration is working as expected though ("did I set the `positionEncodings` capability correctly?")

mfussenegger: > "did I set the positionEncodings capability correctly?"

But that applies to any capability, no?

>  Which makes me wonder if we should be labeling this "Position encoding" rather than "Offset encoding"?

yes

> So if one client says Offset encoding: UTF-8 and another client doesn't say anything, I suspect the reaction will be "why is offset encoding not listed here?" rather than "oh this must just be using the default value".

Fair point. How about showing it everywhere if there are mixed encodings, and otherwise nowhere? (Given that currently mixed encodings are known to cause problems with some of the functions)

gpanders: >Fair point. How about showing it everywhere if there are mixed encodings, and otherwise nowhere? (Given that currently mixed encodings are known to cause problems with some of the functions)

Ok so I went ahead and just made a separate health check for this. We already show a warning when this happens, but now we have it in `:healthcheck` too.

```
vim.lsp: Position Encodings ~
- WARNING Found buffers attached to multiple clients with different position encodings.
  - Buffer 2: UTF-16 (client id(s): 2), UTF-8 (client id(s): 1)
  - ADVICE:
    - Use the positionEncodings client capability to ensure all clients use the same position encoding
```

runtime/lua/vim/lsp/health.lua
```diff
@@ -33,24 +33,30 @@ local function check_active_clients()
...
     for _, client in pairs(clients) do
-      local attached_to = table.concat(vim.tbl_keys(client.attached_buffers or {}), ',')
-      report_info(
+      local cmd ---@type string
+      if type(client.config.cmd) == 'table' then
+        cmd = table.concat(client.config.cmd --[[@as table]], ' ')
+      elseif type(client.config.cmd) == 'function' then
+        cmd = tostring(client.config.cmd)
+      end
+      report_info(table.concat({
```
dundargoc, line 42: giga-nitpick: is it just me or is `report_info` kinda wordy for a local variable? At that point one could just go with the full thing, or shorten it to `info`.

dundargoc:
I can't get "vim.lsp: Position Encodings" check to show up. I think we need to show the check even if it passes.

PR #28983, opened 2024-05-24
fix: "popcount" name conflict on NetBSD

mdehling:
The function popcount added in recent commit 037ea6e786b5d05f4a8965e4c2ba6aa60ec7c01a conflicts with a function by the same name in NetBSD's libc.  I'd like to propose the following change to avoid the issue.

### DIFF
--- a/src/nvim/math.c
+++ b/src/nvim/math.c
@@ -78,13 +78,15 @@ int xctz(uint64_t x)
 }
 
 /// Count number of set bits in bit field.
-int popcount(uint64_t x)
+unsigned xpopcount(uint64_t x)
 {
   // Use compiler builtin if possible.
-#if defined(__clang__) || defined(__GNUC__)
-  return __builtin_popcountll(x);
+#if defined(__NetBSD__)
+  return popcount64(x);
+#elif defined(__clang__) || defined(__GNUC__)
+  return (unsigned)__builtin_popcountll(x);
 #else
-  int count = 0;
+  unsigned count = 0;
   for (; x != 0; x >>= 1) {
     if (x & 1) {
       count++;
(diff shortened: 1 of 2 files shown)


luukvbaal:
TBH I don't think I was finished working on this and forgot this was even part of #28101 when I quickly reworked it to be included as experimental. Probably our function should be named `xpopcount()`, and then include an `#ifdef` for NetBSD that calls `popcount()`?

mdehling:
The problem is that the conflicting symbol breaks the build on NetBSD.

NetBSD defines these functions in string.h:

     unsigned int popcount(unsigned int value);
     unsigned int popcountl(unsigned long value);
     unsigned int popcountll(unsigned long long value);

and these in stdint.h

    unsigned int popcount32(uint32_t value);
    unsigned int popcount64(uint64_t value);

I think @luukvbaal's suggestion might be better than mine.

mdehling:
How about this version?

src/nvim/math.c
```diff
@@ -78,13 +78,15 @@ int xctz(uint64_t x)
 }
 
 /// Count number of set bits in bit field.
-int popcount(uint64_t x)
+unsigned xpopcount64(uint64_t x)
```
luukvbaal, line 81: I don't think we care to add the `64` to our helper function:
```suggestion
unsigned xpopcount(uint64_t x)
```

justinmk:
please add https://github.com/neovim/neovim/pull/28983#issuecomment-2130317588 to the commit message. see also https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md#commit-messages

mdehling:
Thanks for the comments, they should be addressed now.

PR #28986, opened 2024-05-24
refactor(lua): rewrite vim.highlight.range()

zeertzjq:
- Use getregionpos().
- Use a single extmark for non-blockwise selection.


justinmk (approved):

clason:
This introduced a regression with diagnostics: if diagnostics are at the end of the line, the underline for it now extends to the beginning of the next line. I haven't cooked up a minimal example yet, but you can observe this in the `lazy.nvim` update output:
Screenshot 2024-05-26 at 13 04 46

I think there still needs to be some distinction between single- and multiline extmarks (regarding the `hl_eol` option).

zeertzjq:
Oh, I forgot about the existence of `hl_eol`.

zeertzjq:
I don't end of line or `hl_eol` are relevant. I think #29022 should fix this.

clason:
Indeed it does, thank you!

PR #28991, opened 2024-05-25
refactor(eval): use vim.json inside f_json_encode

### LINKED ISSUE #28683: `json_encode()` assert false on c0 or c1 utf-8 chars
### Problem

When passing a c0 or c1 UTF-8 char (except for `\xc0\x80`) to `json_encode()` it will make an assert fail (if neovim is built with debug).

(c0 and c1 UTF-8 characters are the extended versions of the normal ascii characters and are considered invalid (though vim and neovim sometimes handle them fine))

### Steps to reproduce

```
nvim --clean
:echo json_encode("\xc0\x81")
```

### Expected behavior

Don't fail assert.

### Neovim version (nvim -v)

v0.10.0-dev-c18d794  (built with Debug)

no, vim 9.1

### Operating system/version

linux 6.8.9-arch1-2

### Terminal name/version

kitty 0.33.1

xterm-kitty

### Installation

build from repo

glepnir:
Problem: json_encode() uses a C template-based encoder that mishandles
certain invalid UTF-8 sequences (e.g. overlong encodings like \xc0\x81).

Solution: Replace the encoder path with lua_cjson (vim.json.encode).
Add a new typval_encode template (json_lua) in converter.c that performs
JSON validation (float, string, key, recursion checks) while pushing
values onto the Lua stack, then let lua_cjson serialize to JSON.

Fix #28683

### DIFF
--- a/src/nvim/eval/funcs.c
+++ b/src/nvim/eval/funcs.c
@@ -3881,7 +3881,23 @@ static void f_json_decode(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
 static void f_json_encode(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
 {
   rettv->v_type = VAR_STRING;
-  rettv->vval.v_string = encode_tv2json(&argvars[0], NULL);
+  rettv->vval.v_string = NULL;
+
+  MAXSIZE_TEMP_ARRAY(args, 1);
+  Arena arena = ARENA_EMPTY;
+  ADD_C(args, vim_to_object(&argvars[0], &arena, false));
+  Error err = ERROR_INIT;
+  Object result = NLUA_EXEC_STATIC("return vim.json.encode(...)", args, kRetObject, NULL, &err);
+
+  if (ERROR_SET(&err)) {
+    semsg(_("E474: %s"), err.msg);
+  } else if (result.type == kObjectTypeString) {
+    rettv->vval.v_string = xmemdupz(result.data.string.data, result.data.string.size);
+  }
+
+  api_free_object(result);
+  api_clear_error(&err);
+  arena_mem_free(arena_finish(&arena));
 }
 
 /// "keytrans()" function


justinmk:
> some error messages of the test case may require more detailed object handling?

json_encode/json_decode have very detailed error messages, thanks to ZyX's work. And those messages are more granular than Vim's, I believe. But we would likely need to make this tradeoff, if we're going to use cjson instead.

glepnir:
>  And those messages are more granular than Vim's, I believe.
 
that's true.  some converted formats may also need to be processed like space padding

```
Passed in:
(string) '{"d":[]}'
Expected:
(string) '{"d": []}'
```

justinmk:
> some converted formats may also need to be processed

Not sure what you mean, but just to be clear: unless proven otherwise, I think we should just use whatever messages come from cjson. This means tests will need to be updated. We shouldn't try to "convert" the cjson messages to match the old json_encode/json_decode messages. Unless that causes a big problem that I didn't foresee.

clason:
That is not a meaningful difference worth _any_ code to "fix"; just adapt the "expected" test to the new output.

glepnir:
ok I just wanted to confirm if this expected format is intentional to avoid breaking some behavior before i edit some expecteds

justinmk:
Is this WIP? Please mark as Draft if so. I would expect a lot of code to be deleted (that is the purpose of doing this)

test/old/testdir/test_functions.vim
```diff
@@ -2831,7 +2831,7 @@ func Test_call()
   function Mylen() dict
      return len(self.data)
   endfunction
-  let mydict = {'data': [0, 1, 2, 3], 'len': function("Mylen")}
+  let mydict = {'data': [0,1,2,3], 'len': function("Mylen")}
```
zeertzjq: Why does this need change?

test/old/testdir/test_blob.vim
```diff
@@ -707,7 +707,7 @@ endfunc
 func Test_blob_json_encode()
   let lines =<< trim END
       #" call assert_equal('[222,173,190,239]', json_encode(0zDEADBEEF))
```
zeertzjq, line 709: Duplicate of the line below
```suggestion
```

glepnir:
To preserve `_TYPE/_VAL` backward compat, I had to add a full `typval_encode` template instance in converter.c . The reason is `_TYPE` identification relies on C pointer comparison 
`type_di->vval.v_list == eval_msgpack_type_lists[kMPString]` — that identity is lost after `vim_to_object`, so `NLUA_EXEC_STATIC`  can't work. The template was the only way to resolve `_TYPE/_VAL` on the C side while pushing to Lua stack. Also needed a `__tojson` userdata hack for uint64 values that overflow lua doubles...Perhaps there's a better way that idk ? 🤔

If we're okay with dropping `_TYPE/_VAL` support in `json_encode`, the whole thing becomes just an NLUA_EXEC_STATIC call. this would be simpler ...

justinmk:
> If we're okay with dropping `_TYPE/_VAL` support in `json_encode`, the whole thing becomes just an NLUA_EXEC_STATIC call

I think we should drop that feature. What do others think?

justinmk:
> > If we're okay with dropping `_TYPE/_VAL` support in `json_encode`, the whole thing becomes just an NLUA_EXEC_STATIC call
> 
> I think we should drop that feature. What do others think?

still in favor of this.

glepnir:
could give it a try, and then decide after reviewing the cases where the tests failed.

justinmk:
Explored this more in https://github.com/neovim/neovim/pull/39181 . 

Maybe still worth pursuing, but will be a bit of work, and will lose the rather sophisticated roundtripping. But it's not clear if anyone actually uses that, since the roundtripping is mostly for vimscript<=>json, not lua<=>json.

PR #28997, opened 2024-05-25
feat(comment): insert mode command to put markers

tomtomjhj:
Motivation:
No easy way to start a comment with builtin commenting. Some commenting plugins provide commands to start a comment in the next line or at the end of line.
See also: https://github.com/neovim/neovim/discussions/28995.

Solution:
An insert mode command `<C-/>` to put comment markers around cursor. `o<C-/>` starts a comment in the next line, `A <C-/>` starts at the end of the line.

NOTE:
This default mapping overrides the builtin `CTRL-_` command for toggling revins.


runtime/lua/vim/_defaults.lua
```diff
@@ -147,6 +147,16 @@ do
       require('vim._comment').textobject()
     end
     vim.keymap.set({ 'o' }, 'gc', textobject_rhs, { desc = 'Comment textobject' })
+
+    local insert_rhs = function()
+      return require('vim._comment').insert()
+    end
+    vim.keymap.set(
+      { 'i' },
+      '<C-_>', -- <C-/> is registered as <C-_>
```
zeertzjq, line 156: This depends on the terminal emulator. You may need to map both.

XXiaoA:
Sorry to spoil the fun. I am wondering whether the idea and design have been discusssd deeply in other places. And, I dont think `c-/` is uncontroversial or widely recognised among (n)vim users.

runtime/lua/vim/_defaults.lua
```diff
@@ -147,6 +147,22 @@ do
       require('vim._comment').textobject()
     end
     vim.keymap.set({ 'o' }, 'gc', textobject_rhs, { desc = 'Comment textobject' })
+
+    local insert_rhs = function()
+      return require('vim._comment').insert()
```
przepompownia, line 152: `vim._comment` seems to be not public so user cannot rely on it if wants to map `insert` to another shortcut.

justinmk: If this PR is going to trigger requested for a public interface, let's just close this PR.

dundargoc: People are not mindreaders. It was probably put there because most other files are there. Why does `_default` live in `vim` for example?

wookayin:
I'm also not sure about this default option, but if I were to introduce any default keymap `ctrl-/` in the insert mode, I would prefer that it works like toggling (which is a common behavior found in other IDEs and editors).

tomtomjhj:
> I would prefer that it works like toggling (which is a common behavior found in other IDEs and editors) as gcc does.

with this, it's still not easy to add comment at the end of an existing line, e.g.,
```
code; /* comment */
```

tomtomjhj:
> wouldn't this be something an user can write in their config?

this is difficult unless `get_comment_parts` is exposed as public api

ribru17:
If this gets shot down, can we at least make `get_commentstring` public? It's totally fair to say no to additional niceties, though it is a shame that the current implementation is pretty rigid and extending it is only possible with duplicated code (through plugins or in config)

runtime/lua/vim/_defaults.lua
echasnovski: > I'm actually not sure why `_comment` lives on `vim.`.

I thought it was the common way to place the "private" Lua code. I used `vim._system` as a reference. If it is better be moved - no problem.

echasnovski:
I don't think adding comment markers needs a separate built-in mapping, especially with not tried and tested keys.

> If this gets shot down, can we at least make `get_commentstring` public?

I think using `vim.bo.commentstring` directly covers way above overwhelming majority of cases.

That said, either adding something like `vim.treesitter.get_option()` or `position` option for `vim.filetype.get_option()` which can get option value *at certain position* seems reasonable.

clason:
> That said, either adding something like vim.treesitter.get_option() or position option for vim.filetype.get_option() which can get option value at certain position seems reasonable.

Yes, but where would that option value come from, and how can we make sure that this value is consistent with how _other_ (builtin) functionality that relies on `commentstring` behaves? _That_ is the part of this that so far nobody has addressed and is blocking any further progress here.

echasnovski:
Thanks, @github-actions!

My "review" is [here](https://github.com/neovim/neovim/pull/28997#issuecomment-2137228964), as in "I don't think this functionality is a good fit for core".

PR #28998, opened 2024-05-25
fix(snippet): don't override unnamed register on tabstop select

mfussenegger:
Not sure if this is controversial. 

Some context of why I'd find not replacing the register more desirable: https://social.fussenegger.pro/@mathias/112500796424060656


MariaSolOs:
I don't play with my registers enough to have an opinion about this, but I'll approve since I don't think this will be harmful anyway.

@gpanders @clason @justinmk any thoughts?

MariaSolOs (approved):

justinmk (approved):
no objection, and I do play with my registers

PR #29001, opened 2024-05-25
feat(treesitter): support :horizontal mod for :EditQuery

mfussenegger:
Gives users more control over where to open the query window.

If you already have 2 or 3 splits open, adding another one can leave you
with too narrow columns and having it in a horizontal split is more
desirable.

### DIFF
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -1042,7 +1042,7 @@ add_predicate({name}, {handler}, {opts})
                      compatibility). This option will eventually become the
                      default and removed.
 
-edit({lang})                                     *vim.treesitter.query.edit()*
+edit({lang}, {opts})                             *vim.treesitter.query.edit()*
     Opens a live editor to query the buffer you started from.
 
     Can also be shown with *:EditQuery*.
@@ -1055,6 +1055,10 @@ edit({lang})                                     *vim.treesitter.query.edit()*
     Parameters: ~
       • {lang}  (`string?`) language to open the query editor for. If omitted,
                 inferred from the current buffer's filetype.
+      • {opts}  (`table?`) Optional keyword arguments:
+                • {wincmd}? (`string`) command used to create a new window.
+                  Defaults to 60vnew or new depending on inspect window
+                  presence.
 
 get({lang}, {query_name})                         *vim.treesitter.query.get()*
     Returns the runtime query {query_name} for {lang}.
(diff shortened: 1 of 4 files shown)


runtime/doc/treesitter.txt
```diff
@@ -1055,6 +1055,10 @@ edit({lang})                                     *vim.treesitter.query.edit()*
     Parameters: ~
       • {lang}  (`string?`) language to open the query editor for. If omitted,
                 inferred from the current buffer's filetype.
+      • {opts}  (`table?`) Optional keyword arguments:
+                • {wincmd}? (`string`) command used to create a new window.
```
justinmk, line 1059: not a blocker, just for reference:

I did a quick search and didn't find existing `wincmd` key for other cases (in lsp/ts/lua/diagnostics). And I think that's because we usually end up exposing this kind of thing as a callback (e.g. `on_list` for `:help vim.lsp.ListOpts`). E.g. this could be something like:

    {on_open}? (func (...): ...)

However that might complicate the implementation since we can't just pass it to the existing `vim.cmd()` call.

runtime/lua/vim/_defaults.lua
```diff
@@ -22,7 +22,11 @@ do
   end, { desc = 'Inspect treesitter language tree for buffer', count = true })
 
   vim.api.nvim_create_user_command('EditQuery', function(cmd)
-    vim.treesitter.query.edit(cmd.fargs[1])
+    local opts = {}
+    if cmd.smods.horizontal then
+      opts.wincmd = 'new'
+    end
```
wookayin, line 28: Can this also support other split modifiers like `:botright horizontal EditQuery`, if not too difficult, by combining with `smods.split`?

mfussenegger:
Based on the comments I'm thinking that maybe `vim.treesitter.query.edit` shouldn't create a new window at all, but rather just the buffer.
Then the `EditQuery` command can call `new`/`vnew` depending on the mods. And plugin which might call `vim.treesitter.query.edit` directly have even more control over the window creation, or can skip it altogether.

Thoughts?

runtime/doc/treesitter.txt
```diff
@@ -1055,6 +1055,10 @@ edit({lang})                                     *vim.treesitter.query.edit()*
     Parameters: ~
       • {lang}  (`string?`) language to open the query editor for. If omitted,
                 inferred from the current buffer's filetype.
+      • {opts}  (`table?`) Optional keyword arguments:
```
MariaSolOs, line 1058: For newly added `opts` parameters, I always prefer creating a Lua type for it to get that sweet completion support.

crwebb85:
> Based on the comments I'm thinking that maybe vim.treesitter.query.edit shouldn't create a new window at all, but rather just the buffer.

I think this is a good direction. I recently tried to create a keymap in my config to the editor buffer that adds the treesitter query results to the quick fix list. One of the pain points I had for this use case was the lack of control I had over the life cycle of the editor buffer.

PR #29003, opened 2024-05-25
fix(column): crash with 'signcolumn' set to "number"

### LINKED ISSUE #28984: Crash with signcolumn=number and numberwidth=1
### Problem

I caught a:
```
*** Error in `nvim': free(): invalid next size (normal): 0x000000000260b430 ***
```
and abort in neovim 0.10.0

I was able to use Valgrind to track down the location:
```
==29288== Invalid write of size 4
==29288==    at 0x4DE07A: draw_sign (drawline.c:468)
==29288==    by 0x4DE469: draw_lnum_col (drawline.c:553)
==29288==    by 0x4E1A5B: win_line (drawline.c:1547)
==29288==    by 0x4EBC1C: win_update (drawscreen.c:2286)
==29288==    by 0x4E73E5: update_screen (drawscreen.c:647)
==29288==    by 0x65995B: normal_redraw (normal.c:1353)
==29288==    by 0x659BE7: normal_check (normal.c:1455)
==29288==    by 0x733879: state_enter (state.c:40)
==29288==    by 0x657839: normal_enter (normal.c:518)
==29288==    by 0x60027B: main (main.c:664)
==29288==  Address 0x5e5c93c is 4 bytes before a block of size 1,096 alloc'd
==29288==    at 0x402D7BB: malloc (vg_replace_malloc.c:393)
==29288==    by 0x638A59: try_malloc (memory.c:98)
==29288==    by 0x638AD0: xmalloc (memory.c:132)
==29288==    by 0x5A8192: grid_alloc (grid.c:860)
==29288==    by 0x4E6686: default_grid_alloc (drawscreen.c:202)
==29288==    by 0x4E6A51: screen_resize (drawscreen.c:308)
==29288==    by 0x773A45: ui_refresh (ui.c:241)
==29288==    by 0x773FC6: ui_attach_impl (ui.c:401)
==29288==    by 0x486AA8: nvim_ui_attach (ui.c:194)
==29288==    by 0x474409: handle_nvim_ui_attach (dispatch_wrappers.generated.h:6155)
==29288==    by 0x6530A9: request_event (channel.c:407)
==29288==    by 0x55564E: multiqueue_process_events (multiqueue.c:149)
```

The problem is that, in `draw_sign`, the computed `sign_pos` can wind up being negative if `numberwidth=1` and `signcolumn=number`.

I _think_ this only happens with an extmark, not a normal sign.

### Steps to reproduce

Add this patch to provoke a crash if `sign_pos` ever winds up negative:
```diff
diff --git a/src/nvim/drawline.c b/src/nvim/drawline.c
index 283f7d9d6..39004d780 100644
--- a/src/nvim/drawline.c
+++ b/src/nvim/drawline.c
@@ -465,6 +465,7 @@ static void draw_sign(bool nrcol, win_T *wp, winlinevars_T *wlv, int sign_idx, i
     int fill = nrcol ? number_width(wp) + 1 : SIGN_WIDTH;
     draw_col_fill(wlv, schar_from_ascii(' '), fill, attr);
     int sign_pos = wlv->off - SIGN_WIDTH - (int)nrcol;
+    assert(sign_pos >= 0);
     linebuf_char[sign_pos] = sattr.text[0];
     linebuf_char[sign_pos + 1] = sattr.text[1];
   } else {
```
Then make a debug build (`CMAKE_BUILD_TYPE=Debug`) and run with:
```sh
vim --clean -c 'set number numberwidth=1 signcolumn=number' -c 'au VimEnter * call nvim_buf_set_extmark(1, nvim_create_namespace(""), 0, 0, {"sign_text": "X"})'
```

Note that it _doesn't_ reproduce without the VimEnter autocommand. If you just do `-c 'call nvim_buf_set_extmark(...)'` the number column winds up being grown to 2 spaces wide, instead. I think what's happening is that when the sign is placed asynchronously, the number column isn't being grown, and neovim is instead trying to draw a 2 character wide sign in a 1 character wide column, and winds up overwriting the byte before the start of the buffer.

### Expected behavior

No crash 😄

### Neovim version (nvim -v)

0.10.0 commit 27fb62988

n/a (no extmarks)

### Operating system/version

RHEL 7

### Terminal name/version

wsltty

screen-256color

### Installation

build from repo

luukvbaal:
Problem:  Numberwidth may depend on number of signs with text in the
          buffer and is not handled correctly for extmark signs.
Solution: Move legacy sign code for changed numberwidth so that it is
          handled properly for legacy and extmark signs alike.

Fix #28984

### DIFF
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -15,6 +15,7 @@
 #include "nvim/drawscreen.h"
 #include "nvim/extmark.h"
 #include "nvim/fold.h"
+#include "nvim/globals.h"
 #include "nvim/grid.h"
 #include "nvim/grid_defs.h"
 #include "nvim/highlight.h"
@@ -184,6 +185,21 @@ void buf_put_decor(buf_T *buf, DecorInline decor, int row, int row2)
   }
 }
 
+/// When displaying signs in the 'number' column, if the width of the number
+/// column is less than 2, then force recomputing the width after placing or
+/// unplacing the first sign in "buf".
+static void may_force_numberwidth_recompute(buf_T *buf, bool unplace)
+{
+  FOR_ALL_TAB_WINDOWS(tp, wp) {
+    if (wp->w_buffer == buf
+        && wp->w_minscwidth == SCL_NUM
+        && (wp->w_p_nu || wp->w_p_rnu)
+        && (unplace || wp->w_nrwidth_width < 2)) {
+      wp->w_nrwidth_line_count = 0;
+    }
+  }
+}
+
 static int sign_add_id = 0;
 void buf_put_decor_sh(buf_T *buf, DecorSignHighlight *sh, int row1, int row2)
 {
@@ -191,6 +207,7 @@ void buf_put_decor_sh(buf_T *buf, DecorSignHighlight *sh, int row1, int row2)
     sh->sign_add_id = sign_add_id++;
     if (sh->text[0]) {
       buf_signcols_count_range(buf, row1, row2, 1, kFalse);
+      may_force_numberwidth_recompute(buf, false);
     }
   }
 }
@@ -218,6 +235,7 @@ void buf_remove_decor_sh(buf_T *buf, int row1, int row2, DecorSignHighlight *sh)
       if (buf_meta_total(buf, kMTMetaSignText)) {
         buf_signcols_count_range(buf, row1, row2, -1, kFalse);
       } else {
+        may_force_numberwidth_recompute(buf, true);
         buf->b_signcols.resized = true;
         buf->b_signcols.max = buf->b_signcols.count[0] = 0;
       }
--- a/src/nvim/drawline.c
+++ b/src/nvim/drawline.c
@@ -465,6 +465,7 @@ static void draw_sign(bool nrcol, win_T *wp, winlinevars_T *wlv, int sign_idx, i
     int fill = nrcol ? number_width(wp) + 1 : SIGN_WIDTH;
     draw_col_fill(wlv, schar_from_ascii(' '), fill, attr);
     int sign_pos = wlv->off - SIGN_WIDTH - (int)nrcol;
+    assert(sign_pos >= 0);
     linebuf_char[sign_pos] = sattr.text[0];
     linebuf_char[sign_pos + 1] = sattr.text[1];
   } else {
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -1548,6 +1548,7 @@ static void win_update(win_T *wp)
   // Force redraw when width of 'number' or 'relativenumber' column changes.
   if (wp->w_nrwidth != nrwidth_new) {
     type = UPD_NOT_VALID;
+    changed_line_abv_curs_win(wp);
     wp->w_nrwidth = nrwidth_new;
   } else {
     // Set mod_top to the first line that needs displaying because of
--- a/src/nvim/sign.c
+++ b/src/nvim/sign.c
@@ -246,12 +246,6 @@ static int buf_delete_signs(buf_T *buf, char *group, int id, linenr_T atlnum)
     return FAIL;
   }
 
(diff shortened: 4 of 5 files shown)


luukvbaal:
ASAN crash only reproduces with `--clean` and `-c`, not when exchanging either of those with e.g. `-u NONE` and `--cmd`. I guess that further suggests that this is just a startup issue; maybe we can add some logic there instead...

godlygeek:
Setting aside my minimal reproducer, I definitely encountered this bug outside of startup: the buffer loaded, then an LSP client processed the buffer and added a mark (and corrupted the heap, because of this bug), and then I got a crash (assertion failure) on a future call to `free`.

luukvbaal:
Good to know, I can try to add a different test in that case.

godlygeek:
Perhaps just swapping the VimEnter for a CursorHold?

luukvbaal:
OK yeah I was just misled by the involved repro in the linked issue. Exchanged for a simpler test that crashes on master with an ASAN build.

justinmk (approved):

src/nvim/decoration.c
```diff
@@ -191,6 +192,16 @@ void buf_put_decor_sh(buf_T *buf, DecorSignHighlight *sh, int row1, int row2)
     sh->sign_add_id = sign_add_id++;
     if (sh->text[0]) {
       buf_signcols_count_range(buf, row1, row2, 1, kFalse);
+      // Since number_width() is a cached value, it may not be updated accordingly after
+      // adding the first sign so do it here in case it's smaller than 2 and 'signcolumn'
+      // is set to "number" #28984.
+      if (buf_meta_total(buf, kMTMetaSignText) == 1) {
+        FOR_ALL_WINDOWS_IN_TAB(wp, curtab) {
```
zeertzjq, line 199: Should this also be done for other tabpages?

luukvbaal: It seems so. I assumed the nrwidth cache would be invalidated upon switching tabpages but that doesn't seem to be the case.

There is also (currently `static`) `may_force_numberwidth_recompute()` which could be re-used here.

luukvbaal: Updated with a different solution that handles this properly.

src/nvim/decoration.c
```diff
@@ -184,13 +185,28 @@ void buf_put_decor(buf_T *buf, DecorInline decor, int row, int row2)
...
+/// unplacing the first sign in "buf".
+static void may_force_numberwidth_recompute(buf_T *buf, int unplace)
+{
+  FOR_ALL_TAB_WINDOWS(tp, wp)
+  if (wp->w_buffer == buf
+      && wp->w_minscwidth == SCL_NUM
+      && (wp->w_p_nu || wp->w_p_rnu)
+      && (unplace || wp->w_nrwidth_width < 2)) {
+    wp->w_nrwidth_line_count = 0;
+  }
```
zeertzjq, line 199:
```suggestion
  FOR_ALL_TAB_WINDOWS(tp, wp) {
    if (wp->w_buffer == buf
        && wp->w_minscwidth == SCL_NUM
        && (wp->w_p_nu || wp->w_p_rnu)
        && (unplace || wp->w_nrwidth_width < 2)) {
      wp->w_nrwidth_line_count = 0;
    }
  }
```

src/nvim/decoration.c
```diff
@@ -184,13 +185,29 @@ void buf_put_decor(buf_T *buf, DecorInline decor, int row, int row2)
   }
 }
 
+/// When displaying signs in the 'number' column, if the width of the number
+/// column is less than 2, then force recomputing the width after placing or
+/// unplacing the first sign in "buf".
+static void may_force_numberwidth_recompute(buf_T *buf, int unplace)
```
zeertzjq, line 191:
```suggestion
static void may_force_numberwidth_recompute(buf_T *buf, bool unplace)
```

zeertzjq (approved):

PR #29012, opened 2024-05-26
fix(tohtml): properly handle multiple hl groups

ribru17:
Problem: :TOhtml doesn't properly handle virtual text when it has multiple highlight groups. It also improperly calculates position offset for multi-byte virt_text characters.

Solution: Apply the `vim.api.nvim_strwidth` broadly to properly calculate character offset, and handle the case where the `hl` argument is a table of multiple hl groups.

### DIFF
--- a/runtime/lua/tohtml.lua
+++ b/runtime/lua/tohtml.lua
@@ -168,6 +168,8 @@ local background_color_cache = nil
 --- @type string?
 local foreground_color_cache = nil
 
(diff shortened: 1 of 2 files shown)


ribru17:
Still need to add tests. Both problems solved in this PR can be noticed when using the indent blankline plugin

justinmk (approved):
I see a +1 from @altermo so I assume that means we can merge this. Thanks!

PR #29018, opened 2024-05-26
build(deps): bump tree-sitter parsers

clason:
- **build(deps): bump tree-sitter-query to v0.4.0**
- **build(deps): bump tree-sitter-c to v0.21.3**

Query change is breaking for the downstream format-queries script (which defaults to the bundled parser), which is why I would like this change to be in stable as well. (C changes are grammar fixes that don't affect queries, but might as well backport them to prevent surprises when distros package and upgrade the parser.)


dundargoc (approved):
Does this count as breaking change?

clason:
Unclear. I would say "no", since queries and parsers should always be installed together. I'd rather say that this "unbreaks" the downstream script.

PR #29021, opened 2024-05-26
TOhtml fixes

altermo:
I looked through the code and noticed a few bugs and other unwanted behavior:
+ There was an `ipairs()` which should have been a `pairs()`
+ Inlay hints work similarly to semantic tokens, and thus should not be shown
+ When there are multiple notifications, only the last one is shown, so instead add a `(+ %d more warnings)` to the last one.

### DIFF
--- a/runtime/lua/tohtml.lua
+++ b/runtime/lua/tohtml.lua
@@ -481,8 +481,15 @@ end
 
 --- @param state vim.tohtml.state
 --- @param extmark {[1]:integer,[2]:integer,[3]:integer,[4]:vim.api.keyset.set_extmark|any}
-local function _styletable_extmarks_virt_text(state, extmark)
+--- @param namespaces table<integer,string>
+local function _styletable_extmarks_virt_text(state, extmark, namespaces)
   if not extmark[4].virt_text then
+    return
+  end
+  ---TODO(altermo) LSP semantic tokens (and some other extmarks) are only
+  ---generated in visible lines, and not in the whole buffer.
+  if (namespaces[extmark[4].ns_id] or ''):find('vim_lsp_inlayhint') then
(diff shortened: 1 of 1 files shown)


runtime/lua/tohtml.lua
```diff
@@ -611,9 +636,7 @@ local function styletable_folds(state)
     end
   end
   if has_folded and type(({ pcall(vim.api.nvim_eval, vim.o.foldtext) })[2]) == 'table' then
-    vim.notify_once(
-      'Info(TOhtml): foldtext returning a table is half supported, HTML may be incorrect'
-    )
+    notify('foldtext returning a table is half supported, HTML may be incorrect')
```
clason, line 639: What does "half supported" mean? Either it is supported, or not ;) (In other words, this message should be consistent with the others.)

altermo: notification updated

runtime/lua/tohtml.lua
```diff
@@ -57,6 +57,26 @@
...
+--- @type string[]
+local notifications = {}
+
+---@param msg string
+local function notify(msg)
+  if #notifications == 0 then
+    vim.schedule(function()
+      if #notifications > 1 then
+        vim.notify(
+          ('TOhtml: %s (+ %d more warnings)'):format(notifications[1], tostring(#notifications - 1))
```
justinmk, line 69: Maybe vim.notify should do this by default. Eventually `:echo` and all messages will be reworked to behave something like this (with sufficient affordances for seeing the full message).

justinmk (approved):

wookayin (commented):
LGTM too. My review was too late (started writing before merged), but leaving some nit/minor comment for future recommendations/improvements.

PR #29022, opened 2024-05-26
fix(lua): don't highlight empty region

zeertzjq:


clason (approved):

clason:
That fixes my issue, but isn't this throwing out the baby with the bathwater?

PR #29028, opened 2024-05-26

clason:
- **vim-patch:76174e7: runtime(asm): remove the indent plugin since it has too many issues**
- **vim-patch:8607192: runtime(typescriptreact): fix highlighting nested and escaped quotes in string props (vim/vim#14852)**


runtime/indent/asm.vim
zeertzjq, line 1: Delete this file.

clason: That will teach me to rely on the script.

zeertzjq (approved):

PR #29050, opened 2024-05-27
fix(lsp): clear lsp client diagnostics

### LINKED ISSUE #29048: LSP diagnostics not cleared on client stop for unattached buffers
### Problem

Stopping an LSP client does not clear the diagnostics for unattached buffers. 

1. Create the file structure with the following files
```
./
  minimal_init.lua
  lsp_test/
    bad_lua1.lua
    bad_lua2.lua
    .luarc.json
```

minimal_init.lua
```lua
local cmd = { "lua-language-server" }

local root_markers = { ".luarc.json" }

	pattern = "lua",
	callback = function(args)
		vim.lsp.start({
			name = "lua_ls",
			cmd = cmd,
			root_dir = root_dir,
			settings = settings,
		})
		vim.notify("Lsp root_dir: " .. root_dir)
	end,
})

--Notify when Lsp is attached
vim.api.nvim_create_autocmd("LspAttach", {
	group = vim.api.nvim_create_augroup("lsp_attach", { clear = true }),
	desc = "lsp on_attach",
	callback = function(event)
		vim.notify("Attached Lsp client " .. event.data.client_id)
	end,
})

vim.api.nvim_create_user_command("Test", function(_)
	vim.diagnostic.setqflist()
end, {
	desc = "Set qf list with diagnostics",
})

vim.api.nvim_create_user_command("LspStop", function(_)
	for _, client in ipairs(vim.lsp.get_clients()) do
		client.stop()
	end
end, {
	desc = "Stops all lsp clients",
})

vim.api.nvim_create_user_command("LspStop", function(_)
	for _, client in ipairs(vim.lsp.get_clients()) do
		client.stop()
	end
end, {
	desc = "Stops all lsp clients",
})

vim.api.nvim_create_user_command("LspStart", function(_)
	local current_buf_file_name = vim.api.nvim_buf_get_name(0)

	local match = vim.fs.find(root_markers, { path = current_buf_file_name, upward = true })[1]
	vim.lsp.start({
		name = "lua_ls",
		cmd = cmd,
		root_dir = root_dir,
		settings = settings,
	})
	vim.notify("Lsp lua_ls started at root_dir: " .. root_dir)
end, {
	desc = "Start lua_ls lsp client",
})
```

bad_lua1.lua
```lua
---@type string
local a = 5
```

bad_lua2.lua
```lua
---@type integer
local b = "definitely an integer" 
```

.luarc.json
```json
{
  "$schema": "https://raw.githubusercontent.com/sumneko/vscode-lua/master/setting/schema.json",
  "diagnostics": {
    "groupFileStatus": {
      "type-check": "Any"
    }
  }
}
```

2. From the directory with the minimal_init.lua file, run `nvim ./lsp_test -u ./minimal_init.lua`.
3. Using netrw, open file `bad_lua1.lua`.
4. Wait for message notification that the lua language server started.
5. Run command `:Test` to open quickfix list with the lsp diagnostics. 
6. Visually verify that quick fix list has diagnostics for both the file `bad_lua1.lua` and `bad_lua2.lua`.
7. Run command `:LspStop`.
8. Run command `:Test` to open quickfix list with the new list of lsp diagnostics. 
9. Verify that only diagnostics for `bad_lua2.lua` remain in the quick fix list.

### Expected behavior

On the final step, the quickfix list should be empty instead of listing diagnostics for the file `bad_lua2.lua`.

### Neovim version (nvim -v)

v0.11.0-dev

3.7.4

### Operating system/version

Ubuntu 22.04.2 LTS

### Log file

_No response_

crwebb85:
Fixes neovim/neovim#29048

Problem: When an lsp client is stopped, the client will only clear the diagnostics for the attached buffers but not the unattached buffers.
Solution: Reset the diagnostics for the whole namespace rather than for only the attached buffers.

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -391,8 +391,8 @@ end
 local function on_client_exit(code, signal, client_id)
   local client = all_clients[client_id]
 
-  for bufnr in pairs(client.attached_buffers) do
-    vim.schedule(function()
+  vim.schedule(function()
+    for bufnr in pairs(client.attached_buffers) do
       if client and client.attached_buffers[bufnr] then
         api.nvim_exec_autocmds('LspDetach', {
           buffer = bufnr,
@@ -401,15 +401,16 @@ local function on_client_exit(code, signal, client_id)
         })
       end
 
-      local namespace = vim.lsp.diagnostic.get_namespace(client_id)
-      vim.diagnostic.reset(namespace, bufnr)
       client.attached_buffers[bufnr] = nil
 
       if #lsp.get_clients({ bufnr = bufnr, _uninitialized = true }) == 0 then
         reset_defaults(bufnr)
       end
-    end)
-  end
+    end
+
+    local namespace = vim.lsp.diagnostic.get_namespace(client_id)
+    vim.diagnostic.reset(namespace)
+  end)
 
   local name = client.name or 'unknown'
 


runtime/lua/vim/lsp.lua
```diff
@@ -401,7 +402,6 @@ local function on_client_exit(code, signal, client_id)
         })
       end
 
-      local namespace = vim.lsp.diagnostic.get_namespace(client_id)
       vim.diagnostic.reset(namespace, bufnr)
```
justinmk, line 405: why doesn't this work? surely the purpose (at least) of creating a per-client namespace is to be able to clear or reset the related diagnostics without a bunch of extra code.

justinmk:
> the client will only clear the diagnostics for the attached buffers but not the unattached buffers.

why would unattached buffers have diagnostics? and if they do, why aren't those diagnostics linked to the per-client namespace?

mfussenegger:
Do you by chance use a server that supports pull diagnostics? I could imagine the problem you're facing is due to lsp using two namespaces per client. One for push diagnostic, one for pull. Only one gets cleared

runtime/lua/vim/lsp.lua
crwebb85: Actually, I think you are correct. I reread the docs for vim.diagnostic.reset and it does say when the buffer number is omitted, diagnostics are removed for all buffers. I will test if that works as that would be much cleaner.

crwebb85:
The main reason that the lsp can have diagnostics without the buffer attached is that the lsp finds all diagnostics for the workspace. You normally don't open all the workspace files in the buffer so diagnostics need a bufnr to be asigned to. From digging into the code, it seems to me that when the lsp reports a diagnostic for an unopened buffer, the bufnr gets reserved for if the you later open the file and that bufnr is what is used in the diagnostic cache.

runtime/lua/vim/lsp.lua
crwebb85: You were correct vim.diagnostic.reset does have the ability to clear all diagnostics on the namespace without needing to manually iterate over all the buffers. I cleaned up the code with that fix.

crwebb85:
> Do you by chance use a server that supports pull diagnostics? I could imagine the problem you're facing is due to lsp using two namespaces per client. One for push diagnostic, one for pull. Only one gets cleared

I believe the lsp's that I have experienced this with are push diagnostics and it didn't seem from my testing that it was using two different namespaces. The lsp's I have confirmed this behavior with is lua-language-server and omnisharp. Although, I can only get lua-language-server to consistently find workspace diagnostics when I have a .luarc.json file.

crwebb85:
> @crwebb85 you have a failing test.

I took a quick look during lunch but nothing obvious stuck out to me so I will setup a debugger later tonight to see why the test fails.

runtime/lua/vim/lsp.lua
```diff
@@ -411,6 +409,11 @@ local function on_client_exit(code, signal, client_id)
     end)
   end
 
+  vim.schedule(function()
+    local namespace = vim.lsp.diagnostic.get_namespace(client_id)
+    vim.diagnostic.reset(namespace)
+  end)
```
icholy, line 415: I don't think you need a separate `vim.schedule` for this, it can go in the callback on line 418

mfussenegger: I think the above vim.schedule and loop could also be inverted:

```lua
  vim.schedule(function()
    for bufnr in pairs(client.attached_buffers) do
      if client and client.attached_buffers[bufnr] then
        api.nvim_exec_autocmds('LspDetach', {
          buffer = bufnr,
          modeline = false,
          data = { client_id = client_id },
        })
      end

      client.attached_buffers[bufnr] = nil

      if #lsp.get_clients({ bufnr = bufnr, _uninitialized = true }) == 0 then
        reset_defaults(bufnr)
      end
    end
    local namespace = vim.lsp.diagnostic.get_namespace(client_id)
    vim.diagnostic.reset(namespace)
  end)
```

crwebb85: Good point. I moved it to the other `vim.schedule`.

crwebb85: @mfussenegger actually, I like that better but I have to questions if I do that I should probaly also combine it with the final vim.schedule like this

```lua
  vim.schedule(function()
    for bufnr in pairs(client.attached_buffers) do
      if client and client.attached_buffers[bufnr] then
        api.nvim_exec_autocmds('LspDetach', {
          buffer = bufnr,
          modeline = false,
          data = { client_id = client_id },
        })
      end

      client.attached_buffers[bufnr] = nil

      if #lsp.get_clients({ bufnr = bufnr, _uninitialized = true }) == 0 then
        reset_defaults(bufnr)
      end
    end
    local namespace = vim.lsp.diagnostic.get_namespace(client_id)
    vim.diagnostic.reset(namespace)

    all_clients[client_id] = nil

    -- Client can be absent if executable starts, but initialize fails
    -- init/attach won't have happened
    if client then
      changetracking.reset(client)
    end
    if code ~= 0 or (signal ~= 0 and signal ~= 15) then
      local msg = string.format(
        'Client %s quit with exit code %s and signal %s. Check log for errors: %s',
        name,
        code,
        signal,
        lsp.get_log_path()
      )
      vim.notify(msg, vim.log.levels.WARN)
    end
  end)
``` 
I will check to see if that change passes the tests. If there isn't any reason we shouldn't combine all of these `vim.schedule`s, I will go ahead and make that change.

mfussenegger: Afaik the split into at least two schedule's is important to ensure the `LspDetach` events finished processing before the remainder of the cleanup runs.

crwebb85: Although reading this comment before the last vim.schedule make me wonder if combining with the last vim.schedule could break autocommands using LspDetach. Any thoughts on if that is a possibility?

```  
 -- Schedule the deletion of the client object so that it exists in the execution of LspDetach
  -- autocommands
```

crwebb85: > I will check to see if that change passes the tests.

Changing the code to below does pass the existing tests but I still am a bit concerned about if LspDetach autocommands will have access to the client so I am going to test it with my `minimal_init.lua` to see if it is actually safe. If it isn't I will switch to @mfussenegger original suggestion.

```lua
--- @param code integer
--- @param signal integer
--- @param client_id integer
local function on_client_exit(code, signal, client_id)
  local client = all_clients[client_id]

  -- Schedule the deletion of the client object so that it exists in the execution of LspDetach
  -- autocommands
  vim.schedule(function()
    for bufnr in pairs(client.attached_buffers) do
      if client and client.attached_buffers[bufnr] then
        api.nvim_exec_autocmds('LspDetach', {
          buffer = bufnr,
          modeline = false,
          data = { client_id = client_id },
        })
      end

      client.attached_buffers[bufnr] = nil

      if #lsp.get_clients({ bufnr = bufnr, _uninitialized = true }) == 0 then
        reset_defaults(bufnr)
      end
    end

    local name = client.name or 'unknown'

    all_clients[client_id] = nil

    -- Client can be absent if executable starts, but initialize fails
    -- init/attach won't have happened
    if client then
      changetracking.reset(client)
    end
    if code ~= 0 or (signal ~= 0 and signal ~= 15) then
      local msg = string.format(
        'Client %s quit with exit code %s and signal %s. Check log for errors: %s',
        name,
        code,
        signal,
        lsp.get_log_path()
      )
      vim.notify(msg, vim.log.levels.WARN)
    end
  end)
end
```

crwebb85: I added the below LspDetach autocmd to my `minimal_init.lua` and got the same response "Client lua_ls is stopped" for the version of neovim with my code changes and also my unmodified version neovim on my system, so I believe it is a safe change.
```lua
vim.api.nvim_create_autocmd("LspDetach", {
	group = vim.api.nvim_create_augroup("lsp_detach", { clear = true }),
	desc = "lsp on_detach",
	callback = function(event)
		vim.notify("Detached Lsp client " .. event.data.client_id)
		local client = vim.lsp.get_client_by_id(event.data.client_id)

		if client == nil then
			vim.notify("Client is nil")
		elseif client.is_stopped() then
			vim.notify("Client " .. client.name .. " is stopped")
		else
			vim.notify("Client " .. client.name .. " is not stopped")
		end
	end,
})
```

crwebb85: I want to test one more thing but will need to test it after work so I will convert this back to a draft PR.

mfussenegger: I'd still keep at least two schedule's for the backport to 0.10 to avoid risking a regression.
We can follow up with a refactor that puts all in one for 0.11

crwebb85: Sounds good. I updated it to just use two schedules and moved the diagnostic reset up to the top one as it makes more since to be with that section of the code.

wookayin: I think this `vim.schedule` hack (?) should be commented, briefly talking about why we needed to defer its execution.

icholy: @wookayin I think we can just make a followup PR right now to combine it all into one `vim.schedule` instead. @crwebb85 are you interested in doing that?

TilakMaddy:
Doesn't the LSP spec put the owness on the LSP server to clear the diagnostics (by sending an empty arrray of diagnostics to the associated file URI of which, you want to clear)?

MariaSolOs:
> Doesn't the LSP spec put the oweness on the LSP server to clear the diagnostics (by sending an empty array of diagnostics to the associated file URI of which, you want to clear)?

I think that this depends on the diagnostics model (push versus pull diagnostics).

PR #29051, opened 2024-05-27
Fix broken vim.keymap.del error

mattfbacon:
I would get an error saying that the mapping doesn't exist. Weird because the code seems to specifically check that the mapping exists but regardless pcall will suppress it.


clason:
Sorry, but this is just sweeping the mess under the rug. We beed to find out why the error is triggered and fix that.

mattfbacon:
OK, any thoughts? I figured the check that's already there would be enough but clearly not.

clason:
Open an issue with minimal reproducing steps, then we can take a look at it.

wookayin:
Guessing from the code: maparg() returns a global keymap but we are trying to delete a (non-existent) buffer local keymap. Probably need to check the maparg result has buffer == true.

MariaSolOs:
> Probably need to check if the maparg result has buffer == true.

It should be `buffer == 1` btw.

PR #29071, opened 2024-05-28
feat(lsp): support postfix snippet completion; use fuzzy match on filterText 

mfussenegger:
First commit:

The `complete()` mechanism matches completion candidates against
the typed text, so strict pre-filtering isn't necessary.

This is a first step towards supporting postfix snippets (like
`items@insert` in luals)

---

Second commit:

Changes the approach on how `word` is inferred and enables support for postfix snippets

There are two goals here:

1. The `word` should match what the user started typing, so that vim.fn.complete() doesn't
   filter it away, preventing snippet expansion

   For example, if they type `items@ins`, luals returns `table.insert(items, $0)` as
   `textEdit.newText` and `insert` as label.
   There would be no prefix match if `textEdit.newText` is used as `word`

2. If users do not expand a snippet, but continue typing, they should see a somewhat reasonable
   `word` getting inserted.

   For example in:

       insertText: "testSuites ${1:Env}"
       label: "testSuites"

   `testSuites` should have priority as `word`, as long as the full snippet gets expanded on accept (`<c-y>`)

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -201,11 +201,17 @@ function M._lsp_to_complete_items(result, prefix, client_id)
     return {}
   end
 
-  local function matches_prefix(item)
-    return vim.startswith(get_completion_word(item), prefix)
-  end
+  if prefix ~= '' then
+    ---@param item lsp.CompletionItem
+    local function match_prefix(item)
+      if item.filterText then
+        return next(vim.fn.matchfuzzy({ item.filterText }, prefix))
+      end
+      return true
+    end
 
-  items = vim.tbl_filter(matches_prefix, items) --[[@as lsp.CompletionItem[]|]]
+    items = vim.tbl_filter(match_prefix, items) --[[@as lsp.CompletionItem[]|]]
+  end
   table.sort(items, function(a, b)
     return (a.sortText or a.label) < (b.sortText or b.label)
   end)
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -129,18 +129,33 @@ end
 --- @param item lsp.CompletionItem
 --- @return string
 local function get_completion_word(item)
-  if item.textEdit ~= nil and item.textEdit.newText ~= nil and item.textEdit.newText ~= '' then
-    if item.insertTextFormat == protocol.InsertTextFormat.PlainText then
-      return item.textEdit.newText
-    else
-      return parse_snippet(item.textEdit.newText)
-    end
-  elseif item.insertText ~= nil and item.insertText ~= '' then
-    if item.insertTextFormat == protocol.InsertTextFormat.PlainText then
-      return item.insertText
-    else
+  if item.insertTextFormat == protocol.InsertTextFormat.Snippet then
+    if item.textEdit then
+      -- Use label instead of text if text has different starting characters.
+      -- label is used as abbr (=displayed), but word is used for filtering
+      -- This is required for things like postfix completion.
+      -- E.g. in lua:
+      --
+      --    local f = {}
+      --    f@|
+      --      ▲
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/lsp/completion.lua
```diff
@@ -129,18 +129,33 @@ end
...
+      --    local f = {}
+      --    f@|
+      --      ▲
+      --      └─ cursor
+      --
+      --    item.textEdit.newText: table.insert(f, $0)
+      --    label: insert
+      --
+      -- Typing `i` would remove the candidate because newText starts with `t`.
+      local text = item.insertText or item.textEdit.newText
```
MariaSolOs, line 148: Shouldn't this just be `item.textEdit.newText`? If I'm reading the specs correctly `textEdit` has priority over `insertText`.

MariaSolOs, line 148: Shouldn't this just be `item.textEdit.newText`? If I'm reading the specs correctly `textEdit` has priority over `insertText`.

MariaSolOs (approved):

runtime/lua/vim/lsp/completion.lua
mfussenegger: `item.textEdit.newText` still gets used if the snippet is accepted. The `insertText` is used because it looks as if this is the better choice for the `word` (which will get prefix matches by `complete()`)

For example, in jdtls you get snippet candidates to implement methods from a base class or interface.

Say you have an `AstVisitor` class with various `visitXY` methods, and you implement one.
If you type `visit|` you'll get completion candidates for all the various methods, where an individual item can look like this:

```
  detail: Override method in 'AstVisitor'
  filterText: @Override\npublic Symbol visitAssignment(Assignment<?> node, ExpressionAnalysisContext context) {\n    ${0:// TODO Auto-generated method stub\n    return super.visitAssignment(node, context);}\n}
  insertText: visitAssignment
  insertTextFormat: 2
  kind: 2
  label: visitAssignment(Assignment<?> node, ExpressionAnalysisContext context) : Symbol
  sortText: 999999163
  textEdit: table: 0x7a70f453e628
    newText: @Override\npublic Symbol visitAssignment(Assignment<?> node, ExpressionAnalysisContext context) {\n    ${0:// TODO Auto-generated method stub\n    return super.visitAssignment(node, context);}\n}
    range: table: 0x7a70f47f5bf8
      end: table: 0x7a70f47f5c40
        character: 12
        line: 457
      start: table: 0x7a70f40a3be0
        character: 8
        line: 457
  textEditText: @Override\npublic Symbol visitAssignment(Assignment<?> node, ExpressionAnalysisContext context) {\n    ${0:// TODO Auto-generated method stub\n    return super.visitAssignment(node, context);}\n}
```

Using `textEdit.newText` as word would mean that the entry would disappear, because `visit|` doesn't share the prefix with `@Override...`

Granted, this might be over-fitting towards how `jdtls` represents things and I expect that we'll get issues about other language servers. But I'm fairly confident that we can keep tweaking these heuristics until we get a good experience cross languages. E.g. another option is to choose whatever has a prefix match (and is the shortest).

mfussenegger: Another option would be to set `equal = 1` in the items for `complete()`, then it would no longer apply the prefix match and not discard the wrong entries. The downside is that if you don't expand the snippet, you get a less useful `word` added. For example it wouldn't handle newlines well.

It would also mean that we'd have to do manual re-filtering after each character - which means either also invoking a new request, or cache the entries in the `Context`.

MariaSolOs: That sounds reasonable, I'm okay with the current approach then.

> Another option would be to set equal = 1 in the items for complete()

Doesn't that affect fuzzy matching?

MariaSolOs:
As discussed offline, I believe we want to put fuzzy matching behind a completion option.

PR #28981, opened 2024-05-24
feat(checkhealth): allow folding

dundargoc:
Closes https://github.com/neovim/neovim/issues/22796.

Reference:
https://github.com/neovim/neovim/pull/22866
https://github.com/neovim/neovim/pull/28232

PR #29072, opened 2024-05-28
feat(checkhealth): show summary of errors and warnings at top [skip ci]

### LINKED ISSUE #22796: checkhealth: fold "OK" (successful) sections by default

dundargoc:
This way it's easy for user to directly see the most important
information first. If they then wish to know more information they can
read in more detail further down in the checkhealth.

Alternative to https://github.com/neovim/neovim/pull/28981.
Closes https://github.com/neovim/neovim/issues/22796.

Reference:
https://github.com/neovim/neovim/pull/22866
https://github.com/neovim/neovim/pull/28232

### DIFF
--- a/runtime/lua/vim/health.lua
+++ b/runtime/lua/vim/health.lua
@@ -89,6 +89,8 @@
 local M = {}
 
 local s_output = {} ---@type string[]
+local checks = {}
+local current_name ---@type string
 
 -- From a path return a list [{name}, {func}, {type}] representing a healthcheck
 local function filepath_to_healthcheck(path)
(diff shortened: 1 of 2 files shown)


yochem:
This might not be necessary anymore as section headers show a summary now (#33388)

PR #29074, opened 2024-05-28
docs: allow "Neovim" and "neovim"

dundargoc:
No one cares about the correct usage between nvim, Nvim, Neovim, or
neovim. They are all used interchangably.


dundargoc:
Also makes writinf docs super annoying in cases where the usage of Neovim is correct

zeertzjq:
> Also makes writinf docs super annoying in cases where the usage of Neovim is correct

What cases?

dundargoc:
All the cases where it is correct.

Does listing these somehow change your mind about anything?

scripts/gen_help_html.lua
```diff
@@ -35,9 +35,7 @@ local invalid_links = {} ---@type table<string, any>
 local invalid_urls = {} ---@type table<string, any>
 local invalid_spelling = {} ---@type table<string, table<string, string>>
 local spell_dict = {
-  Neovim = 'Nvim',
   NeoVim = 'Nvim',
-  neovim = 'Nvim',
```
gpanders, line 40: What is the case for allowing `neovim`? When is it ever correct to use `neovim` (as opposed to "Neovim", which at least has correct capitalization)?

dundargoc: I don't see why both can't be correct

wookayin:
> `master`/runtime%2Fdoc%2Fbackers.txt#L10

https://github.com/neovim/neovim/blob/v0.10.0/scripts/gen_help_html.lua#L51

We have this. Wouldn't it suffice?

PR #29080, opened 2024-05-29
Remove deprecated rustc arguments

kaffarell:
Currently the :RustExpand command doesn't work because it uses non-existing arguments.
The `--pretty` and `--xpretty` arguments have been deprecated [0] and
removed [1]. They have been renamed to '-Zunpretty' and moved to
nightly [2]. This means that currently, with the default stable compiler
this won't work. You will have to use nightly or wait until it is
stabilized (which is an ongoing effort [3]).

[0]: https://github.com/rust-lang/rust/pull/21441
[1]: https://github.com/rust-lang/rust/pull/83491
[2]: https://github.com/rust-lang/rust/pull/27392
[3]: https://github.com/rust-lang/rust/issues/43364

### DIFF
--- a/runtime/autoload/rust.vim
+++ b/runtime/autoload/rust.vim
@@ -144,14 +144,8 @@ endfunction
 function! s:Expand(dict, pretty, args)
     try
         let rustc = exists("g:rustc_path") ? g:rustc_path : "rustc"
-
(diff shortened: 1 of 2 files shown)


zeertzjq:
This should go to https://github.com/rust-lang/rust.vim

PR #29085, opened 2024-05-29
feat: change 'commentstring' to `// %s` for C/C++

### LINKED ISSUE #29079: Default c/cpp commentstring
### Problem

The new neovim commenting feature is great. However, with the new commenting feature, the program breaks on commenting lines containing comments for c/cpp programs. For example consider

```
int a = 1;
/*int b = 2;*/
int c = 3;
````

Commenting these together by selecting using 'V' would lead to

```
/*int a = 1;*/
/*/*int b = 2;*/*/
/*int c = 3;*/
````

which is not valid C/C++ code. A simple fix for this would be to change the commentstring to '// %s' (in runtime/ftplugin/c.lua). Ideally, from reading previous issues, it seems this should be raised on vim/vim, but this is not a issue for vim due to the lack of inbuilt individual line based commenting (last I checked, feel free to correct me here). I add the same in my ftplugins, but a good default would be better.

### Expected behavior

Commenting

```
int a = 1;
// int b = 2;
int c = 3;
````

would lead to 
```
// int a = 1;
// // int b = 2;
// int c = 3;
````

(and appropriately for /*%s*/)

Edit: I can submit a PR if required, just wanted to confirm if this would be an acceptable change.

### PR DESCRIPTION
**Pull request by sohamshanbhag**
The present commenting plugin uses `/* %s */` for C/C++ comments. This has a problem when commenting commented code. For example

```
int a = 1;
/*int b = 2;*/
int c = 3;
````

when commented together by selecting using 'V' would become

```
/*int a = 1;*/
/*/*int b = 2;*/*/
/*int c = 3;*/
````

which is not valid C/C++ code. This commit changes the commit string to `// %s` which solves this problem.

Closes #29079


clason:
Please check the other ftplugins in Lua that also set a `commentstring`; these derived from the C default and should match unless this is explicitly wrong.

clason:
Yeah, that's what I thought -- you can't just change _one_ thing in Vim. You need to adapt the tests to the new behavior, and make sure the derived ftplugins (objc, ch, xs) can also handle the new comment string -- see https://github.com/neovim/neovim/issues/29079#issuecomment-2138023368

sohamshanbhag:
Yeah, I'll sort the tests out once the discussion on #29079 is complete.

clason:
Here's what I propose to move this forward:

* for any ftplugin in Lua that sets the commentstring for a C-type language (which includes `indent`), change to `// %s`: `c`, `d`, `cs`, `indent`
* for any ftplugin that includes a `runtime! ftplugin/c.vim`, add a separate ftplugin in Lua that sets the commentstring explicitly (as `// %s` if it's C-type: `objc`, `ch`, or the actually correct commentstring if it isn't: `xdefaults`, `xs`, `css`);
* `cpp` can be left as-is (to keep the link for user plugins);
* do not touch any other ftplugins; those are left for Vim to fix (if you are motivated enough, you can contact each maintainer and notify them).

echasnovski:
I agree that `//` sounds better as commentstring for C-like filetypes (where it is supported). And I also agree [with @clason's suggestion](https://github.com/neovim/neovim/pull/29085#issuecomment-2140283298). Ideally, of course, this change is better be made in upstream Vim filetype plugins, but it might be a stretch to ask for it.

clason:
I am _very_ skeptical that Vim will accept that change since they use the "classical" C commentstring in their own sources (unlike us -- which is the main motivation for _us_ to accept this change, which we normally wouldn't).

clason:
Missing `xs` and `css` -- _any_ filetype plugin that (implicitly or explicitly) derives its commentstring from the (C) default should be changed to have its commentstring set explicitly.

clason:
And to be clear: the goal here is _not_ to have "all C-like languages have the commentstring `// %s`"! The only goals are 
1. have C itself have `// %` (because that is the what we use in our own C code);
2. fix the fallout from this breaking change by explicitly setting all derived commentstrings correctly(!) (if at all reasonable, to the same value Vim uses -- because changes to _those_ should go through upstream).

echasnovski:
> I couldn't run tests locally since my tests are hanging on `commenting toggle_lines() respects tree-sitter injections:`, but the tests on github are passing.

I can not reproduce hanging test locally at d0eed1e5bf66. Running `TEST_FILE=test/functional/lua/comment_spec.lua make functionaltest` shows that all tests pass. Try executing `make distclean` before running tests.

clason:
> @clason please check the same.

Still missing xs (needs its own commentstring) and css (needs the old-style c commentstring).

sohamshanbhag:
@clason 

> Missing xs and css -- any filetype plugin that (implicitly or explicitly) derives its commentstring from the (C) default should be changed to have its commentstring set explicitly.

Regarding `xs`, I've raised it on vim-perl/vim-perl, so I'll add the same after they reply. The `css` commentstring is already set in lua to `/*%s*/`, which is correct, so I haven't changed it.

> fix the fallout from this breaking change by explicitly setting all derived commentstrings correctly(!) (if at all reasonable, to the same value Vim uses -- because changes to those should go through upstream).

Sorry I'm getting confused here. In the previous comment you mentioned

> for any ftplugin that includes a `runtime! ftplugin/c.vim`, add a separate ftplugin in Lua that sets the commentstring explicitly (as `// %s` if it's C-type: `objc`, `ch`, or the actually correct commentstring if it isn't: `xdefaults`, `xs`, `css`);

I've checked the ones I've changed accept `// %s` as comments, and have changed them appropriately. I've only added arduino, since it was broken due to the default commentstring removed in 0.9(vim has default `/**/`, so this is not a problem for them).

@echasnovski Probably my setup is borked, and I'm not proficient enough to debug it. Nevermind, since it is passing on your setup.

clason:
Thank you, then I'm satisfied. (I assumed you'd want to add padding to css as well while we're at it -- which we can since we already maintain our own ftplugin for it -- and that you'd want to make the change here already since the current commentstring is broken -- if you'd rather wait for a reply before making a change, that is fine, but we shouldn't wait too long. It's just important here that this is _also_ reported upstream.)

sohamshanbhag:
> I assumed you'd want to add padding to css as well while we're at it

I assume there would be some other filetypes too which you'd want to add padding to too. In that case, it would be better to have a different commit for that similar to vim's. If you want, I can add another PR with those too.

>  if you'd rather wait for a reply before making a change, that is fine

Yeah, I would prefer that. Either `xs` is broken for vim, and then it would be pulled from upstream when it is corrected there, or it is empty. Since I don't know what the correct commentstring is yet, I'd rather not add a wrong one. Let's wait till Monday?

clason:
> I assume there would be some other filetypes too which you'd want to add padding to too. 

No, that we wouldn't do unless Vim did it first. This is just about consistency within _our_ filetypes. (Otherwise we should remove all padding from this PR as well, i.e. `//%s`.)

> Since I don't know what the correct commentstring is yet, I'd rather not add a wrong one. Let's wait till Monday?

Sounds good, but they can be slow to answer, and I feel confident about making this change proactively.

sohamshanbhag:
> This is just about consistency within our filetypes. 

Makes sense. I've now changed calendar.lua and css.lua to include padding. This will concur with vim too once @ribru17's PR is merged there and lands in neovim. If you feel we need not wait for `xs`, feel free to merge this PR.

clason:
I would like to make the change to `xs` in this PR, with or without a reaction from vim-perl. But no great hurry.

sohamshanbhag:
Sure, then I'll wait till Monday, and if I don't receive any information from vim-perl, set it to ` # %s` as per the perldoc.

sohamshanbhag:
I have added the `xs` commentstring. Please feel free to merge this.

sohamshanbhag:
Hello, since you are backporting a lot of commentstring related commits to v0.10, could you please backport this one too? (apologies if there is a specific reason this has not been backported, in which case please ignore my request)

Thanks!

PR #29089, opened 2024-05-29
fix(ui): superfluous showmode due to unset globals

### LINKED ISSUE #29086: nvim sends excessive amount of grid_cursor_goto when pmenu is open with ext_popupmenu
### Problem

With nvim 0.10, nvim sends excessive (endless) amount of `grid_cursor_goto` and `flush` messages when the popupmenu is open (with ext_popupmenu).

### Steps to reproduce

1. use a GUI that as externalized popupmenu implemented and enabled (requires ext_popupmenu)
2. get your nvim to a state that it shows the popupmenu
3. observe the endless cursor goto and flush events

I pushed a branch to gnvim to make testing easier: https://github.com/vhakulinen/gnvim/tree/debug/excessive-cursor-goto (https://github.com/vhakulinen/gnvim/tree/e8b72c3c543e0ed953579fbb8175cb59f2ea6039 for exact commit).

If you have all the requirements installed on your system (i.e. latest gtk4) you can just run `cargo run --bin gnvim -- --nvim-args "-u NONE"` to test with a plain nvim. You can also use flatpak (see the README instructions). The default `make run` uses flatpak, and that branch has the nvim args adjusted.

### Expected behavior

Not sending unnecessary cursor goto events.

### Neovim version (nvim -v)

NVIM v0.10.0

no

### Operating system/version

Pop!_OS 22.04 LTS

### Terminal name/version

gnvim

-

### Installation

build from source

luukvbaal:
Problem:  Unsetting global variables earlier in #28578 to avoid
          recursiveness, caused superfluous or even unlimited
          showmode().
Solution: Partly revert #28578 so that the globals are unset at the end
          of showmode(), and avoid recursiveness for ext UI by adding a
          recursive function guard to each generated UI call that may
          call a Lua callback.

Fix #29086

### DIFF
--- a/src/nvim/generators/gen_api_ui_events.lua
+++ b/src/nvim/generators/gen_api_ui_events.lua
@@ -128,8 +128,16 @@ for i = 1, #events do
     write_signature(call_output, ev, '')
     call_output:write('\n{\n')
     if ev.remote_only then
+      -- Lua callbacks may emit other events or the same event again. Avoid the latter
+      -- by adding a recursion guard to each generated function that may call a Lua callback.
+      call_output:write('  static bool entered = false;\n')
+      call_output:write('  if (entered) {\n')
+      call_output:write('    return;\n')
+      call_output:write('  }\n')
+      call_output:write('  entered = true;\n')
       write_arglist(call_output, ev)
       call_output:write('  ui_call_event("' .. ev.name .. '", ' .. args .. ');\n')
+      call_output:write('  entered = false;\n')
     elseif ev.compositor_impl then
       call_output:write('  ui_comp_' .. ev.name)
       write_signature(call_output, ev, '', true)
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -931,13 +931,7 @@ int showmode(void)
     msg_ext_clear(true);
   }
 
-  // Don't make non-flushed message part of the showmode and reset global
-  // variables before flushing to to avoid recursiveness.
-  bool draw_mode = redraw_mode;
-  bool clear_cmd = clear_cmdline;
-  redraw_cmdline = false;
-  redraw_mode = false;
-  clear_cmdline = false;
+  // Don't make non-flushed message part of the showmode.
   msg_ext_ui_flush();
 
   msg_grid_validate();
@@ -960,8 +954,8 @@ int showmode(void)
     msg_check_for_delay(false);
 
     // if the cmdline is more than one line high, erase top lines
-    bool need_clear = clear_cmd;
-    if (clear_cmd && cmdline_row < Rows - 1) {
+    bool need_clear = clear_cmdline;
+    if (clear_cmdline && cmdline_row < Rows - 1) {
       msg_clr_cmdline();  // will reset clear_cmdline
     }
 
@@ -1083,7 +1077,7 @@ int showmode(void)
     }
 
     mode_displayed = true;
-    if (need_clear || clear_cmd || draw_mode) {
+    if (need_clear || clear_cmdline || redraw_mode) {
       msg_clr_eos();
     }
     msg_didout = false;                 // overwrite this message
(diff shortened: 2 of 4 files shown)


zeertzjq:
Is it possible to add a test?

luukvbaal:
> Is it possible to add a test?

Looking into it. I think there are more requirements than just `ext_popupmenu = true`, i.e. this already passes on master:

```diff
diff --git a/test/functional/ui/popupmenu_spec.lua b/test/functional/ui/popupmenu_spec.lua
index 8f8604eecb..33a48374c1 100644
--- a/test/functional/ui/popupmenu_spec.lua
+++ b/test/functional/ui/popupmenu_spec.lua
@@ -911,6 +911,30 @@ describe('ui/ext_popupmenu', function()
                                                                   |
     ]])
   end)
+
+  it('does not result in recursive showmode()', function()
+    local cursor = 0
+    local screen_cursor = screen._handle_grid_cursor_goto
+    screen._handle_grid_cursor_goto = function(...)
+      screen_cursor(...)
+      cursor = cursor + 1
+    end
+    insert([[foo bar  ]])
+    feed('i<C-N>')
+    screen:expect({
+      grid = [[
+        foo bar foo^                                                 |
+        {1:~                                                           }|*6
+        {2:-- Keyword completion (^N^P) }{5:match 1 of 2}                   |
+      ]], popupmenu={
+      anchor = { 1, 0, 8 },
+      items = { { "foo", "", "", "" }, { "bar", "", "", "" } },
+      pos = 0
+    }
+    })
+    vim.uv.sleep(1000)
+    eq(cursor < 10, true)
+  end)
 end)

 describe("builtin popupmenu 'pumblend'", function()
```

luukvbaal:
Yeah idk. It's easy to reproduce unnecessary showmodes on master but I'm not sure how the infinite state loop is entered in the gnvim repro.

luukvbaal:
> It's easy to reproduce unnecessary showmodes on master but I'm not sure how the infinite state loop is entered in the gnvim repro.

Added a test that at least tests the former.

src/nvim/generators/gen_api_ui_events.lua
```diff
@@ -128,8 +128,16 @@ for i = 1, #events do
     write_signature(call_output, ev, '')
     call_output:write('\n{\n')
     if ev.remote_only then
+      -- Lua callbacks may emit other events or the same event again. Avoid the latter
+      -- by adding a recursion guard to each generated function that may call a Lua callback.
```
luukvbaal, line 132: I proposed this once before to @bfredl but got no response then, feel free to review it here instead. I think it is more sustainable than adding recursion guards manually where needed and still avoids the problem(only present for Lua callbacks so only adding it here).

vhakulinen:
> It's easy to reproduce unnecessary showmodes on master but I'm not sure how the infinite state loop is entered in the gnvim repro.

The input is pretty much just this:
```
ifoobar<CR>foofoo<CR>foo<C-n>
```

What needs to happen is for the popumenu to open for completion. That input puts `foobar` and `foofoo` to the buffer, and then starts the completion from `foo` which triggers the completion for bot `foobar` and `foofoo`.

All this requires for `ext_popupmenu` to be enabled.

luukvbaal:
> All this requires for `ext_popupmenu` to be enabled.

Not true in my testing, I was talking about adding a test that fails on master. I was able to reproduce the issue in gnvim but not in our test environment.

vhakulinen:
I tested with the latest master, and the issue exists there too (not just for v0.10.0).

This PR fixes the issue.

vhakulinen:
> Not true in my testing, I was talking about adding a test that fails on master. I was able to reproduce the issue in gnvim but not in our test environment.

Are you able to reproduce the issue with `ext_messages` disabled?

luukvbaal:
Well no not the exact issue you described in #29086. But the underlying cause is not related to ext_popupmenu and can be tested for without it like in the added test in this PR.

src/nvim/generators/gen_api_ui_events.lua
bfredl: later on, I think we want recursiveness checks to be more semantic, like at least based on the category of events (like any ext_cmdline event handler shouldn't case another ext_cmdline event), but we can iterate on that during the 0.11 cycle. 

This would be okay for now.

justinmk (approved):

PR #29094, opened 2024-05-30
vim-patch:8.1.0716: get warning message when 'completefunc' returns nothing

glepnir:
Problem:    Get warning message when 'completefunc' returns nothing.
Solution:   Allow for returning v:none to suppress the warning message.
(Yasuhiro Matsumoto, closes https://github.com/vim/vim/pull/3789)

https://github.com/vim/vim/commit/cee9bc2e3dc5c16a9d2a8d0e23aa0d5fdefa3a4a


glepnir:
we don't use `v:none` so need changed to `v:null` .

PR #29096, opened 2024-05-30
fix(tui): skip TUI in ui_rgb_attached

### LINKED ISSUE #29056: `synIDattr` returns GUI color names with `notermguicolors`
### Problem

When opening a R markdown file (filetype rmd), receiving the following E421 color error tracing back to the SetupPandocHighlights function in runtime/syntax/pandoc.vim:

Error message:

```
Error executing lua callback: /opt/nvim-linux64/share/nvim/runtime/filetype.lua:35: Error executing lua: /opt/nvim-linux64/share/nvim/runtime/filetype.lua:36: BufReadPost Autocommands for "*"..FileType A
utocommands for "*"..Syntax Autocommands for "*"..function <SNR>5_SynSet[26]..script /opt/nvim-linux64/share/nvim/runtime/syntax/rmd.vim[41]../opt/nvim-linux64/share/nvim/runtime/syntax/pandoc.vim[697]..
function <SNR>25_SetupPandocHighlights, line 111: Vim(highlight):E421: Color name or number not recognized: ctermfg=NvimLightGreen                                                                         
stack traceback:                                                                                                                                                                                           
        [C]: in function 'nvim_cmd'                                                                                                                                                                        
        /opt/nvim-linux64/share/nvim/runtime/filetype.lua:36: in function </opt/nvim-linux64/share/nvim/runtime/filetype.lua:35>                                                                           
        [C]: in function 'nvim_buf_call'                                                                                                                                                                   
        /opt/nvim-linux64/share/nvim/runtime/filetype.lua:35: in function </opt/nvim-linux64/share/nvim/runtime/filetype.lua:10>                                                                           
stack traceback:                                                                                                                                                                                           
        [C]: in function 'nvim_buf_call'                                                                                                                                                                   
        /opt/nvim-linux64/share/nvim/runtime/filetype.lua:35: in function </opt/nvim-linux64/share/nvim/runtime/filetype.lua:10>     
```

**Repeating same in 0.9.5 does not produce this behavior.**

### Steps to reproduce

nvim --clean foo.Rmd

### Expected behavior

File open and with proper syntax highlighting

### Neovim version (nvim -v)

0.10.0

no, vim 8.2.2121

### Operating system/version

ubuntu 22,.04

### Terminal name/version

xterm-256color

### Installation

repo

gpanders:
The ui_rgb_attached function determines if any UI is attached which supports RGB (truecolor). We determine if the TUI supports RGB via the 'termguicolors' option which is checked at the beginning of this function. If the TUI does not support RGB ('termguicolors' is unset), we check to see if any _other_ UI is attached which supports RGB.

Normally, the TUI's "rgb" flag and the 'termguicolors' option are the same. However, they may differ during startup when the "rgb" flag is set by tui/tui.c to indicate to the core that the terminal emulator supports truecolor. The 'termguicolors' option is not actually set until _defaults.lua runs.

Fixes: https://github.com/neovim/neovim/issues/29056

### DIFF
--- a/src/nvim/ui.c
+++ b/src/nvim/ui.c
@@ -144,7 +144,7 @@ void ui_free_all_mem(void)
 /// Returns true if any `rgb=true` UI is attached.
 bool ui_rgb_attached(void)
 {
-  if (!headless_mode && p_tgc) {
+  if (p_tgc) {
     return true;
   }
   for (size_t i = 0; i < ui_count; i++) {
--- a/src/nvim/ui.c
+++ b/src/nvim/ui.c
@@ -148,7 +148,11 @@ bool ui_rgb_attached(void)
     return true;
   }
   for (size_t i = 0; i < ui_count; i++) {
-    if (uis[i]->rgb) {
+    // We do not consider the TUI in this loop because we already checked for 'termguicolors' at the
+    // beginning of this function. In this loop, we are checking to see if any _other_ UIs which
+    // support RGB are attached.
+    bool tui = uis[i]->stdin_tty || uis[i]->stdout_tty;
+    if (!tui && uis[i]->rgb) {
       return true;
     }
   }


bfredl (approved):

src/nvim/ui.c
```diff
@@ -148,7 +148,11 @@ bool ui_rgb_attached(void)
...
   }
   for (size_t i = 0; i < ui_count; i++) {
-    if (uis[i]->rgb) {
+    // We do not consider the TUI in this loop because we already checked for 'termguicolors' at the
+    // beginning of this function. In this loop, we are checking to see if any _other_ UIs which
+    // support RGB are attached.
+    bool tui = uis[i]->stdin_tty || uis[i]->stdout_tty;
+    if (!tui && uis[i]->rgb) {
       return true;
     }
```
zeertzjq, line 157: Since `p_tgc` is not checked above in headless mode, I think extra changes will be needed for headless mode. Something like this:
```suggestion
    }
    if (tui && p_tgc) {
      return true;
    }
```

bfredl: or maybe just change the check on the top to just check for `p_tgc` regardless of `headless_mode`. in headless mode 'termguicolors' can only be true if explicitly set, in case it conveys the user's explicit intent.

zeertzjq: Oh, right.

gpanders: a55dbfe0125fb7c7a049130ef42a375a66199061

zeertzjq (dismissed):

gpanders:
Tests are failing because we set `'termguicolors'` in the test runner, so `synIDattr` always behaves as if an RGB GUI is attached. I'm not sure why termguicolors is set though. I opened https://github.com/neovim/neovim/pull/29103 to see if we can remove it.

zeertzjq (approved):

PR #29103, opened 2024-05-31
test: do not set termguicolors in test runner

gpanders:
**test: do not set termguicolors in test runner**

It's not clear why this is needed and it has adverse side effects on other tests.

**fix(vim.text): remove assert from vim.text.hexdecode**

Instead, return nil plus an error message if the input is invalid. _This change is required to fix a test failure in terminal/tui_spec caused by the first commit._

### DIFF
--- a/test/functional/core/startup_spec.lua
+++ b/test/functional/core/startup_spec.lua
@@ -400,9 +400,6 @@ describe('startup', function()
         read_file('Xtest_startup_ttyout')
       )
     end)
-    if is_os('win') then
-      assert_log('stream write failed. RPC canceled; closing channel', testlog)
-    end
   end)
 
   it('input from pipe: has("ttyin")==0 has("ttyout")==1', function()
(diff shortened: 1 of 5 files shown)


zeertzjq (dismissed):
Please read #26437. Having 'notermguicolors' in test instance will cause terminal tests to be environment-dependent.

test/functional/core/startup_spec.lua
```diff
@@ -400,9 +400,9 @@ describe('startup', function()
         read_file('Xtest_startup_ttyout')
       )
     end)
-    if is_os('win') then
-      assert_log('stream write failed. RPC canceled; closing channel', testlog)
-    end
+    -- if is_os('win') then
+    --   assert_log('stream write failed. RPC canceled; closing channel', testlog)
+    -- end
```
gpanders, line 405: Tests pass with this commented and fail with it uncommented. I guess whatever issue caused this message in the log is fixed?

zeertzjq:
Hmm, I'm wondering whether it really makes sense to have 'termguicolors' decide whether `$COLORTERM` is set in `:terminal`. For a GUI, RGB colors are supported even if 'termguicolors' isn't set, but in that case it's more reasonable for `$COLORTERM` to still be set in `:terminal`.

gpanders:
@zeertzjq Is there some concrete action you want to take (or are expecting me to take) that is blocking this?

zeertzjq (approved):
I think this is OK. Most nested uses of Nvim already use `:set notermguicolors`.

PR #29104, opened 2024-05-31
feat(vim.net): fetch(), download(), and :e URL in lua

### LINKED ISSUE #23232: implement download() function without netrw
### Problem

- No easy way to download a file, though both `curl` and libuv are available in all Nvim targets.
- `:edit <uri>` (example: `:edit https://example.com/foo.txt`) is dependent on netrw for no reason

### Related

- https://github.com/neovim/neovim/issues/23231
- https://github.com/neovim/neovim/issues/4378

### Expected behavior

separate PRs:

1. introduce something like `vim.fetch()` (`vim.net.fetch()` ? idk, look at Go, NodeJS, for precedent)
2. use that to implement `:edit <uri>` without netrw

TheLeoP:
I saw that #23461 was closed, so I rebased it on top of master and addressed a few of the pending comments. In the following days I'll continue working on this PR.

Should close #23232


runtime/lua/vim/health/health.lua
```diff
@@ -397,6 +397,23 @@ local function check_external_tools()
   end
 end
 
+local function check_network()
```
clason, line 400: Can you model this check after the `external` check for `ripgrep`? Besides the version, the location can also be very important (system vs. "overlay package manager"). From painful experience, the full `curl -V` output can also be relevant -- there are package managers that install a curl without ssl support, and then suddenly nothing works...

runtime/doc/news.txt
```diff
@@ -154,13 +154,27 @@ UI
...
+• `:e, :w, :[range]w!, and :r https://*,http://*,ftp://*,scp://*` can now
+  use |vim.net.fetch()|. Opt-in via >lua
+      vim.g.lua_net_enable = true
+<
+• `:e https://*,http://*,ftp://*,scp://*` now automatically matches
+  filetype using the url. Optionally, you can enable (slower) automatic
+  full-text filetype matching for editing remote files that don't have an
+  extension (for example https://neovim.io/ -> `ft=html`) via >lua
+      vim.g.lua_net_ft_full = true
+<
```
clason, line 177: We should avoid adding configuration variables for this wherever possible. Since the decision has already been made that NetRW *delendam est* and we are very early in the dev cycle, we should just enable this by default (with a temporary(!) opt-out, hopefully to be removed by the time 0.11 is released).

Also, why would someone _not_ want actual filetype matching? How much slower are we talking about -- milliseconds or minutes? I think the basic expectation (that we should meet) is that remote editing behaves just like local editing. This means we should just do full matching (with possibly an "gimme _nao_, I know what I'm doing" fast option -- again, hopefully to be removed if we can iron out the issues that make it desirable).

runtime/plugin/net.lua
```diff
@@ -0,0 +1,333 @@
+if not vim.g.lua_net_enable then
+  return
+end
+
+local function url_safe_encode_byte(byte)
+  local char_map = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_'
+  return char_map:sub(byte % 64 + 1, byte % 64 + 1)
+end
+
```
clason, line 9: These functions should not be defined here (since `plugin/` is executed unconditionally) but in `vim.net`. This file should just define autocommands with callbacks that are (private) functions in `vim.net`.

clason: There's also a good chunk of code duplication in these callbacks, which further indicates that these should be refactored into dedicated functions.

clason:
To be honest, I think this code still needs a lot of work before it could be merged. Maybe it would be better to start with only `vim.net.download()`, and only consider `fetch()` after that is done?

runtime/doc/news.txt
TheLeoP: > How much slower are we talking about -- milliseconds or minutes?

I'm gonna disable this for now. If there are performance problems in the future, I'm gonna add it again

runtime/doc/lua.txt
```diff
@@ -4429,4 +4429,216 @@ tohtml.tohtml({winid}, {opt})                         *tohtml.tohtml.tohtml()*
...
+      • {key}    (`string`) Non case-sensitive header name.
+
+vim.net.download({url}, {path}, {opts})                   *vim.net.download()*
+    Asynchronously download a file. To read the response metadata, such as
+    headers and body, use |vim.net.fetch()|.
+
+    See man://curl for supported protocols. Not all protocols are fully
+    tested.
+
+    Please carefully note the option differences with |vim.net.fetch()|,
```
justinmk, line 4508: what are the differences? at a glance doesn't look like they need to be separate functions.

as clason also mentioned, we could start with a minimal version of `vim.net.fetch` which has the interface (signature) that we want, but only supports basic "download" behavior.

TheLeoP:
Netrw seems to handle `https:/some/path/to/a/dir/` showing the contents of the dir inside `:Explore`. Is it expected this to be handled as well?

TheLeoP:
Also, the Netrw implementation uses multiple commands (`ftp` for `ftp://`, for example). Is it expected to handle all of this protocols? My local version of `curl` seems to handle `ftp` just fine. What would be the expectations regarding protocol support for `curl` on the user local machine?

runtime/plugin/netrwPlugin.vim
```diff
@@ -46,10 +46,13 @@ augroup END
...
- au BufReadCmd   ftp://*,rcp://*,scp://*,http://*,https://*,dav://*,davs://*,rsync://*,sftp://*	exe "sil doau BufReadPre ".fnameescape(expand("<amatch>"))|call netrw#Nread(2,expand("<amatch>"))|exe "sil doau BufReadPost ".fnameescape(expand("<amatch>"))
- au FileReadCmd  ftp://*,rcp://*,scp://*,http://*,file://*,https://*,dav://*,davs://*,rsync://*,sftp://*	exe "sil doau FileReadPre ".fnameescape(expand("<amatch>"))|call netrw#Nread(1,expand("<amatch>"))|exe "sil doau FileReadPost ".fnameescape(expand("<amatch>"))
- au BufWriteCmd  ftp://*,rcp://*,scp://*,http://*,file://*,dav://*,davs://*,rsync://*,sftp://*			exe "sil doau BufWritePre ".fnameescape(expand("<amatch>"))|exe 'Nwrite '.fnameescape(expand("<amatch>"))|exe "sil doau BufWritePost ".fnameescape(expand("<amatch>"))
- au FileWriteCmd ftp://*,rcp://*,scp://*,http://*,file://*,dav://*,davs://*,rsync://*,sftp://*			exe "sil doau FileWritePre ".fnameescape(expand("<amatch>"))|exe "'[,']".'Nwrite '.fnameescape(expand("<amatch>"))|exe "sil doau FileWritePost ".fnameescape(expand("<amatch>"))
+ " Enable vimscript implementations if user opts-out
+ if exists("g:lua_net_disable")
```
justinmk, line 50: let's drop this . make it non-optional

justinmk:
> What would be the expectations regarding protocol support for `curl` on the user local machine?

whatever curl can easily handle is fine. or we can start with http[s] and add more later.

btw, please review comments starting from https://github.com/neovim/neovim/pull/23461#issuecomment-1660446261

TheLeoP:
> Could we implement gf and :edit https://... analogously to vim.ui.open (which provides gx)?

Would this look like providing a `vim.ui.edit` that uses the logic that's currently inside the `[Buf/File][Read/Write]Cmd` and using it to create a default `gf` keymap? Then, the autocmd would also call `vim.ui.open`?

runtime/doc/lua.txt
clason: Fetch is _much_ more general, allowing arbitrary requests (including _writing_, needed for `:e`). Download is a simplified interface for just... downloading a file (but not enough to implement `:e` over network). But note that you still need to expose some curl options since network conditions are very heterogeneous.

So both are needed (if both of these different features are wanted). I agree the name `fetch` is bad, though, and should be changed to make that clearer.

clason:
> Also, the Netrw implementation uses multiple commands (`ftp` for `ftp://`, for example). Is it expected to handle all of this protocols? My local version of `curl` seems to handle `ftp` just fine. What would be the expectations regarding protocol support for `curl` on the user local machine?

This was exactly my worry about relying on system curl. I don't know what the minimum expectable baseline is (I just know it's _very_ low). I think we need to rely on runtime checks here -- support everything that is reasonable (not necessarily for first PR) but fail gracefully if the local `curl` lacks support (`curl -V` should print a list of protocols; we probably want to cache that on first checking).

runtime/doc/lua.txt
```diff
@@ -4432,4 +4432,175 @@ tohtml.tohtml({winid}, {opt})                         *tohtml.tohtml.tohtml()*
...
+      • {redirect}           (`"follow"|"error"|"none"?`)
+      • {data}               (`string|table<string, any>?`)
+      • {headers}            (`HeaderTable|table<string, string|string[]>?`)
+      • {download_location}  (`string?`)
+      • {upload_file}        (`string?`)
+      • {credentials}        (`string?`)
+
+
+HeaderTable:append({key})                               *HeaderTable:append()*
+    Append value to header.
```
dundargoc, line 4465: I'm really not a fan of the vim.opt-style custom functions for basic table operations. Could this not be a normal table instead? IIRC even python's requests package use a basic dictionary for providing headers.

dundargoc:
I wonder if making this API public is premature. Could we not start using this internally for a while so we have time to iron out problems that show up encounter? Netrw could still use this under the hood ofc.

clason:
@TheLeoP I really appreciate the effort, but I would like to structure it a bit more. (There's a reason the original PR languished and got closed in the end...) I really don't think trying to do the whole `vim.net` in one go is the best approach here.

So I will close this PR and would like to ask you to open a new, fresh, one where you just extract and polish the functionality needed for `vim.net.download()` _and nothing else_. (This will be bikeshedding enough since we need it to work on all supported platforms and under reasonable network peculiarities, like proxies.)

Once we're happy with that, we can build on that with `vim.net.request()` for the purpose of `:edit`ing over a simple network protocol (manually at first). Once we are happy with _that_, we can slowly replace bits of NetRW with it (which _is_ the end goal).

runtime/doc/lua.txt
justinmk: > Fetch is _much_ more general,

why can't that be an overload? I still don't get why we need two different functions

PR #29111, opened 2024-05-31
refactor(lua): use tuple syntax everywhere

icholy:


runtime/lua/tohtml.lua
```diff
@@ -481,7 +481,7 @@ local function styletable_treesitter(state)
 end
 
 --- @param state vim.tohtml.state
---- @param extmark {[1]:integer,[2]:integer,[3]:integer,[4]:vim.api.keyset.set_extmark|any}
+--- @param extmark [integer,integer,integer,vim.api.keyset.set_extmark|any]
```
wookayin, line 484: While using tuple syntax here as well will be good for consistency, but I am actually not very sure this would read better than the previous one.

wookayin (commented):
As a side note, I think we need a consistent, standardized use of whitespace for better readability.

runtime/lua/tohtml.lua
clason: It's shorter; that alone helps.

icholy:
@wookayin if it was up to me, I'd choose the same whitespace rule that the docs use:
| Type             | Document As:                           |
|------------------|----------------------------------------|
| Union Type       | TYPE_1 \| TYPE_2                       |
| Array            | VALUE_TYPE[]                           |
| Tuple            | [VALUE_TYPE, VALUE_TYPE]               |
| Dictionary       | { [string]: VALUE_TYPE }               |
| Key-Value Table  | table<KEY_TYPE, VALUE_TYPE>            |
| Table literal    | { key1: VALUE_TYPE, key2: VALUE_TYPE } |
| Function         | fun(PARAM: TYPE): RETURN_TYPE          |

See: https://luals.github.io/wiki/annotations/#documenting-types

runtime/lua/tohtml.lua
icholy: Let me know if I should revert it.

dundargoc:
>  I think we need a consistent, standardized use of whitespace for better readability.

Pass. If it's not automated with a formatter then I'm not interested.

runtime/lua/tohtml.lua
wookayin: Since we never enforced formatting and we don't have an automatic formatter or any established convention for luadoc annotation, I fully agree with @dundargoc that one should not nitpick formatting/style issues and that it's OK to have inconsistent whitespaces for luadoc annotations, However, my point was that this line is not quite easy to read, so please consider add some whitespaces if you don't mind.

```suggestion
--- @param extmark [integer, integer, integer, vim.api.keyset.set_extmark|any]
```

icholy: done

MariaSolOs (approved):
If we're making a convention, then I think we should document it in `develop.txt`.

icholy:
Sounds like the best we can do right now is to open a feature request to luals.

justinmk (approved):

PR #29113, opened 2024-05-31
fix(luacats): allow all types inside tuples

icholy:
Found during: https://github.com/neovim/neovim/pull/29111


icholy:
cc @lewis6991

PR #29115, opened 2024-05-31
build(luarc): remove deprecated {3rd} syntax

dundargoc:
Vendor https://github.com/LuaCATS/luv.


mrcjkb:
@dundargoc feel free to assign me so I can pick this up later :smile:

dundargoc:
Sure, although I think this is good to go as is tbh. The type stuff from luals works as expected.

mrcjkb:
> Sure, although I think this is good to go as is tbh. The type stuff from luals works as expected.

Woops, I didn't realise this was a PR :man_facepalming:

dundargoc:
The peoples in the dev chat were not a fan, and we concluded we can worry about this when luals removes this/it becomes a problem.

PR #29116, opened 2024-05-31
feat: add `vim.usercommand` Lua API

mrcjkb:
This is a draft implementation of a lua wrapper around `nvim_create_usercommand`, which adds a basic API for creating scoped subcommands, based on the example in #29073.

See also: https://github.com/neovim/neovim/pull/29073#discussion_r1617940460

This implementation allows for only one level of subcommands. I think more would be overkill, but if desired, it would be possible to achieve more levels by making the `vim.usercommand.SubcommandOpts` type recursive (this could overwhelm users though).

I'd like to get some feedback on whether this is going in the right direction, before continuing with documentation and tests.

Plugins that implement this (with a similar approach):

- [rustaceanvim](https://github.com/mrcjkb/rustaceanvim/blob/2fa45427c01ded4d3ecca72e357f8a60fd8e46d4/lua/rustaceanvim/commands/init.lua#L17)
- [rocks.nvim](https://github.com/nvim-neorocks/rocks.nvim/blob/e52cfb8e78b6543f0d63d74c6d69ad7d98f0e892/lua/rocks/commands.lua#L130)


dundargoc:
I am in telepathic conversation with future justin, he says that `vim.usercommand` is too long and that is should be `vim.command`.

clason:
Nope, already got one of those.

dundargoc:
> Nope, already got one of those.

You mean `vim.cmd`?

mrcjkb:
> But let's focus on whether we really want an extra API, which we're very conservative about. What is it that it offers over `vim.api.nvim_create_user_command`?

It handles the boilerplate of creating completions for subcommands and figuring out which one to invoke.

From a user perspective, it's easier to discover subcommands if you rely on command completions (compared to a plugin polluting the command namespace for each possible action). I've heard feedback from plugin authors that Neovim not having a nice API for that is their reason for not implementing them (also, this feedback on my lua-plugin guide PR: https://github.com/neovim/neovim/pull/29073#discussion_r1617940460)

What I'm wondering is: Is this a "nice API"? Or is it not flexible enough?
If I compare it to a CLI library, it's still quite lacking (no flags, options, etc.). But maybe Neovim commands don't need the same API as a CLI tool?

antonk52:
My two cents on subcommands topic. In this issue as well as the linked discussion #29073 it is noted that people discover subcommands through completion. I have a counter point where some would prefer to have a flat list of commands instead of a nested/scoped subcommands. My rationale is fuzzy search. Telescope has [`builtin.commands`](https://github.com/nvim-telescope/telescope.nvim?tab=readme-ov-file#vim-pickers) and cmp has [command line completion](https://github.com/hrsh7th/nvim-cmp/blob/5260e5e8ecadaf13e6b82cf867a909f54e15fd07/doc/cmp.txt#L120-L129) where both fuzzy search through all available top level commands. This makes top level commands more discoverable than nested commands.

In my personal neovim setup I manually create top level user command for combined plugin subcommands as it is easier and more ergonomic to run a command this way with less key strokes. For example if a full plugin subcommand is `:Telescope live_grep` the fastest one can run it is `:Te<tab> li<tab><cr>`. While with a top level command and fuzzy completion one can do `:Tlg<tab><cr>`. This is also doable with a builtin tab completion in command line.

Another point for flat commands can be VS Code. Where cmd+shift+p opens a fuzzy picker with a flat list of all available commands in a shape of `Command description. [command-id]` (command-id can be omitted from picker). VS Code does not support nested commands/subcommands or passing arguments via this picker. Most commands are namespaced like `extension-name.command-name`. VS Code allows special characters like `.` and `-` in command names unlike neovim(do we want this to change?). Overall this experience leads to all commands to be in a single place and to be easily discoverable.

I wanted to make sure that the higher discoverability point for top most commands is mentioned here before it will be recommended for all plugins to use subcommands.

mrcjkb:
@antonk52 definitely an interesting take and valuable insight.

When I speak of discoverability, I mostly mean discovering commands that I don't know exist or forgot what they are called.
For this, scoping is very helpful, and I don't care about the number of keystrokes. When it comes to saving keystrokes for commands that I use often, I typically rely on keymaps and/or aliases.

With top-level PascalCase commands, typing `Foo<tab>` or `FB<tab>` (fuzzy) could match both one plugin's `FooBar` and another plugin's `FootballBar`, whereas `Foo<space><tab>` will show me only `Foo` subcommands. That decreases discoverability if you're looking for actions related to `Foo`.

IndianBoy42:
maybe if subcommands are 'officially' added here and thus considered first class citizens rather than simply an implementation trick, then those fuzzy command searches could be expanded to included the subcommand too. ie, making `:Tli<tab>` also return `Telescope live_grep`. That would require this PR to make them discoverable/iterable from the root in the same way top level commands are though.

justinmk:
> It handles the boilerplate of creating completions for subcommands and figuring out which one to invoke.

Show BEFORE and AFTER examples. Nothing else matters.

mrcjkb:
@justinmk sure :smile: 

## Before

```lua
---@class FooSubcommand
---@field impl fun(args:string[], opts: table)
---@field complete? fun(subcmd_arg_lead: string): string[]

---@type table<string, FooSubcommand>
local subcommand_tbl = {
    action1 = {
        impl = function(opts)
          -- Implementation
        end,
        -- This subcommand has no completions
    },
    action2 = {
        impl = function(opts)
            -- Implementation
        end,
        complete = function(subcmd_arg_lead, ...)
            -- Simplified example
            local install_args = {
                "first",
                "second",
                "third",
            }
            return vim.iter(install_args)
                :filter(function(install_arg)
                    -- If the user has typed `:Foo action2 fi`,
                    -- this will match 'first'
                    return install_arg:find(subcmd_arg_lead) ~= nil
                end)
                :totable()
        end,
        -- ...
    },
}

---@param opts table :h lua-guide-commands-create
local function foo_cmd(opts)
    local fargs = opts.fargs
    local subcommand_key = fargs[1]
    -- Get the subcommand's arguments, if any
    local args = #fargs > 1 and vim.list_slice(fargs, 2, #fargs) or {}
    local subcommand = subcommand_tbl[subcommand_key]
    if not subcommand then
        vim.notify("Foo: Unknown command: " .. subcommand_key, vim.log.levels.ERROR)
        return
    end
    -- Invoke the subcommand
    subcommand.impl(args, opts)
end

vim.api.nvim_create_user_command("Foo", foo_cmd, {
    nargs = "+",
    desc = "My awesome command with subcommand completions",
    complete = function(arg_lead, cmdline, _)
        -- Get the subcommand.
        local subcmd_key, subcmd_arg_lead = cmdline:match("^Foo[!]*%s(%S+)%s(.*)$")
        if subcmd_key 
            and subcmd_arg_lead 
            and subcommand_tbl[subcmd_key] 
            and subcommand_tbl[subcmd_key].complete
        then
            -- The subcommand has completions. Return them.
            return subcommand_tbl[subcmd_key].complete(subcmd_arg_lead)
        end
        -- Check if cmdline is a subcommand
        if cmdline:match("^Foo[!]*%s+%w*$") then
            -- Filter subcommands that match
            local subcommand_keys = vim.tbl_keys(subcommand_tbl)
            return vim.iter(subcommand_keys)
                :filter(function(key)
                    return key:find(arg_lead) ~= nil
                end)
                :totable()
        end
    end,
})
```

## After

```lua
---@type table<string, FooSubcommand>
local subcommand_tbl = {
    action1 = {
        impl = function(opts)
          -- Implementation
        end,
        -- This subcommand has no completions
    },
    action2 = {
        impl = function(opts)
            -- Implementation
        end,
        complete = function(subcmd_arg_lead, ...)
            -- Simplified example
            local install_args = {
                "first",
                "second",
                "third",
            }
            return vim.iter(install_args)
                :filter(function(install_arg)
                    -- If the user has typed `:Foo action2 fi`,
                    -- this will match 'first'
                    return install_arg:find(subcmd_arg_lead) ~= nil
                end)
                :totable()
        end,
        -- ...
    },
}

vim.usercommand.create("Foo", subcommand_tbl, {
    nargs = "+",
    desc = "My awesome command with subcommand completions",
}
```

The "after" variant also accepts a default command (e.g. if no subcommand was entered):

```lua
vim.usercommand.create('Foo', {
  function() print('foo') end, -- default
  bar = {
    command = function(opts) print('bar') end,
    complete = function(arg_lead, cmdline) return { 'a', 'b', 'c' } end,
    bang = true,
  },
  baz = {
    command = function(opts) print('baz') end,
  },
})
```

or a function without subcommands:

```lua
vim.usercommand.create('Foo', function() print('foo') end, { bang = true })
```

justinmk:
Seems like this is oriented around the desire for subcommands, thus the title is misleading. 

If subcommands are needed as a concept, please create an issue that describes the interface. Shouldn't that be a core change? Else this is a parallel "platform", more like a utility function. Listing and inspecting the subcommands isn't supported, for example. And the subcommands aren't available from `:command <name>`.

PR #29117, opened 2024-05-31
feat(terminal): add support for copying with OSC 52 in embedded terminal

### LINKED ISSUE #20672: Support OSC 52 remote copy in terminal buffer
### Feature description

OSC52 can let us copy text from a remote (e.g., ssh) machine to the local terminal (e.g., Windows Terminal), however, the terminal buffer in Neovim seems didn't support this feature. The behavior this issue wants to accomplish is to allow us to copy some text in the Neovim terminal, such as by executing the command like `printf "\E]52;c;%s\a"` in the terminal, and the result is copied to local.

gpanders:
When libvterm receives the OSC 52 escape sequence it ignores it because Nvim does not set any selection callbacks. Install selection callbacks that forward to the clipboard provider, so that setting the clipboard with OSC 52 in the embedded terminal writes to the system clipboard.

Fixes: #20672

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -138,7 +138,8 @@ STARTUP
 
 TERMINAL
 
-• TODO
+• The |terminal| now understands the OSC 52 escape sequence to write to the
+  system clipboard (copy). Querying with OSC 52 (paste) is not supported.
 
 TREESITTER
 
--- a/runtime/doc/nvim_terminal_emulator.txt
+++ b/runtime/doc/nvim_terminal_emulator.txt
@@ -164,7 +164,22 @@ directory indicated in the request. >lua
 To try it out, select the above code and source it with `:'<,'>lua`, then run
 this command in a :terminal buffer: >
 
-   printf "\033]7;file://./foo/bar\033\\"
+    printf "\033]7;file://./foo/bar\033\\"
+
+OSC 52: write to system clipboard		*terminal-osc52*
+
+Applications in the :terminal buffer can write to the system clipboard by
+emitting an OSC 52 sequence. Example: >
+
+    printf '\033]52;;%s\033\\' "$(echo -n 'Hello world' | base64)"
+
+Nvim uses the configured |clipboard| provider to write to the system
+clipboard. Reading from the system clipboard with OSC 52 is not supported, as
+this would allow any arbitrary program in the :terminal to read the user's
+clipboard.
+
+OSC 52 sequences sent from the :terminal buffer do not emit a |TermRequest|
+event. The event is handled directly by Nvim and is not forwarded to plugins.
 
 ==============================================================================
 Status Variables				*terminal-status*
--- a/src/nvim/base64.c
+++ b/src/nvim/base64.c
@@ -142,6 +142,7 @@ char *base64_encode(const char *src, size_t src_len)
 /// @param [out] out_lenp Returns the length of the decoded string
 /// @return Decoded string
 char *base64_decode(const char *src, size_t src_len, size_t *out_lenp)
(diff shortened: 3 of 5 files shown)


gpanders:
We will need some way to make this configurable (and for pasting, disabled by default).

gpanders:
Also wondering if instead we should just emit a TermRequest event for OSC 52.

EliasA5:
> Also wondering if instead we should just emit a TermRequest event for OSC 52.

I think its better to do TermRequest event, because its more flexible and will allow us to also send the event to terminals that handle the sequence differently, for example tmux expects osc to be wrapped in a special tmux escape sequence.
If we forward a TermRequest event we can define the following autocmd to copy into the system:
```lua
local function write(osc52)
  local success = false

  if vim.fn.filewritable('/dev/fd/2') == 1 then
    success = vim.fn.writefile({osc52}, '/dev/fd/2', 'b') == 0
  else
    success = vim.fn.chansend(vim.v.stderr, osc52) > 0
  end

  return success
end

vim.api.nvim_create_autocmd({ 'TermRequest' }, {
  desc = 'Handles OSC 52',
  callback = function(args)
    if args.data:match('\027]52;c;') then
      local to_copy = args.data:gsub('\027]52;c;', '')
      local osc52 = string.format('\27]52;c;%s\7', to_copy)
      if os.getenv("TMUX")
          or os.getenv("TERM"):match("^tmux")
          or os.getenv("TERM"):match("^screen") then
        osc52 = string.format('\27Ptmux;\27%s\27\\', osc52)
      end
      write(osc52)
    end
  end
})
```

This will send the correct sequence to the controlling terminal be it tmux or another terminal that supports osc52.

EliasA5:
we can also patch libvterm to fallback to our osc handler when selection callback is not defined https://github.com/EliasA5/libvterm/commit/57911e83cbe2d3870b192eecf943da8a7509a4fd, this will send a TermRequest event.
still doesn't solve pasting from osc52.

gpanders:
> we can also patch libvterm to fallback to our osc handler when selection callback is not defined [EliasA5/libvterm@57911e8](https://github.com/EliasA5/libvterm/commit/57911e83cbe2d3870b192eecf943da8a7509a4fd), this will send a TermRequest event. still doesn't solve pasting from osc52.

We don't vendor libvterm so we can't just patch the source code. We would have to send the patch upstream and have them accept it and cut a new release.

>Also wondering if instead we should just emit a TermRequest event for OSC 52.

I've thought about this some more and don't think this is a good idea, for a couple reasons:

1. libvterm already Base64 decodes the OSC sequence before calling the callback. If we send this is a TermRequest event, we would have to re-encode the sequence for the autocommand, only for the plugin to then decode it again!
2. There are security considerations with the OSC 52 sequence, particularly around querying the clipboard. The OSC 52 sequence allows any arbitrary program to query the clipboard contents of the system. Most terminal emulators disallow this by default (or provide a prompt to the user), but if we simply forward this request to a TermRequest event handler we are putting the burden of properly handling this to plugins.

I think that for now we should simply support the OSC 52 sequence in `:terminal` buffers for writing only and ignore "query" requests. Otherwise, we would need to add a new option so that users can specify themselves whether applications in `:terminal` buffers can read the clipboard, and I'm not convinced this is worth it (after all, users can paste directly into the application using `Ctrl-V` or `"+p` without OSC 52 at all).

justinmk (approved):
Is there something we can say about this in `nvim_terminal_emulator.txt`, perhaps at `:help terminal-events` ?

gpanders:
> Is there something we can say about this in `nvim_terminal_emulator.txt`, perhaps at `:help terminal-events` ?

Good catch, added.

PR #29121, opened 2024-06-01
fix(types): support named and vararg returns for `fun()` types

willothy:
This should remove the "redundant return value" warning when returning multiple values from functions passed to `Iter:map()`.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -4079,10 +4079,10 @@ Iter:map({f})                                                     *Iter:map()*
 <
 
     Parameters: ~
-      • {f}  (`fun(...):any`) Mapping function. Takes all values returned from
-             the previous stage in the pipeline as arguments and returns one
-             or more new values, which are used in the next pipeline stage.
-             Nil return values are filtered from the output.
+      • {f}  (`fun(...):...:any`) Mapping function. Takes all values returned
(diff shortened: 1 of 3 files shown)


willothy:
Hmm seems like the docgen doesn't like this. Maybe just `...` will work?

willothy:
Ok I just ended up updating the LuaCats parser to support vararg returns. Let me know if that part looks okay, I've never used Lpeg before :)

willothy:
cc @lewis6991 re: the parser edits

icholy:
@willothy can you the docs where this syntax is explained? Not sure if I'm blind, but I can't find anything here https://luals.github.io/wiki/annotations/

willothy:
Now that I'm looking I can't find anything either. Seems to be only documented with vararg params. It does work in lua-ls though.

icholy:
Yeah it works as expected. How did you learn about this syntax?

willothy:
> Yeah it works as expected. How did you learn about this syntax?

Must've seen it in another plugin, or just figured it out by chance. I don't actually remember, sorry :)

willothy:
Honestly maybe I should PR a doc example into lua-ls too, that could be nice to have.

justinmk (approved):

PR #29122, opened 2024-06-01
fix(lsp): trim trailing whitespace from completion words

mfussenegger:
the `complete()` mechanism doesn't play nicely with trailing newlines or
tabs. A newline causes it to insert a null character, showing up as
`^@`.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -153,7 +153,8 @@ local function get_completion_word(item)
       return item.label
     end
   elseif item.textEdit then
-    return item.textEdit.newText
+    local word = item.textEdit.newText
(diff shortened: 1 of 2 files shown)


MariaSolOs (approved):

luozhiya:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

> 	 * *Note:* The range of the edit must be a single line range and it must
>	 * contain the position at which completion has been requested.

The LSP specification restricts `textEdit` to a `single line` while **allowing** it to include spaces. However, merely cutting the first word might lead to an incomplete result.

For example `local i = 10`

Trigger
```lua
loc
   ^
```

Expected completion
```lua
al i = 10
```

Cutting
```
al
```

PR #29123, opened 2024-06-01
fix(completion): make sure the info preview window is highest one

glepnir:
Problem: when use completion in a float window the info window will be shown under float window.

Solution: give info preview window zindex a max value.


justinmk:
still needed?

glepnir:
No

PR #29129, opened 2024-06-01
Fix tutor

Empathy117:
Visual selection should use uppercase V instead of lowercase v button


clason:
Why? These are different things (character wise vs line wise). Which of these does not work with charwise motions?

Empathy117:
> Why? These are different things (character wise vs line wise). Which of these does not work with charwise motions?

Of course it's work, but if use v it's not really obvious, you can't see any difference from your cursor but use V the line will be selected, so I think it's better to use capital V rather than lowercase v or explain more about visual mode, it's just quite brief.

clason:
Then I am sorry, but this goes against the intent here.

Issue #16062, opened 2021-10-18
`vim.diagnostic.goto_next / goto_prev` doesn't automatically show diagnostics at the end of a line


jose-elias-alvarez:
### Neovim version (nvim -v)

NVIM v0.6.0-dev+472-gffc28dcbd

### Language server name/version

Observed with `tsserver`, but could happen with any server / diagnostic source

### Operating system/version

macOS Big Sur 11.6

### :checkhealth

```markdown
lsp: health#lsp#check
========================================================================
  - ERROR: Failed to run healthcheck for "lsp" plugin. Exception:
    function health#check[24]..health#lsp#check, line 2
    Vim(lua):E5108: Error executing lua [string ":lua"]:1: attempt to call field 'check_health' (a nil value)

lspconfig: require("lspconfig.health").check()
========================================================================
  - INFO: tsserver: configuration checked.
  - INFO: sumneko_lua: configuration checked.
  - INFO: null-ls: configuration checked.

nvim: health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $TERM_PROGRAM='tmux'
  - INFO: $COLORTERM='truecolor'

## tmux
  - OK: escape-time: 0
  - INFO: Checking stuff
  - OK: focus-events: on
  - INFO: $TERM: tmux-256color

nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found  0.20.0 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v16.11.1 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl" }
    Version: Apple clang version 13.0.0 (clang-1300.0.29.3)
  - OK: Neovim was compiled with tree-sitter runtime ABI version 13 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - json           ✓ ✓ ✓ ✓ . 
  - tsx            ✓ ✓ ✓ ✓ ✓ 
  - c              ✓ ✓ ✓ ✓ ✓ 
[...]

PR #29134, opened 2024-06-01
fix(diagnostic): fix float scope filtering

atoerien:
This fixes two bugs with the diagnostic float popup.

The first is simple, the fix in #28301 only changed the behaviour of the default `scope = "line"`, if `scope = "cursor"` then hovering on multi-line diagnostics doesn't work.

The second is that the filtering treats the end of the range as being inclusive, when it should be exclusive (well this is what is used everywhere else, it doesn't seem to be strictly specified in the docs...).
This is easy to see when using `scope = "cursor"` - just open the popup with the cursor on the character directly after the end of the underline.
When `scope = "line"`, the easiest way to see it is to add a multi-line diagnostic that has `end_col = 0`:
```
:lua vim.diagnostic.set(vim.api.nvim_create_namespace("test"), 0, { { lnum = 1, col = 0, end_lnum = 2, end_col = 0, message = "test" } })
```
then open the popup with the cursor on the line below the underline.

A corner case that I came across was how to deal with zero-width diagnostics, where `lnum == end_lnum` and `col == end_col`. Currently, underlining treats these as having width 1, so I've made both `scope = "cursor"` and `scope = "line"` do the same. However this behaviour seems to have been deemed a bug due to #28986, and #29022 will remove underlines for zero-width diagnostics again. 

So might it be better to treat zero-width diagnostics as spanning the whole line? Meaning that opening the popup will work from anywhere in the line even when `scope = "cursor"` ? The reasoning being that since there's no underline, a user wouldn't know where to put the cursor for the popup to work, so it should just work anywhere.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -1857,16 +1857,19 @@ function M.open_float(opts, ...)
   if scope == 'line' then
     --- @param d vim.Diagnostic
     diagnostics = vim.tbl_filter(function(d)
-      return lnum >= d.lnum and lnum <= d.end_lnum
+      return lnum >= d.lnum
+        and lnum <= d.end_lnum
+        and (d.lnum == d.end_lnum or lnum ~= d.end_lnum or d.end_col ~= 0)
     end, diagnostics)
   elseif scope == 'cursor' then
-    -- LSP servers can send diagnostics with `end_col` past the length of the line
+    -- If `col` is past the end of the line, show if the cursor is on the last char in the line
     local line_length = #api.nvim_buf_get_lines(bufnr, lnum, lnum + 1, true)[1]
     --- @param d vim.Diagnostic
     diagnostics = vim.tbl_filter(function(d)
-      return d.lnum == lnum
-        and math.min(d.col, line_length - 1) <= col
-        and (d.end_col >= col or d.end_lnum > lnum)
+      return lnum >= d.lnum
+        and lnum <= d.end_lnum
+        and (lnum ~= d.lnum or col >= math.min(d.col, line_length - 1))
+        and ((d.lnum == d.end_lnum and d.col == d.end_col) or lnum ~= d.end_lnum or col < d.end_col)
     end, diagnostics)
   end
 
--- a/test/functional/lua/diagnostic_spec.lua
+++ b/test/functional/lua/diagnostic_spec.lua
@@ -2291,6 +2291,38 @@ describe('vim.diagnostic', function()
         return lines
       ]]
       )
+
+      -- End position is exclusive
+      eq(
+        vim.NIL,
+        exec_lua [[
+        local diagnostics = {
+          make_error("Syntax error", 1, 1, 2, 0),
+        }
+        vim.api.nvim_win_set_buf(0, diagnostic_bufnr)
+        vim.diagnostic.set(diagnostic_ns, diagnostic_bufnr, diagnostics)
+        vim.api.nvim_win_set_cursor(0, {1, 1})
+        local _, winnr = vim.diagnostic.open_float(0, {header=false, pos={2,0}})
+        return winnr
+      ]]
+      )
+
+      -- Works when width == 0
+      eq(
+        { '1. Syntax error' },
+        exec_lua [[
+        local diagnostics = {
+          make_error("Syntax error", 2, 0, 2, 0),
+        }
+        vim.api.nvim_win_set_buf(0, diagnostic_bufnr)
+        vim.diagnostic.set(diagnostic_ns, diagnostic_bufnr, diagnostics)
+        vim.api.nvim_win_set_cursor(0, {1, 1})
+        local float_bufnr, winnr = vim.diagnostic.open_float(0, {header=false, pos={2,1}})
+        local lines = vim.api.nvim_buf_get_lines(float_bufnr, 0, -1, false)
+        vim.api.nvim_win_close(winnr, true)
+        return lines
+      ]]
+      )
     end)
 
     it('can show diagnostics from a specific position', function()
@@ -2299,7 +2331,7 @@ describe('vim.diagnostic', function()
         { 'Syntax error' },
         exec_lua [[
         local diagnostics = {
-          make_error("Syntax error", 1, 1, 1, 2),
+          make_error("Syntax error", 1, 1, 1, 3),
           make_warning("Some warning", 1, 3, 1, 4),
         }
         vim.api.nvim_win_set_buf(0, diagnostic_bufnr)
@@ -2317,7 +2349,7 @@ describe('vim.diagnostic', function()
         { 'Some warning' },
         exec_lua [[
         local diagnostics = {
-          make_error("Syntax error", 1, 1, 1, 2),
+          make_error("Syntax error", 1, 1, 1, 3),
           make_warning("Some warning", 1, 3, 1, 4),
         }
         vim.api.nvim_win_set_buf(0, diagnostic_bufnr)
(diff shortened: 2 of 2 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -1857,16 +1857,37 @@ function M.open_float(opts, ...)
...
+      else
+        -- end-exclusive range, don't show if on end_lnum and end_col == 0
+        return lnum ~= d.end_lnum or d.end_col ~= 0
+      end
     end, diagnostics)
   elseif scope == 'cursor' then
-    -- LSP servers can send diagnostics with `end_col` past the length of the line
-    local line_length = #api.nvim_buf_get_lines(bufnr, lnum, lnum + 1, true)[1]
+    -- LSP servers can send diagnostics with `col` past the end of the line
+    local lnum_length = #api.nvim_buf_get_lines(bufnr, lnum, lnum + 1, true)[1]
```
gpanders, line 1874: This comment changes `end_col` to `col`. Was this intentional? Was the old comment incorrect?

If the server can send both `col` and `end_col` past the end of the line we should mention both here.

Do you have an example of a server sending `col` (not `end_col`) past the end of the line? That seems very peculiar to me.

gpanders, line 1874: Any reason to rename this variable? The old name `line_length` was fine, the new name `lnum_length` is more confusing (`lnum` is the line number, so `lnum_length` doesn't really make sense).

runtime/lua/vim/diagnostic.lua
```diff
@@ -1857,16 +1857,37 @@ function M.open_float(opts, ...)
...
+        return false
+      end
+      if d.lnum == d.end_lnum then
+        -- if diagnostic only spans one line, always show
+        -- allows showing diagnostics with zero width (where lnum == end_lnum and col == end_col)
+        return true
+      else
+        -- end-exclusive range, don't show if on end_lnum and end_col == 0
+        return lnum ~= d.end_lnum or d.end_col ~= 0
+      end
```
gpanders, line 1870: It looks like the logic here is the same as before, except for the addition of the end-exclusive range. I find the refactored logic here hard to follow (multiple returns and if statements), we can keep the original structure and just add the new condition:

```suggestion
      return (lnum >= d.lnum and lnum <= d.end_lnum) and (lnum < d.end_lnum or d.end_col ~= 0)

```

runtime/lua/vim/diagnostic.lua
AThePeanut4: This was intentional yes - the issue that was meant to fix (#16062) gives this example, where `col` is past the end of the line:

> The easiest way to reproduce the issue is to directly create diagnostics using `vim.diagnostic.set`. Open a buffer with the following content:
> 
> ```
> hello
> hello
> ```
> 
> Then run the following Lua code to set a diagnostic at the end of the first line:
> 
> ```lua
> vim.diagnostic.set( vim.api.nvim_create_namespace("test"), 0, { { lnum = 0, end_lnum = 1, col = 5, end_col = 0, message = "test" } })
> ```
> 
> (Note that the range is meant to specifically denote the end of the first line, not including the final `o` in `hello`. More on this below.)

Everything is zero-based, so `col = 5` is past the end of the line. `end_col` can also be past the end of the line, but that doesn't need special treatment - the condition `col < d.end_col` will always be true and the float will just _always_ be shown. `col` past the end of the line does need special treatment, because in that case the condition `col < d.col` will always be true and the float will _never_ be shown.

As for an actual example, the #16062 mentions `tsserver`, but "didn't have an example handy", and neither do I :( I can try and find one if you'd like.

AThePeanut4: The reasoning behind that was that the length is for the line referred to by `d.lnum`, if the diagnostic was multiple lines long. But now that I think about it, that is more confusing - I'll revert it back.

runtime/lua/vim/diagnostic.lua
AThePeanut4: That condition isn't the same - if the diagnostic is `{ lnum = 0, end_lnum = 0, col = 0, end_col = 0 }` (zero width) then that will return `false` when the cursor is on line 0, meaning the diagnostic will never be "hoverable".

gpanders: I did make an edit, so it should be the same now. I ran it against your test cases and they all pass.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1857,16 +1857,37 @@ function M.open_float(opts, ...)
...
+        return false
+      end
+      if d.lnum == d.end_lnum and d.col == d.end_col then
+        -- treat zero-width diagnostics as having width 1
+        -- condition is effectively this:
+        -- return lnum ~= d.end_lnum or col < (d.col + 1)
+        return col == d.col
+      else
+        return lnum ~= d.end_lnum or col < d.end_col
+      end
```
gpanders, line 1890: I'm realizing that we already have all of this logic in our `_range` module, but that is currently namespaced under `vim.treesitter` so wouldn't be appropriate to use here. A follow up PR would be to move `_range.lua` into a higher-level/more generic namespace, and refactor some of the diagnostics code to use that instead.

runtime/lua/vim/diagnostic.lua
AThePeanut4: Originally I tried to keep both branches as one big conditional statement, but the `scope = 'cursor'` case was too complex for that, so I decided to change both to use if-then-returns for consistency. But it's certainly not a big deal - I don't mind changing it.

AThePeanut4:
@gpanders Thoughts on this?
> So might it be better to treat zero-width diagnostics as spanning the whole line? Meaning that opening the popup will work from anywhere in the line even when scope = "cursor" ? The reasoning being that since there's no underline, a user wouldn't know where to put the cursor for the popup to work, so it should just work anywhere.

The change would just be to replace `return col == d.col` with `return true` here:
https://github.com/neovim/neovim/blob/036b8e1fb8002d986fef8a2c62ac7a1914aa7bf2/runtime/lua/vim/diagnostic.lua#L1876-L1883
It might also then be possible to make a big conditional statement rather than if-returns.

gpanders:
> @gpanders Thoughts on this?
> 
> > So might it be better to treat zero-width diagnostics as spanning the whole line? Meaning that opening the popup will work from anywhere in the line even when scope = "cursor" ? The reasoning being that since there's no underline, a user wouldn't know where to put the cursor for the popup to work, so it should just work anywhere.


What are some examples of zero-width diagnostics? Should we even treat those as valid at all? The LSP specification doesn't mention anything about this that I could find, and I am not sure that zero-width diagnostics makes sense outside of LSP either. cc @mfussenegger.

Even if zero-width diagnostics are considered valid, they are still navigable with `vim.diagnostic.jump()` (maybe?). But to answer the quoted question, I think I need to first understand what these zero-width diagnostics even are.

mfussenegger:
> What are some examples of zero-width diagnostics? Should we even treat those as valid at all? The LSP specification doesn't mention anything about this that I could find, and I am not sure that zero-width diagnostics makes sense outside of LSP either. cc @mfussenegger.

nvim-lint has quite a few linters with error formats like `'%f:%l: %m'` which leads to diagnostics like:

```
    message: fqcn[action-core]: Use FQCN for builtin module actions (command).
    end_col: 0
    lnum: 30
    col: 0
    bufnr: 1
    namespace: 6
    end_lnum: 30
    source: ansible-lint
    severity: 3
```

The current behavior that the float works on the whole line for them makes sense to me.

Having end-lnum/end-col is in some ways a newer trend.

gpanders:
Sounds good. @AThePeanut4 go ahead and proceed with that then.

>It might also then be possible to make a big conditional statement rather than if-returns.

Let's see how this looks and make a decision.

AThePeanut4:
@gpanders Done. It's a fairly big conditional, but I think it's fine.

gpanders:
> @gpanders Done. It's a fairly big conditional, but I think it's fine.

I agree, looks good to me. Thanks!

gpanders (approved):

PR #29143, opened 2024-06-02
fix(lsp): use client.id instead of pairs index

ippachi:
Completion side effects not working randomly. 
Because when creating the table of LSP responses, the table index was used, 
but this is not the same as the actual client_id, so it was changed to use the client_id directly.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -349,7 +349,7 @@ function M._convert_results(
   return matches, server_start_boundary
 end
 
---- @param clients table<integer, vim.lsp.Client>
+--- @param clients table<integer, vim.lsp.Client> # keys != client_id
 --- @param bufnr integer
 --- @param win integer
 --- @param callback fun(responses: table<integer, { err: lsp.ResponseError, result: vim.lsp.CompletionResult }>)
@@ -359,7 +359,8 @@ local function request(clients, bufnr, win, callback)
   local request_ids = {} --- @type table<integer, integer>
   local remaining_requests = vim.tbl_count(clients)
 
-  for client_id, client in pairs(clients) do
+  for _, client in pairs(clients) do
+    local client_id = client.id
     local params = lsp.util.make_position_params(win, client.offset_encoding)
     local ok, request_id = client.request(ms.textDocument_completion, params, function(err, result)
       responses[client_id] = { err = err, result = result }


MariaSolOs:
I see. This works when invoking `trigger` from the newly-added completion code, since `buf_handles` keeps a dictionary from ID to client. But when invoking the function from `_omnifunc`, the input client list is the result of `vim.lsp.get_clients`, which returns a list and not a dict.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -359,7 +359,8 @@ local function request(clients, bufnr, win, callback)
   local request_ids = {} --- @type table<integer, integer>
   local remaining_requests = vim.tbl_count(clients)
 
-  for client_id, client in pairs(clients) do
+  for _, client in pairs(clients) do
```
MariaSolOs, line 362: To prevent confusion in the future, could you please also change the parameter type of `clients` to `vim.lsp.Client[]`? That way it's clear that we're not expecting a dictionary or doing anything with the keys.

MariaSolOs (approved):

runtime/lua/vim/lsp/completion.lua
ippachi: Thank you for your quick response and explanation! Certainly I use omnifunc.

I changed it.

mfussenegger: > could you please also change the parameter type of clients to vim.lsp.Client[]? That way it's clear that we're not expecting a dictionary or doing anything with the keys.

The downside of that is that it would give the impression that it's safe to use ipairs which isn't the case if the method is called from the autotrigger path.

I think adding a comment that keys!=client id would be clearer. 

Changing the type to `table<integer, vim.lsp.Client>|vim.lsp.Client[]` would be another option but somebody might take that as a mistake and try to clean it up

icholy: I think it should be `table<unknown, vim.lsp.Client>`

runtime/lua/vim/lsp/completion.lua
```diff
@@ -359,7 +359,8 @@ local function request(clients, bufnr, win, callback)
...
+    local client_id = client.id
```
icholy, line 363: This extra variable seem superfluous. You can replace all subsequent usages of `client_id` with `client.id`.

mfussenegger: It's perfectly fine to use the extra variable. Please no style nitpicking if there isn't a substantial advantage to do it in another way.

runtime/lua/vim/lsp/completion.lua
ippachi: I prefer `vim.lsp.Client[]` or `table<integer, vim.lsp.Client>|vim.lsp.Client[]`. 

> I think adding a comment that keys!=client id would be clearer.

I think just only adding a comment, the role of the keys in the table is unclear and might confuse someone. (it might be because of I am beginner of neovim code base)

MariaSolOs: I don't feel too strongly about this, let's not overthink it. Changing the type and/or adding a comment is okay, the main goal is to prevent this from happening again and clarifying the use and expected structure of the parameter.

ippachi: Okay. I only added comment.

@mfussenegger 
Comment position is okay? Or top of for-loop?

PR #29155, opened 2024-06-03
fix(tui)!: correct behavior of 'nottimeout'

### LINKED ISSUE #29047: initial file open sometimes replaces character under cursor with 'g'
### Problem

Frequently, but not always, the first file opened from the command line will have the character under the cursor replaced by the letter 'g'. This seems to only occur when `set nottimeout` is enabled. The inconsistency suggests a startup race condition.

### Steps to reproduce

`nvim --clean --cmd 'set nocompatible' --cmd 'set nottimeout' ~/.vimrc`

Removing the `--cmd 'set nottimeout'` seems to greatly reduce if not eliminate the issue.

### Expected behavior

I expect the file to appear in the buffer as it does on disk. The buffer state should initially be unmodified.

### Neovim version (nvim -v)

% nvim -v NVIM v0.10.0 Build type: Release LuaJIT 2.1.1713773202

no, vim 9.0.1544 (nor a previous forgotten nvim 0.9.x version)

### Operating system/version

macOS 11.7.10

### Terminal name/version

Reproduced with two: iTerm2, Wezterm

iTerm2:xterm-256color, Wezterm:wezterm

### Installation

brew install neovim

zeertzjq:
Problem:  'nottimeout' causes escape sequences to time out immediately
          instead having no timeout.
Solution: Don't use termkey_getkeys_force().

Fix #29047

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -115,6 +115,8 @@ LUA
 OPTIONS
 
 • 'shelltemp' defaults to "false".
+• 'nottimeout' now consistently makes Nvim wait indefinitely for a complete
+  sequence instead of sometimes behaving as if 'ttimeoutlen' is 0.
 
 PLUGINS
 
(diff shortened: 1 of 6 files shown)


zeertzjq:
Hmm, the behavior is still different from Vim.

zeertzjq:
It's not possible to match the behavior of Vim because Vim handles this in getchar.c, which has some special handling for ESC.

zeertzjq:
Actually this still isn't worth it.

PR #29159, opened 2024-06-03
docs(luv): Clarify types in luv docs and remove duplicate section

adigitoleo:
Some libuv filesystem functions expect the numeric `mode` argument to be an octal representation of the equivalent `chmod(1)` mode. This is potentially surprising behaviour to anyone who hasn't used either `libuv` or `glibc` from C. The documentation for e.g. `fs_mkdir` in libuv defers to `mkdir(2)` which defers to `mode_t(3)` which only says "It is an integer type". Furthermore, `fs_access` uses a different `mode` type, validated using
[luv_check_amode](https://github.com/luvit/luv/blob/7233e6dea92498a244feb51b790c1ba51e8abbff/src/fs.c#L184), which can accept strings ("symbolic mode" in `chmod(1)`). I chose not to mention that this one still technically accepts an integer because I have no idea about the restrictions of said integer.

### DIFF
--- a/runtime/doc/luvref.txt
+++ b/runtime/doc/luvref.txt
@@ -2743,7 +2743,7 @@ uv.fs_open({path}, {flags}, {mode} [, {callback}])                *uv.fs_open()*
                 Parameters:
                 - `path`: `string`
                 - `flags`: `string` or `integer`
-                - `mode`: `integer`
+                - `mode`: `integer` (octal representation of `chmod(1)` mode)
                 - `callback`: `callable` (async version) or `nil` (sync
                   version)
                   - `err`: `nil` or `string`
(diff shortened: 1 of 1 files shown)


dundargoc:
Heads up: these are taken verbatim from luv documentation. You should change that first if you haven't.

adigitoleo:
https://github.com/luvit/luv/pull/703

dundargoc:
I think we can go ahead an close this for the time being. We will update the documentation when we bump the luv version that includes the documentation changes anyway, and it makes more sense to include the changes together with the luv version bump. Thanks for the PR though, much appreciated!

adigitoleo:
That's fine. For posterity: note that the duplicated section was unique to the NeoVim helpfile.

adigitoleo:
Done here: #29161

PR #29160, opened 2024-06-03
fix(gx): allow `@` in url

dundargoc:
This will make `gx` work for links for the form
https://hachyderm.io/@neovim.

### DIFF
--- a/runtime/lua/vim/ui.lua
+++ b/runtime/lua/vim/ui.lua
@@ -179,7 +179,13 @@ function M._get_url()
       current_node = current_node:parent()
     end
   end
-  return vim.fn.expand('<cfile>')
+
+  local old_isfname = vim.o.isfname
+  vim.cmd [[set isfname+=@-@]]
+  local url = vim.fn.expand('<cfile>')
+  vim.o.isfname = old_isfname
+
+  return url
 end
 
 return M


runtime/lua/vim/ui.lua
```diff
@@ -179,7 +179,13 @@ function M._get_url()
       current_node = current_node:parent()
     end
   end
-  return vim.fn.expand('<cfile>')
+
+  local old_isfname = vim.o.isfname
+  vim.cmd [[set isfname+=@-@]]
+  local url = vim.fn.expand('<cfile>')
+  vim.o.isfname = old_isfname
```
justinmk, line 186: off topic: we could introduce an internal/private `vim._with()` https://github.com/neovim/neovim/issues/19832 to start discovering places where it's useful and figure out its interface.

justinmk (approved):

runtime/lua/vim/ui.lua
```diff
@@ -179,7 +179,13 @@ function M._get_url()
       current_node = current_node:parent()
     end
   end
-  return vim.fn.expand('<cfile>')
+
+  local old_isfname = vim.o.isfname
+  vim.cmd [[set isfname+=@-@]]
+  local url = vim.fn.expand('<cfile>')
```
wookayin, line 185: This approach is not quite perfect because L185 can often throw an error:
```
Vim:E446: No file name under cursor
```
in which case `isfname` will never be recovered. We'll probably need to protect call for this kind of approach unless it's guaranteed to not throw.

runtime/lua/vim/ui.lua
dundargoc: https://github.com/neovim/neovim/pull/29244

runtime/lua/vim/ui.lua
dundargoc: This will now be protected in https://github.com/neovim/neovim/pull/29244 so this won't be a problem after that.

PR #29161, opened 2024-06-03
docs(luv): remove duplicate `fs_chmod` section

adigitoleo:
~~It is already documented on L3096.~~

Nope, they are `chmod` and `fchmod`, I need more sleep.


dundargoc:
> It is already documented on L3096.

That line is about `fs_chmod` though, while this is removing `fs_fchmod`. I think the current docs are correct actually.

adigitoleo:
Woops, my bad, missed the difference.

PR #29162, opened 2024-06-03
fix(lsp): check if buffer is valid before LspDetach autocmd

coffebar:
Preventing `Invalid buffer id` error when loading saved sessions.

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -393,7 +393,7 @@ local function on_client_exit(code, signal, client_id)
 
   vim.schedule(function()
     for bufnr in pairs(client.attached_buffers) do
-      if client and client.attached_buffers[bufnr] then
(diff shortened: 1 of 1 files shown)


mfussenegger:
Could you add a test case for this?

coffebar:
It is difficult for me to do a functional test. There must be some kind of asynchronous solution in the test.
I can tell you about use-case when it happens to me. 
I'm using a [project-manager](https://github.com/coffebar/neovim-project) plugin, that uses `:mksession` under the hood. Each project has 2-5 lsp servers, 5-20 opened buffers. When I switch these projects, sometimes it can be no errors, sometimes 2 or more of these. It started from v0.10, but I don't have a clear understanding how this happens. At least I believe that no matter how bad plugins are, they should not be able to break the core.

icholy:
@coffebar did you manage to reproduce this on master before submitting this PR? Pretty sure it was already fixed in https://github.com/neovim/neovim/pull/29029

PR #29164, opened 2024-06-03
feat(lsp): include end_col and end_lnum in vim.lsp.buf.locations_to_items

tom-anders:
Noticed this by accident when after the Trouble v3 update, `:Trouble quickfix` would not display the correct range of items in the preview.

Might be worth a 0.10.0 backport? I assume many people upgraded to Trouble v3 and see this (currently only the first character of each item is highlighted in the preview, which looks a bit weird)

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -113,6 +113,7 @@ LSP
 
 • Completion side effects (including snippet expansion, execution of commands
   and application of additional text edits) is now built-in.
+• |vim.lsp.util.locations_to_items()| sets `end_col` and `end_lnum` fields.
 
 LUA
 
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -1721,7 +1721,9 @@ end)
 ---@inlinedoc
 ---@field filename string
 ---@field lnum integer 1-indexed line number
+---@field end_lnum integer 1-indexed end line number
 ---@field col integer 1-indexed column
+---@field end_col integer 1-indexed end column
 ---@field text string
 ---@field user_data lsp.Location|lsp.LocationLink
 
@@ -1748,7 +1750,7 @@ function M.locations_to_items(locations, offset_encoding)
   end
 
   local items = {}
-  ---@type table<string, {start: lsp.Position, location: lsp.Location|lsp.LocationLink}[]>
+  ---@type table<string, {start: lsp.Position, end: lsp.Position, location: lsp.Location|lsp.LocationLink}[]>
   local grouped = setmetatable({}, {
     __index = function(t, k)
       local v = {}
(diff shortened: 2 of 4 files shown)


runtime/doc/news.txt
```diff
@@ -113,6 +113,7 @@ LSP
 
 • Completion side effects (including snippet expansion, execution of commands
   and application of additional text edits) is now built-in.
+• |vim.lsp.util.locations_to_items()| correctly sets `end_col` field.
```
MariaSolOs, line 116: I wouldn't consider this a new feature (it almost feels like a bug fix), but I'm unsure.

MariaSolOs (approved):

runtime/doc/news.txt
mfussenegger: I kinda disagree because `end_col` in qf items is relatively new.

clason: Sorry, the "correctly" in the news threw me.

tom-anders:
Thanks for the quick feedback! I removed "correctly" from the news.txt entry, as discussed.

I also noticed that `end_lnum` was missing as well, so I added that as well.

tom-anders:
(I noticed that the calculation for `end_col` would possibly yield false results if `lnum!=end_lnum`, the last force-push fixes that and adapts the test accordingly

rockyzhang24:
Great. And I think the doc for `vim.lsp.util.location_to_items()` should be updated as well.

tom-anders:
> Great. And I think the doc for `vim.lsp.util.locations_to_items()` (its `Return`) should be updated as well.

Done!

One more thing: Right now we set `end_col` and `end_lnum` even if they are actually the same as `col` and `lnum` - This doesn't seem to cause any problems, but the docs of `setqflist` at least seem to indicate that `end_col`/`end_lnum` should only be set if they're different from `col`/`lnum`. Is it worth changing (and complicating) the code for this?

runtime/lua/vim/lsp/util.lua
```diff
@@ -1760,7 +1762,7 @@ function M.locations_to_items(locations, offset_encoding)
     -- locations may be Location or LocationLink
     local uri = d.uri or d.targetUri
     local range = d.range or d.targetSelectionRange
-    table.insert(grouped[uri], { start = range.start, location = d })
+    table.insert(grouped[uri], { start = range.start, ['end'] = range['end'], location = d })
```
rockyzhang24, line 1765: Should the annotation for `grouped` above be adjusted accordingly?
```diff
- ---@type table<string, {start: lsp.Position, location: lsp.Location|lsp.LocationLink}[]>
+ ---@type table<string, {start: lsp.Position, end: lsp.Position, location: lsp.Location|lsp.LocationLink}[]>
```

ofseed:
Found this is duplicated with https://github.com/neovim/neovim/pull/28724. Feel free to close it if you decided to merge this one.

justinmk (approved):

tom-anders:
whoops, I force pushed to add the type annotation suggested by @rockyzhang24 , but accidentally reverted another change (adding `end_col` and `end_lnum` to the docs) - The second force push fixes this. So, looks like this is ready to be merged?

PR #29165, opened 2024-06-03
docs(autocommand): fix example for last-position-jump

mikehaertl:


runtime/doc/usr_05.txt
```diff
@@ -120,7 +120,7 @@ This switches on three very clever mechanisms:
 				*restore-cursor* *last-position-jump*  >
     augroup RestoreCursor
       autocmd!
-      autocmd BufRead * autocmd FileType <buffer> ++once
+      autocmd BufRead * ++once
```
mikehaertl, line 123: Actually now that I review the change I'm not sure anymore. I thought the previous version was messed up by accident. Now I noticed the `++once`. Not sure what was the idea here.

In any case: The initial version did not work for me.

mikehaertl: It seems that `BufRead` is too late to attach to `FileType`. `BufReadPre` works fine, though. Commit updated.

justinmk (approved):

dundargoc:
Thank you for your contribution! We collect all small documentation fixes into a single pull request and merge it once it gets big enough: https://github.com/neovim/neovim/pull/29410

PR #29166, opened 2024-06-03
refactor(gx): modify isfname with lua method

atusy:
a9c89bcbf uses `vim.cmd [[set isfname+=@-@]]`, but `vim.opt.isfname:append('@')` fits the Neovim way.


wookayin:
Ref #29160

PR #29195, opened 2024-06-04
Update BUILD.md in regards to nightly builds on Nix

vE5li:
Since [0e9c92a](https://github.com/neovim/neovim/commit/0e9c92a900435fb09440398674687e45cc802b08), the information in `BUILD.md` in regards to Neovim nightly on Nix is no longer accurate.
I think we should instead refer to `github:nix-community/neovim-nightly-overlay`.
Alternatively we can just remove these paragraphs completely.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -365,13 +365,7 @@ and replacing `neovim-unwrapped` with `neovim-dev`:
 nix-shell '<nixpkgs>' -A neovim-dev
 ```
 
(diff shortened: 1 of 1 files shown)


dundargoc:
I think something like this is already included in the docs PR

dundargoc:
See https://github.com/neovim/neovim/pull/28837

PR #29196, opened 2024-06-04
fix(lsp): check for nil response from server

### LINKED ISSUE #27395: Omnisharp Lsp server return "null" not handled in Neovim
### Problem

Omnisharp version 1.39.11
Neovim version 0.9.5
Platform: Windows & Linux (not tested on mac)

Omnisharp language server seems to return "null" on some communications.
This are currently not handled in neovim and therefore we see this error.
Would be interesting if someone who is more knowledge regarding LSP's could chime in on this.

There is an issue open in Omnisharp's github too regarding this: https://github.com/OmniSharp/omnisharp-roslyn/issues/2574

I have pinpointed the "error" to Client:handle_body function in rpc.lua
the pcall to vim.json.decode returns ok and therefore the code will continue and start checking underlying properties on decoded (that is nil) variable.

```lua
  local ok, decoded = pcall(vim.json.decode, body, { luanil = { object = true } })
  if not ok then
    self:on_error(client_errors.INVALID_SERVER_JSON, decoded)
    return
  end
  local _ = log.debug() and log.debug('rpc.receive', decoded)

  if type(decoded.method) == 'string' and decoded.id then
  -- ...
```

If we would include an check if decoded is nil and return the error stop happening. But I'm unsure if that is an expected behavior.

```lua
  local ok, decoded = pcall(vim.json.decode, body, { luanil = { object = true } })
  -- ...

  if decoded == nil then
    return
  end
  if type(decoded.method) == 'string' and decoded.id then
  -- ...
```

Would be interesting if someone who is more knowledge regarding LSP's could chime in on this.

Stacktrace:
```text
Error executing luv callback:
C:\opt\nvim\share\nvim\runtime/lua/vim/lsp/rpc.lua:393: attempt to index local 'decoded' (a nil value)
stack traceback:
        C:\opt\nvim\share\nvim\runtime/lua/vim/lsp/rpc.lua:393: in function 'handle_body'
        C:\opt\nvim\share\nvim\runtime/lua/vim/lsp/rpc.lua:743: in function 'handle_body'
        C:\opt\nvim\share\nvim\runtime/lua/vim/lsp/rpc.lua:263: in function <C:\opt\nvim\share\nvim\runtime/lua/vim/lsp/rpc.lua:247>
```

open an .cs file. it is a ~70% chance that Omnisharp lsp server will return null

### Expected behavior

_No response_

### Neovim version (nvim -v)

0.9.5

Omnisharp 1.39.11

### Operating system/version

Windows 11

### Log file

_No response_

TamaMcGlinn:
This fixes #27395. OmniSharp (and possibly other LSP servers) can sometimes return a nil value, which rpc.lua chokes on with this error message:

```
Error executing luv callback:                                                                                                                                                                              
...nt_nvimGcA7m0/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:410: attempt to index local 'decoded' (a nil value)                                                                                                      
stack traceback:                                                                                                                                                                                                     
        ...nt_nvimGcA7m0/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:410: in function 'handle_body'                                                                                                                   
        ...nt_nvimGcA7m0/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:761: in function 'handle_body'                                                                                                                   
        ...nt_nvimGcA7m0/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:267: in function <...nt_nvimGcA7m0/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:251>
```

We should not crash, no matter what the LSP server throws at us. Note that #27148 would be even worse; silently ignoring the error. Rather, we should provide a proper error message stating the LSP server gave us something invalid, so that it is clear that the error is with the LSP server, rather than being a crash inside NeoVim runtime scripts. We are already doing a lot of validation in rpc.lua, it's just that nil was being overlooked here.

By adding a check first that `decoded` isn't nil, we end up falling back to the else case:

```
self:on_error(M.client_errors.INVALID_SERVER_MESSAGE, decoded)
```

which is going to correctly inform the user that the LSP server is broken, not NeoVim.

### DIFF
--- a/runtime/lua/vim/lsp/rpc.lua
+++ b/runtime/lua/vim/lsp/rpc.lua
@@ -407,7 +407,9 @@ function Client:handle_body(body)
   end
   log.debug('rpc.receive', decoded)
 
-  if type(decoded.method) == 'string' and decoded.id then
+  if type(decoded) ~= 'table' then
+    self:on_error(M.client_errors.INVALID_SERVER_MESSAGE, decoded)
+  elseif type(decoded.method) == 'string' and decoded.id then
     local err --- @type lsp.ResponseError|nil
     -- Schedule here so that the users functions don't trigger an error and
     -- we can still use the result.


icholy:
You could make this more robust by checking if it's an actual table instead of just not nil.

``` lua
if type(decoded) ~= "table" then
	self:on_error(M.client_errors.INVALID_SERVER_MESSAGE, decoded)
elseif --- etc ...
```

runtime/lua/vim/lsp/rpc.lua
```diff
@@ -407,7 +407,7 @@ function Client:handle_body(body)
   end
```
wookayin, line 407: Well, I think instead of such (repetitive) nil checks, we should handle the invalid response and throw early. LSP is not supposed to return a NIL as a RPC response, so if `decoded == nil` then we should raise an error as early as here.

TamaMcGlinn:
@wookayin @icholy I incorporated your improvements. Any else needed to merge this?

justinmk (approved):

justinmk:
is the commit message up to date? the PR description looks like it has useful details that are missing in the commit message

MariaSolOs (approved):

MariaSolOs:
Slightly unrelated question but this PR makes me wonder: Can we distinguish a decoding error versus a legitimate invalid server response?

mfussenegger:
> > Can we distinguish a decoding error versus a legitimate invalid server response?
> 
> We already do; if there was an error decoding the json we get INVALID_SERVER_JSON, whereas nil, although it is valid json, is an invalid server response, and then we get INVALID_SERVER_MESSAGE.
> 
> But I agree that we should specify further and add "INVALID_SERVER_MESSAGE_NIL" for this case, since it does add some useful information; we want the user to immediately search for why their language server is giving nil, rather than just "something invalid". Please hold on merging until we've changed the messages.
> 
> I'm not sure the case of non-nil but still not a table will ever come up, but that could be yet another message: INVALID_SERVER_MESSAGE_NONTABLE. The original "invalid" meant that decoded.method was nil or not a string, that should then also get a specific name: INVALID_SERVER_MESSAGE_NO_METHOD.

Please don't make this more complex than necessary. The existing error codes are sufficient.

runtime/lua/vim/lsp/rpc.lua
```diff
@@ -407,7 +407,10 @@ function Client:handle_body(body)
...
   log.debug('rpc.receive', decoded)
 
-  if type(decoded.method) == 'string' and decoded.id then
+  if type(decoded) ~= 'table' then
+    -- Invalid server message
```
mfussenegger, line 411:
```suggestion
```

The next line says `on_error(M.client_errors.INVALID_SERVER_MESSAGE` making this pretty obvious

icholy: @mfussenegger FYI https://github.com/neovim/neovim/blob/37bf4c572a8fa20dc4a8433524ecc8c8d68fe53c/runtime/lua/vim/lsp/rpc.lua#L526-L527

MariaSolOs:
> Please don't make this more complex than necessary. The existing error codes are sufficient.

Agreed, I wasn't suggesting that we add more error codes but I was just wondering if we're able to correctly separate a client-side error (JSON decoding) from a truly misbehaving server.
Either way, changes for that are out of scope for this PR.

PR #29197, opened 2024-06-04
fix(tui): move $COLORTERM check to _defaults.lua

### LINKED ISSUE #28776: DECRQSS +q5463;524742;73657472676266;73657472676262$qm appears in terminal
### Problem

When I start nvim 0.10.0 I see `+q5463;524742;73657472676266;73657472676262$qm` appear at the top left of the terminal window, with the `$qm` in italic and darker. After a fraction of a second it disappears and the regular initial window contents appear.

Using a terminal to a remote Mac, I see `p$qm+q5463;524742;73657472676266;73657472676262$qm+q4D73` with the `$qm+q4D73` in dark italic.

I didn't see this with 0.9.5.

It still does it with `nvim --clean`.

I'm guessing this is due to DECRQSS terminal escapes that aren't implemented by macOS terminal. Apparently [they aren't implemented for alacritty either](https://github.com/alacritty/alacritty/issues/7184).

### Steps to reproduce

`nvim --clean`

### Expected behavior

Initial `NVIM v0.10.0` blank window appears.

### Neovim version (nvim -v)

NVIM v0.10.0

no, 9.0.2142

### Operating system/version

macOS 14.5

### Terminal name/version

macOS terminal

xterm-256color

### Installation

Homebrew

gpanders:
We currently check $COLORTERM in the TUI process to determine if the terminal supports 24 bit color (truecolor). If $COLORTERM is "truecolor" or "24bit" then we automatically assume that the terminal supports truecolor, but if $COLORTERM is set to any other value we still query the terminal.

The `rgb` flag of the UI struct is a boolean which only indicates whether the UI supports truecolor, but does not have a 3rd state that we can use to represent "we don't know if the UI supports truecolor". We currently use `rgb=false` to represent this "we don't know" state, and we use XTGETTCAP and DECRQSS queries to determine at runtime if the terminal supports truecolor. However, if $COLORTERM is set to a value besides "truecolor" or "24bit" (e.g. "256" or "16) that is a clear indication that the terminal _does not_ support truecolor, so it is incorrect to treat `rgb=false` as "we don't know" in that case.

Instead, in the TUI process we only check for the terminfo capabilities. This must be done in the TUI process because we do not have access to this information in the core Neovim process when `_defaults.lua` runs. If the TUI cannot determine truecolor support from terminfo alone, we set `rgb=false` to indicate "we don't know if the terminal supports truecolor yet, keep checking". When we get to `_defaults.lua`, we can then check $COLORTERM and only query the terminal if it is unset.

This means that users can explicitly opt out of truecolor determination by setting `COLORTERM=256` (or similar) in their environment.

Fixes: #28776

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -470,10 +470,14 @@ do
     --- response indicates that it does support truecolor enable 'termguicolors',
     --- but only if the user has not already disabled it.
     do
-      if tty.rgb then
-        -- The TUI was able to determine truecolor support
+      local colorterm = os.getenv('COLORTERM')
+      if tty.rgb or colorterm == 'truecolor' or colorterm == '24bit' then
+        -- The TUI was able to determine truecolor support or $COLORTERM explicitly indicates
+        -- truecolor support
         setoption('termguicolors', true)
-      else
+      elseif colorterm == nil or colorterm == '' then
+        -- Neither the TUI nor $COLORTERM indicate that truecolor is supported, so query the
+        -- terminal
         local caps = {} ---@type table<string, boolean>
         require('vim.termcap').query({ 'Tc', 'RGB', 'setrgbf', 'setrgbb' }, function(cap, found)
           if not found then
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -1855,20 +1855,12 @@ static int unibi_find_ext_bool(unibi_term *ut, const char *name)
   return -1;
 }
 
-/// Determine if the terminal supports truecolor or not:
+/// Determine if the terminal supports truecolor or not.
 ///
-/// 1. If $COLORTERM is "24bit" or "truecolor", return true
-/// 2. Else, check terminfo for Tc, RGB, setrgbf, or setrgbb capabilities. If
-///    found, return true
-/// 3. Else, return false
+/// If terminfo contains Tc, RGB, or both setrgbf and setrgbb capabilities, return true.
 static bool term_has_truecolor(TUIData *tui, const char *colorterm)
 {
-  // Check $COLORTERM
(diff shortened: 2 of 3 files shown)


src/nvim/tui/tui.c
```diff
@@ -1857,10 +1857,7 @@ static int unibi_find_ext_bool(unibi_term *ut, const char *name)
...
 /// Determine if the terminal supports truecolor or not:
 ///
-/// 1. If $COLORTERM is "24bit" or "truecolor", return true
-/// 2. Else, check terminfo for Tc, RGB, setrgbf, or setrgbb capabilities. If
-///    found, return true
-/// 3. Else, return false
+/// If terminfo contains Tc, RGB, or both setrgbf and setrgbb capabilities, return true.
 static bool term_has_truecolor(TUIData *tui, const char *colorterm)
 {
   // Check $COLORTERM
```
zeertzjq, line 1863: Why are there only comment changes?

zeertzjq:
Maybe it's possible to add a test?

src/nvim/tui/tui.c
gpanders: Not sure what happened there, but fixed now.

runtime/lua/vim/_defaults.lua
```diff
@@ -470,10 +470,14 @@ do
...
     do
-      if tty.rgb then
-        -- The TUI was able to determine truecolor support
+      local colorterm = os.getenv('COLORTERM')
+      if tty.rgb or colorterm == 'truecolor' or colorterm == '24bit' then
+        -- The TUI was able to determine truecolor support or $COLORTERM explicitly indicates
+        -- truecolor support
         setoption('termguicolors', true)
-      else
+      elseif colorterm == nil then
```
zeertzjq, line 478:
```suggestion
      elseif colorterm == nil or colorterm == '' then
```

PR #29198, opened 2024-06-05
docs(lua): add return type to vim.str_byteindex

icholy:


icholy:
@dundargoc for future reference, does a PR like that always exist?

dundargoc:
Ye. It should be automated but I haven't gotten around to fixing that yet.

PR #29207, opened 2024-06-05
feat: command parsing recognizes visual range

mg979:
Problem: it's not possible to determine if a command is run over a visual range, rather than an ordinary range (line1, line2). Knowing we're handling a visual range is important, so that we can read the marks to know also the columns at which the range starts and ends.

Solution: introduce a `visualrange` boolean in the output of nvim_parse_cmd, and also pass it to lua callbacks for user commands, in the arguments table.

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -1902,6 +1902,8 @@ nvim_parse_cmd({str}, {opts})                               *nvim_parse_cmd()*
           if command doesn't accept a range. Otherwise, has no elements if no
           range was specified, one element if only a single range item was
           specified, or two elements if both range items were specified.
+        • visualrange: (boolean) (optional) Whether the range is the visual
+          selection ("'<,'>").
         • count: (number) (optional) Command <count>. Omitted if command
           cannot take a count.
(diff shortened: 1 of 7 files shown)


mg979:
Or maybe it would be better to pass the range as string, so that the user can parse it? It would also cover `%`, `.` and empty string (that could be handled differently from `current line` that could be the default).

zeertzjq (dismissed):
There may be whitespace before or inside the command address, and the Visual area can also be indicated by `:*`, so this doesn't always work.

zeertzjq (changes requested):
This is just a simple comparison, so why not just pass the full Ex command as another field to Lua user command callbacks instead of adding a special-cased field for Visual range only?

mg979:
> This is just a simple comparison, so why not just pass the full Ex command as another field to Lua user command callbacks instead of adding a special-cased field for Visual range only?

Yes it makes more sense to pass the full command line. Or at least the part that comes before the command name, since the rest can be deduced from the rest (`args.name`, `args.fargs` etc). Moreover this would fail for more complex command lines that have other stuff before that range.

PR #29212, opened 2024-06-06
fix(lsp): fix reverse sorting of same position text edits

### LINKED ISSUE #29202: Text edits with the same start position are applied in reverse
### Problem

Text edits returned from the LSP that have the same start position are being applied in reverse. For example:

For the following dummy text,
```elixir
defmodule Dummy do
  alias B
  alias A

  def hello do
  end
end
```

and following edits:
```
[DEBUG][2024-06-02 08:39:45] .../vim/lsp/rpc.lua:387	"rpc.receive"	{
  id = 10,
  jsonrpc = "2.0",
  result = { {
      edit = {
        changes = {
          ["file:///Users/scottming/Code/dummy/lib/dummy.ex"] = { {
              newText = "",
              range = {
                ["end"] = {
                  character = 0,
                  line = 3
                },
                start = {
                  character = 0,
                  line = 2
                }
              }
            }, {
              newText = "\n",
              range = {
                ["end"] = {
                  character = 0,
                  line = 2
                },
                start = {
                  character = 0,
                  line = 1
                }
              }
            }, {
              newText = "  alias A\n  alias B",
              range = {
                ["end"] = {
                  character = 0,
                  line = 1
                },
                start = {
                  character = 0,
                  line = 1
                }
              }
            } }
        }
      },
      kind = "source.organizeImports",
      title = "Organize aliases"
    } }
}
```

Results in the following:
```elixir
defmodule Dummy do

  alias B  alias B

  def hello do
  end
end
```

According to the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEditArray), text edits with the same start position should be applied in the order received.

Also, in the [vscode implementation](https://github.com/microsoft/vscode-languageserver-node/blob/main/textDocument/src/main.ts#L371), similar start position text edits aren't sorted in reverse order

Additional context can be found on an issue filed on the lexical repo
https://github.com/lexical-lsp/lexical/issues/765

Using this minimal init,
```lua
--- CHANGE THESE
local pattern = { 'elixir', 'eelixir' }
local cmd = { '/path/to/lexical/_build/dev/package/lexical/bin/start_lexical.sh' }

  pattern = pattern,
  callback = function(args)
    vim.lsp.start({
      name = 'bugged-ls',
      cmd = cmd,
      root_dir = root_dir,
      settings = settings,
    })
  end,
})
```

and this dummy file:
```elixir
# dummy.exs
defmodule Dummy do
  alias B
  alias A

  def hello do
  end
end
```

1. Open up the dummy file with the command `nvim -u minimal_init.lua dummy.exs`
2. Navigate to the first alias and run the command `:lua vim.lsp.buf.code_action()`
3. Run the `Organize aliases` code action to sort the aliases

### Expected behavior

The code action should result in the file:
```elixir
# dummy.exs
defmodule Dummy do
  alias A
  alias B

  def hello do
  end
end
```

### Neovim version (nvim -v)

v0.10.0 release

lexical v0.6.1

### Operating system/version

macos 14.4.1

### Log file

https://gist.github.com/alcolmenar/fd784a80a8f91c6a254e74cb3c1ed3b6

alcolmenar:
### Problem
Text edits with the same position (both line and character) were being
reverse sorted prior to being applied which differs from the lsp spec

### Solution
Change the sort order for just the same position edits

Fixes #29202

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -391,7 +391,7 @@ function M.apply_text_edits(text_edits, bufnr, offset_encoding)
       return a.range.start.character > b.range.start.character
     end
     if a._index ~= b._index then
-      return a._index > b._index
+      return a._index < b._index
     end
   end)
 
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -1794,9 +1794,9 @@ describe('LSP', function()
       }
       exec_lua('vim.lsp.util.apply_text_edits(...)', edits, 1, 'utf-16')
       eq({
-        '',
-        '123',
-        'fooFbar',
+        '3',
+        'foo',
+        '12Fbar',
         '123irst guy',
         'baz line of text',
         'The next line of text',
@@ -1818,9 +1818,9 @@ describe('LSP', function()
       }
       exec_lua('vim.lsp.util.apply_text_edits(...)', edits, 1, 'utf-16')
       eq({
-        '',
-        '123',
-        'fooFbar',
+        '3',
+        'foo',
+        '12Fbar',
         '123irst guy',
         'baz line of text',
         'The next line of text',


megalithic:
interesting, why were those text edits being sorted in the first place? anyone core contribs know?

PR #29216, opened 2024-06-06
refactor: inline short functions

IAKOBVS:
These are short wrapper functions that call other functions. For short strings, inlining them should make them faster because we're avoiding the overhead of calling the wrapper function, as if they were a macro.

### DIFF
--- a/src/nvim/memory.c
+++ b/src/nvim/memory.c
@@ -223,61 +223,6 @@ void *xmemdupz(const void *data, size_t len)
   return memcpy(xmallocz(len), data, len);
 }
 
-/// Copies `len` bytes of `src` to `dst` and zero terminates it.
-///
-/// @see {xstrlcpy}
-/// @param[out]  dst  Buffer to store the result.
-/// @param[in]  src  Buffer to be copied.
-/// @param[in]  len  Number of bytes to be copied.
-void *xmemcpyz(void *dst, const void *src, size_t len)
-  FUNC_ATTR_NONNULL_ALL FUNC_ATTR_NONNULL_RET
-{
-  memcpy(dst, src, len);
-  ((char *)dst)[len] = NUL;
-  return dst;
-}
-
-#ifndef HAVE_STRNLEN
-size_t xstrnlen(const char *s, size_t n)
-  FUNC_ATTR_NONNULL_ALL FUNC_ATTR_PURE
-{
-  const char *end = memchr(s, NUL, n);
-  if (end == NULL) {
-    return n;
-  }
-  return (size_t)(end - s);
-}
-#endif
-
-/// A version of strchr() that returns a pointer to the terminating NUL if it
-/// doesn't find `c`.
-///
-/// @param str The string to search.
-/// @param c   The char to look for.
-/// @returns a pointer to the first instance of `c`, or to the NUL terminator
-///          if not found.
-char *xstrchrnul(const char *str, char c)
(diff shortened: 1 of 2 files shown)


dundargoc:
You'd think an optimizing compiler would easily figure out low-hanging optimizations like this on its own. Not that I'm against this change per se, but I would personally be more comfortable with some data in the form of benchmarking before changing things around. That said others in the team know C way more so I'll defer to them if they disagree with my assessment.

clason:
To add to that, these refactors are not free because 1. they introduce churn (make bisecting/blaming harder) and 2. introduce possible conflict with upstream Vim, which makes porting patches (even) harder.

So we tend to be not very welcoming of "I was bored and it sounded like a good idea" refactors and instead ask for objective justification. (There is a reason why we expect PR descriptions to come in a "Problem -- Solution" form -- the "problem" part is extremely important since it allows us to weigh the benefits against the listed drawbacks.)

zeertzjq:
And please fix the lint failures.

famiu:
Agree with dundar and clason here. Micro-optimizations like this should be avoided, especially if it's reasonable to assume that the compiler can handle it. Unless there is a concrete benchmark showing a significant performance increase, I am against this change

IAKOBVS:
> Micro-optimizations like this should be avoided, especially if it's reasonable to assume that the compiler can handle it.

If x.c includes memory.h, uses a function f from memory.h that is implemented in memory.c, and is compiled with memory.o, the compiler can not inline f in x.c because the implementation of f is invisible to the compiler when compiling x.c as memory.h only exposes the function declaration of f.

For example:
https://github.com/IAKOBVS/neovim-inline-demo

In the not-inlined directory, my_strlen which calls strlen is declared in f.h and defined in f.c. my_strlen is called in main.c and main.c is compiled with f.o. The objdump shows that main.c calls my_strlen, whereas in the inlined directory, where my_strlen is defined in f.h, it shows that main.c calls strlen, which is optimized out since the argument is a literal string.

famiu:
> > Micro-optimizations like this should be avoided, especially if it's reasonable to assume that the compiler can handle it.
> 
> If x.c includes memory.h, uses a function f from memory.h that is implemented in memory.c, and is compiled with memory.o, the compiler can not inline f in x.c because the implementation of f is invisible to the compiler when compiling x.c as memory.h only exposes the function declaration.
> 
> For example:
> https://github.com/IAKOBVS/neovim-inline-demo
> 
> In the not-inlined directory, my_strlen which calls strlen is declared in f.h and defined in f.c. my_strlen is called in main.c and main.c is compiled with f.o. The objdump shows that main.c still calls my_strlen, whereas in the inlined directory, where my_strlen is defined in f.h, it shows that main.c calls strlen instead of my_strlen, and the strlen is optimized out in this case since the argument is a literal string.

Okay, but what's the actual overhead of not inlining them? Are there any benchmarks that show a significant performance change?

dundargoc:
> For example:
https://github.com/IAKOBVS/neovim-inline-demo

You didn't enable LTO which should take care of that. I'm not sure if it does though, I can test at home.

In any case: while there's nothing wrong with this PR (aka there's no real harm in merging this IMO) I think it's the wrong way to go to battle performance problems. You've made a a couple of performance-related PR so it seems it's something you care about. I would personally advice that you make a more targetted approach: identify features with performance problems (or look for issues with the `performance` label) and identify and fix the hot spots for these. I think that would be much more useful, and also make benchmarking easier as you've already identified what to benchmark (the slow feature).

PR #29219, opened 2024-06-06
refactor: use S_LEN macro

IAKOBVS:
Use S_LEN(s) instead of writing s, n where n is a literal number for the length of s which may be error-prone. Also prefer memmove over strmove when the length of the source string is known to avoid a strlen.


zeertzjq:
Please fix the lint failures.

PR #29220, opened 2024-06-06
feat: store the parsed command range as string

mg979:
Problem: it's not possible to determine if a command is run over a visual range, rather than an ordinary range (line1, line2). Knowing we're handling a visual range can be important, since we can read the marks to know also the columns at which the range starts and ends.

Solution: store the parsed range as string in the `rangestr` key, so that the user can access it in the lua table that is passed to the command callback as argument.

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -1902,6 +1902,8 @@ nvim_parse_cmd({str}, {opts})                               *nvim_parse_cmd()*
           if command doesn't accept a range. Otherwise, has no elements if no
           range was specified, one element if only a single range item was
           specified, or two elements if both range items were specified.
+        • rangestr: (string) (optional) The parsed range extracted from the
+          command line.
         • count: (number) (optional) Command <count>. Omitted if command
           cannot take a count.
         • reg: (string) (optional) Command <register>. Omitted if command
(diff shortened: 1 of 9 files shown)


mg979:
Alternative to https://github.com/neovim/neovim/pull/29207.

justinmk:
https://github.com/neovim/neovim/pull/29207#pullrequestreview-2100420758 suggested returning the full Ex command, not only the range part.

> Knowing we're handling a visual range can be important

Could we have a enum-like field that could represent the kind of range (potentially supporting "V" and "^v" in the future, not only "v", as well as marks/extmarks...)? It is planned "some day" to support columnwise range (e.g. with backticks):

    :`<,`>s/foo/bar

mg979:
> https://github.com/neovim/neovim/pull/29207#pullrequestreview-2100420758 suggested returning the full Ex command, not only the range part.

Yes, I thought that the full command line could be very long for some commands and just the range would be enough (and easier to parse), since you get the rest of the command line in other fields, but I can use the full command line if it makes more sense (but it does? @zeertzjq).

> Could we have a enum-like field that could represent the kind of range

You can get that with `visualmode()`, once you're sure that you're handling a visual range, that is. For example with:
```lua
-- test from some visual selection
vim.api.nvim_create_user_command('Testcmd', function(args)
  print(args.rangestr:find("'<,'>") and vim.fn.visualmode() or 'not visual mode')
end, { range = true })
```

>  It is planned "some day" to support columnwise range (e.g. with backticks)

Once the range is available as string, it would be enough to allow backticks in the range (treating them just like simple quotes), then let the user parse the range to see what it actually is.

mg979:
An alternative would be to have fields also for columns (not just lines) for the range, inside `struct exarg`. When the range is parsed and its lines are determined, also columns could be, if the kind of range allows it (and backticks are supported). But it's more complex.

PR #29232, opened 2024-06-07
refactor: quadratic behavior in vim_findfile_stopdir

IAKOBVS:
As in #27827, vim_findfile_stopdir uses STRMOVE in a loop and has a worst-case of O(n^2). The new implementation copies characters in-place and has a worst-case of O(n).

### DIFF
--- a/src/nvim/file_search.c
+++ b/src/nvim/file_search.c
@@ -527,6 +527,7 @@ start:
   *dst = NUL;
   if (*buf == ';') {
 is_semicolon:
+    *buf = NUL;
     buf++;
   } else {  // if (*buf == NUL)
 is_nul:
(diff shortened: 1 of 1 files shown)


src/nvim/file_search.c
```diff
@@ -505,24 +505,35 @@ void *vim_findfile_init(char *path, char *filename, char *stopdirs, int level, i
...
+      *dst++ = *buf++;
     }
-    r_ptr++;
   }
-  if (*r_ptr == ';') {
-    *r_ptr = 0;
-    r_ptr++;
-  } else if (*r_ptr == NUL) {
-    r_ptr = NULL;
+  *dst = NUL;
```
zeertzjq, line 527:
```suggestion
  assert(dst != buf);
  *dst = NUL;
```

zeertzjq (approved):

PR #29234, opened 2024-06-07
fix(lsp): apply LspInlayHint to inlay hint padding

ribru17:
Problem: The whitespace padding around inlay hints is not given a highlight group, making it confusing for colorschemes that apply a background color to inlay hints. The padding looks like a real space even though it is an inlay hint space.

Solution: Apply the content highlight to the padding characters.

Before:
beforeinlay
After:
afterinlay

### DIFF
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -350,11 +350,11 @@ api.nvim_set_decoration_provider(namespace, {
             end
             local vt = {} --- @type [string, string?][]
             if hint.paddingLeft then
-              vt[#vt + 1] = { ' ' }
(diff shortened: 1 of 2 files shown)


ribru17:
I think test failures are unrelated?

MariaSolOs:
> I think test failures are unrelated?

Yeah, the tests are flaky.

MariaSolOs:
Technically this change violates the protocol. The spec says "Padding should use the editor's background color, not the background color of the hint itself. That means padding can be used
to visually align/separate an inlay hint.": https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHint

ribru17:
Ah, I see. Thanks for clarifying!

PR #29236, opened 2024-06-08
fix(glob): handle overlapping `{}` condition elements

### LINKED ISSUE #28931: vim.glob.to_lpeg doesn't always work as expected
### Problem

Group conditions do not properly work when there are multiple groups which are the same at the beginning, for example, `vim.glob.to_lpeg('**/*.{tsx,ts}'):match('test.tsx')` matches, but `vim.glob.to_lpeg('**/*.{ts,tsx}'):match('test.tsx')` doesn't

### Steps to reproduce

1. `nvim --clean`
2. `:lua=vim.glob.to_lpeg('**/*.{tsx,ts}'):match('test.tsx')` (returns 9)
3. `:lua=vim.glob.to_lpeg('**/*.{ts,tsx}'):match('test.tsx')` (returns nil)

### Expected behavior

Both of the globs match.

### Neovim version (nvim -v)

v0.11.0-dev-57+gcd48b72b6

n/a

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.34.1

xterm-kitty

### Installation

build from repo

nojnhuh:
This change fixes an issue where glob patterns like `{a,ab}` would not match `ab` because the first option `a` matches, then the end of the string is expected but `b` is found, and LPeg does not backtrack to try the next option `ab` which would match. The fix here is to also append the rest of the pattern to the generated LPeg pattern for each option. This changes a glob `{a,ab}` from being parsed as

    ("a" or "ab") "end of string"

to

    ("a" "end of string" or "ab" "end of string")

Here, matching against `ab` would try the first option, fail to match, then proceed to the next option, and match.

The sacrifice this change makes is dropping support for nested `{}` conditions, which VSCode doesn't seem to support or test AFAICT.

Fixes #28931

### DIFF
--- a/test/functional/lua/glob_spec.lua
+++ b/test/functional/lua/glob_spec.lua
@@ -161,7 +161,7 @@ describe('glob', function()
       eq(false, match('{ab,cd}', 'a'))
       eq(true, match('{ab,cd}', 'cd'))
       eq(true, match('{a,b,c}', 'c'))
-      eq(true, match('{a,{b,c}}', 'c'))
+      eq(false, match('{a,{b,c}}', 'c')) -- {} cannot nest
     end)
 
     it('should match [] groups', function()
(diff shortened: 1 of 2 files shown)


test/functional/lua/glob_spec.lua
```diff
@@ -223,6 +223,17 @@ describe('glob', function()
       eq(true, match('{[0-9],[a-z]}', '0'))
       eq(true, match('{[0-9],[a-z]}', 'a'))
       eq(false, match('{[0-9],[a-z]}', 'A'))
+
+      -- glob is from willRename filter in typescript-language-server
+      -- https://github.com/typescript-language-server/typescript-language-server/blob/b224b878652438bcdd639137a6b1d1a6630129e4/src/lsp-server.ts#L266
+      eq(true, match('**/*.{ts,js,jsx,tsx,mjs,mts,cjs,cts}', 'test.js'))
+      eq(true, match('**/*.{ts,js,jsx,tsx,mjs,mts,cjs,cts}', 'test.ts'))
```
nojnhuh, line 230: Thank you @faergeek for the tests!

test/functional/lua/glob_spec.lua
```diff
@@ -161,7 +161,7 @@ describe('glob', function()
       eq(false, match('{ab,cd}', 'a'))
       eq(true, match('{ab,cd}', 'cd'))
       eq(true, match('{a,b,c}', 'c'))
-      eq(true, match('{a,{b,c}}', 'c'))
+      eq(false, match('{a,{b,c}}', 'c')) -- {} cannot nest
```
justinmk, line 164: does `:help vim.glob.to_lpeg()` need to mention this?

PR #29238, opened 2024-06-08
feat(help): use treesitter for table of contents

clason:
Problem: Creating the table of contents for `gO` is complicated.

Solution: Use treesitter instead.

We could normalize the toc entries (sentence case), but I think there'd be too many headings where that would not work. I'd also love to get rid of the non-breaking whitespace for indentation, but the loclist view seems to strip normal leading whitespace.


runtime/ftplugin/help.vim
```diff
@@ -21,77 +21,5 @@ endif
...
-        " Any line that's yelling is important.
-        let has_sub_section = 1
-        let level = has_section
-        let add_text = matchstr(text, '.\{-}\ze\s*\%([ \t]\+\*.\+\*\)\?$')
-      elseif text =~# '\~$'
-            \ && matchstr(text, '^\s*\zs.\{-}\ze\s*\~$') !~# '\t\|\s\{2,}'
-            \ && getline(lnum - 1) =~# '^\s*<\?$\|^\s*\*.*\*$'
-            \ && getline(lnum + 1) =~# '^\s*>\?$\|^\s*\*.*\*$'
-        " These lines could be headers or code examples.  We only want the
-        " ones that have subsequent lines at the same indent or more.
```
clason, line 65: And this is why we can't have nice things... 😞 

It's impossible to replace this sort of hackery with syntax tree parsing unless we take structured documentation a lot more seriously.

clason:
I think this is blocked by us finally biting the bullet and strictly separating tabular material from headings.

runtime/ftplugin/help.vim
dundargoc: What part of `text =~# '\~$ && matchstr(text, '^\s*\zs.\{-}\ze\s*\~$') !~# '\t\|\s\{2,} && getline(lnum - 1) =~# '^\s*<\?$\|^\s*\*.*\*$' && getline(lnum + 1) =~# '^\s*>\?$\|^\s*\*.*\*$'` do you find hacky?

clason: The fact that the function is building its own heading toc hierarchy based on heuristics instead of relying on syntax. (Understandably, given the way docs are written.)

dundargoc (approved):
LGTM

dundargoc:
The modelines sometimes show up in TOC for e.g. `news.txt` and `help.txt` and sometimes not, e.g. `starting.txt`. Unsure why there's a discrepancy.

clason:
Because these files are not formatted with proper parsing in mind. There's no hope for things like this to work robustly as long as the files are all over the place. (In this case, the delimiter line makes it a heading instead of a modeline.)

clason:
Hmm, I just realized that the different header levels (h1, h2, h3) are complete fan fiction by the vimdoc parser. Vim itself
* only documents `====` (h1) heading markers
* silently accepts `----` heading markers (as the same level!) as well in the syntax file
* silently highlights SCREAMING CASE as headlines as well.

Any attempt of generating a consistent hierarchical TOC is bound to fail.

justinmk:
> the different header levels (h1, h2, h3) are complete fan fiction by the vimdoc parser.

they are not "complete fiction", they are a reasonable starting point. Old files like syntax.txt are not authoritative, they just have quirks that are not present in newer help files.

This doesn't need to "block" anything if we get reasonable results for most help files.

runtime/ftplugin/help.vim
justinmk: Why is it "impossible"? `gen_help_html.lua` has workarounds by inspecting the text content.

justinmk (approved):

clason:
> they are not "complete fiction", they are a reasonable starting point

No, I stand by my point. Neither `:h help-writing` nor `syntax/help.vim` make these distinctions, and Vim syntax files use `====` and `----` as whim (vim?) dictates (but don't tend to mix).

They are certainly reasonable starting points for _defining_ our own flavor of "nvimdoc", but they are not descriptive of legacy "vimdoc". I'm all in favor of it, but then we should take it seriously and

1. start by defining a _full_ nvimdoc spec that makes sense to _us_;
2. (re)writing the parser to parse that _and only that_;
3. rewrite the legacy help files to align.

Otherwise we'll keep having to crawl over sandpaper to implement such features that rely on structural parsing.


(And personally, I think three levels are enough. I'd replace SCREAMING CASE headings with (a variant) of the Column headings ~ and define a custom syntax for tables/preformatted text.)

runtime/ftplugin/help.vim
clason: Yes, and this sort of "check the contents to see if it's _really_ a heading" is exactly what I want to replace with just being able to rely on a proper syntax tree. (I wrote "impossible to replace", not "impossible to do".)

justinmk:
> they are not descriptive of legacy "vimdoc"

The presence of some quirky old files like syntax.txt does not mean the table was flipped. Most help files use `====` and `----` as h1 and h2.

> start by defining a full nvimdoc spec that makes sense to us;

that is tree-sitter-vimdoc already.

> 2\. (re)writing the parser to parse that _and only that_;

What does this mean? The parser works already, we are using it to generate HTML.

> 3\. rewrite the legacy help files to align.

Sure, the h1 and h2 headings might need updating in a few files. 

The "rewrite everything!!" declarations sound more dramatic than what I would guess is needed.

clason:
Obviously I can't read, so I'll leave this in your capable hands. Proceed with this PR as you wish.

justinmk:
I tried this locally and compared the `gO` results from this PR vs the old vimscript implmentation on intro.txt, syntax.txt, eval.txt, and netrw.txt. The results with this PR are comparable or better than the old implementation, except for the known, fixable, non-blocking issue that modeline is incorrectly recognized as a header.

notes:
- intro.txt has a couple tables that need a small fixup to avoid being treated as h2. 
- netrw.txt: the new gO captures many more UPPERCASE headings that were missed by the old gO
- eval.txt has h2 headings that are now correctly recognized, unlike the old gO.
- syntax.txt results are nearly identical

PR #29239, opened 2024-06-08
feat(defaults): make `splitright` default option

dundargoc:
`splitright=true` makes more sense as the current buffer stays in the
same place when splitting the window. Otherwise the entire buffer is
switched on split which is a bit disorienting.

### DIFF
--- a/src/nvim/options.lua
+++ b/src/nvim/options.lua
@@ -7990,7 +7990,7 @@ return {
     },
     {
       abbreviation = 'spr',
-      defaults = { if_true = false },
+      defaults = { if_true = true },
       desc = [=[
         When on, splitting a window will put the new window right of the
         current one. |:vsplit|


clason:
Hmm, not sure I am excited about this change, especially `splitbelow` -- that would open `:help` under the current buffer, which feels... wrong? I think the editor layout varies so wildly that you cannot assume that one direction is uniformly better (even by a margin) than any other. In such a situation, you basically toss a coin -- which Vim did for us -- and live with it.

dundargoc:
I concede `splitbelow` is up for grabs, but I think `splitright` is absolutely an improvement. Let's ditch `splitbelow` and focus on `splitright` then, I think the improvement there is more clear-cut. Pivoting to only changing `splitright`.

antonk52:
While I see the reason to hesitate to change the defaults. The proposed change is the default is all modern popular nvim configuration distributions which have `splitbelow` and `splitright` enabled by default. 
- [LazyVim](https://github.com/LazyVim/LazyVim/blob/51854046dbb0bacf690ddc217cc76bf98c8ec1aa/lua/lazyvim/config/options.lua#L86-L88)
- [LunarVim](https://github.com/LunarVim/LunarVim/blob/85ccca97acfea9a465e354e18bb2f6109ba417f8/lua/lvim/config/settings.lua#L29-L30)
- [AstroNvim](https://github.com/AstroNvim/AstroNvim/blob/e1e08e29d2df6a2be92a012edba52a1b7f7eaa05/lua/astronvim/plugins/_astrocore_options.lua#L36-L37)
- [NvChad](https://github.com/NvChad/NvChad/blob/164e8cc7fcb9006a1edd4ddfc98bf8c7f4fe2e0d/lua/nvchad/options.lua#L37-L38)
- [mini.ai](https://github.com/echasnovski/mini.nvim/blob/1d4772e9f4f3ddc125a5ec1386fd4f85c7e3bb18/lua/mini/basics.lua#L465-L466)

runtime/lua/vim/_defaults.lua
```diff
@@ -577,4 +577,6 @@ do
     vim.o.grepprg = 'rg --vimgrep -uu '
     vim.o.grepformat = '%f:%l:%c:%m'
   end
+
+  vim.o.splitright = true
```
gpanders, line 581: Change the default value in options.lua

https://github.com/neovim/neovim/blob/9afa1fd35510c5fe485f4a1dfdabf94e5f051a1c/src/nvim/options.lua#L7993

justinmk:
I used to have 'splitright' but found that the current default is... equivalent. The "current window" to stays leftmost, which is presumably more comfortable. `:vsplit foo` surely means you want to work on "foo". With 'splitright', each vertical split moves your "primary" window further to the right.

@tpope may have better insights :)

tpope:
> I used to have 'splitright' but found that the current default is... equivalent, if not advantageous. With 'nosplitright', the "current window" to stays leftmost, which is presumably more comfortable. `:vsplit foo` surely means you want to work on "foo". With 'splitright', each vertical split moves your "primary" window further to the right, which is weird.
> 
> @tpope may have better insights :)

Another way to phrase this is that `splitright` shifts the content so you don't have to shift your eyes.

> This is one of those cases that does not seem like an overwhelming win.

This would be my opinion too. I mean I favor `'nosplitright'`, but if the discussion were the other way around, I would not be in support of rolling out such a disorienting change.

dundargoc:
Welp, this wasn't the easy-peasy, slamdunk default "everyone can agree on" that I initially thought. It was worth a shot, thanks for the input everyone.

PR #29244, opened 2024-06-08
refactor: use `vim._with` where possible

dundargoc:
This mostly means replacing `nvim_buf_call` and `nvim_win_call` with
`vim._with`.

### DIFF
--- a/runtime/lua/vim/ui.lua
+++ b/runtime/lua/vim/ui.lua
@@ -180,10 +180,9 @@ function M._get_url()
     end
   end
 
-  local old_isfname = vim.o.isfname
-  vim.cmd [[set isfname+=@-@]]
-  local url = vim.fn.expand('<cfile>')
-  vim.o.isfname = old_isfname
+  local url = vim._with({ go = { isfname = vim.o.isfname .. ',@-@' } }, function()
+    return vim.fn.expand('<cfile>')
+  end)
 
   return url
 end
--- a/test/functional/autocmd/autocmd_spec.lua
+++ b/test/functional/autocmd/autocmd_spec.lua
@@ -350,11 +350,11 @@ describe('autocmd', function()
     }
 
     -- Create specific layout and ensure it's left unchanged.
-    -- Use nvim_buf_call on a hidden buffer so aucmd_win is used.
+    -- Use vim._with on a hidden buffer so aucmd_win is used.
     exec_lua [[
       vim.cmd "wincmd s | wincmd _"
       _G.buf = vim.api.nvim_create_buf(true, true)
-      vim.api.nvim_buf_call(_G.buf, function() vim.cmd "wincmd J" end)
+      vim._with({buf = _G.buf}, function() vim.cmd "wincmd J" end)
     ]]
     screen:expect [[
       ^                                                  |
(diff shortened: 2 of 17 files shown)


test/functional/autocmd/autocmd_spec.lua
```diff
@@ -367,26 +367,26 @@ describe('autocmd', function()
...
+      vim._with({buf = _G.buf}, function()
         vim.fn.win_splitmove(vim.fn.winnr(), vim.fn.win_getid(1))
       end)
     ]]
     screen:expect_unchanged()
 
     -- Also check with nvim_win_set_config().
     matches(
-      ': Failed to move window %d+ into split$',
+      'Failed to move window %d+ into split$',
```
zeertzjq, line 385:
```suggestion
      '^Failed to move window %d+ into split$',
```

justinmk (approved):
Do we have (explicit) tests that verify that options/etc are restored even when `f()` throws an error?

dundargoc:
> Do we have (explicit) tests that verify that options/etc are restored even when `f()` throws an error?

https://github.com/neovim/neovim/pull/29269

dundargoc:
> Do we have (explicit) tests that verify that options/etc are restored even when `f()` throws an error?

https://github.com/neovim/neovim/pull/29280

runtime/lua/vim/ui.lua
```diff
@@ -180,10 +180,9 @@ function M._get_url()
     end
   end
 
-  local old_isfname = vim.o.isfname
-  vim.cmd [[set isfname+=@-@]]
-  local url = vim.fn.expand('<cfile>')
-  vim.o.isfname = old_isfname
+  local url = vim._with({ o = { isfname = vim.o.isfname .. ',@-@' } }, function()
```
echasnovski, line 183: I'd recommend using more targeted scopes for options. As `isfname` is global option, I'd use `go` instead of `o`.
It seems to make code more explicit while avoiding [extra step in implementation](https://github.com/neovim/neovim/blob/fc9b70826ec88ca2e6c0624c522b872e87aa7ac1/runtime/lua/vim/shared.lua#L1183) that decided which scope should be used.

test/functional/autocmd/autocmd_spec.lua
```diff
@@ -367,26 +367,26 @@ describe('autocmd', function()
...
+      vim._with({buf = _G.buf}, function()
         vim.fn.win_splitmove(vim.fn.winnr(), vim.fn.win_getid(1))
       end)
     ]]
     screen:expect_unchanged()
 
     -- Also check with nvim_win_set_config().
     matches(
-      ': Failed to move window %d+ into split$',
+      '^Failed to move window %d+ into split$',
```
echasnovski, line 385: Hmm... I wonder why this test changes was necessary? On paper `vim._with({ buf = buf }, f)` seems to work exactly as `vim.api.nvim_buf_call(buf, f)`.

echasnovski (approved):

test/functional/autocmd/autocmd_spec.lua
dundargoc: Just using different error functions I think. `vim._with` uses `nlua_push_errstr` while `nvim_buf_call` uses `api_set_error`. I don't think it makes a real difference which you use.

runtime/lua/vim/ui.lua
dundargoc: I don't think that's more correct. If `isfname` becomes global-local in the future, won't `vim.go` be incorrect (since we're changing the default/global value rather than the value of the buffer)?

echasnovski: *If* 'isfname' scopes are changed, *then* using `bo` or `wo` here will be better. Right now this option is sensible to be used with `go`, as its scope is global. Using `o` as scope here will only lead to more work done during this function execution with the same result as `go`.

That said, It is not particularly strong opinion per se.

dundargoc: OK, fixed.

PR #29250, opened 2024-06-09
refactor(comment): replace nvim_buf_set_lines with nvim_buf_set_text

altermo:
`nvim_buf_set_lines` is less precise, which causes an incorrect change in the extmarks position, so use `nvim_buf_set_text` instead.
Other benefits of this refactor:
+ `lockmarks = true` can be removed as regular marks should mostly work as intended.
+ `nvim_buf_attach` (and similar functions) reports only the regions that have changed, not the whole line.

### DIFF
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -132,45 +132,45 @@ end
 --- Make a function which comments a line
 ---@param parts vim._comment.Parts
 ---@param indent string
----@return fun(line: string): string
+---@return fun(line: string): [number,number,string][]
 local function make_comment_function(parts, indent)
-  local prefix, nonindent_start, suffix = indent .. parts.left, indent:len() + 1, parts.right
   local blank_comment = indent .. vim.trim(parts.left) .. vim.trim(parts.right)
 
   return function(line)
     if is_blank(line) then
-      return blank_comment
+      return { { 0, #line, blank_comment } }
     end
-    return prefix .. line:sub(nonindent_start) .. suffix
+    return { { #indent, #indent, parts.left }, { #line, #line, parts.right } }
   end
 end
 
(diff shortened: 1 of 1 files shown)


echasnovski:
Thanks for the PR.

Unfortunately, preserving extmarks is not a universally good side effect. I explored the idea in 'mini.comment' prior to making built-in commenting PR and decided that using `nvim_buf_set_lines()` is a better approach right now. [Here is full motivation](https://github.com/echasnovski/mini.nvim/blob/1d4772e9f4f3ddc125a5ec1386fd4f85c7e3bb18/lua/mini/comment.lua#L260-L287). TL;DR: preserving extmarks result into flickering (as some extmarks highlight text ranges) while calling `nvim_buf_set_text()` twice per line instead of a single `nvim_buf_set_lines()` can be costly.

clason:
> nvim_buf_attach (and similar functions) reports only the regions that have changed, not the whole line.

That part seems like a worthwhile change, though -- if it can be isolated.

echasnovski:
> > nvim_buf_attach (and similar functions) reports only the regions that have changed, not the whole line.
> 
> That part seems like a worthwhile change, though -- if it can be isolated.

I'd say "maybe". This boils down to whether "add/remove comment" should be perceived as simple text edit ("add/remove comment markers at these positions") or a more general semantic one ("replace these lines with their commented/uncommented version"). Using the second one seems like more useful on practice.

clason:
Oh, sorry, my bad -- I somehow read "buffer" for "line". Then the current behavior is consistent, of course (commenting replaces a code line with a comment line, and vice versa).

dundargoc:
@echasnovski mind adding a comment and/or link to your reasoning? Just as a precaution so people know why the code is written like this.

PR #29252, opened 2024-06-09
fix(runtime): add commentstring for glsl ftplugin

samuelborn:
# Problem 

The builtin commenting was not able to comment in `.glsl` shader files

# Solution

Add `ftplugin/glsl.lua` with glsl commentstring

# Reference Pull Request

https://github.com/neovim/neovim/pull/23039


clason:
Thanks!

PR #29255, opened 2024-06-09
fix(api): allow `scope = 'local'` with `buf` when using `nvim_get_option_value`

dundargoc:
`nvim_get_option_value` throws a warning if both `scope` and `buf`
options are used at the same time. This is because using `buf` always
implies `scope` is local, and is therefore not needed. There's however
no need to error if `scope` is already set "local" as it's the correct
value.

### DIFF
--- a/src/nvim/api/options.c
+++ b/src/nvim/api/options.c
@@ -54,6 +54,10 @@ static int validate_option_value_args(Dict(option) *opts, char *name, OptIndex *
   }
 
   if (HAS_KEY_X(opts, buf)) {
+    VALIDATE(!(HAS_KEY_X(opts, scope) && *scope == OPT_GLOBAL), "%s",
+             "cannot use both global 'scope' and 'buf'", {
+      return FAIL;
+    });
     *scope = OPT_LOCAL;
     *req_scope = kOptReqBuf;
     *from = find_buffer_by_handle(opts->buf, err);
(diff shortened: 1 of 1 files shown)


src/nvim/api/options.c
```diff
@@ -68,8 +68,8 @@ static int validate_option_value_args(Dict(option) *opts, char *name, OptIndex *
     return FAIL;
   });
 
-  VALIDATE((!HAS_KEY_X(opts, scope) || !HAS_KEY_X(opts, buf)), "%s",
-           "cannot use both 'scope' and 'buf'", {
+  VALIDATE(!(HAS_KEY_X(opts, scope) && *scope == OPT_GLOBAL && HAS_KEY_X(opts, buf)), "%s",
+           "cannot use both 'scope=global' and 'buf'", {
```
zeertzjq, line 72: Other error messages don't use this kind of quoting for `key=val`

dundargoc: I did a quick sweep of other `VALIDATE` usages but couldn't find one that specifically complained about wrong value. I changed it, if it's wrong please lemme know how it should look like.

src/nvim/api/options.c
```diff
@@ -54,6 +54,10 @@ static int validate_option_value_args(Dict(option) *opts, char *name, OptIndex *
   }
 
   if (HAS_KEY_X(opts, buf)) {
+    VALIDATE(!(HAS_KEY_X(opts, scope) && *scope == OPT_GLOBAL), "%s",
+             "cannot use both 'scope' set to 'global' and 'buf'", {
```
zeertzjq, line 58:
```suggestion
             "cannot use both global 'scope' and 'buf'", {
```

PR #29259, opened 2024-06-09
fix(runtime): add commentstring for swift

bruinscolin:
# Problem
The builtin commenting was not able to comment in swift files

# Solution
Add ftplugin/swift.lua with swift commentstring

# Reference 
https://github.com/neovim/neovim/pull/23039#issuecomment-1505645253


clason:
Thanks, but could you please add the remaining missing ones from that comment as well?

PR #29268, opened 2024-06-10
fix(vim._with): don't modify global-local options

### LINKED ISSUE #29253: `vim._with()` does not properly restore global-local options
### Problem

### Steps to reproduce

    ```lua

    _G.log = {}
    end)
    ```
3. `:source`
4. `:=log`.

```
:=log
{ { "Before",
    bo = "-- %s",
    go = "- %s"
  }, { "Inside",
    bo = "## %s",
    go = "## %s"
  }, { "After",
    bo = "-- %s",
    go = "-- %s"
  } }
```

### Expected behavior

### Neovim version (nvim -v)

### Operating system/version

### Terminal name/version

st-0.9

st-256color

### Installation

appimage

dundargoc:
The global options are not properly restored after using `vim._with`.
Circumvent this problem by always using "local" scope. A solution to
allow modifying global options might be added later on, but this is a
good enough fix for the time being.

Closes https://github.com/neovim/neovim/issues/29253.

### DIFF
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -1200,8 +1200,8 @@ function vim._with(context, f)
     previous_options = {}
     for k, v in pairs(context.options) do
       previous_options[k] =
-        vim.api.nvim_get_option_value(k, { win = context.win, buf = context.buf })
-      vim.api.nvim_set_option_value(k, v, { win = context.win, buf = context.buf })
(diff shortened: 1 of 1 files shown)


justinmk (approved):

dundargoc:
Superseded by https://github.com/neovim/neovim/pull/29280

PR #29269, opened 2024-06-10
feat(vim._with): allow specifying scope-specific options

### LINKED ISSUE #29253: `vim._with()` does not properly restore global-local options

dundargoc:
Replace `options` context with `o`, `bo`, `wo` and `go` contexts. This
allows users to specify which scope the option should work on.

Also add `env` context for changing `vim.env`.

Closes https://github.com/neovim/neovim/issues/29253.

---

Alternative to https://github.com/neovim/neovim/pull/29268

Will add tests once we're fairly sure about the design/function signature.

### DIFF
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -1141,8 +1141,11 @@ end
 
 --- @nodoc
 --- @class vim.context.mods
+--- @field bo? table<string, any>
 --- @field buf? integer
 --- @field emsg_silent? boolean
+--- @field env? table<string, any>
+--- @field go? table<string, any>
 --- @field hide? boolean
 --- @field horizontal? boolean
 --- @field keepalt? boolean
@@ -1151,11 +1154,12 @@ end
 --- @field keeppatterns? boolean
 --- @field lockmarks? boolean
 --- @field noautocmd? boolean
---- @field options? table<string, any>
+--- @field o? table<string, any>
 --- @field sandbox? boolean
 --- @field silent? boolean
 --- @field unsilent? boolean
 --- @field win? integer
+--- @field wo? table<string, any>
 
 --- Executes function `f` with the given context specification.
 ---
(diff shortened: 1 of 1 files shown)


dundargoc:
> I like #29268 because it is DWIM. Could we use the DWIM approach for `o` (analogous to `vim.o`)?

Done.

echasnovski:
> I like #29268 because it is DWIM. Could we use the DWIM approach for `o` (analogous to `vim.o`)?

It is not quite flexible enough for a general "context". Using `buf` and/or `win` in context does not necessarily imply that caller wants/needs temporary options be for that buffer / window.

For example there might be the following cases:
- Explicitly set in context global option, even though there is `buf` / `win`.
- Set one context option locally while the other - globally.

Using `bo`, `wo`, `go` contexts allows caller to be explicit and independent of whether `buf` / `win` is set in context. Plus opens a door for a more aligned `env` context (for environment variables).

echasnovski:
I think the upside of this design is also an ability for quite concise and expressive basic implementation (takes care only for options):

```lua
_G.with = function(context, f)
  -- Save and set
  local cache = { bo = {}, env = {}, go = {}, o = {}, wo = {} }
  for scope, scope_cache in pairs(cache) do
    for name, context_value in pairs(context[scope] or {}) do
      scope_cache[name] = vim[scope][name]
      vim[scope][name] = context_value
    end
  end

  -- Execute
  local ok, res = pcall(f)

  -- Restore
  for scope, scope_cache in pairs(cache) do
    for name, cached_value in pairs(scope_cache) do
      vim[scope][name] = cached_value
    end
  end

  -- Return
  if not ok then error(res, 0) end
  return res
end
```

justinmk:
> Using `buf` and/or `win` in context does not necessarily imply that caller wants/needs temporary options be for that buffer / window.

using `o` does imply that?

dundargoc:
Superseded by https://github.com/neovim/neovim/pull/29280

Issue #27046, opened 2024-01-16
Assertion `buf->b_signcols.count[width - 1] >= 0` failed


przepompownia:
### Problem

From few days I get crashes when updating the buffer e.g. undo. 

```cpp
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007fea9eef315f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007fea9eea5472 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fea9ee8f4b2 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007fea9ee8f3d5 in __assert_fail_base (fmt=0x7fea9f003dc8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x5565767fe960 "buf->b_signcols.count[width - 1] >= 0",
    file=file@entry=0x5565767fe748 ".../src/nvim/decoration.c", line=line@entry=848, function=function@entry=0x55657679f310 <__PRETTY_FUNCTION__.1.lto_priv.8> "buf_signcols_count_range") at ./assert/assert.c:92
#5  0x00007fea9ee9e3a2 in __assert_fail (assertion=assertion@entry=0x5565767fe960 "buf->b_signcols.count[width - 1] >= 0", file=file@entry=0x5565767fe748 ".../src/nvim/decoration.c", line=line@entry=848,
    function=function@entry=0x55657679f310 <__PRETTY_FUNCTION__.1.lto_priv.8> "buf_signcols_count_range") at ./assert/assert.c:101
#6  0x00005565764923be in buf_signcols_count_range (buf=0x55657a5cb7f0, row1=20, row2=20, add=-1, clear=kFalse) at .../src/nvim/decoration.c:848
#7  0x000055657648f9c0 in buf_remove_decor_sh (buf=buf@entry=0x55657a5cb7f0, row1=row1@entry=20, row2=row2@entry=20, sh=sh@entry=0x556578824ac0) at .../src/nvim/decoration.c:258
#8  0x000055657648fa46 in buf_decor_remove (buf=0x55657a5cb7f0, row1=20, row2=20, decor=..., free=true) at .../src/nvim/decoration.c:227
[...]

PR #29272, opened 2024-06-10
fix(column): clamp line number for legacy signs

luukvbaal:
Problem:  Legacy :sign API still allows placing signs beyond the end of
          the buffer. This is unaccounted for by the signcolumn tracking
          logic and is disallowed in general for the extmark API which
          implements it now.
Solution: Clamp legacy sign line number to the length of the buffer.

### DIFF
--- a/src/nvim/sign.c
+++ b/src/nvim/sign.c
@@ -126,8 +126,8 @@ static void buf_set_sign(buf_T *buf, uint32_t *id, char *group, int prio, linenr
                          | (has_hl ? MT_FLAG_DECOR_SIGNHL : 0);
 
   DecorInline decor = { .ext = true, .data.ext = { .vt = NULL, .sh_idx = decor_put_sh(sign) } };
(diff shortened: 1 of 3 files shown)


luukvbaal:
Ref https://github.com/neovim/neovim/issues/27046#issuecomment-2158531077, unclear if this is the desired solution.

TamaMcGlinn:
I've just verified on my end that this fixes the assert crash described in #27046.

Unfortunately, I haven't been able to slim down my vimrc config to something easily reproducible, because it is dependent on the interaction of several plugins with a lot of configuration; LSP needs to be working fully for a C++ file, with a linter configured and giving errors, and also the CtrlSF search plugin installed. [My vimrc](https://github.com/TamaMcGlinn/vimrc) is open source, and comes with [an installer](https://github.com/TamaMcGlinn/dotfiles/blob/master/install_vimrc), it just takes a long time. I'm happy to just quickly test whatever you need on my end, of course.

Thank you @luukvbaal for the quick fix!

PR #29280, opened 2024-06-11
Add several updates to `vim._with` (tests, granular option contexts, `env` context)

### LINKED ISSUE #29253: `vim._with()` does not properly restore global-local options

echasnovski:
This PR evolved into three improvements:

- *Massively* cover existing functionality with tests. Some (many) tests are marked as pending because the particular context does not work. Those tests are meant as a reference for future fixing.
- Replace `options` context with `o`, `bo`, `wo`, `go` granular contexts. This aligns with existing Lua options terminology and is more flexible than single `options`.
- Add `env` context for environment variables. As this also aligns with `vim.env`.

More details in commits' messages.

------

Resolves #29253

Alternative to #29269 and #29268

### DIFF
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -1144,7 +1144,6 @@ end
 --- @field buf? integer
 --- @field emsg_silent? boolean
 --- @field hide? boolean
---- @field horizontal? boolean
 --- @field keepalt? boolean
 --- @field keepjumps? boolean
 --- @field keepmarks? boolean
@@ -1159,6 +1158,15 @@ end
 
 --- Executes function `f` with the given context specification.
 ---
+--- Notes:
+--- - Context `{ buf = buf }` has no guarantees about current window when
+---   inside context.
+--- - Context `{ buf = buf, win = win }` is yet not allowed, but this seems
+---   to be an implementation detail.
+--- - There should be no way to revert currently set `context.sandbox = true`
+---   (like with nested `vim._with()` calls). Otherwise it kind of breaks the
+---   whole purpose of sandbox execution.
+---
 --- @param context vim.context.mods
 function vim._with(context, f)
   vim.validate('context', context, 'table')
@@ -1167,7 +1175,6 @@ function vim._with(context, f)
   vim.validate('context.buf', context.buf, 'number', true)
   vim.validate('context.emsg_silent', context.emsg_silent, 'boolean', true)
   vim.validate('context.hide', context.hide, 'boolean', true)
-  vim.validate('context.horizontal', context.horizontal, 'boolean', true)
   vim.validate('context.keepalt', context.keepalt, 'boolean', true)
   vim.validate('context.keepjumps', context.keepjumps, 'boolean', true)
   vim.validate('context.keepmarks', context.keepmarks, 'boolean', true)
@@ -1192,6 +1199,10 @@ function vim._with(context, f)
     if not vim.api.nvim_win_is_valid(context.win) then
       error('Invalid window id: ' .. context.win)
     end
+    -- TODO: Maybe allow it?
+    if context.buf and vim.api.nvim_win_get_buf(context.win) ~= context.buf then
+      error('Can not set both `buf` and `win` context.')
+    end
   end
 
   -- Store original options
@@ -1214,7 +1225,7 @@ function vim._with(context, f)
     end
   end
 
-  return unpack(retval)
+  return unpack(retval, 1, table.maxn(retval))
 end
 
 return vim
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -1141,8 +1141,10 @@ end
 
 --- @nodoc
 --- @class vim.context.mods
+--- @field bo? table<string, any>
 --- @field buf? integer
 --- @field emsg_silent? boolean
+--- @field go? table<string, any>
 --- @field hide? boolean
 --- @field keepalt? boolean
 --- @field keepjumps? boolean
@@ -1150,11 +1152,47 @@ end
 --- @field keeppatterns? boolean
 --- @field lockmarks? boolean
 --- @field noautocmd? boolean
---- @field options? table<string, any>
+--- @field o? table<string, any>
 --- @field sandbox? boolean
 --- @field silent? boolean
 --- @field unsilent? boolean
 --- @field win? integer
+--- @field wo? table<string, any>
+
+--- @nodoc
+--- @class vim.context.state
+--- @field bo? table<string, any>
+--- @field go? table<string, any>
+--- @field wo? table<string, any>
+
+local scope_map = { buf = 'bo', global = 'go', win = 'wo' }
+local scope_order = { 'o', 'wo', 'bo', 'go' }
+local state_restore_order = { 'bo', 'wo', 'go' }
+
+--- Gets data about current state, enough to properly restore specified options/env/etc.
+--- @param context vim.context.mods
+--- @return vim.context.state
+local get_context_state = function(context)
+  local res = { bo = {}, go = {}, wo = {} }
+
+  -- Use specific order from possibly most to least intrusive
+  for _, scope in ipairs(scope_order) do
+    for name, _ in pairs(context[scope] or {}) do
+      local sc = scope == 'o' and scope_map[vim.api.nvim_get_option_info2(name, {}).scope] or scope
+
+      -- Do not override already set state and fall back to `vim.NIL` for
+      -- state `nil` values (which still needs restoring later)
+      res[sc][name] = res[sc][name] or vim[sc][name] or vim.NIL
+
+      -- Always track global option value to properly restore later.
+      -- This matters for at least `o` and `wo` (which might set either/both
+      -- local and global option values).
+      res.go[name] = res.go[name] or vim.go[name]
+    end
+  end
+
+  return res
+end
 
 --- Executes function `f` with the given context specification.
 ---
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/shared.lua
```diff
@@ -1151,34 +1154,70 @@ end
...
 --- @field noautocmd? boolean
---- @field options? table<string, any>
+--- @field o? table<string, any>
 --- @field sandbox? boolean
 --- @field silent? boolean
 --- @field unsilent? boolean
 --- @field win? integer
+--- @field wo? table<string, any>
+
+local save_set_context = function(context)
```
justinmk, line 1164: The name might need a bit of explanation, is this correct:
```suggestion
--- Gets option :set context.
local save_set_context = function(context)
```

Nit: "save" implies side-effects, so "get" would normally be preferrable, but `get_set_context` would be strange.

justinmk: > How about `cache_and_set_context()`?

doesn't seem better than "save"

justinmk: > How about `cache_and_set_context()`?

doesn't seem better than "save".  I see now that this is setting options so it does in fact have side effects, so ignore my previous comment. Just maybe a brief docstring would help.

runtime/lua/vim/shared.lua
```diff
@@ -1151,34 +1154,70 @@ end
...
+    ['context.keeppatterns'] = { context.keeppatterns, 'boolean', true },
+    ['context.lockmarks'] = { context.lockmarks, 'boolean', true },
+    ['context.noautocmd'] = { context.noautocmd, 'boolean', true },
+    ['context.o'] = { context.o, 'table', true },
+    ['context.sandbox'] = { context.sandbox, 'boolean', true },
+    ['context.silent'] = { context.silent, 'boolean', true },
+    ['context.unsilent'] = { context.unsilent, 'boolean', true },
+    ['context.win'] = { context.win, 'number', true },
+    ['context.wo'] = { context.wo, 'table', true },
+  })
```
gpanders, line 1220: Out of curiosity, is there a reason to change this?

The single-line `vim.validate` calls are basically free. I don't think changing this is a readability win either way (there's a small argument to be made for eliminating the repeated `vim.validate`, but that's it). My main concern is about establishing patterns. Future contributors will copy-paste what they find elsewhere, and I think we should try to use the fast version of `vim.validate` wherever possible for that reason.

echasnovski: My bad, it was more of an assumption that single call to `vim.validate()` is more performant than multiple. However, I did not account that those are the newly added style of vararg calls instead of creating a table every single time.

I've profiled two versions (single `vim.validate()` with table argument vs multiple with vararg) and indeed vararg version seems to be faster: around 5e-07 seconds for single `vim.validate()` and around 8e-08 seconds for vararg (time for all current 19 calls).

I'll change back to multiple vararg calls.

runtime/lua/vim/shared.lua
echasnovski: I resolved this by compactifying implementation so that it can be reasonably fit inside `vim._with()` body. No helper function - no problem.

runtime/lua/vim/shared.lua
```diff
@@ -1192,29 +1221,60 @@ function vim._with(context, f)
...
+
+    -- Use specific order from possibly most to least intrusive
+    for _, scope in ipairs(scope_order) do
+      for name, context_value in pairs(context[scope] or {}) do
+        local sc = scope == 'o' and scope_map[vim.api.nvim_get_option_info2(name, {}).scope]
+          or scope
+
+        -- Do not override already set cache and fall back to `vim.NIL` to
+        -- cache `nil` values (which needs restoring later)
+        cache[sc][name] = cache[sc][name] or vim[sc][name] or vim.NIL
```
dundargoc, line 1243: Do I understand it correctly that e.g. if both `o` and `bo` scopes are used with different values then `o` will be used since it has higher priority?

dundargoc (approved):
I haven't checked the test code but I assume they hold the standard echa quality. The option code is somewhat hard to follow, but I'm unsure if there's a good way to simplify it tbh. Maybe we can look into that later. All in all looks great. All the test cases are super valuable.

runtime/lua/vim/shared.lua
```diff
@@ -1151,21 +1154,46 @@ end
...
   vim.validate('context', context, 'table')
   vim.validate('f', f, 'function')
 
+  vim.validate('context.bo', context.bo, 'table', true)
   vim.validate('context.buf', context.buf, 'number', true)
   vim.validate('context.emsg_silent', context.emsg_silent, 'boolean', true)
+  vim.validate('context.env', context.env, 'table', true)
+  vim.validate('context.go', context.go, 'table', true)
   vim.validate('context.hide', context.hide, 'boolean', true)
   vim.validate('context.horizontal', context.horizontal, 'boolean', true)
```
dundargoc, line 1198: This is an error from my part. This context does not exists in `vim._with_c` and should probably be removed for the time being.

runtime/lua/vim/shared.lua
```diff
@@ -1192,29 +1221,60 @@ function vim._with(context, f)
...
-      previous_options[k] =
-        vim.api.nvim_get_option_value(k, { win = context.win, buf = context.buf })
-      vim.api.nvim_set_option_value(k, v, { win = context.win, buf = context.buf })
+  -- Decorate so that save-set-restore options is done in correct window-buffer
+  local callback = function()
+    -- Save and set context (NOTE: triggers `OptionSet` event)
+    local cache = { bo = {}, env = {}, go = {}, wo = {} } ---@type table<string, any>
+
+    -- Use specific order from possibly most to least intrusive
+    for _, scope in ipairs(scope_order) do
```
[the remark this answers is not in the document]
echasnovski: I am not sure how to make it less confusing, as it is caching and setting inside for loops.

Here is an example of what it does if it would make it easier to follow. For `{ o = { winblend = 25 }, bo = { undolevels = 0 } }`, it does the following sequence:
- Process `o` context:
    - Get its "effective"/"cachable" context via `nvim_get_option_info2()`. The result is one of `bo`, `go`, or `wo`. It is `wo` for `winblend`.
    - Cache it via `cache.wo.winblend = cache.wo.winblend or vim.wo.winblend or vim.NIL`. The first `cache.wo.winblend` is there in order to write to cache only once (as the same option name can be supplied in three different `context` fields). The fallback to `vim.NIL` is there mostly for `env` context (as `vim.env` can return `nil`).
    - Cache its global value via `cache.go.winblend = cache.go.winblend or vim.go.winblend`. This is needed because later setting option might affect global values (like for `o` or `wo` contexts).
    - Set it with `vim.o.winblend = 25`.
- Process `bo` context:
    - `cache.bo.undolevels = cache.bo.undolevels or vim.bo.undolevels or vim.NIL`.
    - `cache.go.undolevels = cache.go.undolevels or vim.go.undolevels`.
    - `vim.bo.undolevels = 0`.
- Execute callback `f`.
- Restore cache:
    - `vim.bo.undolevels = cache.bo.undolevels`
    - `vim.wo.winblend = cache.wo.winblend`
    - `vim.go.undolevels = cache.go.undolevels` and `vim.go.winblend = cache.go.winblend`.

runtime/lua/vim/shared.lua
echasnovski: On the contrary, `bo` will be used as it will be set *after* setting the `o` context.

Using `cache[sc][name] = cache[sc][name] or vim[sc][name]` instead of direct `cache[sc][name] = vim[sc][name]` guards exactly from this kind of case. Without it, cache would have been set to user's value from the `o` context and later (erroneously) restored to it.

runtime/lua/vim/shared.lua
echasnovski: I removed it from code, but left pending tests just in case.

runtime/lua/vim/shared.lua
echasnovski: But good question, as I realized that there was no test case specifically if same option is supplied in several contexts.

runtime/lua/vim/shared.lua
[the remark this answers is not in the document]
dundargoc: Ah ok, that clears things up a bit. Hmm, yeah I can't come up with a simpler approach for the time being. Let's go with this 👍

runtime/lua/vim/shared.lua
echasnovski: I've recently had a realization that if there is a misplaced option scope (like `{ bo = { winblend = 10 } }`), then the approach of "cache and set" inside same double loop would not properly restore already set context values (due to error during `vim.bo.winblend = 10`). So I indeed had to separate `get_context_state()` into a separate function (with brief description of what it does).

runtime/lua/vim/shared.lua
```diff
@@ -1141,44 +1141,105 @@ end
...
+--- @field bo? table<string, any>
+--- @field env? table<string, any>
+--- @field go? table<string, any>
+--- @field wo? table<string, any>
+
+local scope_map = { buf = 'bo', global = 'go', win = 'wo' }
+local scope_order = { 'o', 'wo', 'bo', 'go', 'env' }
+local state_restore_order = { 'bo', 'wo', 'go', 'env' }
+
+--- Get data about current state, enough to properly undo later context setting
```
justinmk, line 1174:
```suggestion
--- Gets data about current state, enough to properly restore specified options/env/etc.
```

test/functional/lua/with_spec.lua
```diff
@@ -7,286 +7,1620 @@ local api = n.api
...
     ]]
-    t.eq(mark, { 1, 0 })
+    local get_error = function(string_args)
+      return exec_lua('return get_error(' .. string_args .. ')')
+    end
+
+    matches('context.*table', get_error("'a', function() end"))
+    matches('f.*function', get_error('{}, 1'))
+
+    local validate_context = function(bad_context, expected_type)
```
justinmk, line 1596: we are fairly consistent about using `assert_` prefix for tests
```suggestion
    local assert_context = function(bad_context, expected_type)
```

justinmk (approved):

dundargoc:
This is super valuable. Thanks for championing this.

Follow-up work (not necessarily by you, I'm talking in general): ensure all the tests marked as `pending` passes. It looks like it involved primarily ensuring that `vim._with` can be nested which will need adjustments in the C code.

echasnovski:
> Follow-up work (not necessarily by you, I'm talking in general): ensure all the tests marked as `pending` passes. It looks like it involved primarily ensuring that `vim._with` can be nested which will need adjustments in the C code.

Not only those, I am afraid. Quite a few context fields coming from command modifiers don't seem to work:
- [`emsg_silent`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L306)
- [`hide`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L512)
- [`horizontal`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L565) (although it is not even implemented in C part)
- [`keepalt`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L591)
- [`keepjumps`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L625)
- [`keepmarks`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L654)
- [`keeppatterns`](https://github.com/neovim/neovim/blob/76dd07e572935742f6dfa0d3364e262ebc047dcc/test/functional/lua/with_spec.lua#L687)

Soo... basically most part of the flags :)

PR #29290, opened 2024-06-11
Add StatusLineTerm highlight group and move terminal buffer defaults

gpanders:
This PR contains two distinct but related changes.

The first is to add the `StatusLineTerm` and `StatusLineTermNC` highlight groups to highlight the status line in terminal buffers. This allows users to easily define separate statusline highlights in terminal buffers.

These highlight groups are defined in Vim and Vim by default uses separate colors for the statusline in terminal buffers. For now we just link `StatusLineTerm` and `StatusLineTermNC` to `StatusLine` and `StatusLineNC` (respectively), but we could later on define independent default highlights for these as part of the default colorscheme (I'll leave that to somebody more qualified tho, cc @echasnovski).

This essentially adds the following (from `vim_diff`) as a default:

```
Highlight groups:
- *hl-StatusLineTerm* *hl-StatusLineTermNC* are unnecessary because Nvim
  supports 'winhighlight' window-local highlights. For example, to mimic Vim's
  StatusLineTerm:  >vim
      hi StatusLineTerm ctermfg=black ctermbg=green
      hi StatusLineTermNC ctermfg=green
      autocmd TermOpen,WinEnter * if &buftype=='terminal'
        \|setlocal winhighlight=StatusLine:StatusLineTerm,StatusLineNC:StatusLineTermNC
        \|else|setlocal winhighlight=|endif
```

Note that some of the bundled colorschemes already have `StatusLineTerm` highlight definitions (e.g. `industry`), so this change allows those colorschemes to more closely match their Vim counterparts.

The other change moves the default options for `:terminal` buffers out of `terminal_open` (in C) into a `TermOpen` autocommand in `_defaults.lua`. There is no functional benefit or difference of doing it this way, the motivation is mostly to consolidate default settings and move code out of C and into Lua (where it makes sense to do so, which I think it does in this case). We also add a new default `'winhighlight'` option to use the new `StatusLineterm` and `StatusLineTermNC` highlights in the terminal window.

One potential risk/downside of this is that a user could delete the default `TermOpen` autocommand and the `:terminal` buffer would no longer have those default options. But this could also be seen as a plus by some users, I suppose. Presumably, if someone deletes the autocommand they know what they are doing and are willing to deal with the consequences.

---

**feat(highlight): add StatusLineTerm and StatusLineTermNC groups**

These highlight groups are used for the statusline in :terminal windows. By default they link to StatusLine and StatusLineNC (respectively), so there is no visual difference unless a colorscheme defines these groups separately.

**refactor(terminal): move :terminal defaults to _defaults.lua**


runtime/doc/syntax.txt
```diff
@@ -5104,6 +5104,11 @@ SpellRare	Word that is recognized by the spellchecker as one that is
 StatusLine	Status line of current window.
 							*hl-StatusLineNC*
 StatusLineNC	Status lines of not-current windows.
+							*hl-StatusLineTerm*
+StatusLineTerm	Status line of |terminal| window.
+							*hl-StatusLineTermNC*
+StatusLineTermNC
+		Status line of non-current |terminal| window.
```
echasnovski, line 5111:
```suggestion
		Status line of non-current |terminal| windows.
```

echasnovski (approved):
Linking to `StatusLine` and `StatusLineNC` look like the most reasonable approach.

runtime/doc/vim_diff.txt
```diff
@@ -665,17 +665,6 @@ Events:
 - *SafeStateAgain*
 - *SigUSR1* Use |Signal| to detect `SIGUSR1` signal instead.
 
-Highlight groups:
-- *hl-StatusLineTerm* *hl-StatusLineTermNC* are unnecessary because Nvim
-  supports 'winhighlight' window-local highlights. For example, to mimic Vim's
```
justinmk, line 670: should we keep a smaller note like (in the `:help nvim-changed` section):

    |hl-StatusLineTerm| |hl-StatusLineTermNC| are implemented as 'winhighlight' window-local highlights which are set by the default TermOpen handler.

I assume that is different than the Vim implementation.

justinmk (approved):

runtime/lua/vim/_defaults.lua
```diff
@@ -282,6 +282,21 @@ do
...
+    desc = 'Default settings for :terminal buffers',
+    callback = function(args)
+      local buf = args.buf ---@type integer
+      vim.bo[buf].modifiable = false
+      vim.bo[buf].undolevels = -1
+      vim.bo[buf].scrollback = vim.o.scrollback < 0 and 10000 or math.max(1, vim.o.scrollback)
+      vim.bo[buf].textwidth = 0
+      vim.wo[0][buf].wrap = false
+      vim.wo[0][buf].list = false
+      vim.wo[0][buf].winhighlight = 'StatusLine:StatusLineTerm,StatusLineNC:StatusLineTermNC'
```
echasnovski, line 296: I've only just realized that this will completely override possible user's 'winhighlight'.

It'll probably be better to use something like `vim.opt_local.winhighlight:append()`, but that doesn't support non-current window-buffers. So either using the new `vim._with({ buf = buf }, function() ... end)` or string manipulation like `vim.wo[0][buf].winhighlight = vim.wo[0][buf].winhighlight .. ',...'` should do the trick.

justinmk (approved):

zeertzjq:
Should these two be all to runtime/colors/vim.lua?

echasnovski:
> Should these two be added to runtime/colors/vim.lua?

Good question. I am not sure what is the best thing to do here (and when adding new highlight group in general).

The `:color vim` was initially intended to be a way to ["restore previous definitions'](https://github.com/neovim/neovim/blob/0231265c8cf8c13f317bbddcfcbac2639e7022cd/runtime/doc/news-0.10.txt#L36). Both `StatusLineTerm` / `StatusLineTermNC` were not part of previous definitions.

On the other hand, `:color vim` can be treated (and is documented as such) as "Vim's default color scheme". In which case it should have [green background](https://github.com/vim/vim/blob/02f3ebacfbfa1f892347d7532278f24620e68300/src/highlight.c#L346-L347).

I am leaning towards the second option, but don't have clear preference.

PR #29295, opened 2024-06-12
docs(lsp): format the handwritten part

ofseed:
This part of the documentation seems to be handwritten, while other parts are generated by luacats comments. Unify the style. I suggest backporting this patch.

* Before:
image

* After:
image

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -220,26 +220,26 @@ Each response handler has this signature: >
     function(err, result, ctx, config)
 <
     Parameters: ~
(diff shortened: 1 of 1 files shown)


justinmk (approved):

wookayin:
Related work: #27101. I don't mind if this gets merged first.

ofseed:
@wookayin Thank you for your hard work! At first, I tried to find out if I could do it like you did, I didn't know you had such a PR. That needs a lot of work, and the documentation for `lsp-handler` is written pretty well in this way, so I still recommend merging this as a tiny fix.

PR #29297, opened 2024-06-12
fix(meta): mod operator lpeg meta

TheLeoP:
The mod operator does not produce any captures, but within the lua-ls type system it makes sense that it returns a Capture.

Upstream PR https://github.com/LuaCATS/lpeg/pull/10


TheLeoP:
Already merged #29328

PR #29304, opened 2024-06-12
refactor(lsp): use metatable for buf_versions

icholy:
This reduces the number of nil checks around buf_versions usage

### DIFF
--- a/runtime/lua/vim/lsp/client.lua
+++ b/runtime/lua/vim/lsp/client.lua
@@ -916,18 +916,16 @@ function Client:_text_document_did_open_handler(bufnr)
   if not api.nvim_buf_is_loaded(bufnr) then
     return
   end
-  local filetype = vim.bo[bufnr].filetype
 
-  local params = {
+  local filetype = vim.bo[bufnr].filetype
+  self.notify(ms.textDocument_didOpen, {
     textDocument = {
-      version = 0,
+      version = lsp.util.buf_versions[bufnr],
       uri = vim.uri_from_bufnr(bufnr),
       languageId = self.get_language_id(bufnr, filetype),
       text = lsp._buf_get_full_text(bufnr),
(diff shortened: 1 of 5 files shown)


runtime/lua/vim/lsp/client.lua
```diff
@@ -916,18 +916,16 @@ function Client:_text_document_did_open_handler(bufnr)
...
     return
   end
-  local filetype = vim.bo[bufnr].filetype
 
-  local params = {
+  local filetype = vim.bo[bufnr].filetype
+  self.notify(ms.textDocument_didOpen, {
     textDocument = {
-      version = 0,
+      version = lsp.util.buf_versions[bufnr],
```
justinmk, line 923: this particular case seems like it would be clearer to keep the explicit 0 assignment

icholy: I'm not opposed, but the spec doesn't say that `textDocument/didOpen` version needs to be 0. I plan on opening a follow-up PR to start the version at 1 to avoid possible issues with the way version=0 is handled https://github.com/neovim/neovim/commit/b67f689e4cb08ac3f33432038ae9aec81383a9f3

icholy: Actually, I think that sending an explicit 0 here could result in a bug when there are 2 LSP servers. Example:

``` mermaid
sequenceDiagram
    nvim ->>+ LSP1: initialize
    nvim ->>+ LSP2: initialize
    LSP1 ->>- nvim: initialized
    nvim ->> LSP1: textDocument/didOpen version=0
    note over nvim: user modifies content <br/> version=1
    LSP2 ->>- nvim: initialized
    nvim ->> LSP1: textDocument/didOpen version=0
    note over LSP1: should have recieved version=1
```

PR #29310, opened 2024-06-13
fix: Use `vim_dialog_yesno` for the "file changed since reading" message

murtaza64:
In order to allow more flexibility in how UI plugins like [Noice handle this specific prompt](https://github.com/folke/noice.nvim/issues/720), switch from using `ask_yesno` to `vim_dialog_yesno`, which sends a `msg_send.confirm` event which can be handled by `vim.ui_attach`. This also brings it in line with other user prompts such as the swapfile dialog.

Before:
image

After:
image

With Noice:
image

Questions/feedback requested:
- I felt like combining the existing two messages together creates a better UX for this flow, but happy to leave it as is if that's preferred
- Even without combining the two strings onto one line, I probably would still need to update localization (at least to add the question mark that `ask_yesno` used to add). I'm happy to take a crack at that, but I'm having trouble testing it. Doing `VIMRUNTIME=runtime ./build/bin/nvim --clean` and then setting `:lang es_ES` doesn't seem to use Spanish strings :/

### DIFF
--- a/src/nvim/bufwrite.c
+++ b/src/nvim/bufwrite.c
@@ -355,9 +355,11 @@ static int check_mtime(buf_T *buf, FileInfo *file_info)
       && time_differs(file_info, buf->b_mtime_read, buf->b_mtime_read_ns)) {
     msg_scroll = true;  // Don't overwrite messages here.
     msg_silent = 0;     // Must give this prompt.
-    // Don't use emsg() here, don't want to flush the buffers.
-    msg(_("WARNING: The file has been changed since reading it!!!"), HL_ATTR(HLF_E));
-    if (ask_yesno(_("Do you really want to write to it"), true) == 'n') {
+    if (vim_dialog_yesno(VIM_WARNING,
+                         NULL,
+                         _(
+                          "WARNING: The file has been changed since reading it. Do you really want to write to it?"),
+                         2) == VIM_NO) {
       return FAIL;
     }
     msg_scroll = false;  // Always overwrite the file message now.
--- a/src/nvim/po/es.po
+++ b/src/nvim/po/es.po
@@ -2387,19 +2387,9 @@ msgstr "[no hay fin de línea]"
 msgid "[Incomplete last line]"
 msgstr "[Última línea incompleta]"
 
-# don't overwrite messages here
-# must give this prompt
-# don't use emsg() here, don't want to flush the buffers
-#. don't overwrite messages here
-#. must give this prompt
-#. don't use emsg() here, don't want to flush the buffers
-#: ../fileio.c:3865
-msgid "WARNING: The file has been changed since reading it!!!"
-msgstr "ADVERTENCIA: ¡¡¡El archivo ha cambiado desde que se leyó!!!"
-
-#: ../fileio.c:3867
-msgid "Do you really want to write to it"
-msgstr "¿Desea realmente escribir al archivo?"
+#: ../bufwrite.c:362
+msgid "WARNING: The file has been changed since reading it. Do you really want to write to it?"
+msgstr "ADVERTENCIA: El archivo ha cambiado desde que se leyó. ¿Desea realmente escribir en él?"
 
 #: ../fileio.c:4648
 #, c-format
--- a/test/functional/core/fileio_spec.lua
+++ b/test/functional/core/fileio_spec.lua
@@ -295,10 +295,10 @@ describe('fileio', function()
     -- use async feed_command because nvim basically hangs on the prompt
     feed_command('w')
     screen:expect([[
-      {9:WARNING: The file has been changed since}|
-      {9: reading it!!!}                          |
-      {6:Do you really want to write to it (y/n)?}|
-      ^                                        |
+      {6:WARNING: The file has been changed since}|
+      {6: reading it. Do you really want to write}|
+      {6: to it?}                                 |
+      {6:(Y)es, [N]o: }^                           |
     ]])
 
     feed('n')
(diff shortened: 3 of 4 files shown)


murtaza64:
This is my first PR in neovim core so I'm not sure if I should assign anyone or who I could reach out to for feedback, but feedback is very much appreciated!

murtaza64:
@tjdevries you're the only core maintainer I know the name of so tagging you for feedback and/or to point me in the right direction. This is my first PR in neovim :)

src/nvim/bufwrite.c
```diff
@@ -355,9 +355,11 @@ static int check_mtime(buf_T *buf, FileInfo *file_info)
       && time_differs(file_info, buf->b_mtime_read, buf->b_mtime_read_ns)) {
     msg_scroll = true;  // Don't overwrite messages here.
     msg_silent = 0;     // Must give this prompt.
-    // Don't use emsg() here, don't want to flush the buffers.
-    msg(_("WARNING: The file has been changed since reading it!!!"), HL_ATTR(HLF_E));
-    if (ask_yesno(_("Do you really want to write to it"), true) == 'n') {
+    if (vim_dialog_yesno(VIM_WARNING,
+                         NULL,
+                         _(
+                          "WARNING: The file has been changed since reading it. Do you really want to write to it?"),
```
wookayin, line 361: Just out of curiosity, any special reason to change the message slightly (e.g. dropping the exclamation marks)?

src/nvim/po/es.po
```diff
@@ -2387,19 +2387,9 @@ msgstr "[no hay fin de línea]"
...
-#: ../fileio.c:3865
-msgid "WARNING: The file has been changed since reading it!!!"
-msgstr "ADVERTENCIA: ¡¡¡El archivo ha cambiado desde que se leyó!!!"
-
-#: ../fileio.c:3867
-msgid "Do you really want to write to it"
-msgstr "¿Desea realmente escribir al archivo?"
+#: ../bufwrite.c:362
+msgid "WARNING: The file has been changed since reading it. Do you really want to write to it?"
+msgstr "ADVERTENCIA: El archivo ha cambiado desde que se leyó. ¿Desea realmente escribir en él?"
```
wookayin, line 2392: What about other locale files other than Spanish?

test/functional/core/fileio_spec.lua
```diff
@@ -295,10 +295,10 @@ describe('fileio', function()
     -- use async feed_command because nvim basically hangs on the prompt
     feed_command('w')
     screen:expect([[
-      {9:WARNING: The file has been changed since}|
-      {9: reading it!!!}                          |
-      {6:Do you really want to write to it (y/n)?}|
-      ^                                        |
+      {6:WARNING: The file has been changed since}|
```
wookayin, line 298: It'd be just a personal taste, but I would actually prefer shouting, intimidating warning messages that are put in red or more discernible than simple `MoreMsg` highlights.

test/functional/ui/input_spec.lua
```diff
@@ -357,36 +358,23 @@ describe('input non-printable chars', function()
...
-      {3:                                                            }|
-      "Xtest-overwrite"                                           |
-      {9:WARNING: The file has been changed since reading it!!!}      |
-      {6:Do you really want to write to it (y/n)?}u                   |
-      {6:Do you really want to write to it (y/n)?}                    |
-      {6:Do you really want to write to it (y/n)?}n                   |
-      {6:Press ENTER or type command to continue}^                     |
-    ]])
-
-    feed('<cr>')
```
wookayin, line 389: Q: why removing those tests?

murtaza64: With `vim_dialog_yesno`, after pressing 'n', the "Press ENTER" line doesn't appear

src/nvim/po/es.po
murtaza64: I was having trouble testing this one, `lang es_ES` on my dirty build doesn't seem to do anything. Do I have to put the localization files somewhere for it to pick them up?

src/nvim/bufwrite.c
murtaza64: No special reason, but I always found the three exclamation marks a bit over the top in addition to the all caps WARNING. Don't feel too strongly about it though.

test/functional/core/fileio_spec.lua
murtaza64: Hmm, I guess we could still print the first line in WARNING colors like it did before, but the main idea behind this change is to allow `ui_attach` to handle this prompt gracefully, and splitting the message into two makes it harder to do that

murtaza64:
>this part is common to vim. I suggest you submit it to vim first.

I could give it a shot, but I suspect they won't see the value in this change since the main reason is to make it play nicely with `ui_attach` which is neovim only

test/functional/core/fileio_spec.lua
murtaza64: I could also look into the implementation of `do_dialog` to see if we can use different highlights for the message

justinmk (approved):
If `vim_dialog_yesno` helps something, that seems useful. But please drop the formatting/wording changes, that's not worth the trouble (or at least should be a separate PR).

test/functional/ui/input_spec.lua
justinmk: That seems like an improvement.

justinmk:
> switch from using `ask_yesno` to `vim_dialog_yesno`, which sends a `msg_send.confirm` event which can be handled by `vim.ui_attach`

This needs test coverage. Should only require small updates, at least a new it() case here: https://github.com/neovim/neovim/blob/b7e896671500a6e9a0c773bc6bac2d073e588eba/test/functional/ui/messages_spec.lua#L27

PR #29313, opened 2024-06-13
fix(highlight): add `StatusLineTerm`/`StatusLineTermNC` to `:color vim`

echasnovski:
Problem: both `StatusLineTerm`/`StatusLineTermNC` are now explicitly
  used, but `:color vim` does not set them to the values used in Vim.
  This might be fine if `:color vim` is treated as "the state of default
  color scheme prior the big update", but it seems to be better treated
  as "Vim's default color scheme" (how it is documented in its header).

Solution: add `StatusLineTerm`/`StatusLineTermNC` definitions to
  'runtime/colors/vim.lua'.
  Use explicit foreground colors ('Whte'/'Black') instead of `guifg=bg`
  used in source, as the latter caused some problems in the past (if
  `Normal` is not defined, `nvim_set_hl()` can't recognize `'bg'` as the
  foreground value).
  Also realign the rest of the background conditional highlight groups.

------

Also update some functional tests to make them pass.

------

Follow up to #29290.
[Original context](https://github.com/neovim/neovim/pull/29290#issuecomment-2164862061).
Vim's references: [light background](https://github.com/vim/vim/blob/02f3ebacfbfa1f892347d7532278f24620e68300/src/highlight.c#L344-L347), [dark background](https://github.com/vim/vim/blob/02f3ebacfbfa1f892347d7532278f24620e68300/src/highlight.c#L433-L436).


zeertzjq:
Needs rebase on #29326

echasnovski:
> Needs rebase on #29326

Done.

test/functional/terminal/tui_spec.lua
```diff
@@ -1552,6 +1557,29 @@ describe('TUI', function()
       [3] = { { bold = true }, { bold = true } },
       [4] = { { fg_indexed = true, foreground = tonumber('0xe0e000') }, { foreground = 3 } },
       [5] = { { foreground = tonumber('0xff8000') }, {} },
+      [6] = {
+        { bold = true, background = Screen.colors.LightGreen, foreground = Screen.colors.Black },
+        { bold = true },
+      },
```
zeertzjq, line 1563: Attribute ID 2 is now unused, so you can replace attribute 2 with this.

test/functional/terminal/tui_spec.lua
```diff
@@ -1056,14 +1056,19 @@ describe('TUI', function()
     if is_ci('github') then
       pending('tty-test complains about not owning the terminal -- actions/runner#241')
     end
+    screen:set_default_attr_ids({
+      [1] = { reverse = true }, -- focused cursor
+      [3] = { bold = true },
+      [19] = { bold = true, background = 121, foreground = Screen.colors.Black }, -- StatusLineTerm
```
zeertzjq, line 1062:
```suggestion
      [19] = { bold = true, background = 121, 0 }, -- StatusLineTerm
```

test/functional/terminal/tui_spec.lua
```diff
@@ -1552,6 +1557,29 @@ describe('TUI', function()
...
+      },
+      [7] = {
+        {
+          fg_indexed = true,
+          bg_indexed = true,
+          bold = true,
+          background = tonumber('0x66ff99'),
+          foreground = Screen.colors.Black,
+        },
+        { bold = true, background = 121, foreground = Screen.colors.Black },
```
zeertzjq, line 1572:
```suggestion
        { bold = true, background = 121, foreground = 0 },
```

test/functional/terminal/tui_spec.lua
```diff
@@ -1552,6 +1557,29 @@ describe('TUI', function()
...
+      },
+      [8] = {
+        {
+          fg_indexed = true,
+          bg_indexed = true,
+          background = tonumber('0x66ff99'),
+          foreground = Screen.colors.Black,
+        },
+        { background = 121, foreground = Screen.colors.Black },
```
zeertzjq, line 1581:
```suggestion
        { background = 121, foreground = 0 },
```

test/functional/ui/wildmode_spec.lua
```diff
@@ -199,9 +199,13 @@ describe("'wildmenu'", function()
     feed((':terminal "%s" REP 5000 !terminal_output!<cr>'):format(testprg('shell-test')))
     feed('G') -- Follow :terminal output.
     feed([[:sign <Tab>]]) -- Invoke wildmenu.
+    screen:set_default_attr_ids {
+      [31] = { foreground = Screen.colors.Grey0, background = Screen.colors.Yellow },
```
zeertzjq, line 203:
```suggestion
      [31] = { foreground = Screen.colors.Black, background = Screen.colors.Yellow },
```

test/functional/ui/wildmode_spec.lua
```diff
@@ -259,9 +263,13 @@ describe("'wildmenu'", function()
 
     feed([[<C-\><C-N>]])
     feed([[:<Tab>]]) -- Invoke wildmenu.
+    screen:set_default_attr_ids {
+      [31] = { foreground = Screen.colors.Grey0, background = Screen.colors.Yellow },
```
zeertzjq, line 267:
```suggestion
      [31] = { foreground = Screen.colors.Black, background = Screen.colors.Yellow },
```

zeertzjq (approved):

PR #29316, opened 2024-06-13
fix: screenchar works when multigrid enabled

### LINKED ISSUE #29135: screenchar()/screenstring() with multigrid
### Problem

When multigrid is enabled, the screenchar/screenstring functions do not return the correct data. See https://github.com/neovide/neovide/issues/2569 for the original report.

### Steps to reproduce

Modify the `screenchar_spec.lua` test to use multigrid and observe that it's failing.

### Expected behavior

Multigrid should not affect the functionality of `screenchar`.

### Neovim version (nvim -v)

NVIM v0.10.0

N/A

### Operating system/version

Arch Linux

### Terminal name/version

Neovide 0.13.1

xterm-256color

### Installation

pacman

TwIStOy:
fix #29135

### DIFF
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -803,6 +803,7 @@ void ui_ext_win_position(win_T *wp, bool validate)
   if (!wp->w_floating) {
     ui_call_win_pos(wp->w_grid_alloc.handle, wp->handle, wp->w_winrow,
                     wp->w_wincol, wp->w_width, wp->w_height);
+    ui_put_grid(&wp->w_grid_alloc, false);
     return;
   }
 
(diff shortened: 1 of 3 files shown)


src/nvim/window.c
```diff
@@ -803,6 +803,7 @@ void ui_ext_win_position(win_T *wp, bool validate)
   if (!wp->w_floating) {
     ui_call_win_pos(wp->w_grid_alloc.handle, wp->handle, wp->w_winrow,
                     wp->w_wincol, wp->w_width, wp->w_height);
+    ui_put_grid(&wp->w_grid_alloc, false);
```
bfredl, line 806: I don't think this should be done. in multigrid mode, there is no guarantee that the grid size of the specific windows will match the space in default_grid (that's the entire point of ext_multigrid mode), so things will get weird in other ways. Also it breaks an abstraction boundary to ui_compositor.c which is not something I like for a edge case unrelated to ordinary use.

justinmk:
#31788 looks lower-risk.

PR #29321, opened 2024-06-13
fix(diagnostics): don't apply extmarks to invalid lines

ribru17:
Problem: If there are errors in the last line of a buffer, something like `Gdk` or `G2k3J` will produce an error (at least with `lua_ls`):
```
Error executing vim.schedule lua callback: .../neovim/share/nvim/runtime/lua/vim/diagnostic.lua:1446: Invalid 'line': out of range
```
Solution: Only set extmarks if the target buffer line still exists

---

I had difficulty writing a test for this, it kept passing with and without the change...

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -1359,6 +1359,10 @@ M.handlers.signs = {
     bufnr = get_bufnr(bufnr)
     opts = opts or {}
 
+    if not api.nvim_buf_is_loaded(bufnr) then
+      return
+    end
+
     if opts.signs and opts.signs.severity then
       diagnostics = filter_by_severity(opts.signs.severity, diagnostics)
     end
@@ -1441,8 +1445,10 @@ M.handlers.signs = {
     local numhl = opts.signs.numhl or {}
     local linehl = opts.signs.linehl or {}
 
+    local line_count = api.nvim_buf_line_count(bufnr)
+
     for _, diagnostic in ipairs(diagnostics) do
-      if api.nvim_buf_is_loaded(diagnostic.bufnr) then
+      if diagnostic.lnum <= line_count then
         api.nvim_buf_set_extmark(bufnr, ns.user_data.sign_ns, diagnostic.lnum, 0, {
           sign_text = text[diagnostic.severity] or text[M.severity[diagnostic.severity]] or 'U',
           sign_hl_group = sign_highlight_map[diagnostic.severity],


runtime/lua/vim/diagnostic.lua
```diff
@@ -1440,9 +1440,10 @@ M.handlers.signs = {
 
     local numhl = opts.signs.numhl or {}
     local linehl = opts.signs.linehl or {}
+    local line_count = api.nvim_buf_line_count(bufnr)
```
icholy, line 1443: This will fail if the buffer isn't loaded.

icholy:
Do you have a repro for this?

runtime/lua/vim/diagnostic.lua
ribru17: Ah, right, thanks!

ribru17:
Do you prefer repros without `nvim-lspconfig`? If so I can make one in a second, I was still figuring that part out. If you don't mind a repro with that plugin I can make one quicker

icholy:
I usually do something like this:

``` lua
        pattern = "go",
        callback = function()
	        vim.lsp.start({ cmd = { "gopls" } })
        end
})
```

But one with `nvim-lspconfig` is better than nothing.

runtime/lua/vim/diagnostic.lua
icholy: You can probably move the `api.nvim_buf_is_loaded(bufnr)` check higher up in this function.

ribru17:
Ok. Minimum repro:
```lua
-- Create a custom namespace. This will aggregate signs from all other
-- namespaces and only show the one with the highest severity on a
-- given line
local ns = vim.api.nvim_create_namespace('my_namespace')
-- Get a reference to the original signs handler
local orig_signs_handler = vim.diagnostic.handlers.signs
-- Override the built-in signs handler
vim.diagnostic.handlers.signs = {
  show = function(_, bufnr, _, opts)
    -- Get all diagnostics from the whole buffer rather than just the
    -- diagnostics passed to the handler
    local diagnostics = vim.diagnostic.get(bufnr)
    -- Find the "worst" diagnostic per line
    local max_severity_per_line = {}
    for _, d in pairs(diagnostics) do
      local m = max_severity_per_line[d.lnum]
      if not m or d.severity < m.severity then
        max_severity_per_line[d.lnum] = d
      end
    end
    -- Pass the filtered diagnostics (with our custom namespace) to
    -- the original handler
    local filtered_diagnostics = vim.tbl_values(max_severity_per_line)
    orig_signs_handler.show(ns, bufnr, filtered_diagnostics, opts)
  end,
  hide = function(_, bufnr)
    orig_signs_handler.hide(ns, bufnr)
  end,
}

vim.api.nvim_create_autocmd({ 'BufEnter', 'BufWinEnter' }, {
  pattern = { '*.lua' },
  callback = function()
    vim.lsp.start {
      name = 'lua_ls',
      cmd = { 'lua-language-server' },
      root_dir = vim.fs.root(0, {
        '.luarc.json',
        '.luarc.jsonc',
        '.luacheckrc',
        ' .stylua.toml',
        'stylua.toml',
        'selene.toml',
        'selene.yml',
        '.git',
      }),
    }
  end,
})
```

Looks like this error has to do with the default sign handler example. For a given `test.lua` file:

```
-- comment
-- here
err
```

If you open with `nvim --clean -u minimum_repro.lua test.lua`, then wait for the diagnostic signs to be published, doing `Gdk` gives the above error

runtime/lua/vim/diagnostic.lua
ribru17: Alright I've moved it out up top, also changing the parameter from `diagnostic.bufnr` to just the given `bufnr`. Hopefully this is fine

runtime/lua/vim/diagnostic.lua
```diff
@@ -1440,9 +1440,11 @@ M.handlers.signs = {
 
     local numhl = opts.signs.numhl or {}
     local linehl = opts.signs.linehl or {}
+    local buf_loaded = api.nvim_buf_is_loaded(bufnr)
+    local line_count = buf_loaded and api.nvim_buf_line_count(bufnr) or 0
 
     for _, diagnostic in ipairs(diagnostics) do
-      if api.nvim_buf_is_loaded(diagnostic.bufnr) then
+      if buf_loaded and diagnostic.lnum <= line_count then
```
icholy, line 1447: No need to iterate over each diagnostic if the bufnr is bad. Return early way higher up in the function.

icholy:
Code LGTM. But I'm not clear on why your signs override triggers the bug and the default one doesn't.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1441,8 +1441,14 @@ M.handlers.signs = {
     local numhl = opts.signs.numhl or {}
     local linehl = opts.signs.linehl or {}
 
+    if not api.nvim_buf_is_loaded(bufnr) then
+      return
```
justinmk, line 1445: is it intentional that this is done so late in the function? is that because the `text` and `opts` state needs to be updated even if the buffer isn't loaded?

justinmk (approved):

runtime/lua/vim/diagnostic.lua
ribru17: I don't think so -- I'll move it to the beginning. The `underline.show` function does this and seems to be fine

ribru17:
> Code LGTM. ~But I'm not clear on why your signs override triggers the bug and the default one doesn't. 🤔~

Did you ever figure out why this only happens with the override? :joy:

PR #29331, opened 2024-06-14
fix(lsp): tune completion word extraction for decorated labels

### LINKED ISSUE #29301: [LSP Completion] - dot character remains in completion results
### Problem

When using Neovim's new built in LSP completion, the "•" character is also provided in the completion. 

A video to demonstrate the issue is below:

https://github.com/neovim/neovim/assets/5642902/69beeccd-a8ee-4309-84e2-21a4a93fbe9e

Please notice that this is Clangd and it occurs with *most* symbols but not all. 

Once I switch back to mini.nvim completion the dots are no longer present in the completion. Screen cap below:

https://github.com/neovim/neovim/assets/5642902/faa86710-4420-42c7-9616-8c241bd76624

### Steps to reproduce

Use this on_attach function to enable lsp completion when LSP is attached:

```
-- on attach method for lsps
local on_attach = function(client, bufnr)
	vim.lsp.completion.enable(true, client.id, bufnr, { autotrigger = false })
end
```

This is all I need to do to reproduce this issue. 

### Expected behavior

The expected behavior is for the dot characters to be removed from the completion text. 

### Neovim version (nvim -v)

NVIM v0.11.0-dev+226-g721551210

N/A - vim does not have this feature

### Operating system/version

Fedora 40

### Terminal name/version

kitty 0.35.1 created by Kovid Goyal

xterm-kitty

### Installation

https://copr.fedorainfracloud.org/coprs/agriffis/neovim-nightly/

mfussenegger:
Problem:

For snippets lsp.completion prefers the label if it is shorter than the
insertText or textEdit to support postfix completion cases but clangd
adds decoration characters to labels. E.g.: `•INT16_C(c)`

Solution:

Use parse_snippet on insertText/textEdit before checking if it is
shorter than the label.

Fixes https://github.com/neovim/neovim/issues/29301

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -145,8 +145,8 @@ local function get_completion_word(item)
       --    label: insert
       --
       -- Typing `i` would remove the candidate because newText starts with `t`.
-      local text = item.insertText or item.textEdit.newText
-      return #text < #item.label and text or item.label
+      local text = parse_snippet(item.insertText or item.textEdit.newText)
+      return #text < #item.label and vim.fn.matchstr(text, '\\k*') or item.label
     elseif item.insertText and item.insertText ~= '' then
       return parse_snippet(item.insertText)
     else
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -81,9 +81,20 @@ describe('vim.lsp.completion: item conversion', function()
       -- plain text
       {
         label = 'foocar',
-        sortText = 'k',
+        sortText = 'g',
         insertText = 'foodar(${1:var1})',
         insertTextFormat = 1,
+      },
+      {
+        label = '•INT16_C(c)',
+        insertText = 'INT16_C(${1:c})',
+        insertTextFormat = 2,
+        filterText = 'INT16_C',
+        sortText = 'h',
+        textEdit = {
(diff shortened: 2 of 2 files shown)


runtime/lua/vim/lsp/completion.lua
```diff
@@ -145,8 +145,8 @@ local function get_completion_word(item)
       --    label: insert
       --
       -- Typing `i` would remove the candidate because newText starts with `t`.
-      local text = item.insertText or item.textEdit.newText
-      return #text < #item.label and text or item.label
+      local text = parse_snippet(item.insertText or item.textEdit.newText)
+      return #text < #item.label and vim.fn.matchstr(text, '\\k*') or item.label
```
mfussenegger, line 149: I'm not sure about the matchstr here. Without it we'd use `INT16_C(c)` as word. With it `INT16_C`

My opinion is that the latter is better if you _don't_ expand the snippet. Because after selecting an entry you can continue typing `(whatever`. If `(c)` is inserted too you instead need to backtrack.

If you expand the snippet, it will insert `INT16_C(c)` with `c` being the tabstop/placeholder as expected.

MariaSolOs: I agree that it's nice to not expand the snippet. Another argument if favour of that is that it might also be confusing to insert placeholders without starting a proper snippet session.

MariaSolOs (approved):

PR #29332, opened 2024-06-14
Added Georgian QWERTY keymap

namper:


clason:
Thank you, but we don't maintain these files. Please contribute them to Vim first, and we will port any additions.

PR #29335, opened 2024-06-14
Added a way to install the program in ALT Linux

Krator3:


justinmk:
Don't think we need to mention how to use apt-get for every debian derivative.

Thanks for your help , though.

Krator3:
> Don't think we need to mention how to use apt-get for every debian derivative.
> 
> Thanks for your help , though.

Sorry, but ALT is a standalone distribution. In addition, it does not support apt, only apt-get. Besides, I don't think that adding one line to a file will make anyone worse. Maybe you'll change your mind?

Krator3:
Other than that, ALT doesn't support sudo by default, so I think it's worth taking out separately. Well, or at least indicate ALT next to some APT-get command. And add a note to it that you need to use not sudo, but su

justinmk:
There are thousands of linux distros. We don't need to have a section for every distro. Especially distros that use apt. 

> And add a note to it that you need to use not sudo, but su

Why does that need to be mentioned in the install instructions for Nvim? Surely the steps are the same for every other package on this linux distro.

iDevPro (commented):
This is work variant, checked.

iDevPro:
> There are thousands of linux distros. We don't need to have a section for every distro. Especially distros that use apt.
> 
> > And add a note to it that you need to use not sudo, but su
> 
> Why does that need to be mentioned in Nvim's docs? Surely the steps are the same for every other package on this linux distro, this is not a Nvim-specific quirk.

Hello jusinmk,
This is not a regular Apt-like distribution, in AltLinux apt-get is used for rpm packages.
And, as with other distributions, they all have some deviations in the use of installation commands.
In addition, ALT is not preinstalled by default and the ability to use sudo is disabled, which distinguishes the installation process of programs.
What needs to be clarified or changed here in order for this patch to be reviewed by you?

dundargoc:
To be fair, we already have installation instructions for the most random/unheard of linux distros, so in that sense it's not egregious to at least be consistent on that part.

That being said, I kinda wonder if we should just remove all trivial installation instructions (i.e. anything that is just a `<package_manager> install neovim`). There's a new linux distro every other second and I don't see much point in us trying to continuously keep track of them.

clason:
We can also just link to https://repology.org/project/neovim/versions.

PR #29340, opened 2024-06-14
feat(snippet): support multiple sessions

MariaSolOs:
Part of https://github.com/neovim/neovim/issues/25696

Had kept this as a draft because the dynamic behavior of the snippet navigation keymaps was making this behavior not actually work. But now that [those are permanently set](https://github.com/neovim/neovim/pull/31887) the changes here are much simpler and the unit test passes now.

After a rebase this PR is basically the same as https://github.com/neovim/neovim/pull/38746, so credits to @notfirefox as well.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -89,12 +89,7 @@ HIGHLIGHTS
 
 LSP
 
-• LSP capabilities:
-  • `textDocument/foo` …
-    https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_codeLens
-  • `textDocument/bar` …
-    https://microsoft.github.io/language-server-protocol/specification/#textDocument_colorPresentation
-• todo
+• Support for nested snippets.
 
 LUA
 
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -89,6 +89,11 @@ HIGHLIGHTS
 
 LSP
 
+• LSP capabilities:
+  • `textDocument/foo` …
+    https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_codeLens
+  • `textDocument/bar` …
+    https://microsoft.github.io/language-server-protocol/specification/#textDocument_colorPresentation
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/snippet.lua
```diff
@@ -397,7 +397,7 @@ local function setup_autocmds(bufnr)
       local cursor_row, cursor_col = cursor_pos()
 
       -- The cursor left the snippet region.
-      local snippet_range = get_extmark_range(bufnr, M._session.extmark_id)
+      local snippet_range = get_extmark_range(bufnr, M._current_session.extmark_id)
```
justinmk, line 400: the old name seems fine. and is less verbose :)
```suggestion
      local snippet_range = get_extmark_range(bufnr, M._session.extmark_id)
```

MariaSolOs: Fair enough, I blame C# for making me use extra verbose names lol

runtime/doc/news.txt
```diff
@@ -89,12 +89,7 @@ HIGHLIGHTS
 
 LSP
 
-• LSP capabilities:
-  • `textDocument/foo` …
-    https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_codeLens
-  • `textDocument/bar` …
-    https://microsoft.github.io/language-server-protocol/specification/#textDocument_colorPresentation
```
justinmk, line 96: this "LSP capabilities" skeleton is intentional so that we remember to use it

runtime/doc/news.txt
```diff
@@ -89,12 +89,7 @@ HIGHLIGHTS
...
-• todo
+• Support for nested snippets.
```
justinmk, line 97:
```suggestion
• LSP capabilities:
  • `textDocument/foo` …
    https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_codeLens
  • `textDocument/bar` …
    https://microsoft.github.io/language-server-protocol/specification/#textDocument_colorPresentation
• Support for nested snippets.
```

PR #25461, opened 2023-10-01
fix(jumplist): Ctrl+o, Ctrl+i behave unintuitively when deleting buffers

vE5li:
Closes #25365.

These changes fix the weird behavior described in #25365, by applying two simple changes:
- remove all entries of a buffer from the jump list when deleting it
- not adding a new entry to the jump list if the next buffer to be displayed was also found in the jump list

This makes the jump list behave exactly as you would expect when deleting buffers

PR #29347, opened 2024-06-15
feat(jumplist): allow opting out of removing unloaded buffers

### LINKED ISSUE #28968: After closing a buffer, `Ctrl-o` no longer re-opens the buffer
### Problem

In v0.9.5, closing a buffer and then pressing `ctrl-o` re-opened the closed buffer. In v0.10.0, this is no longer the case, and `ctrl-o` jumps back to the previous position only in currently open buffers.

### Steps to reproduce

Assuming you have binaries for both v0.10.0 and v0.9.5 named `nvim-0.10.0` and `nvim-0.9.5` respectively:

```sh
nvim-0.9.5 --clean file1 file2
:bdelete
<C-o> # opens buffer that was just deleted
```

```sh
nvim-0.10.0 --clean file1 file2
:bdelete
<C-o> # does not open buffer that was just deleted
```

### Expected behavior

`<C-o>` should presumably behave as it did in 0.9.5, by opening deleted buffers

### Neovim version (nvim -v)

0.10.0 release, and latest prerelease 0e9c92a

Vim 9.1.400 does not have this issue

### Operating system/version

macOS 14.5

### Terminal name/version

Alacritty 0.13.2

tmux-256color

### Installation

Homebrew

zeertzjq:
Problem:  Cannot opt out of removing unloaded buffers from the jumplist.
Solution: Only enable that with "unload" flag in 'jumpoptions'.

Close #28968
Follow-up: #30418
The behavior before #25461 can now be achieved with `:set jumpoptions-=clean` or `:set jumpoptions=`.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -3619,7 +3619,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 	Otherwise only one space is inserted.
 
 						*'jumpoptions'* *'jop'*
-'jumpoptions' 'jop'	string	(default "")
+'jumpoptions' 'jop'	string	(default "unload")
 			global
 	List of words that change the behavior of the |jumplist|.
 	  stack         Make the jumplist behave like the tagstack.
@@ -3631,6 +3631,9 @@ A jump table for the options with a short description can be found at |Q_op|.
 	  view          When moving through the jumplist, |changelist|,
 			|alternate-file| or using |mark-motions| try to
 			restore the |mark-view| in which the action occurred.
+
+	  unload        Remove unloaded buffers from the jumplist.
+			EXPERIMENTAL: this flag may change in the future.
 
 						*'keymap'* *'kmp'*
 'keymap' 'kmp'		string	(default "")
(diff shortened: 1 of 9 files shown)


justinmk:
I was also hoping that @vE5li was still planning to simplify the jumplist code: https://github.com/neovim/neovim/pull/25461#issuecomment-2147506565

vE5li:
I already started working on it

hyhugh:
Should we make `-unload` the default behaviour to align with vim? as a person who uses vim daily, i recently upgraded to nvim and immediately noticed that <C-o> cannot reopen buffers that are closed via bdelete anymore, usually they should be removed from jump list only if i call `bwipeout`.

zeertzjq:
I'm considering backporting this as it's a flag that allows restoring the old behavior. Nevertheless I'm not entirely sure as backporting this means settling down on the name `unload`. There is now also #29639 that does more cleanup for `:bwipeout`.

runtime/doc/options.txt
```diff
@@ -3632,6 +3632,9 @@ A jump table for the options with a short description can be found at |Q_op|.
 			|alternate-file| or using |mark-motions| try to
 			restore the |mark-view| in which the action occurred.
 
+	  unload        Remove unloaded buffers from the jumplist.
+			EXPERIMENTAL: this flag may change in the future.
```
justinmk, line 3636: It's hard to think of a good name for this, but perhaps `clean` ?

zeertzjq: Ah yes, sounds good.

zeertzjq:
Follow-up: #30418

Issue #25027, opened 2023-09-05
homebrew: "Could NOT find Lpeg (missing: LPEG_LIBRARY)" on M2 Mac


nathanjcollins:
### Problem

I'm trying to update neovim from HEAD on an M2 MBP and I'm receiving "Could NOT find Lpeg (missing: LPEG_LIBRARY)" on its attempt to build. I was not receiving this yesterday, so likely something that was recently added?

I'm using `brew upgrade neovim --fetch-HEAD`

I've also tried removing and installing using  `brew install neovim --HEAD` but it results in the same.

```
==> Fetching neovim
==> Downloading https://raw.githubusercontent.com/Homebrew/homebrew-core/cdbc18aa2874d6efac7423fc61461537c1efb0e8/Formula/n/neovim.rb
Already downloaded: /Users/nathancollins/Library/Caches/Homebrew/downloads/88effdb6f5edccc0830c745d386456795bafb48bfb2e9a47862af2f3cb7a7347--neovim.rb
==> Cloning https://github.com/neovim/neovim.git
Updating /Users/nathancollins/Library/Caches/Homebrew/neovim--git
==> Checking out branch master
Already on 'master'
Your branch is up to date with 'origin/master'.
HEAD is now at c3e176f6e fix(options): correct condition for calling did_set_option() (#25026)
==> cmake -S . -B build -DLUV_LIBRARY=/opt/homebrew/opt/luv/lib/libluv.dylib -DLIBUV_LIBRARY=/opt/homebrew/opt/libuv/lib/libuv.dylib
Last 15 lines from /Users/nathancollins/Library/Logs/Homebrew/neovim/01.cmake:
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found Libuv: /opt/homebrew/opt/libuv/lib/libuv.dylib (Required is at least version "1.28.0")
-- Found Luv: /opt/homebrew/opt/luv/lib/libluv.dylib (Required is at least version "1.43.0")
-- Found Iconv: /Library/Developer/CommandLineTools/SDKs/MacOSX13.sdk/usr/include
CMake Error at /opt/homebrew/Cellar/cmake/3.27.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Lpeg (missing: LPEG_LIBRARY)
Call Stack (most recent call first):
[...]

PR #23410, opened 2023-04-30
build: make lpeg an imported library

dundargoc:
Cmake truncates the full link path to a shared library if it is missing
an SONAME in some undocumented scenarios. This causes builds in some
systems to fail if "lpeg" isn't a library on the system path.

The path of imported libraries aren't modified by cmake, so we can use
that as a workaround until a proper solution for this has been
identified.

Closes https://github.com/neovim/neovim/issues/23395.

PR #29375, opened 2024-06-17
build: fix CMake find_library search

albestro:
I faced a problem building Neovim from source on macOS. In particular the problem looks very similar/related to what has been reported in other issues/workarounds

- https://github.com/neovim/neovim/issues/25041
- https://github.com/neovim/neovim/issues/25027
- https://github.com/Homebrew/homebrew-core/pull/141673

After some digging I figured out that 

https://github.com/neovim/neovim/blob/20a7eebec086129e605041d32916f36df50890de/cmake/FindLpeg.cmake#L1

translates to:
- on linux `... NAMES lpeg_a lpeg liblpeg_a lpeg.so`
- on macos `... NAMES lpeg_a lpeg liblpeg_a lpeg.dylib`

## How it works

The CMake search mechanism turns out to look for either one of the given name "as is", or it "manipulates" the given name to create a regex with prefixes and suffixes.

Taking the second name given (i.e. `lpeg`) as example, CMake creates this regex

```
(lib)lpeg(\.tbd|\.dylib|\.so|\.a)
```

Turns out to either match with `lpeg` or 
- `liblpeg.tbd`
- `liblpeg.dylib`
- `liblpeg.so`
- `liblpeg.a`

None of this will match on Linux either...but in the given names we had also `lpeg.so`, obtained thanks to `lpeg${CMAKE_SHARED_LIBRARY_SUFFIX}`, so when it is its turn, it will find the match.

It is a bit of a workaround having it listed, because generally it is expected to find it with `lpeg`.

## Why does not work on macOS?

The problem is that on both platforms the `lpeg` library gets installed as `lpeg.so`, but `lpeg${CMAKE_SHARED_LIBRARY_SUFFIX}` ends up to be:
- on linux `lpeg.so`
- on macOS `lpeg.dylib`

Hence, for macOS the workaround of giving the exact name to look for does not work because of how the name is constructed.

## The fix

IMHO the fix is to make `find_library` work at its best. Since we are interested in looking for libraries that do not have the classic "lib" prefix, we instruct `find_library` to consider also this possibility, i.e. the library name to look for might also have an "empty" prefix and not just "lib".

In this way the regex ends up being (note the `|` before lib, indicating an empty option)

```
(|lib)lpeg(\.tbd|\.dylib|\.so|\.a)
```

that ends up looking for `lpeg` or:
- `lpeg.tbd`
- `lpeg.dylib`
- `lpeg.so`
- `lpeg.a`
- `liblpeg.tbd`
- `liblpeg.dylib`
- `liblpeg.so`
- `liblpeg.a`

Note: I took the chance to cleanup also other `find_library` calls to not explicitly have to deal with the prefix "lib".

It would be nice if someone having the issue with building with Homebrew could test it too.

Hope you find it useful and you are interested in having it merged.

### DIFF
--- a/cmake/Find.cmake
+++ b/cmake/Find.cmake
@@ -17,6 +17,7 @@ endfunction()
 # 2. Only search in .deps directory. Search all libraries
 # 3. Search everywhere, all libraries
 function(find_library2)
+  set(CMAKE_FIND_LIBRARY_PREFIXES "" "lib")
   find_library_nvim(STATIC ${ARGV})
   find_library_nvim(${ARGV})
   find_library(${ARGV})
--- a/cmake/FindIconv.cmake
+++ b/cmake/FindIconv.cmake
@@ -2,7 +2,7 @@
 # file can be removed once we decide to upgrade minimum cmake version.
 
 find_path2(ICONV_INCLUDE_DIR NAMES iconv.h)
-find_library2(ICONV_LIBRARY NAMES iconv libiconv)
+find_library2(ICONV_LIBRARY NAMES iconv)
 find_package_handle_standard_args(Iconv DEFAULT_MSG
   ICONV_INCLUDE_DIR)
 mark_as_advanced(ICONV_INCLUDE_DIR ICONV_LIBRARY)
--- a/cmake/FindLibuv.cmake
+++ b/cmake/FindLibuv.cmake
@@ -1,5 +1,5 @@
 find_path2(LIBUV_INCLUDE_DIR uv.h)
-find_library2(LIBUV_LIBRARY NAMES uv_a uv libuv)
+find_library2(LIBUV_LIBRARY NAMES uv_a uv)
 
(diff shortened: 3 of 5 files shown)


albestro:
It is worth adding that this fixes just the CMake side of the thing, but not the `lpeg` related issue on macOS with bundle vs dylib.

```console
ld: unsupported mach-o filetype (only MH_OBJECT and MH_DYLIB can be linked) in /lib/lua/5.1/lpeg.so'
```

But that's another story (if I will ever have time I will look into it, but this latter one does not sound like a `neovim` problem).

albestro:
After digging in the problem about linking mentioned in the last comment, I see that this cmake fix might interfere with other existing workarounds (e.g. https://github.com/neovim/neovim/pull/23410)

There are too many workarounds to circumvent this problem with Lua modules, and in particular with Lpeg. And I don't know enough about them. So, I will close this PR hoping that in the future it will be possible to finally fix the root problem and remove all these workarounds (and related patches existing in various package managers to deal with it).

PR #29380, opened 2024-06-17
Windows tilde expansion followup

rkitover:
@justinmk @zeertzjq as promised, here's the stuff left over from our discussion in #28515.

### DIFF
--- a/src/nvim/os/env.c
+++ b/src/nvim/os/env.c
@@ -412,6 +412,17 @@ void os_get_hostname(char *hostname, size_t size)
 static char *homedir = NULL;
 static char *os_uv_homedir(void);
 
+/// Public accessor for the cached "real", resolved user home directory. See
+/// comment on `homedir`.
+const char *os_get_homedir(void)
+{
+  if (!homedir) {
+    emsg("os_get_homedir failed: homedir not initialized");
+    return NULL;
+  }
+  return homedir;
+}
+
 void init_homedir(void)
 {
   // In case we are called a second time.
--- a/src/nvim/os/env.c
+++ b/src/nvim/os/env.c
@@ -400,17 +400,17 @@ void os_get_hostname(char *hostname, size_t size)
 ///   2. if $HOME is not set, try the following
 /// For Windows:
 ///   1. assemble homedir using HOMEDRIVE and HOMEPATH
-///   2. try os_homedir()
+///   2. try os_uv_homedir()
 ///   3. resolve a direct reference to another system variable
 ///   4. guess C drive
 /// For Unix:
-///   1. try os_homedir()
+///   1. try os_uv_homedir()
 ///   2. go to that directory
 ///     This also works with mounts and links.
 ///     Don't do this for Windows, it will change the "current dir" for a drive.
 ///   3. fall back to current working directory as a last resort
 static char *homedir = NULL;
-static char *os_homedir(void);
+static char *os_uv_homedir(void);
 
 void init_homedir(void)
 {
@@ -440,7 +440,7 @@ void init_homedir(void)
     }
   }
   if (var == NULL) {
-    var = os_homedir();
+    var = os_uv_homedir();
   }
 
   // Weird but true: $HOME may contain an indirect reference to another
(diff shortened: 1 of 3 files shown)


src/nvim/os/env.c
```diff
@@ -412,6 +412,15 @@ void os_get_hostname(char *hostname, size_t size)
 static char *homedir = NULL;
 static char *os_homedir(void);
 
+const char *os_get_homedir(void)
```
zeertzjq, line 415: Add a function about the difference of this function with `os_homedir()`?

zeertzjq: Or maybe one of these functions should be renamed, as it's confusing to have both `os_homedir()` and `os_get_homedir()`.

rkitover: You mean describe the differences in the commit message?

zeertzjq: Sorry, I mean add a comment above this function.

justinmk: does it make sense to rename the existing `os_homedir` to `uv_homedir` (and keep it private/static), then name this new function `os_homedir` ? And the docstring on the new function can mention:

    /// Gets the "real", resolved homedir. See comment on `homedir`.

test/functional/core/main_spec.lua
```diff
@@ -194,3 +194,41 @@ describe('command-line option', function()
     matches('Compilation: .*Run :checkhealth', fn.system({ nvim_prog_abs(), '-V1', '-v' }))
   end)
 end)
+
+if is_os('win') then
+  describe('Windows features', function()
+    it('expands ~/', function()
```
justinmk, line 200: this it() can live in the describe() block starting on line 17. Else this is missing before_all(...).

```suggestion
    it('expands ~/ on Windows', function()
```

test/functional/core/main_spec.lua
```diff
@@ -194,3 +194,41 @@ describe('command-line option', function()
...
+          '-es',
+          '+%print',
+          '+q',
+          '~/nvim_test.txt'
+        }):gsub('\n', '')
+      )
+      os.remove(fname)
+    end)
+
+    it('expands ~\\', function()
```
justinmk, line 217: test looks very redundant, can the code be deduplicated

src/nvim/os/env.c
zeertzjq: `uv_` prefix is for libuv functions. This is a wrapper around a libuv function, so I think it should at least be `os_uv_homedir`.

rkitover:
How's this?

src/nvim/os/env.c
rkitover: This good?

zeertzjq: I think you should reference the comment above the `homedir` variable.
> ```
> /// Gets the "real", resolved user home directory. See comment on `homedir`.
> ```

rkitover:
Good to go now?

test/functional/core/main_spec.lua
```diff
@@ -193,4 +193,25 @@ describe('command-line option', function()
...
+          'some text',
+          fn.system({
+            nvim_prog_abs(),
+            '-es',
+            '+%print',
+            '+q',
+            prefix .. 'nvim_test.txt',
+          }):gsub('\n', '')
+        )
+        os.remove(fname)
```
zeertzjq, line 213: I don't think it's necessary to use `os.remove()` before writing to the file. However, it's better to use `finally()`, so that the file is deleted even if the test fails.
```suggestion
        finally(function()
          os.remove(fname)
        end)
        write_file(fname, 'some text')
        eq(
          'some text',
          fn.system({
            nvim_prog_abs(),
            '-es',
            '+%print',
            '+q',
            prefix .. 'nvim_test.txt',
          }):gsub('\n', '')
        )
```

rkitover:
Fixed.

zeertzjq (approved):

zeertzjq:
Because 0.11 isn't released yet

rkitover:
Sorry, I forgot how the release branches work.

I fixed the merge conflict [here](https://github.com/neovim/neovim/pull/31254).

PR #29389, opened 2024-06-18
Move default statusline from C implementation to 'stl' expression

### LINKED ISSUE #28809: statusline: reimplement default, include LSP progress
### Problem

the default 'statusline' is implemented in C, it doesn't use the 'statusline' DSL. And it isn't readable as a 'statusline' expression, so user configs/plugins can't augment it.

### Expected behavior

- remove the C impl of 'statusline'. reimplement it as a 'statusline' expression
- show LspProgress in the default 'statusline' https://github.com/neovim/neovim/pull/28569
- it should continue to be minimal and fast, but consider other useful things we can add

yochem:
Partly fixes: #28809. After this is merged, we can look into implementing a Lua/Vimscript version of a more modern default statusline.

Let me know if/what docs I need to update. The current default is taken from :h 'statusline', where they describe it as `Emulate standard status line with 'ruler'`. I can change this to `Default statusline with 'ruler' set`. Should there also be a news entry?


famiu:
I cannot check to make sure as I do not have access to my PC right now, but this does not seem like the correct way of setting the new default to me. This only sets the initial default value instead of a fallback value. E.g. if you did `vim.o.statusline = ""`, I assume it wouldn't use the default statusline anymore, which breaks old behavior. Do correct me if I'm wrong.

yochem:
Yes, you're right. I forgot that setting strings option to an empty string usually reverts it to the default value. I'll look into your proposed approach.

famiu:
> The easiest way to set the default statusline would be to modify the `p_stl` or `wp->w_p_stl` when `NUL`:
> 
> ```c
>   } else if (*p_stl != NUL || *wp->w_p_stl != NUL) {
>     // redraw custom status line
>     redraw_custom_statusline(wp);
>   } else {
>     // default statusline case
> 
>     // define default statusline
>     char *default_stl = (char*)xmalloc(sizeof("hallo"));
>     strcpy(default_stl, "hallo");
>     p_stl = default_stl;
> 
>     redraw_custom_statusline(wp);
>   }
> ```
> (excuse the probably really bad C code)
> 
> But this seems like an antipattern to me:
> 
> ```lua
> vim.o.statusline = ""
> print(vim.o.statusline)
> -- prints "hallo" when built with the C code above
> ```
> 
> As @famiu suggested, another option would be to modify the functions that draw the stl (`win_redr_status`, `redraw_custom_statusline`, `win_redr_custom`, `build_stl_str_hl`) to accept a string value for the statusline, which could be set to the default in the else clause above.
> 
> The drawback here is that the default statusbar format is then still defined in statusbar.c, and I can imagine it not being ideal if we later want to define the default statusline from Lua/Vimscript.
> 
> Also, how would `nvim_eval_statusline` then be able to evaluate the statusline? Either `build_stl_str_hl` should know the default value of the statusline, or maybe a function in statusline.c `default_stl_str`, that returns the default statusline format? `nvim_eval_statusline` can then call this function to get the default.
> 

I don't think we should change option values as a side effect when building the statusline, that sounds like a recipe for disaster.

Instead, we can define a function that returns the default value, and use that in relevant places. This keeps the default value contained to a single place while avoiding any unnecessary hacks

src/nvim/statusline.c
```diff
@@ -90,89 +103,11 @@ void win_redr_status(win_T *wp)
     wp->w_redr_status = true;
   } else if (*p_stl != NUL || *wp->w_p_stl != NUL) {
     // redraw custom status line
-    redraw_custom_statusline(wp);
+    redraw_custom_statusline(wp, NULL);
```
famiu, line 106: Instead of passing `NULL` when the default format is not being used and the default format otherwise. Why not just always pass the statusline string that will be used (user-configured statusline string if it exists and default statusline string otherwise) instead of letting `redraw_custom_statusline` choose the statusline string when `NULL` is passed?

yochem: I wanted to still include a fallback if the fmt string was NULL, so that's why I did this. But makes sense! I'll change that when I start my work on this PR again.

yochem:
Closing in favor of https://github.com/neovim/neovim/pull/30721.

PR #29405, opened 2024-06-19
feat(filetype): detect `.envrc`, `.prettierignore`

tmillr:
Support ft detection for `.envrc` files used by direnv, and `.prettierignore` files used by prettier.


max397574:
afaik normally contributions like this should first be made to vim

clason:
All filetype logic is taken from Vim (even though the code is refactored). Please contribute any changes to there, and we will port them once they're merged.

tmillr:
Oh ok makes sense. I will do that. I wasn't totally sure because nvim has a different implementation for ft detection.

PR #29418, opened 2024-06-19
feat(diff): cancel linematch if ctrl-c is pressed while it's running

jwhite510:
@lewis6991 
In the case of running a linematch algorithm with a really large diff hunk that takes forever, the algorithm will hang and the user must stop the neovim process. With this pull request, I added an ability for the running algorithm to respect the ctrl + c press, similar to when you have a word completion (ctrl-p / ctrl-n) huge word list coming from many open buffers or large files, it will hang until you press ctrl + c to cancel it.

I'm not sure if i handled the capture of the ctrl-c the best way, but it seems to work. so some input on that would be appreciated if anyone knows a better way

### DIFF
--- a/src/nvim/buffer_defs.h
+++ b/src/nvim/buffer_defs.h
@@ -793,8 +793,8 @@ struct diffblock_S {
   diff_T *df_next;
   linenr_T df_lnum[DB_COUNT];           // line number in buffer
   linenr_T df_count[DB_COUNT];          // nr of inserted/changed lines
-  bool is_linematched;  // has the linematch algorithm ran on this diff hunk to divide it into
-                        // smaller diff hunks?
+  int is_linematched;  // has the linematch algorithm ran on this diff hunk to divide it into
+                       // smaller diff hunks? 0: not run yet / 1: success / -1: canceled
 
   bool has_changes;     ///< has cached list of inline changes
   garray_T df_changes;  ///< list of inline changes (diffline_change_T)
--- a/src/nvim/diff.c
+++ b/src/nvim/diff.c
@@ -501,7 +501,7 @@ static void diff_mark_adjust_tp(tabpage_T *tp, int idx, linenr_T line1, linenr_T
     }
 
     // check if this block touches the previous one, may merge them.
-    if ((dprev != NULL) && !dp->is_linematched && !diff_busy
+    if ((dprev != NULL) && (dp->is_linematched == 0) && !diff_busy
         && (dprev->df_lnum[idx] + dprev->df_count[idx] == dp->df_lnum[idx])) {
       for (int i = 0; i < DB_COUNT; i++) {
         if (tp->tp_diffbuf[i] != NULL) {
@@ -559,7 +559,7 @@ static diff_T *diff_alloc_new(tabpage_T *tp, diff_T *dprev, diff_T *dp)
 {
   diff_T *dnew = xcalloc(1, sizeof(*dnew));
 
-  dnew->is_linematched = false;
+  dnew->is_linematched = 0;
   dnew->df_next = dp;
   if (dprev == NULL) {
     tp->tp_first_diff = dnew;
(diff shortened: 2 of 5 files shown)


justinmk:
This is definitely wanted, thanks! Typically the way to check for ctrl-c is one of the breakcheck functions https://github.com/neovim/neovim/blob/b923fcbaf06276051372c17177a9970c4adc8e3d/src/nvim/os/input.c#L210-L228

E.g. if this is in a very hot loop, use `veryfast_breakcheck`. Does that not work?

justinmk:
@jwhite510 can you rebase this? any thoughts on my comment above?

jwhite510:
@justinmk 
rebased, and now using the the veryfast_breakcheck, that's exactly what I was looking for.

src/nvim/linematch.c
```diff
@@ -353,24 +362,46 @@ size_t linematch_nbuffers(const char **diff_blk, const int *diff_len, const size
...
-    (*decisions)[n_optimal++] = startNode->df_choice[j];
-    startNode = startNode->df_decision[j];
-  }
-  // reverse array
-  for (size_t i = 0; i < (n_optimal / 2); i++) {
-    int tmp = (*decisions)[i];
-    (*decisions)[i] = (*decisions)[n_optimal - 1 - i];
-    (*decisions)[n_optimal - 1 - i] = tmp;
+  if (canceled == true) { // if canceled, compare the lines guessing that the new lines are at the
+                          // bottom of the longer text
```
justinmk, line 371: would it simplify things we just bailed out of diff-mode entirely? I don't think we need to bother trying to construct partial "guessed" result.

justinmk:
Needs a rebase, and a test.

jwhite510:
@justinmk 
> Needs a rebase, and a test. And hopefully the code will be much simpler after removing the attempt to "guess" when the user cancels.

I removed the simplified guessed result, so the error handling is now moved outside linematch_nbuffers. I stored the failed state in diff_T.is_linematched, which is now an integer. 0: not yet ran, 1: success, -1: canceled. I'm not sure in the lua test how to simulate the ctrl-c press of the user to cancel it while it is still running.

test/functional/ui/linematch_spec.lua
```diff
@@ -1084,6 +1084,38 @@ something
       end
     )
   end)
+  describe('setup a diff with 2 files consisting of a large diff, cancel running linematch algorithm with ctrl-c and set linematch:8000', function()
+    before_each(function()
+      local f1 = [[
+a
+a
```
justinmk, line 1025: can use `t.dedent()` to make this look nicer :)

justinmk:
sorry for missing this, if you can rebase and CI looks good let's merge it.

src/nvim/diff.c
```diff
@@ -372,7 +372,7 @@ static void diff_mark_adjust_tp(tabpage_T *tp, int idx, linenr_T line1, linenr_T
     if (last >= line1 - 1) {
       // 6. change below line2: only adjust for amount_after; also when
       // "deleted" became zero when deleted all lines between two diffs.
-      if (dp->df_lnum[idx] - (deleted + inserted != 0) > line2 - dp->is_linematched) {
+      if (dp->df_lnum[idx] - (deleted + inserted != 0) > line2 - (dp->is_linematched == 1 ? 1 : 0)) {
```
justinmk: dropped in 4f0ab9877bef18d78714a2517dd2165c712a0be0

src/nvim/buffer_defs.h
```diff
@@ -793,8 +793,8 @@ struct diffblock_S {
   diff_T *df_next;
   linenr_T df_lnum[DB_COUNT];           // line number in buffer
   linenr_T df_count[DB_COUNT];          // nr of inserted/changed lines
-  bool is_linematched;  // has the linematch algorithm ran on this diff hunk to divide it into
-                        // smaller diff hunks?
+  int is_linematched;  // has the linematch algorithm ran on this diff hunk to divide it into
```
justinmk, line 796: `TriState` ? https://github.com/neovim/neovim/blob/516c0ba8a34c504f8806453537e4110299ccd8c8/src/nvim/types_defs.h#L48

PR #29425, opened 2024-06-20
build(vim-patch.sh): don't ignore changes to version*.txt

### LINKED ISSUE #29422: vim-patch PRs could also contribute to the news.txt
### Problem

i recently found `wildoptions=fuzzy` this amazing feature, as i did not follow news from the vim side, and i saw no mentions about that in news.txt, so i have missed that for the entire v0.9 period. which is such a pity that i could not discover it earlier to make life happier.

i just checked the recently added compleopt=fuzzy feature, it's not included in the news.txt either.

### Expected behavior

some features/changes brought by vim-patch PRs could be mentioned in the news.txt.

zeertzjq:
Suggest adding them to news.txt instead. Close #29422

Also don't ignore changes to intro.txt and sponsor.txt, as they don't
change much these days, and it's necessary to consider whether to
include their changes in Nvim's intro.txt.

### DIFF
--- a/runtime/doc/news-0.9.txt
+++ b/runtime/doc/news-0.9.txt
@@ -92,7 +92,7 @@ The following new APIs or features were added.
   showing a text representation of the nodes in a language tree for the current
   buffer.
 
-• |'statuscolumn'| option to customize the area to the side of a window,
+• 'statuscolumn' option to customize the area to the side of a window,
   normally containing the fold, sign and number columns. This new option follows
   the 'statusline' syntax and can be used to transform the line numbers, create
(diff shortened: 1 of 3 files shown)


justinmk:
The reason for skipping vim patches in our release notes is because it makes it much harder to see Neovim-specific improvements. We could have a single VIM PATCHES section in news.txt, that addresses this. It shouldn't include minor changes though.

zeertzjq:
Currently news.txt already contains some features introduced in Vim patches, so I think moving Vim patches to a separate section should be decided in a separate PR.

justinmk (approved):

PR #29428, opened 2024-06-20
feat(commenting): Add 'inner comment' text object

### LINKED ISSUE #29318: commenting: add 'inner comment' text object
### Problem

The builtin default commenting feature also ads the `gc` text object, so e.g. `dgc` allows to delete the comment block. It would also be useful if there is an 'inner comment' to delete everything but the commentstring.

Currently, this can also be done using nvim-treesitter-textobjects with the `@comment.inner` capture, although that works slightly different, as it does not grab the 'largest contiguous block of non-blank commented lines around the cursor'.

It's something I, and I expect others too, use quite often to rewrite a whole comment. It would be nice to not depend on nvim-treesitter-textobjects for this (small) functionality.

### Expected behavior

Example:
```lua
-- this is a comment
-- block
```
`cigc`:
```lua
-- |
```
Where `|` marks the cursor position.

Or a begin-end commentstring like HTML:
```html
<!-- html comment -->
```
`cigc`:
```html
<!-- | -->
```

yochem:
## Problem

No way to change a comment while keeping the markers. 

Currently changing a html comment:

```html
<!-- a |comment -->
```

Type: `cgc<!-- new comment -->`

```html
<!-- new comment -->|
```

I think changing an entire comment is done quite often by programmers.

## Solution

Add an 'inner comment' text object, which is in line with `it`, `i{`, `i[`, and other surrounding text objects.

The default keymap for this text object is `igc`.

With this change:
```html
<!-- a |comment -->
```

Type: `cigcnew comment`

```html
<!-- new comment| -->
```

Fixes #29318.

### DIFF
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -270,18 +270,7 @@ local function textobject(inner)
   end
 
   -- Select range from line:col to line:col for operator to act upon
-  vim.cmd(
-    'normal! '
-      .. lnum_from
-      .. 'G'
-      .. lcol_from
-      .. '|'
-      .. visual_cmd
-      .. lnum_to
-      .. 'G'
-      .. lcol_to
-      .. '|'
-  )
+  vim.cmd(('normal! %sG%s|%s%sG%s|'):format(lnum_from, lcol_from, visual_cmd, lnum_to, lcol_to))
 end
 
 return { operator = operator, textobject = textobject, toggle_lines = toggle_lines }
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -230,7 +230,8 @@ local function operator(mode)
 end
 
 --- Select contiguous commented lines at cursor
-local function textobject()
+---@param inner boolean?
+local function textobject(inner)
   local lnum_cur = vim.fn.line('.')
   local parts = get_comment_parts({ lnum_cur, vim.fn.col('.') })
   local comment_check = make_comment_check(parts)
@@ -251,8 +252,36 @@ local function textobject()
     lnum_to = lnum_to + 1
   end
 
-  -- Select range linewise for operator to act upon
-  vim.cmd('normal! ' .. lnum_from .. 'GV' .. lnum_to .. 'G')
+  -- Column does not matter if not inner, because of linewise selection
+  local lcol_from = 0
+  local lcol_to = 0
+  local visual_cmd = 'V'
+
+  -- Find the columns where the comment markers end
+  if inner then
+    local line_from = vim.api.nvim_buf_get_lines(0, lnum_from - 1, lnum_from, true)
+    local indent, _ = get_lines_info(line_from, parts)
+    lcol_from = indent:len() + parts.left:len() + 1
+
+    local line_to = vim.fn.getline(lnum_to)
+    lcol_to = line_to:len() - parts.right:len()
+
+    visual_cmd = 'v'
+  end
+
+  -- Select range from line:col to line:col for operator to act upon
+  vim.cmd(
+    'normal! '
+      .. lnum_from
+      .. 'G'
+      .. lcol_from
+      .. '|'
+      .. visual_cmd
+      .. lnum_to
+      .. 'G'
+      .. lcol_to
+      .. '|'
+  )
 end
 
 return { operator = operator, textobject = textobject, toggle_lines = toggle_lines }
--- a/test/functional/lua/comment_spec.lua
+++ b/test/functional/lua/comment_spec.lua
@@ -706,7 +706,7 @@ describe('commenting', function()
     end)
 
     it('leaves both comment markers in text', function()
-      set_commentstring("<!-- %s -->")
+      set_commentstring('<!-- %s -->')
       set_lines({ 'aa', '<!-- this is a comment -->', 'aa' })
       set_cursor(2, 0)
       feed('d', 'igc')
--- a/test/functional/lua/comment_spec.lua
+++ b/test/functional/lua/comment_spec.lua
@@ -648,4 +648,69 @@ describe('commenting', function()
       eq(get_lines(), { 'lua << EOF', 'EOF' })
     end)
   end)
+
+  describe('Inner textobject', function()
+    it('works', function()
+      set_lines({ 'aa', '# aa', '# aa', 'aa' })
+      set_cursor(2, 0)
+      feed('d', 'igc')
+      eq(get_lines(), { 'aa', '# ', 'aa' })
+    end)
+
+    it('allows dot-repeat', function()
+      set_lines({ 'aa', '# aa', '# aa', 'aa', '# aa' })
+      set_cursor(2, 0)
+      feed('d', 'igc')
+      set_cursor(4, 0)
+      feed('.')
+      eq(get_lines(), { 'aa', '# ', 'aa', '# ' })
+    end)
+
+    it('does nothing when not inside textobject', function()
+      -- Builtin operators
+      feed('d', 'igc')
+      eq(get_lines(), example_lines)
+
+      -- Comment operator
+      local validate_no_action = function(line, col)
+        set_lines(example_lines)
+        set_cursor(line, col)
+        feed('gc', 'gc')
+        eq(get_lines(), example_lines)
+      end
+
+      validate_no_action(1, 1)
+      validate_no_action(2, 2)
+    end)
+
+    it('respects tree-sitter injections', function()
+      setup_treesitter()
+      local lines = {
+        '"set background=dark',
+        '"set termguicolors',
+        'lua << EOF',
+        '-- print(1)',
+        '-- print(2)',
+        'EOF',
+      }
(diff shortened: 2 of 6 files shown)


echasnovski:
Thanks for the PR!

I personally don't think this is a good addition to built-in commenting for at least the following reasons:
- The "inner comment textobject" is mostly useful only in the `cigc` situation. The already present `cgc` preserves the indent (which is a tedious part to add) while inserting comment marks is *usually* not a big issue. Even if comment marks are verbose, ability to add them in Insert mode looks more ergonomic and can be mostly achieved with manual Insert mode mapping.
- Adding `igc` default textobject will interfere having `ig` textobject for users.

yochem:
> The "inner comment textobject" is mostly useful only in the `cigc` situation.

The `cigc` point is very true. I do however think this is something that many users do many times: rewriting an entire comment. They might expect the comment object to be in line with other surrounding objects, like tags.

> while inserting comment marks is _usually_ not a big issue.

This could also be said about inserting `()`, `[]`, and (html) tags. It's not that typing the comment marks is a lot of work, but rather that the current situation feels like there is missing something. I also don't think it's ideal to type `cgci<!-- comment -->` instead of `cigc comment`. 

> ability to add them in Insert mode looks more ergonomic and can be mostly achieved with manual Insert mode mapping.

This is an interesting idea, but might be hard for users to add, without the useful functions currently in this module. One would need to get the commentstring, parse left and right parts, insert those around the current cursor position. Not trivial.

> * Adding `igc` default textobject will interfere having `ig` textobject for users.

Haven't heard of this textobject, and can't find it in the docs. Where can I find this in the help pages?

echasnovski:
> This is an interesting idea, but might be hard for users to add, without the useful functions currently in this module. One would need to get the commentstring, parse left and right parts, insert those around the current cursor position. Not trivial.

I mean, maybe not trivial, but pretty straightforward in my opinion:

```lua
local insert_comment_parts = function()
  local cms = vim.bo.commentstring
  local offset = cms:find('%%s')
  local insert = cms:gsub('%%s', '')

  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
  vim.api.nvim_buf_set_text(0, row - 1, col, row - 1, col, { insert })
  vim.api.nvim_win_set_cursor(0, { row, col + offset - 1 })
end
vim.keymap.set('i', '<C-b>', insert_comment_parts)
```

The only drawback is that it uses buffer's commentstring instead of the one "at cursor", which built-in commenting uses. Exporting it (and the built-in variant of this functionality) was discussed in #28997.

runtime/lua/vim/_comment.lua
```diff
@@ -251,8 +252,36 @@ local function textobject()
...
+
+    visual_cmd = 'v'
+  end
+
+  -- Select range from line:col to line:col for operator to act upon
+  vim.cmd(
+    'normal! '
+      .. lnum_from
+      .. 'G'
+      .. lcol_from
```
justinmk, line 277: use ('%s...%s'):format(...)

justinmk:
> I also don't think it's ideal to type `cgci<!-- comment -->` instead of `cigc comment`.

Please add that to the Problem description :) That is a stronger, clearer motivating case.

test/functional/lua/comment_spec.lua
```diff
@@ -648,4 +648,69 @@ describe('commenting', function()
       eq(get_lines(), { 'lua << EOF', 'EOF' })
     end)
   end)
+
+  describe('Inner textobject', function()
```
justinmk, line 652:
```suggestion
  describe('inner textobject', function()
```

justinmk (approved):
Hard to argue against this. Code looks minimal, and for comments like `/** ... */` and `<!-- ... -->` it's definitely useful.

echasnovski:
@justinmk, [there are issues with this](https://github.com/neovim/neovim/pull/29428#issuecomment-2180222641). If this type of functionality is useful, exporting [this type of Insert mode mapping](https://github.com/neovim/neovim/pull/29428#issuecomment-2180369913) is a more versatile approach.

justinmk:
> @justinmk, [there are issues with this](https://github.com/neovim/neovim/pull/29428#issuecomment-2180222641). If this type of functionality is useful, exporting [this type of Insert mode mapping](https://github.com/neovim/neovim/pull/29428#issuecomment-2180369913) is a more versatile approach.

Yes, I read that and almost closed this issue, but the use-case is strong enough to make the case for this being a default.

>  Even if comment marks are verbose, ability to add them in Insert mode looks more ergonomic 

I don't get this argument. Textobjects are the primary text manipulation interface for vim. Insert-mode stuff is what vscode does.

>  Adding `igc` default textobject will interfere having `ig` textobject for users.

We added commenting so we need to claim a textobject. We can choose `igc` or `ic`. I'm fine with either.

echasnovski:
> >  Even if comment marks are verbose, ability to add them in Insert mode looks more ergonomic 
> 
> I don't get this argument. Textobjects are the primary text manipulation interface for vim. Insert-mode stuff is what vscode does.

The argument is that "insert comment marks" is a more versatile approach. For example, it can be used together with `o` or `O` to start a comment.

The suggested "inner textobject" is primarily (if not even only) useful in one particular situation of `cigc`. This can be the same as `cgc` followed by an "insert comment marks".

> >  Adding `igc` default textobject will interfere having `ig` textobject for users.
> 
> We added commenting so we need to claim a textobject. We can choose `igc` or `ic`. I'm fine with either. 

We did with `gc`. Claiming `igc` for built-in will break any potential `ig` user textobject. Claiming `ic` without `ac` is doable, but seems like wasting mapping space.

------

My suggestion is to do either of the following (from most to least preferred):
- Export function which will be able to "get commentstring at cursor". Either as a more general "get option at cursor" or the one in `vim._comment`.
- Do nothing.
- Make default Insert mode mapping to insert comment parts at cursor. This is lower than "Do nothing" because this is fairly easy to add to config.

Making built-in separate textobject that is useful only for a single operator is not on the list, I am afraid.

max397574:
> We did with gc. Claiming igc for built-in will break any potential ig user textobject. Claiming ic without ac is doable, but seems like wasting mapping space.

well then we couldn't add any new builtin mappings?
also I can't think of any textobject which you'd abbreviate with ig

and imo we should also add `ac` or `agc` as a textobject including comment markers
this is especially usefull to delete comments imo

yochem:
> The suggested "inner textobject" is primarily (if not even only) useful in one particular situation of cigc. This can be the same as cgc followed by an "insert comment marks".

Yanking only the text of a comment is definitely also a useful operator to have 'inner comment' for.

max397574:
because we just map `gc` and not `cgc`
having a mapping starting with `c` was the problem described in this issue afaict

echasnovski:
> well then we couldn't add any new builtin mappings?

Of course adding built-in mappings is possible. But having `igc` as Operator-pending mode mapping in particular means that users will have to wait for `timeoutlen` milliseconds after `cig` if they want to use `ig` as their textobject. This creates "either `ig` or `igc`" type of situation which I described is "breaking potential use of `ig`"

> ... also I can't think of any textobject which you'd abbreviate with ig

Me neither off the top of my head. But that doesn't mean not accounting for it is bad. Plus [nvim-various-textobjects](https://github.com/chrisgrieser/nvim-various-textobjs) uses it for "greedyOuterIndentation" together with `ag`.

> and imo we should also add `ac` or `agc` as a textobject including comment markers this is especially usefull to delete comments imo

This is already possible with `dgc`. If I understand correctly the suggestion ("outer" variant of the proposed here `igc` variant), then I doubt that "delete comment but leave indentation of its first line" is a common operation.

> Yanking only the text of a comment is definitely also a useful operator to have 'inner comment' for.

I agree, it is indeed useful. But the proposed `igc` textobject that works well for `cigc` will not work well for `yigc`. Because with multiline comment it will yank the whole charwise region **including** all comment parts except left of first line and right of last line.

> btw, why is `cgc` acceptable but we had to [revert](https://github.com/neovim/neovim/issues/28634) the `cr` mappings?

There is no `cgc` Normal mode mapping: there is built-in `c` operator and `gc` Operator-pending mode mapping (claimed as it is the established pattern for comment textobject, together with `gc` operator).

Creating `crr`/ `crn` *Normal mode* mappings also removes the possibility for `r` and `rr` / `rn` *Operator-pending* mode mappings. Meaning that:
- If user want to have `r` textobject, it will work with `c` built-in operator only after `timeoutlen` milliseconds wait.
- User can not use `rr`/`rn` textobjects with `c` operator but can with others.

echasnovski:
> Yes, that's true for multiple singleline comments following each other. However, it's something that could be fixed if this behaviour is desired.

As built-in commenting is designed to be used only with single line comments, I don't see a way to make `igc` textobject work as desired with both `c` (to delete comment block except indent and single comment markers) and `y` (to yank only commented text without per line comment markers) operators.

brianhuster:
> Could be changed to ic if we want to prevent that, or not create a default map for it.

Vim uses `ic` for inner comment textobject, though it seems to have a different meaning from this PR https://github.com/vim/vim/pull/16938

PR #23401, opened 2023-04-30
vim.ui.open: "gx" without netrw

mrshmllow:
Fix #23231

`vim.ui.os_open` is partially based off https://github.com/folke/lazy.nvim/blob/main/lua/lazy/util.lua#L13-L44, as recommended by @folke on the original issue.

- [x] Implement `vim.ui.os_open(uri: string | vim.uri.Uri)`
    - [x] Accept a `string` as `uri`
    - [ ] ~~Accept a `vim.uri.Uri` as `uri`~~
- [x] Implement "gx" in lua. The implementation drops current netrw options, including `g:netrw_gx`.
- [ ] Ability to change impl by using `vim.func.on_func(vim.ui, 'open_uri', …)`

PR #29443, opened 2024-06-21
feat(vim.ui.open): try PowerShell before Explorer

sybereal:
This comes from a similar motivation as #28424: explorer.exe does not work particularly well for opening arbitrary paths and URLs. However, in addition, due to an outstanding issue with WSL,[^1] explorer.exe always exits with a status of 1, resulting in an a superfluous error message being displayed.
PowerShell, on the other hand, does not suffer from this issue and preserves exit codes. Additionally, it handles URLs correctly that explorer.exe fails on.

[^1]: microsoft/WSL#6565

### DIFF
--- a/runtime/lua/vim/ui.lua
+++ b/runtime/lua/vim/ui.lua
@@ -151,12 +151,15 @@ function M.open(path)
     end
   elseif vim.fn.executable('wslview') == 1 then
     cmd = { 'wslview', path }
+  elseif vim.fn.executable('powershell.exe') == 1 then
+    cmd = { 'powershell.exe', '-c', 'start', path }
   elseif vim.fn.executable('explorer.exe') == 1 then
     cmd = { 'explorer.exe', path }
   elseif vim.fn.executable('xdg-open') == 1 then
     cmd = { 'xdg-open', path }
   else
-    return nil, 'vim.ui.open: no handler found (tried: wslview, explorer.exe, xdg-open)'
+    return nil,
+      'vim.ui.open: no handler found (tried: wslview, powershell.exe, explorer.exe, xdg-open)'
   end
 
   return vim.system(cmd, { text = true, detach = true }), nil
--- a/test/functional/lua/ui_spec.lua
+++ b/test/functional/lua/ui_spec.lua
@@ -153,7 +153,7 @@ describe('vim.ui', function()
         vim.fn.executable = function() return 0 end
       ]]
       eq(
-        'vim.ui.open: no handler found (tried: wslview, explorer.exe, xdg-open)',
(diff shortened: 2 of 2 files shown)


justinmk:
Thanks. What is the difference between `cmd.exe start` and `powershell start`? I would guess they invoke the same thing. But `cmd.exe` is presumably faster and more reliable (not sure if it returns the same exit codes though).

The only reason we used explore.exe in the first place was because of some purported problem with `start` (I don't remember what). Does `start` handle all URLs and filepaths (and director paths!) that explorer.exe handles? Does it open pdfs, images, directories, and URLs, in the same (or better) way?

justinmk:
Also in https://github.com/neovim/neovim/pull/23401#issuecomment-1621187483 I mentioned that netrw uses `rundll ...` which was presumably chosen because it is the most reliable and effective choice. So please test that and confirm whether it works as good or better.

sybereal:
I chose PowerShell because that is what wslview, which is currently already at a higher priority than explorer.exe due to the issue linked in my initial comment, uses internally. Therefore, what this change would achieve is that WSL users who don't have wslview installed would receive (mostly) the same functionality, avoiding the fallback to explorer.exe.

Alternatively, since you mentioned `rundll32`, that could also be used from WSL by removing the `has('win32')` check and adding an explicit `.exe`, so that it can be found from WSL's PATH. Since you seem to be partial to `rundll32`, would that solution be acceptable to you?

justinmk:
> I chose PowerShell because that is what wslview ... uses internally

seems like a detail worth mentioning in a one-line code comment, and the commit message.



> Since you seem to be partial to `rundll32`, would that solution be acceptable to you?

I'm partial to whatever works best. Up to now, no one has really confirmed/disconfirmed whether rundll32 works best. And it seems worth confirming that since this is the 5th PR that has tweaked this feature :)

sybereal:
I experimented with a number of other options, but came to the conclusion that using wslview is probably preferrable for most of my use cases.

I was wondering if maybe it could be nice to have an option to override the command used by `vim.ui.open` without overriding the entire function, but I guess the function is simple enough that copying and modifying it is justifiable.

My apologies for the back and forth. I'll be closing this.

justinmk:
> I experimented with a number of other options, but came to the conclusion that using wslview is probably preferrable for most of my use cases.

Thanks for confirming that
 
> I was wondering if maybe it could be nice to have an option to override the command used by `vim.ui.open` without overriding the entire function

In Nvim HEAD it supports an `opt.cmd` parameter

sybereal:
> In Nvim HEAD it supports an opt.cmd parameter

Ah great, that seems to be largely what I was thinking of. Thank you for your work.

PR #29454, opened 2024-06-22
fix(runtime): add commentstring for editorconfig ftplugin

matthewsia98:
Problem: No commentstring is set for `editorconfig` buffers

Solution: Add `ftplugin/editorconfig.vim` with `setlocal commentstring=#\ %s`

### DIFF
--- a/runtime/ftplugin/editorconfig.vim
+++ b/runtime/ftplugin/editorconfig.vim
@@ -1,0 +1,1 @@
+setlocal commentstring=#\ %s


clason:
Thank you, but can you contribute this to Vim first? They also have editorconfig and commenting plugins, and their default string doesn't apply here either.

matthewsia98:
seems like `.editorconfig` files have filetype `dosini` in Vim whereas they have filetype `editorconfig` in Neovim

clason:
That needs to be fixed then -- our filetype detection logic is supposed to be 100% identical to Vim's. 

@gpanders

Issue #12374, opened 2020-05-24
Prevent cursor from moving after using an operator


seb-mueller:
- `nvim --version`: v0.4.3
- `vim -u DEFAULTS` (version: ) behaves differently? vim behaves the same (all version I'm aware of)
- Operating system/version: Ubuntu 19.10
- Terminal name/version: yakuake
- `$TERM`: xterm-256color

### Steps to reproduce using `nvim -u NORC`

```
nvim -u NORC
```
Copy the following paragraph in a buffer and position the cursor somewhere on line2:
```
line1
line2
```
Type `yip` to copy the whole paragraph in the unnamed register.

### Actual behaviour
cursor **moves** to line1

### Desired behaviour
cursor should **stay** in the same position it was typing `yip`

Above is only an example for yank. This behaviour is the same for other operators as well and it would be ideal to have an global option or flat to prevent the cursor from moving. 

### Further info
Having asked a [question](https://vi.stackexchange.com/questions/25503/keep-cursor-from-moving-when-using-operators/25507?noredirect=1#comment44904_25507) on stackexchange, it turns out this feature is not available in (n)vim. It's a behaviour that has annoyed me for quite some time.

teto:
agreed that would be a nice fix. Patches welcome. I hope it won't break plugins too much.

seb-mueller:
> I hope it won't break plugins too much.

True. Maybe it would be best to keep it as it is as default,  but introduce a flag such that it can be turned off optionally, i.e. `set nocursor_movement` or so.

seb-mueller:
Just checking back to see if there is been any progress on this issue? It seems quite popular, so maybe a good enough reason to implement it?

teto:
seems like it was not a good enough reason for people to devote time to this yet. You are welcome to try.

seb-mueller:
Thanks, yes it seems like it, I'd love to try, but it's unfortunately beyond my programming skills (I'd probably need some hand holding), but I'll mention it in my next donation, maybe that bumps it up a little :)

justinmk:
related 1789

gh-liu:
You could use the `ModeChanged` event and the `vim.v.operator` variable to achieve this.

```lua
local api = vim.api
local autocmd = api.nvim_create_autocmd
local augroup = api.nvim_create_augroup

local g = augroup("user/keep_yank_position", { clear = true })

autocmd("ModeChanged", {
	pattern = { "n:no", "no:n" },
	group = g,
[...]

PR #29462, opened 2024-06-23
feat: add undokeepcurs option to keep the cursor position after undo

Grolaf:
feat: add undokeepcurs option to keep the cursor to the position prior to the undone operation.

Problem : 

By default, the cursor is set at the top of the undo block. That means that, no matter what kind of operation it was (e.g. auto-remove for import statement), your cursor will be set to the top position of the undo block after the undo operation. You may want more flexibilité on your cursor position depending on the operation).

Solution : 

Use the `:set undokeepcurs` option. Once you've set this option, your cursor will be set to the position where you were just after the action you want to undo. This can be useful, for example, if you are using autocommands to format your import declarations.

See an example of the result below :

https://github.com/neovim/neovim/assets/75337003/1963d7d4-87a7-4a15-b6a2-8b621588b1bb

This is my first contribution to neovim, feel free to blame me if I broke something, either the integrity or philosophy.


justinmk:
There was a request(s) for this in the issue tracker. Please link to it/them, and if there were tradeoffs mentioned, explain how this addresses the tradeoffs or why it should override the tradeoffs. And review any alternatives/workarounds discussed.

Is there an existing `fooopt` or `foooptions` option where this would fit? Possibly as a new 'cpoptions' flag.

Also needs tests.

gh-liu:
#12374

Issue #26575, opened 2023-12-14
Parse keycode into dictionary


altermo:
### Problem

As I know, there are no ways to parse a keycode into a dictionary.

### Expected behavior

A function that can parse a keycode into a dictionary, examples:
`<C-a>` -> `{mod={'ctrl'},key='A'}`
`<Home>` -> `{mod={},key='Home'}`
`<char-65>` -> `{mod={},key='A'}`

Maybe even support parsing multiple keycodes at once, for example:
`<C-a>B<char-67>` -> `{{mod={'ctrl'},key='A'},{mod={},key='B'}.{mod={},key='C'}}`

PR #29464, opened 2024-06-23
feat(lua): add vim.keymap.get

### LINKED ISSUE #24333: Feature request: add `vim.keymap.get`
### Problem

Currently, the only easy way to get a mapping with its callback is using `maparg()` which is not lua.

### Expected behavior

I'm basically suggesting to make `vim.keymap.get` to a wrapper around `maparg()` (or `nvim_get_keymap`)
You can use it as if it was like `maparg()`:
```lua
local map=vim.keymap.get('i','K',{})
if map.lhsraw==':w\rK' then
  --..
end
```
I'm also suggesting an option to have an opt `return_format='map'` which would return the same table as before but extended with such values that unpack it would give the right arguments to `vim.keymap.set` to redefine the map.
```lua
local map=vim.keymap.get('i','K',{return_format='map'})
--...
vim.keymap.set(unpack(map))
```

max397574:
Fix https://github.com/neovim/neovim/issues/24333

Adds a function to get a keymapping. This is really useful for saving and restoring mappings.

Example case: Neorg adds mappings (insert and normal mode) for it's api functions for each buffer and wants to remove and restore them
Before:
```lua
-- Get mappings
local mappings = vim
  .iter({ vim.api.nvim_buf_get_keymap(0, 'n'), vim.api.nvim_buf_get_keymap(0, 'n') })
  :flatten()
  :filter(function(mapping)
    if string.match(mapping.rhs, '^<Plug>(Neorg') then
      return true
    else
      return false
    end
  end)

-- Restore
for _, mapping in ipairs(mappings) do
  vim.keymap.set({ 'n', 'i' }, mapping.lhs, mapping.rhs, { desc = mapping.desc, buffer = true })
end
```

After:
```lua
-- Get mappings
local mappings = vim.keymap.get({ 'n', 'i' }, { rhs = '^<Plug>(Neorg' })

-- Restore
for _, mapping in ipairs(mappings) do
  vim.keymap.set({ 'n', 'i' }, mapping.lhs, mapping.rhs, mapping.opts)
end
```

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2891,6 +2891,31 @@ vim.keymap.del({modes}, {lhs}, {opts})                      *vim.keymap.del()*
     See also: ~
       • |vim.keymap.set()|
 
+vim.keymap.get({modes}, {filter})                           *vim.keymap.get()*
+    Gets mappings in a format easily usable for vim.keymap.set Examples: >lua
+        -- Gets all normal mode mappings
+        vim.keymap.get('n')
+
+        -- Gets a mapping which maps to a certain rhs
+        vim.keymap.get('n', { rhs = "<Plug>(MyAmazingFunction)" })
+<
+
+    Parameters: ~
+      • {modes}   (`string|string[]`)
+      • {filter}  (`table?`) A table with the following fields:
+                  • {lhs}? (`string`) Lhs of mapping
+                  • {rhs}? (`string`) Patter to match against rhs of mapping
+                  • {buffer}? (`integer|boolean`) Get a mapping for a certain
+                    buffer
+
+    Return: ~
+        (`table[]`) A list of objects with the following fields:
+        • {lhs} (`string`) Lhs of mapping
+        • {rhs} (`string|function`) Rhs of mapping (can be callback)
+        • {mode} (`string`) Mode of the mapping
+        • {opts}? (`vim.keymap.get.Return.Opts`) Get a mapping for a certain
+          buffer
+
 vim.keymap.set({mode}, {lhs}, {rhs}, {opts})                *vim.keymap.set()*
     Defines a |mapping| of |keycodes| to a function or keycodes.
 
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -330,7 +330,8 @@ LUA
 • |vim.fs.relpath()| gets relative path compared to base path.
 • |vim.fs.dir()| and |vim.fs.find()| now follow symbolic links by default,
   the behavior can be turn off using the new `follow` option.
-• |vim.text.indent()| indents/dedents text.
+- |vim.text.indent()| indents/dedents text.
+- Added |vim.keymap.get()| to get keymappings.
 
 OPTIONS
 
--- a/runtime/lua/vim/keymap.lua
+++ b/runtime/lua/vim/keymap.lua
@@ -41,7 +41,20 @@ local keymap = {}
 ---@see |maparg()|
 ---@see |mapcheck()|
 ---@see |mapset()|
+---@overload fun(opts: vim.keymap.get.Return)
 function keymap.set(mode, lhs, rhs, opts)
+  if type(mode) == 'table' and not lhs then
+    ---@type vim.keymap.get.Return
+    local keymap_get_opts = mode
+    mode = keymap_get_opts.mode
+    lhs = keymap_get_opts.lhs
+    rhs = keymap_get_opts.rhs
+    opts = {
+      expr = keymap_get_opts.opts.expr,
+      buffer = keymap_get_opts.opts.buffer,
+      desc = keymap_get_opts.opts.desc,
+    }
+  end
   vim.validate('mode', mode, { 'string', 'table' })
   vim.validate('lhs', lhs, 'string')
   vim.validate('rhs', rhs, { 'string', 'function' })
@@ -127,6 +140,113 @@ function keymap.del(modes, lhs, opts)
       vim.api.nvim_buf_del_keymap(buffer, mode, lhs)
     end
   end
+end
+
+--- @class vim.keymap.get.Filter
+--- @inlinedoc
+---
+--- Lhs of mapping
+--- @field lhs? string
+---
+--- Patter to match against rhs of mapping
+--- @field rhs? string
+---
+--- Get a mapping for a certain buffer
+--- @field buffer? integer|boolean
+
+--- @class vim.keymap.get.Return.Opts
+--- @inlinedoc
+---
+--- If they mapping is an expr mapping
+--- @field expr boolean
+---
+--- If they mapping is a buffer-local mapping
+--- @field buffer boolean
+---
+--- Description of the mapping
+--- @field desc? string
+
+--- @class vim.keymap.get.Return
+--- @inlinedoc
+---
+--- Lhs of mapping
+--- @field lhs string
+---
+--- Rhs of mapping (can be callback)
+--- @field rhs string|function
+--- Mode of the mapping
+--- @field mode string
+---
+--- Get a mapping for a certain buffer
+--- @field opts? vim.keymap.get.Return.Opts
+
+--- Gets mappings in a format easily usable for vim.keymap.set
+--- Examples:
+---
+--- ```lua
+--- -- Gets all normal mode mappings
+--- vim.keymap.get('n')
+---
+--- -- Gets a mapping which maps to a certain rhs
+--- vim.keymap.get('n', { rhs = "<Plug>(MyAmazingFunction)" })
+--- ```
+---
+---@param modes string|string[]
+---@param filter? vim.keymap.get.Filter
+---@return vim.keymap.get.Return[]
+function keymap.get(modes, filter)
+  filter = filter or {}
+
+  vim.validate('modes', modes, { 'string', 'table' })
+  vim.validate('filter', filter, { 'table' })
+
+  modes = type(modes) == 'string' and { modes } or modes
+  --- @cast modes string[]
+
+  local mappings = {}
+  for _, mode in ipairs(modes) do
+    if filter.buffer then
+      table.insert(
+        mappings,
+        ---@diagnostic disable-next-line: param-type-mismatch
+        vim.api.nvim_buf_get_keymap(filter.buffer == true and 0 or filter.buffer, mode)
+      )
+    else
+      table.insert(mappings, vim.api.nvim_get_keymap(mode))
+    end
+  end
+
+  local function matches(mapping)
+    local match = true
+    if filter.lhs then
+      match = match and vim.keycode(filter.lhs) == vim.keycode(mapping.lhs)
+    end
+    if filter.rhs then
+      match = match and vim.keycode(filter.rhs) == vim.keycode(mapping.rhs)
+    end
+    return match
+  end
+  return vim
+    .iter(mappings)
+    :flatten()
+    :filter(function(mapping)
+      return matches(mapping)
+    end)
+    :map(function(mapping)
+      ---@type vim.keymap.get.Return
+      return {
+        lhs = mapping.lhs,
+        -- For roundtripping: keymap.set() only accepts "rhs".
+        rhs = mapping.callback or mapping.rhs,
+        mode = mapping.mode,
+        opts = {
+          desc = mapping.desc,
+          buffer = mapping.buffer == 1,
+          expr = mapping.buffer == 1,
+        },
+      }
+    end)
+    :totable()
 end
 
 return keymap


clason:
What is the benefit over the already existing nvim API for this?

max397574:
a qol wrapper around it
could say the same for `vim.keymap.del`

clason:
That was a different time. New additions need to stand on their own feet.

clason:
By which I mean to say, what is the QoL exactly?

max397574:
well it's basically the same as `vim.keymap.del`
the user can just provide string arguments and doesn't have to write the filter logic himself.
If that's not reason enough for you to add this you can close this pr.

justinmk:
QoL improvements are very encouraged, but let's make sure they are strong QoL improvements, not thin wrappers around the existing API. (No comment on this particular case.) And such PRs should always show "before" and "after" examples.

max397574:
I don't have very strong arguments for it
I thought I'd just put together sth rq and we can discuss it here
I'll add a before and after later

dundargoc:
I understand. Shall we close the PR then if it's not desired in its current form?

clason:
If there is no interest by the OP in engaging in a wider discussion, I guess so.

clason:
We're still waiting for the promised before/after for your proposal...

max397574:
@clason @justinmk added before-after

clason:
Does it work with Lua functions as rhs?

max397574:
do you mean passing a function to `vim.keymap.get` or getting one e.g. by passing a lhs which maps a function which has a function as rhs?
former doesn't, latter does
wanted to test if former is also possible though

clason:
The latter, especially for the `get`--`set` use case. (The former doesn't really make sense.)

max397574:
well the former one would also be implementable I think
and could be useful to e.g. detect which mappings there which map to your api

which could be really useful

max397574:
okay
it seems like function as argument to `vim.keymap.get` is no implementable
I guess this is due to some things happening internally in neovim

runtime/lua/vim/keymap.lua
```diff
@@ -135,4 +135,112 @@ function keymap.del(modes, lhs, opts)
...
+--- vim.keymap.get('n')
+---
+--- -- Gets a mapping which maps to a certain rhs
+--- vim.keymap.get('n', { rhs = "<Plug>(MyAmazingFunction)" })
+--- ```
+---
+---@param modes string|string[]
+---@param opts? vim.keymap.get.Opts
+---@return vim.keymap.get.Return[]
+function keymap.get(modes, opts)
```
justinmk, line 188: is `opts` a filter here? I think we prefer to call it such if so. Example: https://github.com/neovim/neovim/blob/069451bb214bd9d97273ac92b37a25054df0f1a8/runtime/lua/vim/lsp.lua#L773-L775

```suggestion
---@param opts? vim.keymap.get.Filter
---@return vim.keymap.get.Return[]
function keymap.get(modes, filter)
```

runtime/lua/vim/keymap.lua
```diff
@@ -135,4 +135,112 @@ function keymap.del(modes, lhs, opts)
...
+    else
+      table.insert(keymaps, vim.api.nvim_get_keymap(mode))
+    end
+  end
+
+  local function matches(mapping)
+    local match = true
+    if opts.lhs then
+      opts.lhs = opts.lhs:gsub('<leader>', vim.g.mapleader or '')
+      opts.lhs = opts.lhs:gsub('<localleader>', vim.g.maplocaleader or '')
```
justinmk, line 215: If we use `maparg()` does it do this for us instead of having to hack it with `gsub()` ?

max397574: I'll try that out later

runtime/lua/vim/keymap.lua
```diff
@@ -135,4 +135,112 @@ function keymap.del(modes, lhs, opts)
...
+  return vim
+    .iter(keymaps)
+    :flatten()
+    :filter(function(mapping)
+      return matches(mapping)
+    end)
+    :map(function(mapping)
+      return {
+        lhs = mapping.lhs,
+        rhs = mapping.callback or mapping.rhs,
```
justinmk, line 234: This and similar cases is a good argument for adding keymap.get(). Because otherwise the programmer has to manually map this stuff; the result of `nvim_get_keymap` / `maparg()` is not useable with `keymap.set`.

runtime/lua/vim/keymap.lua
```diff
@@ -135,4 +135,112 @@ function keymap.del(modes, lhs, opts)
...
+--- @inlinedoc
+---
+--- Lhs of mapping
+--- @field lhs string
+---
+--- Rhs of mapping (can be callback)
+--- @field rhs string|function
+---
+--- Get a mapping for a certain buffer
+--- @field opts? integer|boolean
```
justinmk, line 172: This type should be `vim.keymap.set.Opts` , right?

runtime/lua/vim/keymap.lua
```diff
@@ -135,4 +135,112 @@ function keymap.del(modes, lhs, opts)
...
+--- If they mapping is an expr mapping
+--- @field expr boolean
+---
+--- If they mapping is a buffer-local mapping
+--- @field buffer boolean
+---
+--- Description of the mapping
+--- @field desc? string
+
+--- @class vim.keymap.get.Return
```
justinmk, line 162: Can/should `keymap.set` accept an overload so that this works:

    vim.keymap.set(vim.keymap.get(...))

runtime/lua/vim/keymap.lua
max397574: no
it's not exactly the same
this was also wrong though

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,115 @@ function keymap.del(modes, lhs, opts)
...
+--- @field mode string
+---
+--- Get a mapping for a certain buffer
+--- @field opts? vim.keymap.get.Return.Opts
+
+--- Gets mappings in a format easily usable for vim.keymap.set
+--- Examples:
+---
+--- ```lua
+--- -- Gets all normal mode keymaps
```
zeertzjq, line 191:
```suggestion
--- -- Gets all normal mode mappings
```

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,115 @@ function keymap.del(modes, lhs, opts)
...
+  vim.validate({
+    mode = { modes, { 's', 't' } },
+    filter = { filter, 't', true },
+  })
+
+  filter = filter or {}
+  modes = type(modes) == 'string' and { modes } or modes
+  --- @cast modes string[]
+
+  local keymaps = {}
```
zeertzjq, line 211:
```suggestion
  local mappings = {}
```

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,115 @@ function keymap.del(modes, lhs, opts)
...
+  for _, mode in ipairs(modes) do
+    if filter.buffer then
+      table.insert(
+        keymaps,
```
zeertzjq, line 215:
```suggestion
        mappings,
```

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,115 @@ function keymap.del(modes, lhs, opts)
...
+        ---@diagnostic disable-next-line: param-type-mismatch
+        vim.api.nvim_buf_get_keymap(filter.buffer == true and 0 or filter.buffer, mode)
+      )
+    else
+      table.insert(keymaps, vim.api.nvim_get_keymap(mode))
```
zeertzjq, line 220:
```suggestion
      table.insert(mappings, vim.api.nvim_get_keymap(mode))
```

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,115 @@ function keymap.del(modes, lhs, opts)
...
+    end
+    if filter.rhs then
+      filter.rhs = filter.rhs:gsub('<leader>', vim.g.mapleader or '')
+      filter.rhs = filter.rhs:gsub('<localleader>', vim.g.maplocaleader or '')
+      match = match and string.match(mapping.rhs, filter.rhs)
+    end
+    return match
+  end
+  return vim
+    .iter(keymaps)
```
zeertzjq, line 239:
```suggestion
    .iter(mappings)
```

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,115 @@ function keymap.del(modes, lhs, opts)
...
+    local match = true
+    if filter.lhs then
+      filter.lhs = filter.lhs:gsub('<leader>', vim.g.mapleader or '')
+      filter.lhs = filter.lhs:gsub('<localleader>', vim.g.maplocaleader or '')
+      match = match and filter.lhs == mapping.lhs
+    end
+    if filter.rhs then
+      filter.rhs = filter.rhs:gsub('<leader>', vim.g.mapleader or '')
+      filter.rhs = filter.rhs:gsub('<localleader>', vim.g.maplocaleader or '')
+      match = match and string.match(mapping.rhs, filter.rhs)
```
zeertzjq, line 234: This won't work for `<Leader>` and `<LocalLeader>` as they have different capitalization.

runtime/lua/vim/keymap.lua
zeertzjq: Another solution is to use `vim.keycode()` on both the mapping and the filter.

zeertzjq:
Related to the capitalization problem: #26575

runtime/lua/vim/keymap.lua
```diff
@@ -133,4 +146,111 @@ function keymap.del(modes, lhs, opts)
...
+    .iter(mappings)
+    :flatten()
+    :filter(function(mapping)
+      return matches(mapping)
+    end)
+    :map(function(mapping)
+      ---@type vim.keymap.get.Return
+      return {
+        lhs = mapping.lhs,
+        rhs = mapping.callback or mapping.rhs,
```
justinmk, line 244:
```suggestion
        -- For roundtripping: keymap.set() only accepts "rhs".
        rhs = mapping.callback or mapping.rhs,
```

max397574:
@justinmk anything that needs to be done on my side for this?

justinmk:
@max397574 please rebase. also the docs and news CI checks are giving some hints :)

max397574:
@justinmk rebased and fixed the CI
there is just one test failing but I can't see how that's related to my changes

max397574:
@justinmk rebased and fixed things up
anything left to do?

justinmk:
I want to get this in for 0.11, will check before EOW. Main thing I want to confirm is that roundtripping works:

    vim.keymap.set(..., vim.keymap.get(...))

max397574:
that should indeed work due to the new vim.keymap.set overload

justinmk:
> Definitely needs tests.

yes, that is required.

runtime/lua/vim/keymap.lua
```diff
@@ -41,7 +41,20 @@ local keymap = {}
 ---@see |maparg()|
 ---@see |mapcheck()|
 ---@see |mapset()|
+---@overload fun(opts: vim.keymap.get.Return)
 function keymap.set(mode, lhs, rhs, opts)
+  if type(mode) == 'table' and not lhs then
```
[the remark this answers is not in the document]
justinmk: Overloads are a basic mechanism that we must be allowed to leverage.

runtime/doc/lua.txt
```diff
@@ -2891,6 +2891,31 @@ vim.keymap.del({modes}, {lhs}, {opts})                      *vim.keymap.del()*
     See also: ~
       • |vim.keymap.set()|
 
+vim.keymap.get({modes}, {filter})                           *vim.keymap.get()*
+    Gets mappings in a format easily usable for vim.keymap.set Examples: >lua
+        -- Gets all normal mode mappings
+        vim.keymap.get('n')
+
+        -- Gets a mapping which maps to a certain rhs
+        vim.keymap.get('n', { rhs = "<Plug>(MyAmazingFunction)" })
```
justinmk, line 2900: show a roundtrip example.

runtime/doc/lua.txt
```diff
@@ -2891,6 +2891,31 @@ vim.keymap.del({modes}, {lhs}, {opts})                      *vim.keymap.del()*
...
+    Parameters: ~
+      • {modes}   (`string|string[]`)
+      • {filter}  (`table?`) A table with the following fields:
+                  • {lhs}? (`string`) Lhs of mapping
+                  • {rhs}? (`string`) Patter to match against rhs of mapping
+                  • {buffer}? (`integer|boolean`) Get a mapping for a certain
+                    buffer
+
+    Return: ~
+        (`table[]`) A list of objects with the following fields:
```
justinmk, line 2912:
```suggestion
        (`table[]`) List of keymaps with these fields:
```

runtime/doc/news.txt
```diff
@@ -330,7 +330,8 @@ LUA
 • |vim.fs.relpath()| gets relative path compared to base path.
 • |vim.fs.dir()| and |vim.fs.find()| now follow symbolic links by default,
   the behavior can be turn off using the new `follow` option.
-• |vim.text.indent()| indents/dedents text.
+- |vim.text.indent()| indents/dedents text.
+- Added |vim.keymap.get()| to get keymappings.
```
justinmk, line 334:
```suggestion
• |vim.text.indent()| indents/dedents text.
• Added |vim.keymap.get()| to get keymappings.
```

max397574:
will look at the reviews and add tests later this week.

max397574:
Closing this as I was quite directly told that "contributions from me aren't welcomed here"

Even though I got no way to enforce this I'd appreciate it if my code wouldn't just get directly reused to open another pr.

Issue #29468, opened 2024-06-24
vim.filetype.add: Impossible to override asm extensions (.S, .s)


levioneyh:
### Problem

Related: #19996 

.s and .S files are not forced and instead recognized as the default "asm" filetype. I've tried:

1. Put the code in .config/nvim/after/ftdetect/ and .config/nvim/ftdetect directories. 
2. Delete .local/share/nvim and .local/state/nvim.
3. Repeat the same steps for an arch VM without GUI where i never downloaded any nvim plugin or made any configuration changes.
4. Download the nvim.appimage (NVIM v0.11.0-dev-302+gbe999e6a0) from [releases](https://github.com/neovim/neovim/releases/nightly) and repeat steps 1 and 2.

### Steps to reproduce

In init.lua:

```lua
vim.filetype.add({
	extension = {
		c = "lua",
		TEST = "lua",
		s = "lua",
		S = "lua",
		asm = "lua",
	},
})
```

only .c, .TEST, and .asm files are overriden correctly

### Expected behavior

.s and .S files are correctly forced to the specified types.

### Neovim version (nvim -v)

NVIM v0.10.0 Build type: Release LuaJIT 2.1.1716656478

### Vim (not Nvim) behaves the same?

-

### Operating system/version

Arch linux, 6.9.6-arch1-1

### Terminal name/version

kitty, linux console

### $TERM environment variable

xterm-kitty, linux

### Installation

Pacman

clason:
Try adding `priority`.

clason:
Works for me; please always test with `nvim --clean -u init.lua` to make sure you don't have plugins interfering. If that is the case, try adding a `priority`.

zeertzjq:
It works if you add it in the `pattern` table:
```lua
vim.filetype.add({
	extension = {
		c = "lua",
		TEST = "lua",
		asm = "lua",
	},
	pattern = {
		['.*%.[sS]'] = "lua",
	},
})
```

clason:
>  However, I think there should be a way for user extensions to override builtin patterns.

That is difficult due to the architecture of the filetype detection (if not impossible without a full,  burn-it-to-the-ground-first, redesign). 

I think this should just be documented better, including with a concrete example. (The `vim.filetype.add` documentation implies this, but only after you have grokked the logic.)

zeertzjq:
Moving `asm` patterns to the `extension` table in Nvim will also make this work:
```diff
[...]

PR #29472, opened 2024-06-24
refactor(filetype): change some patterns to extensions

zeertzjq:
Ref #29468


gpanders:
You beat me to it. I was in the process of making this change too.

I also found some patterns that could be filename entries, e.g. `justfile` and `Justfile`. But maybe we can save that for another PR and just stick with extensions for now.

gpanders (approved):

justinmk (approved):

PR #29483, opened 2024-06-25
refactor(typval)!: remove binary distinction of binary and nonbinary strings

bfredl:
This is a modest proposal of a breaking change which will make refactor of typval and shada code a lot easier. In particular, code that would use or check for `v:msgpack_types.binary` in the wild would be broken.

Also some cases where `v:msgpack_type.string` would be used to represent a binary string of "string" type, we use a BLOB instead, which is vim script's native type for binary blobs. (we don't "innovate" in vim script, but since we have included blobs, we can just as well use them)

`msgpackdump(msgpackparse(data))` will no longer preserve the distinction of BIN and STR strings. This is very common behavior for language-specific msgpack bindings. Nvim uses msgpack as a tool to serialize its data. Nvim is not a tool to bit-perfectly manipulate arbitrary msgpack data out in the wild.

The changed tests should indicate how behavior changes in various edge cases.

Usage of `v:msgpack_types` in the wild seems very rare, at least from a cursory glance at github code search.


clason:
> Usage of v:msgpack_types in the wild seems very rare, at least from a cursory glance at github code search.

In fact, outside of Neovim or its forks, I can only find https://github.com/bfredl/nvim-x11/blob/d3b3c8ef17ca07dcb4a55a121a68073d7364a2fa/autoload/provider/clipboard.vim#L26 ;)

PR #29490, opened 2024-06-26
feat(vim.ui.open): configurable opener

### LINKED ISSUE #29488: Make the command run in `vim.ui.open` configurable
### Problem

In the bad old days when `netrw` was used, it was possible to configure the callback to `gx` command with the variable `vim.g.netrw_browsex_viewer` and it was possible to stuff almost anything there. So, I had in my configuration

```lua
vim.g.netrw_browsex_viewer='setsid osurl'
```

where [`osurl`](https://git.sr.ht/~mcepl/vim-suse-changes/tree/master/item/osurl) is a short bash script preprocessing [SUSE common abbreviations](https://en.opensuse.org/openSUSE:Packaging_Patches_guidelines#Current_set_of_abbreviations). The result is that when I run `gx` on one of those abbreviations (e.g., `gh#neovim/neovim#29443`) it would expand the abbreviation to the full URL (`https://github.com/neovim/neovim/pull/29443`) and open that URL with `xdg-open`.

`vim.ui.open` doesn’t have any such extension capability, but it should be relatively easy to do so:

```diff
diff --git a/runtime/lua/vim/ui.lua b/runtime/lua/vim/ui.lua
index f168da4955..cf92accd2a 100644
--- a/runtime/lua/vim/ui.lua
+++ b/runtime/lua/vim/ui.lua
@@ -141,7 +141,9 @@ function M.open(path)
 
   local cmd --- @type string[]
 
-  if vim.fn.has('mac') == 1 then
+  if vim.fn.executable(vim.g.netrw_browsex_viewer) == 1 then
+    cmd = { vim.g.netrw_browsex_viewer, path}
+  elseif vim.fn.has('mac') == 1 then
     cmd = { 'open', path }
   elseif vim.fn.has('win32') == 1 then
     if vim.fn.executable('rundll32') == 1 then
```

(of course, `vim.g.netrw_browsex_viewer` is just a legacy placeholder some better name should be found)

### Expected behavior

See above

mcepl:
With the welcomed demise of `netrw`, we lost an opportunity to use an arbitrary command for opening URL-like strings with `gx`. This should suggest a way how to recover this functionality.

Of course, new name of the variable `vim.g.netrw_browsex_viewer` should be developed (I am not sure what are the standards for creating new options).

Another thing which is missing are tests. I tried to create one, but I got completely lost and frankly `neovim` is not my primary editor, so I don’t have a time to dive too deep into its innards.

Will eventually Fix: https://github.com/neovim/neovim/issues/29488

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2559,7 +2559,7 @@ vim.ui.input({opts}, {on_confirm})                            *vim.ui.input()*
                       typed (it might be an empty string if nothing was
                       entered), or `nil` if the user aborted the dialog.
 
-vim.ui.open({path})                                            *vim.ui.open()*
+vim.ui.open({path}, {opt})                                     *vim.ui.open()*
     Opens `path` with the system default handler (macOS `open`, Windows
     `explorer.exe`, Linux `xdg-open`, …), or returns (but does not show) an
     error message on failure.
@@ -2570,6 +2570,8 @@ vim.ui.open({path})                                            *vim.ui.open()*
         -- Asynchronous.
         vim.ui.open("https://neovim.io/")
         vim.ui.open("~/path/to/file")
+        -- Use the "osurl" command to handle the path or URL.
+        vim.ui.open("gh#neovim/neovim!29490", { cmd = { 'osurl' } })
         -- Synchronous (wait until the process exits).
         local cmd, err = vim.ui.open("$VIMRUNTIME")
         if cmd then
@@ -2579,6 +2581,8 @@ vim.ui.open({path})                                            *vim.ui.open()*
 
     Parameters: ~
       • {path}  (`string`) Path or URL to open
+      • {opt}   (`{ cmd?: string[] }?`) Options
+                • cmd string[]|nil Command used to open the path or URL.
 
     Return (multiple): ~
         (`vim.SystemObj?`) Command object, or nil if not found.
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -194,7 +194,10 @@ TUI
 
 UI
 
-• TODO
+• |vim.ui.open()| (by default bound to |gx|) accepts an `opt.cmd` parameter
+  which controls the tool used to open the given path or URL. If you want to
+  globally set this, you can override vim.ui.open using the same approach
+  described at |vim.paste()|.
 
 ==============================================================================
 CHANGED FEATURES                                                 *news-changed*
--- a/runtime/lua/vim/ui.lua
+++ b/runtime/lua/vim/ui.lua
@@ -117,6 +117,8 @@ end
 --- -- Asynchronous.
 --- vim.ui.open("https://neovim.io/")
 --- vim.ui.open("~/path/to/file")
+--- -- Use the "osurl" command to handle the path or URL.
+--- vim.ui.open("gh#neovim/neovim!29490", { cmd = { 'osurl' } })
 --- -- Synchronous (wait until the process exits).
 --- local cmd, err = vim.ui.open("$VIMRUNTIME")
 --- if cmd then
@@ -125,12 +127,14 @@ end
 --- ```
 ---
 ---@param path string Path or URL to open
+---@param opt? { cmd?: string[] } Options
+---     - cmd string[]|nil Command used to open the path or URL.
 ---
 ---@return vim.SystemObj|nil # Command object, or nil if not found.
 ---@return nil|string # Error message on failure, or nil on success.
 ---
 ---@see |vim.system()|
-function M.open(path)
+function M.open(path, opt)
   vim.validate({
     path = { path, 'string' },
   })
@@ -139,12 +143,13 @@ function M.open(path)
     path = vim.fs.normalize(path)
   end
 
-  local cmd --- @type string[]
-  local opts --- @type vim.SystemOpts
+  opt = opt or {}
+  local cmd ---@type string[]
+  local job_opt = { text = true, detach = true } --- @type vim.SystemOpts
 
-  opts = { text = true, detach = true }
-
-  if vim.fn.has('mac') == 1 then
+  if opt.cmd then
+    cmd = vim.list_extend(opt.cmd --[[@as string[] ]], { path })
+  elseif vim.fn.has('mac') == 1 then
     cmd = { 'open', path }
   elseif vim.fn.has('win32') == 1 then
     if vim.fn.executable('rundll32') == 1 then
@@ -154,8 +159,8 @@ function M.open(path)
     end
   elseif vim.fn.executable('xdg-open') == 1 then
     cmd = { 'xdg-open', path }
-    opts.stdout = false
-    opts.stderr = false
+    job_opt.stdout = false
+    job_opt.stderr = false
   elseif vim.fn.executable('wslview') == 1 then
     cmd = { 'wslview', path }
   elseif vim.fn.executable('explorer.exe') == 1 then
@@ -164,7 +169,7 @@ function M.open(path)
     return nil, 'vim.ui.open: no handler found (tried: wslview, explorer.exe, xdg-open)'
   end
 
-  return vim.system(cmd, opts), nil
+  return vim.system(cmd, job_opt), nil
 end
 
 --- Returns all URLs at cursor, if any.
--- a/test/functional/lua/ui_spec.lua
+++ b/test/functional/lua/ui_spec.lua
@@ -157,5 +157,28 @@ describe('vim.ui', function()
         exec_lua [[local _, err = vim.ui.open('foo') ; return err]]
       )
     end)
+
+    it('opt.cmd #29490', function()
+      t.matches(
+        'ENOENT: no such file or directory',
+        t.pcall_err(exec_lua, function()
+          vim.ui.open('foo', { cmd = { 'non-existent-tool' } })
+        end)
+      )
+
+      eq(
+        {
+          code = 0,
+          signal = 0,
+          stderr = '',
+          stdout = 'arg1=arg1;arg2=https://example.com;',
+        },
+        exec_lua(function(cmd_)
+          local cmd, err = vim.ui.open('https://example.com', { cmd = cmd_ })
+          assert(cmd and not err)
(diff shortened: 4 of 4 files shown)


mcepl:
I have no idea what’s the `docs / docs` test problem.

gpanders:
This seems like another eventual candidate for `vim.ui.config()`: https://github.com/neovim/neovim/issues/25514

In the meantime though, supporting this variable for backwards compatibility seems fine to me. Although I am not sure that we should document it (we don't want _new_ code to start using this `netrw` variable).

cc @justinmk

runtime/lua/vim/ui.lua
```diff
@@ -141,7 +141,9 @@ function M.open(path)
 
   local cmd --- @type string[]
 
-  if vim.fn.has('mac') == 1 then
+  if vim.fn.executable(vim.g.netrw_browsex_viewer) == 1 then
```
gpanders, line 144: `vim.fn.executable()` will error with a `nil` argument.

```suggestion
  if vim.g.netrw_browsex_viewer and vim.fn.executable(vim.g.netrw_browsex_viewer) == 1 then
```

runtime/doc/lua.txt
```diff
@@ -2555,6 +2555,9 @@ vim.ui.open({path})                                            *vim.ui.open()*
     `explorer.exe`, Linux `xdg-open`, …), or returns (but does not show) an
     error message on failure.
 
+    Instead of the system default handler the function calls a command set in
+    the 'vim.g.netrw_browsex_viewer', if it is present.
+
```
gpanders, line 2560: My preference would be to leave support for this `netrw` variable undocumented (we can silently support it to make people transitioning from netrw easier).

If we want to document another way to do this, I'd prefer using a different name for this variable. Perhaps this could even be a variable in the `vim.ui` module/namespace (e.g. `vim.ui.opener = 'foobar'`).

mcepl: Yes, not documenting it is one way, and of course, I haven’t expected (and I mentioned it above) that this name would survive, but I don’t know enough about naming conventions in `neovim` to suggest a better name.

justinmk (changes requested):
supporting this old variable is not acceptable, it will prolong its life.

instead,

1. could add a optional `cmd` parameter: `open(path, cmd)`
2. users can override `vim.ui.open` using the pattern described in `:help vim.paste` and pass a hardcoded `cmd` of their choice.

> This seems like another eventual candidate for `vim.ui.config()`: #25514

in this case it seems that `on_func` would be sufficient : https://github.com/neovim/neovim/pull/22598

runtime/lua/vim/ui.lua
```diff
@@ -117,6 +120,8 @@ end
 --- -- Asynchronous.
 --- vim.ui.open("https://neovim.io/")
 --- vim.ui.open("~/path/to/file")
+--- -- Command 'osurl' knows what to do with this reference
+--- vim.ui.open("gh#neovim/neovim!29490", "osurl")
```
justinmk, line 124: is `osurl` something we should add as a fallback to the builtin impl?

mcepl: No, it is just [my script](https://git.sr.ht/~mcepl/vim-suse-changes/tree/master/item/osurl) which handles [SUSE abbreviated URLs](https://en.opensuse.org/openSUSE:Packaging_Patches_guidelines#Current_set_of_abbreviations). It is meant just as an example of a random script.

runtime/doc/news.txt
```diff
@@ -185,7 +185,10 @@ TUI
 
 UI
 
-• TODO
```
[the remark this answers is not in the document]
dundargoc: `|vim.ui.open|` should be `|vim.ui.open()|`, `'gx'` should be `|gx|` and `cmdpath` doesn't even seem to exist.

mcepl:
On Sun Sep 1, 2024 at 2:01 PM CEST, dundargoc wrote:
> `|vim.ui.open|` should be `|vim.ui.open()|`, `'gx'` should be `|gx|` and `cmdpath` doesn't even seem to exist.

`cmdpath` is the new optional parameter of `vim.ui.open()` function.

runtime/lua/vim/ui.lua
```diff
@@ -125,12 +127,15 @@ end
 --- ```
 ---
 ---@param path string Path or URL to open
+---@param cmdpath string|string[] optional command to be used for opening URL
```
justinmk, line 130: don't need to mention "optional" in the docstring, just use the `?` syntax. Also would you mind changing this to a `opts` style param (`:help kwargs`). 

Then this PR should be good to go. I have a followup here (I'll rebase it after merging this one): https://github.com/neovim/neovim/pull/30271

```suggestion
---@param opts? string|string[] command to be used for opening URL
```

runtime/lua/vim/ui.lua
```diff
@@ -117,6 +117,8 @@ end
 --- -- Asynchronous.
 --- vim.ui.open("https://neovim.io/")
 --- vim.ui.open("~/path/to/file")
+--- -- Command 'osurl' knows what to do with this reference
+--- vim.ui.open("gh#neovim/neovim!29490", "osurl")
```
justinmk, line 121:
```suggestion
--- -- Try the "osurl" command to handle the path, before falling back to other tools.
--- vim.ui.open("gh#neovim/neovim!29490", { cmd = 'osurl' })
```

runtime/lua/vim/ui.lua
```diff
@@ -139,12 +144,17 @@ function M.open(path)
...
   local opts --- @type vim.SystemOpts
 
   opts = { text = true, detach = true }
 
-  if vim.fn.has('mac') == 1 then
+  if cmdpath then
+    -- not checking for the existence of the cmd command, because
+    -- it could be a more complicated string or even list of
+    -- string, not just the name of the executable
+    cmd = { cmdpath, path }
```
justinmk, line 156: if cmdpath is a list then this won't work. rather than supporting `string|string[]`, let's require `string[]`.

```suggestion
    cmd = vim.list_extend(cmdpath, { path })
```

justinmk, line 156: if cmdpath is a list then this won't work. rather than supporting `string|string[]`, let's require `string[]`.

```suggestion
    cmd = vim.list_extend(cmdpath, { path })
```

runtime/lua/vim/ui.lua
```diff
@@ -139,12 +144,17 @@ function M.open(path)
     path = vim.fs.normalize(path)
   end
 
-  local cmd --- @type string[]
+  local cmd --- @type string[]|string[][]
```
justinmk, line 147:
```suggestion
  local cmd --- @type string[]
```

runtime/doc/news.txt
```diff
@@ -188,7 +188,10 @@ TUI
...
+• Behaviour of the function |vim.ui.open()| (by default bound to |gx| key
+  shortcut) can now be configured by setting the its second optional parameter
+  `cmdpath` to the name of the command which should be run with the selected
+  string as the only parameter.
```
justinmk, line 194:
```suggestion
• |vim.ui.open()| (by default bound to |gx|) accepts an optional `opts.cmd`
  parameter which will be tried before falling back to other default tools.
```

runtime/lua/vim/ui.lua
mcepl: I don’t care about harsh, we are engineers and what matters is the result, if your suggestions make the result better, then they should be accepted. And they do.

(And yes, I live too close to Germany, and in these parts, we don’t pussyfoot about something being wrong, we just say it).

runtime/lua/vim/ui.lua
```diff
@@ -125,12 +127,16 @@ end
 --- ```
 ---
 ---@param path string Path or URL to open
+---@param opts? table Additional options
+---     - cmd (string[]|string)
+---           Alternative URL handler: if set, editor will use this command
+---           to handle 'path' as its only command argument.
```
justinmk, line 133: single quotes in vim help files link to vim options
```suggestion
---           to handle `path` as its last command argument.
```

runtime/lua/vim/ui.lua
```diff
@@ -140,11 +146,15 @@ function M.open(path)
...
+  local conf --- @type vim.SystemOpts
 
-  opts = { text = true, detach = true }
+  conf = opts or {}
+  conf.text = opts.text or true
+  conf.detach = opts.detach or true
 
-  if vim.fn.has('mac') == 1 then
+  if conf.cmd then
+    cmd = vim.list_extend(conf.cmd, { path })
```
justinmk, line 156: looks like `opts` and `conf` are being mixed up. they aren't related. this should just be `opts.cmd`.

runtime/lua/vim/ui.lua
```diff
@@ -140,11 +146,15 @@ function M.open(path)
   end
 
   local cmd --- @type string[]
-  local opts --- @type vim.SystemOpts
+  local conf --- @type vim.SystemOpts
 
-  opts = { text = true, detach = true }
+  conf = opts or {}
+  conf.text = opts.text or true
+  conf.detach = opts.detach or true
```
justinmk, line 153: looks like `opts` and `conf` are being mixed up. they aren't related.

runtime/lua/vim/ui.lua
mcepl: see above

runtime/lua/vim/ui.lua
justinmk: i guess `opts` could be a way to pass these options, but it wasn't really something we discussed, and it's not documented... should wait until a different PR so we can wrap this one up

mcepl: Ehm, can I ask for your suggestion here? I really don’t understand what would you like to be done.

justinmk: What is named `conf` here is unrelated to the `opts` passed to `vim.ui.open`. If we want to also allow callers to configure the `vim.system()` options, that would be a separate `opts` field such as `opts.cmd_opts` or something like that.

For now I removed that.

justinmk (approved):
image

PR #29491, opened 2024-06-26
fix(lsp): pre-filter matches on label if filterText is missing

### LINKED ISSUE #29287: Built-in LSP completion not filtering on label
### Problem

According to LSP specification, completion items should be filtered by `filterText` property, and if it does not exist, it should filter by `label`. Quoting from https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem:

```
	/**
	 * A string that should be used when filtering a set of
	 * completion items. When omitted the label is used as the
	 * filter text for this item.
	 */
	filterText?: string;
```

New built-in LSP completion filters only by `filterText` if it is defined, but it does not apply filtering on the `label` as a fallback.

https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/completion.lua#L241-L243

### Steps to reproduce

1. Install `vtsls` language server (this one was easiest to reproduce)
```
npm install -g @vtsls/language-server
```
2. Use this minimal init:
```lua
  pattern = 'javascript',
  callback = function()
    vim.lsp.start({
      cmd = { 'vtsls', '--stdio' },
      on_attach = function(client, bufnr)
        vim.lsp.completion.enable(true, client.id, bufnr, { autotrigger = true })
      end,
    })
  end,
})

vim.keymap.set('i', '<C-space>', function()
 vim.lsp.completion.trigger()
end)
```
3. `nvim -u minimal_init.lua`
4. `:edit test.js`
5. Type `Math.mi`, this should show correct completion since it's doing "live" filtering
6. Press `<C-space>`, it will show a lot of unmatched items
screenshot_2024_06_11_18_14_41

### Expected behavior

Expected to properly filter by the prefix, which is in my example `mi`.

### Neovim version (nvim -v)

NVIM v0.11.0-dev-37bf4c5

N/A

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.35.1

xterm-kitty

### Installation

bob (https://github.com/MordechaiHadad/bob

mfussenegger:
Although the built-in pum completion mechanism will filter anyway on the
next input it is odd if the initial popup shows entries which don't
match the current prefix.

Using fuzzy match on the label/prefix is compatible with
`completeopt+=fuzzy` and also doesn't seem to break postfix snippet
cases

Closes https://github.com/neovim/neovim/issues/29287

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -235,14 +235,20 @@ function M._lsp_to_complete_items(result, prefix, client_id)
     return {}
   end
 
-  local matches = prefix == '' and function()
-    return true
-  end or function(item)
-    if item.filterText then
-      return next(vim.fn.matchfuzzy({ item.filterText }, prefix))
+  ---@type fun(item: lsp.CompletionItem):boolean
+  local matches
+  if prefix == '' then
+    matches = function(_)
+      return true
     end
-    return true
+  else
+    ---@param item lsp.CompletionItem
+    matches = function(item)
+      local text = item.filterText or item.label
+      return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
+    end
   end
+
   local candidates = {}
   for _, item in ipairs(items) do
     if matches(item) then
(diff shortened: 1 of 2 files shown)


MariaSolOs (approved):
Should we add a test for this case?

mfussenegger:
> Should we add a test for this case?

Good call. Added it

PR #29497, opened 2024-06-27
docs(lsp): add examples to lsp-quickstart for completion and autoformatting

gpanders:
Auto-completion and auto-formatting are common (though certainly not universal) features that many users want. We can document how to accomplish this in lsp-quickstart so that users that do want these features can easily find examples of how to configure them.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -34,16 +34,16 @@ Follow these steps to get LSP features:
     vim.api.nvim_create_autocmd('FileType', {
       -- This handler will fire when the buffer's 'filetype' is "python"
       pattern = 'python',
-      callback = function(ev)
+      callback = function(args)
         vim.lsp.start({
           name = 'my-server-name',
           cmd = {'name-of-language-server-executable', '--option', 'arg1', 'arg2'},
 
           -- Set the "root directory" to the parent directory of the file in the
-          -- current buffer (`ev.buf`) that contains either a "setup.py" or a
+          -- current buffer (`args.buf`) that contains either a "setup.py" or a
           -- "pyproject.toml" file. Files that share a root directory will reuse
           -- the connection to the same LSP server.
-          root_dir = vim.fs.root(ev.buf, {'setup.py', 'pyproject.toml'}),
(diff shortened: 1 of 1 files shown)


MariaSolOs (approved):

mfussenegger (approved):
👍 

Should we also mention the snippet keymaps somehow?

MariaSolOs:
> 👍
> 
> Should we also mention the snippet keymaps somehow?

Well those are set by default when a snippet session is active, so there's no need to do anything about them here.

mfussenegger:
> > 👍
> > Should we also mention the snippet keymaps somehow?
> 
> Well those are set by default when a snippet session is active, so there's no need to do anything about them here.

But how are users supposed to know that that happens and which keys they can use?
Or that snippet expansion is even a thing

MariaSolOs:
> But how are users supposed to know that that happens and which keys they can use? Or that snippet expansion is even a thing

Oh I see what you mean. I had misunderstood your original suggestion as mentioning "how to setup snippet keymaps".

mfussenegger:
> > But how are users supposed to know that that happens and which keys they can use? Or that snippet expansion is even a thing
> 
> Oh I see what you mean. I had misunderstood your original suggestion as mentioning "how to setup snippet keymaps".

Sorry, I could've explained my point better.
I also just saw that the `vim.lsp.completion.enable` docs doesn't mention that aspect at all. It's not just the autotrigger, but opt-in to completion side effect handling - including snippet expansion. We should mention that somehow

mfussenegger:
Merging this as it's already an improvement. We can follow up on the snippet-expansion and side effect aspects later

runtime/doc/lsp.txt
```diff
@@ -34,16 +34,16 @@ Follow these steps to get LSP features:
     vim.api.nvim_create_autocmd('FileType', {
       -- This handler will fire when the buffer's 'filetype' is "python"
       pattern = 'python',
-      callback = function(ev)
+      callback = function(args)
```
justinmk, line 37: `ev` is an intentional convention to teach users that these args are event data.

PR #29501, opened 2024-06-27
fix(lsp): don't send foreign diagnostics to servers in buf.code_action

### LINKED ISSUE #29500: Nvim lsp code action send doesn't respect encoding
### Problem

Here we translate vim diagnostic to lsp:
https://github.com/neovim/neovim/blob/724d1110b1e4699a34f489e9cdb2d25098746499/runtime/lua/vim/lsp/diagnostic.lua#L164
But it's doing it incorrectly. It doesn't respect LSP offset_encoding and makes wrong encoding. So, nvim sends wrong position in such case like:

```go
func привет() { // if such code will have diagnostic pointing to 'привет', nvim will send to lsp position as end=17 instead of end=11
}
```

I believe it's a bug cause at other places offset_encoding is handling correctly, like here:

https://github.com/neovim/neovim/blob/724d1110b1e4699a34f489e9cdb2d25098746499/runtime/lua/vim/lsp/buf.lua#L902

### Neovim version (nvim -v)

NVIM v0.10.0

mfussenegger:
`buf.code_action` always included diagnostics on a given line from all
clients. Servers should only receive diagnostics they published, and in
the exact same format they sent it.

Should fix https://github.com/neovim/neovim/issues/29500

### DIFF
--- a/runtime/lua/vim/lsp/diagnostic.lua
+++ b/runtime/lua/vim/lsp/diagnostic.lua
@@ -122,13 +122,7 @@ local function diagnostic_lsp_to_vim(diagnostics, bufnr, client_id)
       code = diagnostic.code,
       _tags = tags_lsp_to_vim(diagnostic, client_id),
       user_data = {
-        lsp = {
-          -- usage of user_data.lsp.code is deprecated in favor of the top-level code field
-          code = diagnostic.code,
-          codeDescription = diagnostic.codeDescription,
-          relatedInformation = diagnostic.relatedInformation,
-          data = diagnostic.data,
-        },
+        lsp = diagnostic,
       },
     }
   end, diagnostics)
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -366,6 +363,7 @@ end
 ---              Structured: { [1] = {...}, [5] = {.... } }
 ---@private
 function M.get_line_diagnostics(bufnr, line_nr, opts, client_id)
+  vim.deprecate('vim.lsp.diagnostic.get_line_diagnostics', 'vim.diagnostic.get', '0.12')
```
MariaSolOs, line 366: Is it necessary to deprecate this function if it's marked as `private`?

mfussenegger: It's a bit grey area:

- It wasn't always marked as `@private` but became private in https://github.com/mfussenegger/neovim/commit/a5bbb932f9094098bd656d3f6be3c58344576709
- It's not prefixed with `_`, users who inspect modules by other means might have seen it

I also still use it myself in nvim-jdtls.
But I accelerated it to 0.12 because of it and I'm not sure if it also warrants the deprecated.txt entry.

clason:
@mfussenegger needs a rebase

PR #29502, opened 2024-06-27
build(deps): fix formatc target on MSVC

rkitover:
Set CMAKE_RUNTIME_OUTPUT_DIRECTORY for CMAKE_ARGS for the uncrustify ExternalProject to point to DEPS_BIN_DIR because the uncrustify cmake sources do not install to DESTDIR/bin under MSVC but to DESTDIR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ get_externalproject_options(uncrustify ${DEPS_IGNORE_SHA})
 ExternalProject_Add(uncrustify
   DOWNLOAD_DIR ${DEPS_DOWNLOAD_DIR}/uncrustify
   CMAKE_ARGS ${DEPS_CMAKE_ARGS}
+    -D CMAKE_RUNTIME_OUTPUT_DIRECTORY=${DEPS_BIN_DIR}
   EXCLUDE_FROM_ALL TRUE
   ${EXTERNALPROJECT_OPTIONS})
 


dundargoc:
This isn't the correct solution because you'll need to do this for all programs forever

rkitover:
Should I fix how the `uncrustify` program is installed into the deps folder then?

rkitover:
@dundargoc do you think this solution is more correct, this is probably the best I can do here.

dundargoc:
I'll take a closer look tomorrow

rkitover:
What do you think about adding `stylua` to deps for the `formatlua` target? I won't do that here though.

dundargoc:
Better workaround IMO:

```diff
---
 CMakeLists.txt | 1 +
 1 file changed, 1 insertion(+)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1b070fef2..90f00131b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ get_externalproject_options(uncrustify ${DEPS_IGNORE_SHA})
 ExternalProject_Add(uncrustify
   DOWNLOAD_DIR ${DEPS_DOWNLOAD_DIR}/uncrustify
   CMAKE_ARGS ${DEPS_CMAKE_ARGS}
+  -D CMAKE_RUNTIME_OUTPUT_DIRECTORY=${DEPS_BIN_DIR}
   EXCLUDE_FROM_ALL TRUE
   ${EXTERNALPROJECT_OPTIONS})
 
---
```

I guess the real solution is to send a PR to uncrustify to fix their build system but I don't have the energy for that lol.

dundargoc:
> What do you think about adding `stylua` to deps for the `formatlua` target? I won't do that here though.

Uh, I don't understand what you mean. We already use stylua when we run `formatlua`. Do you mean that it should automatically install stylua like with uncrustify?

rkitover:
Yeah that's what I meant, it would be more difficult though because it uses cargo not cmake.

dundargoc:
Eh, couldn't hurt I suppose although I don't see too much benefit from it either. As long as we don't think about downloading cargo if it's missing because that will lead to pain.

rkitover:
> I guess the real solution is to send a PR to uncrustify to fix their build system but I don't have the energy for that lol.

I can do that, but how would you suggest handling that? Their install commands are correct for Windows and GNUInstallDirs are not. Perhaps adding some variables to make the install targets more configurable?

rkitover:
Ok I've applied your commit and tested it on MSVC and Linux.

dundargoc (approved):

rkitover:
> Eh, couldn't hurt I suppose although I don't see too much benefit from it either. As long as we don't think about downloading cargo if it's missing because that will lead to pain.

So apparently scoop has stylua, and it works perfectly fine with the formatlua target, so I think that's good enough.

rkitover:
@dundargoc so I opened this:

https://github.com/uncrustify/uncrustify/pull/4348

rkitover:
@zeertzjq can you merge please, we have settled on a solution here.

PR #29519, opened 2024-06-30
feat(win32): embed executable icon

### LINKED ISSUE #27466: Include the Neovim icon in `nvim.exe` just like `nvim-qt.exe`
### Problem

I prefer to associate various file types with `nvim.exe` for direct opening in the Windows Terminal via double clicks. However, unlike `nvim-qt.exe`, nvim.exe lacks its own icon, so every file gets the default executable icon, which is confusing.

### Expected behavior

Associated files should get the Neovim icon.

alex-tdrn:
Problem: on windows, the neovim executable (and thus the filetypes associated to open with neovim) has no embedded icon

Solution: create a windows resource file pointing to the icon, and add it to the nvim binary target

fixes #27466

before:
image

after:
image

### DIFF
--- a/runtime/windows_icon.rc
+++ b/runtime/windows_icon.rc
@@ -1,0 +1,4 @@
+// NOTE: this resource file *must* be in the same folder as the icon.
+// Otherwise, absolute paths would need to be used.
+// see https://learn.microsoft.com/en-us/windows/win32/menurc/icon-resource
+NEOVIM_ICON ICON "neovim.ico"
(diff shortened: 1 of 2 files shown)


alex-tdrn:
I marked this as draft because of the news.txt warning, but maybe a news entry is not necessary? If yes, under what section would I put the news entry?

dundargoc (approved):
Works without problems for me. Well done.

ni554n:
Sorry, I'm not familiar with Neovim's release schedule, but this feature seems to be complete. Why has it been in the backlog for so long?

dundargoc:
Should be available on v0.11. I suppose we could backport it since it's arguably a fix?

ni554n:
That would be awesome. I have been waiting forever for this to merge.

dundargoc:
Should be available in the next release after that PR is merged.

PR #29521, opened 2024-06-30
docs(lsp): add enum type annotation `vim.lsp.protocol.Methods`

rcasia:
Added the enum type annotation `vim.lsp.protocol.Methods` to provide some intellisense support,
improving developer experience with autocomplete suggestions and documentation.


MariaSolOs (approved):

mfussenegger:
You'd also need to run the script to update the `protocol.lua`, but I don't really see any change with lua-ls, with or without this change?

runtime/lua/vim/lsp/protocol.lua
```diff
@@ -942,5 +943,14 @@ protocol.Methods = {
   --- The `workspace/workspaceFolders` is sent from the server to the client to fetch the open workspace folders.
   workspace_workspaceFolders = 'workspace/workspaceFolders',
 }
+local function freeze(t)
```
max397574, line 946: why would you add this here?
perhaps rebased on wrong branch?

rcasia: This was autogenerated when running `nvim -l scripts/gen_lsp.lua gen --methods`.

It is hardcoded in the script: https://github.com/neovim/neovim/blob/8f5e90811025bcc652146168d5116f8d73fe8f81/scripts/gen_lsp.lua#L114-L127

It seems someone forgot to run it in the past. For what I checked, every reference to lsp.Methods complies with this changes.

max397574: 🤦‍♂️ I didn't realize that this also was part of the autogenerated things

rcasia:
@mfussenegger I initially thought this would be handled by GitHub Actions, but it seems not. I've already updated it.

MariaSolOs (approved):

rcasia:
@mfussenegger Hi, just following up on this PR. I’ve addressed all requested changes and am happy to assist further. Thanks for your time!

runtime/lua/vim/lsp/protocol.lua
justinmk: freeze was removed in https://github.com/neovim/neovim/commit/0a0962a2e8d9ffa5f03e492fa293e964650a965e , please regen and rebase.

rcasia: Just completed the regen and rebase. A different set of changes was generated due to the regen script, but they seem appropriate.

PR #29522, opened 2024-06-30
fix(lsp): default to `label` for newText

seblyng:
Problem: 

If the server doesn't provide newText, item.textEditText or item.insertText, `newText` will be `nil`.
This will result in an error on line 157 trying to index a nil value.

Solution:

From what I could read from the spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

the client should default back to using `label` if `textEditText` or `insertText` is not provided.


MariaSolOs (approved):

mfussenegger:
Could you also add a test case?

seblj:
> Could you also add a test case?

Done

PR #29531, opened 2024-07-01
perf(treesitter): detach parser when it's no longer active

zeertzjq:
Problem:
When calling `vim.treesitter.get_parser()` with different languages on the same buffer, the active parser of the buffer is changed, but the old parser is still attached to the buffer even though it's not reachable any more. This can be observed as delay when typing text after executing the following script:
```lua
    for _ = 1, 100000 do
      vim.treesitter.get_parser(0, 'lua')
      vim.treesitter.get_parser(0, 'query')
    end
```
Solution:
When a parser is longer the active one in its buffer, detach from the buffer by returning `true` in `nvim_buf_attach` callbacks.

This is extracted from #29530.


zeertzjq:
This seems incorrect, as the old parser may still be used somewhere.

PR #29536, opened 2024-07-02
fix(lua): use `rawget()` to get `__call` in `vim.is_callable()`

tmillr:
Lua 5.1 uses a ***raw get*** to retrieve `__call` from a metatable to determine if a table is callable. Mirror this behavior in `vim.is_callable()`.

Reopens #27457

### DIFF
--- a/test/functional/lua/vim_spec.lua
+++ b/test/functional/lua/vim_spec.lua
@@ -1495,6 +1495,60 @@ describe('lua stdlib', function()
     ]])
     )
 
+    eq(
+      { false, false },
+      exec_lua([[
+      local meta = { __call = {} }
+      assert(meta.__call)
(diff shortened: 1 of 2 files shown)


test/functional/lua/vim_spec.lua
```diff
@@ -1495,6 +1495,60 @@ describe('lua stdlib', function()
...
+      end
+      local not_callable = new()
+      return { pcall(function() not_callable() end), vim.is_callable(not_callable) }
+    ]])
+    )
+    eq(
+      { false, false },
+      exec_lua([[
+      local function new()
+        return { __call = function()end }
```
zeertzjq, line 1514:
```suggestion
        return { __call = function() end }
```

PR #29548, opened 2024-07-03
build(macos): disable deduplication in link step

clason:
Problem: Apple Clang 16 comes with a new deduplication algorithm that is
enabled by default in release builds (https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes#New-Features-in-Xcode-16-Beta) which breaks the built
nvim binary (locks and spins at 100% CPU).

Solution: Disable deduplication on macOS. Tested on Clang 15 as well and
seems to make no difference to binary size anyway.

### DIFF
--- a/src/nvim/CMakeLists.txt
+++ b/src/nvim/CMakeLists.txt
@@ -160,6 +160,9 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # ENABLE_EXPORTS is set to true. See
   # https://github.com/neovim/neovim/issues/25295
   target_link_options(nvim_bin PRIVATE "-Wl,-export_dynamic")
+  # Apple Clang 16's new deduplication pass leads to lock-up on start
+  # TODO(clason): verify that problem remains after release or remove
+  target_link_options(nvim_bin PRIVATE "-Wl,-no_deduplicate")
 elseif(CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
   target_link_libraries(main_lib INTERFACE pthread c++abi)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "SunOS")


clason:
I'm not 100% sure we should merge this, as I'd like to understand _why_ the new algorithm breaks Nvim. It's possible that this is a Clang bug (which persists in Beta 2) that will get fixed before release, but I thought it'd be good to document the issue and workaround in the meantime for others that test the macOS betas.

dundargoc:
Somewhat conflicted since it might get solved on non-beta like you mentioned. Maybe make it more explicit in the comment that it's mac's own beta-clang 16 and add a TODO to recheck again later when clang 16 comes out?

dundargoc (approved):

clason:
Not in a big hurry, although it's unclear that this actually is a bug and not something to do with our `nlua0` bootstrapping. As I said, my primary motivation is documentation; I'm fine with parking this for a few more prereleases.

dundargoc (approved):

src/nvim/CMakeLists.txt
```diff
@@ -160,6 +160,8 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # ENABLE_EXPORTS is set to true. See
   # https://github.com/neovim/neovim/issues/25295
   target_link_options(nvim_bin PRIVATE "-Wl,-export_dynamic")
+  # Apple Clang 16's new deduplication pass leads to lock-up on start (as of beta 2)
```
dundargoc, line 163: Just so we know when we can possible remove this?

```suggestion
  # Apple Clang 16's new deduplication pass leads to lock-up on start (as of beta 2)
  # TODO(clason): test if this is still needed once Apple Clang 16 is not on beta.
```

dundargoc:
IMO we can merge this for now, I don't think we need to wait. The updated comment is good enough to give context.

dundargoc (approved):

clason:
Still happens in beta3, so probably here to stay. According to `lldb`, it's not actually hanging but spending all its time in LuaJIT, e.g., in 
```
   256  int32_t lj_lib_optint(lua_State *L, int narg, int32_t def)
   257  {
-> 258    TValue *o = L->base + narg-1;
   259    return (o < L->top && !tvisnil(o)) ? lj_lib_checkint(L, narg) : def;
   260  }
 ```

Since I understand neither C nor LuaJIT nor lldb, I'll have to leave it at that.

clason:
Backporting since 0.11 probably won't be out before Xcode 16.

PR #29555, opened 2024-07-04
fix(diagnostic): make docs agree with code

telemachus:
Problem: the code and docs for vim.diagnostic.JumpOpts.float send mixed signals about what the default should be. When the option is first set, in the `global_diagnostic_options` table, the comment clearly says that the default is `false`. Later in the code, in `goto_diagnostic`, there's a line that sets the default to `true` if no default is present. Finally, the docs say that the default is `true`.

Solution: Change the docs to reflect the new default of `false` and fix the `goto_diagnostic` function.


clason:
Your commit does the complete opposite of what your PR description says?

telemachus:
@clason I apologize. I managed to step on my own feet with `git force --push`. I'll close this and try again.

Before I do though, I thought the change here was right. That is, I want to update the docs, but I also think I want to remove the call to `if_nil`. If the new default is `false`, this line `local float = if_nil(opts.float, true)` is unnecessary (and backwards). Am I confused about that?

PR #29556, opened 2024-07-04
feat(runtime): add filetype plugin for terraform

JannoTjarks:
- Adds commentstring property for terraform
- Adds comments property for terraform


JannoTjarks:
I think, the news.txt can be skipped? Its just a small ftplugin, but i can at a line, if necassary :)

clason:
Thank you, but we don't maintain these files. Please wait for your Vim PR to be merged, then we will port it as a matter of course.

PR #29581, opened 2024-07-06
fix(treesitter.foldexpr): robustness against ctrl-c

### LINKED ISSUE #28516: vim.treesitter.foldexpr: attempt to index field 'foldupdate_range' (a nil value)
### Problem

Using `vim.opt.foldexpr = "v:lua.vim.treesitter.foldexpr()"` while triggering some kind of autocompletion on insert mode and then going to normal mode without closing the autocompletion menu causes the following error:

```lua
Se ha detectado un error al procesar InsertLeave Auto-órdenes para "<buffer=1>":
Error executing lua callback: ...s\Neovim\share\nvim\runtime/lua/vim/treesitter/_fold.lua:285: attempt to index field 'foldupdate_range' (a nil value)
stack traceback:
        ...s\Neovim\share\nvim\runtime/lua/vim/treesitter/_fold.lua:285: in function 'do_foldupdate'
        ...s\Neovim\share\nvim\runtime/lua/vim/treesitter/_fold.lua:274: in function <...s\Neovim\share\nvim\runtime/lua/vim/treesitter/_fold.lua:273>
```

### Steps to reproduce

1. `nvim -u minimal.lua minimal.lua`
2. `G` go to last line, `o` insert new line under current line and enter insert mode
3. `local function something() end`
4. Without closing the `pummenu`, `<esc>` go to normal mode
5. The error is shown

minimal.lua:
```lua
function ExampleOmnifunc(find_start, base)
	if find_start == 1 then
		return #vim.api.nvim_get_current_line()
	elseif find_start == 0 then
		return { "end" }
	end
end

vim.opt.omnifunc = "v:lua.ExampleOmnifunc"

vim.opt.foldenable = true
vim.opt.completeopt = { "menuone", "noselect", "noinsert" }
vim.opt.foldcolumn = "1"
vim.o.foldlevelstart = 99
vim.o.foldmethod = "expr"
vim.opt.foldexpr = "v:lua.vim.treesitter.foldexpr()"

local triggers = { "d" }
vim.api.nvim_create_autocmd("InsertCharPre", {
	buffer = vim.api.nvim_get_current_buf(),
	callback = function()
		if vim.fn.pumvisible() == 1 or vim.fn.state("m") == "m" then
			return
		end
		local char = vim.v.char
		if vim.list_contains(triggers, char) then
			local key = vim.keycode("<C-x><C-o>")
			vim.api.nvim_feedkeys(key, "m", false)
		end
	end,
})
```

### Expected behavior

There should be no errors and folds should work as expected

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2999+g38b9c322c

Doesn't apply

### Operating system/version

Windows 11 Pro 23H2

### Terminal name/version

Doesn't apply

Doesn't apply

### Installation

nvim-win64.msi from the github releases page

tomtomjhj:
Problem:
Exiting the insert mode with ctrl-c does not trigger InsertLeave autocmd. This may lead to nil error in treesitter foldexpr.

Solution:
Check nil. Folds still can be stale after exiting the insert mode with ctrl-c, but it will be eventually updated correctly.

An alternative solution would be to ensure that exiting the insert mode always triggers do_foldupdate. This can be done either by "fixing" ctrl-c or with on_key callback that checks ctrl-c (nvim-cmp does this).

Fixes #28516

### DIFF
--- a/runtime/lua/vim/treesitter/_fold.lua
+++ b/runtime/lua/vim/treesitter/_fold.lua
@@ -264,6 +264,15 @@ end
 
 ---@package
 function FoldInfo:do_foldupdate(bufnr)
+  -- InsertLeave is not executed when <C-C> is used for exiting the insert mode, leaving
+  -- do_foldupdate untouched. If another execution of foldupdate consumes foldupdate_range, the
+  -- InsertLeave do_foldupdate gets nil foldupdate_range. In that case, skip the update. This is
+  -- correct because the update that consumed the range must have incorporated the range that
+  -- InsertLeave meant to update.
+  if not self.foldupdate_range then
+    return
+  end
+
   local srow, erow = self.foldupdate_range[1], self.foldupdate_range[2]
   self.foldupdate_range = nil
   for _, win in ipairs(vim.fn.win_findbuf(bufnr)) do


justinmk:
> either by "fixing" ctrl-c or with on_key callback that checks ctrl-c

ctrl-c is a feature that allows the user to intentionally skip extra processing. We shouldn't try to work around that.

Users need to understand that ctrl-c from insert mode skips various features. That's intentional.

PR #29583, opened 2024-07-06
fix(lua): change some vim.fn.expand() to vim.fs.normalize()

### LINKED ISSUE #29579: vim.ui.open can't handle all file names
### Problem

The Lua function `vim.ui.open()` can't open all files. Some filenames with `[]` characters don't work. Here's what happens when I open `foo[2024-01-01].txt`:
```
E5108: Error executing lua Vim:E944: Reverse range in character class
stack traceback:
        [C]: in function 'expand'
        /usr/share/nvim/runtime/lua/vim/ui.lua:139: in function 'open'
```
The documentation at `:help vim.ui.open()` seems to imply that any filename works.

### Steps to reproduce

```
touch "foo[2024-01-01].txt"
nvim --clean
:lua vim.ui.open("foo[2024-01-01].txt")
```

### Expected behavior

the file is opened, with whatever xdg-open decides is the right program.

### Neovim version (nvim -v)

NVIM v0.10.0

no

### Operating system/version

Arch Linux x86_64

### Terminal name/version

qterminal 2.0.1 + tmux 3.4

tmux-256color

### Installation

Arch Repository neovim-0.10.0-5

zeertzjq:
Unlike `vim.fn.expand()`, `vim.fs.normalize()` doesn't expand wildcards.

Fix #29579

### DIFF
--- a/runtime/lua/vim/health/health.lua
+++ b/runtime/lua/vim/health/health.lua
@@ -50,11 +50,11 @@ local function check_config()
 
   local init_lua = vim.fn.stdpath('config') .. '/init.lua'
   local init_vim = vim.fn.stdpath('config') .. '/init.vim'
-  local vimrc = vim.env.MYVIMRC and vim.fn.expand(vim.env.MYVIMRC) or init_lua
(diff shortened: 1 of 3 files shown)


zeertzjq:
Maybe this will cause problems on Windows?

PR #29593, opened 2024-07-06
fix(diagnostic): fix backwards compatibility for goto_next and goto_prev

### LINKED ISSUE #29591: [backwards-incompatibility] breaking change to `vim.diagnostic.goto_next`
### Problem

8ba73f0e4cc6 deprecated `vim.diagnostic.goto_next` in favor of `vim.diagnostic.jump`, and reimplements `goto_next` in terms of `jump`.  This is fine.  However, when reimplementing `goto_next`, we forgot that the default behavior of `goto_next` is to show a floating window with the error (the `float` option).

We should update `goto_next`'s implementation to call `jump` with `float=true`.

1. use neovim 0.10
2. open file with LSP errors
3. run `:lua vim.diagnostic.goto_next()`
4. note that you visit the next diag and a float opens with the error
5. switch to neovim> 8ba73f0e4cc6
6. repeat steps 2-3

### Expected behavior

Expected: Should open float with the error.

Actual: Doesn't open float.

I suggest two changes:
1. We should update `goto_next`'s implementation to call `jump` with `float=true`.
2. We should update the deprecated message to say "Use |vim.diagnostic.jump()| with `{count = 1, float=true}` instead."

### Neovim version (nvim -v)

v0.11.0-3a8265

N/A

### Operating system/version

N/A

### Log file

_No response_

vegerot:
Summary:

8ba73f0e4cc6 deprecated `vim.diagnostic.goto_next` in favor of
`vim.diagnostic.jump`, and reimplements `goto_next` in terms of `jump`.  This
is fine.  However, when reimplementing `goto_next`, we forgot that the default
behavior of `goto_next` is to show a floating window with the error (the
`float` option).

This commit adds the `float` option to the `opts` table in `goto_next` and
`goto_prev`, and sets it to `true` by default.  This is the behavior that
`goto_next` and `goto_prev` had before the deprecation.  This commit also
updates the documentation to reflect this change.

Test Plan:

1. use neovim 0.10
2. run `:lua vim.diagnostic.goto_next()`
3. verify that the error is shown in a floating window
4. run `:lua vim.diagnostic.goto_next({float=false})`
5. verify that the error is not shown in a floating window
----
6. use previous commit to this commit
7. run `:lua vim.diagnostic.goto_next()`
8. verify that the error is NOT shown in a floating window
----
9. use this commit
10. run `:lua vim.diagnostic.goto_next()`
11. verify that the error is shown in a floating window

Fixes #29591

### DIFF
--- a/runtime/doc/deprecated.txt
+++ b/runtime/doc/deprecated.txt
@@ -23,8 +23,8 @@ LUA
 - vim.region()		Use |getregionpos()| instead.
 
 DIAGNOSTICS
-- *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count = 1}` instead.
-- *vim.diagnostic.goto_prev()*	Use |vim.diagnostic.jump()| with `{count = -1}` instead.
+- *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count = 1, float=true}` instead.
+- *vim.diagnostic.goto_prev()*	Use |vim.diagnostic.jump()| with `{count = -1, float=true}` instead.
 - *vim.diagnostic.get_next_pos()*
 	Use the "lnum" and "col" fields from the return value of
 	|vim.diagnostic.get_next()| instead.
--- a/runtime/doc/deprecated.txt
+++ b/runtime/doc/deprecated.txt
@@ -23,8 +23,8 @@ LUA
 - vim.region()		Use |getregionpos()| instead.
 
 DIAGNOSTICS
-- *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count = 1, float=true}` instead.
-- *vim.diagnostic.goto_prev()*	Use |vim.diagnostic.jump()| with `{count = -1, float=true}` instead.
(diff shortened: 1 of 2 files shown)


runtime/doc/deprecated.txt
```diff
@@ -23,8 +23,8 @@ LUA
 - vim.region()		Use |getregionpos()| instead.
 
 DIAGNOSTICS
-- *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count = 1}` instead.
-- *vim.diagnostic.goto_prev()*	Use |vim.diagnostic.jump()| with `{count = -1}` instead.
+- *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count = 1, float=true}` instead.
+- *vim.diagnostic.goto_prev()*	Use |vim.diagnostic.jump()| with `{count = -1, float=true}` instead.
```
gpanders, line 27: Nit, use spaces around `=` in both or in neither, but don't mix (since we're space constrained, probably omit spaces):

```suggestion
- *vim.diagnostic.goto_next()*	Use |vim.diagnostic.jump()| with `{count=1, float=true}` instead.
- *vim.diagnostic.goto_prev()*	Use |vim.diagnostic.jump()| with `{count=-1, float=true}` instead.
```

PR #29615, opened 2024-07-08
fix: check buffer valid when inlay_hint loading

Parsifa1:
fix for #29614

### DIFF
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -70,6 +70,9 @@ function M.on_inlayhint(err, result, ctx, _)
     return
   end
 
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -70,6 +70,9 @@ function M.on_inlayhint(err, result, ctx, _)
     return
   end
 
+  if not vim.api.nvim_buf_is_valid(bufnr) then
```
MariaSolOs, line 73: I feel like this is not the right place to change. If the buffer is still being tracked in the internal state table then it must have missed some unloading/deletion buffer signal (and that's what we should change instead).

Parsifa1: Is there a better way to solve this problem? :(

PR #29624, opened 2024-07-09
build(deps): drop unused bundled bash, python parsers and queries

clason:
Problem: Neovim bundles treesitter parsers for bash and python but does not use them by default. This dilutes the messaging about the bundled parsers being required for functionality or reasonable out-of-the-box experience. It also increases the risk of query incompatibilities for no
gain.

Solution: Stop bundling bash and python parser and queries.

@jamessan @lewis6991


zeertzjq:
Update news-0.10?

dundargoc (approved):

clason:
Turns out the python parser _is_ used -- in a foldexpr test 😞 @ribru17

ribru17:
Oops, thanks for changing that. New test looks good

PR #29627, opened 2024-07-09
build(macos): disable deduplication in link step

clason:
Problem: Apple Clang 16 comes with a new deduplication algorithm that is
enabled by default in release builds (https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes#New-Features-in-Xcode-16-Beta) which breaks the built
nvim binary (locks and spins at 100% CPU).

Solution: Disable deduplication on macOS. Tested on Clang 15 as well and
seems to make no difference to binary size anyway.


clason:
Slowpoke bot 😠

PR #29628, opened 2024-07-09
refactor!: use utf8proc full casefolding

dundargoc:
According to `CaseFolding-15.1.0.txt`, full casefolding should be
preferred over simple casefolding as it's considered to be more correct.
Since utf8proc already provides full casefolding it makes sense to
switch to it. This will also remove a lot of unnecessary build code.

Temporary exceptions are made for two sets characters:

- `ß` will still be considered `ß` (instead of `ss`) as using a full
  casefolding requires interfering with upstream spell files in some
  form.
- `İ` will still be considered `İ` (instead of `i̇`) as using full
  casefolding requires making a value judgement on the "correct"
  behavior. There are two, equally valid case-insensetive comparison for
  this character according to unicode. It is essentially up to the
  implementor to decide which conversion is correct. For this reason it
  might make sense to allow users to decide which conversion should be
  done as an added option to `casemap` in a future PR.


clason:
@dundargoc rebase, and can we give this a test-drive on `master`? The PR description sounds like an unambiguous win to me.

zeertzjq:
It seems that the problem mentioned here has been reported before in #12456

dundargoc:
> @dundargoc rebase, and can we give this a test-drive on `master`? The PR description sounds like an unambiguous win to me.
> 
> I wouldn't mind adapting the tests instead of adding a hack around them.

We could, but would need to adjust the commit/PR description since I currently just skip doing full casefolding for sharp s and turkish i because the tests fail. I'm pretty sure there's a bug in the regexp code but I haven't had the time to figure out exactly what's happening. We have two options I believe:

- Adjust commit/PR message and merge. This would entail full casefolding for all characters except for the two I mentioned with the intention of fixing these later on.

- Adapt/yeet the tests that cause a problem. Note that I'm fairly sure there is an actual bug since the regex engines behaves differently.

clason:
When rebasing, I ran the tests myself and there's a larger number of failures than I expected. So I am fine with merging with the workaround (niggle: TODO instead of HACK).

But maybe we can at least drop the ß hack? (The Turkish i is tested explicitly so needs more work.)

bfredl:
> Adjust commit/PR message and merge. This would entail full casefolding for all characters except for the two I mentioned with the intention of fixing these later on.

This sounds preferable to me. Getting rid of `src/unicode/` files for things that utf8proc can do is valuable by in itself, then we can work with edge/compat cases affecting tests in follow up PR:s.

clason (approved):

PR #29641, opened 2024-07-10
feat: let getmousepos accepts row, col params

haolian9:
hi folks, i have the need to resolve a point(screenrow,screencol) to window id, col and row (which share the same meaning with col/line(), not winline/wincol()). and i found the getmousepos() offers almost identical functionality, except it accepts no screen (or mouse) row/col.
so this pr tries to add these two parameters.

the reason i do not use getwininfo to compute it by myself is that i found it's rather difficult to deal with a line contains `<tab>` and utf-8 characters.

currently it's just my very early attempt: the code is not mature enough, no tests. yet it does work in my plugin. 

what do you think?

### DIFF
--- a/src/nvim/eval.lua
+++ b/src/nvim/eval.lua
@@ -4107,6 +4107,8 @@ M.funcs = {
     signature = 'getmatches([{win}])',
   },
   getmousepos = {
+    args = {0, 2},
+    base = 0,
     desc = [=[
       Returns a |Dictionary| with the last known position of the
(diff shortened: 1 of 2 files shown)


zeertzjq:
This seems like an XY problem. If you want to highlight a specific position on screen, highlight a character in buffer won't always work, because there are end-of-line and end-of-buffer positions.

haolian9:
yeah, i know that. and matchaddpos() just silently ignores them, i belive. but since i have tried different approaches, this is the best one to me.

PR #29645, opened 2024-07-10
fix: use utf8proc to detect emojis

dundargoc:
More accurately, it will check if the codepoint has the
"Extended_Pictographic" property, which according to
https://www.unicode.org/reports/tr51/#Emoji_Properties_and_Data_Files is
described as:

"The Extended_Pictographic characters contain all the Emoji characters
except for some Emoji_Component characters." This should in most cases
align with what people refer to when they say "emoji".

### DIFF
--- a/src/nvim/generators/gen_unicode_tables.lua
+++ b/src/nvim/generators/gen_unicode_tables.lua
@@ -12,9 +12,9 @@
 --    2 then interval applies only to first, third, fifth, … character in range.
 --    Fourth value is number that should be added to the codepoint to yield
 --    folded codepoint.
--- 4. emoji_wide and emoji_all tables: sorted lists of non-overlapping closed
---    intervals of Emoji characters.  emoji_wide contains all the characters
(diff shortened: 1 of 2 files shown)


clason:
How is this change breaking precisely?

gpanders:
> I think this new solution should be superior but it wouldn't hurt if some emoji experts/connoisseurs tested this PR.

I am neither an expert nor a connoisseur, but the test that I use is the "heart test": write a single ❤️ to a file and make sure it displays properly with cursor movements and `:set cursorline`.

I am happy to report that with this PR the heart test passes!

clason:
Could also count as "fix" (not breaking) if we declare utf8proc as the source of truth for this (which arguably we should)?

zeertzjq:
It's strange to replace `emoji_all` without replacing `emoji_wide`.

dundargoc:
So it turns out it's not possible to determine from codepoint alone if a character is an emoji, as there are codepoints that can be both text and emoji depending on variant selector(?). Closing for the time being until we have a better solution in mind.

PR #29659, opened 2024-07-11
fix(lua)!: do not use typed table for empty dict

### LINKED ISSUE #29218: Error with built-in lsp completion
### Problem

I'm testing new built-in lsp completion, and I ran into some issues with `vtsls` language server. I thought issue is on their side and I opened up an issue (https://github.com/yioneko/vtsls/issues/164) but it looks like that the issue is on the Neovim side. 
It seems that `vim.empty_dict()` somehow gets converted to an object that is not JSON serializable

### Steps to reproduce

1. Create this minimal.lua file:
```lua
local function trigger()
  local item = {
    word = "a",
    abbr = "a",
    kind = "v",
    user_data = {
      some_field = vim.empty_dict(),
    },
  }
  vim.fn.complete(1, { item })
end

vim.opt.completeopt = 'menu,menuone'
vim.keymap.set("i", "<C-a>", trigger, { nowait = true, noremap = true, silent = false })

vim.api.nvim_create_autocmd("CompleteDone", {
  callback = function()
    local reason = vim.api.nvim_get_vvar("event").reason --- @type string
    if reason == "accept" then
      local completed_item = vim.api.nvim_get_vvar("completed_item")
      print(vim.inspect(completed_item))
      print(vim.json.encode(completed_item))
    end
  end,
})
```
2. Start nvim with minimal init: `nvim --clean -u minimal.lua`
3. Enter insert mode, press `<C-A>`, and then accept the completion with `<C-y>`
4. See the error
```
Error detected while processing CompleteDone Autocommands for "*":                                                                                                                                                                              
Error executing lua callback: /home/kristijan/neovim-config/test.lua:22: Cannot serialise boolean: table key must be a number or string                                                                                                         
stack traceback:                                                                                                                                                                                                                                
        [C]: in function 'encode'                                                                                                                                                                                                               
        /home/kristijan/neovim-config/test.lua:22: in function </home/kristijan/neovim-config/test.lua:17> 
```

### Expected behavior

Expecting to not get an error and have the object JSON serializable.

### Neovim version (nvim -v)

NVIM v0.11.0-dev-189+g78d3f4742

N/A

### Operating system/version

Arch Linux

### Terminal name/version

Kitty

xterm-kitty

### Installation

Bob (https://github.com/MordechaiHadad/bob)

amitds1997:
Fixes #29218


src/nvim/lua/converter.c
```diff
@@ -703,14 +703,10 @@ void nlua_push_Boolean(lua_State *lstate, const Boolean b, int flags)
...
-  if (dict.size == 0 && (flags & kNluaPushSpecial)) {
-    nlua_create_typed_table(lstate, 0, 0, kObjectTypeDictionary);
-  } else {
-    lua_createtable(lstate, 0, (int)dict.size);
-    if (dict.size == 0 && !(flags & kNluaPushSpecial)) {
-      nlua_pushref(lstate, nlua_global_refs->empty_dict_ref);
-      lua_setmetatable(lstate, -2);
-    }
+  lua_createtable(lstate, 0, (int)dict.size);
+  if (dict.size == 0 && !(flags & kNluaPushSpecial)) {
```
bfredl, line 707: if we do this, we should also remove the `!(flags & kNluaPushSpecial)` condition here, so that the (less intrusive) metatable tag is used instead.

clason:
And this special-cased test will have to be adapted:
https://github.com/neovim/neovim/blob/45b7a2c50335e1943a36715101e40eda5a1423f3/test/functional/lua/vim_spec.lua#L1191-L1192

amitds1997:
> And this special-cased test will have to be adapted:
> 
> https://github.com/neovim/neovim/blob/45b7a2c50335e1943a36715101e40eda5a1423f3/test/functional/lua/vim_spec.lua#L1208-L1209

Thanks for the heads-up. Fixed it in the latest [ee08b7e](https://github.com/neovim/neovim/pull/29659/commits/ee08b7ec32b0fdbe0b70f108673e500c1fd1feee). The now failing CI check seems to be flaky but I don't have permissions to re-run it.

test/functional/lua/vim_spec.lua
```diff
@@ -1205,8 +1205,13 @@ describe('lua stdlib', function()
     ]])
     eq(true, exec_lua([[return next(vim.fn.FooFunc(3)) == nil ]]))
     eq(3, eval('g:test'))
-    -- compat: nvim_call_function uses "special" value for empty dict
-    eq(true, exec_lua([[return next(vim.api.nvim_call_function("FooFunc", {5})) == true ]]))
+    eq(
+      { {}, true, true },
+      exec_lua([[
+    local ret = vim.api.nvim_call_function("FooFunc", {5})
+    return {ret, vim.tbl_isempty(ret), next(ret) == nil}
```
bfredl, line 1212: these are a bit redundant, it is enough to use `return vim.tbl_isempty(vim.api.nvim_call_function("FooFunc", {5}))` (tbl_isempty throws an error for a non-table)

bfredl:
Needs a note in `news.txt` as well, in `LUA` section in `BREAKING CHANGES` .  otherwise LGTM

dundargoc:
Annoying nitpick: since it's technically a breaking change the commit message should be "fix!(lua):".

amitds1997:
> Nitpick: since it's technically a breaking change the commit message should be "fix(lua)!:".

Done. I did look at the `news.txt` blame before squashing to see if all breaking changes are marked so, but it did not look like that 😄

test/functional/lua/vim_spec.lua
amitds1997: Done in [c6902b4](https://github.com/neovim/neovim/pull/29659/commits/c6902b4c979e7a37c8ffb8d485ba667c1a159e4a)

runtime/doc/news.txt
```diff
@@ -69,7 +69,8 @@ LSP
 
 LUA
 
-• TODO
+• Empty dict are converted to empty Lua table instead of a "special" typed
```
bfredl, line 72: "API functions consistently return an empty dictionary as |vim.empty_dict()|. Before a |lua-special-tbl| was sometimes used. "

amitds1997: Thanks, that sounds so much better than the original statement. Done in https://github.com/neovim/neovim/pull/29659/commits/970a27927eb31bdc735f0d7d5e3d07784486c6de

PR #29670, opened 2024-07-12
fix(snippet): modify base indentation when there's actually whitespace

### LINKED ISSUE #29658: vim.snippet.expand has problems with circumflex (`^`) in placeholders
### Problem

The new native snippet expansion seems to have problems with VSCode snippets if a placeholder contains a circumflex and is followed by a newline. A minimal example snippet (in json syntax) showing the problem is

```
  "bug": {
    "prefix": ["bug"],
    "body": "${1:foo^bar}\n"
  }
```

One can also test it directly via `:lua vim.snippet.expand("${1:foo^bar}\n")`. Expanding this adds an extra `^` character in the last line of the expansion. That is, the result of the expansion is

```
foo^bar
^
```

The problem does not appear if the newline at the end of the snippet is removed.

I tested the same snippet with LuaSnip which does not result in the extra `^` character. Also, I checked the VSCode documentation whether maybe `^` needs to be escaped in placeholders, which does not seem to be the case (and does not help, in any case).

### Steps to reproduce

vim --clean
:lua vim.snippet.expand("${1:foo^bar}\n")

### Expected behavior

expand to

```
foo^bar

```

### Neovim version (nvim -v)

NVIM v0.10.0

I don't think vim has a native snippet expansion

### Operating system/version

macOS 14.5

### Terminal name/version

kitty 0.35.2

xterm-kitty

### Installation

via nix-darwin (nixpkgs-unstable)

MariaSolOs:
Fixes #29658

### DIFF
--- a/test/functional/lua/snippet_spec.lua
+++ b/test/functional/lua/snippet_spec.lua
@@ -58,7 +58,13 @@ describe('vim.snippet', function()
   end)
 
   it('adds indentation based on the start of snippet lines', function()
+    local curbuf = api.nvim_get_current_buf()
+
     test_expand_success({ 'if $1 then', '  $0', 'end' }, { 'if  then', '  ', 'end' })
+
+    -- Regression test: #29658
+    api.nvim_buf_set_lines(curbuf, 0, -1, false, {})
+    test_expand_success({ '${1:foo^bar}\n' }, { 'foo^bar', '' })
   end)
 
   it('replaces tabs with spaces when expandtab is set', function()
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -514,7 +514,7 @@ function M.expand(input)
     local snippet_lines = text_to_lines(snippet_text)
     -- Get the base indentation based on the current line and the last line of the snippet.
     if #snippet_lines > 0 then
-      base_indent = base_indent .. (snippet_lines[#snippet_lines]:match('(^%s*)%S') or '') --- @type string
+      base_indent = base_indent .. (snippet_lines[#snippet_lines]:match('(^%s+)%S') or '') --- @type string
     end
 
     local shiftwidth = vim.fn.shiftwidth()


test/functional/lua/snippet_spec.lua
```diff
@@ -48,7 +48,7 @@ describe('vim.snippet', function()
     matches(err, pcall_err(exec_lua, string.format('vim.snippet.expand("%s")', snippet)))
   end
 
-  it('adds base indentation to inserted text', function()
+  it('adds correct base indentation to inserted text', function()
```
mfussenegger, line 51: Why the change here? `correct` is a bit redundant for test cases as you usually don't write a case for something to behave incorrect

clason:
Also needs a rebase now :)

test/functional/lua/snippet_spec.lua
MariaSolOs: True, I just felt like I needed to update the description of this message to better reflect the added case but this change is silly.

zsugabubus:
Sorry and please ignore me if I'm mistaken, I've just randomly came across this PR and I couldn't really understand the fix.

From the intention of the code, the change `%s*` -> `%s+` should make no difference. Previously, the empty string was `match()`ed (= `''`), now isn't but `nil or '' = ''` is just the same.

Let's try to reproduce the bug with some added space:

```vim
vim.snippet.expand('foo^  bar$1\nbaz')
```

```
foo^  bar
^  baz
```

Something is very wrong: `^` was matched in the middle of the line and it somehow ended up in `base_indent` that should be whitespace only. Beep. The only way this can happen if `^` has lost its special meaning and matched literally.

I guess the pattern is meant to be `^(%s*)%S`.

But to be honest I do not really get why `base_indent` has to be accumulated. It either has zero importance (no tests) or it's something totally unwanted, so maybe the real fix is to ditch the entire if?

```vim
vim.snippet.expand('^  foo$1\nbar$1\nbiz$1\nbaz')
```

```
^  foo
^  bar
^  ^  biz
^  ^  ^  baz
```

vs

```vim
vim.snippet.expand('^  foo\nbar\nbiz\nbaz')
```

```
^  foo
bar
biz
baz
```

mfussenegger:
Could you create a new issue for that? I noticed while working on the SnippetTextEdit support that there's something wrong with the indentation and the issue you're describing sounds similar, but I didn't have time to look into it yet

80avin:
@zsugabubus You're right, we in reality don't need this update of `base_indent`.
Check: https://github.com/neovim/neovim/pull/32970

PR #29676, opened 2024-07-12
feat: add count param to fn.getchangelist()

haolian9:
hi folks, i want to show inline extmarks for `g;` and `g,`, and i need to keep updating these extmarks while pressing `g;`/`g,`. but currently fn.getchangelist() returns the whole changelist, which could be expensive.

so this pr propuses to add a count parameter to it, which could be negative or positive that indicates the iterating direction relatived to changelistindex, and 0 for the current change only. the current change will always be included, i actually have no particular reason.

does it sound good? should i continue?

### DIFF
--- a/src/nvim/eval.lua
+++ b/src/nvim/eval.lua
@@ -3373,7 +3373,7 @@ M.funcs = {
     signature = 'getcellwidths()',
   },
   getchangelist = {
-    args = { 0, 1 },
+    args = { 0, 2 },
     base = 1,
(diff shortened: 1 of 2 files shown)


clason:
Thank you, but that function is taken from Vim -- please contribute the change there first, and we will then port it in the form it is merged there. This prevents divergence so that plugins can rely on consistent Vimscript API and makes sure both projects profit from the improvement!

PR #29683, opened 2024-07-13
vim-patch:8.2.3716: Vim9: range without a command is not compiled

zeertzjq:
#### vim-patch:8.2.3716: Vim9: range without a command is not compiled

Problem:    Vim9: range without a command is not compiled.
Solution:   Add the ISN_EXECRANGE byte code.

https://github.com/vim/vim/commit/e4eed8c6db693a9183b776032570ce2f89dcffb6

Co-authored-by: Bram Moolenaar <Bram@vim.org>


zeertzjq:
Moved to #29691

PR #29684, opened 2024-07-13
test(old): enable Test_address_line_overflow()

zeertzjq:
Nvim doesn't use atol() in getdigits() and doesn't need to check for
size of long.


zeertzjq:
Moved to #29691

PR #29685, opened 2024-07-13
build(cmake)!: standardize package naming across architectures

minhanghuang:
this PR enhances the clarity of the generated .deb and .tar.gz packages by including the architecture (x86_64 or aarch64) in their names.


zeertzjq:
Please also update documentation (e.g. BUILD.md) to reference the updated file names.

PR #29688, opened 2024-07-13
Add systemd socket activation docs for --listen

### LINKED ISSUE #22602: nvim --listen should accept sockets from systemd
### Problem

If nvim --listen accepted sockets from systemd, this would allow for headless neovim services to be started up on demand.

### Expected behavior

nvim --listen could take a special value like "systemd-socket" or nvim could learn a new flag for enabling socket activation.
This would allow neovim to be started by systemd service and socket units.

ananthb:
Closes #22602


justinmk:
Added to https://github.com/neovim/neovim/pull/30240 (with credit).

PR #29689, opened 2024-07-13
vim-patch: fix prolog file detection

clason:
- **vim-patch:9.1.0558: filetype: prolog detection can be improved**
- **vim-patch:7347642: runtime(filetype): Fix Prolog file detection regex**

(I didn't want to convert vim regex twice so I skipped the bad regex and went straight for the corrected one in the first patch.)


clason:
(waiting for the followup patch)

clason:
I think it's better to make a clean PR of the three patches @zeertzjq

PR #29690, opened 2024-07-13
fix(lsp): don't show codelens for buffers that don't support it

### LINKED ISSUE #29156: Unsupported method warning issued when calling `vim.lsp.codelens.run`
### Problem

The following warning is emitted when calling `vim.lsp.codelens.run`:

> method textDocument/codeLens is not supported by any of the servers registered for the current buffer

Neovim 0.10 changed the behavior of `vim.lsp.codelens.refresh` to refresh all buffers by default in https://github.com/neovim/neovim/pull/27253. The warning is written when an LSP client that does not support code lenses is attached to a different buffer at the time the code lens executes.

The source appears to be a call to `vim.lsp.codelens.refresh` out of the [`execute_lens` handler](https://github.com/neovim/neovim/blob/27fb62988e922c2739035f477f93cc052a4fee1e/runtime/lua/vim/lsp/codelens.lua#L53). Prior to 0.10, this would only refresh lenses for the current buffer but now it attempts to refresh lenses for all buffers.

1. Open a file and attach an LSP that does not support code lenses (e.g. dockerls)
2. Open a different file in another buffer and attach an LSP that does support code lenses
3. Execute a code lens in the second buffer
4. Observe that a warning is written

### Expected behavior

Running a code lens should not result in warnings associated with unrelated clients that are not attached to the current buffer.

### Neovim version (nvim -v)

NVIM v0.10.0

gopls v0.15.3 and dockerls v0.12.0

### Operating system/version

macOS Sonoma and Ubuntu Linux 22.04

### Log file

https://gist.github.com/sykesm/55a67323b8af9bea59d61ee7d9a0d7fe

ribru17:
Fixes #29156

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -855,17 +855,20 @@ api.nvim_create_autocmd('VimLeavePre', {
 ---@param params table|nil Parameters to send to the server
 ---@param handler? lsp.Handler See |lsp-handler|
 ---       If nil, follows resolution strategy defined in |lsp-handler-configuration|
----
+---@param on_unsupported? fun()
+---       The function to call when the buffer has no clients that support the given method.
+---       Defaults to an `ERROR` level notification.
 ---@return table<integer, integer> client_request_ids Map of client-id:request-id pairs
 ---for all successful requests.
 ---@return function _cancel_all_requests Function which can be used to
 ---cancel all the requests. You could instead
 ---iterate all clients and call their `cancel_request()` methods.
-function lsp.buf_request(bufnr, method, params, handler)
+function lsp.buf_request(bufnr, method, params, handler, on_unsupported)
   validate({
     bufnr = { bufnr, 'n', true },
     method = { method, 's' },
     handler = { handler, 'f', true },
+    on_unsupported = { on_unsupported, 'f', true },
   })
 
   bufnr = resolve_bufnr(bufnr)
@@ -887,7 +890,11 @@ function lsp.buf_request(bufnr, method, params, handler)
 
   -- if has client but no clients support the given method, notify the user
   if next(clients) and not method_supported then
-    vim.notify(lsp._unsupported_method(method), vim.log.levels.ERROR)
+    if on_unsupported == nil then
+      vim.notify(lsp._unsupported_method(method), vim.log.levels.ERROR)
+    else
+      on_unsupported()
+    end
     vim.cmd.redraw()
     return {}, function() end
   end
--- a/runtime/lua/vim/lsp/codelens.lua
+++ b/runtime/lua/vim/lsp/codelens.lua
@@ -307,7 +307,13 @@ function M.refresh(opts)
       }
       active_refreshes[buf] = true
 
-      local request_ids = vim.lsp.buf_request(buf, ms.textDocument_codeLens, params, M.on_codelens)
+      local request_ids = vim.lsp.buf_request(
+        buf,
+        ms.textDocument_codeLens,
+        params,
+        M.on_codelens,
+        function() end
+      )
       if vim.tbl_isempty(request_ids) then
         active_refreshes[buf] = nil
       end


runtime/lua/vim/lsp/codelens.lua
```diff
@@ -307,9 +307,20 @@ function M.refresh(opts)
       }
       active_refreshes[buf] = true
 
-      local request_ids = vim.lsp.buf_request(buf, ms.textDocument_codeLens, params, M.on_codelens)
-      if vim.tbl_isempty(request_ids) then
-        active_refreshes[buf] = nil
+      local supports_lens = false
+      for _, client in pairs(vim.lsp.get_clients({ bufnr = buf })) do
+        if client.server_capabilities.codeLensProvider then
```
MariaSolOs, line 312: You should use `supports_method` to check for server capabilities.

MariaSolOs:
I don't think this is the right approach. `vim.lsp.buf_request` already checks the capabilities of each server before sending the request, so this feels like duplicating that logic.

I instead suggest adding a parameter to `vim.lsp.buf_request` that's a callback to be invoked when no clients in the given buffer support the respective method. That way we can no-op (or log) in this scenario while preserving the notification as the default.

runtime/lua/vim/lsp.lua
```diff
@@ -855,17 +855,20 @@ api.nvim_create_autocmd('VimLeavePre', {
 ---@param params table|nil Parameters to send to the server
 ---@param handler? lsp.Handler See |lsp-handler|
 ---       If nil, follows resolution strategy defined in |lsp-handler-configuration|
----
+---@param unsupported_method_callback? fun(method: string)
```
MariaSolOs, line 858: Nit, but given that the method is a parameter to this function, it feels redundant to me to pass it to the callback.

MariaSolOs (approved):

runtime/lua/vim/lsp.lua
ribru17: Makes sense :+1:

runtime/lua/vim/lsp.lua
```diff
@@ -855,17 +855,20 @@ api.nvim_create_autocmd('VimLeavePre', {
...
+---@param unsupported_method_callback? fun()
+---       The function to call when the buffer has no clients that support the given method.
+---       Defaults to an `ERROR` level notification.
 ---@return table<integer, integer> client_request_ids Map of client-id:request-id pairs
 ---for all successful requests.
 ---@return function _cancel_all_requests Function which can be used to
 ---cancel all the requests. You could instead
 ---iterate all clients and call their `cancel_request()` methods.
-function lsp.buf_request(bufnr, method, params, handler)
+function lsp.buf_request(bufnr, method, params, handler, unsupported_method_callback)
```
mfussenegger, line 866:
```suggestion
function lsp.buf_request(bufnr, method, params, handler, on_unsupported)
```

runtime/lua/vim/lsp.lua
```diff
@@ -887,7 +890,11 @@ function lsp.buf_request(bufnr, method, params, handler)
 
   -- if has client but no clients support the given method, notify the user
   if next(clients) and not method_supported then
-    vim.notify(lsp._unsupported_method(method), vim.log.levels.ERROR)
+    if on_unsupported == nil then
+      vim.notify(lsp._unsupported_method(method), vim.log.levels.ERROR)
+    else
+      on_unsupported()
```
justinmk, line 896: Is it not acceptable to throw an error instead (which the caller can choose to handle or not)? Adding random callbacks to provide error handling seems clunky.

Is this another case where we're suffering from the lack of a promise/future abstraction?

mfussenegger: I also wasn't too keen on adding a new parameter but raising the error would have been a breaking change, and although handling errors with `pcall` works we afaik don't have an established pattern for how to deal with different types of errors. (E.g. if we introduce error codes like some of the uv API does, or if we recommend to parse the error message)

`buf_request` was also already removed from the docs with the intention to replace it eventually, but there wasn't much progress towards a suitable replacement. (Other than the existing `buf_request_all`)

 > Is this another case where we're suffering from the lack of a promise/future abstraction?

I'm not sure if it would help in this case.

justinmk: > `buf_request` was also already removed from the docs with the intention to replace it eventually

Ok that reduces the concern here :)

PR #29707, opened 2024-07-14
fix(lsp): inlay hints are rendered in the correct order

### LINKED ISSUE #29647: Wrong ordering of inlay hints with equal position
### Problem

LSP specifies:

> If multiple hints have the same position, they will be shown in the order they appear in the response.

Neovim currently shows inlay hints in different order.

No need for `minimal_init.lua`, the issue reproduces with `nvim --clean` by running two commands:

```
:lua vim.lsp.start { name = "r3", cmd = {"python3", "/path/to/repro.py"}, root_dir = "." }
:lua vim.lsp.inlay_hint.enable()
```

where `repro.py` is a minimal language server that sends digits 0 to 4 as four separate inlay hints at position 0: https://gist.github.com/GoldsteinE/702327e04ddc3aedd3e6af7e9a70e68e

Neovim currently shows `04321`.

The issue also reproduces with rust-analyzer, see https://github.com/rust-lang/rust-analyzer/issues/17574 for a minimal config and example.

### Expected behavior

`01234` (in that order) appears at position 0.

### Neovim version (nvim -v)

NVIM v0.11.0-dev-393+g3c5abf01b

custom, attached

### Operating system/version

NixOS

amitds1997:
Fixes #29647 

Problem:
When there are multiple inlay hints present at the same position, they should be rendered in the order they are received in the response from LSP as per the LSP spec. Currently, this is not respected.

Reason:
When each extmark call is done, the mark is stored in a MarkTree which keeps track of all marks. When these marks get applied during redraw, the order of access does not preserve the order of insertion ([source](https://github.com/amitds1997/neovim/blob/54b9646210e8bc4889ef7dd45792909630e4ec10/src/nvim/decoration.c?plain=1#L593-L600)). So, adding marks in order of insertion could be other solution too but since `MarkTree` is used for tracking other kind of marks (like signs) I was unsure if such a traversal method makes sense. If that might be a better solution, let me know.

Solution:
Gather all hints for a given position, and then set it in a single extmark call instead of multiple `nvim_buf_set_extmark` calls. This leads to fewer `nvim_buf_set_extmark` calls and correct inlay hints being rendered.

### DIFF
--- a/test/functional/plugin/lsp/inlay_hint_spec.lua
+++ b/test/functional/plugin/lsp/inlay_hint_spec.lua
@@ -12,7 +12,8 @@ local api = n.api
 local clear_notrace = t_lsp.clear_notrace
 local create_server_definition = t_lsp.create_server_definition
 
-local text = dedent([[
+describe('vim.lsp.inlay_hint', function()
+  local text = dedent([[
 auto add(int a, int b) { return a + b; }
 
 int main() {
@@ -22,7 +23,7 @@ int main() {
 }
 }]])
 
-local response = [==[
+  local response = [==[
 [
 {"kind":1,"paddingLeft":false,"label":"-> int","position":{"character":22,"line":0},"paddingRight":false},
 {"kind":2,"paddingLeft":false,"label":"a:","position":{"character":15,"line":5},"paddingRight":true},
@@ -30,7 +31,7 @@ local response = [==[
 ]
 ]==]
 
-local grid_without_inlay_hints = [[
+  local grid_without_inlay_hints = [[
   auto add(int a, int b) { return a + b; }          |
                                                     |
   int main() {                                      |
@@ -42,7 +43,7 @@ local grid_without_inlay_hints = [[
                                                     |
 ]]
 
-local grid_with_inlay_hints = [[
+  local grid_with_inlay_hints = [[
   auto add(int a, int b){1:-> int} { return a + b; }    |
                                                     |
   int main() {                                      |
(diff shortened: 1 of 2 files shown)


mfussenegger:
Looks good to me but could you also add a test case?

clason:
(and rebase)

amitds1997:
> Looks good to me but could you also add a test case?

Thanks for the review. Done in the latest push.

test/functional/plugin/lsp/inlay_hint_spec.lua
```diff
@@ -258,3 +258,74 @@ describe('vim.lsp.inlay_hint', function()
     end)
   end)
 end)
+
+describe('Inlay hints are correctly rendered', function()
```
mfussenegger, line 262:
```suggestion
describe('Inlay hints handler', function()
```

Correctly is redundant/implied for a test case. See suggestion below for a variant that also communicates the expected outcome

test/functional/plugin/lsp/inlay_hint_spec.lua
```diff
@@ -258,3 +258,74 @@ describe('vim.lsp.inlay_hint', function()
...
+    vim.api.nvim_win_set_buf(0, bufnr)
+
+    client_id = vim.lsp.start({ name = 'dummy', cmd = server.cmd })
+  ]],
+      response
+    )
+    insert(text)
+  end)
+
+  it('when multiple inlay hints are at same character position', function()
```
mfussenegger, line 321:
```suggestion
  it('renders hints with same position in received order, function()
```

mfussenegger:
Thanks for the test case addition. I left a minor comment regarding naming, otherwise lgtm

test/functional/plugin/lsp/inlay_hint_spec.lua
amitds1997: Makes sense. 
Done in the latest commit: [7248b73](https://github.com/neovim/neovim/pull/29707/commits/7248b7363ffdd00e0ac3288a92d4c0d2f7396a5c). Also, rebased with current master.

test/functional/plugin/lsp/inlay_hint_spec.lua
amitds1997: Done in [7248b73](https://github.com/neovim/neovim/pull/29707/commits/7248b7363ffdd00e0ac3288a92d4c0d2f7396a5c).

clason:
Backport?

MariaSolOs:
> Backport?

I think we should.

PR #29735, opened 2024-07-16
test: fix reporting "no flush received" too early

zeertzjq:


zeertzjq:
This fixes the flakiness of "TUI draws screen lines with leading spaces correctly" test.
- Before: https://github.com/zeertzjq/neovim/actions/runs/9951361101/job/27490719001
- After: https://github.com/zeertzjq/neovim/actions/runs/9951381624/job/27490774052

But I'm not entirely sure if this is the correct solution.

bfredl (approved):

PR #29743, opened 2024-07-16
ci: run workflows on release branches

dundargoc:
Github doesn't allow workflows to be run from the `github-actions`
account, which is the default account. This caused the CI on backport
PRs to not be run. The way to circumvent this is to use a token that
essentially "pretends" to be another user, which in turn triggers the CI
as desired.

Also run lintcommit on release branches as that is now a required check,
meaning it must always be run.

### DIFF
--- a/.github/workflows/lintcommit.yml
+++ b/.github/workflows/lintcommit.yml
@@ -4,6 +4,7 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - 'master'
+      - 'release-[0-9]+.[0-9]+'
 jobs:
   lint-commits:
     runs-on: ubuntu-latest
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -12,12 +12,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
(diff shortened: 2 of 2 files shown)


.github/workflows/lintcommit.yml
```diff
@@ -4,6 +4,7 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - 'master'
+      - 'release-[0-9]+.[0-9]+'
```
clason, line 7: do we have (or plan to have) `release-x.y` branches?

clason:
If we are really trusting, we can add https://github.com/peter-evans/enable-pull-request-automerge to the mix. (Probably want to watch a few PRs in action and see how it goes first.)

.github/workflows/lintcommit.yml
dundargoc: Huh, good catch. I just copy pasted it from other workflows but all of them are wrong lol.

dundargoc: Wait no, this is right though. We do have branches of that form: `release-0.5` would be `x=0` and `y=5`.

clason: Ah, didn't think that far ahead. Smurt.

PR #29745, opened 2024-07-16
fix(float): close preview float window when no selected

glepnir:
Problem: float preview window still exist when back at origianl.

Solution: if selected is -1 mean back to origianl close preview window first.

### DIFF
--- a/test/functional/ui/popupmenu_spec.lua
+++ b/test/functional/ui/popupmenu_spec.lua
@@ -1684,25 +1684,26 @@ describe('builtin popupmenu', function()
           }
         end
 
-        -- info window position should be adjusted when new leader add
-        feed('<C-P>o')
+        -- delete one character make the pum width smaller than before
+        -- info window position should be adjusted when popupmenu width changed
+        feed('<BS>')
         if multigrid then
-          screen:expect {
+          screen:expect({
             grid = [[
-          ## grid 1
-            [2:----------------------------------------]|*10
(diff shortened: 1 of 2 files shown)


test/functional/ui/popupmenu_spec.lua
```diff
@@ -1717,22 +1718,89 @@ describe('builtin popupmenu', function()
...
+              {2:-- }{8:Back at original}                     |
+            ]],
+          })
         end
 
         -- test nvim__complete_set_info
-        feed('<ESC>cc<C-X><C-O><C-N><C-N>')
-        vim.uv.sleep(10)
+        feed('<ESC>S<C-X><C-O><C-N><C-N>')
+        exec_lua([[vim.wait(100)]])
```
zeertzjq, line 1803: What is the point of this wait?

glepnir: because when first select item 3 it doesn't have info so the preview window will close, but CompleteChanged callback is set info of 3. so wait this callback. there is test for this info set from event callback  so use vim.wait here.

zeertzjq: I don't understand how that requires you to use `vim.wait()`. The test runs fine even if you remove this line.

glepnir: emm i test it again. you are right it passed..before it not passed so i think maybe i need wait editor..

PR #29746, opened 2024-07-16
fix(column): modifying a sign should update placed signs

zeertzjq:
Problem:  Modifying a sign no longer updates already placed signs.
Solution: Keep track of sh_idx indexes of a sign and update a sign's
          DecorSignHighlight entries when modifying it.

### DIFF
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -292,6 +292,9 @@ static void decor_free_inner(DecorVirtText *vt, uint32_t first_idx)
   while (idx != DECOR_ID_INVALID) {
     DecorSignHighlight *sh = &kv_A(decor_items, idx);
     if (sh->flags & kSHIsSign) {
+      if (sh->sign_name) {
+        sign_after_delete(sh->sign_name, idx);
+      }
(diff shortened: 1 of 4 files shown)


zeertzjq:
I checked the performance impact of this PR with the following script:
```lua
vim.cmd([[call setline(1, range(1, 100000))]])
vim.fn.sign_define('s1', { text = '!!' })

local t = vim.uv.hrtime()
for i = 1, 100000 do
  vim.fn.sign_place(i, '', 's1', '', { lnum = i })
end
print((vim.uv.hrtime() - t) * 1e-9)

vim.cmd.redraw()

local t = vim.uv.hrtime()
for i = 1, 100000 do
  vim.fn.sign_unplace('', { buffer = '', id = i })
end
print((vim.uv.hrtime() - t) * 1e-9)
```
This seems to have less than 5% impact on the performance of placing and unplacing signs.

lewis6991 (approved):
I haven't looked at the sign code in a while since there has been lots of changes to it.

Since this is a fix with a test and the perf drop is minor, then I think this is good to merge.

PR #29750, opened 2024-07-16
fix(column): modifying a sign should update placed signs

luukvbaal:
Problem:  Modifying a sign no longer updates already placed signs.
Solution: Loop over (newly-exposed) placed decorations when modifying a
          sign definition. Update placed decor if it belongs to the sign
          that is modified.

### DIFF
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -31,9 +31,6 @@
 # include "decoration.c.generated.h"
 #endif
 
-// TODO(bfredl): These should maybe be per-buffer, so that all resources
-// associated with a buffer can be freed when the buffer is unloaded.
-kvec_t(DecorSignHighlight) decor_items = KV_INITIAL_VALUE;
 uint32_t decor_freelist = UINT32_MAX;
 
 // Decorations might be requested to be deleted in a callback in the middle of redrawing.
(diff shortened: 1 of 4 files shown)


src/nvim/sign.c
```diff
@@ -441,6 +435,27 @@ static int sign_define_by_name(char *name, char *icon, char *text, char *linehl,
     }
   }
 
+  // Update already placed signs and redraw if necessary when modifying a sign.
+  if (!new_sign) {
+    bool should_redraw = true;
+    for (size_t i = 0; i < kv_size(decor_items); i++) {
+      if (strequal(name, kv_A(decor_items, i).sign_name)) {
```
zeertzjq, line 442: Check for `kSHIsSign` first?

zeertzjq: Also, `kv_A(decor_items, i)` is used a lot of times here, so it may be worth assigning it to a variable.

luukvbaal: > Check for `kSHIsSign` first?

`strequal` checks if `sign_name` is non-NULL first, which is a smaller set (i.e. legacy signs only) for which `kSHIsSign` should always be set.

zeertzjq:
I've thought about this approach before, but didn't choose it, because `decor_items` may contain a lot of unrelated items (e.g. a lot of extmark signs).

luukvbaal:
> I've thought about this approach before, but didn't choose it, because `decor_items` may contain a lot of unrelated items (e.g. a lot of extmark signs).

Yes, but it seems preferable to me as this only affects redefining a sign, which should be infrequent.

src/nvim/decoration.c
```diff
@@ -293,6 +290,7 @@ static void decor_free_inner(DecorVirtText *vt, uint32_t first_idx)
     DecorSignHighlight *sh = &kv_A(decor_items, idx);
     if (sh->flags & kSHIsSign) {
       xfree(sh->sign_name);
+      sh->sign_name = NULL;  // Still referenced in "decor_items", avoid accessing.
```
zeertzjq, line 293: `XFREE_CLEAR`

PR #29763, opened 2024-07-17
fix(health): fix pyenv root and python exepath detect issue

AbaoFromCUG:
**This PR fix 2 issues**
When I run `:checkhealth provider.python` in `v.0.10.0`, I get
```shell
provider.python: require("provider.python.health").check()

Python 3 provider (optional) ~
- pyenv: Path: /home/abao/.pyenv/libexec/pyenv
- pyenv: $PYENV_ROOT is not set. Infer from `pyenv root`.
- WARNING pyenv: Root does not exist: /home/abao/.pyenv
  . Ignoring pyenv for all following checks.
- WARNING No Python executable found that can `import neovim`. Using the first available executable for diagnostics.
- WARNING Could not load Python :
  /home/abao/.pyenv/shims/python3 does not have the "neovim" module.
  /home/abao/.pyenv/shims/python3.12 does not have the "neovim" module.
  /home/abao/.pyenv/shims/python3.11 does not exist: 
  /home/abao/.pyenv/shims/python3.10 does not have the "neovim" module.
  /home/abao/.pyenv/shims/python3.9 does not exist: 
  python3.8 not found in search path or not executable.
  python3.7 not found in search path or not executable.
  /home/abao/.pyenv/shims/python does not have the "neovim" module.
  - ADVICE:
    - See :help |provider-python| for more information.
    - You may disable this provider (and warning) by adding `let g:loaded_python3_provider = 0` to your init.vim
- Executable: Not found

Python virtualenv ~
- ERROR Failed to run healthcheck for "provider.python" plugin. Exception:
  /usr/share/nvim/runtime/lua/provider/python/health.lua:20: 
  ```

## pyenv root detection issue
When I don't  set `PYENV_ROOT` environment variable, neovim will detect pyenv's root via `pyenv root` command, but which will be always faild, because `vim.fn.system()` return result with additional `\n`,  so we not to trim it before check whether  it is exists
> By the way, I use pure lua function `vim.system` replace `vim.fn.system`

## python executable path detection issue
Filter unrelated `python-config`, in my case, which includes `python-config`, `python3.10-config`, `python3.11-config`, `python3.12-config` because  I install mulitiple version python. So I simply filter out all the `python.*%-config` result

### DIFF
--- a/runtime/lua/vim/provider/health.lua
+++ b/runtime/lua/vim/provider/health.lua
@@ -366,7 +366,17 @@ local function check_for_pyenv()
   local pyenv_root = vim.fn.resolve(os.getenv('PYENV_ROOT') or '')
 
   if pyenv_root == '' then
-    pyenv_root = vim.fn.system({ pyenv_path, 'root' })
+    local p = vim.system({ pyenv_path, 'root' }):wait()
+    if p.code ~= 0 then
+      local message = string.format(
+        'pyenv: Failed to infer the root of pyenv by running `%s root` : %s. Ignoring pyenv for all following checks.',
+        pyenv_path,
+        p.stderr
+      )
+      health.warn(message)
+      return { '', '' }
+    end
+    pyenv_root = vim.trim(p.stdout)
     health.info('pyenv: $PYENV_ROOT is not set. Infer from `pyenv root`.')
   end
 
@@ -754,7 +764,7 @@ local function python()
   local venv_bins = vim.fn.glob(string.format('%s/%s/python*', virtual_env, bin_dir), true, true)
   venv_bins = vim.tbl_filter(function(v)
     -- XXX: Remove irrelevant executables found in bin/.
-    return not v:match('python%-config')
+    return not v:match('python.*%-config')
   end, venv_bins)
   if vim.tbl_count(venv_bins) > 0 then
     for _, venv_bin in pairs(venv_bins) do

### CODE REVIEW

File: runtime/lua/vim/provider/health.lua [hunk 1]
```diff
@@ -366,7 +366,8 @@ local function check_for_pyenv()
   local pyenv_root = vim.fn.resolve(os.getenv('PYENV_ROOT') or '')
 
   if pyenv_root == '' then
-    pyenv_root = vim.fn.system({ pyenv_path, 'root' })
+    local p = vim.system({ pyenv_path, 'root' }):wait()
```
glepnir, line 369: I think it's better add stderr/code error handling.

AbaoFromCUG: Done, add check if  `p.code~=0` and call `health.warn` to report error

glepnir: Just checking the exit code is not enough. I prefer also add stderr check.

AbaoFromCUG: Are you mean stderr maybe nil? I don't understand,  the content of stderr is irrelevant with neovim, I prefer just show it to user

glepnir: some programs will output errors to stderr but the exit code is normal. not sure if it will appear here. so I suggest checking stderr.

AbaoFromCUG: No, `pyenv root` just a warapper of [`pyenv-root`](https://github.com/pyenv/pyenv/blob/master/libexec/pyenv-root), before execute `pyenv-root` , pyenv set the `PYENV_PATH` variable in script scope, so the `pyenv-root` only need call `echo "$PYENV_ROOT"`  as show in the `pyenv-root` script's content, the `echo` command isn't the case which you mention.

In the other words, `pyenv root` just a complex `echo $PYENV_ROOT`. So, I think it is enought to display the stderror

To be honest, if `pyenv` is a executable (which is ensured in previous code), `pyenv root` must work as our want,  refer to https://github.com/pyenv/pyenv/tree/master?tab=readme-ov-file#locating-pyenv-provided-python-installations

AbaoFromCUG:
@glepnir  hi, any issue do you have about this issue,  could this PR be merged? thanks

glepnir:
cc @dundargoc  PTAL

dundargoc (approved):
Looks good. Next time don't forget to add the full explanation/reasoning in the commit message (which is more important than PR message).

clason:
Backport? (After release is before release.)

Issue #27509, opened 2024-02-17
Press/release key event support


gpanders:
### Problem

The [Kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol) supports sending press/repeat/release events for key presses. We currently do not enable this mode and termkey is not currently able to parse these events, but we certainly could.

A possible use case for press/release events would be to show a floating window only while a key is held down (e.g. for LSP hover or diagnostics).

Adding support to libtermkey is fairly straightforward (I already have a working prototype in a branch). The more difficult problem is plumbing this into Neovim's own event handling.

At least one question is how a press/release event should be represented. We could limit key events only to the API (via `nvim_set_keymap`) so that we don't need to modify `:map` and friends.

Related: https://github.com/neovim/neovim/issues/14400

### Expected behavior

N/A

justinmk:
> We should be able to leave "press" events as-is and only add special casing for "release" events

Currently if I hold down key `x` (without release) it emits the same keycode as if I press-and-release `x`.

I suppose it makes sense for us to formally define that as [keydown](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event), and that doesn't really change the semantics of existing mappings/scripts, because there was never an expectation that a keycode implied press-and-release.

ColinKennedy:
Curious - if press down is supported in keymaps, would it then be possible to produce "key progression" mappings?

e.g.

(press asdf one finger at a time until all 4 keys are pressed down and then let go of a s d then f one finger at a time)
```lua
vim.keymap.set(
    "n",
    "<PDown>a<PDown>s<PDown>d<PDown>f<PUp>a<PUp>s<PUp>d<PUp>f",
    'echomsg "hello"', {}
)
```

Or maybe even "chord" mappings?

(press asdf down in any order [or at the same time] and then release in any order [or at the same time])
```lua
vim.keymap.set(
    "n",
    "<PDownStart>asdf<PDownEnd><PUpStart>asdf<PUpEnd>",
    'echomsg "hello"', {}
)
```

The potential would be pretty interesting, assuming Terminal emulators support that kind of input. I have no idea though, terminal inner workings are pretty foreign to me.

gpanders:
With https://github.com/neovim/neovim/pull/29805 merged, the next step for this is to figure out this part:

>The more difficult problem is plumbing this into Neovim's own event handling.
>
[...]

PR #29805, opened 2024-07-19
feat(tui): parse CSI subparams in termkey

gpanders:
libtermkey does not know how to parse CSI subparameters (parameters separated by ':', ASCII 0x3A) and currently just ignores them. However, many important CSI sequences sent by the terminal make use of subparameters, most notably key events when using the kitty keyboard protocol [1]. Enabling subparameters is a prerequisite for [expanding kitty keyboard protocol](https://github.com/neovim/neovim/issues/27509) support in Neovim.

Concretely, we do this by returning pointers into the internal termkey buffer for each CSI parameter rather than parsing them into integers directly. When a caller wants to actually use the parameter as an integer, they must call termkey_interpret_csi_param, which parses the full parameter string into an integer parameter and zero or more subparameters.

The pointers into the internal buffer will become invalidated when new input arrives from the terminal so it is important that the individual params are used and parsed right away. All of our code (and libtermkey's code) does this, so this is fine for now, but is something to keep in mind moving forward.

[1]: https://sw.kovidgoyal.net/kitty/keyboard-protocol/

### DIFF
--- a/src/nvim/tui/input.c
+++ b/src/nvim/tui/input.c
@@ -596,10 +596,10 @@ static void handle_unknown_csi(TermInput *input, const TermKeyKey *key)
 {
   // There is no specified limit on the number of parameters a CSI sequence can
   // contain, so just allocate enough space for a large upper bound
-  long args[16];
-  size_t nargs = 16;
+  TermKeyCsiParam params[16];
+  size_t nparams = 16;
   unsigned long cmd;
-  if (termkey_interpret_csi(input->tk, key, args, &nargs, &cmd) != TERMKEY_RES_KEY) {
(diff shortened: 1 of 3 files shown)


justinmk:
Do we need test coverage for this? I see `TermKeyCsiParam` in https://github.com/neovim/neovim/blob/9742a5d060d95cc1b9f57bc947cc69efec994e5b/test/unit/termkey_spec.lua#L881-L884 , but I assume that was a direct port of libtermkey's tests and doesn't cover the new logic from this PR. cc @dundargoc

dundargoc:
Yup, you are right. This was a straight port of almost all libvterm tests without any modification.

PR #29809, opened 2024-07-20
fix(loader): follow the style of the error message for built-in loaders

futsuuu:
The error message should be started with `"\n\t"` and the module name should be surrounded by single quotes.

before

```
:lua require("foo")
E5108: Error executing lua [string ":lua"]:1: module 'foo' not found:
	no field package.preload['foo']
cache_loader: module foo not found
cache_loader_lib: module foo not found
	no file '.\foo.lua'
	...
```

after

```
:lua require("foo")
E5108: Error executing lua [string ":lua"]:1: module 'foo' not found:
	no field package.preload['foo']
	cache_loader: module 'foo' not found
	cache_loader_lib: module 'foo' not found
	no file '.\foo.lua'
	...
```

### DIFF
--- a/runtime/lua/vim/loader.lua
+++ b/runtime/lua/vim/loader.lua
@@ -200,7 +200,7 @@ function Loader.loader(modname)
     return chunk or error(err)
   end
   Loader._hashes = nil
-  return '\ncache_loader: module ' .. modname .. ' not found'
+  return ("\n\tcache_loader: module '%s' not found"):format(modname)
 end
 
 --- The `package.loaders` loader for libs
@@ -222,7 +222,7 @@ function Loader.loader_lib(modname)
     local chunk, err = package.loadlib(ret.modpath, 'luaopen_' .. funcname:gsub('%.', '_'))
     return chunk or error(err)
   end
-  return '\ncache_loader_lib: module ' .. modname .. ' not found'
+  return ("\n\tcache_loader_lib: module '%s' not found"):format(modname)
 end
 
 --- `loadfile` using the cache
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/loader.lua
```diff
@@ -200,7 +200,7 @@ function Loader.loader(modname)
     return chunk or error(err)
   end
   Loader._hashes = nil
-  return '\ncache_loader: module ' .. modname .. ' not found'
+  return ("\n\tcache_loader: module '%s' not found"):format(modname)
```
glepnir, line 203: could you also add a test case in `test/functional/lua/load_spec.lua` ? something like

```lua
  it('correct indent on error message', function()
    exec_lua [[
      vim.loader.enable()
    ]]
    exec_capture("lua require('foo')")
  end)
```

futsuuu: Sure! Sorry, I couldn't find loader_spec file when I created the PR...

glepnir: internal assertions is wired. you can capture the output, simply case like 
```lua
 it('correct indent on error message (#29809)', function()
   local result = exec_lua [[
      vim.loader.enable()
      local _, errmsg = pcall(require, 'non_existent_module')
      errmsg = vim.split(errmsg, '\n')
      return {errmsg[3], errmsg[4]}
    ]]
    eq({'\tcache_load: module non_exit_module not found','\tcache_load_lib: module non_exit_module not found'}, result)
  end)
```

lewis6991 (approved):

PR #29837, opened 2024-07-23
Fixed deleting backslash file. Copy of chrisbra's Vim commit

zer0reaction:
Found the issue with deleting file with name \ (backslash). chrisbra fixed it in the Vim repository, I copied his commit here

More on the commit and the issue here:
https://github.com/vim/vim/commit/581d4a7b356395bcb8606c1717ded65d47d26c68
https://github.com/vim/vim/issues/15330


clason:
Please use the vim-patch script; this will take care of the necessary(!) bookkeeping.

clason:
(But thank you!)

PR #29857, opened 2024-07-25
'keywordprg' was set to `:help` twice in the vim ftplugin

Futarimiti:
I don't see a particular reason for this. Not observed in vim.

### DIFF
--- a/runtime/ftplugin/vim.vim
+++ b/runtime/ftplugin/vim.vim
@@ -47,6 +47,7 @@ setlocal fo-=t fo+=croql
 " keyword character.  E.g., for netrw#Nread().
 setlocal isk+=#
 
+" Prefer Vim help instead of manpages.
 " Use :help to lookup the keyword under the cursor with K.
 setlocal keywordprg=:help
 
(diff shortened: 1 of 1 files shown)


clason:
Probably happened mistakenly during porting runtime file updates.

runtime/ftplugin/vim.vim
```diff
@@ -47,6 +47,7 @@ setlocal fo-=t fo+=croql
 " keyword character.  E.g., for netrw#Nread().
 setlocal isk+=#
 
+" Prefer Vim help instead of manpages.
 " Use :help to lookup the keyword under the cursor with K.
 setlocal keywordprg=:help
```
clason, line 52:
```suggestion
```
Please remove this and leave the other place to match the upstream ftplugin (helps prevent this cropping up again).

runtime/ftplugin/vim.vim
```diff
@@ -47,8 +47,6 @@ setlocal fo-=t fo+=croql
 " keyword character.  E.g., for netrw#Nread().
 setlocal isk+=#
 
-" Use :help to lookup the keyword under the cursor with K.
-setlocal keywordprg=:help
```
zeertzjq, line 51: Why are both instances removed?

zeertzjq (changes requested):

runtime/ftplugin/vim.vim
zeertzjq: As mentioned above, you should keep this one and remove the one below.

runtime/ftplugin/vim.vim
```diff
@@ -47,9 +47,6 @@ setlocal fo-=t fo+=croql
...
```
zeertzjq, line 51: As mentioned above, you should keep this one and remove the one below.

Futarimiti:
@zeertzjq Does this look good to you?

zeertzjq (approved):

PR #27011, opened 2024-01-14
perf(extmarks): add metadata for efficient filtering of special decorations

bfredl:
This expands on the global "don't pay for what you don't use" rules for these special extmark decorations:

- inline virtual text, which needs to be processed in plines.c when we calculate the size of text on screen
- virtual lines, which are needed when calculating "filler" lines
- signs, with text and/or highlights, both of which needs to be processed for the entire line already at the beginning of a line.

This adds a count to each node of the marktree, for how many special marks of each kind can be found in the subtree for this node. This makes it possible to quickly handle extra checks, when working in regions of the buffer not containing these kind of marks, instead of before where this could just be skipped if the entire _buffer_ didn't contain such marks.

Note: this PR only covers some of the places where this extra information is useful. For instance I haven't touched the the "b_signcols" code yet as there is ongoing work #26898 by @luukvbaal and it would benefit from some discussion.

PR #29858, opened 2024-07-25
fix(decor): don't draw invalidated virtual lines

luukvbaal:


src/nvim/decoration.c
```diff
@@ -924,7 +924,7 @@ int decor_virt_lines(win_T *wp, linenr_T lnum, VirtLines *lines, TriState has_fo
   while (true) {
     MTKey mark = marktree_itr_current(itr);
     DecorVirtText *vt = mt_decor_virt(mark);
-    if (ns_in_win(mark.ns, wp)) {
+    if (!mt_invalid(mark) && ns_in_win(mark.ns, wp)) {
```
zeertzjq, line 927: While at it, I wonder if `vt` should be checked here first before the `ns_in_win()` check, as `ns_in_win()` is more costly.
```suggestion
    if (vt && !mt_invalid(mark) && ns_in_win(mark.ns, wp)) {
```
The same seems to apply to `decor_redraw_col()` too, where `ns_in_win()` is currently the first check, but it will be better for it to be the last.

zeertzjq: Oh, it seems that `mt_end()` was actually removed from this function in #27011, but it wasn't removed elsewhere.

luukvbaal: Hmm yeah because we are iterating with the `kMTMetaLines` filter, I would assume `mark.flags & MT_FLAG_DECOR_VIRT_LINES` to always be true. Presumably `mt_end()` is still relevant and should be checked?

zeertzjq: Oh, I didn't notice the filter.

PR #29868, opened 2024-07-26
Correct synonym(s) for "s" in the documentation

rwmpelstilzchen:
"cl" has the same effect as "s" only when 'rightleft' is not set (the default), while "c<Space>" has the same effect under all conditions and is thus preferred as a general-purpose synonym.


zeertzjq:
This isn't really true. It's equivalent to `cl` from a mapping, but not typed `cl`.

clason:
And this documentation is taken verbatim form Vim so should be changed there (if at all).

PR #29877, opened 2024-07-27
fix(lsp): revert text edit application order change

mfussenegger:
Reverts https://github.com/neovim/neovim/pull/29212 and adds a few
additional test cases

After re-reading the spec several times:

> All text edits ranges refer to positions in the document they are
> computed on. They therefore move a document from state S1 to S2 without
> describing any intermediate state. Text edits ranges must never overlap,
> that means no part of the original document must be manipulated by more
> than one edit. However, it is possible that multiple edits have the same
> start position: multiple inserts, or any number of inserts followed by a
> single remove or replace edit. If multiple inserts have the same
> position, the order in the array defines the order in which the inserted
> strings appear in the resulting text.

I think the fix was actually wrong. The important part is here:

> If multiple inserts have the same position, the order in the array
> defines the order in which the inserted strings appear in the
> resulting text.

Emphasis on _appear in the resulting text_

Which means that in:

    local edits1 = {
      make_edit(0, 3, 0, 3, { 'World' }),
      make_edit(0, 3, 0, 3, { 'Hello' }),
    }

`World` must appear before `Hello` in the final text. That means the old
logic was correct, and the fix was wrong.

cc @alcolmenar

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -396,10 +396,7 @@ function M.apply_text_edits(text_edits, bufnr, offset_encoding)
     if a.range.start.character ~= b.range.start.character then
       return a.range.start.character > b.range.start.character
     end
-    if a._index ~= b._index then
-      return a._index < b._index
-    end
-    return false
+    return a._index > b._index
   end)
 
   -- save and restore local marks since they get deleted by nvim_buf_set_lines
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -1756,6 +1756,58 @@ describe('LSP', function()
     }
   end
 
+  describe('apply vscode text_edits', function()
+    it('single replace', function()
+      insert('012345678901234567890123456789')
+      local edits = {
+        make_edit(0, 3, 0, 6, { 'Hello' }),
+      }
+      exec_lua('vim.lsp.util.apply_text_edits(...)', edits, 1, 'utf-16')
+      eq({ '012Hello678901234567890123456789' }, buf_lines(1))
+    end)
+    it('two replaces', function()
+      insert('012345678901234567890123456789')
+      local edits = {
+        make_edit(0, 3, 0, 6, { 'Hello' }),
+        make_edit(0, 6, 0, 9, { 'World' }),
+      }
+      exec_lua('vim.lsp.util.apply_text_edits(...)', edits, 1, 'utf-16')
+      eq({ '012HelloWorld901234567890123456789' }, buf_lines(1))
+    end)
+    it('same start pos insert are kept in order', function()
+      insert('012345678901234567890123456789')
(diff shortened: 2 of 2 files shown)


alcolmenar:
hey @mfussenegger , I'm going to have to disagree on the original fix being incorrect. Others have mentioned that it moved away from VSCode's implementation, but looking at the implementation [here](https://github.com/microsoft/vscode-languageserver-node/blob/81daeee7362603c63311416f7d6cb8cdf74572ff/textDocument/src/main.ts?plain=1#L366-L372):

```typescript
		const sortedEdits = mergeSort(edits.map(getWellformedEdit), (a, b) => {
			const diff = a.range.start.line - b.range.start.line;
			if (diff === 0) {
				return a.range.start.character - b.range.start.character;
			}
			return diff;
		});
...
		const ret = compare(left[leftIdx], right[rightIdx]);
		if (ret <= 0) {
			// smaller_equal -> take left to preserve order
			data[i++] = left[leftIdx++];
		} else {
			// greater -> take right
			data[i++] = right[rightIdx++];
		}
```

I think the fix moves us closer to the reference implementation. I haven't really had a chance to actually test it in practice though.

Thank you for looking at this again though and sorry for all the extra work that this might've caused.

max397574:
As far as I know (iirc this was an lsp spec issue) vscode doesn't implement this as it is written in the spec.
And I think neovim should do as the spec says and not what vscode does.

mfussenegger:
> I think the fix moves us closer to the reference implementation so I'm a bit confused as to why others have mentioned that. I haven't really had a chance to actually test it in practice though.

No it didn't. The PR here actually copies some of the vscode test cases and they failed with your fix, but passed on our original implementation.

> As far as I know (iirc this was an lsp spec issue) vscode doesn't implement this as it is written in the spec.
And I think neovim should do as the spec says and not what vscode does.

vscode has test cases here:

https://github.com/microsoft/vscode-languageserver-node/blob/main/textDocument/src/test/edits.test.ts

The PR here copies some of these.

One that's relevant to the "same start position" case is:

```
assert.equal(applyEdits(input, [TextEdit.insert(Position.create(0, 3), 'World'), TextEdit.replace(Range.create(Position.create(0, 3), Position.create(0, 6)), 'Hello')]), '012WorldHello678901234567890123456789');
```

Which clearly matches the text in the specification:

> If multiple inserts have the same position, the order in the array
> defines the order in which the inserted strings appear in the
> resulting text.

Note that this doesn't say that the text edits that appear first in the array must be applied first. It says that the *the order in the array defines the order in which the inserted strings appear in the resulting text*

With that my interpretation is that:

- vscode matches the specification
- The fix merged in neovim was wrong and no longer matched the spec
- This PR brings neovim back in line with the specification _and_ in line with the vscode implementaiton

alcolmenar:
That makes sense. Thank you for the breakdown and apologies for the extra work!

MariaSolOs (approved):

PR #29881, opened 2024-07-27
docs(insert): mark vim code blocks as such

ribru17:


clason:
Sorry, but we try not to touch the help pages taken from Vim beyond correcting factual inaccuracies due to divergence (otherwise I would have done this a long time ago...)

ribru17:
I see; would this be a valid PR to vim (and then a patch here)?

clason:
> I see; would this be a valid PR to vim (and then a patch here)?

No, because they don't use language injections. Until we finally decide to take ownership of _all_ docs and rewrite them according to _our_ (stricter) vimdoc spec, they will have to stay as they are.

ribru17:
Got it, thanks

PR #29882, opened 2024-07-28
feat(lsp): support lsp.completion "deprecated" field

glepnir:
Problem: CompletionItem in lsp spec mentioned the deprecated attribute

Solution: when item has deprecated attribute set hl_group to DiagnosticDeprecated
                 in complete function

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -253,6 +253,13 @@ function M._lsp_to_complete_items(result, prefix, client_id)
   for _, item in ipairs(items) do
     if matches(item) then
       local word = get_completion_word(item)
+      local hl_group = ''
+      if
+        item.deprecated
+        or vim.list_contains((item.tags or {}), protocol.CompletionTag.Deprecated)
+      then
+        hl_group = 'DiagnosticDeprecated'
+      end
       table.insert(candidates, {
         word = word,
         abbr = item.label,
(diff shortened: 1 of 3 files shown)


mfussenegger:
Looks like we should also set `deprecatedSupport` to true in protocol.lua 
And the `deprecated` property itself is deprecated in favour of `tags` so we should check that too.

Could you also add test cases for both scenarios? One where `deprecated` is set and one where `tags` are used?

runtime/lua/vim/lsp/completion.lua
```diff
@@ -253,6 +253,10 @@ function M._lsp_to_complete_items(result, prefix, client_id)
   for _, item in ipairs(items) do
     if matches(item) then
       local word = get_completion_word(item)
+      local hl_group = ''
+      if item.deprecated or vim.list_contains((vim.tbl_get(item, 'tags') or {}), 1) then
```
glepnir, line 257: there use a constant more better ..

runtime/lua/vim/lsp/completion.lua
```diff
@@ -253,6 +253,13 @@ function M._lsp_to_complete_items(result, prefix, client_id)
   for _, item in ipairs(items) do
     if matches(item) then
       local word = get_completion_word(item)
+      local hl_group = ''
+      if
+        item.deprecated
+        or vim.list_contains((vim.tbl_get(item, 'tags') or {}), protocol.CompletionTag.Deprecated)
```
mfussenegger, line 259:
```suggestion
        or vim.list_contains(item.tags or {}, protocol.CompletionTag.Deprecated)
```

mfussenegger (approved):

MariaSolOs (approved):

PR #29894, opened 2024-07-28
Better consistency with anonymous nodes

ribru17:
This PR has three different commits, each with relatively small changes (but ones that change the API). In order of when they appear:

### feat(treesitter): add node_for_range function

- This is identical to `named_node_for_range` except that it includes
anonymous nodes. This maintains consistency in the API because we
already have `descendant_for_range` and `named_descendant_for_range`.

### feat(treesitter): allow get_node to return anonymous nodes

- Adds a new field `include_anonymous` to the `get_node` options to allow
anonymous nodes to be returned.

### fix(treesitter): highlight anonymous nodes in inspect_tree

- **Problem:** With anonymous nodes toggled in the inspect tree, only
named nodes will be highlighted when moving the cursor in the source
code buffer.
\
**Solution:** Retrieve the anonymous node at the cursor (when toggled on
in the inspect tree) and highlight them when appropriate, for better
clarity/specificity.

### DIFF
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -1375,6 +1375,19 @@ LanguageTree:named_node_for_range({range}, {opts})
     Return: ~
         (`TSNode?`)
 
+                                               *LanguageTree:node_for_range()*
+LanguageTree:node_for_range({range}, {opts})
+    Gets the smallest node that contains {range}.
+
+    Parameters: ~
+      • {range}  (`Range4`) `{ start_line, start_col, end_line, end_col }`
+      • {opts}   (`table?`) A table with the following fields:
+                 • {ignore_injections}? (`boolean`, default: `true`) Ignore
+                   injected languages
+
+    Return: ~
+        (`TSNode?`)
+
 LanguageTree:parse({range})                             *LanguageTree:parse()*
     Recursively parse all regions in the language tree using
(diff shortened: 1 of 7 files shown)


clason:
Can the fix commit be separated out so we can backport it?

ribru17:
The fix commit depends on the previous changes; would it be crazy to do a separate backport PR using the existing changes (ie descendant function) with a TODO to update the logic?

runtime/doc/treesitter.txt
```diff
@@ -1366,6 +1368,19 @@ LanguageTree:language_for_range({range})
...
+      • {range}  (`Range4`) `{ start_line, start_col, end_line, end_col }`
+      • {opts}   (`table?`) A table with the following fields:
+                 • {ignore_injections}? (`boolean`, default: `true`) Ignore
+                   injected languages
+
+    Return: ~
+        (`TSNode?`)
+
+                                               *LanguageTree:node_for_range()*
+LanguageTree:node_for_range({range}, {opts})
```
justinmk, line 1381: Do people actually like `named_` variants of all these functions, or would it make sense to make "named vs anon" a parameter?

ribru17: That sounds good to me! I don't have a strong preference either way.

PR #22949, opened 2023-04-08
fix(packaging): revert installation context to per-machine on Windows

sitiom:
Fixes #22933. Revert this change until a proper solution is implemented, such as https://github.com/neovim/neovim/issues/22933#issuecomment-1500766519.

This change should also be backported to the https://github.com/neovim/neovim/tree/release-0.9 branch.

PR #29895, opened 2024-07-28
fix(win-msi): set installer scope to machine

### LINKED ISSUE #29885: `ALLUSERS` property missing in neovim 0.10.1 MSI
### Problem

The 0.10.1 neovim MSI installer doesn't have an `ALLUSERS` property in the MSI property table despite it having the property in 0.10.0. This means that the installer is defined as installing in user scope.

I have documented this in https://github.com/russellbanks/Komac/issues/709#issuecomment-2254422052.

Also see https://github.com/microsoft/winget-pkgs/issues/164944 for issues this causes.

### Steps to reproduce

Open the neovim MSI in [Orca](https://learn.microsoft.com/windows/win32/msi/orca-exe). There is no `ALLUSERS` property in the property table.

### Expected behavior

There should be an `ALLUSERS` property of 1, as there is in neovim 0.10.0.

### Neovim version (nvim -v)

0.10.1

N/A

### Operating system/version

N/A

### Terminal name/version

N/A

N/A

### Installation

N/A

a-mnich:
Problem:
The windows installer did not have the AllUsers property which leads to the installer being misidentified as per user installer. Currently the installer already requires administrative privileges and installs into the system-wide ProgramFiles directory, but the start menu entry and uninstaller registration are created only for the current user. 

https://cmake.org/cmake/help/latest/cpack_gen/wix.html#variable:CPACK_WIX_INSTALL_SCOPE

Solution:
With setting CPACK_WIX_INSTALL_SCOPE to "perMachine" the generated msi installer includes the Property ALLUSERS=1.
Additionally the start menu entries and uninstaller registration will be created for all users.

resolves #29885

fixed-installer

### DIFF
--- a/cmake.packaging/CMakeLists.txt
+++ b/cmake.packaging/CMakeLists.txt
@@ -38,6 +38,7 @@ if(WIN32)
   # Create start menu and desktop shortcuts
   set(CPACK_WIX_PROGRAM_MENU_FOLDER "${CPACK_PACKAGE_NAME}")
   set(CPACK_PACKAGE_EXECUTABLES "nvim" "Neovim")
+  set(CPACK_WIX_INSTALL_SCOPE "perMachine")
 
   set(CPACK_WIX_UI_REF "WixUI_CustomInstallDir")
   list(APPEND CPACK_WIX_EXTRA_SOURCES ${CMAKE_CURRENT_LIST_DIR}/WixUI_CustomInstallDir.wxs)


dundargoc:
This looks familiar. I suspect we may have already done this before. Git blame is needed to understand what happened, there's been a lot of back and forth with the ALLUSERS property

zeertzjq:
#22949 looks related

a-mnich:
I currently don't understand why the 0.10.0 installer from the release page includes the ALLUSERS attribute at all.  The 0.9.5 installer doesn't include it and I couldn't find any changes in the git blame between those versions that would explain the inclusion.  

Additionally when [running the release worklow](https://github.com/a-mnich/neovim/actions/runs/10134181978) on the `v0.10.0` tag again the installer artifact doesn't include the attribute anymore. I thought that I should be able to reproduce the installer with this.
grafik

justinmk:
> This looks familiar. I suspect we may have already done this before ... #22949 looks related

Nice catch. https://wixtoolset.org/docs/schema/wxs/packagescopetype/ mentions `perUserOrMachine` which we haven't tried before, but is apparently equivalent to what was reverted in #22949 . 

> perUserOrMachine: Set this value to declare that the package is [dual-purpose that can install per-user or per-machine](https://learn.microsoft.com/en-us/windows/win32/msi/single-package-authoring). Sets the ALLUSERS property to 2 and MSIINSTALLPERUSER property to 1.

# Resolution

We haven't tried `ALLUSERS=1` explicitly yet, so let's do so. 

> I currently don't understand why the 0.10.0 installer from the release page includes the ALLUSERS attribute at all

We don't set anything explicitly, so we are subject to system defaults? Whatever we end up deciding, we should explicitly specify it via `CPACK_WIX_INSTALL_SCOPE`.

PR #29914, opened 2024-07-30
fix(watch): .git was not excluded

wilriker:
`inotifywait` man page specifies:
The file must be specified with a relative or absolute path according to whether a relative or absolute path is given for watched directories.

So it would only work this way in case the path is relative (which at least for `gopls` it is not)


mfussenegger:
@cryptomilk does this look okay to you?

cryptomilk:
If you're in a subdirectory of the working directory, you might better use an absolute path. I think this is fine. :+1:

PR #29956, opened 2024-08-02
feat(lsp): announce codeLens resolveSupport

mfussenegger:
The codelens implementation can resolve command and title via
`codeLens/resolve`.

The spec added client capabilities for that:

https://github.com/microsoft/language-server-protocol/pull/1979

### DIFF
--- a/runtime/lua/vim/lsp/protocol.lua
+++ b/runtime/lua/vim/lsp/protocol.lua
@@ -427,6 +427,12 @@ function protocol.make_client_capabilities()
           properties = { 'edit' },
         },
       },
+      codeLens = {
+        dynamicRegistration = false,
+        resolveSupport = {
+          properties = { 'command' },
+        },
+      },
       formatting = {
         dynamicRegistration = true,
       },


runtime/lua/vim/lsp/protocol.lua
```diff
@@ -427,6 +427,12 @@ function protocol.make_client_capabilities()
           properties = { 'edit' },
         },
       },
+      codeLens = {
+        dynamicRegistration = false,
+        resolveSupport = {
+          properties = { 'command', 'title' },
```
MariaSolOs, line 433: Can the title be resolved lazily? If I understand [the code](https://github.com/neovim/neovim/blob/0a0962a2e8d9ffa5f03e492fa293e964650a965e/runtime/lua/vim/lsp/codelens.lua#L234) correctly only the command is resolved.

mfussenegger: Good point. The title is actually part of the command

MariaSolOs (approved):

PR #29976, opened 2024-08-03
feat(ui): allow statusline-like ui elements to combine highlight attrs

### LINKED ISSUE #26037: The background in WinBar is taken from `Normal` highlight
### Problem

When setting a custom highlight in WinBar (using `%#..#`), the background is taken from `Normal` highlight

### Steps to reproduce

1) Define `Normal` and `WinBar` highlights with different `bg`.
2) 

```
vim.opt_local.winbar = 'hello'
```

Result - Ok, background is from `WinBar`

3) 

```
vim.opt_local.winbar = '%#TestBug#hello'
```

`TestBug` is no existing highlight.
Result - Ok, background is from `WinBar`

4)
Now define `TestBug`, but without `bg`.
Result - background is from `Normal`

### Expected behavior

Background is from `WinBar `

### Neovim version (nvim -v)

NVIM v0.10.0-dev-1507+g267c346e2-Homebrew

dont know

### Operating system/version

macOS 14.1

### Terminal name/version

Alacritty fork

xterm-256color

### Installation

brew

ribru17:
Fixes #26037

Patch courtesy of @zeertzjq (see https://github.com/neovim/neovim/issues/26037#issuecomment-1838548013)


echasnovski:
A slightly drive-by comment: I think it is a good idea to have combining highlight attributes not only in winbar, but also other "UI" parts. Like statusline, tabline, statuscolumn, and pmenu.

For example, without combining there is no good way of changing default highlighting of `PMenuMatch` and `PMenuMatchSel` to be bold or underline. Right now it will lead to showing character matches in pmenu bold but with `Normal` background and foreground.

ribru17:
Makes sense. I'll edit the commit/PR description, as this commit also gives attr inheritance to statusline, not just winbar. I can look into the others in a separate PR

justinmk (approved):
idk why the news check passed, but this probably should be mentioned in news.

justinmk (approved):

dundargoc:
> idk why the news check passed, but this probably should be mentioned in news.

It is not magic. It is only triggered if a commit message (not PR title) has type `feat` or if the commit message indicates it is breaking. How would you like to improve on it? Were you thinking it should also work if it had a "breaking-change" label, but no `!` in the commit message to indicate a breaking change?

ribru17:
Sorry, my fault; I didn't realize the commit message was "fix"

justinmk:
> It is not magic.

I thought our CI was running on magic. My mistake

echasnovski:
I've just tested functionality from this PR. Although it is indeed an improvement, from the title I expected a slightly different behavior.

Statusline-like ui element (statusline, tabline, winbar) can define highlighting via `%#HLname#` which *sets this highlighting for all characters to its right*. So by "combine highlight attributes" I'd expect that adding new highlight group would combine attributes with ones that are currently active. Here is an example:

1. `:hi AAA guibg=Red`
2. `:hi FG guifg=Black`
3. `:set showtabline=2`
4. `:lua vim.o.tabline = '%#AAA#  %#FG#my text'`. I would expect that "my text" (and everything to its right) would be shown with red background and black foreground. Instead it is shown with background from `TabLine` and black foreground.

This behavior also fixes the original #26037 (assuming that initial attributes to combine with come from `TabLine` / `WinBar` / `WinBarNC` / `StatusLine` / `StatusLineNC` highlight group).

@ribru17, @zeertzjq: does this all make sense?

zeertzjq:
No

echasnovski:
Why?

ribru17:
imo this would be too breaking- seems like it would be impossible to just define regular, non cascading highlights (red then blue then green etc)

echasnovski:
> imo this would be too breaking- seems like it would be impossible to just define regular, non cascading highlights (red then blue then green etc)

It is breaking in exactly the same way as this PR:
- Now attributes are combined with `TabLine` / `WinBar` / `WinBarNC` / `StatusLine` / `StatusLineNC`. So if `%#HL#` has only `guifg`, background is inferred from one reference highlight group.
- With the suggestion combining is done with whichever attributes are now "active". Background from previous case is inferred from whichever attributes are set previously. As initial attributes are the same as the one reference group from the previous case, I don't see how this will be more breaking.

ribru17:
> > imo this would be too breaking- seems like it would be impossible to just define regular, non cascading highlights (red then blue then green etc)
> 
> 
> 
> It is breaking in exactly the same way as this PR:

Not really... if I wanted a fancy pants statusline with maybe some strikethrough in one character then maybe underline at one part... you get the idea, the last characters of the statusline text would be a muddled bunch of colors/attributes and the more hl groups you have before it the more unwieldy things would get

echasnovski:
> Not really... if I wanted a fancy pants statusline with maybe some strikethrough in one character then maybe underline at one part... you get the idea, the last characters of the statusline text would be a muddled bunch of colors/attributes and the more hl groups you have before it the more unwieldy things would get

Aha, yeah, I can see the problem with `gui` / `cterm` attributes (underline, undercurl, etc.). However, this is not the case with foreground and background colors, as those are not defined as flags but as values.

I'd say colors (`fg`, `bg`, `sp`) cover vast majority of cases so combining them with what are "active" ones and not come from `WinBar` / etc.will still be hugely beneficial (and breaking to the same extent as this PR).

> And already you can always define highlights that inherit what you want and then just specify those custom ones in the statusline text

This same argument applies to the change in this PR, so not really a point.

ribru17:
> Aha, yeah, I can see the problem with `gui` / `cterm` attributes (underline, undercurl, etc.). However, this is not the case with foreground and background colors, as those are not defined as flags but as values.
> 
> I'd say colors (`fg`, `bg`, `sp`) cover vast majority of cases so combining them with what are "active" ones and not come from `WinBar` / etc.will still be hugely beneficial (and breaking to the same extent as this PR).

Personally I still disagree that this should be done (even if attributes are less widely used) but if the head honchos agree with you I will accept my fate :) It would be nice if there was just some "end delimiter" of the highlight group section, then I think nested highlight group specifications would be possible and one could achieve inheritance without arbitrary obliteration of rightmost text

echasnovski:
> It would be nice if there was just some "end delimiter" of the highlight group section, then I think nested highlight group specifications would be possible and one could achieve inheritance without arbitrary obliteration of rightmost text

There is `%*` which (in theory) can be made to force *exactly* the `WinBar` (`StatusLine`, etc.) group with all its attributes. While using `%#WinBar#` would inherit fg-bg-sp attributes.

chrisbra:
Guys, this change broke vim-airline.

justinmk:
The change is useful and intuitive. How difficult is it for plugins (vim-airline) to adapt to this change?

chrisbra:
I don't know yet, I just received a complaint

matveyt:
> The change is useful and intuitive. How difficult is it for plugins (vim-airline) to adapt to this change?

It is absolutely not. Say, default hl-StatusLine is "reverse". Then any custom color in the status line will be in reverse too. How is this intuitive?!

echasnovski:
> ... Say, default hl-StatusLine is "reverse". Then any custom color in the status line will be in reverse too. ...

Not if highlight group in question contains `gui=nocombine`. See `:h attr-list`.

matveyt:
@echasnovski Then one will need to use only "special nocombine" groups in statis line but never "a normal group", right?

echasnovski:
> @echasnovski Then one needs to use only "special nocombine" groups in status line but never "a normal group", right?

If one wants *any* highlight group to never combine its attributes with other highlight groups that affect same cell, then one needs to add `nocombine` attribute to it.

It was always like this and became more apparent here when the "base" highlight group with which attributes are combined has changed from `Normal` to a more relevant one. This question could have still appeared if user defined `gui=reverse` in `Normal` group.

matveyt:
No one ever sets "Normal" to be "reverse". While "StatusLine" is often reverse. Even by default until now.

I totally understand that the problem is (mainly) due to background inheritance. And that the Vim's legacy is flawed. Nonetheless, the medicine appears to be much worse than the illness.

PR #29989, opened 2024-08-06
vim-patch:9.1.0465: missing filecopy() function

zeertzjq:
#### vim-patch:9.1.0465: missing filecopy() function

Problem:  missing filecopy() function
Solution: implement filecopy() Vim script function
          (Shougo Matsushita)

closes: vim/vim#12346

https://github.com/vim/vim/commit/60c8743ab6c90e402e6ed49d27455ef7e5698abe

Co-authored-by: Shougo Matsushita <Shougo.Matsu@gmail.com>

### DIFF
--- a/src/nvim/fileio.c
+++ b/src/nvim/fileio.c
@@ -2654,7 +2654,6 @@ static int rename_with_tmp(const char *const from, const char *const to)
 int vim_rename(const char *from, const char *to)
   FUNC_ATTR_NONNULL_ALL
 {
-  char *errmsg = NULL;
   bool use_tmp_file = false;
 
   // When the names are identical, there is nothing to do.  When they refer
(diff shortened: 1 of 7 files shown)


src/nvim/fileio.c
```diff
@@ -2698,21 +2697,57 @@ int vim_rename(const char *from, const char *to)
...
   vim_acl_T acl = os_get_acl(from);
   int fd_in = os_open(from, O_RDONLY, 0);
   if (fd_in < 0) {
     os_free_acl(acl);
-    return -1;
+    return FAIL;
   }
 
   // Create the new file with same permissions as the original.
   int fd_out = os_open(to, O_CREAT|O_EXCL|O_WRONLY|O_NOFOLLOW, perm);
```
justinmk, line 2746: Do we gain anything if we use `os_copy(..., UV_FS_COPYFILE_FICLONE)` ?

https://github.com/neovim/neovim/blob/8df6736ca14d09f87cf0a8486758ac5708819434/src/nvim/bufwrite.c#L1715

zeertzjq: I'm not sure if `UV_FS_COPYFILE_FICLONE` should be used here, but `UV_FS_COPYFILE_EXCL` is needed. With `UV_FS_COPYFILE_EXCL`, `os_copy()` does seems to be a drop-in replacement here, and is faster and handles permissions better.

PR #29992, opened 2024-08-06
docs(treesitter): generate inline docs for `Range`s

ofseed:
The docs did not explain the internal type names like `Range`. I'm not sure if squash is appropriate as this contains several tiny fixes.


lewis6991 (approved):

clason:
@ofseed yes, I think it would be good to squash the `docs` commits (but keep the `feat` commit separate). Thanks!

ofseed:
I think it should be done now.

PR #30014, opened 2024-08-08
support emojis with ZWJ and variant selectors

### LINKED ISSUE #7151: Emoji modifiers are broken
- `nvim --version`: v0.2.0
- OSX 10.11.6
- iTerm 3.0.15
- `$TERM`: xterm-256color-italic

### Steps to reproduce using `nvim -u NORC`

```
nvim -u NORC
👍🏼
💪🏼
```

### Actual behaviour
screen shot 2017-08-11 at 10 47 28 am

### Expected behaviour
screen shot 2017-08-11 at 10 37 03 am

Emoji's are suddenly displaying odd. I'm not sure if this is a UTF-8 issue or issue specific to emoji but neovim seems to insist to display both characters that make up an emoji. When there is only one character (ie:👖) then neovim will display emptyspace after the emoji.

### LINKED ISSUE #22014: Flag emojis display as squares
### Describe the bug

Given this text file as `t.txt`:

```
* 🇺🇸 
* 🇰🇷 
```

on mac ventura, neovim displays the flag emojis incorrectly on `iterm2` and `terminal.app`, but correctly on `kitty`.

`iterm` and `terminal.app` themselves have no trouble displaying the emoji.

`nvim -u none t.txt` (v0.8.2 from homebrew) on `iterm` (3.4.19, the latest as of right now), looks like this:

image

while `cat t.txt` shows it properly:

image

`nvim -u none t.txt` on `terminal.app` looks similarly:

image

but on kitty (0.26.5), it works properly:

image

I tried a few `TERM` values, as I noticed kitty had `TERM=xterm-kitty`, but nvim on iterm failed with `TERM` set to all of: `ansi-256color`, `xterm-256color`, `xterm-kitty` and `xterm-new`

I reproduced this behavior on a second macbook with the same versions of these softwares.

All of these programs also display the flags correctly in iterm: `bat`, `nano`, `emacs`, `helix`

`TextEdit`, `code` and `VimR` do as well

### Steps to reproduce

1. edit a file with neovim 0.8.2 that contains flag emojis on mac ventura, with iterm 3.4.19

### Expected behavior

I expect it to show flag emojis

### Neovim version (nvim -v)

NVIM v0.8.2

yes, version 9.0

### Operating system/version

macOS 13.1 ventura

### Terminal name/version

iterm 3.4.19, see above for further details

xterm-256color

### Installation

homebrew

bfredl:
The implementation of grapheme clusters   was upgraded to closely follow extended grapheme clusters as defined by UAX#29  in the unicode standard. Noteworthily, this enables proper display of many  more emoji characters than before, including those encoded with multiple
emoji codepoints combined with ZWJ (zero width joiner) codepoints and variant selectors.

Fix #7151
Fix #22014


test/old/testdir/test_normal.vim
```diff
@@ -3905,7 +3905,7 @@ func Test_normal_gj_on_extra_wide_char()
   put =text
   call cursor(2,1)
   norm! gj
-  call assert_equal([0,2,25,0], getpos('.'))
+  call assert_equal([0,2,29,0], getpos('.'))
```
zeertzjq, line 3908: This test is for unprintable character. When ZWJ is no longer unprintable some other unprintable character between 0x1000 and 0xFFFF should be used.

test/old/testdir/test_normal.vim
```diff
@@ -3897,9 +3897,9 @@ func Test_normal_count_after_operator()
   bw!
 endfunc
 
-func Test_normal_gj_on_extra_wide_char()
+func Test_normal_gj_on_unprintable_char()
```
zeertzjq, line 3900: "extra wide" refers to the unprintable char taking 6 cells. "unprintable" alone doesn't necessarily mean what.

bfredl: "extra wide" absolutely does not imply that something is at least 6 cells wide. 
"unprintable" refers to a documented, pre-existing thing which is the subject of this test.

if the exact width of 6 is important I will add a comment.

runtime/doc/mbyte.txt
```diff
@@ -646,6 +646,12 @@ widespread as file format.
 A composing or combining character is used to change the meaning of the
 character before it.  The combining characters are drawn on top of the
 preceding character.
+
+Nvim largely follows the definition of extended grapheme clusters in UAX#29
+in the unicode standard, with some modifications. An ascii char will always
```
clason, line 651:
```suggestion
in the unicode standard, with some modifications: An ascii char will always
```

runtime/doc/mbyte.txt
```diff
@@ -646,6 +646,12 @@ widespread as file format.
...
+start a new cluster. In addition 'arabicshape' enables the combining of some
+arabic letters, when they are shaped to be displayed togheter in a single cell.
```
clason, line 653:
```suggestion
start a new cluster. In addition, 'arabicshape' enables the combining of some
arabic letters when they are shaped to be displayed together in a single cell.
```

runtime/doc/options.txt
```diff
@@ -2217,9 +2217,12 @@ A jump table for the options with a short description can be found at |Q_op|.
 			global
 	When on all Unicode emoji characters are considered to be full width.
 	This excludes "text emoji" characters, which are normally displayed as
-	single width.  Unfortunately there is no good specification for this
-	and it has been determined on trial-and-error basis.  Use the
-	|setcellwidths()| function to change the behavior.
+	single width. However, Such "text emoji" are treated as full-width
```
clason, line 2220:
```suggestion
	single width. However, such "text emoji" are treated as full-width
```

src/nvim/mbyte.c
```diff
@@ -793,29 +812,45 @@ int mb_cptr2char_adv(const char **pp)
...
-bool utf_composinglike(const char *p1, const char *p2)
+/// comes after "p1".
+///
+/// We use the definition in UAX#29 as implemented by utf8proc with some
+/// exceptions:
+///
+/// - ASCII chars always begin a new cluster. This is a long assumed invariant
+///   in the code base and very useful for performance (we can exit early for ASCII
+///   all over the place, branch predictor go brrr in ascii-only text).
+///   as of unicode 15.1 this will only break BOUNDCLASS_PREPEND followed by ascii,
```
clason, line 829:
```suggestion
///   As of unicode 15.1 this will only break BOUNDCLASS_PREPEND followed by ascii,
```

src/nvim/mbyte.c
```diff
@@ -793,29 +812,45 @@ int mb_cptr2char_adv(const char **pp)
...
+///   which should be exceedingly rare (these PREPEND chars are expected to be
+///   followed by multibyte chars within the same script family)
+///
+/// - When 'arabicshape' is active, some pairs of arabic leters "ab" is replaced with
```
clason, line 833:
```suggestion
/// - When 'arabicshape' is active, some pairs of arabic letters "ab" is replaced with
```

test/functional/ui/multibyte_spec.lua
```diff
@@ -381,15 +473,28 @@ describe('multibyte rendering: statusline', function()
...
+      {3:xx}{9:🧑‍💻}{3:yy                                  }|
+                                              |
+    ]],
+    }
+  end)
+
+  it('unprintable chars in filename with custom stl', function()
+    command('set statusline=xx%#ErrorMsg#%f%##yy')
+    command('file 🧑​💻')
     -- TODO: this is also wrong but also avoids a crash
```
zeertzjq, line 492:
```suggestion
```

runtime/doc/mbyte.txt
```diff
@@ -646,6 +646,12 @@ widespread as file format.
 A composing or combining character is used to change the meaning of the
 character before it.  The combining characters are drawn on top of the
 preceding character.
+
+Nvim largely follows the definition of extended grapheme clusters in UAX#29
+in the Unicode standard, with some modifications: An ascii char will always
+start a new cluster. In addition 'arabicshape', enables the combining of some
+arabic letters, when they are shaped to be displayed toghether in a single cell.
```
zeertzjq, line 653:
```suggestion
arabic letters, when they are shaped to be displayed together in a single cell.
```

src/nvim/mbyte.c
```diff
@@ -996,9 +1031,10 @@ int utfc_ptr2len_len(const char *p, int size)
     return 1;
   }
 
-  // Check for composing characters.  We can handle only the first six, but
+  // Check for composing characters.  we can only display a limited amount, but
```
zeertzjq, line 1034:
```suggestion
  // Check for composing characters.  We can only display a limited amount, but
```

src/nvim/mbyte.c
```diff
@@ -1679,50 +1736,107 @@ int utf_head_off(const char *base_in, const char *p_in)
...
+  return 0;
+}
+
+/// Assumes caller already handles ascii. see `utfc_next`
+StrCharInfo utfc_next_impl(StrCharInfo cur)
+{
+  int32_t prev_code = cur.chr.value;
+  uint8_t *next = (uint8_t *)(cur.ptr + cur.chr.len);
+  GraphemeState state = GRAPHEME_STATE_INIT;
+  assert(*next >= 128);
```
zeertzjq, line 1818: Using `0x80` here seems more consistent with similar functions
```suggestion
  assert(*next >= 0x80);
```

src/nvim/plines.c
```diff
@@ -146,7 +146,7 @@ CharSize charsize_regular(CharsizeArg *csarg, char *const cur, colnr_T const vco
   } else if (cur_char < 0) {
     size = kInvalidByteCells;
   } else {
-    size = char2cells(cur_char);
+    size = ptr2cells(cur);
```
zeertzjq, line 149: This kind of makes it pointless to pass in `cur_char`, as `utf_ptr2cells()` already handles illegal bytes, and the `cur_char > 0x80` check can be replaced with `MB_BYTE2LEN(*cur) > 1`.

Or the logic in `utf_ptr2cells()` can be replicated here without the first `utf_ptr2char()` to avoid decoding first char twice, but then that will also require passing in `ci.chr.len`, so not sure if that's worth it.

src/nvim/plines.c
```diff
@@ -352,7 +352,7 @@ static inline CharSize charsize_fast_impl(win_T *const wp, bool use_tabstop, col
     if (cur_char < 0) {
       width = kInvalidByteCells;
     } else {
-      width = char2cells(cur_char);
+      width = ptr2cells(cur);
```
zeertzjq, line 355: ditto

src/nvim/plines.c
```diff
@@ -371,23 +371,23 @@ static inline CharSize charsize_fast_impl(win_T *const wp, bool use_tabstop, col
...
-int charsize_nowrap(buf_T *buf, bool use_tabstop, colnr_T vcol, int32_t cur_char)
+int charsize_nowrap(buf_T *buf, const char *cur, bool use_tabstop, colnr_T vcol, int32_t cur_char)
 {
   if (cur_char == TAB && use_tabstop) {
     return tabstop_padding(vcol, buf->b_p_ts, buf->b_p_vts_array);
   } else if (cur_char < 0) {
     return kInvalidByteCells;
   } else {
-    return char2cells(cur_char);
+    return ptr2cells(cur);
```
zeertzjq, line 390: ditto

src/nvim/plines.c
bfredl: ye, multiple decoding also happens in other places like the main win_line() loop. we probably want a specialized version of CharInfo which also includes the ptr2cells() width calculated at the same time as the byte length. Although I am thinking of that as a follow-up perf PR while only focusing on correctness (and no larger regressions) in this PR.

GitMurf:
@bfredl sorry to ping you but I have been searching everywhere to try and figure out a solution to my problem and I believe it is similar to the problem(s) you were aiming to fix with this PR. I posted the repro and details in a post in the Neovim reddit here: https://www.reddit.com/r/neovim/comments/1f6z9da/help_with_1_keycap_digit_1_emoji_sequence_with/

I am happy to give you more details or move this conversation somewhere else if you prefer, but the TLDR is that the emoji 1️⃣ (and the other similar numbers 2-9) are having problems and I believe it is due to the multiple code points. It is comprised of `U+31 + U+FE0F + U+20E3` and doing the `str2list` returns: `{ 49, 65039, 8419 }`. Thanks in advance!

zeertzjq:
That probably can't be fixed due to performance reasons.

GitMurf:
> That probably can't be fixed due to performance reasons.

@zeertzjq thanks for the quick reply! Is there some sort of "fallback" workaround that I could implement in my config? Like an autocmd that would render emojis like these to a broken icon emoji (something I would choose)? My team uses these number emojis a lot in comments so it is not an option for me to just remove. But I am totally fine if I just replace emojis like these to compatible ones and would just create a mapping list and add to it anytime these pop up. I'm just not sure if there is a reasonable way to do this (presumably an autocmd)? Thanks!!

bfredl:
we could at least mark anything + 0xFE0F as having ambiguous terminal width ( `utf_ambiguous_width`) . That would fix rendering of the rest of the line getting out of sync, even if it is not having the correct width (which is hard to encode as `setcellwidths()` does not handle clusters by design).

GitMurf:
> we could at least mark anything + 0xFE0F as having ambiguous terminal width ( utf_ambiguous_width) . That would fix rendering of the rest of the line getting out of sync

Thanks @bfredl ! That sounds great to me! My issue is not the display of the emoji itself but the fact it throws the rest of the line (and often surrounding lines) off. A couple questions:

1. are you able to confirm that my scenario is actually a problem (as opposed to just being an issue on my side)? Like does it make sense that I am having a problem with this 1️⃣ emoji?
2. can you think of any temporary workaround for the time being that I can apply on my end? I assume the `utf_ambiguous_width` is something you have to do in neovim core and not something I can do on my end?

Thanks so much for the quick response!

clason:
PSA: _Please_ don't leave tangential comments on (especially) on a (merged) PR! If you have a problem, open an issue (yes, that means filling out the template. it's annoying but there for a reason!).

That would also allow a PR to be linked to it, so you wouldn't have missed https://github.com/neovim/neovim/pull/30232.

PR #30028, opened 2024-08-12
fix(lsp): avoids an additional request.

glepnir:
Problem: if result.item is empty there still invoke complete function.

Solution: when matches is empty no need invoke complete.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -403,6 +403,10 @@ local function trigger(bufnr, clients)
   reset_timer()
   Context:cancel_pending()
 
+  if tonumber(vim.fn.pumvisible()) == 1 and Context.isIncomplete then
+    return
+  end
+
   local win = api.nvim_get_current_win()
   local cursor_row, cursor_col = unpack(api.nvim_win_get_cursor(win)) --- @type integer, integer
   local line = api.nvim_get_current_line()


mfussenegger (approved):
Does invoking complete cause a user-observable issue? If so - what kind?

I vaguely remember that calling it with empty matches was important in nvim-lsp-compl, but that might have been the case due to the different fuzzy implementation (which set `equal=1` on completion items) so I think the change here should be okay

glepnir:
When passing empty matches to complete this will immediately clear the current popupmenu. I guess the problem is here?

mfussenegger:
> When passing empty matches to complete this will immediately clear the current popupmenu.

Isn't that expected behavior if the server doesn't return any matches?

glepnir:
I have a completion wrokaround snippet for enable completion on characters and auto completion when prev character in triggerCharacters  .  

```lua
local function debounce(fn, delay)
  local timer = nil ---[[uv_timer_t]]
  local function safe_close()
    if timer and timer:is_active() and not timer:is_closing() then
      timer:stop()
      timer:close()
      timer = nil
    end
  end
  return function(...)
    local args = { ... }
    safe_close()
    timer = assert(uv.new_timer())
    timer:start(
      delay,
      0,
      vim.schedule_wrap(function()
        safe_close()
        xpcall(function()
          fn(args)
        end, function(err)
          vim.notify('Error in debounced trigger function ' .. err, vim.log.levels.ERROR)
        end)
      end)
    )
  end
end

-- hack can completion on any triggerCharacters
local function auto_trigger(bufnr)
  local debounced_trigger = debounce(completion.trigger, 100)
  au(TextChangedI, {
    buffer = bufnr,
    callback = function(args)
      local clients = lsp.get_clients({ bufnr = args.buf, method = ms.textDocument_completion })
      if #clients == 0 then
        return
      end
      vim.iter(clients):any(function(client)
          client,
          'server_capabilities',
          'completionProvider',
          'triggerCharacters'
        ) or {}
        if has_word_before(triggerchars) then
          debounced_trigger()
          return true
        end
        return false
      end)
    end,
  })
end

au('LspAttach', {
  callback = function(args)
    local bufnr = args.buf
    local client_id = args.data.client_id
    completion.enable(true, client_id, bufnr, { autotrigger = true })
    auto_trigger(bufnr)
  end,
})
```

Eg:  i have a variable named hashmap in C , type `h` textchagnedi callback invoke then there show a pum with some completion items.  then type `p`  pum is disappeared.  because `autotrigger = true` register  InsertCharePre then `p` also trigger it and pumvisible is true then i invoke trigger again.  `hp` get an empty lsp result then complete function is set with an empty matches. 

https://github.com/neovim/neovim/blob/9768e88f3891b5594eb2d556bd3bdf40c61d46e1/runtime/lua/vim/lsp/completion.lua#L461-L473

glepnir:
I could also implement InsertCharPre myself instead of autotrigger. So not sure.

glepnir:
nvm ...I think current is correct ..

glepnir:
I thought about it again. If multiple lsp clients return completion, the first client returns the completion, and the second client returns nothing. Doesn't this clear the current pum?

mfussenegger:
> I thought about it again. If multiple lsp clients return completion, the first client returns the completion, and the second client returns nothing. Doesn't this clear the current pum?

The results from all clients are merged before it calls `fn.complete` so this shouldn't be a problem.

>  because autotrigger = true register InsertCharePre then p also trigger it and pumvisible is true then i invoke trigger again. hp get an empty lsp result then complete function is set with an empty matches.

I'm a bit confused about this. If you trigger completion manually then the combination of checking `pumvisible()` and `Context.isIncomplete` should avoid the autotrigger from triggering completion again because `isIncomplete` should be false. It should hit the early return here:

https://github.com/neovim/neovim/blob/40bc48479dca0d4a02c6445ceccab700847dcbe9/runtime/lua/vim/lsp/completion.lua#L475

glepnir:
Nope `pumvisible()` will be 0 in TextchangedI . the trigger time is different.  I use InsertcharPre instead of Textchangedi so it works well on my local agian. hp can also get fuzzy result in pum. So do you think there still need change?

mfussenegger:
I see. If I understand it right then maybe we could do something like this instead:

```diff
diff --git a/runtime/lua/vim/lsp/completion.lua b/runtime/lua/vim/lsp/completion.lua
index 6a6659fd2..eea914b27 100644
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -400,11 +400,13 @@ local function request(clients, bufnr, win, callback)
 end
 
 local function trigger(bufnr, clients)
   reset_timer()
   Context:cancel_pending()
-
+  if tonumber(vim.fn.pumvisible()) == 1 and Context.isIncomplete then
+    return
+  end
   local win = api.nvim_get_current_win()
   local cursor_row, cursor_col = unpack(api.nvim_win_get_cursor(win)) --- @type integer, integer
   local line = api.nvim_get_current_line()
   local line_to_cursor = line:sub(1, cursor_col)
   local word_boundary = vim.fn.match(line_to_cursor, '\\k*$')
```

This would have the advantage that it avoids an additional request.

Otherwise I'd tend to keep things as they are. Removing the popup window if there are no more matches - which can be the case for `isIncomplete=true` cases - is imho the right thing to do.

glepnir:
make sense

runtime/lua/vim/lsp/completion.lua
```diff
@@ -403,6 +403,10 @@ local function trigger(bufnr, clients)
   reset_timer()
   Context:cancel_pending()
 
+  if tonumber(vim.fn.pumvisible()) == 1 and Context.isIncomplete then
```
mfussenegger, line 406: Ah, just realized I was too quick with the merge. This should be `not Context.isIncomplete`, no?
If the previous result was incomplete it must fire another request to fetch more candidates.

glepnir: i mixedup the var name with boolean value T.T

PR #30044, opened 2024-08-14
fix(lsp): when prefix is non word add all result into matches

glepnir:
Problem: prefix can be a symbol like period, the fuzzy mathes can't handle it correctly.

Solution: when prefix is empty or a symbol add all lsp completion result into matches.


glepnir:
@mfussenegger  PTAL

PR #30047, opened 2024-08-14
refactor: adopt termkey and eliminate duplicate code

dundargoc:
Termkey is abandoned and it's now our code, so there's no reason not to
treat it as such. An alternative approach could be to have a proper repo
that we maintain such as with unibilium, although with this approach we
can make a few assumptions that will allow us to remove more code.

Also eliminate duplicate code from both termkey and libvterm.


bfredl:
should we maybe place the files under `src/nvim/tui/` if it is just part of the TUI and not seen as a separate library?

dundargoc:
Note that the code removal is more dramatic than the diffstats might indicate. A lot of lines were added because braces were added when formatting.

src/vterm/screen.c
```diff
@@ -2,9 +2,11 @@
 
 #include <stdio.h>
 #include <string.h>
+#include "nvim/mbyte.h"
+#include "nvim/tui/termkey.h"
 
 #include "rect.h"
-#include "utf8.h"
+#undef small
```
clason, line 9: Isn't that already `undef`ed in our `win_defs.h`?

dundargoc: Yeah, but I assume the problematic windows header is being included somewhere else. I suspect libuv. I'll just rename all variables named "small" to "little" or something instead, this is annoying to guard against.

clason: "tiny"

dundargoc: OK, found a workaround. `WIN32_LEAN_AND_MEAN` only includes core set of headers and not kitchen sink, so if CI passes with that let's go with that.

zeertzjq:
You are changing libvterm too

src/nvim/tui/termkey_defs.h
```diff
@@ -0,0 +1,228 @@
...
+  OptInt ttimeoutlen;
+  TermKey *tk;
+  TermKey_Terminfo_Getstr_Hook *tk_ti_hook_fn;  ///< libtermkey terminfo hook
+  uv_timer_t timer_handle;
+  Loop *loop;
+  RStream read_stream;
+  TUIData *tui_data;
+  char key_buffer[KEY_BUFFER_SIZE];
+  size_t key_buffer_len;
+} TermInput;
```
zeertzjq, line 217: Why are `KeyEncoding` and `TermInput` moved to this file?

clason:
> You are changing libvterm too

Yes, because there is shared code, and shared problems.

src/nvim/tui/termkey_defs.h
dundargoc: Fix clint warnings. I can revert and include termkey_defs instead if you prefer

dundargoc: Actually keyencoding has to be there to silence clint warning. So this is correct.

zeertzjq: What clint warning? There are no clint warnings on [`c180eef` (#30047)](https://github.com/neovim/neovim/pull/30047/commits/c180eef87187ec5b1b85b8bc831450d3665b8f1a)

dundargoc:
> You are changing libvterm too

Ye, I figured I could deal with the low hanging fruits from there as well. Should I postpone it to a future PR?

src/nvim/tui/termkey_defs.h
zeertzjq: Why? I don't see it used in any other header.

dundargoc: Must've looked at the wrong place or something. Fixed.

clason:
Note for later: as we're taking ownership of `vterm` by integrating it more tightly with Neovim utf8 functionality (to guarantee that core and embedded terminal have the same idea of, say, character width), we will move it to `src/nvim/term`, similarly to the `termkey` move here. The current placement is just a staging ground for that until we start to make bigger changes.

clason:
Not sure we want to yeet all debug output? That can be helpful.

dundargoc:
> Not sure we want to yeet all debug output? That can be helpful.

Yeah I changed my mind and reverted it.

src/nvim/tui/driver-csi.c
```diff
@@ -131,17 +130,16 @@ static TermKeyResult handle_csifunc(TermKey *tk, TermKeyKey *key, int cmd, TermK
     int mod = key->modifiers;
     (*tk->method.emit_codepoint)(tk, args[2], key);
     key->modifiers |= mod;
-  }
-  else if(args[0] >= 0 && args[0] < NCSIFUNCS) {
+  } else if (args[0] >= 0 && (size_t)args[0] < NCSIFUNCS) {
```
zeertzjq, line 133: This cast is not needed anymore
```suggestion
  } else if (args[0] >= 0 && args[0] < NCSIFUNCS) {
```

src/nvim/tui/driver-csi.c
```diff
@@ -153,7 +151,7 @@ static TermKeyResult handle_csifunc(TermKey *tk, TermKeyKey *key, int cmd, TermK
 
 static void register_csifunc(TermKeyType type, TermKeySym sym, int number)
 {
-  if(number >= NCSIFUNCS) {
+  if ((size_t)number >= NCSIFUNCS) {
```
zeertzjq, line 154: ditto
```suggestion
  if (number >= NCSIFUNCS) {
```

src/nvim/tui/driver-csi.c
```diff
@@ -547,14 +559,15 @@ static int register_keys(void)
...
-  for(i = 0; i < 64; i++) {
+  for (i = 0; i < 64; i++) {
     csi_ss3s[i].sym = TERMKEY_SYM_UNKNOWN;
-    ss3s[i].sym     = TERMKEY_SYM_UNKNOWN;
+    ss3s[i].sym = TERMKEY_SYM_UNKNOWN;
     ss3_kpalts[i] = 0;
   }
 
-  for(i = 0; i < NCSIFUNCS; i++)
+  for (i = 0; (size_t)i < NCSIFUNCS; i++) {
```
zeertzjq, line 568: ditto
```suggestion
  for (i = 0; i < NCSIFUNCS; i++) {
```

src/nvim/tui/driver-csi.c
```diff
@@ -632,15 +646,18 @@ static int register_keys(void)
...
       return NULL;
+    }
+  }
 
-  TermKeyCsi *csi = malloc(sizeof *csi);
-  if(!csi)
+  TermKeyCsi *csi = xmalloc(sizeof *csi);
+  if (!csi) {
     return NULL;
+  }
```
zeertzjq, line 660:
```suggestion
```

src/nvim/tui/driver-ti.c
```diff
@@ -125,126 +119,116 @@ struct trie_node_key {
...
 static int insert_seq(TermKeyTI *ti, const char *seq, struct trie_node *node);
 
 static struct trie_node *new_node_key(TermKeyType type, TermKeySym sym, int modmask, int modset)
 {
-  struct trie_node_key *n = malloc(sizeof(*n));
-  if(!n)
+  struct trie_node_key *n = xmalloc(sizeof(*n));
+  if (!n) {
     return NULL;
+  }
```
zeertzjq, line 133:
```suggestion
```

src/nvim/tui/driver-ti.c
```diff
@@ -125,126 +119,116 @@ struct trie_node_key {
...
 }
 
 static struct trie_node *new_node_arr(unsigned char min, unsigned char max)
 {
-  struct trie_node_arr *n = malloc(sizeof(*n) + ((int)max-min+1) * sizeof(n->arr[0]));
-  if(!n)
+  struct trie_node_arr *n = xmalloc(sizeof(*n) + (max - min + 1) * sizeof(n->arr[0]));
+  if (!n) {
     return NULL;
+  }
```
zeertzjq, line 150:
```suggestion
```

src/nvim/tui/driver-ti.c
```diff
@@ -278,288 +261,260 @@ static int load_terminfo(TermKeyTI *ti)
...
 
-static void *new_driver(TermKey *tk, const char *term)
+void *new_driver_ti(TermKey *tk, const char *term)
 {
-  TermKeyTI *ti = malloc(sizeof *ti);
-  if(!ti)
+  TermKeyTI *ti = xmalloc(sizeof *ti);
+  if (!ti) {
     return NULL;
+  }
```
zeertzjq, line 368:
```suggestion
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1333 @@
...
+  *strcamelp = strcamel;
+  return *str - *strcamel;
+}
+
+static TermKey *termkey_alloc(void)
+{
+  TermKey *tk = xmalloc(sizeof(TermKey));
+  if (!tk) {
+    return NULL;
+  }
```
zeertzjq, line 318:
```suggestion
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1333 @@
...
+  return tk;
+}
+
+static int termkey_init(TermKey *tk, const char *term)
+{
+  tk->buffer = xmalloc(tk->buffsize);
+  if (!tk->buffer) {
+    return 0;
+  }
+
```
zeertzjq, line 366:
```suggestion
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1333 @@
...
+  tk->keynames = xmalloc(sizeof(tk->keynames[0]) * (size_t)tk->nkeynames);
+  if (!tk->keynames) {
+    goto abort_free_buffer;
+  }
```
zeertzjq, line 370:
```suggestion
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1333 @@
...
+    (*p->driver->free_driver)(p->info);
+    struct TermKeyDriverNode *next = p->next;
+    xfree(p);
+    p = next;
+  }
+
+abort_free_keynames:
+  xfree(tk->keynames);
+
+abort_free_buffer:
```
zeertzjq, line 439:
```suggestion
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1333 @@
...
+{
+  return tk->buffsize;
+}
+
+int termkey_set_buffer_size(TermKey *tk, size_t size)
+{
+  unsigned char *buffer = xrealloc(tk->buffer, size);
+  if (!buffer) {
+    return 0;
+  }
```
zeertzjq, line 624:
```suggestion
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1333 @@
...
+{
+  if (!sym) {
+    sym = tk->nkeynames;
+  }
+
+  if (sym >= tk->nkeynames) {
+    const char **new_keynames = xrealloc(tk->keynames, sizeof(new_keynames[0]) * ((size_t)sym + 1));
+    if (!new_keynames) {
+      return -1;
+    }
```
zeertzjq, line 1085:
```suggestion
```

src/nvim/tui/termkey_defs.h
```diff
@@ -0,0 +1,199 @@
+#pragma once
```
justinmk, line 1:
```suggestion
// Adapted from ...

#pragma once
```

src/nvim/tui/termkey.c
```diff
@@ -0,0 +1,1315 @@
+#include <ctype.h>
```
justinmk, line 1:
```suggestion
// Adapted from ...

#include <ctype.h>
```

justinmk:
Currently this  PR mixes all the files together, e.g. `src/nvim/tui/driver-csi.c` lives next to other Nvim TUI files. Whereas previously, termkey was at least somewhat "abstracted", it was a self-contained thing. Is it not useful to continue to treat it like a self-contained thing in some respect, perhaps by putting it in a subdirectory or using a common prefix for all the files? Would that keep the door open for migrating to something else in the future (such as libtikit or similar)?

dundargoc:
Moved the code to `tui/termkey`. Kinda starting to wonder if the PR is worth it tho if we want to keep the two codes separated.

clason:
I think (slowly) moving utf8-related code to be in sync with Neovim is absolutely worth it. Stylistic polish and dead code removal is low-effort/low-reward by comparison, though (albeit not unwelcome).

clason:
@dundargoc rebase, and possibly look into using more utf8 code after bfredls bobbytablectomy (not necessarily in this PR)? Would like to move forward with (slowly) cleaning up the inherited code.

MAINTAIN.md
```diff
@@ -160,7 +160,6 @@ These dependencies are "vendored" (inlined), we must update the sources manually
     * Needs to be updated when LPeg is updated.
 * `src/bit.c`: only for PUC lua: port of `require'bit'` from luajit https://bitop.luajit.org/
 * `runtime/lua/coxpcall.lua`: coxpcall (only needed for PUC lua, builtin to luajit)
-* `src/termkey`: [libtermkey](https://github.com/neovim/libtermkey)
```
justinmk, line 163: why not list it here still? even if the upstream is frozen, we still vendored it and it's useful to have a reference list of all the things we vendored. we still list unibilium too.

justinmk (approved):

MAINTAIN.md
zeertzjq: Unibilium is not vendered

dundargoc: > is no longer true: the sources now live in-tree; there will no longer be an external repo from which we need to pull updates. (The repo is already archived.)

That was pretty much my thought process. It's not something I'm for or against either way so I don't mind re-adding it.

clason:
@dundargoc do you want to squash (some of) the commits?

justinmk:
coverity:

```
*** CID 509228:  Insecure data handling  (INTEGER_OVERFLOW)
/src/nvim/tui/termkey/driver-ti.c: 413 in start_driver_ti()
407       }
408     #endif
409     
410       // Can't call putp or tputs because they suck and don't give us fd control
411       len = strlen(start_string);
412       while (len) {
>>>     CID 509228:  Insecure data handling  (INTEGER_OVERFLOW)
>>>     "(unsigned int)len", which might have underflowed, is passed to "write(tk->fd, start_string, (unsigned int)len)".
413         size_t written = (size_t)write(tk->fd, start_string, (unsigned)len);
414         if (written == (size_t)-1) {
415           return 0;
416         }
417         start_string += written;
418         len -= written;

________________________________________________________________________________________________________
*** CID 509227:  Insecure data handling  (INTEGER_OVERFLOW)
/src/nvim/tui/termkey/driver-ti.c: 451 in stop_driver_ti()
445       // The terminfo database will contain keys in application cursor key mode.
446       // We may need to enable that mode
447     
448       // Can't call putp or tputs because they suck and don't give us fd control
449       len = strlen(stop_string);
450       while (len) {
>>>     CID 509227:  Insecure data handling  (INTEGER_OVERFLOW)
>>>     "(unsigned int)len", which might have underflowed, is passed to "write(tk->fd, stop_string, (unsigned int)len)".
451         size_t written = (size_t)write(tk->fd, stop_string, (unsigned)len);
452         if (written == (size_t)-1) {
453           return 0;
454         }
455         stop_string += written;
456         len -= written;
```

PR #30050, opened 2024-08-14
feat(term): support OSC 8 hyperlinks in :terminal

### LINKED ISSUE #28789: Render hyperlinks (OSC 8) in embedded terminal
### Problem

Make hyperlinks in command outputs clickable in the embedded terminal (:term) in neovim if the hosting terminal or UI supports it.

Commands such as ls --hyperlink=always (GNU coreutils) and ripgrep are examples of utils that can output hyperlink escapes.

### Expected behavior

ls -l --hyperlink=always should produce clickable links when used inside neovim's embedded terminal (:term). Compare with how this works in standalone terminals (such as gnome-terminal, kitty, or others, also in tmux) which make these links clickable. Use the existing neovim 0.10 support for OSC 8 (in extmarks in that case) to accomplish this.

### Discussion

Briefly discussed here https://github.com/neovim/neovim/discussions/28620#discussioncomment-9459364 where @gpanders shared a thought

gpanders:
Fixes #28789.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -166,6 +166,8 @@ TERMINAL
 • The terminal buffer now supports reflow (wrapped lines adapt when the buffer
   is resized horizontally). Note: Lines that are not visible and kept in
   'scrollback' are not reflown.
+• The |terminal| now supports OSC 8 escape sequences and will display
+  hyperlinks in supporting host terminals.
 
 TREESITTER
 
--- a/src/nvim/terminal.c
+++ b/src/nvim/terminal.c
@@ -227,11 +227,66 @@ static void schedule_termrequest(Terminal *term, char *payload, size_t payload_l
                  term->pending_send);
 }
 
+static int parse_osc8(VTermStringFragment frag, int *attr)
+  FUNC_ATTR_NONNULL_ALL
+{
+  // Parse the URI from the OSC 8 sequence and add the URL to our URL set.
+  // Skip the ID, we don't use it (for now)
+  size_t i = 0;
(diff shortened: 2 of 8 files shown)


zeertzjq:
Add a test?

bluss:
Thank you!! :tada:

runtime/doc/news.txt
```diff
@@ -166,6 +166,8 @@ TERMINAL
 • The terminal buffer now supports reflow (wrapped lines adapt when the buffer
   is resized horizontally). Note: Lines that are not visible and kept in
   'scrollback' are not reflown.
+• The |terminal| now supports OSC 8 escape sequences and will display
+  hyperlinks in supporting host terminals.
```
justinmk, line 170: Can `gx` take advantage of this in :terminal buffers?

gpanders: `gx` doesn’t currently use the `url` extmark attribute but it certainly could!

src/nvim/terminal.c
```diff
@@ -227,19 +227,74 @@ static void schedule_termrequest(Terminal *term, char *payload, size_t payload_l
...
+  }
+
+  const size_t len = i - start;
+  if (len == 0) {
+    // Empty OSC 8, no URL
+    *attr = 0;
+    return 1;
+  }
+
+  char *url = xmemdupz(&frag.str[start], len + 1);
```
zeertzjq, line 265: This is reading 1 byte over the end of `frag.str`. The `+ 1` shouldn't be here.

PR #30053, opened 2024-08-15
revert: "fix(man): handle absolute paths as `:Man` targets (#20624)"

zeertzjq:
This reverts commit 39911d76be560c998cc7dee51c5d94f811164f66.

Restores support for the :Man {path} mentioned in the help.


zeertzjq:
Oh, I see the problem.

Issue #29792, opened 2024-07-18
Add LSP client flag to pull diagnostics on save


Delerme:
### Problem

Now that neovim 10 supports pull diagnostics, I think it'd be useful to expose some additional flags to control how often diagnostic requests are sent.

Right now, diagnostics are requested after each `didChange` event, which can be controlled partially via existing lsp client flags like `debounce_text_changes`: https://neovim.io/doc/user/lsp.html#_lua-module:-vim.lsp.client

Some lsp servers like eslint expose options to run diagnostics on save, since they can be a little slow. [With pull diagnostics, those settings have to be on the client-side to function](https://github.com/neovim/nvim-lspconfig/issues/3211).

### Expected behavior

To start, I propose adding a new flag `pull_diagnostics_on_save` that sends diagnostic requests after `didSave` instead of `didChange` for that lsp client.

Any thoughts or concerns? I'm happy to try implementing if it sounds like a reasonable idea!

Delerme:
cc @catlee for any thoughts or guidance, or to catch any mistakes in my understanding. Thanks for adding pull diagnostic support!

catlee:
Yes, I think it makes sense to be able to configure the frequency of when diagnostics are requested per client or to change it to be on save only.

According to https://github.com/neovim/neovim/blob/7031949be065870f0daf74a9f1be3df47f83312c/runtime/doc/lsp.txt#L1027, you can already configure the debounce timer per client.

MariaSolOs:
This sounds like a good idea. @Delerme feel free to start working on a draft implementation!

MariaSolOs:
@justinmk @mfussenegger based on the conversation from https://github.com/neovim/neovim/pull/30055, it seems like we've agreed that this behavior can be achieved today by calling `vim.diagnostic.enable` in autocommands. I guess we can close this?

thatsmydoing:
I don't believe that's the case. I think what they were referring to is `vim.lsp.diagnostic._enable` which is accessible but it's technically private. `vim.diagnostic.enable` just turns off the display of the diagnostics but doesn't prevent them from being requested (which is what we want to stop).

liskin:
[...]

PR #30055, opened 2024-08-15
feat(lsp): support pull diagnostics on save via flag

Delerme:
Problem:
Some lsp servers can be configured to refresh diagnostics on save to avoid performance issues. The introduction of pull diagnostics in 0.10 goes around those settings by requesting diagnostics with every change event. See #29792

Solution:
Add an optional lsp client flag to request diagnostics on save if the lsp server supports pull diagnostics. By default, the existing behavior is maintained and diagnostics are requested with every change event, if the server supports pull diagnostics.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1033,6 +1033,11 @@ Lua module: vim.lsp.client                                        *lsp-client*
                                   `150`) Debounce `didChange` notifications to
                                   the server by the given number in
                                   milliseconds. No debounce occurs if `nil`.
+                                • {pull_diagnostics_on_save} (`boolean`,
+                                  default: `false`) Request diagnostics from
+                                  the server on `didSave` events instead of
+                                  `didChange` events. Has no effect if the
+                                  server does not support pull diagnostics.
                                 • {exit_timeout} (`integer|false`, default:
                                   `false`) Milliseconds to wait for server to
                                   exit cleanly after sending the "shutdown"
@@ -1204,6 +1209,11 @@ Lua module: vim.lsp.client                                        *lsp-client*
                                 `150`) Debounce `didChange` notifications to
                                 the server by the given number in
                                 milliseconds. No debounce occurs if `nil`.
+                              • {pull_diagnostics_on_save} (`boolean`,
+                                default: `false`) Request diagnostics from the
+                                server on `didSave` events instead of
+                                `didChange` events. Has no effect if the
+                                server does not support pull diagnostics.
                               • {exit_timeout} (`integer|false`, default:
                                 `false`) Milliseconds to wait for server to
                                 exit cleanly after sending the "shutdown"
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -130,6 +130,8 @@ LSP
 
 • Completion side effects (including snippet expansion, execution of commands
   and application of additional text edits) is now built-in.
+• |vim.lsp.Client| flag `pull_diagnostic_on_save` sends diagnostic requests on
+  save instead of change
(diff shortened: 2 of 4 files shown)


justinmk:
Does it really need to be optional? And, is it not possible to "configure" this via client/server capabilities?

Delerme:
> Does it really need to be optional?
I can make it required

> Is it not possible to "configure" this via client/server capabilities?
The user could override the capabilities to disable pull diagnostics entirely, but otherwise I'm not aware of something in the capabilities that would work. It seems like pull diagnostics are the "future" of lsp diagnostics so that the client is in control of when to pull, instead of every server having different implementations and settings. If so, it seems useful for neovim to start supporting settings for it. I could imagine this becoming a full field on vim.lsp.ClientConfig later, like `pull_diagnostics_on` with settings like `change`, `save`, `normal_mode`, etc. Open to other thoughts and ideas!

Delerme:
I may need some help with writing the functional test

I tried duplicating [this test](https://github.com/neovim/neovim/blob/a8fbe1d409e08c68b05bc26b096486020ae3162b/test/functional/plugin/lsp_spec.lua#L1295) and adjusted:
- added `diagnosticProvider` capability to enable pull diagnostics
- check for `textDocument/diagnostic` event sent to lsp sever

I added assertions checking that the capabilities were set correctly and that the [correct autocmd](https://github.com/neovim/neovim/blob/a8fbe1d409e08c68b05bc26b096486020ae3162b/runtime/lua/vim/lsp/inlay_hint.lua#L267) was on the buffer but I couldn't get it to fire. I even tried manually execing the autocmd in an `exec_lua` block but the callback didn't fire. Not sure if there's something special with autocmds in tests, or if you have any other advice!

mfussenegger:
> Some lsp servers can be configured to refresh diagnostics on save to avoid performance issues. The introduction of pull diagnostics in 0.10 goes around those settings by requesting diagnostics with every change event. See https://github.com/neovim/neovim/issues/29792

Couldn't you turn off pull in that case and rely on push?

> It seems like pull diagnostics are the "future" of lsp diagnostics so that the client is in control of when to pull, instead of every server having different implementations and settings. If so, it seems useful for neovim to start supporting settings for it. I could imagine this becoming a full field on vim.lsp.ClientConfig later, like pull_diagnostics_on with settings like change, save, normal_mode, etc. Open to other thoughts and ideas!

Afaik part of the rationale for a pull in addition to a push was that the client could have more control over which document to get diagnostics for first, to get them faster. It's not replacing push but complementing. 

Adding this new flag would imho go a bit against that and we definitely don't want to have a dozen settings to control when it should trigger.
Instead what we could do is expose enable and refresh and leave it up to the user to setup autocmds as they please. That would give maximum flexibility without having to introduce just one more option whenever someone can think of a new scenario. 

But I'm overall not convinced this is needed

Delerme:
> Couldn't you turn off pull in that case and rely on push?
> Afaik part of the rationale for a pull in addition to a push was that the client could have more control over which document to get diagnostics for first, to get them faster. It's not replacing push but complementing.

Ah that's helpful context, thanks!

> Adding this new flag would imho go a bit against that and we definitely don't want to have a dozen settings to control when it should trigger. Instead what we could do is expose enable and refresh and leave it up to the user to setup autocmds as they please. That would give maximum flexibility without having to introduce just one more option whenever someone can think of a new scenario.

Yes, given that info that makes more sense. What do you think about for now adding a simpler setting in the lsp client config to disable pull diagnostics? Right now it requires some relatively deep knowledge of lsp servers, capabilities, and how neovim implemented pull diagnostics. I'll admit when I tried to disable them I couldn't get it to work. It could be in `flags` or in the `ClientConfig` directly, but something like `pull_diagnostics boolean` or `disable_pull_diagnostics boolean`? What do you think?

PR #30060, opened 2024-08-15
feat(lsp): completion opts support custom format of abbr

glepnir:
Problem: Some items of completion response have params signatures this will cause the pum ver long when a function have more params, because pum always align with the longest word/abbr.

Solution: add custom format function that can  remove params.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1630,6 +1630,9 @@ Lua module: vim.lsp.completion                                *lsp-completion*
     Fields: ~
       • {autotrigger}?  (`boolean`) Whether to trigger completion
                         automatically. Default: false
+      • {convert}?      (`fun(item: lsp.CompletionItem): table`) An optional
+                        function used to customize the transformation of an
+                        LSP CompletionItem to |complete-items|.
 
 
                                                  *vim.lsp.completion.enable()*
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -23,6 +23,7 @@ local ns_to_ms = 0.000001
 --- @class vim.lsp.completion.BufHandle
 --- @field clients table<integer, vim.lsp.Client>
 --- @field triggers table<string, vim.lsp.Client[]>
+--- @field convert? fun(item: lsp.CompletionItem): table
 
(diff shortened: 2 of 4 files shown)


glepnir:
any thoughts before add test case :)

runtime/lua/vim/lsp/completion.lua
```diff
@@ -250,9 +251,18 @@ function M._lsp_to_complete_items(result, prefix, client_id)
...
     if matches(item) then
       local word = get_completion_word(item)
+      local abbr = item.label
+      if buf_handles[bufnr].abbr_format then
+        if abbr then
+          abbr = buf_handles[bufnr].abbr_format(abbr)
+        else
+          word = buf_handles[bufnr].abbr_format(word)
+        end
+      end
```
mfussenegger, line 265: `item.label` is afaik a mandatory property so the `if abbr then` case should always be true and the else brach would be dead code.


I'm overall not sure about this. Seems a bit too specific. If servers report long labels maybe it should be brought up with them.

Or otherwise another option could be to set a `convert(item)` function that can return the full structure, including word, kind, info, etc.. ` and also apply a `tbl_extend("keep", ...)` on the result to fill in missing properties.

This way you could customize `abbr` as well as any of the other properties.

runtime/doc/lsp.txt
```diff
@@ -1630,6 +1630,7 @@ Lua module: vim.lsp.completion                                *lsp-completion*
     Fields: ~
       • {autotrigger}?  (`boolean`) Whether to trigger completion
                         automatically. Default: false
+      • {abbr_format}?  (`function`)
```
justinmk, line 1633:
```suggestion
      • {abbr_format}?  (`function`) Decides the format of ...
```

MariaSolOs: Mentioning the default format would also be helpful.

runtime/lua/vim/lsp/completion.lua
glepnir: > I'm overall not sure about this. Seems a bit too specific. If servers report long labels maybe it should be brought up with them.

lsp no universal standard  here.  our api functions luals also return function name with all parameters.

convert is a more general way.

mfussenegger (approved):
Could you check the CI failures? Otherwise lgtm

PR #30064, opened 2024-08-15
feat(lsp): export diagnostic conversion functions

MariaSolOs:
Now that `vim.diagnostic.get` is the only way to obtain diagnostics, it's sometimes necessary to convert these to their LSP equivalent for use with other functions.

Example: When sending a code action request and including the diagnostics in the current line as context.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1486,6 +1486,15 @@ workspace_symbol({query}, {opts})             *vim.lsp.buf.workspace_symbol()*
 ==============================================================================
 Lua module: vim.lsp.diagnostic                                *lsp-diagnostic*
 
+from({diagnostics})                                *vim.lsp.diagnostic.from()*
+    Converts the input `vim.Diagnostic`s to LSP diagnostics.
+
+    Parameters: ~
+      • {diagnostics}  (`vim.Diagnostic[]`)
+
+    Return: ~
+        (`lsp.Diagnostic[]`)
+
                                           *vim.lsp.diagnostic.get_namespace()*
 get_namespace({client_id}, {is_pull})
     Get the diagnostic namespace associated with an LSP client
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -68,6 +68,8 @@ LSP
 • Add convert field in |vim.lsp.completion.BufferOpts| of
   |vim.lsp.completion.enable()| an optional function used to customize the
   transformation of an Lsp CompletionItem to |complete-items|.
+• |vim.lsp.diagnostic.from()| can now be used to convert a given list of `vim`
+  diagnostics to their LSP representation.
 
 LUA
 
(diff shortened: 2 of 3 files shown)


mfussenegger (commented):
I'm not sure about this.

- The `lsp_to_vim` version should never be needed. Or what case do you have in mind where a user needs to call `diagnostic.set` with LSP diagnostics that isn't handled by the existing push/pull support?
- The mentioned use-case for `vim_to_lsp` would be kinda broken for multi-clients, as currently the code action function doesn't offer a way to provide _per_ client diagnostics. We'd have to fix that API first - and if we do we could consider accepting vim diagnostic directly instead of having users do the conversion outside.

MariaSolOs:
> The `lsp_to_vim` version should never be needed. Or what case do you have in mind where a user needs to call `diagnostic.set` with LSP diagnostics that isn't handled by the existing push/pull support?

You're right. I don't have a particular use case for this direction.

> as currently the code action function doesn't offer a way to provide _per_ client diagnostics.

Are you referring to `vim.lsp.buf.code_action`? If so, that's not actually the function I'm interested in in my example. To be more concrete, I have a "VS Code-like" lightbulb in my configuration that's displayed when the line where my cursor is at has available fixes. For this I need to get the code actions via `vim.lsp.buf_request`, since `vim.lsp.buf.code_action` handles the result with `vim.ui.select` (which I don't need here). So modifying the code action API doesn't handle this case where I'm constructing the arguments for `vim.lsp.buf_request`.

mfussenegger:
I see. In that case I'd keep `lsp_to_vim` private and I guess we can expose the other one.
Not sure about the name, given that it's already in the `lsp.diagnostic` namespace I think we could omit the `diagnostics_` prefix. `vim_to_lsp`, or maybe even `to_lsp`  - although that might be a bit confusing in the lsp namespace and would make more sense if it were in the `vim.diagnostic` namespace (which I don't think we want to change?)

cc @gpanders

MariaSolOs:
@mfussenegger @gpanders Another idea is to add an option to `vim.diagnostic.get` to return LSP diagnostics instead of vim ones.

mfussenegger:
> @mfussenegger @gpanders Another idea is to add an option to `vim.diagnostic.get` to return LSP diagnostics instead of vim ones, and keep the conversion code private.

Sounds intriguing at first but the downside of that is that we can't express the relationship between parameter and return value with the type system so we'd end up with `vim.Diagnostic[]|lsp.Diagnostic[]`. Not a fan of that.

MariaSolOs:
> Sounds intriguing at first but the downside of that is that we can't express the relationship between parameter and return value with the type system so we'd end up with `vim.Diagnostic[]|lsp.Diagnostic[]`. Not a fan of that.

Hmmm we can try generics, but I think generic annotations aren't powerful enough for this yet.

MariaSolOs:
@gpanders I would appreciate your input here :)

clason:
`get()`?

gpanders (approved):
Just needs a news entry

gpanders (approved):

Issue #29892, opened 2024-07-28
`editorconfig.trim_trailing_whitespace` breaks repeat last substitute on save


mateuszmandera:
### Problem

`trim_trailing_whitespace` when active, since it works by using `:s`, changes the last substitute string on every `BufWritePre`:

https://github.com/neovim/neovim/blob/5aa1a9532cbac835ad027ebdf04311c7e8fb7007/runtime/lua/editorconfig.lua#L147-L170

This is consistent with the original plugin I believe (https://github.com/editorconfig/editorconfig-vim/issues/175), but nonetheless is pretty surprising behavior as it silently breaks things like `&` (repeat last substitute) or `:s/newpattern/~`.
What adds to the confusion is that since it does preserve the original search history, then if you do:
```
:s/foo/replacement
:w
&
```
you're going to get substitution of `foo` with the empty string.

### Steps to reproduce

```
# in a project with `.editorconfig` with trim_trailing_whitespace = true
nvim --clean .
:s/foo/replacement
:w
&
```

### Expected behavior

In my view this would ideally preserve the original last substitute value.

### Neovim version (nvim -v)

NVIM v0.10.0

justinmk:
If you remove `keeppatterns` does it fix the issue? https://github.com/neovim/neovim/blob/ab561302a3b4dbb161aa6ef3ea39a6d1410a72fe/runtime/lua/editorconfig.lua#L159

zeertzjq:
#30066 preserves the substitute string for `&`. If you want to preserve the string for `~` as well please open another issue.

PR #30066, opened 2024-08-16
vim-patch: :keepp does not retain last substitute string

zeertzjq:
#### vim-patch:9.1.0677: :keepp does not retain the substitute pattern

Problem:  :keeppatterns does not retain the substitute pattern
          for a :s command
Solution: preserve the last substitute pattern when used with the
          :keeppatterns command modifier (Gregory Anders)

closes: vim/vim#15497

https://github.com/vim/vim/commit/3b59be4ed8a145d3188934f1a5cd85432bd2433d

Co-authored-by: Gregory Anders <greg@gpanders.com>

### DIFF
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -348,7 +348,7 @@ terminals)
 
 :keepp[atterns] {command}			*:keepp* *:keeppatterns*
 		Execute {command}, without adding anything to the search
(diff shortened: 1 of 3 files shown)


gpanders:
This will also resolve https://github.com/neovim/neovim/issues/29892.

justinmk (approved):

PR #30068, opened 2024-08-16
vim-patch:9.1.{0678,0679}

zeertzjq:
#### vim-patch:9.1.0678: [security]: use-after-free in alist_add()

Problem:  [security]: use-after-free in alist_add()
          (SuyueGuo)
Solution: Lock the current window, so that the reference to
          the argument list remains valid.

This fixes CVE-2024-43374

https://github.com/vim/vim/commit/0a6e57b09bc8c76691b367a5babfb79b31b770e8

#### vim-patch:9.1.0679: Rename from w_closing to w_locked is incomplete

Problem:  Rename from w_closing to w_locked is incomplete
          (after 9.1.0678).
Solution: Rename remaining occurrences of w_closing to w_locked and
          update comments (zeertzjq).

closes: vim/vim#15504

https://github.com/vim/vim/commit/bc11f6d9d4f580179d3963f3c489881904d6cd62

### DIFF
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -2471,7 +2471,7 @@ void close_windows(buf_T *buf, bool keep_curwin)
   // When the autocommand window is involved win_close() may need to print an error message.
   for (win_T *wp = lastwin; wp != NULL && (is_aucmd_win(lastwin) || !one_window(wp));) {
     if (wp->w_buffer == buf && (!keep_curwin || wp != curwin)
-        && !(wp->w_closing || wp->w_buffer->b_locked > 0)) {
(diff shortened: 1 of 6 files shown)


src/nvim/window.c
```diff
@@ -2798,9 +2798,6 @@ int win_close(win_T *win, bool free_buf, bool force)
   // to split a window to avoid trouble.
   split_disallowed++;
 
-  // let terminal buffers know that this window dimensions may be ignored
```
zeertzjq, line 2801: `w_closing` hasn't been used in terminal.c for very long.

PR #23751, opened 2023-05-25
Expose a couple of diagnostic helper functions

WhyNotHugo:
These helpers are currently used internally but are extremely useful for plugin developers writing plugins that handle diagnostics in any way.

These two functions themselves use other private functions, so re-implementing this externally requires copy-pasting around 90 lines of code into each site that needs this functionality.

Fixes: https://github.com/neovim/neovim/issues/22900

PR #30070, opened 2024-08-17
feat(diagnostic)!: filter diagnostics by severity before passing to handlers

gpanders:
BREAKING CHANGE: This changes the list of diagnostics that are passed to
a diagnostic handler. If a handler is already filtering by severity
itself then this won't break anything, since the handler's filtering
will become a no-op. But handlers which depend on receiving the full
list of diagnostics may break.

Note that diagnostics are only filtered if the handler's configuration
has the `severity` option set. If `severity` is not set, the handler
still receives the full list of diagnostics.

Ref: https://github.com/neovim/neovim/pull/23751#issuecomment-1563498953

cc @WhyNotHugo

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -68,6 +68,8 @@ DIAGNOSTICS
 
 • The "underline" diagnostics handler sorts diagnostics by severity when using
   the "severity_sort" option.
+• Diagnostics are filtered by severity before being passed to a diagnostic
+  handler |diagnostic-handlers|.
 
 EDITOR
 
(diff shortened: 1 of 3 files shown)


runtime/doc/news.txt
```diff
@@ -42,6 +42,9 @@ DIAGNOSTICS
 
 • |vim.diagnostic.config()| accepts a "jump" table to specify defaults for
   |vim.diagnostic.jump()|.
+• Diagnostics are filtered by severity before being passed to a diagnostic
```
MariaSolOs, line 45: I think we should also document this in [`diagnostic.txt`](https://github.com/neovim/neovim/blob/9e0d40f7e45f483e54d38be1266f63240808b4b0/runtime/doc/diagnostic.txt#L73).

MariaSolOs (approved):

runtime/doc/news.txt
gpanders: Good call, added.

PR #30075, opened 2024-08-17
fix(vim.text): handle very long strings

### LINKED ISSUE #29981: Error using vim.text.hexdecode on large (?) strings.
### Problem

Consider:
```lua
-- https://github.com/neovim/neovim/blob/0c2860d9e5ec5417a94db6e3edd237578b76d418/test/functional/lua/text_spec.lua#L14
local input, output = string.rep("😂", 2 ^ 16), string.rep("F09F9882", 2 ^ 16)
assert(output, vim.text.hexencode(input))
assert(input, vim.text.hexdecode(output))
```
The I get an error `string slice is too long`. It seems the string. functions have a problem with large strings.
I think a better implementation would be to use the jit string.buffer:
```lua
local buffer = require("string.buffer")
local bit = require("bit")

local function hexencode2(str)
  local bytes = buffer.new()
  local enc = buffer.new()
  bytes:set(str)
  local ptr, len = bytes:ref()
  for i = 0, len - 1 do
    -- enc:put(bit.tohex(ptr[i], -2))
    enc:putf("%02X", ptr[i])
  end
  return enc:tostring()
end

local function hexdecode2(enc)
  local dec = buffer.new()
  local enc_buf = buffer.new()
  enc_buf:set(enc)
  local ptr, len = enc_buf:ref()
  for i = 0, len - 1, 2 do
    local hex = (ptr[i] % 0x20 + 0x9) % 0x19 * 0x10 + (ptr[i + 1] % 0x20 + 0x9) % 0x19
    dec:putf("%c", hex)
  end
  return dec:tostring()
end
local input, output = string.rep("😂", 2 ^ 16), string.rep("F09F9882", 2 ^ 16)
assert(output == hexencode2(input))
assert(input == hexdecode2(output))
```

### Steps to reproduce

```lua
-- https://github.com/neovim/neovim/blob/0c2860d9e5ec5417a94db6e3edd237578b76d418/test/functional/lua/text_spec.lua#L14
local input, output = string.rep("😂", 2 ^ 16), string.rep("F09F9882", 2 ^ 16)
assert(output == vim.text.hexencode(input))
assert(input == vim.text.hexdecode(output))
```

### Expected behavior

success

### Neovim version (nvim -v)

NVIM v0.10.1

not applicable

### Operating system/version

Ubuntu 24.04

### Terminal name/version

wezterm

wezterm

### Installation

Release from github

gpanders:
Lua's string.byte has a maximum (undocumented) allowable length, so vim.text.hencode fails on large strings with the error "string slice too long".

Instead of converting the string to an array of bytes up front, convert each character to a byte one at a time.

Fixes: https://github.com/neovim/neovim/issues/29981

### DIFF
--- a/runtime/lua/vim/text.lua
+++ b/runtime/lua/vim/text.lua
@@ -7,10 +7,9 @@ local M = {}
 --- @param str string String to encode
 --- @return string : Hex encoded string
 function M.hexencode(str)
-  local bytes = { str:byte(1, #str) }
   local enc = {} ---@type string[]
-  for i = 1, #bytes do
-    enc[i] = string.format('%02X', bytes[i])
+  for i = 1, #str do
+    enc[i] = string.format('%02X', str:byte(i, i + 1))
   end
   return table.concat(enc)
 end
--- a/test/functional/lua/text_spec.lua
+++ b/test/functional/lua/text_spec.lua
@@ -20,5 +20,11 @@ describe('vim.text', function()
         eq(input, vim.text.hexdecode(output))
       end
     end)
+
+    it('works with very large strings', function()
+      local input, output = string.rep('😂', 2 ^ 16), string.rep('F09F9882', 2 ^ 16)
+      eq(output, vim.text.hexencode(input))
+      eq(input, vim.text.hexdecode(output))
+    end)
   end)
 end)


pedrohgmacedo:
I think it would be more efficient to chunk it in larger pieces. I don't know if it is a lua thing, but in luajit the code is https://github.com/LuaJIT/LuaJIT/blob/ae4735f621d89d84758769b76432d2319dda9827/src/lib_string.c#L57

gpanders:
I also tried a chunked approach:

```lua
function M.hexencode(str)
  local enc = {} ---@type string[]
  local chunk = 4096
  for i = 1, #str, chunk do
    chunk = math.min(chunk, #str - i + 1)
    enc[#enc + 1] = string.rep('%02X', chunk):format(str:byte(i, i + chunk))
  end

  return table.concat(enc)
end
```

But this is not any faster than the simple approach implemented in this PR (both take around 40ms for a string of length 2^16 on my M1 Macbook Pro).

If we really needed this function to be much faster we could implement it in C, but I don't think that's necessary right now.

PR #30076, opened 2024-08-17
fix: alignment in Lesson 1.1

bart1024:


zeertzjq:
No, `j` and `k` should not be aligned.

PR #30083, opened 2024-08-18
refactor: adopt vterm

dundargoc:
There's a lot of unicode-related code in libvterm which is duplicated by
the functionality in mbyte.c and utf8proc. To ensure a consistent
experience we have to adopt the code sections we're already
implementing.


dundargoc:
@gpanders @justinmk yay or nay on this? It would essentially entail giving it the same treatment as termkey. The one difference is that vterm is supposedly not deprecated.

dundargoc:
https://github.com/neovim/neovim/pull/30047 should be merged beforehand if we want to move forward with this.

src/nvim/tui/termkey.c
clason, line 1: I don't think we want to do that; the termkey files should stay quarantined in a separate (sub)directory.

clason: (Sorry, looked like you had just rebased so I thought that was new.)

dundargoc:
People have expressed that they would like to import the libvterm tests before attempting any major refactors such as this. This is frozen until that happens.

dundargoc:
The tests have been [ported](https://github.com/neovim/neovim/pull/30335) so this is good to go, but @bfredl has some ongoing work in this in https://github.com/neovim/neovim/pull/31635 so don't think it's worth keeping this open. This can be done after bfredl is done with his fixes.

dundargoc:
Continued in https://github.com/neovim/neovim/pull/31846

PR #30098, opened 2024-08-20
fix(lsp): improve lsp request name to capability map

atusy:
by adding items based on specifications
(<https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentClientCapabilities>)

- textDocument/documentColor
- textDocument/inlineValue
- textDocument/linkedEditingRange
- textDocument/moniker
- textDocument/onTypeFormatting
- textDocument/selectionRange

In particular, I needed `textDocument/documentColor` so that `ccc.nvim` correctly understands the LSP configurations.

https://github.com/uga-rosa/ccc.nvim/blob/4fb5abaef2f2e0540fe22d4d74a9841205fff9e4/lua/ccc/handler/lsp.lua?plain=1#L58-L64

The commit uses `fix` tag to be aligned with the following PR

- https://github.com/neovim/neovim/pull/28838


runtime/lua/vim/lsp.lua
```diff
@@ -71,6 +71,12 @@ lsp._request_name_to_capability = {
   [ms.textDocument_didOpen] = { 'textDocumentSync', 'openClose' },
   [ms.textDocument_willSave] = { 'textDocumentSync', 'willSave' },
   [ms.textDocument_willSaveWaitUntil] = { 'textDocumentSync', 'willSaveWaitUntil' },
+  [ms.textDocument_documentColor] = { 'colorProvider' },
```
MariaSolOs, line 74: Super nit which should be a separate commit, but to better keep track of what's been added so far it would be useful to sort these keys alphabetically.

MariaSolOs (approved):

runtime/lua/vim/lsp.lua
atusy: Definitely! I will.

atusy: Done. Thanks!

https://github.com/neovim/neovim/pull/30098/commits/06246472b3da7351a06206e19525f8081ba239b4

justinmk (approved):
should this be done by `scripts/gen_lsp.lua` ?

MariaSolOs:
> should this be done by `scripts/gen_lsp.lua` ?

Don’t think we can (or at least it wouldn’t be useful) since the metamodel doesn’t have information about the respective server capability for each request.

MariaSolOs:
We can ask for that information to be included though.

PR #30105, opened 2024-08-20
fix(tohtml): quote font-family names

yayoyuyu:
Problem:
The font is not applied when the font-family name contains spaces.

Solution:
Enclose the font-family name in quotation marks.

### DIFF
--- a/runtime/lua/tohtml.lua
+++ b/runtime/lua/tohtml.lua
@@ -1293,9 +1293,25 @@ local function opt_to_global_state(opt, title)
   local fonts = {}
   if opt.font then
     fonts = type(opt.font) == 'string' and { opt.font } or opt.font --[[@as (string[])]]
+    for i, v in pairs(fonts) do
+      fonts[i] = ('"%s"'):format(v)
+    end
   elseif vim.o.guifont:match('^[^:]+') then
-    table.insert(fonts, vim.o.guifont:match('^[^:]+'))
+    -- Example:
+    -- Input: "Font,Escape\,comma, Ignore space after comma"
+    -- Output: { "Font","Escape,comma","Ignore space after comma" }
+    local prev = ''
+    for name in vim.gsplit(vim.o.guifont:match('^[^:]+'), ',', { trimempty = true }) do
+      if vim.endswith(name, '\\') then
+        prev = prev .. vim.trim(name:sub(1, -2) .. ',')
+      elseif vim.trim(name) ~= '' then
+        table.insert(fonts, ('"%s%s"'):format(prev, vim.trim(name)))
+        prev = ''
+      end
+    end
   end
+  -- Generic family names (monospace here) must not be quoted
+  -- because the browser recognizes them as font families.
   table.insert(fonts, 'monospace')
   --- @type vim.tohtml.state.global
   local state = {
--- a/test/functional/plugin/tohtml_spec.lua
+++ b/test/functional/plugin/tohtml_spec.lua
@@ -137,27 +137,26 @@ local function run_tohtml_and_assert(screen, func)
 end
 
 ---@param guifont boolean
-local function test_generates_html(guifont)
+local function test_generates_html(guifont, expect_font)
   insert([[line]])
   exec('set termguicolors')
   local bg = fn.synIDattr(fn.hlID('Normal'), 'bg#', 'gui')
   local fg = fn.synIDattr(fn.hlID('Normal'), 'fg#', 'gui')
-  if guifont then
-    exec_lua [[
-      vim.o.guifont="Font,Escape\\,comma, Ignore space after comma"
-      local outfile = vim.fn.tempname() .. '.html'
-      local html = require('tohtml').tohtml(0,{title="title"})
-      vim.fn.writefile(html, outfile)
-      vim.cmd.split(outfile)
-    ]]
-  else
-    exec_lua [[
-      local outfile = vim.fn.tempname() .. '.html'
-      local html = require('tohtml').tohtml(0,{title="title",font={ "dumyfont","anotherfont" }})
-      vim.fn.writefile(html, outfile)
-      vim.cmd.split(outfile)
-    ]]
-  end
+  local tmpfile = t.tmpname()
+
+  exec_lua(
+    [[
+    local guifont, outfile = ...
+    local html = (guifont
+      and require('tohtml').tohtml(0,{title="title"})
+      or require('tohtml').tohtml(0,{title="title",font={ "dumyfont","anotherfont" }}))
+    vim.fn.writefile(html, outfile)
+    vim.cmd.split(outfile)
+  ]],
+    guifont,
+    tmpfile
+  )
+
   local out_file = api.nvim_buf_get_name(api.nvim_get_current_buf())
   eq({
     '<!DOCTYPE html>',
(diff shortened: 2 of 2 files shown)


runtime/lua/tohtml.lua
```diff
@@ -1293,6 +1293,9 @@ local function opt_to_global_state(opt, title)
   elseif vim.o.guifont:match('^[^:]+') then
     table.insert(fonts, vim.o.guifont:match('^[^:]+'))
   end
+  for i, v in pairs(fonts) do
+    fonts[i] = ('"%s"'):format(v)
+  end
```
justinmk, line 1298: since we're already doing a table.concat below, perhaps it makes sense to do this there. instead of storing this "formatting" info in the `fonts` list (although it's short lived so doesn't technically matter currently)

    font = ('"%s"'):format(table.concat(fonts, '","'))

yayoyuyu: Generic family names (`monospace` here) must not be quoted because the browser recognizes them as font families. so, I think we should enclose specific font families in quotation marks before adding `monospace` to the `fonts`.

justinmk: then we need to call that out in a code comment

yayoyuyu: When multiple fonts are specified in `vim.o.guifont`, they are currently being inserted as a single element. It seems we need to modify this so that each font is inserted as a separate element.

and, Is it appropriate to write the comment like this above the for loop?
```
-- Generic family names (monospace here) must not be quoted
-- because the browser recognizes them as font families.
```

yayoyuyu: I mistakenly thought it was about specifying multiple fonts in `vim.o.guifont`.(A fix is necessary for the above problem.)
If `monospace` is specified in `vim.o.guifont`, it is treated as a regular font family name, resulting in something like:
```
{font-family: "monospace",monospace}
```
It is unlikely that a generic family will be specified in `vim.o.guifont`.

runtime/lua/tohtml.lua
```diff
@@ -1291,7 +1291,29 @@ local function opt_to_global_state(opt, title)
...
+    while i <= #guifont do
+      char = guifont:sub(i, i)
+      if char == ',' then
+        if char_pre == '\\' then
+          str = str .. char
+        else
+          table.insert(fonts, str)
+          str = ''
+        end
+      elseif char ~= '\\' and not (char == ' ' and char_pre == ',') then
```
justinmk, line 1307: would `vim.iter` be helpful for simplifying this code?

need to find a way to simplify all of this. too hard to follow currently

yayoyuyu: Hmm... I thought about it, but I couldn't come up with a good implementation. However, using for like this might make it a bit simpler:
```
local str, char_pre = '', ''
for _, char in ipairs(vim.split(vim.o.guifont:match('^[^:]+'), '')) do
    if char == ',' then
        if char_pre == '\\' then
            str = str .. char
        else
            table.insert(fonts, str)
            str = ''
        end
    elseif char ~= '\\' and not (char == ' ' and char_pre == ',') then
        str = str .. char
    end
    char_pre = char
end
table.insert(fonts, str)
```

justinmk: please also use `('...'):format()` where possible. string concat with `..` is harder to visualize, and often more verbose.

justinmk: please add a comment that shows a "input => output" example. that is 100x more readable than a bunch of string concatenations.

runtime/lua/tohtml.lua
```diff
@@ -1291,7 +1291,29 @@ local function opt_to_global_state(opt, title)
   if opt.font then
     fonts = type(opt.font) == 'string' and { opt.font } or opt.font --[[@as (string[])]]
   elseif vim.o.guifont:match('^[^:]+') then
-    table.insert(fonts, vim.o.guifont:match('^[^:]+'))
+    -- Example:
+    -- Input: "Font,Escape\,comma, Ignore space after comma"
+    -- Output: { "Font","Escape,comma","Ignore space after comma" }
+    local str, char_pre = '', ''
+    for _, char in ipairs(vim.split(vim.o.guifont:match('^[^:]+'), '')) do
```
justinmk, line 1298: IIUC this is splitting on empty string as a way to inspect each character? And most of the logic is trying to deal with escaped commas (`\,`) ? Why not use `vim.gsplit` instead of `ipairs(vim.split)` ?

This seems to work:

```lua
local fonts = {}
local prev = ''
for name in vim.gsplit(vim.o.guifont:match('^[^:]+'), ',', {trimempty=true}) do
  if vim.endswith(name, '\\') then
    prev = prev .. vim.trim(name:sub(1, -2) .. ',')
  elseif vim.trim(name) ~= '' then
    table.insert(fonts, ('"%s%s"'):format(prev or '', vim.trim(name)))
    prev = ''
  end
end
```

runtime/lua/tohtml.lua
```diff
@@ -1291,7 +1291,29 @@ local function opt_to_global_state(opt, title)
...
+        str = str .. char
+      end
+      char_pre = char
+    end
+    table.insert(fonts, str)
+  end
+  -- Generic family names (monospace here) must not be quoted
+  -- because the browser recognizes them as font families.
+  for i, v in pairs(fonts) do
+    fonts[i] = ('"%s"'):format(v)
```
justinmk, line 1316: not needed if you use the code from https://github.com/neovim/neovim/pull/30105/files#r1741902695

runtime/lua/tohtml.lua
yayoyuyu: Yes, that’s right. That implementation didn’t occur to me...
As for the quotation marks, since opt.font also needs to be enclosed in them, I think it would be better to do that below rather than here.

justinmk: > since opt.font also needs to be enclosed in them

can we just change line 1292 to:

    fonts = { ('"%s"'):format(type(opt.font) == 'string' and opt.font or opt.font[1]) }

yayoyuyu: With this code, if multiple fonts are specified in opt.font, only the first item is applied. Is this the expected behavior?

justinmk: see different suggestion: https://github.com/neovim/neovim/pull/30105/files#r1742248186

justinmk: yes. the point is that logic is only needed for that if-branch, there is no reason to repeat it for both branches.

test/functional/plugin/tohtml_spec.lua
```diff
@@ -162,7 +162,7 @@ describe(':TOhtml', function()
       '<title>title</title>',
       ('<meta name="colorscheme" content="%s"></meta>'):format(api.nvim_get_var('colors_name')),
       '<style>',
-      '* {font-family: dumyfont,monospace}',
+      '* {font-family: "dumyfont",monospace}',
```
justinmk, line 165: can we add another font or two on line 152 so that this covers all the logic from this PR?

justinmk:
code LGTM. can you make the [test change](https://github.com/neovim/neovim/pull/30105/files#r1743845724), assuming it makes sense?

yayoyuyu:
It seems there is no test for the case where `opt.font` is not specified and a font is selected from `vim.o.guifont`. Should we add that as well?

yayoyuyu:
> factor out the existing test into a test_generates_html(guifont) function

Is this what you mean...? -> de1df14d1a5de2fdb3e06be1cd893320d4c09ef2

test/functional/plugin/tohtml_spec.lua
```diff
@@ -143,37 +190,11 @@ describe(':TOhtml', function()
   end)
 
   it('expected internal html generated', function()
```
justinmk, line 192:
```suggestion
  it('generates html', function()
```

test/functional/plugin/tohtml_spec.lua
```diff
@@ -143,37 +190,11 @@ describe(':TOhtml', function()
...
-      'line',
-      '',
-      '</pre>',
-      '</body>',
-      '</html>',
-    }, fn.readfile(out_file))
+    test_generates_html(false)
+  end)
+
+  it('expected internal html generated using guifont', function()
```
justinmk, line 196:
```suggestion
  it("generates html, respects 'guifont'", function()
```

PR #2674, opened 2015-05-16
[RFC]fix flicker in popup menu

SolaWing:
as #2661 says, I spend a day to fix the annoying flicker.

In detail, I `update_screen` before restart_complete, and remove `update_screen` in `ins_compl_new_leader`, so the screen will stay with last popup menu, until the complete function returns  and update screen. Instead of block without popup menu, which disappear and reappear, causing flicker.

Another change is when typed char, call the `ins_compl_new_leader` even `refresh_always` is setted. So complete func can immediately get called after change leader char. Instead of have to waiting for `CursorMoved` event and reinvoke complete function by feedkeys, which also cause flicker. 

The last change is if restart_complete with new_leader, and in user complete func, refresh_always is set, free `compl_leader`, so popup menu doesn't use prefix as a filter. this cause all the result returned by complete func valid. This make more sense for fuzzy complete func, which need completely control complete and filter. The return candidate certainly may not match with the prefix, since it use itself match rule.

PR #30111, opened 2024-08-21
vim-patch:9.1.0684: completion is inserted on Enter with "noselect"

glepnir:
Problem:  completion is inserted on Enter with "noselect"
          (Carman Fu)
Solution: check noselect before update compl_shown_match
          (glepnir)

closes: vim/vim#15530

https://github.com/vim/vim/commit/753794bae8a9401903b82e5c5d1f35a106aa912a

### DIFF
--- a/test/old/testdir/test_ins_complete.vim
+++ b/test/old/testdir/test_ins_complete.vim
@@ -2664,6 +2664,7 @@ func Test_complete_fuzzy_match()
   call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
   call assert_equal('hello help hero h', getline('.'))
 
+  " issue #15526
   set completeopt=fuzzy,menuone,menu,noselect
   call setline(1, ['Text', 'ToText', ''])
   call cursor(2, 1)
--- a/test/old/testdir/test_ins_complete.vim
+++ b/test/old/testdir/test_ins_complete.vim
@@ -2664,6 +2664,12 @@ func Test_complete_fuzzy_match()
   call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
   call assert_equal('hello help hero h', getline('.'))
 
+  set completeopt=fuzzy,menuone,menu,noselect
+  call setline(1, ['Text', 'ToText', ''])
+  call cursor(2, 1)
+  call feedkeys("STe\<C-X>\<C-N>x\<CR>\<Esc>0", 'tx!')
+  call assert_equal('Tex', getline('.'))
+
   " clean up
   set omnifunc=
   bw!
--- a/src/nvim/insexpand.c
+++ b/src/nvim/insexpand.c
@@ -1244,7 +1244,9 @@ static int ins_compl_build_pum(void)
         if (comp->cp_score > max_fuzzy_score) {
           did_find_shown_match = true;
           max_fuzzy_score = comp->cp_score;
-          compl_shown_match = comp;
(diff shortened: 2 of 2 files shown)


glepnir:
guess no need a test in popupmenu_spec.because it test for behavior of <CR> so just equal the line data..

glepnir:
waht's the different there of vim.. <CR> insert newline but when i do this operator in build nvim it doesn't insert newline..
test_case

glepnir:
@zeertzjq  any opinion on there ..

zeertzjq:
I think the difference is caused by #2674

glepnir:
Oh thanks ！！,the test passed by splitting it into two sends...

zeertzjq:
Meanwhile, it seems that removing this condition will make the behavior here match Vim:
```diff
diff --git a/src/nvim/insexpand.c b/src/nvim/insexpand.c
index 6446a4e70c..b55c918bf7 100644
--- a/src/nvim/insexpand.c
+++ b/src/nvim/insexpand.c
@@ -1835,8 +1835,7 @@ static void ins_compl_new_leader(void)
   ins_compl_show_pum();
 
   // Don't let Enter select the original text when there is no popup menu.
-  // Don't let Enter select when use user function and refresh_always is set
-  if (compl_match_array == NULL || ins_compl_need_restart()) {
+  if (compl_match_array == NULL) {
     compl_enter_selects = false;
   }
 }
```
And it doesn't seem mentioned in #2674 what the purpose of this condition is.

glepnir:
will try to remove it later to see if it breaks the test of #2674..

zeertzjq:
Hmm I just tested it and the condition does seem wrong. I'll open a PR to remove it.

zeertzjq:
```vim
function! TestCompletion(findstart, base) abort
  if a:findstart
    let line = getline('.')
    let start = col('.') - 1
    while start > 0 && line[start - 1] =~ '\a'
      let start -= 1
    endwhile
    return start
  else
    let ret = []
    for m in split("January February March April May June July August September October November December")
      if m =~ a:base  " match by regex
        call add(ret, m)
      endif
    endfor
    return {'words':ret, 'refresh':'always'}
  endif
endfunction

set completeopt=menuone,noselect
set completefunc=TestCompletion
```
Hmm, with the script above, it seems that the condition affects the following situation:
1. Type `iJ<C-X><C-U>`
2. Type `u`
3. Type `<CR>`

With the condition, the `<CR>` inserts a newline, which is correct. Without the condition, the `<CR>` just closes the completion menu. In Vim, the `<CR>` inserts a newline, however the completion menu isn't show after typing `u` in the first place.

Although that doesn't mean `compl_was_interrupted` can't be excluded from the condition.

glepnir:
smart

test/old/testdir/test_ins_complete.vim
```diff
@@ -2664,6 +2664,12 @@ func Test_complete_fuzzy_match()
   call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
   call assert_equal('hello help hero h', getline('.'))
 
+  set completeopt=fuzzy,menuone,menu,noselect
+  call setline(1, ['Text', 'ToText', ''])
+  call cursor(2, 1)
```
glepnir, line 2669: It should be 3 but I made a mistake. I will fix it in follow up PR. Then port into #30189.

zeertzjq (approved):

PR #30141, opened 2024-08-26
fix(filetype): move fallback logic to `vim.filetype.match()`

### LINKED ISSUE #30100: vim.filetype.match doesn't detect file type but :filetype detect does
### Problem

Sometimes, vim is able to automatically determine the filetype of a file and highlight it correctly, but `vim.filetype.match` does not detect it

### Steps to reproduce

Open a file with the following contents:

```conf
# foo
bar "baz"
```

`:set filetype?` shows the filetype as `conf`, and the syntax is highlighted correctly, but `vim.filetype.match({buf = 0})` returns `nil`.

### Expected behavior

`vim.filetype.match` should detect the filetype when `:filetype` is able to detect it

### Neovim version (nvim -v)

NVIM v0.11.0-dev+599-g8df6736ca

N/a, related to the nvim api

### Operating system/version

Fedora Silverblue 40

### Terminal name/version

wezterm 20240203-110809-5046fc22

xterm-256color

### Installation

nvim-nightly Fedora Copr

JonnyKong:
Problem:
Previously, the fallback logic to ".conf" was located outside of `vim.filetype.match()` and directly within the AutoCmd definition. As a result, `vim.filetype.match()` would return nil instead of ".conf" for fallback cases (https://github.com/neovim/neovim/issues/30100).

Solution:
Added a boolean return value to `vim.filetype.match()` that indicates whether the match was the result of fallback. If true, the filetype will be set using `setf FALLBACK <ft>` instead of `setf <ft>`.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2414,6 +2414,11 @@ vim.filetype.match({args})                              *vim.filetype.match()*
         (`function?`) A function that modifies buffer state when called (for
         example, to set some filetype specific buffer variables). The function
         accepts a buffer number as its only argument.
+        (`boolean?`) Return true if a match was found by falling back to a
+        generic configuration file (i.e., ".conf"). If true, the filetype
+        should be set with `:setf FALLBACK conf`, which enables a later
+        |:setf| command to override the filetype. See `:help setf` for more
+        information.
 
 
 ==============================================================================
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2415,7 +2415,7 @@ vim.filetype.match({args})                              *vim.filetype.match()*
         example, to set some filetype specific buffer variables). The function
         accepts a buffer number as its only argument.
         (`boolean?`) Return true if a match was found by falling back to a
-        generic configuration file (i.e., ".conf"). If true, the filetype
+        generic filetype (e.g. ".conf"). If true, the filetype
         should be set with `:setf FALLBACK conf`, which enables a later
         |:setf| command to override the filetype. See `:help setf` for more
         information.
--- a/runtime/filetype.lua
+++ b/runtime/filetype.lua
@@ -11,21 +11,13 @@ vim.api.nvim_create_autocmd({ 'BufRead', 'BufNewFile', 'StdinReadPost' }, {
     if not vim.api.nvim_buf_is_valid(args.buf) then
       return
     end
-    local ft, on_detect = vim.filetype.match({
+    local ft, on_detect, is_fallback = vim.filetype.match({
       -- The unexpanded file name is needed here. #27914
       -- However, bufname() can't be used, as it doesn't work with :doautocmd. #31306
       filename = args.file,
       buf = args.buf,
     })
-    if not ft then
-      -- Generic configuration file used as fallback
-      ft = require('vim.filetype.detect').conf(args.file, args.buf)
-      if ft then
-        vim._with({ buf = args.buf }, function()
-          vim.api.nvim_cmd({ cmd = 'setf', args = { 'FALLBACK', ft } }, {})
-        end)
-      end
-    else
+    if ft then
       -- on_detect is called before setting the filetype so that it can set any buffer local
       -- variables that may be used the filetype's ftplugin
       if on_detect then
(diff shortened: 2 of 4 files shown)


gpanders (commented):
Implementation looks ok to me, just some feedback on documentation.

runtime/doc/lua.txt
```diff
@@ -2783,6 +2783,7 @@ vim.filetype.match({args})                              *vim.filetype.match()*
         (`function?`) A function that modifies buffer state when called (for
         example, to set some filetype specific buffer variables). The function
         accepts a buffer number as its only argument.
+        (`boolean?`) True if a match was found via fallback
```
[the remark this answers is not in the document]
justinmk: Also seems like there is no test coverage?

JonnyKong:
Thanks a lot @gpanders @justinmk for the valuable feedback. I've amended the commit to address your comments. 🙂

FYI, it seems the failing macos clang functionaltest is unrelated; That test is also failing on the master branch.

clason:
Yes, that test is flaky and fails all the time. You can ignore it.

gpanders (approved):
Looks good, thanks!

justinmk (approved):
Please explain the problem being solved in 1-2 sentences.

Waiting for @clason or another FileType owner to merge this

clason:
In particular, I would like to know the rationale behind the added return value, which seems unrelated to the issue this solves? (The rest of the changes -- moving the fallback into vim.filetype.match -- looks good to me.)

JonnyKong:
> Please explain the problem being solved in 1-2 sentences.

Hi @justinmk -- Previously, the fallback logic to ".conf" was located outside of `vim.filetype.match()` and directly within the AutoCmd definition. As a result, `vim.filetype.match()` would return nil instead of .conf for fallback cases (#30100).

> the rationale behind the added return value

Hi @clason -- The added return value is needed to choose between `setf <ft>` or `setf FALLBACK <ft>`:

https://github.com/neovim/neovim/blob/ed76447fb6530604c79da8aa5a66dfb6b31a2b8e/runtime/filetype.lua#L28-L31

clason:
Can you document this _in the commit message_ and PR description? I'll have to think about whether such a boolean return is the best way to handle this.

justinmk:
Will help to keep this rebased

runtime/doc/lua.txt
```diff
@@ -2870,6 +2870,10 @@ vim.filetype.match({args})                              *vim.filetype.match()*
         (`function?`) A function that modifies buffer state when called (for
         example, to set some filetype specific buffer variables). The function
         accepts a buffer number as its only argument.
+        (`boolean?`) Return true if a match was found by falling back to a
+        generic configuration file (".conf"). If true, the filetype should be
+        set with `:setf FALLBACK conf`, which enables a later `:setf` command
+        to override the filetype. See `:help setf` for more information.
```
justinmk, line 2876: is ".confg" an example?
```suggestion
        generic configuration file (e.g. ".conf"). If true, the filetype should
        be set with `:setf FALLBACK conf`, which enables a later |:setf| command
        to override the filetype.
```

runtime/doc/lua.txt
```diff
@@ -2870,6 +2870,10 @@ vim.filetype.match({args})                              *vim.filetype.match()*
         (`function?`) A function that modifies buffer state when called (for
         example, to set some filetype specific buffer variables). The function
         accepts a buffer number as its only argument.
+        (`boolean?`) Return true if a match was found by falling back to a
```
justinmk, line 2873: Should it be a string in case there are other variants in the future?
```suggestion
        (`string?`) "FALLBACK" if a match was found by falling back to a
```

justinmk:
do you plan to finish this?

JonnyKong:
Hi @justinmk, thanks for your attention to this PR.

Do you still recommend your proposed changes (on Dec 20) based on my earlier response? Thanks!

justinmk:
https://github.com/neovim/neovim/pull/30141#discussion_r1893807433 "FALLBACK" is not a blocker, and on reflection the current boolean is probably fine.

the [docs clarification](https://github.com/neovim/neovim/pull/30141#discussion_r1893806639) seems worthwhile (but please use your judgement).

please rebase then we can merge this, thanks!

JonnyKong:
Rebased.

runtime/doc/lua.txt
```diff
@@ -2414,6 +2414,11 @@ vim.filetype.match({args})                              *vim.filetype.match()*
...
+        generic configuration file (i.e., ".conf"). If true, the filetype
```
justinmk: Is this feature specific to config files? That would be weird. If not, let's make that clear (the "??" is asking you to put a reasonable value in there):

```suggestion
        generic filetype (e.g. ".conf"). If true, the filetype
```

PR #30145, opened 2024-08-26
fix(lsp): return call hierarchy item, not the index

### LINKED ISSUE #30122: LSP incoming/outgoing calls triggers error when selecting from multiple symbols at cursor
### Problem

When calling `vim.lsp.buf.incoming_calls()` or `vim.lsp.buf.outgoing_calls()` when more than one candidate symbol matches at the cursor, `runtime/lua/vim/lsp/buf.lua::pick_call_hierarchy_item()` asks the user to select one via `vim.fn.inputlist()`, then returns the index instead of the value, which gets passed to the language server, triggering an error like this:
```
bugged-ls: -32602: Failed to deserialize callHierarchy/incomingCalls: invalid type: integer `1`, expected struct CallHierarchyItem; {"item":1}
```
This error comes from `rust-analyzer`, but most or all language servers are likely to complain in some way, for obvious reasons.

The offending line is line 467 here:
https://github.com/neovim/neovim/blob/6af9ca4926e05ac698d828ab6aeef2f45ccccdf8/runtime/lua/vim/lsp/buf.lua#L463-L467

This patch fixes the issue:
```patch
--- runtime/lua/vim/lsp/buf.lua	2024-08-23 14:33:38.790772646 +1200
+++ runtime/lua/vim/lsp/buf.lua	2024-08-23 14:35:39.631982250 +1200
@@ -464,7 +464,7 @@
   if choice < 1 or choice > #items then
     return
   end
-  return choice
+  return call_hierarchy_items[choice]
 end
 
 --- @param method string
```

This assumes `rust-analyzer` is installed and accessible in `$PATH`, but the issue should be reproducible with any language server, provided the language you are testing with can express a function call with more than one possible symbol resolution.

In a suitable scratch directory, create these files:

`minimal_init.lua`
```lua
local pattern = 'rust'
local cmd = {'rust-analyzer'}
local root_marker = 'Cargo.toml'

  pattern = pattern,
  callback = function(args)
    local match = vim.fs.find(root_marker, { path = args.file, upward = true })[1]
    vim.lsp.start({
      name = 'bugged-ls',
      cmd = cmd,
      root_dir = root_dir,
    })
  end
})
```

`Cargo.toml`
```toml
[package]
name = "nvim_lsp_repro"
version = "0.1.0"
edition = "2021"
```

`src/lib.rs`
```rust
macro_rules! make_fn {
    ($block:block) => {
        pub fn foo_func() $block
    }
}

make_fn!({});

pub fn calls_func() {
    foo_func();
}
```

Run `nvim --clean -u minimal_init.lua src/lib.rs`, then:
1. Place the cursor on the call to `foo_func()` (`131go`)
2. Run `:lua vim.lsp.buf.incoming_calls()`
3. Select either of the two list items with `1<cr>` or `2<cr>`

### Expected behavior

_No response_

### Neovim version (nvim -v)

0.10.1

rust-analyzer 1.80.0-nightly (791adf7 2024-05-21)

### Operating system/version

Arch Linux up-to-date as of 2024-08-18

### Log file

_No response_

MariaSolOs:
Fixes #30122

### DIFF
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -464,7 +464,7 @@ local function pick_call_hierarchy_item(call_hierarchy_items)
   if choice < 1 or choice > #items then
     return
   end
-  return choice
+  return call_hierarchy_items[choice]
 end
 
 --- @param method string


MariaSolOs:
Needs to be backported.

PR #27490, opened 2024-02-16
build: do not use cmake -S and -B flags in Makefile

dundargoc:
This partially reverts some changes in
ae3eed53d6100598b6d26fe58e3e97541e03f3c1.

Using -S and -B instead of cd:ing causes strange problems for some
users, where the `build` and `.deps` directory aren't created and the
process aborts with `make: cmake: Permission denied`. The cause of this
hasn't been identified yet.

Closes https://github.com/neovim/neovim/issues/27475.
Closes https://github.com/neovim/neovim/issues/27488.

PR #30146, opened 2024-08-26
build: work around bug in `make` when `PATH` includes `cmake` as dir

jez:
There appears to be a bug in `make` where if a rule asks `make` to
invoke a command called `foo`, and `foo` exists somewhere in `$PATH` as
a directory (not an executable file), `make` will attempt to `execve`
that directory instead of continuing to search in later parts of the
`$PATH` for `foo` as a true executable.

You can confirm this as follows:

    mkdir tmp
    cd tmp
    tmpdir="$(mktemp -d)"
    mkdir "$tmpdir/touch"
    export PATH="$tmpdir:$PATH"
    echo $'all:\n\ttouch foo.txt' > Makefile
    make

On my system, with `GNU Make 4.3` on Pop!\_OS 22.04, I'm greeted with
this output:

    touch foo.txt
    make: touch: Permission denied
    make: *** [Makefile:2: all] Error 127

A similar problem affected the user who opened #27490, but that issue
was closed because the problem was misdiagnosed (it was claimed that the
user was at fault for having their `$PATH` configured to include `.`).
In reality, this appears to be a bug in GNU Make--a Unix shell would not
mistake a directory (which technically have the execute permission set)
with a runnable file, while GNU Make does appear to do that.

It's easy enough to workaround this bug by using `command -v` to resolve
the `cmake` command, similar to what is already being done with
`cmake3`. Continuing to include `... || echo cmake` at the end means
that if neither `cmake3` nor `cmake` are installed, the user will still
see a message about CMake being missing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ filter-true = $(strip $(filter-out 1 on ON true TRUE,$1))
 
 all: nvim
 
-CMAKE ?= $(shell (command -v cmake3 || echo cmake))
+CMAKE ?= $(shell (command -v cmake3 || command -v cmake || echo cmake))
 CMAKE_FLAGS := -DCMAKE_BUILD_TYPE=$(CMAKE_BUILD_TYPE)
 # Extra CMake flags which extend the default set
 CMAKE_EXTRA_FLAGS ?=


dundargoc:
The problem was that *prepending* `.` to PATH, instead of *appending* it, was the real problem. I found a quick link on the problems with this: https://cets.seas.upenn.edu/answers/dot-path.html.

The problem with this PR is that it does not solve it the actual problem, it only provides a workaround for "cmake". If we wanted to (for some reason) create a directory named "ninja", "ls" or any other name that clashes with an executable, then that would cause the same problem. The crux of your argument is that prepending . to PATH is a perfectly legitimate usecase/workflow, but I've not seen any indications of this. In fact, doing so is a huge security risk because the user never really knows if they're running a command from some unsecured folder or a normal system command.

jez:
Thanks for taking the time to respond!

> The problem was that _prepending_ `.` to PATH, instead of _appending_ it, was the real problem. I found a quick link on the problems with this: https://cets.seas.upenn.edu/answers/dot-path.html.

You'll notice that the steps I included above did not require including `.` on the `PATH` to reproduce—using `.` on the PATH is a red herring!

> The crux of your argument is that prepending . to PATH is a perfectly legitimate usecase/workflow

Nope, even if I happened to have some `cmake` folder somewhere in a directory that was already on my path, this behavior would still happen.

> The problem with this PR is that it does not solve it the actual problem, it only provides a workaround for "cmake".

Yep, this is right!

But at this point I figured that I was not going to be able to convince you otherwise that it might be worth working around this in the Neovim build files (because this `make` bug had affected me plus at least one other person I thought maybe it would be worth working around, so that the "first Neovim contributor" experience might be a little smoother.

No worries, I understand not wanting to add code that is merely a workaround and not a root cause solution.

With that in mind I set out to find the root cause, and I traced it back to a bug in Make 4.3 which stems from their use of the `findprog` function in Gnulib. This was reported to the Make maintainers here:

https://savannah.gnu.org/bugs/index.php?57962

and then forwarded to the Gnulib maintainers here:

https://github.com/coreutils/gnulib/commit/7b1de4a62f8766787160f785217671b074e0f0f2?w=1

Make 4.4 does not have this bug, and I can confirm that I'm able to run `make` in the Neovim repo with no further modifications to my system than upgrading the version of `make` I'm using to 4.4 or 4.4.1.

Thanks again,
Jake Z.

dundargoc:
> Nope, even if I happened to have some cmake folder somewhere in a directory that was already on my path, this behavior would still happen.

Hmm, yeah, you are right actually.

> But at this point I figured that I was not going to be able to convince you otherwise that it might be worth working around this in the Neovim build files. (Because this make bug had affected me plus at least one other person, I thought maybe it would be worth working around so that the "first Neovim contributor" experience might be a little smoother).

Maybe chill? I would like to have the opportunity to ask some questions to clarify things without people packing their bags and deserting. I think with the added context of the make version there was a bug in the commit message/PR description that it's worth having this workaround.

dundargoc:
Would you mind reopening and adding the context of your closing comment to the commit message? It is a workaround but it's cheap enough that I think it could be worth it.

bwalshe:
I was hit by this today on my Ubuntu 24.04.1 LTS computer.  It might be worth having a note about this somewhere more prominent as this was a pretty confusing issue for me at least.

dundargoc:
Extracted commit and the relevant comment information to https://github.com/neovim/neovim/pull/30432

PR #30155, opened 2024-08-27
Add missing translations and fix a typo

VernalUmbrella:
I noticed the Japanese vimtutor has a few lines still in English. I chose to ignore (for now) the practice lines which the user is intended to interact with.
I'm not a native Japanese speaker, but I got firsthand confirmation from one that these edits sound natural.


justinmk:
why still in draft?

PR #30176, opened 2024-08-29
docs(eval): wrong return type of getcharsearch

### LINKED ISSUE #30174: lua: `vim.fn.getcharsearch()` has incorrect return type
### Problem

Minor issue but `vim.fn.getcharsearch()` has an incorrect return type. It's supposed to be `table` and not `table[]`. I was going to send a PR but I couldn't find/figure out what to change (where the type is defined/derived from). I figure this is probably a quick fix.

### Steps to reproduce

n/a

### Expected behavior

n/a

### Neovim version (nvim -v)

0.10.1

n/a

### Operating system/version

n/a

### Terminal name/version

n/a

n/a

### Installation

nixpkgs-unstable

glepnir:
Problem: return type of getcharsearch in eval.lua is not correct.

Solution: change to table.

Fix #30174

### DIFF
--- a/runtime/lua/vim/_meta/vimfn.lua
+++ b/runtime/lua/vim/_meta/vimfn.lua
@@ -2861,7 +2861,7 @@ function vim.fn.getcharpos(expr) end
 ---   nnoremap <expr> , getcharsearch().forward ? ',' : ';'
 --- <Also see |setcharsearch()|.
 ---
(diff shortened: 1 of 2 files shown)


zeertzjq:
`char` is not an integer

glepnir:
oops yup.

PR #30177, opened 2024-08-29
docs: misc

dundargoc:


CMakeLists.txt
```diff
@@ -5,6 +5,15 @@
...
 #     - variable_watch https://cmake.org/cmake/help/latest/command/variable_watch.html
 
+# We currently have no systematic way to determine the minimum required
+# version. A useful rule of thumb to ensure we don't drop support for too many
+# users is to follow the "End of Standard Support" in
+# https://ubuntu.com/about/release-cycle and then compare it with
+# https://packages.ubuntu.com/search?keywords=cmake&searchon=names&suite=all&section=all
+# to see which cmake versions ubuntu still supports. We may want to enforce
+# this rule in the future, but for the time being we still ultimately decide
+# based on cost/benefit analysis for each individual case.
```
dundargoc, line 15: Could add in `MAINTAIN.md` if it fits there better.

BUILD.md
```diff
@@ -284,7 +283,7 @@ General requirements (see [#1469](https://github.com/neovim/neovim/issues/1469#i
 
 - Clang or GCC version 4.9+
 - CMake version 3.16+, built with TLS/SSL support
-  - Optional: Get the latest CMake from an [installer](https://github.com/Kitware/CMake/releases) or the [Python package](https://pypi.org/project/cmake/) (`pip install cmake`)
+  - Optional: Get the latest CMake from https://cmake.org/download/.
```
justinmk, line 286:
```suggestion
  - Optional: Get the latest CMake from https://cmake.org/download/
    - Provides a shell script which works on most Linux systems. After running it, ensure the resulting `cmake` binary is in your $PATH so the the Nvim build will find it.
```

Also not sure if cmake is smart enough to keep looking in $PATH if it finds an outdated `cmake` before finding the newer one. If not, we should mention that the downloaded `cmake` should be the first item in $PATH.

BUILD.md
```diff
@@ -12,7 +12,6 @@
     - To build on Windows, see the [Building on Windows](#building-on-windows) section. _MSVC (Visual Studio) is recommended._
 4. `sudo make install`
     - Default install location is `/usr/local`
-    - On Debian/Ubuntu, instead of installing files directly with `sudo make install`, you can run `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This should help ensuring the clean removal of installed files.
```
justinmk, line 15: Why not mention this?

dundargoc: I thought we concluded deb packages was ultracringe because it doesn't really have any granularity with other dependencies and that it's just a shitty container.

justinmk: We don't provide a deb (for good reason), but mentioning `cpack -G DEB` is a nice hint for people that want to explore it. 

I don't remember the exact problems caused by the deb, did it leave the system in a bad state?

dundargoc: Not really, from my conversations with ubuntu users this was never really what they wanted since it didn't update neovim when upgrading with apt. I don't think anyone really uses this, nor does there seem to be any good reason to honestly.

dundargoc: Can re-add if want, I'm just going by gut feeling here.

justinmk: generally it's high-leverage when we can give a hint about an easy (one-liner) way to unlock powerful but non-obvious things (in this case, cmake generators are not typically known), but if it's not useful then can drop it.

dundargoc: Makes sense. I still have a suspicion that this is a net negative because it will probably give incorrect hint for the majority of users that would be better served by using a package manager like apt/PPA, but are simply not aware of it. Let's remove it for the time being unless someone protests or it becomes relevant in the future.

PR #30192, opened 2024-08-30
feat(ui): gx: use url extmark attribute and tree-sitter directive

gpanders:
Use the "url" extmark attribute as well as the "url" tree-sitter metadata key to determine if the cursor is over something Nvim considers a URL.

### DIFF
--- a/runtime/lua/vim/ui.lua
+++ b/runtime/lua/vim/ui.lua
@@ -167,29 +167,63 @@ function M.open(path)
   return vim.system(cmd, opts), nil
 end
 
---- Gets the URL at cursor, if any.
-function M._get_url()
-  if vim.bo.filetype == 'markdown' then
-    local range = vim.api.nvim_win_get_cursor(0)
-    vim.treesitter.get_parser():parse(range)
-    -- marking the node as `markdown_inline` is required. Setting it to `markdown` does not
-    -- work.
-    local current_node = vim.treesitter.get_node { lang = 'markdown_inline' }
-    while current_node do
-      local type = current_node:type()
-      if type == 'inline_link' or type == 'image' then
-        local child = assert(current_node:named_child(1))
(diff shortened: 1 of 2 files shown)


justinmk (approved):
To confirm, we no longer need to use the `markdown_inline` approach because in a markdown file, any urls will be marked with the `url` attribute? 

So besides adding support for `:terminal` URLs, this also adds support for all treesitter grammars (not only markdown) that annotate URLs?

Nice!

gpanders:
> To confirm, we no longer need to use the `markdown_inline` approach because in a markdown file, any urls will be marked with the `url` attribute?

Correct.

> 
> So besides adding support for `:terminal` URLs, this also adds support for all treesitter grammars (not only markdown) that annotate URLs?

Unfortunately this _does not_ add support for `:terminal` URLS :( `nvim_buf_get_extmark` does not retrieve ephemeral extmarks, so this won't apply for terminal buffers. Tree-sitter also uses ephemeral extmarks, but in that case we can use the query itself to get the `url` attribute.

runtime/lua/vim/ui.lua
```diff
@@ -167,29 +167,65 @@ function M.open(path)
...
+  local highlighter = vim.treesitter.highlighter.active[bufnr]
+  if highlighter then
+    local range = { row, col, row, col }
+    local ltree = highlighter.tree:language_for_range(range)
+    local lang = ltree:lang()
+    local query = vim.treesitter.query.get(lang, 'highlights')
+    if query then
+      local tree = ltree:tree_for_range(range)
+      for pattern, match, metadata in
+        query:iter_matches(tree:root(), bufnr, row, row + 1, { all = true })
```
gpanders, line 200: If https://github.com/neovim/neovim/pull/30193 merges first, be sure to remove `{ all = true }`

PR #30193, opened 2024-08-30
feat(treesitter)!: make correct match behavior for quantified captures the default

gpanders:
For context, see https://github.com/neovim/neovim/pull/24738. Before that PR, Nvim did not correctly handle captures with quantifiers. That PR made the correct behavior opt-in to minimize breaking changes, with the intention that the correct behavior would eventually become the default. Users can still opt-in to the old (incorrect) behavior for now, but this option will eventually be removed completely.

BREAKING CHANGE: Any plugin which uses `Query:iter_matches()` must update their call sites to expect an array of nodes in the `match` table, rather than a single node.

## Migration guide

**Before**

```lua
for pattern, match, metadata in query:iter_matches(...) do
  for id, node in pairs(match) do
    -- Do something with node
  end
end
```

**After**

```lua
for pattern, match, metadata in query:iter_matches(...) do
  for id, nodes in pairs(match) do
    for _, node in ipairs(nodes) do
      -- Do something with node
    end
  end
end
```

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -89,7 +89,11 @@ PLUGINS
 
 TREESITTER
 
-• TODO
+• |Query:iter_matches()| correctly returns all matching nodes in a match
+  instead of only the last node. This means that the returned table maps
+  capture IDs to a list of nodes that need to be iterated over. For
+  backwards compatibility, an option `all=false` (only return the last
+  matching node) is provided that will be removed in a future release.
 
 TUI
 
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -1036,9 +1036,8 @@ add_directive({name}, {handler}, {opts})
                      the same name
                    • {all}? (`boolean`) Use the correct implementation of the
                      match table where capture IDs map to a list of nodes
-                     instead of a single node. Defaults to false (for backward
-                     compatibility). This option will eventually become the
-                     default and removed.
+                     instead of a single node. Defaults to true. This option
+                     will be removed in a future release.
 
                                         *vim.treesitter.query.add_predicate()*
 add_predicate({name}, {handler}, {opts})
@@ -1049,14 +1048,13 @@ add_predicate({name}, {handler}, {opts})
       • {handler}  (`fun(match: table<integer,TSNode[]>, pattern: integer, source: integer|string, predicate: any[], metadata: vim.treesitter.query.TSMetadata)`)
                    • see |vim.treesitter.query.add_directive()| for argument
                      meanings
-      • {opts}     (`table`) A table with the following fields:
+      • {opts}     (`table?`) A table with the following fields:
                    • {force}? (`boolean`) Override an existing predicate of
                      the same name
                    • {all}? (`boolean`) Use the correct implementation of the
                      match table where capture IDs map to a list of nodes
-                     instead of a single node. Defaults to false (for backward
-                     compatibility). This option will eventually become the
-                     default and removed.
+                     instead of a single node. Defaults to true. This option
+                     will be removed in a future release.
 
 edit({lang})                                     *vim.treesitter.query.edit()*
     Opens a live editor to query the buffer you started from.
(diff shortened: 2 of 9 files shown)


clason (approved):
Change LGTM (should not be breaking to downstream as of now), but documentation is a bit wordy maybe and could be boiled down? (I don't think we need to call out the specific version in the docstring, for example; that's what `news.txt` is for.)

gpanders:
>I don't think we need to call out the specific version in the docstring, for example; that's what news.txt is for.

The old docstring did that but was removed in this PR. I admit the news blurb is a bit wordy, but I'm not sure how to pare it down without either reducing clarity or eliminating remediation steps (which is valuable for breaking change notes).

runtime/doc/news.txt
```diff
@@ -89,7 +89,13 @@ PLUGINS
 
 TREESITTER
 
-• TODO
+• |Query:iter_matches()| now defaults to returning all matching nodes in a
+  match, rather than simply the last node in a match. In particular, the
```
MariaSolOs, line 93: Nit, but I think this bit is redundant:
```suggestion
  match, rather than simply the last node. In particular, the
```

runtime/doc/treesitter.txt
```diff
@@ -1252,8 +1244,9 @@ Query:iter_matches({node}, {source}, {start}, {stop}, {opts})
                   • all (boolean) When set, the returned match table maps
                     capture IDs to a list of nodes. Older versions of
                     iter_matches incorrectly mapped capture IDs to a single
-                    node, which is incorrect behavior. This option will
-                    eventually become the default and removed.
+                    node, which is incorrect behavior. As of Nvim 0.11 this
```
clason, line 1247: I was referring to this bit here ☝️

gpanders: Thanks, updated.

runtime/doc/news.txt
gpanders: Changed as suggested.

runtime/lua/vim/treesitter/query.lua
```diff
@@ -925,9 +919,9 @@ end
 ---   - max_start_depth (integer) if non-zero, sets the maximum start depth
 ---     for each match. This is used to prevent traversing too deep into a tree.
 ---   - match_limit (integer) Set the maximum number of in-progress matches (Default: 256).
----   - all (boolean) When set, the returned match table maps capture IDs to a list of nodes.
----     Older versions of iter_matches incorrectly mapped capture IDs to a single node, which is
----     incorrect behavior. This option will eventually become the default and removed.
+---   - all (boolean) When set (the default), the returned match table maps capture IDs to a list of
+---     nodes. Older versions of iter_matches incorrectly mapped capture IDs to a single node, which
+---     is incorrect behavior. This option will be removed in a future release.
```
clason, line 924:
```suggestion
--- - all (boolean) When `false` (default `true`), the returned table maps capture IDs to a single
---   (last) node instead of the full list of matching nodes. This option is only for backward 
---   compatibility and will be removed in a future release.
```

runtime/doc/news.txt
```diff
@@ -89,7 +89,13 @@ PLUGINS
...
 TREESITTER
 
-• TODO
+• |Query:iter_matches()| now defaults to returning all matching nodes in a
+  match, rather than simply the last node. In particular, the `match` table
+  returned by |Query:iter_matches()| now maps capture IDs to a list of nodes,
+  rather than a single node. The old behavior is incorrect, but can be opted
+  into by setting `all=false` in the `opts` argument to
+  |Query:iter_matches()|. This option will be removed completely in a future
+  release.
```
clason, line 98:
```suggestion
• |Query:iter_matches()| correctly returns all matching nodes in a match
  instead of only the last node. This means that the returned table maps
  capture IDs to a list of nodes that need to be iterated over. For
  backwards compatibility, an option `all=false` (only return the last
  matching node) is provided that will be removed in a future release.
```

clason (commented):
Suggestions for condensed text; up to you if you want to include (parts of) them; feel free to merge then.

runtime/lua/vim/treesitter/query.lua
gpanders: Changed as suggested.

runtime/doc/news.txt
gpanders: Changed as suggested.

glepnir:
after this patch `TSNode:parent` can be nil ..

```
Error in decoration provider treesitter/highlighter.line:
Error executing lua: ...nvim-treesitter/lua/nvim-treesitter/query_predicates.lua:80: attempt to call method 'parent' (a nil value)
````

clason:
Nvim-treesitter problem; looks like the (compatibility) predicate needs to be adapted (or, better yet, removed).

glepnir:
aha sry for noise then hide comment ..

PR #30202, opened 2024-08-31
docs(syntax): use `mkdir -p` to save a command

ada4a:


clason:
Thank you, but we don't maintain these files. Please open a PR at vim/vim, and we'll port any changes made.

PR #30203, opened 2024-08-31
feat(treesitter): return supertypes in language.inspect()

ribru17:
This makes sense imo since they can still be queried as `(supertype_name) @capture`.

This commit just puts them with the other symbols, in that same table.

### DIFF
--- a/src/nvim/lua/treesitter.c
+++ b/src/nvim/lua/treesitter.c
@@ -13,6 +13,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <tree_sitter/api.h>
+#include <tree_sitter/parser.h>
 #include <uv.h>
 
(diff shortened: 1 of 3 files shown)


src/nvim/lua/treesitter.c
```diff
@@ -13,6 +13,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <tree_sitter/api.h>
+#include <tree_sitter/parser.h>
```
clason, line 16: This is not part of the public tree-sitter API (it's meant for parsers, not consumers), afaik. We would have to hack the cmake file to pull in that header as well.

ribru17: Makes sense, maybe not worth it for such a tiny change

PR #30215, opened 2024-09-01
feat(clipboard): try cygutils, clip on Windows

Freed-Wu:


runtime/autoload/provider/clipboard.vim
```diff
@@ -140,6 +140,18 @@ function! provider#clipboard#Executable() abort
...
     return 'win32yank'
+  elseif executable('putclip') && executable('getclip')
+    let s:copy['+'] = ['putclip']
+    let s:paste['+'] = ['getclip']
+    let s:copy['*'] = s:copy['+']
+    let s:paste['*'] = s:paste['+']
+    return 'putclip'
+  elseif executable('clip') && executable('powershell')
+    let s:copy['+'] = ['clip']
+    let s:paste['+'] = ['powershell', '-command', 'Get-Clipboard']
```
justinmk, line 151: can you confirm that `-NoProfile -NoLogo` works and update this if so? `-NoProfile` should help with performance.
```suggestion
    let s:paste['+'] = ['powershell', '-NoProfile', '-NoLogo', '-command', 'Get-Clipboard']
```

PR #15380, opened 2021-08-15
fix(screen): missing search highlights when redrawing from timer

gelguy:
Fixes https://github.com/neovim/neovim/issues/14064.

I'm actually not sure of the effects of this PR: I reverted subsets of the changes in 910bbc3cca796f7fa941e0f6176cd0061de0e01c and removing these lines fixed the issue. Functional tests all pass so I assume nothing major breaks.

I wasn't able to write a lua test that fails either. I tried adding a test which tries the steps in #14064 to `functional/ui/searchhl_spec.lua` but the test doesn't fail before the changes are applied.
```vim
  it('correctly applies highlighting when redrawing in timer', function()
    feed_command('set hlsearch')
    feed_command('set incsearch')

    insert([[






      line 0
      line 1
      line 2
      line 3
      line 4
      line 5
      line 6
    ]])

    command("autocmd CmdlineChanged * call timer_start(0, {-> execute('redraw')})")

    feed("gg/0")
    screen:expect([[
                                              |
                                              |
                                              |
                                              |
                                              |
        line {3:0}                                |
      /0^                                      |
    ]])
  end)
```

IMO the whole patch can be reverted - I manually performed the actions in the Vim test and it doesn't fail even if the patch isn't applied.

PR #30217, opened 2024-09-01
vim-patch:9.1.0708: Recursive window update does not account for reset skipcol

luukvbaal:
Problem:  Window is updated with potentially invalid skipcol in recursive
          window update path. I.e. cursor outside of visible range in
          large line that does not fit.
Solution: Make sure it is valid (Luuk van Baal).

https://github.com/vim/vim/commit/3d5065fc7553c8de3d0555c3085165bdd724663b


luukvbaal:
It seems this code has fallen out of sync with Vim so the comment diff couldn't really be ported. I'll try to figure out if porting an earlier patch is appropriate.

zeertzjq:
It was reverted in #15380

zeertzjq (approved):

PR #30228, opened 2024-09-02
fix(diagnostic): add cursorlinehl in vim.diagnotic.opts.Signs

### LINKED ISSUE #30226: Support cursorline_hl_group for diagnostic signs
### Problem

Currently, for `vim.diagnostic` we may configure `numhl` and `linehl` properties. These configuration values are "forwarded" to `nvim_buf_set_extmark()`'s `number_hl_group` and `line_hl_group` options. However, `cursorline_hl_group` is a valid option to `nvim_buf_set_extmark` as well, and is necessary to get a coherent cursorline extending through the signcolumn. 

Below is a picture of the problem: (We desire for the background of the sign to take on the CursorLineSign background as well.

Screenshot 2024-09-01 at 1 37 58 PM 

### Expected behavior

Proposed solution: we add `culhl` configuration property for `vim.diagnostic.config({})`, mapping diagnostic severity to the highlight group used for the sign when the cursorline is on the line. Just like `numhl` and `linehl`, we forward this value to `cursorline_hl_group` option of `nvim_buf_set_extmark()` here:

https://github.com/neovim/neovim/blob/master/runtime/lua/vim/diagnostic.lua#L1455-L1461

My personal opinion is this is a small, low-cost change that does not add complexity to the codebase. Please consider!

Gitsigns.nvim added support for culhl, so this is the last frontier for me to get that nice coherent cursorline :]

glepnir:
Problem: extmark also support cursorline_hl_group when cursorline is on.

Solution: support it on vim.diangostic.optss.Signs

Fix #30226

### DIFF
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -526,28 +526,32 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
 *vim.diagnostic.Opts.Signs*
 
     Fields: ~
-      • {severity}?  (`vim.diagnostic.SeverityFilter`) Only show virtual text
-                     for diagnostics matching the given severity
-                     |diagnostic-severity|
-      • {priority}?  (`integer`, default: `10`) Base priority to use for
-                     signs. When {severity_sort} is used, the priority of a
-                     sign is adjusted based on its severity. Otherwise, all
-                     signs use the same priority.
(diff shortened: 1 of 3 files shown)


justinmk:
still needed?

glepnir:
`cursorlinehl` is a field of `nvim_buf_set_extmark` . But we have current_line in vim.diagnostic.config , so it probably doesn't work. ping @mizlan

PR #30232, opened 2024-09-02
fix(mbyte): mark any 0xFE0F sequence as a TUI ambiguous width char

bfredl:
Some sequences beginning with ASCII might be rendered as emoji, as for instance emoji 1️⃣  which is encoded as ascii 0x31 + U+FE0F + U+20E3.

While it is tricky to make the width of such sequences configurable, we can make TUI be careful with such sequences and reset the cursor, just like for Extended_Pictogram based sequences.

### DIFF
--- a/src/nvim/mbyte.c
+++ b/src/nvim/mbyte.c
@@ -1339,13 +1339,22 @@ int utf_class_tab(const int c, const uint64_t *const chartab)
 
 bool utf_ambiguous_width(const char *p)
 {
-  int c = utf_ptr2char(p);
-  if (c < 0x80) {
+  // be quick if there is nothing to print or ASCII-only
+  if (p[0] == NUL || p[1] == NUL) {
     return false;
   }
 
-  const utf8proc_property_t *prop = utf8proc_get_property(c);
-  return prop->ambiguous_width || prop_is_emojilike(prop);
+  CharInfo info = utf_ptr2CharInfo(p);
+  if (info.value >= 0x80) {
+    const utf8proc_property_t *prop = utf8proc_get_property(info.value);
+    if (prop->ambiguous_width || prop_is_emojilike(prop)) {
+      return true;
+    }
+  }
+
+  // check if second sequence is 0xFE0F VS-16 which can turn things into emoji,
+  // safe with NUL (no second sequence)
+  return memcmp(p + info.len, "\xef\xb8\x8f", 3) == 0;
 }
 
 // Return the folded-case equivalent of "a", which is a UCS-4 character.  Uses


src/nvim/mbyte.c
```diff
@@ -1339,13 +1339,22 @@ int utf_class_tab(const int c, const uint64_t *const chartab)
...
+  CharInfo info = utf_ptr2CharInfo(p);
+  if (info.value >= 0x80) {
+    const utf8proc_property_t *prop = utf8proc_get_property(info.value);
+    if (prop->ambiguous_width || prop_is_emojilike(prop)) {
+      return true;
+    }
+  }
+
+  // check if second sequence is 0xFE0F VS-16 which can turn things into emoji,
+  // safe with NUL (no secound sequence)
```
clason, line 1356:
```suggestion
  // safe with NUL (no second sequence)
```

zeertzjq:
This may lead to worse user experience in terminal emulators where drawing on the second half of a double-width char clears the first half, as the whole char will just be invisible there.

bfredl:
? how is that different from any other char already guarded by `utf_ambiguous_width` . The goalpost here is not to make emojis magically be rendered correctly even when nvim and the terminal disagrees about the width of a sequence, it is about avoiding redraw _glitches_ due to the cursor position getting out of sync:

image

GitMurf:
@clason sorry about commenting on the old PR RE: your comment here https://github.com/neovim/neovim/pull/30014#issuecomment-2324702784 ... understood!

Thanks for the quick work on this @bfredl et al. Let me know if there is any help you need testing it? I am on nightly as well so can update / test there too when / if it lands.

GitMurf:
Ok @bfredl I was able to get your branch to build on Windows. It looks like it has resolved the issue I was having before with throwing off the characters around the emoji and the lines close to it 👍 It does just render the emoji as the regular digit 1 (which I am fine with).

image

There are some other issues I noticed though mainly with visual selection as it seems to visually select the multi sequence emoji still as I can see the original rendered emoji sometimes when selecting over it... that is no problem for me. But I noticed that when you are visually selecting if you stop selection on / around the emoji then little cursor remnants get left behind. See the video below.

https://github.com/user-attachments/assets/fdd71816-50b4-486c-a7b1-d4b0e50b7cd6

GitMurf:
The concept of setcellwidth() works really well for one off issues with more "legacy" emojis (non complex multi codepoint sequence emojis). As you've mentioned, it doesn't really work with these more complex situations. What we would need is a way to either:

1. Set the width of a sequence of codepoints.
2. Or a way to set certain codepoints to 0 (instead of the current limitation of only allowable sizes of 1 or 2).

Are either of these possible and a potential solution?

zeertzjq:
The biggest problem here is that when the first char is ASCII, allowing 0xFE0F to change its width will cause performance degradations.

GitMurf:
> The biggest problem here is that when the first char is ASCII, allowing 0xFE0F to change its width will cause performance degradations.

But you already can change the width of 0xFE0F with setcellwidths(). Is my idea 1 an option? Where instead of only allowing setcellwidths() for individual codepoints, you could provide a combination of codepoints? So I could set "u31 + FE0F + 20E3" together and set to 2?

GitMurf:
Curious if using something like conceal would work here as a local workaround where if corner cases occur with emojis like this, users can be directed to replacing / concealing the bad emoji(s)? So I could create a concealer to replace this complicated emoji with something simpler that I know works.

Or as an alternative idea, would it be something to consider for neovim to provide a built-in function that could be used similar to setcellwidths() but instead a mapping table of "Emoji A" --> "Emoji B". So this table would just use emoji B in place of emoji A when loading a buffer that an emoji in your mapping table exists in.

Not sure if that is a crazy idea but just trying to contribute some thoughts from an end user perspective :-) 🤷🏻‍♂️

bfredl:
> Or a way to set certain codepoints to 0 (instead of the current limitation of only allowable sizes of 1 or 2).

This part is really handled as clustering, i e multiple codepoints becomes one cluster which then _as a cluster_ either have a width of one or two. Once codepoints are part of a cluster it doesn't really make sense to think of them as separate chars anymore (they only become separate when rendering glitches out).

Having 'emoji' option also apply to ` 0x31 + U+FE0F` is a reasonable long-term goal if this is how a lot of fonts render it in practice, but it is going to require quite a bit of refactors with perf considerations as zeer notes ( i e a lot of fast paths which now applies to ascii+combining will only apply to pure ASCII  with no combining codepoints or at very least ASCII followed by another ASCII char).

bfredl:
>  but instead a mapping table of "Emoji A" --> "Emoji B". So this table would just use emoji B in place of emoji A when loading a buffer that an emoji in your mapping table exists in.

This might be feasible strictly in the display layer. we already use a hash set to memoize long UTF-8 sequences which doesn't directly fit in the screen grid, it would be possible to extend it to a map which also can change the presentation (providing the width is the same for both sequences).

llimllib:
Thank you for all this work! I just wanted to let you know that it's still not quite right for me here following this PR:
Screenshot 2024-09-03 at 11 21 58 AM

in kitty, it's not displaying the number emojis at all

The contents of the file are:

```
🇺🇸✅1️⃣🇻🇳🤢🇹🇭2️⃣👨‍👩‍👦

```

bfredl:
https://mitchellh.com/writing/grapheme-clusters-in-terminals posting this for future reference (there is a table of terminals which are known to work or to not work)

PR #30237, opened 2024-09-02
feat(startup): validate --listen address

### LINKED ISSUE #30123: "nvim --headless --listen ..." does not unlink socket on signal exit/Ctrl-C
### Problem

If Neovim is started with `--headless --listen /some/unix/socket`, and `bind` returns `EADDRINUSE` because the socket file already exists, Neovim does not report any error and just continues running, but will not accept any connections on the socket.

### Steps to reproduce

`nvim --headless --listen ./hello.sock`
(Ctrl-C, check that ./hello.sock exists now)
`nvim --headless --listen ./hello.sock`
(Try to connect via `nvim --remote-ui --server ./hello.sock`)

If needed, use `strace nvim --headless --listen ./hello.sock |& grep hello` to verify that `bind` errors.

### Expected behavior

Neovim should handle this situation gracefully, either by exiting with an error, or trying to determine if the socket is actually used, and unlinking/recreating it if not.

### Neovim version (nvim -v)

NVIM v0.11.0-dev-655+g6af9ca492

N/A

### Operating system/version

Arch Linux

### Terminal name/version

wezterm

xterm-256color

### Installation

`neovim-nightly-bin` AUR package (downloads latest github release)

justinmk:
## Problem:
`nvim --listen` does not error on EADDRINUSE. #30123

## Solution:
Now that `$NVIM_LISTEN_ADDRESS` is deprecated and input *only* (instead of the old, ambiguous situation where it was both an input *and* an output), we can be fail fast instead of trying to "recover". This reverts the "recovery" behavior of 704ba4151e7f67999510ee0ac19fdabb595d530c, but that was basically a workaround for the fragility of `$NVIM_LISTEN_ADDRESS`. 

close #30123

### DIFF
--- a/src/nvim/main.c
+++ b/src/nvim/main.c
@@ -332,12 +332,6 @@ int main(int argc, char **argv)
 #endif
   bool use_builtin_ui = (has_term && !headless_mode && !embedded_mode && !silent_mode);
 
-  // don't bind the server yet, if we are using builtin ui.
-  // This will be done when nvim server has been forked from the ui process
-  if (!use_builtin_ui) {
-    server_init(params.listen_addr);
-  }
-
   if (params.remote) {
     remote_request(&params, params.remote, params.server_addr, argc, argv,
                    use_builtin_ui);
@@ -355,11 +349,19 @@ int main(int argc, char **argv)
     ui_client_channel_id = rv;
   }
 
+  // NORETURN: Start builtin UI client.
   if (ui_client_channel_id) {
     time_finish();
     ui_client_run(remote_ui);  // NORETURN
   }
   assert(!ui_client_channel_id && !use_builtin_ui);
+  // Nvim server...
+
+  int listen_rv = server_init(params.listen_addr);
+  if (listen_rv != 0) {
+    mainerr("Failed to --listen", listen_rv < 0
+            ? os_strerror(listen_rv) : (listen_rv == 1 ? "empty address" : NULL));
+  }
 
   TIME_MSG("expanding arguments");
 
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -800,8 +800,7 @@ describe('LSP', function()
           eq(table.remove(expected_handlers), { err, result, ctx }, 'expected handler')
           if ctx.method == 'start' then
             local tmpfile_old = tmpname()
-            local tmpfile_new = tmpname()
-            os.remove(tmpfile_new)
+            local tmpfile_new = tmpname(false)
             exec_lua(function(oldname, newname)
               _G.BUFFER = vim.api.nvim_get_current_buf()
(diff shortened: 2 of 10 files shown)


src/nvim/main.c
```diff
@@ -332,12 +332,6 @@ int main(int argc, char **argv)
 #endif
   bool use_builtin_ui = (has_term && !headless_mode && !embedded_mode && !silent_mode);
 
-  // don't bind the server yet, if we are using builtin ui.
-  // This will be done when nvim server has been forked from the ui process
-  if (!use_builtin_ui) {
-    server_init(params.listen_addr);
```
justinmk, line 338: Moved to after `ui_client_run`. That should be equivalent, right? Avoids the need for an explanatory comment and `!use_builtin_ui` condition.

src/nvim/main.c
```diff
@@ -355,11 +349,19 @@ int main(int argc, char **argv)
...
   if (ui_client_channel_id) {
     time_finish();
     ui_client_run(remote_ui);  // NORETURN
   }
   assert(!ui_client_channel_id && !use_builtin_ui);
+  // Nvim server...
+
+  int listen_rv = server_init(params.listen_addr);
+  if (listen_rv != 0) {
+    mainerr("Failed to --listen", listen_rv < 0
```
justinmk, line 362: Without `--headless`, this error message doesn't print nicely. But that same problem affects `nvim -W` validation and probably other startup validation.

TODO: if the server fails, can the UI client properly teardown the screen and print the server error?

test/functional/plugin/lsp_spec.lua
```diff
@@ -2388,9 +2386,7 @@ describe('LSP', function()
     it(
       'Supports file creation in folder that needs to be created with CreateFile payload',
       function()
-        local tmpfile = tmpname()
-        os.remove(tmpfile) -- Should not exist, only interested in a tmpname
-        tmpfile = tmpfile .. '/dummy/x/'
+        local tmpfile = tmpname(false) .. '/dummy/x/'
```
justinmk, line 2389: Looks like many tests need this functionality so I added a `create:boolean` param to `t.tmpname()`. Will followup in another PR to improve it.

Shougo:
When I use `nvim --listen ~/.cache/nvim/server.pipe` command, my neovim will exit.
It is intended behavior?

justinmk:
> When I use `nvim --listen ~/.cache/nvim/server.pipe` command, my neovim will exit.

It should show an error explaining why it exited. If you [can't see the error](https://github.com/neovim/neovim/pull/30237#discussion_r1741057134), try with `--headless`:

    nvim --headless --listen ~/.cache/nvim/server.pipe

Shougo:
```
% nvim --headless --listen ~/.cache/nvim/server.pipe
nvim: Failed to --listen: "address already in use"
```

When the address is free?

PR #30238, opened 2024-09-02
feat(lsp): support hostname in rpc.connect

### LINKED ISSUE #29513: LSP: vim.lsp.rpc.connect() with hostname instead of IP address
### Problem

Does anyone know if vim.lsp.rpc.connect() should work with a hostname or is it IP only.

It works perfectly with 127.0.0.1 but localhost returns the below. In the docs the field is called host, so I don't see anything that suggests it should be IP only. 

Anybody more familiar with the LUV TCP stuff?

Perhaps there should be a resolution step in rpc.lua to be able to function with hostnames too?

```/usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:706: Invalid IP address or port [localhost:5000]```

No

### Expected behavior

To connect via hostname 

### Neovim version (nvim -v)

0.11 dev

Any LSP over tcp

### Operating system/version

Ubuntu 22.04

### Log file

_No response_

### PR DESCRIPTION
**Pull request by tris203**
Updated the `rpc.connect` function to support connecting to LSP servers using hostnames, not just IP addresses. This change includes updates to the documentation and additional test cases to verify the new functionality.

- Modified `connect` function to resolve hostnames.
- Updated documentation to reflect the change.
- Added test case for connecting using hostname.

closes #29513

### DIFF
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -33,6 +33,7 @@ local create_server_definition = t_lsp.create_server_definition
 local fake_lsp_code = t_lsp.fake_lsp_code
 local fake_lsp_logfile = t_lsp.fake_lsp_logfile
 local test_rpc_server = t_lsp.test_rpc_server
+local create_tcp_echo_server = t_lsp.create_tcp_echo_server
 
 local function get_buf_option(name, bufnr)
   bufnr = bufnr or 'BUFFER'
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by mfussenegger — APPROVED** [review]

File: test/functional/plugin/lsp_spec.lua [hunk 1]
```diff
@@ -5022,6 +5022,35 @@ describe('LSP', function()
       eq('initialize', result.method)
     end)
 
+    it('can connect to lsp server via rpc.connect using hostname', function()
```
justinmk, line 5025: almost identical to the test above, would you mind factoring out the code into a function?
```suggestion
    it('connects to lsp server via rpc.connect(hostname)', function()
```

justinmk (approved):
Nice, thanks for looking into this! LGTM after test deduplication.

tris203:
Will sort shortly @justinmk 

Should I be concerned about the freeBSD test failure? I suspect it's going to be something about resolution in whatever container that cirrus uses.

Shall I just put a skip() for cirrus in that test?

justinmk:
> Shall I just put a skip() for cirrus in that test?

for the "hostname" case only, sure we can skip it for now, like this: `skip(is_os('bsd')` https://github.com/neovim/neovim/blob/ef8067a19d981388a14407ea08245811cf5b3604/test/functional/lua/watch_spec.lua#L37

justinmk (approved):

PR #30243, opened 2024-09-02
fix(highlight): disable blending for TermCursor

### LINKED ISSUE #30241: Terminal-mode cursor not visible in floating terminals when winblend=100
### Problem

I want a cool transparent terminal in neovide, so I set winblend=100 in neovim. The cursor in the terminal is not longer visible.

~~I'm using floaterm as the floating terminal. Reporting here because I think it uses the same rendering code as neovim's built-in teminal (non-floating)~~
I have found a way to make a floating terminal in vanilla nvim, which is what I've written in "Steps to reproduce"

Here it is with winblend=100 (gnome-terminal, alacritty, kitty, neovide):
Screenshot from 2024-09-02 19-11-44

And with winblend=50:
Screenshot from 2024-09-02 19-12-17

(The screenshots are made with floaterm, but the same thing happens in vanilla nvim)

### Steps to reproduce

config.vim
```
function! Floatbuffer() abort
    " Define the size of the floating window
    let width = 50
    let height = 25

    set winblend=100

    " Create the scratch buffer displayed in the floating window
    let buf = nvim_create_buf(v:false, v:true)

    " Get the current UI
    let ui = nvim_list_uis()[0]

    " Create the floating window
    let opts = {'relative': 'editor',
                \ 'width': width,
                \ 'height': height,
                \ 'col': (ui.width/2) - (width/2),
                \ 'row': (ui.height/2) - (height/2),
                \ 'anchor': 'NW',
                \ 'style': 'minimal',
                \ }
    let win = nvim_open_win(buf, 1, opts)
endfunction
```

```
nvim -u config.vim
```

```
:call Floatbuffer()
:term
<i>
```

The cursor vanishes

### Expected behavior

The cursor should have the same opacity regardless of winblend. (Is is intentional?)

### Neovim version (nvim -v)

NVIM v0.10.1 Build type: Release LuaJIT 2.1.1713773202

no, vim doesn't have winblend

### Operating system/version

NixOS unstable

### Terminal name/version

Gnome Terminal, kitty, alacritty, neovide

xterm-256color

### Installation

nixpkgs (nixvim)

zeertzjq:
Fix #30241


zeertzjq:
I'm not entirely sure if this is the right solution.

PR #30252, opened 2024-09-03
fix(runtime): add remaining missing commentstrings

LosFarmosCTL:
See https://github.com/neovim/neovim/pull/23039#issuecomment-1505645253, adds all the remaining commentstrings that have been missing since removing the default C-style value. Found #29259 after running into this issue while writing swift, but since it seems to be dead I've decided to just pick it up myself.

**filetypes added in this PR:** `swift`, `dot`, `faust`, `stata`, `supercollider`

----

`autohotkey`, `cpp`, `dart`, `kotlin`, `nix`, `solidity`, `sql` and `terraform` are set in their respective vimscript ftplugins

`cuda` inherits from `cpp`, no need to update
`typescriptreact` and `javascriptreact` inherit from `typescript` and `javascript`, no need to update

`arduino`, `cs` and `glsl` already got a new lua ftplugin since that comment was written

### DIFF
--- a/runtime/ftplugin/dot.lua
+++ b/runtime/ftplugin/dot.lua
@@ -1,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '') .. '\n setl commentstring<'
--- a/runtime/ftplugin/faust.lua
+++ b/runtime/ftplugin/faust.lua
@@ -1,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '') .. '\n setl commentstring<'
--- a/runtime/ftplugin/stata.lua
+++ b/runtime/ftplugin/stata.lua
@@ -1,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '') .. '\n setl commentstring<'
--- a/runtime/ftplugin/supercollider.lua
+++ b/runtime/ftplugin/supercollider.lua
@@ -1,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '') .. '\n setl commentstring<'
--- a/runtime/ftplugin/swift.lua
+++ b/runtime/ftplugin/swift.lua
@@ -1,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'


clason:
Thank you! Just to make sure: You checked that for each of these languages, `// ` is the official comment marker?

LosFarmosCTL:
> Thank you! Just to make sure: You checked that for each of these languages, // is the official comment marker?

yep, it is for all of the added ones, some of the ones that are in vimscript have different ones, I guess that's why they were added upstream and the rest wasn't.

> Also, I see that some of the added ftplugins are missing an undo statement (or don't extend them properly) -- could you add these as well (separate commit) while you're at it?

Sure, I can add those.

clason:
> some of the ones that are in vimscript have different ones, I guess that's why they were added upstream and the rest wasn't.

I think as long as the commentstring is not _wrong_, we should follow upstream here (or contribute the fixes there first).

LosFarmosCTL:
> I think as long as the commentstring is not _wrong_, we should follow upstream here (or contribute the fixes there first).

yeah the vimscript ones are fine, no need to change anything there

runtime/ftplugin/d.lua
```diff
@@ -1,3 +1,3 @@
 vim.bo.commentstring = '// %s'
 
-vim.b.undo_ftplugin = 'setl commentstring<'
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: This leads to an error (concatenation with `nil`).

runtime/ftplugin/dot.lua
```diff
@@ -0,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: ditto

runtime/ftplugin/faust.lua
```diff
@@ -0,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: ditto

runtime/ftplugin/glsl.lua
```diff
@@ -1,3 +1,3 @@
 vim.bo.commentstring = '// %s'
 
-vim.b.undo_ftplugin = 'setl commentstring<'
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: ditto

runtime/ftplugin/stata.lua
```diff
@@ -0,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: ditto

runtime/ftplugin/supercollider.lua
```diff
@@ -0,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: ditto

zeertzjq (changes requested):

runtime/ftplugin/d.lua
LosFarmosCTL: > This leads to an error (concatenation with `nil`).

Oh yeah I see, those languages don't have vimscript ftplugins at all so undo_ftplugin was never set, makes sense.

Should that be a null check (something like vim.b.undo_ftplugin or '') or can we assume that those filetypes won't be added in upstream vim and remove the concatenation altogether?

zeertzjq: I have an idea: instead of ` |`, use `\n` to separate commands in a Lua ftplugin (preferably without a preceding space). That'll allow using `(vim.b.undo_ftplugin or '')`, and will not be treated as a part of an `:unmap` command like `|` does.

zeertzjq: Maybe it can be `(vim.b.undo_ftplugin or '') .. '\n setl commentstring<'` here instead. `\n` needs to be used instead of ` |` as using `|` without a command is not no-op.

LosFarmosCTL: I've updated the offending files, should the rest of the lua ftplugins also get that updated concatenation, or, since they do all have corresponding vimscript plugins that fill the undo_ftplugin value, be left as is?
Very unlikely, but if for some reason upstream vim removes an ftplugin or all undo statements in there, that could potentially create the same nil-concatenation error in the future.

zeertzjq: Let's leave that out of this PR?

runtime/ftplugin/dart.lua
```diff
@@ -0,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = vim.b.undo_ftplugin .. ' | setl commentstring<'
```
zeertzjq, line 3: `runtime/ftplugin/dart.vim` already sets this.

zeertzjq:
> some of the ones that are in vimscript have different ones, I guess that's why they were added upstream and the rest wasn't.

I have some trouble understanding this sentence. What does this mean?

runtime/ftplugin/dart.lua
LosFarmosCTL: True, must've missed this while reviewing the vimscript files, I've removed `dart.lua`

LosFarmosCTL:
> I have some trouble understanding this sentence. What does this mean?

Oh yeah, sorry, bad wording. I just meant that some of the languages listed in the original comment do actually use something other than `/* */` and `//` for comments, which is why they were likely added to upstream vim already, while the others were not, since they'd just inherit the C default there.

Also, your review comment made me take a closer look at the vimscript files again, just in case I missed something else and I've noticed that the second commit should be completely unnecessary. While the existing lua ftplugins largely did not set `undo_ftplugin`, the corresponding vimscript ones either already set it themselves or inherit from c.vim, which also does so. The only thing that was changed in the lua files was to update the commentstring from `/*%s*/` to `// %s`, but the original undo statements should still be fine.
I've gone ahead and dropped that commit, the only reason I could see for those undo statements to remain is if you're concerned about vim removing those commentstrings and their undos in the future, which seems very unlikely to me. If they should still be added, just let me know and I'll reflog it back on there.

-----

One remaining thing, I've added `stata`, `supercollider`, `faust` and `dot` as filetypes where there is no corresponding vimscript one. The undo statements right now are set like this:
```lua
vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '') .. '\nsetl commentstring<'
```

I've tested that and it works fine, but I did notice that other filetypes with only a lua ftplugin and no corresponding vimscript one, just set their `undo_ftplugin` directly, without any concatenation (e.g. [glsl](https://github.com/neovim/neovim/blob/master/runtime/ftplugin/glsl.lua)). Choosing a unified style seems like the reasonable thing to me, the question being whether it's likely that vim adds those filetypes in the future and setting the undo directly would potentially override anything that the vimscript ftplugins may set, or if concatenating is unnecessary for filetypes without corresponding vimscript ftplugins.

zeertzjq (approved):
The `vim.b.undo_ftplugin` lines should be fine now. Since there are already different ways `vim.b.undo_ftplugin` is set in existing Lua ftplugins, the consistency can be left out of this PR.

The remaining question is whether the 'commentstring' value for these filetypes should be `// %s` or `/* %s */`, as in Vim they are `/* %s */`, but Nvim already has different 'commentstring' values for filetypes like `c`, `cpp`, and `cs`. @clason What do you think?

clason:
I have not checked all the filetypes, but I believe they are correct: Vim just derived from the default C commentstring, which wasn't even correct in all cases (although nobody complained); other cases allow both C++ and C-style commentstrings, and I think it's reasonable to prefer the former (in any case more consistent).

So the change LGTM as well.

PR #30253, opened 2024-09-03
fix(lsp): handle locations exceeding line length

### LINKED ISSUE #28281: LSP: tolerate invalid paths in locations_to_items
### Problem

I sometimes get an error when trying to go to definition of find code references. When the error occurs I get no results (no quick fix list). However, when checking the LspLog I see that a list of results is returned.
The list of results from LSP contains some non existing paths. The source of the problem is probably somewhere between sourcekit-lsp itself and my configuration files for the iOS project.
However, I see no reason to give up on some of the correct results in the list just because of a few bad apples in it.

This is the error:
```
Error executing vim.schedule lua callback: ...lar/neovim/0.9.4/share/nvim/runtime/lua/vim/lsp/util.lua:1828: index out of range
stack traceback:
        [C]: in function '_str_byteindex_enc'
        ...lar/neovim/0.9.4/share/nvim/runtime/lua/vim/lsp/util.lua:1828: in function 'locations_to_items'
        ...neovim/0.9.4/share/nvim/runtime/lua/vim/lsp/handlers.lua:398: in function 'handler'
        ...w/Cellar/neovim/0.9.4/share/nvim/runtime/lua/vim/lsp.lua:1393: in function 'cb'
        vim/_editor.lua:263: in function <vim/_editor.lua:262>
```

I propose to tolerate errors and just report them to the user. And this can also be put behind some config flag if ignoring errors feels to outrageous.
In `runtime/lua/vim/lsp/util.lua`, in function `M.locations_to_items`:
```
      local callOk, col = pcall(M._str_byteindex_enc, line, pos.character, offset_encoding)
      if callOk then
        table.insert(items, {
          filename = filename,
          lnum = row + 1,
          col = col + 1,
          text = line,
          user_data = temp.location,
        })
      else
        -- errorCount will be defined above, outside of the for loop
        errorCount = errorCount + 1
      end
    end
    -- out of for loop. notify on found errors
    if errorCount > 0 then
      print("Error found when processing LSP RPC.......")
    end
``` 

In the unlikely scenario of working on an iOS xcworkspace with multiple xcodeproj projects and pod dependencies,
Run either `vim.lsp.buf.definition` or `vim.lsp.buf.references` lua functions.

### Expected behavior

_No response_

### Neovim version (nvim -v)

0.9.4

sourcekit

### Operating system/version

macOS 14.4.1

### Log file

_No response_

tris203:
Problem:

LSP spec [states](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position) that "if the character value is greater than the line length it defaults back to the line length", but `locations_to_items` gives the error in that case.

Solution:
Adjust locations_to_items to follow the spec.

closes #28281

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -1795,8 +1795,16 @@ function M.locations_to_items(locations, offset_encoding)
       local row = pos.line
       local end_row = end_pos.line
       local line = lines[row] or ''
-      local col = M._str_byteindex_enc(line, pos.character, offset_encoding)
-      local end_col = M._str_byteindex_enc(lines[end_row] or '', end_pos.character, offset_encoding)
+      local line_len = vim.fn.strcharlen(line)
+      local end_line = lines[end_row] or ''
+      local end_line_len = vim.fn.strcharlen(end_line)
+      local col = pos.character <= line_len
+          and M._str_byteindex_enc(line, pos.character, offset_encoding)
+        or line_len
+      local end_col = end_pos.character <= end_line_len
+          and M._str_byteindex_enc(end_line, end_pos.character, offset_encoding)
+        or end_line_len
+
       table.insert(items, {
         filename = filename,
         lnum = row + 1,
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -1798,6 +1798,8 @@ function M.locations_to_items(locations, offset_encoding)
       local line_len = vim.fn.strcharlen(line)
       local end_line = lines[end_row] or ''
       local end_line_len = vim.fn.strcharlen(end_line)
+      -- LSP spec: if character > line length, default to the line length.
+      -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
       local col = pos.character <= line_len
           and M._str_byteindex_enc(line, pos.character, offset_encoding)
         or line_len
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -2699,6 +2699,7 @@ describe('LSP', function()
             uri = 'file:///fake/uri',
             range = {
               start = { line = 0, character = 2 },
+              -- LSP spec: if character > line length, default to the line length.
               ['end'] = { line = 1, character = 10000 },
             },
           },
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -2673,7 +2673,7 @@ describe('LSP', function()
 
   describe('lsp.util.locations_to_items', function()
     it('Convert Location[] to items', function()
-      local expected = {
+      local expected_template = {
         {
           filename = '/fake/uri',
           lnum = 1,
@@ -2681,20 +2681,11 @@ describe('LSP', function()
           col = 3,
           end_col = 4,
           text = 'testing',
-          user_data = {
-            uri = 'file:///fake/uri',
-            range = {
-              start = { line = 0, character = 2 },
-              ['end'] = { line = 1, character = 3 },
-            },
-          },
+          user_data = {},
         },
       }
-      local actual = exec_lua(function()
-        local bufnr = vim.uri_to_bufnr('file:///fake/uri')
-        local lines = { 'testing', '123' }
-        vim.api.nvim_buf_set_lines(bufnr, 0, 1, false, lines)
-        local locations = {
+      local test_params = {
+        {
           {
             uri = 'file:///fake/uri',
             range = {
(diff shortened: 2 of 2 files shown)


test/functional/plugin/lsp_spec.lua
```diff
@@ -2708,6 +2708,42 @@ describe('LSP', function()
       eq(expected, actual)
     end)
 
+    it('location that exceeds the line length returns end_col of line length', function()
```
justinmk, line 2711: can we just modify the above test? this is a lot of (duplicate) code just to test an offset.

tris203: happy to do that if you prefer @justinmk 

I just thought it was better to leave the test for the "happy path" intact

runtime/lua/vim/lsp/util.lua
```diff
@@ -1795,8 +1795,18 @@ function M.locations_to_items(locations, offset_encoding)
...
+      ---@type boolean, integer, integer
+      local ok, col, end_col
+      ok, col = pcall(_str_byteindex_enc, line, pos.character, offset_encoding)
+      if not ok then
+        col = vim.fn.strcharlen(line) - 1
+      end
+      ok, end_col = pcall(_str_byteindex_enc, end_line, end_pos.character, offset_encoding)
+      if not ok then
+        end_col = vim.fn.strcharlen(end_line) - 1
+      end
```
justinmk, line 1808: The type annotation can be avoided by assigning the `local` declaration.

```suggestion
      local ok, col = pcall(_str_byteindex_enc, line, pos.character, offset_encoding)
      if not ok then
        col = vim.fn.strcharlen(line) - 1
      end
      local ok2, end_col = pcall(_str_byteindex_enc, end_line, end_pos.character, offset_encoding)
      if not ok2 then
        end_col = vim.fn.strcharlen(end_line) - 1
      end
```

runtime/lua/vim/lsp/util.lua
```diff
@@ -1795,8 +1795,18 @@ function M.locations_to_items(locations, offset_encoding)
       local row = pos.line
       local end_row = end_pos.line
       local line = lines[row] or ''
-      local col = M._str_byteindex_enc(line, pos.character, offset_encoding)
-      local end_col = M._str_byteindex_enc(lines[end_row] or '', end_pos.character, offset_encoding)
+      local end_line = lines[end_row] or ''
+      ---@type boolean, integer, integer
+      local ok, col, end_col
+      ok, col = pcall(_str_byteindex_enc, line, pos.character, offset_encoding)
```
justinmk, line 1801: instead of swallowing errors with pcall() is there an easy way to check that the length is valid? then the code would be clearer and more direct:

```lua
local col = pos.character <= ... and M._str_byteindex_enc(line, pos.character, offset_encoding) or (vim.fn.strcharlen(line) - 1)
```

tris203: I can check if pos.character is larger than the `vim.fn.strcharlen()` of the string but i was trying to avoid calling the len function until it was needed

test/functional/plugin/lsp_spec.lua
justinmk: that's a good practice for sure. but adding a `character` param to the `exec_lua` call in the above tests, would allow us to make one extra call with character=10000. 

so that allows testing the happy path as well as some variants. and it's clearer for others to follow, because it's only ~one call instead of 30+ lines of code :)

runtime/lua/vim/lsp/util.lua
justinmk: true. i think it's probably worth it though. pcall() can hide other kinds of errors, which would make other bugs hard to find.

test/functional/plugin/lsp_spec.lua
```diff
@@ -2673,39 +2673,48 @@ describe('LSP', function()
...
-        local lines = { 'testing', '123' }
-        vim.api.nvim_buf_set_lines(bufnr, 0, 1, false, lines)
-        local locations = {
+        {
           {
             uri = 'file:///fake/uri',
             range = {
               start = { line = 0, character = 2 },
-              ['end'] = { line = 1, character = 3 },
+              ['end'] = { line = 1, character = 10000 },
```
justinmk, line 2702:
```suggestion
              // LSP spec: if character > line length, default to the line length.
              ['end'] = { line = 1, character = 10000 },
```

runtime/lua/vim/lsp/util.lua
```diff
@@ -1795,8 +1795,16 @@ function M.locations_to_items(locations, offset_encoding)
       local row = pos.line
       local end_row = end_pos.line
       local line = lines[row] or ''
-      local col = M._str_byteindex_enc(line, pos.character, offset_encoding)
-      local end_col = M._str_byteindex_enc(lines[end_row] or '', end_pos.character, offset_encoding)
+      local line_len = vim.fn.strcharlen(line)
+      local end_line = lines[end_row] or ''
+      local end_line_len = vim.fn.strcharlen(end_line)
+      local col = pos.character <= line_len
```
justinmk, line 1801:
```suggestion
      -- LSP spec: if character > line length, default to the line length.
      -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
      local col = pos.character <= line_len
```

PR #30257, opened 2024-09-04
feat(treesitter): support modelines in `query.set()`

### LINKED ISSUE #26965: Enable extending queries via `vim.treesitter.query.set`

MariaSolOs:
Closes #26965.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -358,6 +358,8 @@ TREESITTER
   descendant itself.
 • |LanguageTree:parse()| optionally supports asynchronous invocation, which is
   activated by passing the `on_parse` callback parameter.
+• |vim.treesitter.query.set()| can now inherit and/or extend runtime file
+  queries in addition to overriding.
 
 TUI
 
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -1339,7 +1339,7 @@ parse({lang}, {query})                          *vim.treesitter.query.parse()*
       `info.captures`).
     • `info.patterns`: information about predicates.
 
-    Example (to try it, use `g==` or select the code then run `:'<,'>lua`): >lua
+    Example: >lua
         local query = vim.treesitter.query.parse('vimdoc', [[
           ; query
           ((h1) @str
@@ -1456,8 +1456,18 @@ Query:iter_matches({node}, {source}, {start}, {stop}, {opts})
 set({lang}, {query_name}, {text})                 *vim.treesitter.query.set()*
     Sets the runtime query named {query_name} for {lang}
 
-    This allows users to override any runtime files and/or configuration set
-    by plugins.
+    This allows users to override or extend any runtime files and/or
+    configuration set by plugins.
+
+    For example, you could enable spellchecking of `C` identifiers with the
+    following code: >lua
+        vim.treesitter.query.set(
+          'c',
+          'highlights',
+          [[;inherits c
+          (identifier) @spell]])
+        ]])
+<
 
     Parameters: ~
       • {lang}        (`string`) Language to use for the query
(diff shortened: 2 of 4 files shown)


MariaSolOs:
I'll wait for https://github.com/neovim/neovim/pull/30227 to be merged first, since the changes in `_memoize` will be relevant in this PR.

lucario387:
Question: Even with `_memoize._invalidate`, how do we determine the query order after deleting some extension? Or is it going to be kept with the order of when it's added in?

clason:
I think any manual `set` should (and probably automatically will) come after queries on `rtp`. Note that _deleting_ queries is not part of this proposal.

runtime/doc/news.txt
```diff
@@ -358,6 +358,8 @@ TREESITTER
   descendant itself.
 • |LanguageTree:parse()| optionally supports asynchronous invocation, which is
   activated by passing the `on_parse` callback parameter.
+• |vim.treesitter.query.set()| can now inherit and/or extend runtime file
+  queries in addition to overriding.
```
justinmk, line 362: Is there a small practical example we can show here? E.g. mentioning 'spell' explicitly would be very helpful.

MariaSolOs: Good idea. I'll add the example to `treesitter.txt` though.

lewis6991 (approved):

MariaSolOs:
@clason Let me know if this implementation matches what you had in mind for the feature :)

runtime/doc/treesitter.txt
```diff
@@ -1456,8 +1456,19 @@ Query:iter_matches({node}, {source}, {start}, {stop}, {opts})
 set({lang}, {query_name}, {text})                 *vim.treesitter.query.set()*
     Sets the runtime query named {query_name} for {lang}
 
-    This allows users to override any runtime files and/or configuration set
-    by plugins.
+    This allows users to override or extend any runtime files and/or
+    configuration set by plugins.
+
+    For example, you could enable spellchecking of `C` identifiers with the
+    following code (to try it, use `g==` or select the code then run
```
clason, line 1463:
```suggestion
    following code
```

runtime/doc/treesitter.txt
```diff
@@ -1456,8 +1456,19 @@ Query:iter_matches({node}, {source}, {start}, {stop}, {opts})
...
+    `:'<,'>lua`): >lua
```
clason, line 1464:
```suggestion
    following code: >lua
```
(no need to call this out in this specific spot)

runtime/lua/vim/treesitter/query.lua
```diff
@@ -255,15 +255,26 @@ local explicit_queries = setmetatable({}, {
 
 --- Sets the runtime query named {query_name} for {lang}
 ---
---- This allows users to override any runtime files and/or configuration
+--- This allows users to override or extend any runtime files and/or configuration
 --- set by plugins.
 ---
+--- For example, you could enable spellchecking of `C` identifiers with the
+--- following code (to try it, use `g==` or select the code then run `:'<,'>lua`):
```
clason, line 262:
```suggestion
--- following code:
```

MariaSolOs: Fair enough. I just added this as this is what `M.parse` does:
https://github.com/neovim/neovim/blob/bad3c9e5eccaf8d273bf52e68c82ae1c7dabf6a8/runtime/lua/vim/treesitter/query.lua#L346

PR #30259, opened 2024-09-04
fix(float): re-sort layers when grid zindex changed

### LINKED ISSUE #18142: Updating zindex of a floating window does not work
### Neovim version (nvim -v)

NVIM v0.7.0-dev+1280-gf2e5f509d Build type: RelWithDebInfo LuaJIT 2.1.0-beta3 Compilation: /bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions     -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security     -fstack-clash-protection -fcf-protection -Wp,-U_FORTIFY_SOURCE -Wp,-D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/var/cache/aura/vcs/neovim-git/neovim-git/src/build/config -I/var/cache/aura/vcs/neovim-git/neovim-git/src/neovim-git/src -I/usr/include -I/var/cache/aura/vcs/neovim-git/neovim-git/src/build/src/nvim/auto -I/var/cache/aura/vcs/neovim-git/neovim-git/src/build/include Compiled by ronan@XPS15  Features: +acl +iconv +tui See ":help feature-compile"     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/usr/share/nvim"

N/A

### Operating system/version

Arch Linux

### Terminal name/version

Kitty 0.23.1

xterm-kitty

### Installation

AUR

```lua
local win = vim.api.nvim_open_win(0, false, {
  relative = "editor",
  anchor = "NW",
  width = 50,
  height = 4,
  col = 1,
  row = 1,
  border = "double",
  style = "minimal",
})

vim.schedule(function()
  vim.api.nvim_win_set_config(win, { zindex = 400 })

  vim.api.nvim_open_win(0, false, {
    zindex = 300,
    relative = "editor",
    anchor = "NW",
    style = "minimal",
    width = 40,
    height = 5,
    row = 1,
    col = 1,
    border = "single",
  })
end)
```

### Expected behavior

The first window (double border) should appear above the second.

Remove the schedule to see expected behaviour
```lua
local win = vim.api.nvim_open_win(0, false, {
  relative = "editor",
  anchor = "NW",
  width = 50,
  height = 4,
  col = 1,
  row = 1,
  border = "double",
  style = "minimal",
})

vim.api.nvim_win_set_config(win, { zindex = 400 })

vim.api.nvim_open_win(0, false, {
zindex = 300,
relative = "editor",
anchor = "NW",
style = "minimal",
width = 40,
height = 5,
row = 1,
col = 1,
border = "single",
})
```

### Actual behavior

The second window appears on top despite having a lower zindex.

glepnir:
Problem: insert_at value is not correct when using vim.schedule.

Solution: range insert_at from head of layers to find the place to insert grid.

Fix #18142

### DIFF
--- a/src/nvim/ui_compositor.c
+++ b/src/nvim/ui_compositor.c
@@ -101,6 +101,35 @@ bool ui_comp_should_draw(void)
   return composed_uis != 0 && valid_screen;
 }
 
+/// Raises or lowers the layer, syncing comp_index with zindex.
+///
+/// This function adjusts the position of a layer in the layers array
+/// based on its zindex, either raising or lowering it.
+///
+/// @param[in]  layer_idx  Index of the layer to be raised or lowered.
+/// @param[in]  raise      Raise the layer if true, else lower it.
+void ui_comp_layers_adjust(size_t layer_idx, bool raise)
+{
+  size_t size = layers.size;
+  ScreenGrid *layer = layers.items[layer_idx];
+
+  if (raise) {
+    while (layer_idx < size - 1 && layer->zindex > layers.items[layer_idx + 1]->zindex) {
+      layers.items[layer_idx] = layers.items[layer_idx + 1];
+      layers.items[layer_idx]->comp_index = layer_idx;
+      layer_idx++;
+    }
+  } else {
+    while (layer_idx > 0 && layer->zindex < layers.items[layer_idx - 1]->zindex) {
+      layers.items[layer_idx] = layers.items[layer_idx - 1];
+      layers.items[layer_idx]->comp_index = layer_idx;
+      layer_idx--;
+    }
+  }
+  layers.items[layer_idx] = layer;
+  layer->comp_index = layer_idx;
+}
+
 /// Places `grid` at (col,row) position with (width * height) size.
 /// Adds `grid` as the top layer if it is a new layer.
 ///
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -852,7 +852,13 @@ void ui_ext_win_position(win_T *wp, bool validate)
       }
     }
 
+    bool resort = wp->w_grid_alloc.comp_index != 0
+                  && wp->w_grid_alloc.zindex != wp->w_config.zindex;
+    bool raise = resort && wp->w_grid_alloc.zindex < wp->w_config.zindex;
     wp->w_grid_alloc.zindex = wp->w_config.zindex;
+    if (resort) {
+      ui_comp_layers_adjust(wp->w_grid_alloc.comp_index, raise);
+    }
     if (ui_has(kUIMultigrid)) {
       String anchor = cstr_as_string(float_anchor_str[c.anchor]);
       if (!c.hide) {
--- a/test/functional/ui/float_spec.lua
+++ b/test/functional/ui/float_spec.lua
@@ -8542,6 +8542,111 @@ describe('float window', function()
                                                   |
         ]]}
       end
+
+      --
+      -- Check that floats are positioned correctly after changing the zindexes.
+      --
+      command('fclose')
+      exec_lua([[
+        local win1, win3 = ...
+        vim.api.nvim_win_set_config(win1, { zindex = 400, title = 'win_400', title_pos = 'center', border = 'double' })
+        vim.api.nvim_win_set_config(win3, { zindex = 300, title = 'win_300', title_pos = 'center', border = 'single' })
+      ]], win1, win3)
+      if multigrid then
+        screen:expect({
+          grid = [[
+          ## grid 1
+            [2:----------------------------------------]|*6
+            [3:----------------------------------------]|
+          ## grid 2
+            ^                                        |
+            {0:~                                       }|*5
+          ## grid 3
+                                                    |
+          ## grid 4
+            {5:╔══════}{11:win_400}{5:═══════╗}|
+            {5:║}{7:                    }{5:║}|
+            {5:║}{7:~                   }{5:║}|*2
+            {5:╚════════════════════╝}|
+          ## grid 6
+            {5:┌──────}{11:win_300}{5:───────┐}|
+            {5:│}{8:                    }{5:│}|
+            {5:│}{8:~                   }{5:│}|*2
+            {5:└────────────────────┘}|
+          ]], float_pos={
+          [4] = {1001, "NW", 1, 1, 5, true, 400};
+          [6] = {1003, "NW", 1, 3, 7, true, 300};
+        }, win_viewport={
+          [2] = {win = 1000, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+          [4] = {win = 1001, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+          [6] = {win = 1003, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+        }, win_viewport_margins={
+          [2] = { bottom = 0, left = 0, right = 0, top = 0, win = 1000 },
+          [4] = { bottom = 1, left = 1, right = 1, top = 1, win = 1001 },
+          [6] = { bottom = 1, left = 1, right = 1, top = 1, win = 1003 }
+        }})
+      else
+        screen:expect({
+          grid = [[
+            ^                                        |
+            {0:~    }{5:╔══════}{11:win_400}{5:═══════╗}{0:             }|
+            {0:~    }{5:║}{7:                    }{5:║─┐}{0:           }|
+            {0:~    }{5:║}{7:~                   }{5:║}{8: }{5:│}{0:           }|*2
+            {0:~    }{5:╚════════════════════╝}{8: }{5:│}{0:           }|
+                   {5:└────────────────────┘}           |
+          ]]
+        })
+      end
+      exec_lua([[
+        local win1, win3 = ...
+        vim.api.nvim_win_set_config(win1, { zindex = 100, title='win_100' })
+        vim.api.nvim_win_set_config(win3, { zindex = 150, title='win_150' })
+      ]], win1, win3)
+      if multigrid then
+        screen:expect({
+          grid = [[
+          ## grid 1
+            [2:----------------------------------------]|*6
+            [3:----------------------------------------]|
+          ## grid 2
+            ^                                        |
+            {0:~                                       }|*5
+          ## grid 3
+                                                    |
+          ## grid 4
+            {5:╔}{11:win_100}{5:═════════════╗}|
+            {5:║}{7:                    }{5:║}|
+            {5:║}{7:~                   }{5:║}|*2
+            {5:╚════════════════════╝}|
+          ## grid 6
+            {5:┌}{11:win_150}{5:─────────────┐}|
+            {5:│}{8:                    }{5:│}|
+            {5:│}{8:~                   }{5:│}|*2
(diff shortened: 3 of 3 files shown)


zeertzjq (dismissed):
This does not fix the problem. It just makes it happen in the opposite direction:
```lua
local win = vim.api.nvim_open_win(0, false, {
  zindex = 300,
  relative = "editor",
  anchor = "NW",
  width = 40,
  height = 5,
  col = 1,
  row = 1,
  border = "single",
  style = "minimal",
})

vim.schedule(function()
  vim.api.nvim_win_set_config(win, { zindex = 100 })

  vim.api.nvim_open_win(0, false, {
    zindex = 150,
    relative = "editor",
    anchor = "NW",
    style = "minimal",
    width = 50,
    height = 4,
    row = 1,
    col = 1,
    border = "double",
  })
end)
```

glepnir:
because when run open_win in vim.schedule layers is `[zindex 0, zindex 400, zindex200]` then insert_at is 3 so problem maybe is where there has a zindex 200 ..

glepnir:
now should work.

src/nvim/window.c
```diff
@@ -839,7 +839,11 @@ void ui_ext_win_position(win_T *wp, bool validate)
       }
     }
 
+    bool resort = wp->w_grid_alloc.zindex != wp->w_config.zindex;
```
zeertzjq, line 842:
```suggestion
    bool resort = wp->w_grid_alloc.comp_index != 0
                  && wp->w_grid_alloc.zindex != wp->w_config.zindex;
```

glepnir: oh thanks

test/functional/ui/float_spec.lua
```diff
@@ -9389,6 +9389,179 @@ describe('float window', function()
         })
       end
     end)
+
+    it('ensures correct zindex order when using vim.schedule', function()
```
justinmk, line 9393: can the existing test be modified? why do we need a separate test?  https://github.com/neovim/neovim/blob/e36e68d35cfa285b28f4c87782311e913a306c92/test/functional/ui/float_spec.lua#L8388

glepnir: screen height is needed to make the position of the two windows clear in the test. Special trigger condition vim.schedule

justinmk: why can't we just modify the existing test with those changes?

justinmk: by "modify the existing test", I don't mean to paste the entire new test in the existing test scope. I mean just change the existing test. Why do we need 200 lines of test code? Just change the _existing_ test for z-index behavior so that it uses `vim.schedule` and whatever else.

glepnir: because of some line wrap. existing tests are the same and each window uses a new config. prediction code for the two-window screen is  long.

glepnir: maybe need reuse and clean . trying..

test/functional/ui/float_spec.lua
```diff
@@ -8434,6 +8434,111 @@ describe('float window', function()
...
         ]]}
       end
+      command('fclose')
+      exec_lua([[
+        local win1, win3 = ...
+        vim.schedule(function()
+          vim.api.nvim_win_set_config(win1, { zindex = 400, title = 'win_400', title_pos = 'center', border = 'single' })
+          vim.api.nvim_win_set_config(win3, { zindex = 300, title = 'win_300', title_pos = 'center', border = 'single' })
+        end)
+      ]], win1, win3)
```
zeertzjq, line 8444: I don't think `vim.schedule()` is needed, as you are already using `exec_lua()`.

test/functional/ui/float_spec.lua
```diff
@@ -8434,6 +8434,107 @@ describe('float window', function()
                                                   |
         ]]}
       end
+      command('fclose')
+      exec_lua([[
+        local win1, win3 = ...
+        vim.api.nvim_win_set_config(win1, { zindex = 400, title = 'win_400', title_pos = 'center', border = 'single' })
+        vim.api.nvim_win_set_config(win3, { zindex = 300, title = 'win_300', title_pos = 'center', border = 'single' })
```
zeertzjq, line 8441: Using different border styles may make the order clearer.

glepnir: okay done.

test/functional/ui/float_spec.lua
```diff
@@ -8385,7 +8385,7 @@ describe('float window', function()
       end
     end)
 
-    it('can use z-index', function()
+    it('#can use z-index', function()
```
zeertzjq, line 8388:
```suggestion
    it('can use z-index', function()
```

zeertzjq:
BTW, I think it should be "re-sort" in the commit message, not "resort".

src/nvim/ui_compositor.c
```diff
@@ -101,6 +101,21 @@ bool ui_comp_should_draw(void)
   return composed_uis != 0 && valid_screen;
 }
 
+static int compare_zindex(const void *a, const void *b)
+{
+  int za = (*(ScreenGrid **)a)->zindex;
+  int zb = (*(ScreenGrid **)b)->zindex;
+  return za == zb ? 0 : (za < zb ? -1 : 1);
```
zeertzjq, line 108: The sorting is not stable. If two layers have the same zindex, their order should be kept, so compare `comp_index` in that case:
```suggestion
  size_t ia = (*(ScreenGrid **)a)->comp_index;
  size_t ib = (*(ScreenGrid **)b)->comp_index;
  return za == zb ? (ia == ib ? 0 : (ia < ib ? -1 : 1)) : (za < zb ? -1 : 1);
```

glepnir: sound reasonable..

src/nvim/ui_compositor.c
```diff
@@ -101,6 +101,34 @@ bool ui_comp_should_draw(void)
   return composed_uis != 0 && valid_screen;
 }
 
+// Adjusts the position of a layer in the layers array based on its zindex.
+// Moves the layer either forward (if zindex increases) or backward (if zindex decreases).
+//
+// @param[in]  idx       The index of the current layer in the items array.
+// @param[in]  forward   True move forward otherwise move backward.
+void ui_comp_layers_adjust(size_t idx, bool forward)
```
justinmk, line 109: should we use the same language as the existing `ui_comp_raise_grid` function? Instead of referring to forward/backward, refer to higher/lower (or "raise").

Also why does the docstring say `if zindex increases`? The caller makes that decision, not this function.

```suggestion
// Raises or lowers the layer.
//
// @param[in]  layer_idx Index of the layer to be raised/lowered.
// @param[in]  raise     Raise the layer if true, else lower it.
void ui_comp_raise_layer(size_t layer_idx, bool raise)
```

src/nvim/ui_compositor.c
```diff
@@ -101,6 +101,34 @@ bool ui_comp_should_draw(void)
...
+{
+  ScreenGrid **items = layers.items;
+  size_t size = layers.size;
+  ScreenGrid *layer = items[idx];
+
+  if (forward) {
+    while (idx < size - 1 && layer->zindex > items[idx + 1]->zindex) {
```
justinmk, line 116: nit: would it be clearer if we did not alias `layers.items`?
```suggestion
  ScreenGrid *layer = layers.items[idx];

  if (forward) {
    while (idx < size - 1 && layer->zindex > layers.items[idx + 1]->zindex) {
```

src/nvim/window.c
```diff
@@ -839,7 +839,13 @@ void ui_ext_win_position(win_T *wp, bool validate)
       }
     }
 
+    bool resort = wp->w_grid_alloc.comp_index != 0
+                  && wp->w_grid_alloc.zindex != wp->w_config.zindex;
+    bool move_forward = resort && wp->w_grid_alloc.zindex < wp->w_config.zindex;
```
justinmk, line 844:
```suggestion
    bool resort = wp->w_grid_alloc.comp_index != 0
                  && wp->w_grid_alloc.zindex != wp->w_config.zindex;
    bool raise = resort && wp->w_grid_alloc.zindex < wp->w_config.zindex;
```

test/functional/ui/float_spec.lua
```diff
@@ -8434,6 +8434,107 @@ describe('float window', function()
                                                   |
         ]]}
       end
+      command('fclose')
```
justinmk, line 8437:
```suggestion

      --
      -- Check that floats are positioned correctly after changing the zindexes.
      --
      command('fclose')
```

src/nvim/ui_compositor.c
```diff
@@ -101,6 +101,32 @@ bool ui_comp_should_draw(void)
   return composed_uis != 0 && valid_screen;
 }
 
+// Raises or lowers the layer.
+//
+// @param[in]  layer_idx Index of the layer to be raised/lowered.
+// @param[in]  raise     Raise the layer if true, else lower it.
+void ui_comp_layers_adjust(size_t layer_idx, bool raise)
```
justinmk, line 108: what is the primary purpose of this function? is its main purpose to "sync" zindex with comp_index? or is that just a detail, and the main purpose is to raise/lower a layer?

how does it different from `ui_comp_raise_grid` ? (why aren't we using `ui_comp_raise_grid`?)

the answers to the above will help lead to a better name for this function. "adjust" is just meaningless.

glepnir: new implementation mainly avoids the use of full qsort, updates the order of the window grid in the layers according to the change of zindex, and ensures the correct order when the windows are drawn.

justinmk: > updates the order of the window grid in the layers according to the change of zindex, and ensures the correct order when the windows are drawn.

Ok that sounds like "syncs comp_index with zindex". Right? 

Is the *primary* purpose of the function to raise/lower, or to "sync" comp_index? Please update the docstring and function name, based on that.

glepnir: yes.

src/nvim/ui_compositor.c
```diff
@@ -101,6 +101,35 @@ bool ui_comp_should_draw(void)
   return composed_uis != 0 && valid_screen;
 }
 
+// Raises or lowers the layer, syncing comp_index with zindex.
+//
+// This function adjusts the position of a layer in the layers array
+// based on its zindex, either raising or lowering it.
+//
+// @param[in]  layer_idx  Index of the layer to be raised or lowered.
+// @param[in]  raise      Raise the layer if true, else lower it.
```
zeertzjq, line 110:
```suggestion
/// Raises or lowers the layer, syncing comp_index with zindex.
///
/// This function adjusts the position of a layer in the layers array
/// based on its zindex, either raising or lowering it.
///
/// @param[in]  layer_idx  Index of the layer to be raised or lowered.
/// @param[in]  raise      Raise the layer if true, else lower it.
```

PR #30272, opened 2024-09-05
fix(multibyte): handle backspace of wide clusters in replace mode

bfredl:
Make utf_head_off more robust against invalid sequences and embedded NUL chars


bfredl:
news.txt already mentions the feature (extended grapheme clusters). This is a bugfix of a feature from the same cycle, not news.

PR #30288, opened 2024-09-06
fix(lsp): handle out-of-bounds character positions

tris203:
Problem:
str_byteindex_enc could return an error if the index was longer than the line length. This was handled in each of the calls to it individually

Solution:
Fix the call at the source level so that if the index is higher than the line length, line length is returned as per LSP specification

may help with #19237

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -147,6 +147,12 @@ end
 ---@param encoding string utf-8|utf-16|utf-32| defaults to utf-16
 ---@return integer byte (utf-8) index of `encoding` index `index` in `line`
 function M._str_byteindex_enc(line, index, encoding)
+  local len = vim.fn.strlen(line)
+  if index > len then
+    -- LSP spec: if character > line length, default to the line length.
+    -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
+    return len
+  end
   if not encoding then
     encoding = 'utf-16'
   end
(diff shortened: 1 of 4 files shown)


test/functional/plugin/lsp_spec.lua
```diff
@@ -2108,7 +2108,7 @@ describe('LSP', function()
     end)
   end)
 
-  describe('apply_text_edits regression tests for #20116', function()
+  describe('apply_text_edits from lsp regression tests for #20116', function()
```
justinmk, line 2111: this is in a `describe('lsp')` block so we generally don't repeat that info in nested describe() blocks
```suggestion
  describe('apply_text_edits regression tests for #20116', function()
```

runtime/lua/vim/lsp/util.lua
```diff
@@ -436,14 +442,14 @@ function M.apply_text_edits(text_edits, bufnr, offset_encoding)
...
-        -- If the replacement is over the end of a line (i.e. e.end_col is out of bounds and the
+        -- If the replacement is over the end of a line (i.e. e.end_col is equal to the line length and the
         -- replacement text ends with a newline We can likely assume that the replacement is assumed
         -- to be meant to replace the newline with another newline and we need to make sure this
         -- doesn't add an extra empty line. E.g. when the last line to be replaced contains a '\r'
         -- in the file some servers (clangd on windows) will include that character in the line
         -- while nvim_buf_set_text doesn't count it as part of the line.
         if
-          e.end_col > last_line_len
+          e.end_col == last_line_len
```
justinmk, line 452: should we check `>=` to be defensive? or should we throw an error if `e.end_col > last_line_len` ?

justinmk (approved):
Looks like a good change. I assume you checked that the test changes make sense, and thus if all other tests are passing then this seems like the right direction.

tris203:
> Looks like a good change. I assume you checked that the test changes make sense, and thus if all other tests are passing then this seems like the right direction.

Yes, the test changes are because, the previous version of the test wasn't backed by a real buffer, so the string passed was always ```''```
I found the PR where the change was introduced and recreated the buffer from screenshots in that PR. The line numbers changed as I shorted the test file to keep it concise.

mfussenegger (approved):

clason:
Backport?

rustyper:
I have encountered a problem in conjunction with the typst plugin, that caused an index out of range exception:

image

I'm wondering whether this has something to do with the changes made by this PR in semantic_tokens.lua?

I'm using the following software versions:

1. neovim: 0.10.2
2. typst.vim: head

P.S. Sorry for commenting on a already merged PR.

PR #30292, opened 2024-09-07
fix(fs): improve environment variable expansion in `vim.fs.normalize`

jamestrew:
Environment variables in Windows (eg. `%PATH%`, `%USERPROFILE%`, `%TEMP%`) are currently unsupported in `vim.fs.normalize`.

This PR aims to address this.


zeertzjq:
This should be `feat:`

zeertzjq:
I don't think this is good idea. Why can't one use `$` on Windows?

jamestrew:
> I don't think this is good idea. Why can't one just also use `$` on Windows?

Yeah I should at least add back `$` support. Didn't realize that was possible - did this PR on a linux machine.

I still think `%ENV_VAR%` support is worth supporting and valid character set for Windows env vars are different to posix ones.

I'll need to touch up this PR.

PR #30295, opened 2024-09-07
fix(highlight): floating windows inherit NormalFloat from global-ns

### LINKED ISSUE #30262: nvim_set_hl_ns makes style='minimal' floats' background transparent
### Problem

Invoking `nvim_set_hl_ns` with anything but 0 makes `style=minimal` floating windows ignore the NormalFloat background color and instead become transparent:

https://github.com/user-attachments/assets/c5c7949a-5971-4287-9dc1-6badef498215

Is this what it's meant to do? Seems not.

### Steps to reproduce

1. `nvim --clean`
2. `:hi NormalFloat guibg=#ff0000` (make it easier to see, NormalFloat is only a slightly different shade than Normal in the default colorscheme)
3. `:lua vim.api.nvim_open_win(0, false, {relative='win', row=3, col=3, width=12, height=3, style='minimal'})`
4. `:lua vim.api.nvim_set_hl_ns(vim.api.nvim_create_namespace("test1"))`

### Expected behavior

I'd expect the float to stay red, as the new namespace doesn't specify any highlights—I'd expect no changes whatsoever anywhere.

### Neovim version (nvim -v)

v0.11.0-dev-711+g7b7c95dac (reproducible with v0.10.1 and v0.9.5 as well)

N/A (nvim api specific issue)

### Operating system/version

Debian testing

### Terminal name/version

alacritty 0.13.2 (reproducible with rxvt-unicode 9.31 as well)

alacritty (or rxvt-unicode-256color)

### Installation

nvim-linux64.tar.gz from latest GitHub nightly

glepnir:
Problem:
floating windows did not correctly inherit the NormalFloat highlight group from the global namespace when it was not defined in the window-specific namespace. This led to floating windows losing their background highlight when switching between namespaces.

Solution:
Updated the window highlight logic in update_window_hl() to handle the fallback.

Fix #30262

### DIFF
--- a/src/nvim/highlight.c
+++ b/src/nvim/highlight.c
@@ -370,12 +370,15 @@ void update_window_hl(win_T *wp, bool invalid)
 
   // determine window specific background set in 'winhighlight'
   bool float_win = wp->w_floating && !wp->w_config.external;
-  if (float_win && hl_def[HLF_NFLOAT] != 0) {
+  if (float_win && hl_def[HLF_NFLOAT] != 0 && ns_id > 0) {
     wp->w_hl_attr_normal = hl_def[HLF_NFLOAT];
   } else if (hl_def[HLF_COUNT] > 0) {
     wp->w_hl_attr_normal = hl_def[HLF_COUNT];
+  } else if (float_win) {
+    wp->w_hl_attr_normal = HL_ATTR(HLF_NFLOAT) > 0
+                           ? HL_ATTR(HLF_NFLOAT) : highlight_attr[HLF_NFLOAT];
   } else {
-    wp->w_hl_attr_normal = float_win ? HL_ATTR(HLF_NFLOAT) : 0;
+    wp->w_hl_attr_normal = 0;
   }
 
   if (wp->w_floating) {
--- a/test/functional/ui/float_spec.lua
+++ b/test/functional/ui/float_spec.lua
@@ -1344,6 +1344,53 @@ describe('float window', function()
                                                   |
         ]])
       end
+
+      --
+      -- floating windows inherit NormalFloat from global-ns.
+      --
+      command('fclose')
+      command('hi NormalFloat guibg=LightRed')
+      api.nvim_open_win(0, false, { relative = 'win', row = 3, col = 3, width = 12, height = 3, style = 'minimal' })
+      api.nvim_set_hl_ns(api.nvim_create_namespace('test1'))
+      if multigrid then
+        screen:expect({
+          grid = [[
+          ## grid 1
+            [2:----------------------------------------]|*6
+            [3:----------------------------------------]|
+          ## grid 2
+            {14:  1 }^x                                   |
+            {14:  2 }y                                   |
+            {14:  3 }                                    |
+            {0:~                                       }|*3
+          ## grid 3
+                                                    |
+          ## grid 5
+            {22:x           }|
+            {22:y           }|
+            {22:            }|
+          ]], float_pos={
+          [5] = {1002, "NW", 2, 3, 3, true, 50};
+        }, win_viewport={
+          [2] = {win = 1000, topline = 0, botline = 4, curline = 0, curcol = 0, linecount = 3, sum_scroll_delta = 0};
+          [5] = {win = 1002, topline = 0, botline = 3, curline = 0, curcol = 0, linecount = 3, sum_scroll_delta = 0};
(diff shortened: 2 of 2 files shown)


src/nvim/highlight.c
```diff
@@ -347,6 +347,7 @@ void update_window_hl(win_T *wp, bool invalid)
 {
   int ns_id = wp->w_ns_hl;
 
+  int old_nfloat = wp->w_floating && wp->w_hl_attr_normal > 0 ? wp->w_hl_attr_normal : -1;
```
bfredl, line 350: unfortunately this doesn't work as this might be the first time update_window_hl() is called for the window, so there is no cached value to use.

I poked around a little and think this condition is the right one:
```diff
diff --git a/src/nvim/highlight.c b/src/nvim/highlight.c
index 8729c74ce8..8436cd04de 100644
--- a/src/nvim/highlight.c
+++ b/src/nvim/highlight.c
@@ -370,12 +370,14 @@ void update_window_hl(win_T *wp, bool invalid)
 
   // determine window specific background set in 'winhighlight'
   bool float_win = wp->w_floating && !wp->w_config.external;
-  if (float_win && hl_def[HLF_NFLOAT] != 0) {
+  if (float_win && hl_def[HLF_NFLOAT] != 0 && ns_id > 0) {
     wp->w_hl_attr_normal = hl_def[HLF_NFLOAT];
   } else if (hl_def[HLF_COUNT] > 0) {
     wp->w_hl_attr_normal = hl_def[HLF_COUNT];
+  } else if (float_win) {
+    wp->w_hl_attr_normal = HL_ATTR(HLF_NFLOAT) >= 0 ? HL_ATTR(HLF_NFLOAT) : highlight_attr[HLF_NFLOAT];
   } else {
-    wp->w_hl_attr_normal = float_win ? HL_ATTR(HLF_NFLOAT) : 0;
+    wp->w_hl_attr_normal = 0;
   }
 
   if (wp->w_floating) {
```

glepnir: oh right i ignored the `highlight_attr`  thanks

glepnir: emm i give it a try the red background still missing

bfredl: ah, sorry, it should be `HL_ATTR(HLF_NFLOAT) > 0` and not `HL_ATTR(HLF_NFLOAT) >= 0`, updated above

test/functional/ui/float_spec.lua
```diff
@@ -1344,6 +1344,49 @@ describe('float window', function()
                                                   |
         ]])
       end
+      command('fclose')
+      command('hi NormalFloat guibg=LightRed')
+      api.nvim_open_win(0, false, { relative = 'win', row = 3, col = 3, width = 12, height = 3, style = 'minimal' })
+      command("lua vim.api.nvim_set_hl_ns(vim.api.nvim_create_namespace('test1'))")
```
bfredl, line 1350: no need for `command("lua")`, just use `api.` functions directly

test/functional/ui/float_spec.lua
```diff
@@ -1344,6 +1344,49 @@ describe('float window', function()
                                                   |
         ]])
       end
+      command('fclose')
```
justinmk, line 1347:
```suggestion

      --
      -- floating windows inherit NormalFloat from global-ns.
      --
      command('fclose')
```

src/nvim/highlight.c
```diff
@@ -370,12 +370,15 @@ void update_window_hl(win_T *wp, bool invalid)
 
   // determine window specific background set in 'winhighlight'
   bool float_win = wp->w_floating && !wp->w_config.external;
-  if (float_win && hl_def[HLF_NFLOAT] != 0) {
+  if (float_win && hl_def[HLF_NFLOAT] != 0 && ns_id > 0) {
     wp->w_hl_attr_normal = hl_def[HLF_NFLOAT];
   } else if (hl_def[HLF_COUNT] > 0) {
     wp->w_hl_attr_normal = hl_def[HLF_COUNT];
+  } else if (float_win) {
+    wp->w_hl_attr_normal = HL_ATTR(HLF_NFLOAT) >
```
bfredl, line 378: more readable with `0` on this line (also allowed by formatter)

glepnir: yes. I didn't look format carefully.. just make formatc highlight.c .

PR #15124, opened 2021-07-18
feat: `vim.merge`

folke:
This PR includes `vim.merge`, an utility function that is similar to `vim.tbl_deep_extend` with the **force** option, but also works for other types than tables.

> --- Merges the values similar to vim.tbl_deep_extend with the **force** behavior,
--- but the values can be any type, in which case they override the values on the left.
--- Values will me merged in-place in the first left-most table. If you want the result to be in
--- a new table, then simply pass an empty table as the first argument `vim.merge({}, ...)`

## Note

This PR includes a commit from another PR #15094. Not sure how that works 😅. I assume when the other PR is merged, this will also be able to merge.

PR #30309, opened 2024-09-08
fix(lua): revert vim.tbl_deep_extend behavior change and document it.

clason:
# Problem: 

`vim.tbl_deep_extend` had an undocumented feature where arrays (integer-indexed tables) were not merged but compared literally (used for merging default and user config, where one list should
overwrite the other completely). Turns out this behavior was relied on in quite a number of plugins (even though it wasn't a robust solution even for that use case, since lists of tables (e.g., plugin specs) can be array-like as well).

# Solution: 

Revert the removal of this special feature. Check for list-like (contiguous integer indices starting from 1) instead, as this is closer to the intent. Document this behavior.

Follow-up should be to add a new (better) `vim.merge` API that covers all current use cases and deprecates these legacy functions. It should also offer more -- and more explicit -- control. (See, e.g., https://github.com/neovim/neovim/pull/15124#issuecomment-2336685680.)

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2230,6 +2230,12 @@ vim.tbl_count({t})                                           *vim.tbl_count()*
 vim.tbl_deep_extend({behavior}, {...})                 *vim.tbl_deep_extend()*
     Merges recursively two or more tables.
 
+    Only values that are empty tables or tables that are not |lua-list|s
+    (indexed by consecutive integers starting from 1) are merged recursively.
+    This is useful for merging nested tables like default and user
+    configurations where lists should be treated as literals (i.e., are
+    overwritten instead of merged).
+
     Parameters: ~
       • {behavior}  (`'error'|'keep'|'force'`) Decides what to do if a key is
                     found in more than one map:
--- a/test/functional/lua/vim_spec.lua
+++ b/test/functional/lua/vim_spec.lua
@@ -1071,12 +1071,11 @@ describe('lua stdlib', function()
     ]])
     )
 
-    -- Fix github issue #23654
     ok(exec_lua([[
-      local a = { sub = { [1] = 'a' } }
-      local b = { sub = { b = 'a' } }
+      local a = { sub = { 'a', 'b' } }
+      local b = { sub = { 'b', 'c' } }
       local c = vim.tbl_deep_extend('force', a, b)
-      return vim.deep_equal(c, { sub = { [1] = 'a', b = 'a' } })
+      return vim.deep_equal(c, { sub = { 'b', 'c' } })
     ]]))
 
     matches('invalid "behavior": nil', pcall_err(exec_lua, [[return vim.tbl_deep_extend()]]))
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -214,9 +214,6 @@ These existing features changed their behavior.
   more emoji characters than before, including those encoded with multiple
   emoji codepoints combined with ZWJ (zero width joiner) codepoints.
 
-• |vim.tbl_deep_extend()| no longer ignores any values for which |vim.isarray()|
-  returns `true`.
-
 ==============================================================================
 REMOVED FEATURES                                                 *news-removed*
 
--- a/runtime/lua/vim/shared.lua
+++ b/runtime/lua/vim/shared.lua
@@ -354,6 +354,12 @@ function vim.tbl_isempty(t)
   return next(t) == nil
 end
 
+--- We only merge empty tables or tables that are not list-like (indexed by consecutive integers
+--- starting from 1)
+local function can_merge(v)
+  return type(v) == 'table' and (vim.tbl_isempty(v) or not vim.islist(v))
+end
+
 --- Recursive worker for tbl_extend
 --- @param behavior 'error'|'keep'|'force'
 --- @param deep_extend boolean
@@ -368,7 +374,7 @@ local function tbl_extend_rec(behavior, deep_extend, ...)
     local tbl = select(i, ...) --[[@as table<any,any>]]
     if tbl then
       for k, v in pairs(tbl) do
-        if deep_extend and type(v) == 'table' and type(ret[k]) == 'table' then
+        if deep_extend and can_merge(v) and can_merge(ret[k]) then
           ret[k] = tbl_extend_rec(behavior, true, ret[k], v)
         elseif behavior ~= 'force' and ret[k] ~= nil then
           if behavior == 'error' then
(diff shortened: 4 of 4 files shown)


runtime/doc/lua.txt
```diff
@@ -2230,6 +2230,11 @@ vim.tbl_count({t})                                           *vim.tbl_count()*
 vim.tbl_deep_extend({behavior}, {...})                 *vim.tbl_deep_extend()*
     Merges recursively two or more tables.
 
+    Only empty tables or tables that are not list-like (indexed by consecutive
```
justinmk, line 2233: link to  `:help lua-list`  ?

test/functional/lua/vim_spec.lua
```diff
@@ -1071,15 +1071,15 @@ describe('lua stdlib', function()
...
-    -- Fix github issue #23654
-    ok(exec_lua([[
-      local a = { sub = { [1] = 'a' } }
-      local b = { sub = { b = 'a' } }
-      local c = vim.tbl_deep_extend('force', a, b)
-      return vim.deep_equal(c, { sub = { [1] = 'a', b = 'a' } })
-    ]]))
-
-    matches('invalid "behavior": nil', pcall_err(exec_lua, [[return vim.tbl_deep_extend()]]))
+    matches(
```
justinmk, line 1074: this matches() case looks like a style-only change

clason: Yes, the revert dropped a style-only change in the meantime. Will re-revert.

clason:
That was the only use of `isarray` in our codebase -- should I go ahead and deprecate it while I'm at it?

runtime/doc/lua.txt
```diff
@@ -2230,6 +2230,11 @@ vim.tbl_count({t})                                           *vim.tbl_count()*
 vim.tbl_deep_extend({behavior}, {...})                 *vim.tbl_deep_extend()*
     Merges recursively two or more tables.
 
+    Only empty tables or tables that are not |lua-list|s (indexed by
+    consecutive integers starting from 1) are merged. This is useful for
```
echasnovski, line 2234: For reference, the reverted implementation (the one present in 0.10.1) does not do this on "first level" of merging. I.e.:
 - `vim.tbl_deep_extend('force', {'a', 'b'}, {'c'})` -> `{ 'c', 'b' }`
 - `vim.tbl_deep_extend('force', {sub = {'a', 'b'} }, { sub = { 'c' } })` -> `{ sub = { 'c' } }`.

This was the whole point of #23654. Not sure if this still requires a fix or should be documented as intended.

clason: This is very likely intended/expected (note that tables of tables -- like plugin specs -- are also list-like). I'll try to document this.

clason: (And, yes, this is an inconsistency which would be better to fix; alas, we cannot do so without too much breakage, as it turns out. Hence we need a proper deprecation cycle, with a new API that is rational.)

runtime/doc/lua.txt
```diff
@@ -2230,6 +2230,12 @@ vim.tbl_count({t})                                           *vim.tbl_count()*
 vim.tbl_deep_extend({behavior}, {...})                 *vim.tbl_deep_extend()*
     Merges recursively two or more tables.
 
+    Only values that are empty tables or tables that are not |lua-list|s
+    (indexed by consecutive integers starting from 1) are merged recursively.
+    This is useful for merging nested tables like default and user
+    configurations where lists should be treated as literals (i.e., are
```
justinmk, line 2236:
```suggestion
    configurations where nested lists should be treated as literals (i.e., are
```

clason: Does it make a difference (honestly asking)?

runtime/doc/lua.txt
```diff
@@ -2230,6 +2230,12 @@ vim.tbl_count({t})                                           *vim.tbl_count()*
 vim.tbl_deep_extend({behavior}, {...})                 *vim.tbl_deep_extend()*
     Merges recursively two or more tables.
 
+    Only values that are empty tables or tables that are not |lua-list|s
+    (indexed by consecutive integers starting from 1) are merged recursively.
```
justinmk, line 2234: if we are talking about "merging recursively", maybe it's not worth mentioning empty tables?

```suggestion
    Only tables that are not |lua-list|s
    (indexed by consecutive integers starting from 1) are merged recursively.
```

clason: No, these are special cased (_and_ list-like), so this is important to call out.

echasnovski: The current "values that are ..." is a way to document [this behavior](https://github.com/neovim/neovim/pull/30309#discussion_r1749287011).

justinmk (approved):

runtime/doc/lua.txt
justinmk: Are top-level lists merged?

Anyway, I think the "nested tables" on the previous line perhaps is the important part, so my suggestion can be ignored.

justinmk:
> That was the only use of `isarray` in our codebase -- should I go ahead and deprecate it while I'm at it?

Needs a separate PR assuming we want to actually tackle that. IMO it seems useful to keep it so that we have a programmatic definition of the "spec" for an array.

runtime/doc/lua.txt
clason: > Are top-level lists merged?

No, as @echasnovski points out above. Only lists that are _values_ are special-cased.

clason: (Always happy about suggestions for improvement; I'd just like to understand them to learn from them.)

clason:
> Needs a separate PR assuming we want to actually tackle that.

Of course (or at least a separate, non-squashed, commit).

> IMO it seems useful to keep it so that we have a programmatic definition of the "spec" for an array. (Not to mention, if anyone finds a use-case, it would feel pretty bad that we removed it.)

Well, the idea was that adding a deprecation would be the fastest way of finding out (the hard way) ;) 

But no strong feelings either way myself.

mfussenegger (approved):

runtime/doc/lua.txt
folke: fyi, not sure if you're talking about lazy's plugin specs, but there @echasnovski's first example would yield `{'c'}`, which does indeed make more sense. So I'm also a proponent to have this fixed (either now, or in a later iteration)

folke (approved):
Looks good to me!

And thank you for the quick turn-around. I'm sure this will avoid lots of mysterious issues popping up all over Github :)

runtime/doc/lua.txt
clason: I was talking more generally, as an excuse not to get my hands too dirty. If you can promise to keep the general pitchforks at bay, I can look at extending `can_merge` to the top-level scope. That will likely take a bit more time since #dayjob will reassert itself powerfully soon.

folke: The most important thing right now is the PR in its current state.
In the context of option/config merging, the top level tables would not be list-like, so this problem does not happen there at least.

I'm willing to work on a `vim.merge` PR that includes this fix this week (after this PR gets merged).

GitMurf: If either of you need / want testing or feedback from an end user perspective as this progresses, please don't hesitate to ask!

PR #30313, opened 2024-09-08
feat(treesitter): start moving get_parser to return nil

ribru17:
**Problem:** `vim.treesitter.get_parser` will throw an error if no parser can be found.

- This means the caller is responsible for wrapping it in a `pcall`, which is easy to forget
- It's a bit unintuitive since many other `get_*` style functions conventionally return `nil` if no object is found (e.g. `get_node`, `get_lang`, `query.get`, etc.)
- It also makes it slightly harder to potentially memoize `get_parser` in the future

**Solution:** Return `nil` if no parser can be found or created

- This requires a function signature change, and some new assertions in places where the parser will always (or should always) be found.
- This is breaking because users who already `pcall`ed this function will see that it now returns true (success) with a nil value in places where it used to return false (failure) with an error message

### DIFF
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -9,8 +9,8 @@
 local function get_commentstring(ref_position)
   local buf_cs = vim.bo.commentstring
 
-  local has_ts_parser, ts_parser = pcall(vim.treesitter.get_parser)
-  if not has_ts_parser then
+  local ts_parser = vim.treesitter._get_parser()
+  if not ts_parser then
     return buf_cs
   end
 
--- a/runtime/lua/vim/treesitter.lua
+++ b/runtime/lua/vim/treesitter.lua
@@ -74,14 +74,14 @@ end
 
 --- Returns the parser for a specific buffer and attaches it to the buffer
 ---
---- If needed, this will create the parser.
+--- If needed, this will create the parser. If no parser can be found or created, returns `nil`.
 ---
 ---@param bufnr (integer|nil) Buffer the parser should be tied to (default: current buffer)
 ---@param lang (string|nil) Language of this parser (default: from buffer filetype)
 ---@param opts (table|nil) Options to pass to the created language tree
 ---
----@return vim.treesitter.LanguageTree object to use for parsing
-function M.get_parser(bufnr, lang, opts)
+---@return vim.treesitter.LanguageTree? object to use for parsing, or `nil` if not found
+function M._get_parser(bufnr, lang, opts)
   opts = opts or {}
 
   if bufnr == nil or bufnr == 0 then
@@ -94,23 +94,42 @@ function M.get_parser(bufnr, lang, opts)
 
   if not valid_lang(lang) then
     if not parsers[bufnr] then
-      error(
-        string.format(
-          'There is no parser available for buffer %d and one could not be'
-            .. ' created because lang could not be determined. Either pass lang'
-            .. ' or set the buffer filetype',
-          bufnr
-        )
-      )
+      return nil
     end
   elseif parsers[bufnr] == nil or parsers[bufnr]:lang() ~= lang then
-    assert(lang, 'lang should be valid')
-    parsers[bufnr] = M._create_parser(bufnr, lang, opts)
+    local parser = vim.F.npcall(M._create_parser, bufnr, lang, opts)
+    if not parser then
+      return nil
+    end
+    parsers[bufnr] = parser
   end
 
   parsers[bufnr]:register_cbs(opts.buf_attach_cbs)
 
   return parsers[bufnr]
+end
+
+--- Returns the parser for a specific buffer and attaches it to the buffer
+---
+--- If needed, this will create the parser.
+---
+---@param bufnr (integer|nil) Buffer the parser should be tied to (default: current buffer)
+---@param lang (string|nil) Language of this parser (default: from buffer filetype)
+---@param opts (table|nil) Options to pass to the created language tree
+---
+---@return vim.treesitter.LanguageTree object to use for parsing
+function M.get_parser(bufnr, lang, opts)
+  -- TODO(ribru17): Remove _get_parser and move that logic back here once the breaking function
+  -- signature change is acceptable.
+  local parser = M._get_parser(bufnr, lang, opts)
+  if not parser then
+    vim.notify_once(
+      'WARNING: vim.treesitter.get_parser will return nil instead of raising an error in Neovim 0.12',
+      vim.log.levels.WARN
+    )
+    error('Parser not found.')
+  end
+  return parser
 end
 
 --- Returns a string parser
@@ -386,7 +405,7 @@ function M.get_node(opts)
 
   local ts_range = { row, col, row, col }
 
-  local root_lang_tree = M.get_parser(bufnr, opts.lang)
+  local root_lang_tree = M._get_parser(bufnr, opts.lang)
   if not root_lang_tree then
     return
   end
@@ -419,7 +438,11 @@ end
 ---@param lang (string|nil) Language of the parser (default: from buffer filetype)
 function M.start(bufnr, lang)
   bufnr = bufnr or api.nvim_get_current_buf()
-  local parser = M.get_parser(bufnr, lang)
+  local parser = M._get_parser(bufnr, lang)
+  if not parser then
+    vim.notify('No parser for the given buffer.', vim.log.levels.WARN)
+    return
+  end
   M.highlighter.new(parser)
 end
 
--- a/runtime/lua/vim/treesitter/_fold.lua
+++ b/runtime/lua/vim/treesitter/_fold.lua
@@ -114,7 +114,7 @@ local function compute_folds_levels(bufnr, info, srow, erow, parse_injections)
   srow = srow or 0
   erow = erow or api.nvim_buf_line_count(bufnr)
 
-  local parser = ts.get_parser(bufnr)
+  local parser = assert(ts._get_parser(bufnr))
 
   parser:parse(parse_injections and { srow, erow } or nil)
 
@@ -392,7 +392,7 @@ function M.foldexpr(lnum)
   lnum = lnum or vim.v.lnum
   local bufnr = api.nvim_get_current_buf()
 
-  local parser = vim.F.npcall(ts.get_parser, bufnr)
+  local parser = ts._get_parser(bufnr)
   if not parser then
     return '0'
   end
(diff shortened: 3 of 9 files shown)


clason:
I'm very much in favor of this change (`pcall` is not a first-class pattern we want to push for the general API; errors should be reserved for exceptional cases, not "graceful failures"), but the breakage concerns are very real. We should try to come up with a gentle migration strategy before landing this.

ribru17:
Sounds good, lmk if this is what you had in mind

runtime/lua/vim/treesitter.lua
```diff
@@ -94,25 +94,43 @@ function M.get_parser(bufnr, lang, opts)
 
   if not valid_lang(lang) then
     if not parsers[bufnr] then
-      error(
-        string.format(
-          'There is no parser available for buffer %d and one could not be'
-            .. ' created because lang could not be determined. Either pass lang'
-            .. ' or set the buffer filetype',
```
ribru17, line 101: Should this message be kept as a `vim.notify`? It would cause extra noise in e.g. `_comment.lua`

runtime/lua/vim/treesitter.lua
```diff
@@ -94,25 +94,43 @@ function M.get_parser(bufnr, lang, opts)
...
+---@param lang (string|nil) Language of this parser (default: from buffer filetype)
+---@param opts (table|nil) Options to pass to the created language tree
+---
+---@return vim.treesitter.LanguageTree object to use for parsing
+function M.get_parser(bufnr, lang, opts)
+  -- TODO(ribru17): Remove _get_parser and move that logic back here once the breaking function
+  -- signature change is acceptable.
+  local parser = M._get_parser(bufnr, lang, opts)
+  if not parser then
+    error(
```
clason, line 126: I think such messages should be `notify_once` (like `vim.deprecate`) to avoid spam (in addition to a simple `error` to trip `pcall`). We should also be 1. more concrete (specify versions) and 2. more concise (try to keep messages to a single line to avoid "press enter")

echasnovski: My guess is that non-trivial amount of usages rely on `pcall` to test if there is a parser (like in `vim._comment`). So this error will be not visible until "rip the band-aid off" stage of deprecation.

It's just an observation though, as having this is better than immediate behavior change and I don't have better suggestions (apart from trying to advertise this as much as reasonably possible).

runtime/lua/vim/_comment.lua
```diff
@@ -9,8 +9,8 @@
 local function get_commentstring(ref_position)
   local buf_cs = vim.bo.commentstring
 
-  local has_ts_parser, ts_parser = pcall(vim.treesitter.get_parser)
-  if not has_ts_parser then
+  local ts_parser = vim.treesitter._get_parser()
```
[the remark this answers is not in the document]
clason: No, the point is to get rid of pcall. It's not the right idiom for this.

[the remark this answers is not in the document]
echasnovski: The `pcall()` will be used not to "test if there is parser", but to "avoid any error at all cost". Because working with tree-sitter is not the main priority here. But again, this is mostly a style preference.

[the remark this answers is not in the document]
clason: > "avoid any error at all cost".

What error could there be?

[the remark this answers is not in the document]
echasnovski: > What error could there be?

An error not known in advance. Which is how errors commonly appear.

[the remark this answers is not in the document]
clason: Does that mean we will have to start wrapping _every_ function in `pcall`? No. Either there is a parser, in which case it's returned, or there is not (for whatever reason), in which case `nil` is returned. Any errors that arise during the check are the responsibility of `get_parser` and should be handled there.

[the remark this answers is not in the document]
echasnovski: My point was that the act of commenting does not directly linked to tree-sitter. So avoiding any errors from it for a graceful fallback would be my choice here.

However (for the third time), this is a style preference. Meaning I don't really have strong opinion on this (otherwise it would have been stated that way) and defer the choice to the author.

runtime/lua/vim/treesitter/_fold.lua
```diff
@@ -392,7 +394,7 @@ function M.foldexpr(lnum)
   lnum = lnum or vim.v.lnum
   local bufnr = api.nvim_get_current_buf()
 
-  local parser = vim.F.npcall(ts.get_parser, bufnr)
+  local parser = ts._get_parser(bufnr)
```
justinmk, line 397: (off topic) perhaps everywhere we use `npcall` should be reevaluated...

runtime/lua/vim/treesitter/_fold.lua
```diff
@@ -114,7 +114,9 @@ local function compute_folds_levels(bufnr, info, srow, erow, parse_injections)
   srow = srow or 0
   erow = erow or api.nvim_buf_line_count(bufnr)
 
-  local parser = ts.get_parser(bufnr)
+  local parser = ts._get_parser(bufnr)
+
+  assert(parser)
```
justinmk, line 119:
```suggestion
  local parser = assert(ts._get_parser(bufnr))
```

runtime/lua/vim/treesitter/_query_linter.lua
```diff
@@ -172,7 +172,8 @@ function M.lint(buf, opts)
     --- @type (table|nil)
     local parser_info = vim.F.npcall(vim.treesitter.language.inspect, lang)
 
-    local parser = vim.treesitter.get_parser(buf)
+    local parser = vim.treesitter._get_parser(buf)
+    assert(parser, 'Query parser not found.')
```
justinmk, line 176:
```suggestion
    local parser = assert(vim.treesitter._get_parser(buf), 'Query parser not found.')
```

runtime/lua/vim/treesitter/dev.lua
```diff
@@ -539,7 +538,8 @@ local edit_ns = api.nvim_create_namespace('treesitter/dev-edit')
 local function update_editor_highlights(query_win, base_win, lang)
   local base_buf = api.nvim_win_get_buf(base_win)
   local query_buf = api.nvim_win_get_buf(query_win)
-  local parser = vim.treesitter.get_parser(base_buf, lang)
+  local parser = vim.treesitter._get_parser(base_buf, lang)
+  assert(parser)
```
justinmk, line 542:
```suggestion
  local parser = assert(vim.treesitter._get_parser(base_buf, lang))
```

runtime/lua/vim/vimhelp.lua
```diff
@@ -33,7 +33,8 @@ end
 --- Show a table of contents for the help buffer in a loclist
 function M.show_toc()
   local bufnr = vim.api.nvim_get_current_buf()
-  local parser = vim.treesitter.get_parser(bufnr, 'vimdoc')
+  local parser = vim.treesitter._get_parser(bufnr, 'vimdoc')
+  assert(parser, 'Vimdoc parser not found.')
```
justinmk, line 37:
```suggestion
  local parser = assert(vim.treesitter._get_parser(bufnr, 'vimdoc'), 'vimdoc parser not found.')
```

scripts/gen_help_html.lua
```diff
@@ -786,7 +786,8 @@ local function parse_buf(fname, parser_path)
   if parser_path then
     vim.treesitter.language.add('vimdoc', { path = parser_path })
   end
-  local lang_tree = vim.treesitter.get_parser(buf)
+  local lang_tree = vim.treesitter._get_parser(buf)
+  assert(lang_tree, 'Vimdoc parser not found.')
```
justinmk, line 790:
```suggestion
  local lang_tree = assert(vim.treesitter._get_parser(buf), 'vimdoc parser not found.')
```

runtime/lua/vim/_comment.lua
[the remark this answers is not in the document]
clason: My point is to clearly state -- especially for future reference -- the official style (avoid pcall within our own sphere of influence in favor of returning `nil` if possible -- pcall is for guarding _external_ functions you cannot control, but here we control the whole code path).

runtime/lua/vim/treesitter/_fold.lua
clason: That is the hope, yes (one API at a time...)

clason:
@ribru17 I think my suggestion wasn't fully thought out (and the PR was merged before I could get to it): This will spam a warning to users, and plugins don't have a clear migration part until 0.12.

I think we should go with the same strategy as for the `iter_matches` change (add an `error = true` (default) optional parameter and mark the change and deprecation in the documentation only. Treesitter being experimental allows for an accelerated deprecation strategy (so we can remove without flipping default first).

@gpanders @mfussenegger

mathjiajia:
Agree with you. The warning msg is quite annoying.

ofseed:
@clason When proposing https://github.com/neovim/neovim/pull/30353, I thought about the current design and suspect that the design of `vim.treesitter.add` isn't very reasonable too. It doesn't always **add** a parser but rather **loads** it. This means that after we `add("python")`, performing the same action again won't load the python parser twice. So, I actually want to change this behavior as well.

Regarding this PR, I'm wondering if we can merge `vim.treesitter.language.add` and `vim.treesitter.get_parser` into a new function, then let `get_parser` retain its behavior before this PR and eventually deprecate it.

clason:
No, that would be too breaking. (I'm not going to touch naming with a ten-foot pole.)

clason:
But that does raise a point: `vim.treesitter.language.add` should similarly not raise an error but instead provide reasonable return values (`nil` for not possible). 

(Ideally, it should return a Lua object for the loaded `TSLanguage` object, but that requires changes to the C level.)

ribru17:
> @ribru17 I think my suggestion wasn't fully thought out (and the PR was merged before I could get to it): This will spam a warning to users, and plugins don't have a clear migration part until 0.12.
> 
> I think we should go with the same strategy as for the `iter_matches` change (add an `error = true` (default) optional parameter and mark the change and deprecation in the documentation only. Treesitter being experimental allows for an accelerated deprecation strategy (so we can remove without flipping default first).
> 
> @gpanders @mfussenegger

That makes sense- in this case should the `get_parser` Luadoc reflect that it can now return nil?

clason:
Yes, I thought so.

PR #30315, opened 2024-09-09
fix(lsp): wrong startcol and avoid pum deleted

glepnir:
Problem:
1. comletion prefix value sub index not correct. which cause
       input `wp.` completion in there after do and type w the completion
       window will be delete.
 2. wrong complete-word generate this will include symbol like `->` ,
       it will make compl_match_array build failed. because the leader
       is character it can not find in this complete-word(cp_str).

Solution:
1. use max value of server_bounary, word_boundary
2. use vimregex keyword remove unncessary symbols.

before

now

After

after

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -154,8 +154,13 @@ local function get_completion_word(item)
       return item.label
     end
   elseif item.textEdit then
-    local word = item.textEdit.newText
-    return word:match('^(%S*)') or word
+    local word = item.textEdit.newText:gsub('%s+$', '')
+    if word:match('^%W') then
+      -- remove unnecessary symbols in label like "->w_locked" => "w_locked"
+      local matched_word = vim.fn.matchstr(word, '\\k\\+')
+      return matched_word ~= '' and matched_word or word
+    end
+    return word
   elseif item.insertText and item.insertText ~= '' then
     return item.insertText
   end
@@ -365,7 +370,8 @@ function M._convert_results(
   elseif curstartbyte ~= nil and curstartbyte ~= server_start_boundary then
     server_start_boundary = client_start_boundary
   end
-  local prefix = line:sub((server_start_boundary or client_start_boundary) + 1, cursor_col)
+  local prefix =
+    line:sub(math.max(server_start_boundary or 0, client_start_boundary) + 1, cursor_col)
   local matches = M._lsp_to_complete_items(result, prefix, client_id)
   return matches, server_start_boundary
 end
@@ -461,7 +467,7 @@ local function trigger(bufnr, clients)
         vim.list_extend(matches, client_matches)
       end
     end
-    local start_col = (server_start_boundary or word_boundary) + 1
+    local start_col = math.max(server_start_boundary or 0, word_boundary) + 1
     vim.fn.complete(start_col, matches)
   end)
 
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -218,6 +218,49 @@ describe('vim.lsp.completion: item conversion', function()
     eq(expected, result)
   end)
 
+  it('removes unnecessary symbols in label', function()
+    local items = {
+      {
+        detail = 'colnr_T',
+        filterText = 'w_virtcol',
+        insertText = '->w_virtcol',
+        insertTextFormat = 1,
+        kind = 5,
+        label = ' w_virtcol',
+        score = 0.59581798315048,
+        sortText = '40e77879w_virtcol',
+        textEdit = {
+          newText = '->w_virtcol',
+          range = {
+            ['end'] = {
+              character = 0,
+              line = 0,
+            },
+            start = {
+              character = 0,
+              line = 0,
+            },
+          },
+        },
+      },
+    }
+    local result = complete('|', items)
+    result = vim.tbl_map(function(x)
+      return {
+        abbr = x.abbr,
+        word = x.word,
+      }
+    end, result.items)
+
+    local expected = {
+      {
+        abbr = ' w_virtcol',
+        word = 'w_virtcol',
+      },
+    }
+    eq(expected, result)
+  end)
+
   it('prefers wordlike components for snippets', function()
     -- There are two goals here:
     --


runtime/lua/vim/lsp/completion.lua
```diff
@@ -265,7 +270,7 @@ function M._lsp_to_complete_items(result, prefix, client_id)
       end
       local completion_item = {
         word = word,
-        abbr = item.label,
+        abbr = vim.trim(item.label),
```
MariaSolOs, line 273: Not sure about this trim. If there's leading whitespace I feel like either the label is incorrect or the space is intentional.

mfussenegger (commented):
> if isIncomplete is true and next request matches is empty invoke complete function will remove current pum.

I'm not sure about this. As I understand it, the intention behind isIncomplete is to allow both extending and narrowing the result by typing more characters. If there are no more matches the popup menu should disappear. Is this change really necessary to get the desired behaviour? 

The other two changes seem okay to me but an additional test case that illustrates the textEdit -> word change would be good

glepnir:
seems like vscode also add new completion list when length > 0 
https://github.com/microsoft/vscode/blob/789c320a1c67bb0547cdef2b8b82634bf93929a3/extensions/npm/src/features/jsonContributions.ts#L142

runtime/lua/vim/lsp/completion.lua
glepnir: https://github.com/llvm/llvm-project/blob/7ba6768df8181bc270763333969d4a1d6cc2e160/mlir/lib/Tools/mlir-pdll-lsp-server/PDLLServer.cpp#L709

there is no trim so we can assume that the server did this on purpose?

mfussenegger: clangd prefixes some items with a •, so I assume the space is to keep all items aligned:

```
 a                                                                Variable   std::unique_ptr<A>
 A                                                                Struct
•ADJ_ESTERROR                                                     Text
•ADJ_FREQUENCY                                                    Text
```

mfussenegger:
> seems like vscode also add new completion list when length > 0 https://github.com/microsoft/vscode/blob/789c320a1c67bb0547cdef2b8b82634bf93929a3/extensions/npm/src/features/jsonContributions.ts#L142

`items.length > 0 || isIncomplete` - the `|| isIncomplete` would match my point about it being able to narrow/finish the completion if there are no more candidates, no?
I still don't understand why you want to get rid of this behavior.

If I understand the problem description of the PR right, only the `word` change is required, and the trim + empty matches change could be reverted?

glepnir:
Nope .  `wp.` will return completion items with isInComplete = true type w send request and response is empty pum delete. this is not work as expect. vscode works like after gif. still can get completion.

glepnir:
yes you are right . i check it again because the wrong prefix value in _convert_results it sub prefix to `.w` then client_matchs is empty in second request.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -154,8 +154,13 @@ local function get_completion_word(item)
...
   elseif item.textEdit then
-    local word = item.textEdit.newText
-    return word:match('^(%S*)') or word
+    local word = item.textEdit.newText:gsub('%s+$', '')
+    if word:match('^%W') then
+      local matched_word = vim.fn.matchstr(word, '\\k\\+')
+      return matched_word ~= '' and matched_word or word
+    else
+      return word
+    end
```
justinmk, line 163: nit:
```suggestion
    end

    return word
```

runtime/lua/vim/lsp/completion.lua
```diff
@@ -154,8 +154,13 @@ local function get_completion_word(item)
       return item.label
     end
   elseif item.textEdit then
-    local word = item.textEdit.newText
-    return word:match('^(%S*)') or word
+    local word = item.textEdit.newText:gsub('%s+$', '')
+    if word:match('^%W') then
+      local matched_word = vim.fn.matchstr(word, '\\k\\+')
```
justinmk, line 159: It's often useful to show a short input/output example
```suggestion
      -- "..." => "..."
      local matched_word = vim.fn.matchstr(word, '\\k\\+')
```

mfussenegger (approved):

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -218,6 +218,49 @@ describe('vim.lsp.completion: item conversion', function()
     eq(expected, result)
   end)
 
+  it('remove uncessary symbols in label', function()
```
justinmk, line 221:
```suggestion
  it('remove unnecessary symbols in label', function()
```

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -218,6 +218,49 @@ describe('vim.lsp.completion: item conversion', function()
...
 
+  it('removes unnecessary symbols in label', function()
+    local items = {
+      {
+        detail = 'colnr_T',
+        filterText = 'w_virtcol',
+        insertText = '->w_virtcol',
+        insertTextFormat = 1,
+        kind = 5,
+        label = ' w_virtcol',
```
justinmk, line 229: why is this preceded by whitespace (also the `abbr` in the result below)? The test just above this one tests that whitespace is supposed to be trimmed...

glepnir: This is the result of lsp clangd returns some results with `•function_name` The space is to distinguish the period. space in there is necessary.

justinmk:
is the commit message (and subject line) up to date?

justinmk:
It would be really helpful to mention (in PR description and commit message), the user-facing effect. The current description is just details about the implementation. What is the *user-facing* symptom that is being fixed? Like this:

```
before: input => result

    foo => bar

after:

    foo2 => bar2

```

glepnir:
in github page is not latest..i have update commit message before. but not this format what you suggest.. i will update again.

justinmk:
Oh sorry, I should have looked at your screenshots. I think I understand now, but I'll wait for your update. Thank you!

mfussenegger:
It looks like this breaks completion for servers like vscode-html-ls where so far you could complete ending tags. E.g. for `<li>foo<|` you got something like:

```
dap> result
table: 0x7466baa2cf28
  isIncomplete: false
  items: table: 0x7466baa2cfa8
    1: table: 0x7466baa2cff0
      filterText: /li
      insertTextFormat: 1
      kind: 10
      label: /li
      textEdit: table: 0x7466baa2d0d8
        newText: /li>
        range: table: 0x7466baa2d120
          end: table: 0x7466baa2d1e8
          start: table: 0x7466baa2d168
dap> candidates
table: 0x7466ba8d8398
  1: table: 0x7466ba8d8c68
    abbr: /li
    dup: 1
    empty: 1
    hl_group: 
    icase: 1
    info: 
    kind: Property
    menu: 
    user_data: table: 0x7466ba8d90c0
    word: /li>
```

With this change the `word` is just `li`:

```
dap> candidates
table: 0x78a2de10e290
  1: table: 0x78a2ddf663e0
    abbr: /li
    dup: 1
    empty: 1
    hl_group: 
    icase: 1
    info: 
    kind: Property
    menu: 
    user_data: table: 0x78a2de562570
    word: li
```

I'm afraid we'll have to find another solution to this.

glepnir:
I also got `li` in master wired.

glepnir:
I think the problem is not in lua. No matter how you modify it, it won't work in all cases. It's not a good idea to clean word. Some symbols are still useful in some files such as html. So I think the problem is in C. lsp provides filtertext as a filter and we are still using word. In C it is cp_str. So I modified some C code. It works well. Can you try it? Thanks.

```diff
diff --git a/runtime/lua/vim/lsp/completion.lua b/runtime/lua/vim/lsp/completion.lua
index 71ea2df10..dd5f1ee00 100644
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -265,6 +265,7 @@ function M._lsp_to_complete_items(result, prefix, client_id)
       end
       local completion_item = {
         word = word,
+        filter_text = vim.tbl_get(item, 'filterText'),
         abbr = item.label,
         kind = protocol.CompletionItemKind[item.kind] or 'Unknown',
         menu = item.detail or '',
diff --git a/src/nvim/insexpand.c b/src/nvim/insexpand.c
index c17bd27da..c926b5b7e 100644
--- a/src/nvim/insexpand.c
+++ b/src/nvim/insexpand.c
@@ -156,7 +156,8 @@ static char *ctrl_x_mode_names[] = {
 #define CPT_MENU        1   ///< "menu"
 #define CPT_KIND        2   ///< "kind"
 #define CPT_INFO        3   ///< "info"
-#define CPT_COUNT       4   ///< Number of entries
+#define CPT_FILTER      4   ///< "filter_text"
+#define CPT_COUNT       5   ///< Number of entries

 /// Structure used to store one match for insert completion.
 typedef struct compl_S compl_T;
@@ -934,10 +935,13 @@ static bool ins_compl_equal(compl_T *match, char *str, size_t len)
   if (match->cp_flags & CP_EQUAL) {
     return true;
   }
+  const char *filter_text = match->cp_text[CPT_FILTER]
+                            && strlen(match->cp_text[CPT_FILTER]) >
+                            0 ? match->cp_text[CPT_FILTER] : match->cp_str;
   if (match->cp_flags & CP_ICASE) {
-    return STRNICMP(match->cp_str, str, len) == 0;
+    return STRNICMP(filter_text, str, len) == 0;
   }
-  return strncmp(match->cp_str, str, len) == 0;
+  return strncmp(filter_text, str, len) == 0;
 }

 /// Reduce the longest common string for match "match".
@@ -1175,11 +1179,18 @@ static int ins_compl_build_pum(void)
     XFREE_CLEAR(compl_leader);
...skipping...
+  size_t orig_text_len = strlen(compl_orig_text);
+  if (compl_orig_text && orig_text_len > 0 && lead_len > 1 && strncmp(compl_leader, compl_orig_text,
+                                                                      orig_text_len) == 0) {
+    leader = leader + orig_text_len;
+    lead_len = lead_len - (int)orig_text_len;
+  }

   do {
     // When 'completeopt' contains "fuzzy" and leader is not NULL or empty,
@@ -1190,7 +1201,7 @@ static int ins_compl_build_pum(void)

     if (!match_at_original_text(comp)
         && (compl_leader == NULL
-            || ins_compl_equal(comp, compl_leader, (size_t)lead_len)
+            || ins_compl_equal(comp, leader, (size_t)lead_len)
             || (compl_fuzzy_match && comp->cp_score > 0))) {
       compl_match_arraysize++;
     }
@@ -1208,7 +1219,7 @@ static int ins_compl_build_pum(void)
   // match after it, don't highlight anything.
   bool shown_match_ok = match_at_original_text(compl_shown_match);

-  if (strequal(compl_leader, compl_orig_text) && !shown_match_ok) {
+  if (strequal(leader, compl_orig_text) && !shown_match_ok) {
     compl_shown_match = compl_no_select ? compl_first_match : compl_first_match->cp_next;
   }

@@ -1220,7 +1231,7 @@ static int ins_compl_build_pum(void)
   do {
     if (!match_at_original_text(comp)
         && (compl_leader == NULL
-            || ins_compl_equal(comp, compl_leader, (size_t)lead_len)
+            || ins_compl_equal(comp, leader, (size_t)lead_len)
             || (compl_fuzzy_match && comp->cp_score > 0))) {
       if (!shown_match_ok && !compl_fuzzy_match) {
         if (comp == compl_shown_match || did_find_shown_match) {
@@ -2587,6 +2598,7 @@ static int ins_compl_add_tv(typval_T *const tv, const Direction dir, bool fast)
     cptext[CPT_MENU] = tv_dict_get_string(tv->vval.v_dict, "menu", true);
     cptext[CPT_KIND] = tv_dict_get_string(tv->vval.v_dict, "kind", true);
     cptext[CPT_INFO] = tv_dict_get_string(tv->vval.v_dict, "info", true);
+    cptext[CPT_FILTER] = tv_dict_get_string(tv->vval.v_dict, "filter_text", true);

     user_hlname = tv_dict_get_string(tv->vval.v_dict, "hl_group", false);
     user_hlattr = get_user_highlight_attr(user_hlname);

```

glepnir:
it works like this

what

justinmk:
Nice. So filter_text is defined in the same way as in [lsp](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/)?

```typescript
	/**
	 * A string that should be used when filtering a set of
	 * completion items. When omitted the label is used as the
	 * filter text for this item.
	 */
	filterText?: string;
```

glepnir:
here https://github.com/vim/vim/pull/15716 . word may have symbols we use word as `cp_str` for leader match this will cause no any matches found.

glepnir:
word is usually `textEdit.newText` , finally fallback to label.  check our `get_completion_word`. But newText will contain some necessary symbols such as pointer symbols `->` or html tag symbols `/li>`. lsp supports us to trigger completion with a dot after a pointer variable. This means that word as cp_str will contain symbols `->` neither `strncmp` nor `fuzzy_match_str` will get results. `filterText` is a clean existence. When isInComplete is false, filterText as cp_str is a more effective match leader than newText.  most detials can check the test cases which in linked vim PR .  any thought @mfussenegger

glepnir:
since not a lua part problem closed.

Issue #5035, opened 2016-07-11
Please support detaching and reattaching, like screen/tmux


joshtriplett:
With nvim's terminal mode, if it had the ability to detach and reattach, it could serve as a replacement for screen or tmux.

This would require a few additional bits:
- A `:detach` function, to stop running the text UI and daemonize into a --headless instance that will survive the current user session ending.
- A command-line argument to list running nvim instances.
- Command-line arguments to attach to a running nvim instance and start the text UI. These need three variations: one that just fails if the session already has an attached text UI, one that detaches any attached text UI first (like `screen -D -R`), and one that attaches without detaching any other attached text UI (like `screen -x`).  (That last one is optional, but I'd find it quite useful.)

justinmk:
See https://github.com/neovim/neovim/issues/2706#issuecomment-103715550

gsf:
A lightweight way to do this at present that's working well for me is to let abduco handle the session management:

```
abduco -A nvim-session nvim
```

lf-:
I'm currently using dtach for this: `dtach -A /tmp/nvim-sess -ze '^q' -r ctrl_l nvim`, but obviously I'm interested in this happening.

rashil2000:
Are there any plans for this? I use Neovim on Windows, but Windows lacks any terminal multiplexers (tmux works only in Msys2/Cygwin and that too only in mintty terminal). 

Splits/panes etc. for shells works fine, but Neovim is obviously missing the main tmux feature that is attach/detach.

aleksandersumowski:
@rashil2000 have you tried https://wezfurlong.org/wezterm/multiplexing.html?

hoschi:
@rashil2000 original Veonim had that feature as well, here is a fork of it https://github.com/smolck/uivonim
I don't know if there is another client implemented it which supports Windows.

hoschi:
@rashil2000 
my point was that this feature is already implemented in core, but not in the TUI. Neovim is already split up into server/client. In a few works the TUI delivered by core spawns and connects to one Neovim instance. What you miss here is the ability to manage different servers, which is not implemented in TUI.
Veonim implemented the server management. Unfortunately the original repository went away, here is a deep link to the functionality you want from my personal fork: https://github.com/hoschi/veonim/blob/master/docs/readme.md#multiple-concurrent-neovim-instances
I guessed, but not researched, that the fork I posted has the same functionality and you can find it by reading the docs.

[...]

PR #22137, opened 2023-02-06
fix(rpc)!: preseve files when stdio channel is closed

zeertzjq:
Fix #22096

BREAKING CHANGE: Unsaved changes are now preserved rather than discarded when stdio channel is closed.

PR #18414, opened 2022-05-04
docs: --remote alternatives

justinmk:
WIP: sketching out how Nvim `--remote` should work. (Followup to https://github.com/neovim/neovim/pull/17856#pullrequestreview-951127037 )

Key ideas:

* The numerous `--remote-x-y` variants from Vim should not be mindlessly copied. Most use of `--remote` is for small one-liner shortcuts. Any non-trivial use of Vim's `--remote` is not going to work with Nvim because we don't and won't have exactly the same interface. So we can instead just make a better interface.
* If possible, the only `--remote` variant we will support is... `--remote`. Not the other stuff.
* The usual Nvim CLI options like `-p`, `-es`, etc., should "just work". For example,
    * `--remote -p` should work, then `--remote-tab` isn't needed.
    * `--remote -es "+echo 1+1"` should work, then `--remote-expr` isn't needed.
    * `echo <keys> | nvim -s - --remote` should work, then `--remote-send` isn't needed.
        * Alternatively `nvim --remote "+call nvim_input('<keys>')"`

Another note about the current behavior:

    This doesn't work, all arguments after --remote will be used as file names: >
        nvim --remote --server ~/.cache/nvim/server.pipe file.txt

What is the purpose of that? We already have `--` (`:help ---`) for declaring "all args hereafter are filenames".

close #18531
close #18519

PR #30319, opened 2024-09-09
feat(ui): ctrl-z targets only the current UI

### LINKED ISSUE #23093: detach the current TUI (nvim continues headlessly)
### Problem

When connected to a remote ui executing `:q` will close the remote server, I guess this makes sense in the way that the command is being run by the backend and not the UI, but it's not intuitive or desired, I wish to quit without killing the backend

### Steps to reproduce

nvim --listen 127.0.0.1:8888 --headless
nvim --server 127.0.0.1:8888 --remote-ui
:q

### Expected behavior

The UI should disconnect from the server but not to kill it

### Neovim version (nvim -v)

0.10.x

no

### Operating system/version

macOS 13.2.1

### Terminal name/version

iTerm

xterm-256color

### Installation

homebrew

justinmk:
## Problem

- ctrl-z suspends *all* clients (and raises `VimSuspend`)
- No way for users/scripts to identify the "current UI".

Current behavior on master:

- after ctrl-z...
    - you can still connect to the server with a new UI
    - you can `fg` (foreground) any of the suspended UIs
    - the client processes don't exit after ctrl-z, you need to terminate them manually.

## Solution

See below...

### Status:

- `:detach` or `[count]ctrl-z` detaches the current UI. #32385
    - Works for all UIs ([`error_exit` UI event is sent](https://github.com/neovim/neovim/pull/34411) as an optional hint to clients before the channel is closed).
- Builtin TUI calls `nvim_set_client_info` with info including PID (`client.attributes.pid`).

### TODO:

- [ ] `nvim --connect` (without args) attaches to most-recent server.
- [ ] `:quit` should confirm if multiple UIs are attached.
- [ ] `chanclose({detach:boolean})` parameter.
    - ~~Clients can set `detach=true` on a channel, to ensure the server does not "self-exit" if the client closes its end of the channel.~~
- [ ] Change "regular suspend" to target only the "current" UI, instead of all connected UIs.
- [ ] Scripts can ask "what's the current UI?" Set `v:ui` or add a field to `nvim_list_uis()`.
- [ ] Docs:
    - Update `:help ctrl-z`, `:suspend`, ...
    - How to map ctrl-z so that it copies `v:servername` to clipboard.
- [ ] The "tmux use-case": keep the server running after unexpected disconnect/SIGHUP.
    - 🚧 Needs to be opt-in.
    - If the foreground process/your terminal unexpectedly disconnects, the Nvim server stays running. SIGTERM/SIGHUP on the client exits *only* the client, not the server.
    - ~~Notes on daemonizing: https://github.com/neovim/neovim/issues/5035#issuecomment-379408935~~
- [x] Move the impl out of `input.c` and plumb it through as proper keymapping.
    - Works with any UI, not only the TUI.
    - ~~Currently it works by special-casing `ctrl-z` in the TUI input stream.~~
    - Ideally, *input sent from a UI should be mapped to a channel id.*
        - Currently, specific input is not mapped, there is only the concept of a "most-recent (current) UI".
- [x] Require `[count]<c-z>`. Regular `<c-z>` without `[count]` should have its old behavior.
    - ~~For `--remote-ui` clients, ctrl-z old behavior really doesn't seem useful.~~
- [x] Support the normal case where `nvim` spawns the server as a child.
    - [use detach=true](https://github.com/neovim/neovim/issues/23093#issuecomment-2337867984)?
- [x] `:connect [server]` (or `:attach`?) performs "hot swap" (attach current UI to a different server). https://github.com/neovim/neovim/issues/5035#issuecomment-2343537003
    - TBD: don't know how to make this work for the "stdio" case. (workaround: always start the server as a daemon instead of using stdio).
- ~~show a warning if `ctrl-z` / `:suspend` is used server-side when UIs are attached.~~

closes https://github.com/neovim/neovim/issues/23093
closes https://github.com/neovim/neovim/pull/23441

### DIFF
--- a/src/nvim/api/private/helpers.c
+++ b/src/nvim/api/private/helpers.c
@@ -700,6 +700,9 @@ void api_set_error(Error *err, ErrorType errType, const char *format, ...)
   vsnprintf(err->msg, bufsize, format, args2);
   va_end(args2);
 
+  // Log API errors at INFO level. Aligns with emsg_multiline().
+  ILOG("API error: %s", err->msg);
+
   err->type = errType;
 }
 
--- a/src/nvim/api/ui.c
+++ b/src/nvim/api/ui.c
@@ -225,6 +225,14 @@ void ui_attach(uint64_t channel_id, Integer width, Integer height, Boolean enabl
   nvim_ui_attach(channel_id, width, height, opts, err);
 }
 
+void ui_request_detach(uint64_t chan, String msg)
+{
+  RemoteUI *ui = pmap_get(uint64_t)(&connected_uis, chan);
+  if (ui) {
+    remote_ui_detach(ui, msg);
+  }
+}
+
 /// Tells the nvim server if focus was gained or lost by the GUI
 void nvim_ui_set_focus(uint64_t channel_id, Boolean gained, Error *error)
   FUNC_API_SINCE(11) FUNC_API_REMOTE_ONLY
--- a/src/nvim/api/ui_events.in.h
+++ b/src/nvim/api/ui_events.in.h
@@ -46,6 +46,8 @@ void chdir(String path)
 // Stop event is not exported as such, represented by EOF in the msgpack stream.
 void stop(void)
   FUNC_API_NOEXPORT;
+void detach(String msg)
+  FUNC_API_SINCE(13);
 void ui_send(String content)
   FUNC_API_SINCE(14) FUNC_API_REMOTE_IMPL;
 
--- a/src/nvim/channel.c
+++ b/src/nvim/channel.c
@@ -322,6 +322,7 @@ static void channel_destroy_early(Channel *chan)
     abort();
   }
   pmap_del(uint64_t)(&channels, chan->id, NULL);
+  // XXX: ??? this is used in `free_channel_event`, how can this work?!
   chan->id = 0;
 
   if ((--chan->refcount != 0)) {
--- a/src/nvim/ex_docmd.c
+++ b/src/nvim/ex_docmd.c
@@ -5971,9 +5971,19 @@ static void ex_tabs(exarg_T *eap)
 /// ":detach!" with bang (!) detaches all UIs _except_ the current UI.
 static void ex_detach(exarg_T *eap)
 {
+  String msg;
+  size_t n;
+  char **addrs = server_address_list(&n);
+  if (n > 0) {
+    msg.size = (size_t)snprintf(IObuff, IOSIZE, _("Detached from Nvim server: %s"), addrs[0]);
+    msg.data = IObuff;
+  } else {
+    msg = STATIC_CSTR_AS_STRING("Detached from Nvim server.");
+  }
+
   // come on pooky let's burn this mf down
   if (eap && eap->forceit) {
-    emsg("bang (!) not supported yet");
+    ui_call_detach(msg);
   } else {
     // 1. Send "error_exit" UI-event (notification only).
     // 2. Perform server-side UI detach.
@@ -5993,6 +6003,9 @@ static void ex_detach(exarg_T *eap)
     Error detach_err = ERROR_INIT;
     nvim__chan_set_detach(chan->id, true, &detach_err);
     api_clear_error(&detach_err);
+
+    // TODO(justinmk)
+    // ui_request_detach(chan->id, STATIC_CSTR_AS_STRING("Detached from Nvim server."));
 
     // Server-side UI detach. Doesn't close the channel.
     Error err2 = ERROR_INIT;
--- a/src/nvim/msgpack_rpc/channel.c
+++ b/src/nvim/msgpack_rpc/channel.c
@@ -363,6 +363,7 @@ static void request_event(void **argv)
     goto free_ret;
   }
 
+  current_channel_id = channel->id;
   Object result = handler.fn(channel->id, e->args, &e->used_mem, &error);
   if (e->type == kMessageTypeRequest || ERROR_SET(&error)) {
     // Send the response.
--- a/src/nvim/normal.c
+++ b/src/nvim/normal.c
@@ -14,13 +14,16 @@
 #include <string.h>
 #include <time.h>
 
+#include "nvim/api/private/defs.h"
 #include "nvim/api/private/helpers.h"
+#include "nvim/api/ui.h"
 #include "nvim/ascii_defs.h"
 #include "nvim/autocmd.h"
 #include "nvim/autocmd_defs.h"
 #include "nvim/buffer.h"
 #include "nvim/buffer_defs.h"
 #include "nvim/change.h"
+#include "nvim/channel.h"
 #include "nvim/charset.h"
 #include "nvim/cmdhist.h"
 #include "nvim/cursor.h"
@@ -61,11 +64,13 @@
 #include "nvim/message.h"
 #include "nvim/mouse.h"
 #include "nvim/move.h"
+#include "nvim/msgpack_rpc/channel.h"
 #include "nvim/normal.h"
 #include "nvim/ops.h"
 #include "nvim/option.h"
 #include "nvim/option_vars.h"
 #include "nvim/os/input.h"
+#include "nvim/os/os.h"
 #include "nvim/os/time.h"
 #include "nvim/plines.h"
 #include "nvim/profile.h"
@@ -5160,14 +5165,22 @@ static void nv_window(cmdarg_T *cap)
   }
 }
 
-/// CTRL-Z: Suspend
+/// CTRL-Z: suspend UI, or detach if `count` is given.
 static void nv_suspend(cmdarg_T *cap)
 {
   clearop(cap->oap);
   if (VIsual_active) {
     end_visual_mode();                  // stop Visual mode
   }
-  do_cmdline_cmd("stop");
+
+  if (cap->count0 == 0) {
+    // Regular suspend.
+    // TODO(justinmk): target this to the current UI.
+    do_cmdline_cmd("suspend");
+    return;
+  }
+
+  do_cmdline_cmd("detach");
 }
 
 /// "gv": Reselect the previous Visual area.  If Visual already active,
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -18,6 +18,7 @@
 #include "nvim/event/defs.h"
 #include "nvim/event/loop.h"
 #include "nvim/event/multiqueue.h"
+#include "nvim/event/proc.h"
 #include "nvim/event/signal.h"
 #include "nvim/event/stream.h"
 #include "nvim/globals.h"
@@ -1691,6 +1692,37 @@ static void tui_suspend_cb(TUIData *tui)
   ui_client_attach(tui->width, tui->height, tui->term, tui->rgb);
 }
 #endif
+
+// static void detach_event(void **argv)
+//   FUNC_ATTR_NORETURN
+// {
+//   char *data = argv[0];
+//   ui_client_stop();
+//   if (data != NULL) {
+//     os_errmsg(data);
+//     os_errmsg("\n");
+//   }
+//   xfree(data);
+//   os_exit(0);
+// }
+
+void tui_detach(TUIData *tui, String msg)
+{
+  DLOG("UI 'detach' event");
+
+  // tui_stop(tui);
+  if (msg.data) {
+    // fprintf(stdout, "%s\n", msg.data);
+    fprintf(stderr, "%s\n", msg.data);
+  }
+
+  exit_on_closed_chan(0);
+  // multiqueue_put_event(resize_events, event_create(detach_event, 1, msg.data ? strdup(msg.data) : NULL));
+  // ui_client_detach();
+  // ui_client_stop();
+  // kill(0, SIGTERM);
+  // os_exit(0);
+}
 
 void tui_set_title(TUIData *tui, String title)
 {
--- a/test/client/session.lua
+++ b/test/client/session.lua
@@ -14,6 +14,7 @@ local RpcStream = require('test.client.rpc_stream')
 --- @field private _timer uv.uv_timer_t
 --- @field private _is_running boolean true during `Session:run()` scope.
 --- @field data table Arbitrary user data.
+--- @field closed boolean
 local Session = {}
 Session.__index = Session
 if package.loaded['jit'] then
@@ -244,6 +245,7 @@ function Session:_run(request_cb, notification_cb, timeout)
   self._prepare:stop()
   self._timer:stop()
   self._rpc_stream:read_stop()
+  -- self:close()
 end
 
 --- Nvim msgpack-RPC session.
--- a/test/functional/ui/embed_spec.lua
+++ b/test/functional/ui/embed_spec.lua
@@ -356,3 +356,40 @@ describe('--embed --listen UI', function()
     eq({ 'VimEnter', ('UIEnter:%d'):format(api_info[1]) }, var)
   end)
 end)
+
+describe('UI :detach', function()
+  after_each(function()
+    check_close()
+    -- os.remove(testlog)
+  end)
+
+  it('xxx1', function()
+    -- io_extra = pipe.read,
+    clear { args_rm = { '--headless' }, env = { NVIM_LOG_FILE = 'testlog' } }
+
+    -- attach immediately after startup, for early UI
+    -- rpc_async: Avoid hanging. #24888
+    -- local pipe = assert(uv.pipe())
+    -- local screen = Screen.new(40, 8, { stdin_fd = 3 }, false)
+    -- screen.rpc_async = true -- Avoid hanging. #24888
+    -- screen:attach()
+
+    -- n.expect_exit(1, n.command, 'detach')
+    -- eq('EOF was received from Nvim. Likely the Nvim process crashed.',
+    --   t.pcall_err(n.command, 'detach'))
+    local addr = n.eval('v:servername')
+    -- local s = n.get_session()
+
+    -- TODO(justinmk): using request(), because expect_exit() hangs for some reason.
+    -- local _, rv = s:request('nvim_command', 'detach')
+    n.expect_exit(n.command, 'detach')
+    -- XXX: :detach closes the stream, it didn't "crash", .
+    -- eq('EOF was received from Nvim. Likely the Nvim process crashed.', rv and rv[2] or '?')
+    -- ok(s.closed)
+
+    local s2 = n.connect(addr)
+    n.set_session(s2)
+    n.assert_alive()
+    n.expect_exit(1, n.command, 'qall!')
+  end)
+end)
--- a/test/functional/ui/screen.lua
+++ b/test/functional/ui/screen.lua
@@ -312,6 +312,8 @@ function Screen:attach(session)
   if self._default_attr_ids == nil then
     self._default_attr_ids = Screen._global_default_attr_ids
   end
+
+  session:request('nvim_set_client_info', 'screen.lua', '1.0.0', 'ui', {}, {})
 end
 
 function Screen:detach()


LunarLambda:
https://github.com/neovim/neovim/discussions/30314#discussioncomment-10593190

Doesn't work. Ctrl-Z suspended both UIs running in two separate terminals.

NVIM v0.11.0-dev-738+g2aa64df0d

justinmk:
it works for me... 

> Ctrl-Z suspended both UIs

- It shouldn't "suspend", it should close the channel and terminate the process. In `htop` or `ps` you should see the process die. 
- Are you definitely running `./build/bin/nvim` in all the terminals (including the server)?
- Check `$NVIM_LOG_FILE`, you should see `xxx DETACHED` when you hit `ctrl-z`.
    - Check `getpid()` and look for that PID in the logs.

what commands did you use? I did:

- terminal A (server): `NVIM_LOG_FILE=n.log ./build/bin/nvim --listen ./hello.sock`
    - works with and without `--headless`
- terminal B: `NVIM_LOG_FILE=n.log ./build/bin/nvim --remote-ui --server ./hello.sock`
- terminal C: `NVIM_LOG_FILE=n.log ./build/bin/nvim --remote-ui --server  ./hello.sock`
- `ctrl-z` in B terminates (only) nvim B.
- `ctrl-z` in C terminates (only) nvim C.

LunarLambda:
Followed those exact commands and got the same result. Both UIs suspend, not detach

grepping the log file for `xxx` shows:

```
1:ERR 2024-09-09T20:03:35.673 ?.17466    remote_request:969: xxx main:remote_request pid=17466 ui_channel=3
2:ERR 2024-09-09T20:03:35.676 ?.17466    ui_client_attach:111: xxx pid=17466 ui_channel=3
3:ERR 2024-09-09T20:03:57.025 ?.17486    remote_request:969: xxx main:remote_request pid=17486 ui_channel=3
4:ERR 2024-09-09T20:03:57.027 ?.17486    ui_client_attach:111: xxx pid=17486 ui_channel=3
```
[n.log](https://github.com/user-attachments/files/16934769/n.log)

My terminal is wezterm and my shell is zsh, should I try it with a different terminal? I think it might be that the terminal/shell is intercepting the Ctrl-Z and sending SIGSTOP to neovim immediately, instead of Neovim getting the Ctrl-Z and self-suspending. Unsure.

justinmk:
could be a bug in the `tui/input.c` logic, it's very hacky currently. from your logs it looks like `handle_ctrl_z` wasn't entered. Maybe also try `nvim --clean` in case a plugin is getting in the way.

LunarLambda:
Same result with `-u NONE`, same log output.

LunarLambda:
It does work correctly in `xfce4-terminal`, so it seems like it might be a `wezterm` bug. With this it works.

justinmk:
very likely a bug in `handle_ctrl_z` . it's just a sloppy hack ATM.

i have an idea for a more robust approach

cosminadrianpopescu:
Same findings for me also. It works for terminator and terminology terminals, it fails for xterm. For xterm, `<C-z>` still suspends the client instead of disconnecting it.

cosminadrianpopescu:
Even calling `:call feedkeys("\<C-z>", 'n')` or `vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes('<C-z>', true, false, true), 'n', false)` will still suspend the client instead of disconnecting it.

zeertzjq:
> Even calling `:call feedkeys("\<C-z>", 'n')` or `vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes('<C-z>', true, false, true), 'n', false)` will still suspend the client instead of disconnecting it.

That's correct, as that's executed on the server.

LunarLambda:
> Same findings for me also. It works for terminator and terminology terminals, it fails for xterm. For xterm, `<C-z>` still suspends the client instead of disconnecting it.

I suspect the difference might come down to whether the terminal emulator decides to handle Ctrl-Z by intercepting it and sending SIGSTP to the foreground process itself, or whether it writes the Ctrl-Z to the pseudoterminal and letting the kernel decide whether to translate it into a signal or not (see `termios(3)`, flags `ISIG` as well as the `SUSP` character setting)

I'll see if I can test this somehow.

justinmk:
I have a fix incoming which will avoid the terminal quirks

justinmk:
Latest push should work better

cosminadrianpopescu:
For me it's also working in xterm

src/nvim/os/input.c
```diff
@@ -493,7 +495,7 @@ static TriState inbuf_poll(int ms, MultiQueue *events)
     blocking = true;
     multiqueue_process_events(ch_before_blocking_events);
   }
-  DLOG("blocking... events=%d pending=%d", !!events, pending_events(events));
+  DLOG("blocking... events_pending=%s", pending_events(events) ? "true" : "false");
```
zeertzjq, line 498: Why is `!!events` no longer logged? This information is quite useful.

justinmk:
> support the normal case where `nvim` spawns the server as a child. [use detach=true](https://github.com/neovim/neovim/issues/23093#issuecomment-2337867984)?

I haven't been able to get the "stdio" client to detach without also bringing down the server. Here's what I tried:

```diff
diff --git a/src/nvim/main.c b/src/nvim/main.c
index 421277a396f1..22c54d42b762 100644
--- a/src/nvim/main.c
+++ b/src/nvim/main.c
@@ -357,6 +357,19 @@ int main(int argc, char **argv)
   assert(!ui_client_channel_id && !use_builtin_ui);
   // Nvim server...
 
+  // Fails (and not necessary anyway, because the server was started with detach=true so libuv
+  // already calls setsid().
+  int setsidresult = setsid();
+  if (setsidresult != 0) {
+    ELOG("xxx setsid failed: %s", os_strerror(setsidresult));
+  }
+  // Fails (see above)
+  int setpgidresult = setpgid(0, 0);
+  if (setpgidresult != 0) {
+    ELOG("xxx setpgid failed: %s", os_strerror(setpgidresult));
+  }
+
+
   if (!server_init(params.listen_addr)) {
     mainerr(IObuff, NULL, NULL);
   }
@@ -682,12 +695,12 @@ void os_exit(int r)
     ml_close_all(true);           // remove all memfiles
   }
 
-  if (!event_teardown() && r == 0) {
-    r = 1;  // Exit with error if main_loop did not teardown gracefully.
-  }
-  if (used_stdin) {
-    stream_set_blocking(STDIN_FILENO, true);  // normalize stream (#2598)
-  }
+  // if (!event_teardown() && r == 0) {
+  //   r = 1;  // Exit with error if main_loop did not teardown gracefully.
+  // }
+  // if (used_stdin) {
+  //   stream_set_blocking(STDIN_FILENO, true);  // normalize stream (#2598)
+  // }
 
   ILOG("Nvim exit: %d", r);
 
diff --git a/src/nvim/os/signal.c b/src/nvim/os/signal.c
index ecedf144e5fe..6428bbc78cd0 100644
--- a/src/nvim/os/signal.c
+++ b/src/nvim/os/signal.c
@@ -194,6 +194,7 @@ static void on_signal(SignalWatcher *handle, int signum, void *data)
 #endif
 #ifdef SIGPIPE
   case SIGPIPE:
+    ELOG("xxx signal: SIGPIPE");
     // Ignore
     break;
 #endif
@@ -202,6 +203,7 @@ static void on_signal(SignalWatcher *handle, int signum, void *data)
   case SIGQUIT:
 #endif
   case SIGHUP:
+    ELOG("xxx signal: SIGHUP");
     if (!rejecting_deadly) {
       deadly_signal(signum);
     }
diff --git a/src/nvim/ui_client.c b/src/nvim/ui_client.c
index 77c49a1652d8..89d3efbb707e 100644
--- a/src/nvim/ui_client.c
+++ b/src/nvim/ui_client.c
@@ -60,9 +60,10 @@ uint64_t ui_client_start_server(int argc, char **argv)
   CallbackReader on_err = CALLBACK_READER_INIT;
   on_err.fwd_err = true;
 
+  // spawn server with detach=true
   Channel *channel = channel_job_start(args, get_vim_var_str(VV_PROGPATH),
                                        CALLBACK_READER_INIT, on_err, CALLBACK_NONE,
-                                       false, true, true, false, kChannelStdinPipe,
+                                       false, true, true, true, kChannelStdinPipe,
                                        NULL, 0, 0, NULL, &exit_status);
   if (!channel) {
     return 0;
```

@gpanders @bfredl any ideas?

LunarLambda:
Yeah, I had asked about how that would work conceptually in https://github.com/neovim/neovim/discussions/30314#discussioncomment-10598499

I'm not sure if it can work. I think stdio client should just suspend, I'm not sure detach makes sense if there's no `--listen` address.

cosminadrianpopescu:
I tried `:call feedkeys("\<C-z>", "n")` in xterm and this works as expected. it will disconnect the client without killing the server. To me it looks like a good workaround for the moment. That can be wrapped in a lua function or a vim command.

justinmk:
> I'm not sure detach makes sense if there's no `--listen` address.

Nvim always starts a server (found in `v:servername`) even without `--listen`. 

The "stdio" case won't block this PR but I would like to have it if possible.

zeertzjq:
That's been the behavior since v0.8.

zeertzjq:
> I tried `:call feedkeys("\<C-z>", "n")` in xterm and this works as expected. it will disconnect the client without killing the server. To me it looks like a good workaround for the moment. That can be wrapped in a lua function or a vim command.

That been the behavior since v0.8.

LunarLambda:
> Nvim always starts a server (found in `v:servername`) even without `--listen`.

Ah, interesting. Will there be changes made to the default server address format (currently contains the PID of the server, makes it unpredictable) to make it easier to reattach or will the user always have to figure out the correct `--server` address?

Currently in `nmux` I just compute a predictable default server address in a manner reminiscent of tmux. (Though tmux has the benefit of supporting multiple sessions per server, which Neovim currently does not)

LunarLambda:
> > to make it easier to reattach
> 
>     * can document how to map ctrl-z so that it also puts v:serername in the clipboard.
> 
>     * later, will add something like `nvim --foo` which gives a menu of the most recent addresses to connect to.

Sounds great! Maybe --remote-cmd could even allow getting the output of `:ls` of each server too. I think that might already be possible?

justinmk:
> Maybe --remote-cmd could even allow

the plan for `--remote` is here: https://github.com/neovim/neovim/pull/18414

justinmk:
> > I tried `:call feedkeys("\<C-z>", "n")` in xterm and this works as expected. it will disconnect the client without killing the server.
> 
> That's been the behavior since v0.8.

On master, the current behavior is:

- ctrl-z will suspend all clients
- after that, you can still connect to the server with a new UI
- you can `fg` (foreground) any of the suspended UIs
- the client processes don't exit after ctrl-z, you need to terminate them manually.

src/nvim/ui_client.c
```diff
@@ -61,12 +61,17 @@ uint64_t ui_client_start_server(int argc, char **argv)
 
   Channel *channel = channel_job_start(args, get_vim_var_str(VV_PROGPATH),
                                        CALLBACK_READER_INIT, on_err, CALLBACK_NONE,
-                                       false, true, true, false, kChannelStdinPipe,
+                                       false, true, true, true, kChannelStdinPipe,
```
justinmk, line 64: This works just fine. Libuv is valuable here--saves a lot of headache trying to setup a "double fork" correctly.

src/nvim/ui_client.c
```diff
@@ -81,20 +86,30 @@ uint64_t ui_client_start_server(int argc, char **argv)
   return channel->id;
 }
 
+static void on_stdio_close(Stream *stream, void *data)
+  FUNC_ATTR_NORETURN
+{
+  // Channel *chan = data;
+  ELOG("xxxxxxxxxxxxxxxxxxxxxxxxxxx");
+  os_exit(0);
```
justinmk, line 94: The parent (`--remote-ui` client) exits itself when the stdio channel is closed.

justinmk:
After latest push, this now also works with the default "stdio" client:

```
./build/bin/nvim
<c-z>
[server stays running]
```

Currently, this means that if user closes their terminal, the nvim server will continue as a daemon. That's a "feature", if the terminal was accidentally closed... OTOH, swapfiles already address that. I assume we'll need a way to either opt-in or opt-out.

zeertzjq:
> After latest push, this now also works with the default "stdio" client:
> 
> ```
> ./build/bin/nvim
> <c-z>
> [server stays running]
> ```
> 
> Currently, this means that if user closes their terminal, the nvim server will continue as a daemon. That's a "feature", if the terminal was accidentally closed... OTOH, swapfiles already address that. I assume we'll need a way to either opt-in or opt-out.

That may make it hard to kill Nvim in case of a hang: one now needs to find the server process and send a signal to it.

zeertzjq:
> After latest push, this now also works with the default "stdio" client:
> 
> ```
> ./build/bin/nvim
> <c-z>
> [server stays running]
> ```
> 
> Currently, this means that if user closes their terminal, the nvim server will continue as a daemon. That's a "feature", if the terminal was accidentally closed... OTOH, swapfiles already address that. I assume we'll need a way to either opt-in or opt-out.

This seems to be a breaking change for GUIs.

justinmk:
> This seems to be a breaking change for GUIs. #22137 was documented to `news.txt`, so this should be too.

Thanks for mentioning that. The final form of this PR is open to feedback, we could make this behavior opt-in, possibly dependent on the signal kind (HUP vs TERM).

Also the "daemon mode" for a stdio embedder could be special-cased for the TUI? So GUIs would not need to change anything.

BTW, isn't https://github.com/neovim/neovim/pull/22137 only relevant for stdio embedders? What is the common practice for GUIs that use sockets?

zeertzjq:
Most GUIs I know use stdio embedding when they launch Nvim themselves. When using sockets they only connect to a Nvim instance that is already running.

Freed-Wu:
Will it close #5035?

LunarLambda:
It's part of the todo list / future work listed in the issue, so it seems to be planned for.

justinmk:
```
2025-02-05T15:56:43.9141905Z = ==13522==ERROR: AddressSanitizer: heap-use-after-free on address 0x50c000000e28 at pc 0x55a66d763285 bp 0x7fff1d09d890 sp 0x7fff1d09d888
2025-02-05T15:56:43.9142821Z = READ of size 8 at 0x50c000000e28 thread T0
2025-02-05T15:56:43.9143541Z =     #0 0x55a66d763284 in QUEUE_REMOVE /home/runner/work/neovim/neovim/src/nvim/lib/queue_defs.h:94:22
2025-02-05T15:56:43.9144358Z =     #1 0x55a66d763284 in multiqueue_free /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:115:3
2025-02-05T15:56:43.9144960Z =     #2 0x55a66d360899 in channel_destroy /home/runner/work/neovim/neovim/src/nvim/channel.c:300:3
2025-02-05T15:56:43.9145539Z =     #3 0x55a66d3602d4 in channel_free_all_mem /home/runner/work/neovim/neovim/src/nvim/channel.c:76:3
2025-02-05T15:56:43.9146296Z =     #4 0x55a66dcb78ee in free_all_mem /home/runner/work/neovim/neovim/src/nvim/memory.c:865:3
2025-02-05T15:56:43.9146822Z =     #5 0x55a66db1d429 in os_exit /home/runner/work/neovim/neovim/src/nvim/main.c:687:3
2025-02-05T15:56:43.9147333Z =     #6 0x55a66d77019d in exit_event /home/runner/work/neovim/neovim/src/nvim/event/proc.c:426:7
2025-02-05T15:56:43.9148070Z =     #7 0x55a66d765f76 in multiqueue_process_events /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:149:7
2025-02-05T15:56:43.9148706Z =     #8 0x55a66d760cb6 in loop_poll_events /home/runner/work/neovim/neovim/src/nvim/event/loop.c:83:3
2025-02-05T15:56:43.9149479Z =     #9 0x55a66e4bda1d in ui_client_run /home/runner/work/neovim/neovim/src/nvim/ui_client.c:177:5
2025-02-05T15:56:43.9149991Z =     #10 0x55a66db0d5fa in main /home/runner/work/neovim/neovim/src/nvim/main.c:344:5
2025-02-05T15:56:43.9150481Z =     #11 0x7f83bf82a1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
2025-02-05T15:56:43.9150946Z =     #12 0x7f83bf82a28a in __libc_start_main csu/../csu/libc-start.c:360:3
2025-02-05T15:56:43.9151589Z =     #13 0x55a66cfdbf64 in _start (/home/runner/work/neovim/neovim/build/bin/nvim+0xc4bf64) (BuildId: 06884038f9951902fbe64cc8d0ddb6b7ca5050d2)
2025-02-05T15:56:43.9152184Z = 
2025-02-05T15:56:43.9152500Z = 0x50c000000e28 is located 104 bytes inside of 120-byte region [0x50c000000dc0,0x50c000000e38)
2025-02-05T15:56:43.9152889Z = freed by thread T0 here:
2025-02-05T15:56:43.9153414Z =     #0 0x55a66d07bafa in free (/home/runner/work/neovim/neovim/build/bin/nvim+0xcebafa) (BuildId: 06884038f9951902fbe64cc8d0ddb6b7ca5050d2)
2025-02-05T15:56:43.9154096Z =     #1 0x55a66dcb3fc0 in xfree /home/runner/work/neovim/neovim/src/nvim/memory.c:144:3
2025-02-05T15:56:43.9154660Z =     #2 0x55a66d7639fe in multiqueue_free /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:115:3
2025-02-05T15:56:43.9155247Z =     #3 0x55a66d7620d9 in loop_close /home/runner/work/neovim/neovim/src/nvim/event/loop.c:189:3
2025-02-05T15:56:43.9155782Z =     #4 0x55a66db0c272 in event_teardown /home/runner/work/neovim/neovim/src/nvim/main.c:185:10
2025-02-05T15:56:43.9156468Z =     #5 0x55a66db1d3c8 in os_exit /home/runner/work/neovim/neovim/src/nvim/main.c:674:8
2025-02-05T15:56:43.9156972Z =     #6 0x55a66d77019d in exit_event /home/runner/work/neovim/neovim/src/nvim/event/proc.c:426:7
2025-02-05T15:56:43.9157584Z =     #7 0x55a66d765f76 in multiqueue_process_events /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:149:7
2025-02-05T15:56:43.9158217Z =     #8 0x55a66d760cb6 in loop_poll_events /home/runner/work/neovim/neovim/src/nvim/event/loop.c:83:3
2025-02-05T15:56:43.9158917Z =     #9 0x55a66e4bda1d in ui_client_run /home/runner/work/neovim/neovim/src/nvim/ui_client.c:177:5
2025-02-05T15:56:43.9159426Z =     #10 0x55a66db0d5fa in main /home/runner/work/neovim/neovim/src/nvim/main.c:344:5
2025-02-05T15:56:43.9159912Z =     #11 0x7f83bf82a1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
2025-02-05T15:56:43.9160373Z =     #12 0x7f83bf82a28a in __libc_start_main csu/../csu/libc-start.c:360:3
2025-02-05T15:56:43.9161008Z =     #13 0x55a66cfdbf64 in _start (/home/runner/work/neovim/neovim/build/bin/nvim+0xc4bf64) (BuildId: 06884038f9951902fbe64cc8d0ddb6b7ca5050d2)
2025-02-05T15:56:43.9161545Z = 
2025-02-05T15:56:43.9161735Z = previously allocated by thread T0 here:
2025-02-05T15:56:43.9162306Z =     #0 0x55a66d07bd93 in malloc (/home/runner/work/neovim/neovim/build/bin/nvim+0xcebd93) (BuildId: 06884038f9951902fbe64cc8d0ddb6b7ca5050d2)
2025-02-05T15:56:43.9162999Z =     #1 0x55a66dcb3d04 in try_malloc /home/runner/work/neovim/neovim/src/nvim/memory.c:98:15
2025-02-05T15:56:43.9163528Z =     #2 0x55a66dcb3ef4 in xmalloc /home/runner/work/neovim/neovim/src/nvim/memory.c:132:15
2025-02-05T15:56:43.9164103Z =     #3 0x55a66d7652b1 in multiqueue_push /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:234:35
2025-02-05T15:56:43.9164749Z =     #4 0x55a66d764503 in multiqueue_put_event /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:137:3
2025-02-05T15:56:43.9165337Z =     #5 0x55a66d76c9b0 in decref /home/runner/work/neovim/neovim/src/nvim/event/proc.c:316:3
2025-02-05T15:56:43.9165877Z =     #6 0x55a66d75f6e4 in close_cb /home/runner/work/neovim/neovim/src/nvim/event/libuv_proc.c:114:5
2025-02-05T15:56:43.9166785Z =     #7 0x55a66e8a144f in uv__finish_close /home/runner/work/neovim/neovim/.deps/build/src/libuv/src/unix/core.c:363:5
2025-02-05T15:56:43.9167475Z =     #8 0x55a66e89eac5 in uv__run_closing_handles /home/runner/work/neovim/neovim/.deps/build/src/libuv/src/unix/core.c:377:5
2025-02-05T15:56:43.9168298Z =     #9 0x55a66e89e969 in uv_run /home/runner/work/neovim/neovim/.deps/build/src/libuv/src/unix/core.c:475:5
2025-02-05T15:56:43.9168873Z =     #10 0x55a66d760fbd in loop_uv_run /home/runner/work/neovim/neovim/src/nvim/event/loop.c:61:3
2025-02-05T15:56:43.9169432Z =     #11 0x55a66d760c24 in loop_poll_events /home/runner/work/neovim/neovim/src/nvim/event/loop.c:82:26
2025-02-05T15:56:43.9169982Z =     #12 0x55a66db0c225 in event_teardown /home/runner/work/neovim/neovim/src/nvim/main.c:176:3
2025-02-05T15:56:43.9170484Z =     #13 0x55a66db1d3c8 in os_exit /home/runner/work/neovim/neovim/src/nvim/main.c:674:8
2025-02-05T15:56:43.9170989Z =     #14 0x55a66d77019d in exit_event /home/runner/work/neovim/neovim/src/nvim/event/proc.c:426:7
2025-02-05T15:56:43.9171600Z =     #15 0x55a66d765f76 in multiqueue_process_events /home/runner/work/neovim/neovim/src/nvim/event/multiqueue.c:149:7
2025-02-05T15:56:43.9172230Z =     #16 0x55a66d760cb6 in loop_poll_events /home/runner/work/neovim/neovim/src/nvim/event/loop.c:83:3
2025-02-05T15:56:43.9172799Z =     #17 0x55a66e4bda1d in ui_client_run /home/runner/work/neovim/neovim/src/nvim/ui_client.c:177:5
2025-02-05T15:56:43.9173301Z =     #18 0x55a66db0d5fa in main /home/runner/work/neovim/neovim/src/nvim/main.c:344:5
2025-02-05T15:56:43.9173796Z =     #19 0x7f83bf82a1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
2025-02-05T15:56:43.9174261Z =     #20 0x7f83bf82a28a in __libc_start_main csu/../csu/libc-start.c:360:3
2025-02-05T15:56:43.9174899Z =     #21 0x55a66cfdbf64 in _start (/home/runner/work/neovim/neovim/build/bin/nvim+0xc4bf64) (BuildId: 06884038f9951902fbe64cc8d0ddb6b7ca5050d2)
```

mikew:
>  Scripts can ask "what's the current UI?"

Would be huge for me to implement uploads / downloads in https://github.com/mikew/nvrh/issues/5

PR #30323, opened 2024-09-10
feat(fs): improve environment variables expasion

jamestrew:
Adds a more robust environment variable expansion to `vim.fs.normalize`, particularly for Windows. Heavily based on Python's `os.path.expandvars` implementation.

Previously, only `$VAR` style environment variables were expanded.
Now also expands `${VAR}` style environment variables.

On Windows, also expands `%VAR%` form environment variables. Windows environment variables have a larger valid character set than that of posix environments. Variables like `ProgramFiles(x86)` is only expandable through `${VAR}` or `%VAR%` forms.

succeeds https://github.com/neovim/neovim/pull/30292


justinmk:
> Now also expands `${VAR}` style environment variables.
> On Windows, also expands `%VAR%` form environment variables.

That is a shell-specific syntax. Why is that needed in `vim.fs.normalize`? 

The central purpose of `vim.fs.normalize` is convert a file path has a single, unique representation. Any use-case where expansion is needed can call `vim.fn.expand` first.

The "expand" feature of `vim.fs.normalize` is intentionally narrow and I don't actually remember what the motivation was. And the amount of code added in this PR is a strong case against this new scope.

jamestrew:
Yeah I understand it's maybe not a major value add especially for the amount of code added. I could probably "simplify" this down using LPEG if interested.

For what it's worth, if I recall correctly, `vim.fn.expand` only supports the simple `$VAR` expansion as far as environment variables, so I don't think there's anything redundant here that wasn't already there.

If there's no interest, feel free to close.

justinmk:
There are existing environment variable issues that we would like help with: https://github.com/neovim/neovim/issues?q=is%3Aissue+is%3Aopen+environment+variables

PR #30325, opened 2024-09-10
fix(coverity/508289): vterm lookup_colour() dead code

devon-g:
Problem:
Dead code due to CSI arg guard resulting in argcount never zero.

Solution:
Invert guard to avoid repeating return statements.

### DIFF
--- a/src/vterm/pen.c
+++ b/src/vterm/pen.c
@@ -100,11 +100,8 @@ static int lookup_colour(const VTermState *state, int palette, const long args[]
     return 3;
 
   case 5: // XTerm 256-colour mode
-    if (!argcount || CSI_ARG_IS_MISSING(args[0])) {
-      return argcount ? 1 : 0;
-    }
-
-    vterm_color_indexed(col, args[0]);
+    if (argcount && !CSI_ARG_IS_MISSING(args[0]))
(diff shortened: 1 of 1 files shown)


src/vterm/pen.c
```diff
@@ -100,11 +100,8 @@ static int lookup_colour(const VTermState *state, int palette, const long args[]
     return 3;
 
   case 5: // XTerm 256-colour mode
-    if (!argcount || CSI_ARG_IS_MISSING(args[0])) {
-      return argcount ? 1 : 0;
-    }
-
-    vterm_color_indexed(col, args[0]);
+    if (argcount && !CSI_ARG_IS_MISSING(args[0]))
+      vterm_color_indexed(col, args[0]);
```
justinmk, line 104: if we're going to update vterm code, it should follow our standards.  braces after `if` are required by our style guide.

devon-g: I had seen other if statements without them. I updated all of them to have braces along with the change I had made.

clason:
Note that this conflicts with https://github.com/neovim/neovim/pull/30083, which we would want to merge first. Afterwards, such fixes would be very welcome.

PR #30326, opened 2024-09-10
feat(packadd/apply_autocmds): add auevents for :packadd

vpyr:
this should be useful to configure packs in a reactive fashion

### DIFF
--- a/src/nvim/auevents.lua
+++ b/src/nvim/auevents.lua
@@ -79,6 +79,8 @@ return {
     'LspProgress', -- after a LSP progress update
     'MenuPopup', -- just before popup menu is displayed
     'ModeChanged', -- after changing the mode
+    'PackAddPre', -- before loading pack form :packpath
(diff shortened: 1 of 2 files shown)


justinmk:
Review https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md#pull-requests-prs . Especially the part about writing a clear "Problem" description, to explain the motivation.

In any case, this particular change should go to https://github.com/vim/vim , assuming it's actually needed. Because we haven't diverged from vim's package model and this isn't a strong reason to do so.

liskin:
@fjunger I just came looking for this, but it seems like you've got discouraged from pursuing this? I think it'd be really useful to have this — would you consider opening this PR against https://github.com/vim/vim as Justin suggested? I'm sure they'd cherry-pick the feature into neovim in no time if it's accepted into vim, as it's super simple yet very useful.

PR #30327, opened 2024-09-10
refactor(os/input.c): rename os_inchar => input_get

justinmk:
# Problem

The name `os_inchar` (from Vim's old `mch_inchar`) is ambiguous: "inchar" sounds like it could be reading or enqueuing (setting) input. Its docstring is also ambiguous.

# Solution

- Rename `os_inchar` to `input_get`.
- Write some mf'ing docstrings.

cc @zeertzjq @bfredl

### DIFF
--- a/src/nvim/os/input.c
+++ b/src/nvim/os/input.c
@@ -33,12 +33,6 @@
 #define READ_BUFFER_SIZE 0xfff
 #define INPUT_BUFFER_SIZE ((READ_BUFFER_SIZE * 4) + MAX_KEY_CODE_LEN)
 
-typedef enum {
-  kInputNone,
-  kInputAvail,
-  kInputEof,
-} InbufPollResult;
-
 static RStream read_stream = { .s.closed = true };  // Input before UI starts.
 static char input_buffer[INPUT_BUFFER_SIZE];
 static char *input_read_pos = input_buffer;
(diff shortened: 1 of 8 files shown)


src/nvim/os/input.c
```diff
@@ -533,8 +551,7 @@ static void process_ctrl_c(void)
   }
 }
 
-// Helper function used to push bytes from the 'event' key sequence partially
-// between calls to os_inchar when maxlen < 3
+// Pushes bytes from the 'event' key sequence partially between calls to input_get when maxlen < 3.
```
zeertzjq, line 554:
```suggestion
/// Pushes bytes from the 'event' key sequence partially between calls to input_get when maxlen < 3.
```

src/nvim/os/input.c
```diff
@@ -115,13 +115,21 @@ int input_get(uint8_t *buf, int maxlen, int ms, int tb_change_cnt, MultiQueue *e
...
   }
 
-  if (maxlen && input_available()) {
-    restart_cursorhold_wait(tb_change_cnt);
-    size_t to_read = MIN((size_t)maxlen, input_available());
-    memcpy(buf, input_read_pos, to_read);
-    input_read_pos += to_read;
-    return (int)to_read;
-  }
+#define TRY_READ() \
```
zeertzjq, line 118: Add an `#undef` after the last use of this macro?

src/nvim/os/input.c
```diff
@@ -84,57 +78,76 @@ static void cursorhold_event(void **argv)
...
+  do { \
+    if (maxlen && input_available()) { \
+      reset_cursorhold_wait(tb_change_cnt); \
+      assert(maxlen >= 0); \
+      size_t to_read = MIN((size_t)maxlen, input_available()); \
+      memcpy(buf, input_read_pos, to_read); \
+      input_read_pos += to_read; \
+      /* This is safe because INPUT_BUFFER_SIZE fits in an int. */ \
+      assert(to_read <= INT_MAX); \
+      return (int)to_read; \
```
justinmk, line 126: Also added the two `assert()` calls.

PR #30330, opened 2024-09-10
fix(lsp): check buffer is loaded and valid in handler

glepnir:
Problem: buffer mabye not valid when callback handler invoke.

Soliton: check buffer is valid and loaded in handler.

### DIFF
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -43,17 +43,16 @@ function M.on_inlayhint(err, result, ctx, _)
     return
   end
   local bufnr = assert(ctx.bufnr)
-  if util.buf_versions[bufnr] ~= ctx.version then
+  if
+    util.buf_versions[bufnr] ~= ctx.version
+    or not result
+    or not api.nvim_buf_is_loaded(bufnr)
+    or not bufstates[bufnr].enabled
+  then
     return
   end
   local client_id = ctx.client_id
-  if not result then
-    return
-  end
   local bufstate = bufstates[bufnr]
-  if not bufstate.enabled then
-    return
-  end
   if not (bufstate.client_hints and bufstate.version) then
     bufstate.client_hints = vim.defaulttable()
     bufstate.version = ctx.version


runtime/lua/vim/lsp/inlay_hint.lua
```diff
@@ -43,17 +43,11 @@ function M.on_inlayhint(err, result, ctx, _)
     return
   end
   local bufnr = assert(ctx.bufnr)
-  if util.buf_versions[bufnr] ~= ctx.version then
+  if util.buf_versions[bufnr] ~= ctx.version or not result or not bufstates[bufnr].enabled then
```
justinmk, line 46: why not check `nvim_buf_is_loaded` here?

glepnir: bufstates[bufnr] has removed in on_detach  so check before this function invoke . reduce a function call instruction

glepnir: Honestly, it works the same everywhere. this lsp handler just calls this function. your suggest is more uniform. I just wonder if there will be an extra call assembly instruction 🥲

justinmk: I wouldn't worry about performance here, because this condition is unusual. So readability is more important, and keeping the "return" conditions in one place seems very helpful.

justinmk (approved):
is there a related bug report? what steps caused this?

glepnir:
it's part of https://github.com/neovim/neovim/issues/29614 

A simple idea is that handler callbacks are asynchronous. so each one needs a buffer check. Can a simple factory be used to avoid some duplication of code?

```lua
local function handler_factory(fn)
    return function(err, result,ctx,config)
           api.nvim_buf_is_loaded(ctx.bufnr) and fn(err, result, ctx, config) or nil
    end
end
```

justinmk:
> Can a simple factory be used to avoid some duplication of code?

Not a factory, but each handler could call a single `_util.foo()` function that captures common checks for this scenario. 

I believe this has been discussed before? 

@MariaSolOs @mfussenegger @p00f @ofseed any objection to this PR meanwhile?

p00f:
looks good to me!

mfussenegger:
> any objection to this PR meanwhile?

No objection but I'd really love to see some more test coverage in this area as there have been numerous fixes now for inlay hints related to buffer state management and each time without tests - makes me worry we might play ping-pong between different edge cases.

justinmk:
> > Can a simple factory be used to avoid some duplication of code?
> 
> Not a factory, but each handler could call a single `_util.foo()` function that captures common checks for this scenario.
> 
> I believe this has been discussed before? (Note: referring to private/internal utils, whereas #25272 is about public utils)

@glepnir this would be a good thing to look into, and include test coverage as @mfussenegger mentioned. And if this logic is centralized, then we need fewer tests yet we gain better coverage.

PR #30333, opened 2024-09-10
lsp: handle InsertTextMode in completion items

### LINKED ISSUE #30332: [lsp,completion] LSP servers main return  InsertReplaceEdit for textEdit field
### Problem

Native LSP completion is currently failing for `gopls`. 

```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:311: attempt to index field 'range' (a nil value)                                                                                                                                                     
stack traceback:                                                                                                                                                                                                                                                                                    
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:311: in function 'adjust_start_col'                                                                                                                                                                                                      
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:362: in function '_convert_results'                                                                                                                                                                                                      
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:451: in function 'callback'                                                                                                                                                                                                              
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:390: in function 'handler'                                                                                                                                                                                                               
        /usr/share/nvim/runtime/lua/vim/lsp/client.lua:687: in function ''                                                                                                                                                                                                                          
```

Per LSP spec `textEdit` can be one of two values and current code is only handling the case where it is a `TextEdit`. 

```
--- @param lnum integer 0-indexed
--- @param line string
--- @param items lsp.CompletionItem[]
--- @param encoding string
--- @return integer?
local function adjust_start_col(lnum, line, items, encoding)
  local min_start_char = nil
  for _, item in pairs(items) do
    if item.textEdit and item.textEdit.range.start.line == lnum then
      if min_start_char and min_start_char ~= item.textEdit.range.start.character then
        return nil
      end
      min_start_char = item.textEdit.range.start.character
    end
  end
  if min_start_char then
    return lsp.util._str_byteindex_enc(line, min_start_char, encoding)
  else
    return nil
  end
end
```

I have some code which fixes this locally and it no longer breaks `gopls` completion.
However, I do not know how wide spread this assumption is across completion.lua. 

If you'd like to submit my fixes for this I will, but it maybe best to have the author of completion.lua address this as they probably understand how far reaching this conflation is. 

### Steps to reproduce

1. Enable gopls (golang.org/x/tools/gopls v0.16.2) 
2. Enable native lsp completion (NVIM v0.11.0-dev+684-g53af02adb)
3. Attempt to complete some function. 

You will encounter an error since, for example, gopls returns:
```
{
  detail = "interface{...}",
  documentation = {
    kind = "markdown",
    value = "Action represents an action in any supported filter.\n"
  },
  filterText = "Action",
  insertTextFormat = 2,
  kind = 8,
  label = "Action",
  preselect = true,
  sortText = "00000",
  textEdit = {
    insert = {
      ["end"] = {
        character = 9,
        line = 43
      },
      start = {
        character = 9,
        line = 43
      }
    },
    newText = "Action",
    replace = {
      ["end"] = {
        character = 9,
        line = 43
      },
      start = {
        character = 9,
        line = 43
      }
    }
  }
}
```

### Expected behavior

The completion code must expect either a `TextEdit` or a `InsertReplaceEdit` in the completion item's textEdit field and be prepared to handle both. 

### Neovim version (nvim -v)

v0.11.0-dev+684-g53af02adb

no, but we need to enable native lsp to produce issue

### Operating system/version

Fedora 40

### Terminal name/version

kitty

xterm-kitty

### Installation

https://copr.fedorainfracloud.org/coprs/agriffis/neovim-nightly/

ldelossa:
Gopls may return a `InsertReplaceEdit` structure in a LSP completion item depending on settings.

Make completion.lua understand this could occur and extract the appropriate variables.

This fixes a crash in gopls completion:
```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:311: attempt to index field 'range' (a nil value)
stack traceback:
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:311: in function 'adjust_start_col'
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:362: in function '_convert_results'
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:451: in function 'callback'
        /usr/share/nvim/runtime/lua/vim/lsp/completion.lua:390: in function 'handler'
        /usr/share/nvim/runtime/lua/vim/lsp/client.lua:687: in function ''
```

Closes #30332

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -308,11 +308,22 @@ end
 local function adjust_start_col(lnum, line, items, encoding)
   local min_start_char = nil
   for _, item in pairs(items) do
-    if item.textEdit and item.textEdit.range.start.line == lnum then
-      if min_start_char and min_start_char ~= item.textEdit.range.start.character then
+    local start_line = 0
+    local start_char = 0
+
+    if item.textEdit and item.textEdit.newText then
+      start_line = item.textEdit.insert.start.line
+      start_char = item.textEdit.insert.start.chararacter
+    else
+      start_line = item.textEdit.range.start.line
+      start_char = item.textEdit.range.start.character
+    end
+
+    if item.textEdit and start_line == lnum then
+      if min_start_char and min_start_char ~= start_char then
         return nil
       end
-      min_start_char = item.textEdit.range.start.character
+      min_start_char = start_char
     end
   end
   if min_start_char then


mfussenegger (commented):
Neovim doesn't set `insertReplaceSupport` in the [CompletionClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionClientCapabilities) so I'd expect servers to not send a `InsertReplaceEdit`.

This seems like a server bug?

ldelossa:
@mfussenegger - oh, I define my own capabilities, depending on what the LSP offers. I didn't think Neovim was heavy handed about this.

MariaSolOs:
> I guess, at the least, its a bit hard to understand when flicking on an option that an LSP supports will break Neovim.

Well it's not "flicking an option". If the editor announces via capabilities that it lacks support for something, you can't expect it to handle requests correctly.

ldelossa:
@MariaSolOs aha yeah, I see I had the wrong mind set here. 

Moving forward, is there Neovim help file which tell the user what capabilities Neovim supports?

MariaSolOs:
> Moving forward, is there Neovim help file which tell the user what capabilities Neovim supports?

AFAIK there's no help file listing those, and IMO it wouldn't be very helpful anyway since plugins can add extra capabilities (e.g. `nvim-cmp`). But you can always check https://github.com/neovim/neovim/blob/4c5bce9cb417cf1c6a93cdd5ca6d3d52d2bf95d6/runtime/lua/vim/lsp/protocol.lua#L333

PR #30335, opened 2024-09-10
test: port libvterm unit test to neovim test suite

dundargoc:
These were imported from the v0.3.3 git tag
https://github.com/neovim/libvterm/tree/v0.3.3 and not the latest
commit. This is for compatibility reasons as the libvterm code was
imported from v0.3.3.


src/vterm/vterm.h
```diff
@@ -499,18 +499,18 @@ void vterm_state_send_selection(VTermState *state, VTermSelectionMask mask, VTer
 // ------------
 
 typedef struct {
-    unsigned int bold      : 1;
```
bfredl, line 502: removing bitfields are going to expand the size of these structs significantly.

If we need to use these in tests it would be better to write parts of the test logic as C helper functions (there is some predecendent for this, marktree e.g. has `ifndef NDEBUG` functions only for debug builds).

bfredl: or if we don't need them we just need to filter out them so luajit doesn't need to see them and get angy..

dundargoc: I suspect they're not needed since both the vterm and our own tests pass without them, but I am not entirely sure. I did it this way for now to unblock the work.

zeertzjq: Did you not read the first sentence?

dundargoc: I did but I don't know how to filter them out like he suggested.

dundargoc: I did, but I misunderstood. I've opted to filter out all bitfields in unit tests.

test/unit/testutil.lua
```diff
@@ -151,6 +151,16 @@ local function filter_complex_blocks(body)
...
     then
+      -- Remove bitfields from structs as luajit can't seem to handle those well.
+      line = string.gsub(line, 'typedef struct {(.-)}', function(s)
+        return string.format('typedef struct {%s}', s:gsub('%s*:%s*%d+%s*;', ';'))
+      end)
+
+      -- HACK: hardcode to remove difficult bitfield not handled by the previous gsub
+      if line:find('struct VTermState') then
+        line = string.gsub(line, 'state : 8;', 'state;')
+      end
```
dundargoc, line 162: Hacky AF, but it works for now

src/vterm/vterm.c
```diff
@@ -429,3 +429,212 @@ void vterm_check_version(int major, int minor)
...
+    if(b < 0x20 || b == 0x7f || (b >= 0x80 && b < 0xa0)) {
+      break;
+    }
+    fprintf(f, i ? ",%x" : "%x", b);
+  }
+  fprintf(f, "\n");
+  fclose(f);
+
+  return i;
+}
```
dundargoc, line 453: TODO: wonder if I can just write this as a lua function instead.

CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ set(DEPS_IGNORE_SHA FALSE)
 #-------------------------------------------------------------------------------
 set(FUNCS_DATA ${PROJECT_BINARY_DIR}/funcs_data.mpack)
 set(TOUCHES_DIR ${PROJECT_BINARY_DIR}/touches)
+set(VTERM_TEST_FILE ${PROJECT_BINARY_DIR}/test/vterm_test_output)
```
justinmk, line 49: if this is a test artifact, should this go in the build/ dir? that is where we store e.g. tempfiles created by `t.tmpname()`

src/nvim/terminal.c
```diff
@@ -232,7 +232,7 @@ static int parse_osc8(VTermStringFragment frag, int *attr)
 {
   // Parse the URI from the OSC 8 sequence and add the URL to our URL set.
   // Skip the ID, we don't use it (for now)
-  size_t i = 0;
+  int i = 0;
```
justinmk, line 235: 👀 is this temporary change or is it explained somewhere?

CMakeLists.txt
dundargoc: `PROJECT_BINARY_DIR` is the `build/` dir.

src/nvim/terminal.c
dundargoc: I needed to change the type of `flag.len` to `int`, which causes "Comparison of integers of different signs" if I don't change this to `int`. I can just cast it to `int` every time instead if you prefer.

dundargoc: So not temporary. I can keep it like this or revert to `size_t` and cast `flag.len` to `size_t`, both should work equally well. Lemme know if you have a preference.

src/nvim/terminal.c
```diff
@@ -248,21 +248,21 @@ static int parse_osc8(VTermStringFragment frag, int *attr)
...
-  const size_t len = i - start;
+  const int len = i - start;
   if (len == 0) {
     // Empty OSC 8, no URL
     *attr = 0;
     return 1;
   }
 
-  char *url = xmemdupz(&frag.str[start], len + 1);
+  char *url = xmemdupz(&frag.str[start], (size_t)len + 1);
```
justinmk, line 265: might need `assert(len >= 0 && len < MAX_INT32)` for coverity

src/vterm/vterm.c
```diff
@@ -429,3 +430,508 @@ void vterm_check_version(int major, int minor)
 
   // Happy
 }
+
+#ifndef NDEBUG
```
justinmk, line 434:
```suggestion
// For unit tests.
#ifndef NDEBUG
```

justinmk (approved):

test/unit/vterm_spec.lua
```diff
@@ -0,0 +1,3590 @@
...
+    push('Bottom', vt)
+    push('\x1b[1;1H', vt)
+    push('Top', vt)
+
+    -- Output
+    screen_row(23, 'Bottom', screen)
+    screen_row(0, 'Top', screen)
+  end)
+
+  itp('92lp1640917', function()
```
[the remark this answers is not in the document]
dundargoc: That is the name of the test file. Don't ask lol.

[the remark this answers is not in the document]
dundargoc: Shall I change it to something more readable/useful?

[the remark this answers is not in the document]
justinmk: Just checking that it's intentional. If a more meaningful name is possible then great

src/nvim/terminal.c
dundargoc: These pre-emptive coverity warning guessing as well as the post-merged coverity warnings adds a lot of friction. I'll see if I have time to set it up coverity CI check for PRs this week, otherwise we might wanna create a dedicated issue to track this problem.

src/vterm/vterm.h
```diff
@@ -219,7 +220,7 @@ typedef enum {
 
 typedef struct {
   const char *str;
-  size_t      len : 30;
+  int         len : 30;
```
zeertzjq, line 223: Can you mention somewhere why this change is necessary?

dundargoc: The tests fail with `test/unit/testutil.lua:276: invalid value near ':' at line 18'` otherwise.

dundargoc: Reverted it back to size_t by removing the bitfield in tests.

PR #30339, opened 2024-09-10
ci: enable automerge by default when backporting

dundargoc:
This will automatically merge backported PRs without human intervention
if the tests pass.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -26,3 +26,8 @@ jobs:
           pull_title: "${pull_title}"
           label_pattern: "^ci:backport ([^ ]+)$"
           github_token: ${{ steps.app-token.outputs.token }}
+
+      - name: Enable automerge
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


dundargoc:
todo: test the damn thing

clason (approved):

clason:
Followup (brought up in chat): add `needs:backport` label if backport failed (`outputs.was_successful=false`).

(Also check if we can make it fail less, e.g., with https://github.com/korthout/backport-action#merge_commits set to `skip`.)

dundargoc:
> (Also check if we can make it fail less, e.g., with [korthout/backport-action#merge_commits](https://github.com/korthout/backport-action#merge_commits) set to skip.)

I tried it out with and without and I'm pretty sure it doesn't make a difference. I think the merge commit is already skipped since we use `cherry_picking = auto`. I don't mind adding `merge_commits = skip` just to be safe though.

clason:
I think this is specifically about when the action fails, which happens more often than I would like (and understand). I'm not sure if (in combination with our other flags) this will make a difference, but anything else than `fail` simply sounds... less likely to fail ;)

PR #30341, opened 2024-09-11
fix(coverity/509227/509228): tui driver_ti avoid underflow

devon-g:
Problem:
write() can return -1 but is cast to unsigned type causing coverity to detect possible overflowed integer

Solution:
Perform check to ensure all negative values are captured rather than just -1 before casting to unsigned type

### DIFF
--- a/src/nvim/tui/termkey/driver-ti.c
+++ b/src/nvim/tui/termkey/driver-ti.c
@@ -449,10 +449,11 @@ int stop_driver_ti(TermKey *tk, void *info)
   // Can't call putp or tputs because they suck and don't give us fd control
   len = strlen(stop_string);
   while (len) {
-    size_t written = (size_t)write(tk->fd, stop_string, (unsigned)len);
-    if (written == (size_t)-1) {
+    ssize_t result = write(tk->fd, stop_string, (unsigned)len);
+    if (result < 0) {
       return 0;
     }
+    size_t written = (size_t)result;
     stop_string += written;
     len -= written;
   }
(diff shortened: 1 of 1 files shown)


src/nvim/tui/termkey/driver-ti.c
```diff
@@ -410,10 +410,15 @@ int start_driver_ti(TermKey *tk, void *info)
   // Can't call putp or tputs because they suck and don't give us fd control
   len = strlen(start_string);
   while (len) {
-    size_t written = (size_t)write(tk->fd, start_string, (unsigned)len);
-    if (written == (size_t)-1) {
+#ifndef _WIN32
+    ssize_t result = write(tk->fd, start_string, len);
+#else
+    ssize_t result = write(tk->fd, start_string, (unsigned int)len);
```
justinmk, line 416: what was wrong with the `(unsigned)len` cast? was coverity, or non-windows CI, complaining about it?

devon-g: The posix write signature calls for size_t and the windows write signature calls for unsigned int. I was trying to be explicit in this difference since windows ci fails if not cast to unsigned int.

justinmk: if (unsigned) doesn't give coverity/CI warnings I think it would be better to avoid the verbosity of the #ifdefs.

devon-g: fixed it!

PR #30342, opened 2024-09-11
fix(health): :checkhealth completes "test_plug/health/init.lua" as "test_plug.health"

monkoose:
`path2name()` function doesn't process `'pluginname/health/init.lua'` correctly. Instead of returning `'pluginname'` it returns `'pluginname.health'`.


justinmk:
Thanks. I added a test. In the future please explain the actual problem being solved. E.g. in this case, the symptom was "`:checkhealth` completion does X... it should do Y".

PR #30347, opened 2024-09-11
docs(lsp): detail on hover window controls

### LINKED ISSUE #27288: Ability to scroll hover docs floating window without entering it
### Problem

I failed to find a way a native lsp feature to scroll docs. Apparently you have to enter the docs by pressing the same keymap twice, then scroll normally, and then you also have to jump out of the floating window. I don't like it, because it forces me to use additional keymaps when I just want to scroll down the docs once or twice. I saw someone mentioned that the coc.nvim plugin does have this feature. And from what I've read it jumps into the floating window, scrolls and then jump out of it. Although, when I used it I didn't notice anything like that — it just scrolls, and I loved it. At least `<c-f>` and `<c-b>` worked, and this was enough for me (at least from what I can remember). If a user wants to copy something from the docs, then the double keymap press to hop into it can still be used. I also didn't find how to scroll docs when using nvim-cmp, I only know that I can scroll it up and down with Lua API (autocompletion docs in insert mode), but it doesn't work for Neovim's LSP hover docs window.

Related links:
https://github.com/neovim/neovim/issues/15092
https://github.com/hrsh7th/nvim-cmp/issues/1410

### Expected behavior

The feature can be implemented in 2 ways, I suppose:
1. add a scroll function that would do "jump in, scroll, jump out" action
2. add a proper docs scroll for a floating window without entering it

The first one should be much easier. The second probably would require quite a bit of discussion about which windows can be scrolled this way and the API.

I haven't used VS Code in a long time, but I bet you can scroll in the floating windows there, when you hover over something. I think this feature is essential for the DX, and I am constantly reminded that it doesn't exist when I want to use it.

tris203:
closes #27288 

as per Justin's suggestion of adding a map to exit: q is already mapped to exit so document this to clear up confusion

If somebody wants to close the float  (that isn't focused) without moving their cursor, they can either do ```Kq``` or map this in their config to something shorter.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1397,7 +1397,11 @@ format({opts})                                          *vim.lsp.buf.format()*
 
 hover()                                                  *vim.lsp.buf.hover()*
     Displays hover information about the symbol under the cursor in a floating
-    window. Calling the function twice will jump into the floating window.
+    window. The window will be dismissed on cursor move. Calling the function
+    twice will jump into the floating window (thus by default, "KK" will open
+    the hover window and focus it). In the floating window, all commands and
+    mappings are available as usual, except that "q" dismisses the window. You
+    can scroll the contents the same as you would any other buffer.
 
 implementation({opts})                          *vim.lsp.buf.implementation()*
     Lists all the implementations for the symbol under the cursor in the
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -29,7 +29,12 @@ local function request(method, params, handler)
 end
 
 --- Displays hover information about the symbol under the cursor in a floating
---- window. Calling the function twice will jump into the floating window.
+--- window. The window will be dismissed on cursor move.
+--- Calling the function twice will jump into the floating window
+--- (thus by default, "KK" will open the hover window and focus it).
+--- In the floating window, all commands and mappings are available as usual,
+--- except that "q" dismisses the window.
+--- You can scroll the contents the same as you would any other buffer.
 function M.hover()
   local params = util.make_position_params()
   request(ms.textDocument_hover, params)


runtime/doc/lsp.txt
```diff
@@ -1397,7 +1397,10 @@ format({opts})                                          *vim.lsp.buf.format()*
 
 hover()                                                  *vim.lsp.buf.hover()*
     Displays hover information about the symbol under the cursor in a floating
-    window. Calling the function twice will jump into the floating window.
+    window. The window will be dismissed on cursor move. Calling the function
+    twice will jump into the floating window. Once focused in the floating
+    window, `<C-f>` and `<C-b>` can be used to scroll forward and back, and
```
MariaSolOs, line 1402: `<C-d>`/`<C-u>` also work for scrolling.

tris203: > Once focused in the floating window, the content buffer can be scrolled as normal and `q` can be used to dismiss the window.

Does that sound better?

MariaSolOs: I would use "usual" instead of "normal" because of the latter already having a special meaning in vim.

MariaSolOs (approved):

runtime/doc/lsp.txt
```diff
@@ -1397,7 +1397,10 @@ format({opts})                                          *vim.lsp.buf.format()*
 
 hover()                                                  *vim.lsp.buf.hover()*
     Displays hover information about the symbol under the cursor in a floating
-    window. Calling the function twice will jump into the floating window.
+    window. The window will be dismissed on cursor move. Calling the function
+    twice will jump into the floating window. Once focused in the floating
+    window, the buffer content can be scrolled as usual and `q` can be used to
+    dismiss the window.
```
justinmk, line 1403:
```suggestion
    twice will jump into the floating window (thus with the default mappings, "KK" will open the hover window and focus it). In the floating window, all commands and mappings are available as usual, except that "q" dismisses the window. You can scroll the contents the same as you would any other buffer.
```

MariaSolOs: Didn't know we had this convention. Maybe something to document in `:help dev`?

runtime/lua/vim/lsp/buf.lua
```diff
@@ -29,7 +29,13 @@ local function request(method, params, handler)
 end
 
 --- Displays hover information about the symbol under the cursor in a floating
---- window. Calling the function twice will jump into the floating window.
+--- window. The window will be dismissed on cursor move.
+--- Calling the function twice will will jump into the floating window
```
justinmk, line 33:
```suggestion
--- Calling the function twice will jump into the floating window
```

tris203: good spot, words are hard

PR #30349, opened 2024-09-12
fix(lsp): handle empty call hierarchy items

jamestrew:
Ensure that the function `pick_call_hierarchy_item` correctly handles the case where `call_hierarchy_items` is nil or an empty table. This prevents erroneous behavior when the function is called with an empty list.

### Steps to reproduce
Using steps from https://github.com/neovim/neovim/issues/30122
Replace step 1 with: Place the cursor on the call to foo_func() (63go)

### Expected result
To not get a input prompt with no valid selections

### Actual result
Get an input prompt with no valid selections
image

### DIFF
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -447,11 +447,9 @@ function M.document_symbol(opts)
   request_with_opts(ms.textDocument_documentSymbol, params, opts)
 end
 
---- @param call_hierarchy_items lsp.CallHierarchyItem[]?
+--- @param call_hierarchy_items lsp.CallHierarchyItem[]
+--- @return lsp.CallHierarchyItem?
 local function pick_call_hierarchy_item(call_hierarchy_items)
-  if not call_hierarchy_items then
-    return
-  end
   if #call_hierarchy_items == 1 then
     return call_hierarchy_items[1]
   end
@@ -476,7 +474,7 @@ local function call_hierarchy(method)
       vim.notify(err.message, vim.log.levels.WARN)
       return
     end
-    if not result then
+    if not result or vim.tbl_isempty(result) then
       vim.notify('No item resolved', vim.log.levels.WARN)
       return
     end


runtime/lua/vim/lsp/buf.lua
```diff
@@ -448,8 +448,9 @@ function M.document_symbol(opts)
 end
 
 --- @param call_hierarchy_items lsp.CallHierarchyItem[]?
+--- @return lsp.CallHierarchyItem?
 local function pick_call_hierarchy_item(call_hierarchy_items)
-  if not call_hierarchy_items then
+  if not call_hierarchy_items or vim.tbl_isempty(call_hierarchy_items) then
```
MariaSolOs, line 453: I'm unsure of whether we should have this check in this function or in the invocation place instead:
https://github.com/neovim/neovim/blob/512d63df5ee3f10f42afe009a9d32dc0113c32d4/runtime/lua/vim/lsp/buf.lua#L480-L482

Since the message in the snippet above refers to the case where the server didn't resolve any items, it makes sense to handle the empty case there IMO.

MariaSolOs: Although I'm unsure of what `nil` versus an empty array means here, and whether we should treat them differently...

MariaSolOs: Okay the protocol doesn't make a distinction between empty and a nil response, and [VS Code treats them the same too](https://github.com/microsoft/vscode/blob/94cd4ed34ca1f718615c0befb1b31181db42575e/src/vs/workbench/api/browser/mainThreadLanguageFeatures.ts#L845), so I do think we should move the check to the linked `if`.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -448,8 +448,9 @@ function M.document_symbol(opts)
 end
 
 --- @param call_hierarchy_items lsp.CallHierarchyItem[]?
```
MariaSolOs, line 450:
```suggestion
--- @param call_hierarchy_items lsp.CallHierarchyItem[]
```
We only call this function in one place, and there's already a `nil` check there.

runtime/lua/vim/lsp/buf.lua
jamestrew: Yeah I agree, I make the change

MariaSolOs (approved):

PR #30353, opened 2024-09-12
docs(treesitter): some functions are only valid for registered parsers

ofseed:
We do not load parsers automatically, even if the user has a `parser/lang.so`, `get_lang` will return `nil` before it is loaded or registered.

Edit: I mean `get_lang` instead of `get_parser`.

### DIFF
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -970,7 +970,7 @@ add({lang}, {opts})                            *vim.treesitter.language.add()*
                   language to load
 
 get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
-    Get the filetypes associated with the parser named {lang}.
+    Get the filetypes associated with the registered parser named {lang}.
 
     Parameters: ~
       • {lang}  (`string`) Name of parser
@@ -979,6 +979,7 @@ get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
         (`string[]`) filetypes
 
 get_lang({filetype})                      *vim.treesitter.language.get_lang()*
+    Get the registered parser name associated with the provided filetype.
 
     Parameters: ~
       • {filetype}  (`string`)
--- a/runtime/lua/vim/treesitter/language.lua
+++ b/runtime/lua/vim/treesitter/language.lua
@@ -7,7 +7,7 @@ local ft_to_lang = {
   help = 'vimdoc',
 }
 
---- Get the filetypes associated with the parser named {lang}.
+--- Get the filetypes associated with the registered parser named {lang}.
 --- @param lang string Name of parser
 --- @return string[] filetypes
 function M.get_filetypes(lang)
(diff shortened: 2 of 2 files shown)


clason:
I don't think that's quite true; `get_parser` _will_ load a parser for the specified language. Of course, a parser for that language (either literally or via registering) must exist, but I think that's quite obvious?

justinmk:
@clason are you suggesting a change to the wording, or should this be closed?

clason:
First of all, I would like some clarification on whether there's a misunderstanding (and whether it's on ofseed's or my side).

The missing `get_lang` docstring is a welcome addition in any case (modulo wording).

ofseed:
I just realized I made a mistake in my description. I meant to say `get_lang` instead of `get_parser` (so the actual change did not involve `get_parser`), apologies.

`get_parser` can load the corresponding parser because it first uses `get_lang` to get the registered language. If it does not exist, it tries to load it using `filetype` as the parser name and finally registers it.

The implicit issue here is that in the past when we used `vim.treesitter`, we also used `nvim-treesitter`, which would register all parsers after loading. So `get_lang` would rarely return `nil` if we pass a filetype to it. However, once we move away from `nvim-treesitter` (which is the future goal), if the user or plugin does not use `get_parser`, even `get_lang("zig")` could return `nil`. This misunderstanding can easily lead to code like:

```lua
vim.api.nvim_create_autocmd("FileType",{
  pattern = "zig",
  callback = function(args)
    local lang = vim.treesitter.language.get(args.match)
    if lang then
      local parser = vim.treesitter.get_parser(args.buf, lang)
    end
end})
```

Although this step seems redundant, it used to work fine without issues because `nvim-treesitter` would register all parsers. However, once we move away from `nvim-treesitter`, problems arise. So I want the documentation to be clearer.

clason:
That sounds to me more like a design issue than a documentation issue. The common pattern I'm seeing is `get_lang(ft) or ft`, and I would expect this to be the default behavior of the function.

(There are two other uses in our codebase; I'd have to look at these in detail whether they rely on the current behavior. If so, they should call an internal function without the fallback, or we should allow passing an option that disables the fallback.)

clason:
Hmm, this is trickier than I thought because of "composite" filetypes (`html.glimmer`)...

clason:
https://github.com/neovim/neovim/pull/30379

clason:
For the record, this:

> Although this step seems redundant, it used to work fine without issues because nvim-treesitter would register all parsers. However, once we move away from nvim-treesitter, problems arise.

is actually not true; `vim.treesitter.language.add` would register the loaded parser for a filetype, `nvim-treesitter` has nothing to do with it. (It used to have its own lookup table, but we deprecated that.)

ofseed:
> That sounds to me more like a design issue than a documentation issue.

I completely agree. I thought to do something like you are going to do in #30379, but what I'm concerned about is that maybe someone's code relies on this behavior, and if we change this behavior directly instead of using another function, the cause might be difficult to identify. So, I was inclined to leave everything as it was. However, I personally fully agree with making such a change.

justinmk:
shall we merge this or close it?

clason:
This is superseded by the linked PRs, so we shouldn't merge this. Thank you for bringing this up, though, that was helpful!

PR #30359, opened 2024-09-13
fix(lsp): resolve blob to string error in semantic tokens callback

petersid2022:
Hello everyone!

I put together this PR, which is my first contribution to Neovim, without being sure if I should have first created an issue or proposed the PR directly and described the issue here. Please excuse any possible mistakes or oversights on my part.

This PR fixes an issue that I've had for sometime, after successfully migrating from `nvim-cmp` to Neovim's builtin completion. More specifically, when I would trigger the omni completion via `CTRL_X CTRL_O` and try to select a `for ...` snippet, I would get this error message:

unnamed

After, investigating the stacktrace, I managed to fix the issue locally, by wraping the `handler(err, result, context)` call with a `pcall`, modifying this chunk from the `runtime/lua/vim/lsp/client.lua` file:
https://github.com/neovim/neovim/blob/deac7df80a1491ae65b68a1a1047902bcd775adc/runtime/lua/vim/lsp/client.lua#L679-L688

### Neovim version
```
NVIM v0.11.0-dev-deac7df
Build type: Debug
LuaJIT 2.1.1724512491
Run "nvim -V1 -v" for more info
```

For the purposes of easily reproducing the bug, I've created this simple `minimal_config.lua` file, after stumbling across #30122. 

First things first, create the `minimal_config.lua` file (preferably in your current working directory) and paste there the following:
```lua
local pattern = 'lua'

  pattern = pattern,
  callback = function()
    vim.lsp.start({
      name = 'testing-ls',
      cmd = cmd,
    })
  end
})
```
Finally, create a new file called `test.lua` (that's also in your current working directory) and paste there the following:
```lua
-- To trigger the bug, please follow these steps:
-- 1. Uncomment the `fo` line.
-- 2. Place your cursor at the end of the line.
-- 3. Go into insert mode by pressing `a`.
-- 4. Press CTRL-X CTRL-O to activate the completion.
--
-- IMPORTANT: As far as I'm aware this only works with `for ...` snippets,
-- so when testing make sure you select something relevant.

-- fo
```

Now, you should be able to reproduce the bug by running:
```
nvim --clean -u minimal_config.lua test.lua
```
And following the instructions on screen.

### Expected behavior
Neovim should be able to select any option that omnifunc serves inside the pmenu without throwing an exception when coming across a `for ...` snippet.

### Actual behavior
Uppon triggering omni-completion using `CTRL_X CTRL_O`, and selecting a `for ...` snippet, neovim throws this exception:
```
Error executing vim.schedule lua callback: ...local/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:304: Vim:E976: Using a Blob as a String 
stack traceback:
[builtin#36]: at 0x5cfd722462d0
...local/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:304: in function 'handler' /usr/local/share/nvim/runtime/lua/vim/lsp/client.lua:687: in function " vimLeditor.lua: in function <vimLeditor.lua:0> 
Press ENTER or type command to continue 
``` 

Please let me know if there are any adjustments or additional steps I should take. I'm happy to provide more information or make changes based on feedback. Thank you for your time!

### DIFF
--- a/runtime/lua/vim/lsp/client.lua
+++ b/runtime/lua/vim/lsp/client.lua
@@ -684,7 +684,7 @@ function Client:_request(method, params, handler, bufnr)
       params = params,
       version = version,
     }
-    handler(err, result, context)
+    pcall(handler, err, result, context)
   end, function(request_id)
     local request = self.requests[request_id]
     request.type = 'complete'


glepnir:
Thanks for the PR. But unfortunately this may not be correct. It just ignores the error. I think the problem should be for snippet with line breaks in llike ` for xx in \n `https://github.com/neovim/neovim/blob/057314345a7cfc8e52bbe13a595759d6ca52ac20/src/nvim/insexpand.c#L3638

Here the string is converted into bytes and inserted into the buffer. For example \n \t. I have a PR in https://github.com/vim/vim/pull/15373 (draft has not been updated recently) to handle similar special symbols. Instead of inserting blobs. Maybe it will solve this problem. If I am wrong, please ignore me.

petersid2022:
First of all thank you for the quick reply! I understand that my PR, doesn't actually address the issue but instead ignores any errors it produces - thus I'm cool if it doesn't get merged. My solution, is rather hack-y and it shouldn't be merged upstream. After checking out your PR, I'm very happy that this issue is under development and I hope I didn't complicate the process further. Should I close this PR then?

MariaSolOs:
I agree with @glepnir here. This change doesn't fix the issue, but it silently swallows it instead. Feel free to follow @glepnir's suggestion for a true fix though!

mfussenegger:
As the others mentioned this isn't the proper fix and swallowing errors would just mask bugs.

Looks like the issue is caused by a `vim.fn.strlen("for index, value in ipairs(t) do\0\t\0end")` call triggered by semantic tokens updates.

I think https://github.com/neovim/neovim/pull/30315 might prevent this from happening as a side effect - because with it we strip the `word`, preventing it from temporarily inserting the full string.

glepnir:
@petersid2022  work around is `vim.lsp.completion` support a `convert` field in opts table. you can do some check to remove special character then generate a new `word` like

```lua
convert = function (item) return { word = item:gsub('\n') } end 
```

glepnir:
if you are using `vim.lsp.completion` module there has `convert` filed that you can do some hack to remove special character then generate `word` completete-item. if snippet is from lsp. because snippet insert is from textEdit.newtext so modifed is okay

PR #30362, opened 2024-09-13
ci: don't add reviewers for PRs created by a bot

dundargoc:
This will ensure automatic backports created by the backport action does
not request reviewers (since the commit in question has already been
vetted and merged), but manual backports created by users does request
reviewers as these commits has not been vetted previously.

### DIFF
--- a/.github/workflows/reviewers_add.yml
+++ b/.github/workflows/reviewers_add.yml
@@ -5,7 +5,7 @@ on:
   workflow_call:
 jobs:
   request-reviewer:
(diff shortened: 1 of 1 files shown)


clason:
(unless it's a manual backport, but then we can still add them manually)

dundargoc:
That's a good point. Right now we're not distinguishing between automatic and manual backports. We can change it around so that a reviewer is not added the PR was created by a bot.

dundargoc:
This is not working correctly. Maybe neovim-backports is not considerer a bot for some reason. Gonna fallback to just skip adding reviewers on the release branch I think.

PR #30368, opened 2024-09-13
feat(quickfix): handle name display in qflist for hard links

austin880625:
Hi I'm new to this project. Please point out if I did anything dumb here.
This PR is to handle the filename display in quickfix list when some files in the lists are each others' hard links.

## The Problem

Say we have a C++ file `main1.cpp` and make a hard link `main2.cpp` with following content:

```
void sample() {}

int main() {
  sample();
  return 0;
}
```

And generate the `compile_commands.json` by build against `main1.cpp` (by something like `bear -- g++ main1.cpp`). Then with `clangd` LSP server when we open `main2.cpp` then call `vim.lsp.buf.references` to display the quickfix list of the references. The list would be shown as if the references on `main2.cpp` are duplicated. We expect the filename to be different for the two references, one for `main1.cpp` and another for `main2.cpp`.

The major cause is that qflist assumed that the filename passed into `qf_add_entry` always matches the filename opened with the buffer. Implementations like [Telescope](https://github.com/nvim-telescope/telescope.nvim/blob/0.1.8/lua/telescope/builtin/__lsp.lua#L42) collects the response from LSP directly into its picker without passing into `vim.fn.*` APIs, so it doesn't have such issue.

## The Solution

This can be a design decision given that the buffer also holds only one filename if we try to open other hard links. But this PR deals with quickfix to have same behavior(and I personally favored) with Telescope.
I introduced `qf_fname` in the `qfline_S` structure to store the filename for each quickfix line, then use the field in `qf_list_entry`. In this way we can get the filename as passed in from LSP to setloclist/setqflist

### DIFF
--- a/src/nvim/quickfix.c
+++ b/src/nvim/quickfix.c
@@ -90,6 +90,7 @@ struct qfline_S {
   int qf_col;             ///< column where the error occurred
   int qf_end_col;         ///< column when the error has range or zero
   int qf_nr;              ///< error number
+  char *qf_fname;         ///< filename for the line
   char *qf_module;        ///< module name for this error
   char *qf_pattern;       ///< search pattern for the error
   char *qf_text;          ///< description of the error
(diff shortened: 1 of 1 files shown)


austin880625:
seems it broke some functional tests. I will be working on fixing this.

justinmk:
Thanks for the interest.

Enhancements to quickfix should be sent to https://github.com/vim/vim . Especially for a small enhancement like this, it's not worth the trouble of "forking" the quickfix interface.

PR #30372, opened 2024-09-13
refactor(tests): rename terminal/testutil.lua => testterm.lua

justinmk:
This module is generally used by any tests that need the full Nvim TUI instead of `screen.lua`. Thus it should live in `functional/` instead of in `functional/terminal/`.

### DIFF
--- a/test/functional/testterm.lua
+++ b/test/functional/testterm.lua
@@ -1,6 +1,13 @@
--- To test tui/input.c, this module spawns `nvim` inside :terminal and sends
--- bytes via jobsend().  Note: the functional/testutil.lua test-session methods
--- operate on the _host_ session, _not_ the child session.
+-- Functions to test :terminal and the Nvim TUI.
+-- Starts a child process in a `:terminal` and sends bytes to the child via nvim_chan_send().
+-- Note: the global functional/testutil.lua test-session is _host_ session, _not_
+-- the child session.
+--
+-- - Use `setup_screen()` to test `:terminal` behavior with an arbitrary command.
+-- - Use `setup_child_nvim()` to test the Nvim TUI.
+--    - NOTE: Only use this if your test actually needs the full lifecycle/capabilities of the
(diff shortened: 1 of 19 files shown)


zeertzjq:
Apparently that's not true, as a lot of that is running `tty-test` in `:terminal`

justinmk:
True, but `setup_child_nvim` exists for testing the TUI

dundargoc (approved):

zeertzjq:
> True, but `setup_child_nvim` exists for testing the TUI. And is used in 4 tests outside of terminal/. Unless those tests are doing something wrong, I would guess we want more tests to make use of this instead of trying to do things "manually".

I can't understand what you mean. What does this have to do with the naming of the file?

justinmk:
> I'd rather name it `testterm`.

done

test/functional/testterm.lua
```diff
@@ -9,18 +16,20 @@ local exec_lua = n.exec_lua
 local api = n.api
 local nvim_prog = n.nvim_prog
 
-local function feed_data(data)
+local M = {}
+
+function M.feed_data(data)
```
justinmk, line 21: this seems to help LuaLS discover the docs

test/functional/testterm.lua
```diff
@@ -34,60 +43,68 @@ end
...
-local function screen_setup(extra_rows, command, cols, env, screen_opts)
+--- Runs `cmd` in a :terminal, and returns a `Screen` object.
+---
+---@param extra_rows? integer Extra rows to add to the default screen.
+---@param cmd? string[] Command to run in the terminal (default: `{ 'tty-test' }`)
+---@param cols? integer Create screen with this many columns (default: 50)
+---@param env? table Environment set on the `cmd` job.
+---@param screen_opts? table Options for `Screen.new()`.
+---@return test.functional.ui.screen # Screen attached to the global (not child) Nvim session.
+function M.setup_screen(extra_rows, cmd, cols, env, screen_opts)
```
justinmk, line 105: renamed from `screen_setup` to align with `setup_child_nvim`

PR #30379, opened 2024-09-14
feat(treesitter)!: remove unnecessary side effects from  `language.add` and add return value

clason:
# Problem

Language names are only registered for filetype<->language lookups when parsers are actually loaded; this means users cannot rely on `vim.treesitter.language.get_lang()` or `get_filetypes()` to return
the correct value when language and filetype coincide and always need to add explicit fallbacks.

# Solution

 Always return the language name as valid filetype in `get_filetypes()`, and default to the filetype in `get_lang()`. Document this behavior.

Supersedes #30353

---

# Problem

No clear way to check whether parsers are available for a given language.

# Solution

Make `language.add()` return `true` if a parser was successfully added and `nil` otherwise. Use explicit `assert` instead of relying on thrown errors.

Cf. #30382 

---

Do not squash.

---

Note to self: needs (trivial) adapting for in nvim-treesitter and telescope.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -97,12 +97,12 @@ TREESITTER
   capture IDs to a list of nodes that need to be iterated over. For
   backwards compatibility, an option `all=false` (only return the last
   matching node) is provided that will be removed in a future release.
-
 • |vim.treesitter.language.get_filetypes()| always includes the {language}
   argument in addition to explicitly registered filetypes.
-
 • |vim.treesitter.language.get_lang()| falls back to the {filetype} argument
   if no languages are explicitly registered.
+• |vim.treesitter.language.add()| returns `true` if a parser was loaded
+  successfully and `nil,errmsg` otherwise instead of throwing an error.
 
 TUI
 
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -98,6 +98,12 @@ TREESITTER
   backwards compatibility, an option `all=false` (only return the last
   matching node) is provided that will be removed in a future release.
 
+• |vim.treesitter.language.get_filetypes()| always includes the {language}
+  argument in addition to explicitly registered filetypes.
+
+• |vim.treesitter.language.get_lang()| falls back to the {filetype} argument
+  if no languages are explicitly registered.
+
 TUI
 
 • TODO
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -969,14 +969,15 @@ add({lang}, {opts})                            *vim.treesitter.language.add()*
     Parameters: ~
       • {lang}  (`string`) Name of the parser (alphanumerical and `_` only)
       • {opts}  (`table?`) Options:
-                • {filetype}? (`string|string[]`, default: {lang}) Default
-                  filetype the parser should be associated with.
                 • {path}? (`string`) Optional path the parser is located at
                 • {symbol_name}? (`string`) Internal symbol name for the
                   language to load
 
 get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
-    Get the filetypes associated with the parser named {lang}.
+    Returns the filetypes for which a parser named {lang} is used.
+
+    The list includes {lang} itself plus all filetypes registered via
+    |vim.treesitter.language.register()|.
 
     Parameters: ~
       • {lang}  (`string`) Name of parser
@@ -985,6 +986,11 @@ get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
         (`string[]`) filetypes
 
 get_lang({filetype})                      *vim.treesitter.language.get_lang()*
+    Returns the language name to be used when loading a parser for {filetype}.
+
+    If no language has been explicitly registered via
+    |vim.treesitter.language.register()|, default to {filetype}. For composite
+    filetypes like `html.glimmer`, only the main filetype is returned.
 
     Parameters: ~
       • {filetype}  (`string`)
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -964,7 +964,12 @@ add({lang}, {opts})                            *vim.treesitter.language.add()*
     Load parser with name {lang}
 
     Parsers are searched in the `parser` runtime directory, or the provided
-    {path}
+    {path}. Can be used to check for available parsers before enabling
+    treesitter features, e.g., >lua
+          if vim.treesitter.language.add('markdown') then
+            vim.treesitter.start(bufnr, 'markdown')
+          end
+<
 
     Parameters: ~
       • {lang}  (`string`) Name of the parser (alphanumerical and `_` only)
@@ -972,6 +977,10 @@ add({lang}, {opts})                            *vim.treesitter.language.add()*
                 • {path}? (`string`) Optional path the parser is located at
                 • {symbol_name}? (`string`) Internal symbol name for the
                   language to load
+
+    Return (multiple): ~
+        (`boolean?`) True if parser is loaded
+        (`string?`) Error if parser cannot be loaded
 
 get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
     Returns the filetypes for which a parser named {lang} is used.
--- a/runtime/lua/vim/treesitter/language.lua
+++ b/runtime/lua/vim/treesitter/language.lua
@@ -62,8 +62,22 @@ function M.require_language(lang, path, silent, symbol_name)
     return installed
   end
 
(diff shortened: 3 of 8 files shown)


clason:
@echasnovski could you please check whether this still works for built-in comments? (Seems to be in my brief test, but I may have missed some edge cases.)

runtime/doc/treesitter.txt
```diff
@@ -979,6 +982,11 @@ get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
         (`string[]`) filetypes
 
 get_lang({filetype})                      *vim.treesitter.language.get_lang()*
+    Returns the language name to be used when loading a parser for {filetype}.
+
+    If no language has been explicitly registered via
+    |vim.treesitter.language.register()|, default to {filetype}. For composite
+    filetypes like `html.glimmer`, only the main filetype is returned.
```
clason, line 989: @zeertzjq I can't find where the help files talk about this feature. Is there some standard terminology I can use instead?

clason:
Note that this is in effect a change in meaning: Before, the functions returned _actual_ associations (reliable because they were added manually or only on successful loading); now they return _potential_ associations.

But that validation was never really fully usable since nvim-treesitter registered all languages upfront, even if a user did not install them. So I think this is the better meaning, and how it is actually used.

echasnovski:
> @echasnovski could you please check whether this still works for built-in comments? (Seems to be in my brief test, but I may have missed some edge cases.)

I also don't notice issues after some testing. It also seems to be more or less fine after looking at code, because the worst that can happen is [this `vim.filetype.get_option()`](https://github.com/neovim/neovim/blob/3b54adc6c6d25dd146fa1eac21321f6bd612f50a/runtime/lua/vim/_comment.lua#L38) returning default global 'commentstring', which arguably should be `''` and later ignored. So should be fine.

------

That said, I'd personally find it somewhat confusing mixing filetype and language abstractions. Especially after the period of their careful separation.

Also (at least current implementation) may result into duplicating entries (like `vim.treesitter.language.get_filetypes('lua')`), which is somewhat confusing.

clason:
Thanks for checking!

> That said, I'd personally find it somewhat confusing mixing filetype and language abstraction.

It's not, though; this is just explicitly handling _in these functions_ the default fallback that is already made elsewhere (e.g., `vim.treesitter.start()`).

> Also (at least current implementation) may result into duplicating entries (like vim.treesitter.language.get_filetypes('lua')), which is somewhat confusing.

How? Because you explicitly did `vim.treesitter.language.register('lua', 'lua')`? (You shouldn't do this anymore; that's why the commit is marked as "breaking".)

echasnovski:
> How? Because you explicitly did `vim.treesitter.language.register('lua', 'lua')`? (You shouldn't do this anymore; that's why the commit is marked as "breaking".)

I do not have this explicitly in the config. The culprit seems to be `master` branch of 'nvim-treesitter/nvim-treesitter'. Having done `:packadd nvim-treesitter` makes this duplication appear.

clason:
As I said, it's a breaking change, and plugins need to adapt. I'll handle this in nvim-treesitter.

echasnovski:
> But is this really breaking anything?

Not that I can think of right now. Just "... is somewhat confusing.".

runtime/doc/treesitter.txt
```diff
@@ -979,6 +982,11 @@ get_filetypes({lang})                *vim.treesitter.language.get_filetypes()*
         (`string[]`) filetypes
 
 get_lang({filetype})                      *vim.treesitter.language.get_lang()*
+    Returns the language name to be used when loading a parser for {filetype}.
```
justinmk, line 985: Future note: This appears to be the dual of `register()`. Can we name register() to `set_lang()` ? 

We do not use the name "register" anywhere else in stdlib AFAICT, and this thankfully looks like the last such usage.

justinmk (approved):

justinmk:
> `vim.treesitter.language.register('lua', 'lua')`? (You shouldn't do this anymore; that's why the commit is marked as "breaking".)

That practical example may be worth mentioning in the news item.

runtime/doc/treesitter.txt
clason: No, `set_lang` implies you set _the_ language; this adds it to a "register". I don't thinks there's a precedent in the API for this.

clason:
> That practical example may be worth mentioning in the news item.

Nah, the only one who'd be doing that is me, and I'll just stop.

runtime/doc/treesitter.txt
justinmk: we don't have `add_lang`.

clason: True, and that could work, but I feel that that would be confusing with `vim.treesitter.language.add` (which is... not a great name).

To be blunt: without a complete rethink of the API, this sort of thing is putting lipstick on a pig...

clason:
Todo: figure out how this factors into https://github.com/neovim/neovim/blob/3b54adc6c6d25dd146fa1eac21321f6bd612f50a/runtime/lua/vim/treesitter/languagetree.lua#L737

runtime/doc/treesitter.txt
zeertzjq: It's in Vim's `:h improvements-7` :)

clason: Worst documented feature ever ;) I think that should be somewhere in `filetype` -- I've seen repeated misunderstandings of how this works...

runtime/doc/news.txt
```diff
@@ -88,6 +88,16 @@ TREESITTER
   backwards compatibility, an option `all=false` (only return the last
   matching node) is provided that will be removed in a future release.
 
+• |vim.treesitter.language.get_filetypes()| always includes the {language}
+  argument in addition to explicitly registered filetypes.
+
+• |vim.treesitter.language.get_lang()| falls back to the {filetype} argument
+  if no languages are explicitly registered.
+
```
ribru17, line 96: Nit: it looks like the other bulleted lists in these sections do not have empty lines in-between list items

runtime/lua/vim/treesitter/language.lua
```diff
@@ -71,53 +76,54 @@ end
...
 
   if vim.endswith(path, '.wasm') then
-    if not vim._ts_add_language_from_wasm then
-      error(string.format("Unable to load wasm parser '%s': not built with ENABLE_WASMTIME ", path))
-    end
-    vim._ts_add_language_from_wasm(path, lang)
+    return vim._ts_add_language_from_wasm and vim._ts_add_language_from_wasm(path, lang)
   else
-    vim._ts_add_language_from_object(path, lang, symbol_name)
+    return vim._ts_add_language_from_object(path, lang, symbol_name)
```
przepompownia, line 125: The above `else` has become redundant.

runtime/doc/news.txt
```diff
@@ -87,6 +87,13 @@ TREESITTER
   capture IDs to a list of nodes that need to be iterated over. For
   backwards compatibility, an option `all=false` (only return the last
   matching node) is provided that will be removed in a future release.
+• |vim.treesitter.language.get_filetypes()| always includes the {language}
+  argument in addition to explicitly registered filetypes.
```
justinmk, line 91: returning extra data is not usually considered breaking.

justinmk (approved):
Looks great

runtime/doc/news.txt
clason: It's not extra, it's returning where it used to not return -- and it's reasonable to assume that some people relied on the `nil` to be able to, e.g., do something else instead of starting treesitter. We could claim this falls under "treesitter is experimental", but I thought it better to err on the side of caution and give a shoutout here.

runtime/lua/vim/treesitter/language.lua
```diff
@@ -96,29 +102,27 @@ function M.add(lang, opts)
...
     end
     path = paths[1]
   end
 
   if vim.endswith(path, '.wasm') then
-    if not vim._ts_add_language_from_wasm then
-      error(string.format("Unable to load wasm parser '%s': not built with ENABLE_WASMTIME ", path))
-    end
-    vim._ts_add_language_from_wasm(path, lang)
+    return vim._ts_add_language_from_wasm and vim._ts_add_language_from_wasm(path, lang)
```
clason, line 123: @lewis6991 This may prioritize a wasm parser over a native one even though we may not build with WASM and then fail.

On the other hand, we would happily have tried to load a `python.borkbork` even before that...

clason:
Changed the return values of `language.add()` to `true` or `nil,errmsg` as discussed.

ribru17 (approved):

PR #30387, opened 2024-09-15
refactor: rename "process" => "proc"

justinmk:
## Problem:
- "process" is often used as a verb (`multiqueue_process_events`), which is ambiguous for cases where it's used as a topic.
- We use "proc" in several places already.
- The documented naming convention for processes is "proc". `:help dev-name-common`
- Shorter is better, when it doesn't harm readability or discoverability.

## Solution:
Rename `process` => `proc`.

### DIFF
--- a/src/nvim/event/defs.h
+++ b/src/nvim/event/defs.h
@@ -142,30 +142,31 @@ struct socket_watcher {
 };
 
 typedef enum {
-  kProcessTypeUv,
-  kProcessTypePty,
-} ProcessType;
+  kProcTypeUv,
+  kProcTypePty,
+} ProcType;
 
-typedef struct process Process;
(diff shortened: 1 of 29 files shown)


src/nvim/event/defs.h
```diff
@@ -142,30 +142,31 @@ struct socket_watcher {
...
-} ProcessType;
+  kProcTypeUv,
+  kProcTypePty,
+} ProcType;
 
-typedef struct process Process;
-typedef void (*process_exit_cb)(Process *proc, int status, void *data);
-typedef void (*internal_process_cb)(Process *proc);
+/// OS process
+typedef struct proc Proc;
```
justinmk, line 150: This one case may be questionable. Can revert this particular type name if anyone feels strongly about it.

PR #30389, opened 2024-09-15
vim-patch:9.1.0731: inconsistent case sensitive extension matching

clason:
Problem:  inconsistent case sensitive extension matching
Solution: unify case sensitive extension matching (Evgeni Chasnovski).

There are different approaches of how extensions are matched with
respect to case sensitivity. In particular, '\c' flag is used in pattern
whereas in most places case sensitive matching is guarded behind
`has("fname_case")` condition.

Replace all instances of '\c' with an explicit case sensitive pattern
variants guarded by `has("fname_case")`. Strictly speaking, this is a
breaking change because only two (most common and prevailingly tested)
variants are now matched: upper first letter and upper all letters.

closes: vim/vim#15672

https://github.com/vim/vim/commit/59b089c9df238ce940b3c593295bc511e201399a

Co-authored-by: Evgeni Chasnovski <evgeni.chasnovski@gmail.com>


echasnovski:
I was just going to ask if you are working on this :)

zeertzjq (approved):

PR #30392, opened 2024-09-16
docs(eval): update param type of prompt-buffer functions

glepnir:
Problem: wrong type of  functions.

Soluton: update the param type of these functions buf with interger expr with string|function.

### DIFF
--- a/src/nvim/eval.lua
+++ b/src/nvim/eval.lua
@@ -8044,7 +8044,7 @@ M.funcs = {
 
     ]=],
     name = 'prompt_getprompt',
-    params = { { 'buf', 'any' } },
+    params = { { 'buf', 'integer|string' } },
     signature = 'prompt_getprompt({buf})',
   },
   prompt_setcallback = {
@@ -8084,7 +8084,7 @@ M.funcs = {
 
     ]=],
     name = 'prompt_setcallback',
-    params = { { 'buf', 'any' }, { 'expr', 'any' } },
+    params = { { 'buf', 'integer|string' }, { 'expr', 'string|function' } },
     signature = 'prompt_setcallback({buf}, {expr})',
   },
   prompt_setinterrupt = {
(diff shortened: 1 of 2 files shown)


src/nvim/eval.lua
```diff
@@ -8044,7 +8044,7 @@ M.funcs = {
 
     ]=],
     name = 'prompt_getprompt',
-    params = { { 'buf', 'any' } },
+    params = { { 'buf', 'integer' } },
```
zeertzjq, line 8047:
```suggestion
    params = { { 'buf', 'integer|string' } },
```

src/nvim/eval.lua
```diff
@@ -8084,7 +8084,7 @@ M.funcs = {
 
     ]=],
     name = 'prompt_setcallback',
-    params = { { 'buf', 'any' }, { 'expr', 'any' } },
+    params = { { 'buf', 'integer' }, { 'expr', 'string|function' } },
```
zeertzjq, line 8087:
```suggestion
    params = { { 'buf', 'integer|string' }, { 'expr', 'string|function' } },
```

src/nvim/eval.lua
```diff
@@ -8101,7 +8101,7 @@ M.funcs = {
 
     ]=],
     name = 'prompt_setinterrupt',
-    params = { { 'buf', 'any' }, { 'expr', 'any' } },
+    params = { { 'buf', 'integer' }, { 'expr', 'string|function' } },
```
zeertzjq, line 8104:
```suggestion
    params = { { 'buf', 'integer|string' }, { 'expr', 'string|function' } },
```

src/nvim/eval.lua
```diff
@@ -8116,7 +8116,7 @@ M.funcs = {
       <
     ]=],
     name = 'prompt_setprompt',
-    params = { { 'buf', 'any' }, { 'text', 'any' } },
+    params = { { 'buf', 'integer' }, { 'text', 'string' } },
```
zeertzjq, line 8119:
```suggestion
    params = { { 'buf', 'integer|string' }, { 'text', 'string' } },
```

Issue #30305, opened 2024-09-08
LSP: show selected word in quickfix title for LSP commands e.g. "grr"


ColinKennedy:
### Problem

When I type `gr` to show references I get a generic quickfix list title that just says `References`

![image](https://github.com/user-attachments/assets/6421a1e9-bfb8-4ff1-ac83-6762ae7231bb)

Running :chistory on a number of these calls are indistinguishable from one another because they are all titled References

![image](https://github.com/user-attachments/assets/b954fa76-ae49-4aa9-a617-d239efa6fb30)

### Expected behavior

The `gr` command (and other similar LSP commands) include the word of whatever was selected (see `:chistory` screenshot)

![image](https://github.com/user-attachments/assets/ff0c0e69-2706-4d5c-a38c-321839847461)

glepnir:
emm  the screenshot it's  not neovim defualt lsp handler? run `:lua vim.lsp.buf..reference()` it will show the pattern in qflist.

ColinKennedy:
Ah whoops, I forgot I added an old custom keymap. This is my definition for `gr`.

```lua
vim.keymap.set(
    "n",
    "gr",
    vim.lsp.buf.references,
    {
        buffer=buffer,
        desc="[g]o to [r]eferences - Show all locations where a variable is used."
    }
)
```

So the screenshot from before was already calling the lua function you'd recommended for testing. And `grr` is the same result as the screenshot of course.

My setup is fairly custom so I wouldn't be surprised if I accidentally messed with the default LSP handler but FWIW I don't recall changing anything. How can I check that?

ColinKennedy:
Are you sure the issue is on my side? From the source code it looks like the title really is just hard-coded 

https://github.com/neovim/neovim/blob/b40ec083ae30eb64ddfec2b238c909116e936e99/runtime/lua/vim/lsp/handlers.lua#L245-L268

glepnir:
there pass line as text 

https://github.com/neovim/neovim/blob/b40ec083ae30eb64ddfec2b238c909116e936e99/runtime/lua/vim/lsp/util.lua#L1810-L1818


![QQ_1725781801995](https://github.com/user-attachments/assets/0657ef42-f30c-4e46-8450-ae4c968d71cd)

glepnir:
`:help vim.lsp.ListOpts` you can custom it.

```lua
local function on_list(options)
  options.title = 'Reference: ' .. vim.fn.expand('<cword>')
[...]

PR #30394, opened 2024-09-16
feat(lsp): show selected word in quickfix title #30305

Naofal-Helal:
Problem: `vim.lsp.buf.references` creates a quickfix list with a generic title as seen from :chistory :
image

Solution: ~Include the selected word in the title at the time of the request~
add the selected word to a context or vim.lsp.ListOpts

ref #30305


justinmk:
No tests are failing, that means we don't have a test that exercises this?

glepnir:
`vim.lsp.ListOpts` is designed to be configurable.  If you want to hardcode title there need to add it to all lsp methods that call qflist. Otherwise don't add it to any.  `title: word` or `title for word`? Cannot be defined as good or bad. Maybe we should configure the behavior here in our lsp common interface..

Naofal-Helal:
Is it safe to add stuff to the lsp request context/params or would that cause problems.  e.g. here:
https://github.com/neovim/neovim/blob/b40ec083ae30eb64ddfec2b238c909116e936e99/runtime/lua/vim/lsp/buf.lua#L434-L441
adding
```lua
params.targetSymbol = "..."
or
params.context.targetSymbol = "..."
```

mfussenegger:
> Is it safe to add stuff to the lsp request context/params or would that cause problems. e.g. here:

It should work but I wouldn't go that direction because it makes the typing more troublesome.
In order to handle multiple clients we'd need to move away from the current split into buf.references and handler.references - which would allow to pass along the target symbol with a closure.


The whole thing would probably look something like this:

```diff
diff --git a/runtime/lua/vim/lsp/buf.lua b/runtime/lua/vim/lsp/buf.lua
index a43ad8a69..61199368b 100644
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -431,15 +431,53 @@ end
 ---@param context (table|nil) Context for the request
 ---@see https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_references
 ---@param opts? vim.lsp.ListOpts
 function M.references(context, opts)
   validate({ context = { context, 't', true } })
-  local params = util.make_position_params()
-  params.context = context or {
-    includeDeclaration = true,
-  }
-  request_with_opts(ms.textDocument_references, params, opts)
+  local clients = vim.lsp.get_clients({ method = ms.textDocument_references })
+  if not next(clients) then
+    return
+  end
+  local win = api.nvim_get_current_win()
+  local remaining = #clients
+  opts = opts or {}
+
+  local all_items = {}
+  local title = 'References'
+
+  local function on_done()
+    if not next(all_items) then
+      vim.notify('No references found')
+    else
+      local list = { title = title, items = all_items }
+      if opts.loclist then
+        vim.fn.setloclist(0, {}, ' ', list)
+        vim.cmd.lopen()
+      elseif opts.on_list then
+        assert(vim.is_callable(opts.on_list), 'on_list is not a function')
+        opts.on_list(list)
+      else
+        vim.fn.setqflist({}, ' ', list)
+        vim.cmd('botright copen')
+      end
+    end
+  end
+
+  for _, client in ipairs(clients) do
+    local params = util.make_position_params(win, client.offset_encoding)
+    params.context = context or {
+      includeDeclaration = true,
+    }
+    client.request(ms.textDocument_references, params, function(err, result)
+      local items = util.locations_to_items(result or {}, client.offset_encoding)
+      vim.list_extend(all_items, items)
+      remaining = remaining - 1
+      if remaining == 0 then
+        on_done()
+      end
+    end)
+  end
 end
 
 --- Lists all symbols in the current buffer in the quickfix window.
 --- @param opts? vim.lsp.ListOpts
 function M.document_symbol(opts)
 ```

**Comment by Naofal-Helal** [comment]
Sorry for the late response.


> we'd need to move away from the current split into buf.references and handler.references - which would allow to pass along the target symbol with a closure.
> 

 I don't quite understand where the closure would be, and your changes would also need to be done for the other methods (get definition, get declaration, etc.)

After tinkering, the best solution I could figure is to introduce `get_target()` in `vim.lsp.ListOpts` and `target` in `vim.lsp.LocationOpts.OnList`, this would allow users to provide a custom `on_list` with access to the target symbol.

**Comment by justinmk** [comment]
try rebasing and force-pushing.

**Comment by mfussenegger** [comment]
I'm not sure anymore if this is even a good idea to include the `<cword>`.
With language servers like eclipse.jdt.ls you can get references for a function even if the cursor is not above the method name but for example inside the docs or the annotation.

For example:


```java
/**
 * Dummy
 **/
@Nullable
public static String foo() {
    return null;
}
```

Here the cursor could be on a `*` or over `Dummy` or `@Nullable` and you'd get
something like `References: Nullable` shown despite seeing references for
`foo()`.


This would be more confusing than helpful.

PR #30396, opened 2024-09-16
feat(docs) : Gleam

Malix-Labs:
Add Gleam compatibility in Readme


Malix-Labs:
Indeed !

Should I also add it ?

https://github.com/gleam-lang/gleam.vim

clason:
That's a filetype plugin, not an API client -- that's something very different! There's hundreds of these; we're not going to list them all individually...

Malix-Labs:
ok, thks !

PR #30398, opened 2024-09-16
docs(tui): rename term.txt, nvim_terminal_emulator.txt

justinmk:
## Problem:
It has long been a convention that references to the builtin terminal UI should mention "tui", not "term", in order to avoid ambiguity vs the builtin `:terminal` feature. The final step was to rename term.txt; let's take that step.

## Solution:
- rename term.txt => tui.txt
- rename nvim_terminal_emulator.txt => terminal.txt

### DIFF
--- a/scripts/vim-patch.sh
+++ b/scripts/vim-patch.sh
@@ -301,10 +301,6 @@ preprocess_patch() {
   LC_ALL=C sed -Ee 's/( [ab]\/runtime\/doc)\/sponsor\.txt/\1\/intro.txt/g' \
     "$file" > "$file".tmp && mv "$file".tmp "$file"
 
-  # Rename terminal.txt to nvim_terminal_emulator.txt
-  LC_ALL=C sed -Ee 's/( [ab]\/runtime\/doc)\/terminal\.txt/\1\/nvim_terminal_emulator.txt/g' \
-    "$file" > "$file".tmp && mv "$file".tmp "$file"
-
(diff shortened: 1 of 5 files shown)


scripts/vim-patch.sh
```diff
@@ -301,10 +301,6 @@ preprocess_patch() {
   LC_ALL=C sed -Ee 's/( [ab]\/runtime\/doc)\/sponsor\.txt/\1\/intro.txt/g' \
     "$file" > "$file".tmp && mv "$file".tmp "$file"
 
-  # Rename terminal.txt to nvim_terminal_emulator.txt
-  LC_ALL=C sed -Ee 's/( [ab]\/runtime\/doc)\/terminal\.txt/\1\/nvim_terminal_emulator.txt/g' \
-    "$file" > "$file".tmp && mv "$file".tmp "$file"
```
justinmk, line 306: fyi @zeertzjq

justinmk:
Why doesn't git recognize this as a "move" ?

zeertzjq:
Because the old file still exists

justinmk:
Hmm. I thought it would be smarter than that. git is supposed to content-based "move" detection.

Well, I'll add some "redirect" logic to `gen_help_files.lua` instead, that's better anyway.

clason:
> Hmm. I thought it would be smarter than that. git has content-based "move" detection.

Git, yes (and if you look at the diff, that is working), but not Github (on the file level -- only on the repo/org level).

Issue #26346, opened 2023-12-01
Treesitter: syntax highlighting for command line, window


clason:
### Problem

The command line does not have syntax highlighting, despite having a shiny new Vimscript parser at home. Also, enabling treesitter highlighting for vimscript breaks/disables the legacy parser for the command window (`q:`).

### Expected behavior

Hook a treesitter highlighter into the command line (and window) so we can have rainbow goodness (including highlighting of embedded Lua). This requires extending the `TSHighlighter` with a new mode that parses each line independently as its own tree (or writing a custom `TSLineHighlighter`).

This will also allow us to remove the custom (unmaintained) vimscript parser used for `<c-r>=` expression highlighting (although that will require some trickery to parse the whole line as Vim expression rather than statement).

justinmk:
Should we also eliminate the [":help expr-highlight" groups](https://github.com/neovim/neovim/blob/988b472d902e6116353fc2ae605bed32a37e6391/src/nvim/highlight_group.c#L463-L645) (`NvimPlainAssignment`, etc.)? These seem over-specific and I would think we want to use our core highlight groups instead.

rudiejd:
I've been poking around with this a little bit today, and I found that if you set `syntax=vim` in the cmdwin, you get what looks like Treesitter syntax highlighting. This seems like a similar issue to https://github.com/folke/noice.nvim/blob/92433164e2f7118d4122c7674c3834d9511722ba/lua/noice/text/syntax.lua#L12.   I made an autocommand for my config:

```lua
vim.api.nvim_create_autocmd('CmdwinEnter', {
  pattern = '[:>]',
  desc = 'If the treesitter vim parser is installed, set the syntax again to get highlighting in the command window',
  group = vim.api.nvim_create_augroup('cmdwin_syntax', {}),
  callback = function ()
    local is_loadable, _ = pcall(vim.treesitter.language.add, 'vim')
    if is_loadable then
      vim.cmd('set syntax=vim')
    end
  end
})
```

I'm going to keep digging into a less hacky patch for both of these

Sam-programs:
How can treesitter be used in the c code?

rudiejd:
@Sam-programs my idea was to turn the cmdline into a buffer, expose that buffer to lua, then start treesitter in the lua code. I haven't gotten very far with it though

Sam-programs:
[...]

PR #30406, opened 2024-09-17
Color cmdline

altermo:
> [!Important]
> This PR requires another PR (https://github.com/neovim/neovim/pull/28412) to be merged before this one (though this can be changed)
> Because of this, the full PR diff is not accurate, only look at the second commit for the true changes

This implements the logic of using treesitter to color the cmdline.

Questions to the maintainers:
+ Should `g:Nvim_color_cmdline` be replaced by a `vim._*` function?
+ In which file should the code live?


runtime/plugin/color_cmdline.lua
```diff
@@ -0,0 +1 @@
+require 'vim.ui.cmdline'
```
clason, line 1: If you do that, you might as well put everything in this file and cut out the middle man...

(Unless this is only temporary and for testing and you'll hook this up properly later.)

clason (changes requested):

runtime/lua/vim/ui/cmdline.lua
```diff
@@ -0,0 +1,102 @@
...
+      b = math.floor((b - a) / 2) + a
+    else
+      a = math.floor((b - a) / 2) + a + 1
+    end
+  end
+
+  return a
+end
+
+vim._cmdline_highlighter = function(cmdline)
```
clason, line 26:
```suggestion
local M = {}
M.highlighter = function(cmdline)
```
It's already exposed as `vim._cmdline`.

justinmk:
btw, as a followup, it would be interesting to see if we could replace the internals of `nvim_parse_expression()` with a treesitter parser. Or at least deprecate it, and provide a documentation hint about how to use treesitter for parsing Vimscript expressions.

luukvbaal:
`g:Nvim_color_cmdline` and its underlying callback mechanism has been slated for removal since it was added in #6364. Promoting it now, after removal was reiterated in #26346 and #28412 seems odd to me.

I can't blame you or core for entertaining this while the default ext_ui plans lie on hold but getting rid of the parallel message grid path still seems more attractive to me.

justinmk:
> Promoting it now, after removal was reiterated in #26346 and #28412 seems odd to me.

@luukvbaal thanks for mentioning that. I forgot about that and/or didn't realize this was conflicting.

If this is conflicting with that work, then yes, let's wait on @luukvbaal 's work.

justinmk (changes requested):
This should wait until (and may be canceled by) @luukvbaal 's work on the messages grid. #27811

justinmk:
color cmdline implemented in https://github.com/neovim/neovim/pull/27855

PR #30409, opened 2024-09-17
runtime:ftplugin: make &keywordprg Nvim compatible

Konfekt:
Since `:!` does not work for terminal output and `:term` is implemented differently, some adaptions are [necessary](https://gist.github.com/Konfekt/8e484af2955a0c7bfe82114df683ce0f/).

### DIFF
--- a/runtime/ftplugin/sudoers.vim
+++ b/runtime/ftplugin/sudoers.vim
@@ -20,7 +20,7 @@ if has('unix') && executable('less')
     command -buffer -nargs=1 SudoersKeywordPrg
           \ silent exe '!' . 'LESS= MANPAGER="less --pattern=''\b' . <q-args> . '\b'' --hilite-search" man ' . 'sudoers' |
           \ redraw!
-  elseif has('terminal') || has('nvim')
+  elseif executable(':terminal') == 2
     command -buffer -nargs=1 SudoersKeywordPrg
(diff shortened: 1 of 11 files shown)


justinmk:
This seems good, but could you see if upstream vim will accept this? After all, it benefits both projects if we avoid drift where possible.

clason:
I would be very surprised if they agreed to include a bunch of `has('nvim')` in their runtime files...

clason:
There'd be [precedent](https://github.com/search?q=repo%3Avim%2Fvim+has%28%27nvim%27%29&type=code) (but then, half of that was a gpanders PR upstreaming from a plugin and they might simply not have noticed...)

runtime/ftplugin/sudoers.vim
```diff
@@ -16,11 +16,11 @@ let b:undo_ftplugin = "setl com< cms< fo<"
 setlocal comments=:# commentstring=#\ %s formatoptions-=t formatoptions+=croql
 
 if has('unix') && executable('less')
-  if !has('gui_running')
+  if !has('gui_running') && !has('nvim')
     command -buffer -nargs=1 SudoersKeywordPrg
           \ silent exe '!' . 'LESS= MANPAGER="less --pattern=''\b' . <q-args> . '\b'' --hilite-search" man ' . 'sudoers' |
           \ redraw!
-  elseif has('terminal')
+  elseif executable(':terminal') == 2
```
justinmk, line 23: this returns 0 on Vim. did you mean `exists(':terminal')` ?

Konfekt: Oh, typo. Thank you

clason:
Included in #30422 -- thanks!

PR #30410, opened 2024-09-17
feat(api): nvim_buf_set_mark can take a window

### LINKED ISSUE #17861: `<C-o>` ignores context marks set via `nvim_buf_set_mark`
### Neovim version (nvim -v)

NVIM v0.7.0-dev+1319-gb81c310d4

N/A

### Operating system/version

Windows 10

### Terminal name/version

N/A

N/A

### Installation

build from repo

Having a buffer with the content:
```
foo
bar
```
After `:clearjumps`, navigate to `r` in `bar`, without jumping. Press ``m` ``. Navigate to
some other position w/o jumping. Pressing `<c-o>` jumps to `r`.

Repeat the same process, but use ``:lua vim.api.nvim_buf_set_mark(0, '`', 2, 4, {})``.

### Expected behavior

`<c-o>` should jump to the position of the `` ` `` mark.

### Actual behavior

Pressing `<c-o>` has no effect. ``` `` ``` works as expected.

muniter:
Problem:
Setting marks "'" and "`" does not work as expected, because the mark is window specific.

Solution:
Pass the window to nvim_buf_set_mark opts argument, and refactor mark.c to use the window instead of curwin.

Fixes #17861


justinmk:
> Setting marks "'" and "`" does not work as expected, because the mark is window specific

Those marks can't be deleted, which is lucky, because `nvim_buf_del_mark` (and `nvim_del_mark`) doesn't take an `opts` param.

But I'd be in favor of defining just 3 new functions, which replace all of these existing functions:

```
nvim_mark_get(..., opts)
nvim_mark_set(..., opts)
nvim_mark_del(..., opts)
```

That's not a blocker here, but just a suggestion.

muniter:
> Those marks can't be deleted, which is lucky, because `nvim_buf_del_mark` (and `nvim_del_mark`) doesn't take an `opts` param.

Neither does `nvim_buf_get_mark` so there's no way to get this marks for an specific window.

Suggestion sounds good I will try it out.

PR #30414, opened 2024-09-18
vim-patch:5036e69: runtime(systemd): allow for overriding systemd ftplugin settings (vim/vim#13373)

Konfekt:
This change in Vim's runtime was not yet included; maybe for a good reason? Then disregard.


clason:
Can you link to the Vim patch? And check with `vim-patch` whether that is included?

PR #30420, opened 2024-09-19
fix(ui): avoid wrong message scroll up

### LINKED ISSUE #30386: Restoring session prints current file name in cmd line with "Press ENTER or type command to continue" prompt
### Problem

When restoring session with `nvim -S` I get annoying prompt to press enter.
It is not happening with version 0.9-x. Only started happening after upgrade to 0.10
My `cmdheight` option is set to 0

### Steps to reproduce

start `nvim -S ` to restore seesion.
The current open file name will be printed in cmd line with the mentioned prompt.

### Expected behavior

No prompt

### Neovim version (nvim -v)

v0.10.1

no

### Operating system/version

arch linux 6.10.9

### Terminal name/version

xfce terminal

xterm-256color

### Installation

arch official repo

glepnir:
Problem: after refactor the message scoll up triggered early.

Solution: Skip scrolling when bottom of screen is reached and
          current characteris newline/carriage return.

Fix #30386

### DIFF
--- a/src/nvim/message.c
+++ b/src/nvim/message.c
@@ -2187,7 +2187,7 @@ static void msg_puts_display(const char *str, int maxlen, int attr, int recurse)
       msg_row = Rows - 1;
 
       // When no more prompt and no more room, truncate here
-      if (msg_no_more && lines_left == 0) {
+      if ((!recurse && *s == '\r') || (msg_no_more && lines_left == 0)) {
         break;
       }
 


Shougo:
This is still WIP?

glepnir:
Kind of forgot about it. Just an attempt to fix issue, message part I'm not very familiar with yet. Does this work?

PR #30422, opened 2024-09-19

clason:
- **vim-patch:c18a9d5: runtime(netrw): using inefficient highlight pattern for 'mf'**
- **vim-patch:41c7bba: runtime(zsh,sh): set and unset compiler in ftplugin**
- **vim-patch:2307945: runtime(java): Optionally recognise all primitive constants in _switch-case_ labels**


zeertzjq:
There is a syntax error in the last commit:
```
line   26:
E580: :endif without :if: endif
```
I'd suggest you move it to a separate PR.

clason:
Added your fix here.

zeertzjq (approved):

PR #30423, opened 2024-09-19
fix: bracketed paste not stored in macro recording

### LINKED ISSUE #28561: bracketed paste not stored in macro recording
### Problem

When I use Neovim in a terminal and copy paste some text in insert mode, I expect the effect to be more or less as if I wrote that text with the keyboard instead of pasting it. When recording a macro, that's not the case: if I record a macro and paste some text (for example to add a line of text) during the recording, then that text is not added when executing the macro.

### Steps to reproduce

- Copy some text to the clipboard
- Start a terminal that supports bracketed paste mode (I tested with gnome-terminal and wezterm)
- Start Neovim (nvim --clean)
- Start a macro: [q][q]
- Start insert mode: [i]
- Paste the text in the terminal with [Ctrl-Shift-V]
- Exit insert mode and stop the macro: [Esc], [q]
- Replay the macro: [@][q]

Result: the text is not added again.

### Expected behavior

The macro should add the text that was copy-pasted during the recording.

### Neovim version (nvim -v)

v0.9.6-dev-18+g0ef27180e

No (Vim 9.1)

### Operating system/version

Debian testing

### Terminal name/version

gnome-terminal 3.50.1

xterm-256color

### Installation

Build from repo (0.9 branch)

jroopam:
Problem: While macro recording the content from the bracketed paste do not reach the register

Solution: Use feed_keys to simulate typing of the pasted content while recording is going on

Fixes [neovim/neovim#28561](https://github.com/neovim/neovim/issues/28561)

### DIFF
--- a/runtime/lua/vim/_editor.lua
+++ b/runtime/lua/vim/_editor.lua
@@ -266,8 +266,26 @@ do
     if undo_started then
       vim.api.nvim_command('undojoin')
     end
+
+    local recording = vim.fn.reg_recording() -- Check if recording is currently happening
+    local function feed_paste_as_input(text, mode)
+      if mode == 'i' then
+        vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(text, true, false, true), 'it', true)
+      else
+        vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(text, true, false, true), 'nt', true)
+        vim.api.nvim_feedkeys('', 'x', true)
+      end
+    end
+
+    local text = table.concat(lines, "\n")
+
     if mode:find('^i') or mode:find('^n?t') then -- Insert mode or Terminal buffer
-      vim.api.nvim_put(lines, 'c', false, true)
+      if recording ~= '' then
+        -- Feed the pasted text as keypresses so it gets recorded
+        feed_paste_as_input(text, 'i')
+      else
+        vim.api.nvim_put(lines, 'c', false, true)
+      end
     elseif phase < 2 and mode:find('^R') and not mode:find('^Rv') then -- Replace mode
       -- TODO: implement Replace mode streamed pasting
       -- TODO: support Virtual Replace mode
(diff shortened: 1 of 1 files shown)


runtime/lua/vim/_editor.lua
```diff
@@ -266,8 +266,26 @@ do
...
+
+    local text = table.concat(lines, "\n")
+
     if mode:find('^i') or mode:find('^n?t') then -- Insert mode or Terminal buffer
-      vim.api.nvim_put(lines, 'c', false, true)
+      if recording ~= '' then
+        -- Feed the pasted text as keypresses so it gets recorded
+        feed_paste_as_input(text, 'i')
+      else
+        vim.api.nvim_put(lines, 'c', false, true)
```
justinmk, line 287: Appreciate your effort here. Would you be up for trying to update `nvim_put` so that it does this internally? That would hopefully avoid the (severe) performance issues of `nvim_feedkeys`.

How does your current PR perform if you paste 100k lines of input? OTOH, I guess macros aren't designed to handle that...

jroopam: Yes, I'd definitely like to try to update `nvim_put` to make this work internally. Just to confirm, when the pasted content reach `nvim_put` they will be added to the macro register(if recording) from there. Not from `vim.paste` as in the current PR.
Also, it would be really helpful if you could point me to the docs that I can refer to(for let's say checking if recording is going on).
Thank you for pointing out for testing it with large inputs. I haven't tried it with large inputs but will surely do while updating it.

justinmk:
> when the pasted content reach `nvim_put` they will be added to the macro register(if recording) from there. Not from `vim.paste`

that sounds correct to me.

> if you could point me to the docs that I can refer to(for let's say checking if recording is going on).


`nvim_put` is implemented here: https://github.com/neovim/neovim/blob/84e85aeb10ae51b03392fdee12e14c9699c99e66/src/nvim/api/vim.c#L1304

`reg_recording` is a global: https://github.com/neovim/neovim/blob/84e85aeb10ae51b03392fdee12e14c9699c99e66/src/nvim/globals.h#L585

zeertzjq:
Maybe it's necessary to introduce `<PasteStart>` and `<PasteEnd>` to indicate the start and end of the pasted text in the register? Similar to what was introduced in [Vim 8.0.0210](https://github.com/vim/vim/commit/ec2da36ca48b40c0654b32a8d2c9f52e796daa5e), but using `nvim_paste` instead of entering Insert mode.

justinmk:
Looks like https://github.com/neovim/neovim/pull/30438 will address this issue, and other improvements. Thanks for getting this started!

Issue #30306, opened 2024-09-08
Configuration of built-in snippets


echasnovski:
### Problem

Currently there is no way to configure built-in `vim.snippet` behavior. This recently became a problem after introduction of not configurable "session only" mappings for snippet navigation (see #30198).

### Expected behavior

Some way for users to configure how certain aspects of snippets expansion is done. As per [this comment](https://github.com/neovim/neovim/issues/30198#issuecomment-2336620609), it might be a good idea to define the scope before deciding on this.

Here are some suggestions:
- Implement `vim.snippet.config()`, similar in spirit to `vim.diagnostic.config()`. Currently this seems like the best (future-proof, already familiar) design. See also [this comment](https://github.com/neovim/neovim/issues/30198#issuecomment-2336624245) for some counter-points.
- Allow `opts` in `vim.snippet.expand()`. This allows configuration on a call level which might be good for plugins that leverage it. However, it also might be too granular to the point of not solving the issue for built-in functionality (see [this comment](https://github.com/neovim/neovim/issues/30198#issuecomment-2336583394)).
- Implement `SnippetEnter` and `SnippetLeave` events (see #26449) for users to be able to perform actions only inside snippet session. Not *exactly* the snippet configuration, but at least *some* way for users to act on snippet expansion.

MariaSolOs:
Since the current snippet settings aren't specific to a session but instead configure the entire snippet expansion logic, using `opts` in `vim.snippet.expand` might have a lot of "global" properties that IMO don't really belong to that function. I would lean towards `vim.snippet.config`, with maybe a `on_expand` callback to include the behavior from the former point.

mfussenegger:
> Currently this seems like the best (future-proof, already familiar) design

I find this assessment a bit odd given the points already made in the linked issue:

- `vim.diagnostic.config` is the exception, not the rule in core. There are global/local buffer/window options for most of the other configurations. `diagnostic.config` is the exception because it required namespace scoping. That's not required for snippets.
- It would require a keymap DSL, which was already identified as sub-optimal in other issues/plugin guide PRs.

> Allow opts in vim.snippet.expand(). This allows configuration on a call level which might be good for plugins that leverage it. However, it also might be too granular to the point of not solving the issue for built-in functionality (see this comment).

[...]

Issue #30198, opened 2024-08-30
Expanding snippet makes custom `<Tab>` and `<S-Tab>` mappings unusable during active snippet session


echasnovski:
### Problem

Expanding snippet with `vim.snippet.expand()` makes user keymaps `<Tab>` and `<S-Tab>` unusable during snippet session without (straightforward) way to opt-out of this behavior.

### Steps to reproduce

1. Create the following 'init.lua' file:
    ```lua
    vim.keymap.set('i', '<Tab>',   [[pumvisible() ? "\<C-n>" : "\<Tab>"]],   { expr = true })
    vim.keymap.set('i', '<S-Tab>', [[pumvisible() ? "\<C-p>" : "\<S-Tab>"]], { expr = true })

    local go_right = function()
      if vim.snippet.active({ direction = 1 }) then return vim.snippet.jump(1) end
      local snippet = 'for k, v in pairs($1) do\n\t$0\nend'
      vim.snippet.expand(snippet)
    end
    vim.keymap.set({ 'i', 's' }, '<C-l>', go_right)
    ```
2. `nvim --clean -u init.lua`
3. Start Insert mode and press `<C-l>` to expand snippet.
4. Press `<C-n>` to show completion popup.
5. Press `<Tab>`. Currently it moves to the next tabstop. When not inside active snippet session, it select next completion item.

------

This seems to be the result of #27339 inside which I didn't find the discussion about the motivation to not recognize user's global `<Tab>` and `<S-Tab>` mappings.

### Expected behavior

`vim.snippet.expand()` keeps custom `<Tab>` / `<S-Tab>` mappings usable. Either by default or by some configurable mechanism. Here are several suggestions:

- Set default snippet navigation keymaps only if it doesn't conflict with what user has already created (either global or buffer-local). This is how I'd expect it to work. For example, default LSP mappings behave the same way: [explicitly with `K`](https://github.com/neovim/neovim/blob/5f95f1249f464e4f0ceed468ec5a1ba6e810da14/runtime/lua/vim/lsp.lua#L365) and [implicitly with other defaults](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/_defaults.lua) (because they are created early and globally while user can override them inside config).
- Allow `opts` in `vim.snippet.expand()` with fields `lhs_jump_forward` and `lhs_jump_backward` to customize which keys are used as default for navigating left and right. Preferably with an ability to supply empty string to not create any.
[...]

PR #30430, opened 2024-09-20
fix(snippet): don't auto map tab and s-tab when already have define

glepnir:
Problem:
Currently, `vim.snippet.expand` automatically creates mappings for `tab` and `shift-tab`. This is done as part of our efforts to provide some default keybindings. However, silently overriding users' existing tab and shift-tab mappings is not a good approach, especially since this is not even documented. There are too many customization possibilities with mappings. The default behavior should inject mappings only when they don't already exist, rather than overwriting them. I have checked #30306 and previous discussions, but I still believe that overriding users' mappings can cause confusion.

Solution:
Add mapping checks to ensure that mappings are injected only if they do not already exist.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -4368,7 +4368,10 @@ vim.snippet.active({filter})                            *vim.snippet.active()*
 vim.snippet.expand({input})                             *vim.snippet.expand()*
     Expands the given snippet text. Refer to
     https://microsoft.github.io/language-server-protocol/specification/#snippet_syntax
-    for the specification of valid input.
+    for the specification of valid input. If no key mappings for snippet jump
+    (tab/shift-tab) are detected, this function will automatically add
+    mappings for 'tab' and 'shift-tab' in both insert ('i') and select ('s')
+    modes. restore after leave snippet.
 
     Tabstops are highlighted with |hl-SnippetTabstop|.
 
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -224,7 +224,7 @@ end
 function Session:set_keymaps()
   local function maparg(key, mode)
     local map = vim.fn.maparg(key, mode, false, true) --[[ @as table ]]
-    if not vim.tbl_isempty(map) and map.buffer == 1 then
+    if not vim.tbl_isempty(map) then
       return map
     else
       return nil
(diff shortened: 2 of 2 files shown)


clason:
This does not address all the problems brought up in the linked issue. I don't think simply slapping such a check in solves the problem.

glepnir:
This does not solve that issue. I have stated the intent.  question is why silently overwrite the existing user mapping?

clason:
Please see https://github.com/neovim/neovim/issues/30198.

glepnir:
> I have checked https://github.com/neovim/neovim/issues/30306 and previous discussions,

 i think i already checked. but i don't feel like explains the behavior of forcibly overwriting the user mapping.

clason:
To allow -- for now -- coexistence with other snippet plugins. I'm not saying that this is not an issue that need not be solved; I'm saying that this is not the correct solution for it.

glepnir:
again PR is not created to solve that issue. Instead, is it possible to check if the user is already mapped? ? please read commit message carefully..

clason:
I have read it, and that exact issue has been discussed in the linked (closed) issue: There are situations where it's _desired_ to temporarily(!) override an existing mapping in the snippet context. The correct solution is to make that behavior opt-in/out, not remove it entirely.

glepnir:
> There are situations where it's desired to temporarily(!) override an existing mapping

Do you mean this includes breaking the default tab on insert behavior? In the case of indentation that requires a tab, it will jump to the snippet. obviously, this is unreasonable. default tab is changed by this override.??!

clason:
Of course not. Any snippet "tab" mapping will check for a valid placeholder first and fall back to literal `<tab>` otherwise. This is unrelated to whether you set it by default or not.

Again, I am in favor of making it configurable, but not of simply removing it.

glepnir:
Not quite. For example, `namespace $1 { \n $0 }` The second tab to $0 I continue tabbing it will just jump to $1 because $1 is still valid. You can't check like this. First of all, it shouldn't be circular. Second, it shouldn't override any expected user behavior. Because I can handle this situation myself.

clason:
That sounds like a bug that should be fixed; reaching `$0` should exit the snippet.

glepnir:
$0 is the last one section which defined in lsp spec. But this behavior is completely configurable by the user. But why override it? And no we have no documentation to mark this. I can't understand overriding this behavior. This infringes on the user's space.

PR #30431, opened 2024-09-20
fix(snippet): stop jump when current index is final one

glepnir:
Problem: The current snippet jump is circular but as lsp spec mentioned
                 the `$0` is final snippet section should stop when at it

Solution: when current tab index is 0 stopped.

Ref https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionClientCapabilities


runtime/lua/vim/snippet.lua
```diff
@@ -674,7 +674,9 @@ end
 --- @return boolean
 function M.active(filter)
   local active = M._session ~= nil and M._session.bufnr == vim.api.nvim_get_current_buf()
-
+  if M._session.current_tabstop and M._session.current_tabstop.index == 0 then
```
justinmk, line 677: when we have a reference, it's always worth a brief comment to provide that. it's a cheap way to help people reading this code in the future.

```suggestion
  -- "$0 defines the final tab stop".
  -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionClientCapabilities
  if M._session.current_tabstop and M._session.current_tabstop.index == 0 then
```

runtime/lua/vim/snippet.lua
```diff
@@ -674,7 +674,9 @@ end
...
+    return active and false
```
justinmk, line 678: `active and false` will always return falsy. 
```suggestion
    return false
```

runtime/lua/vim/snippet.lua
glepnir: because its too long.. exceeds the 80 column width 🤔

glepnir:
It is not. Just try a snippet. It is circular. the test which i added in this Pr will be failed in master.
what

clason:
I cannot replicate that. Are you sure you're using the built-in LSP snippet expansion? Please test with `nvim --clean`.

glepnir:
I am sure it is but it happen when i use this code snippet , <C-X><C-O> works fine with it.

```lua
local function auto_trigger(bufnr, client)
  au(InsertCharPre, {
    buffer = bufnr,
    callback = function()
      if pumvisible() then
        return
      end
        client,
        'server_capabilities',
        'completionProvider',
        'triggerCharacters'
      ) or {}
      if vim.v.char:match('[%w_]') and not vim.list_contains(triggerchars, vim.v.char) then
        completion.trigger()
      end
    end,
  })
end
au('LspAttach', {
  callback = function(args)
    local bufnr = args.buf
    local client_id = args.data.client_id
    completion.enable(true, client_id, bufnr, {
      autotrigger = true,
      convert = function(item)
        return { abbr = item.label:gsub('%b()', ''), kind = '' }
      end,
    })
    local client = lsp.get_client_by_id(client_id)
    if not client then
      return
    end
    auto_trigger(bufnr, client)
  end,
})

```

PR #30435, opened 2024-09-20
refactor: rename "Dictionary" => "Dict"

justinmk:
This has long been on the todo list.

In the api_info() output:

    :new|put =map(filter(api_info().functions, '!has_key(v:val,''deprecated_since'')'), 'v:val')
    ...

    {'return_type': 'ArrayOf(Integer, 2)', 'name': 'nvim_win_get_position', 'method': v:true, 'parameters': [['Window', 'window']], 'since': 1}

The `ArrayOf(Integer, 2)` return type didn't break clients when we added it, which is evidence that clients don't use the `return_type` field, thus renaming Dictionary => Dict in api_info() is not (in practice) a breaking change.


ribru17:
Would it make sense to add this commit to the [rev ignore list](https://github.com/neovim/neovim/blob/100bb15d186ce855d064e86ff4630d9ad8d43a5e/.git-blame-ignore-revs#L10)?

src/nvim/generators/gen_api_ui_events.lua
```diff
@@ -205,7 +205,8 @@ for _, ev in ipairs(events) do
     ev_exported[attr] = ev[attr]
   end
   for _, p in ipairs(ev_exported.parameters) do
-    if p[1] == 'HlAttrs' then
+    if p[1] == 'HlAttrs' or p[1] == 'Dict' then
+      -- TODO(justinmk): for back-compat, but do clients actually look at this?
       p[1] = 'Dictionary'
     end
```
justinmk, line 211: For now I kept the ui_events parameter type the same, since it's possible that clients use that(?).

test/functional/api/version_spec.lua
```diff
@@ -58,10 +58,18 @@ describe('api metadata', function()
...
+  local function normalize_func_metadata(f)
+    -- Dictionary was renamed to Dict. That doesn't break back-compat because clients don't actually
+    -- use the `return_type` field (evidence: "ArrayOf(…)" didn't break clients).
+    f.return_type = f.return_type:gsub('Dictionary', 'Dict')
+
     f.deprecated_since = nil
     for idx, _ in ipairs(f.parameters) do
+      -- Dictionary was renamed to Dict. Doesn't break back-compat because clients don't actually
+      -- use the `parameters` field of API metadata (evidence: "ArrayOf(…)" didn't break clients).
+      f.parameters[idx][1] = f.parameters[idx][1]:gsub('Dictionary', 'Dict')
```
justinmk, line 71: I suspect this won't break any clients to return the `return_type` in the `api_info` metadata. We already report `ArrayOf(..)` and similar stuff in `return_type`, and that didn't break clients.

I think clients are using the types from the actual RPC calls, not api-metadata. There might be some clients that are generating code from api-metadata, and this would be an easy fix for them (that's not really "breaking" if it's a build-time change).

justinmk:
> Would it make sense to add this commit to the [rev ignore list](https://github.com/neovim/neovim/blob/100bb15d186ce855d064e86ff4630d9ad8d43a5e/.git-blame-ignore-revs#L10)?

will wait until this bakes for a bit

runtime/doc/news.txt
```diff
@@ -35,6 +35,13 @@ API
   now maintain their own list of channels.
   • In the future, |vim.rpcnotify()| may accept a list of channels, if there
     is demand for this use-case.
+• "Dictionary" was renamed to "Dict" internally and in the RPC |api-metadata|.
+  This is not expected to break clients because there are no known clients
+  that actually use the `return_type` field or the parameter type names
+  reported by |--api-info| or |nvim_get_api_info()|.
+• Renamed `nvim__id_dictionary` (unsupported/experimental API) to
+  `nvim__id_dict`.
+
```
zeertzjq, line 44:
```suggestion
```

PR #30438, opened 2024-09-20
fix(paste): improve repeating of pasted text

### LINKED ISSUE #28561: bracketed paste not stored in macro recording

zeertzjq:
- Fixes 'autoindent' being applied during redo.
- Makes redoing a large paste significantly faster.
- Stores pasted text in the register being recorded.

Fix #28561

### DIFF
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -1259,30 +1259,19 @@ Boolean nvim_paste(String data, Boolean crlf, Integer phase, Arena *arena, Error
     draining = true;
     goto theend;
   }
-  if (!(State & (MODE_CMDLINE | MODE_INSERT)) && (phase == -1 || phase == 1)) {
-    ResetRedobuff();
-    AppendCharToRedobuff('a');  // Dot-repeat.
+  if (phase == -1 || phase == 1) {
+    paste_store(kFalse, NULL_STRING, crlf);
   }
   // vim.paste() decides if client should cancel.  Errors do NOT cancel: we
   // want to drain remaining chunks (rather than divert them to main input).
   cancel = (rv.type == kObjectTypeBoolean && !rv.data.boolean);
-  if (!cancel && !(State & MODE_CMDLINE)) {  // Dot-repeat.
-    for (size_t i = 0; i < lines.size; i++) {
-      String s = lines.items[i].data.string;
-      assert(s.size <= INT_MAX);
-      AppendToRedobuffLit(s.data, (int)s.size);
-      // readfile()-style: "\n" is indicated by presence of N+1 item.
-      if (i + 1 < lines.size) {
-        AppendCharToRedobuff(NL);
-      }
-    }
-  }
-  if (!(State & (MODE_CMDLINE | MODE_INSERT)) && (phase == -1 || phase == 3)) {
-    AppendCharToRedobuff(ESC);  // Dot-repeat.
+  if (!cancel) {
+    paste_store(kNone, data, crlf);
   }
 theend:
   if (cancel || phase == -1 || phase == 3) {  // End of paste-stream.
     draining = false;
+    paste_store(kTrue, NULL_STRING, crlf);
   }
 
   return !cancel;
--- a/src/nvim/getchar.c
+++ b/src/nvim/getchar.c
@@ -13,6 +13,7 @@
 
 #include "nvim/api/private/defs.h"
 #include "nvim/api/private/helpers.h"
+#include "nvim/api/vim.h"
 #include "nvim/ascii_defs.h"
 #include "nvim/buffer_defs.h"
 #include "nvim/charset.h"
@@ -308,6 +309,24 @@ static void add_num_buff(buffheader_T *buf, int n)
   add_buff(buf, number, -1);
 }
 
+/// Add byte or special key 'c' to buffer "buf".
+/// Translates special keys, NUL and K_SPECIAL.
+static void add_byte_buff(buffheader_T *buf, int c)
+{
+  char temp[4];
+  if (IS_SPECIAL(c) || c == K_SPECIAL || c == NUL) {
+    // Translate special key code into three byte sequence.
+    temp[0] = (char)K_SPECIAL;
+    temp[1] = (char)K_SECOND(c);
+    temp[2] = (char)K_THIRD(c);
+    temp[3] = NUL;
+  } else {
+    temp[0] = (char)c;
+    temp[1] = NUL;
+  }
+  add_buff(buf, temp, -1);
+}
+
 /// Add character 'c' to buffer "buf".
 /// Translates special keys, NUL, K_SPECIAL and multibyte characters.
 static void add_char_buff(buffheader_T *buf, int c)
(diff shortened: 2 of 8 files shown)


zeertzjq:
This PR makes redoing a large paste significantly faster. With the same build type and a 15MB log file:
- Before this PR, pasting the file takes several seconds, while redoing the paste takes over a minute.
- After this PR, both pasting the file and redoing the paste takes several seconds.

src/nvim/getchar.c
```diff
@@ -3182,3 +3189,121 @@ bool map_execute_lua(bool may_repeat)
...
 }
+
+/// Appends pasted text to redo buffer and/or record buffer if needed.
+/// Escapes all K_SPECIAL and NUL bytes in the content.
+///
+/// @param state  kFalse for the start of a paste
+///               kTrue for the end of a paste
+///               kNone for the content of a paste
+/// @param str    the content of the paste (only used when state is kNone)
+void store_pasted_text(const TriState state, const String str, const bool crlf)
```
justinmk, line 3200: is it worth using a common prefix?
```suggestion
void paste_store(const TriState state, const String str, const bool crlf)
```

src/nvim/getchar.c
```diff
@@ -3182,3 +3189,121 @@ bool map_execute_lua(bool may_repeat)
...
+      }
+      if (need_record) {
+        add_byte_buff(&recordbuff, c);
+      }
+    }
+  }
+}
+
+/// Gets a paste stored by store_pasted_text() from typeahead and repeats it.
+void repeat_pasted_text(int count)
```
justinmk, line 3265:
```suggestion
void paste_repeat(int count)
```

src/nvim/getchar.c
zeertzjq: Yes

src/nvim/api/vim.c
```diff
@@ -1259,30 +1259,19 @@ Boolean nvim_paste(String data, Boolean crlf, Integer phase, Arena *arena, Error
...
-      // readfile()-style: "\n" is indicated by presence of N+1 item.
-      if (i + 1 < lines.size) {
-        AppendCharToRedobuff(NL);
-      }
-    }
-  }
-  if (!(State & (MODE_CMDLINE | MODE_INSERT)) && (phase == -1 || phase == 3)) {
-    AppendCharToRedobuff(ESC);  // Dot-repeat.
+  if (!cancel) {
+    paste_store(kNone, data, crlf);
```
justinmk, line 1269: the `AppendCharToRedobuff` hacks were already attempting to "fixup" the input buffer, so arguably `paste_store` is just formalizing that. 👍

src/nvim/getchar.c
```diff
@@ -3182,3 +3189,121 @@ bool map_execute_lua(bool may_repeat)
...
+
+  got_int = false;
+  while (!aborted) {
+    ga_grow(&ga, 32);
+    uint8_t c1 = (uint8_t)vgetorpeek(true);
+    if (c1 == K_SPECIAL) {
+      c1 = (uint8_t)vgetorpeek(true);
+      uint8_t c2 = (uint8_t)vgetorpeek(true);
+      int c = TO_SPECIAL(c1, c2);
+      if (c == K_PE) {
```
justinmk, line 3280:
```suggestion
      if (c == K_PE) {  // <PasteEnd>
```

src/nvim/getchar.c
```diff
@@ -3182,3 +3189,121 @@ bool map_execute_lua(bool may_repeat)
   ga_clear(&line_ga);
   return true;
 }
+
+/// Appends pasted text to redo buffer and/or record buffer if needed.
```
justinmk, line 3193:
```suggestion
/// Wraps pasted text stream with <PasteStart> and <PasteEnd> and appends to redo buffer and/or record buffer if needed.
```

src/nvim/normal.c
```diff
@@ -6593,6 +6594,11 @@ static void nv_open(cmdarg_T *cap)
   }
 }
 
+static void nv_paste(cmdarg_T *cap)
```
justinmk, line 6597: can we give a hint that mentions `<PasteStart>`. 

```suggestion
/// Handles <PasteStart>, repeats pasted text.
static void nv_paste(cmdarg_T *cap)
```

src/nvim/keycodes.c
```diff
@@ -184,6 +184,9 @@ static const struct key_name_entry {
   { K_KRIGHT,          "kRight" },
   { K_KRIGHT,          "KP6" },
 
+  { K_PS,              "PasteStart" },
```
justinmk, line 187: can we leave a hint about the intention of these keycodes, to emphasize that they are internal-only and not to be sent directly by UIs
```suggestion
  // Delimits pasted text (input from nvim_paste). Internal-only, not directly sent by UIs.
  { K_PS,              "PasteStart" },
```

justinmk (commented):
Is there a way we can add a guardrail to prevent `<pastestart>` from being sent (perhaps accidentally) by `nvim_input` / feedkeys? E.g. show an error.

src/nvim/keycodes.c
zeertzjq: I can just remove these notations.

src/nvim/getchar.c
```diff
@@ -3182,3 +3189,121 @@ bool map_execute_lua(bool may_repeat)
...
+
+  const bool need_redo = !block_redo;
+  const bool need_record = reg_recording != 0 && reg_executing == 0;
+
+  if (!need_redo && !need_record) {
+    return;
+  }
+
+  if (state != kNone) {
+    const int c = state == kFalse ? K_PS : K_PE;
```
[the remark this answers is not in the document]
zeertzjq: I've removes these two notations. Maybe I should remove rename them `K_PASTE_START` and `K_PASTE_END`?

justinmk:
I added this to news.txt in https://github.com/neovim/neovim/pull/30494 because this is a very nice improvement, even if it's technically a "bug fix".

PR #30439, opened 2024-09-20
build(deps): bump wasmtime to v25.0.1

clason:
https://github.com/bytecodealliance/wasmtime/releases/tag/v25.0.0
https://github.com/bytecodealliance/wasmtime/releases/tag/v25.0.1


dundargoc:
Should we change exact wasmtime version requirement to minimum mayhaps?

clason:
> Should we change exact wasmtime version requirement to minimum mayhaps?

A big part of the point of my proactive bump PRs is exactly to find out whether that is safe. (I'll remember to bump these numbers one of these days; the script could also be smarter.)

dundargoc (approved):

clason:
> Should we change exact wasmtime version requirement to minimum mayhaps?

Since we now use tree-sitter's cmake file, which doesn't specify a version at all, that may indeed be reasonable.

PR #30454, opened 2024-09-21
docs(api): `nvim_get_runtime_file()` preserves 'runtimepath' order

echasnovski:
Also:
- Add test that 'runtimepath' order is preserved.
- Tweak example of getting all color scheme files, as '*.lua' files are also allowed.


test/functional/api/vim_spec.lua
```diff
@@ -3253,6 +3253,16 @@ describe('API', function()
         exc_exec("echo nvim_get_runtime_file('{', v:false)")
       )
     end)
+    it('preserves order of runtimpath', function()
```
zeertzjq, line 3256:
```suggestion
    it('preserves order of runtimepath', function()
```

runtime/doc/api.txt
```diff
@@ -957,8 +957,8 @@ nvim_get_runtime_file({name}, {all})                 *nvim_get_runtime_file()*
     Find files in runtime directories
```
justinmk, line 957: important enough to call out here?

```suggestion
    Find files in runtime directories, in 'runtimepath' order.
```

justinmk (approved):

test/functional/api/vim_spec.lua
echasnovski: Thanks! Fixed.

justinmk (approved):

PR #30456, opened 2024-09-21
fix(api): deprecate nvim_subscribe, nvim_unsubscribe

justinmk:
## Problem:
- nvim_subscribe, nvim_unsubscribe were deprecated in   aec4938a21a02d279d13a9eb64ef3b7cc592c374 but this wasn't set in the  API metadata.
- The function annotations
  ```
  FUNC_API_SINCE(1) FUNC_API_REMOTE_ONLY FUNC_API_DEPRECATED_SINCE(13)
  ```
  cause this test to fail:
   ```
   RUN T3 api metadata functions are compatible with old metadata or have new level: 3.00 ms ERR
   test/functional/api/version_spec.lua:135: function vim_subscribe was removed but exists in level 0 which nvim should be compatible with
   stack traceback:
     test/functional/api/version_spec.lua:135: in function <test/functional/api/version_spec.lua:128>
   ```

## Solution:
- Set the API metadata.
- Rearrange the annotations so that FUNC_API_DEPRECATED_SINCE is 2nd:
  ```
  FUNC_API_SINCE(1) FUNC_API_DEPRECATED_SINCE(13) FUNC_API_REMOTE_ONLY
  ```

### DIFF
--- a/src/nvim/api/deprecated.c
+++ b/src/nvim/api/deprecated.c
@@ -798,7 +798,8 @@ theend:
 /// @param channel_id Channel id (passed automatically by the dispatcher)
 /// @param event      Event type string
 void nvim_subscribe(uint64_t channel_id, String event)
-  FUNC_API_SINCE(1) FUNC_API_REMOTE_ONLY
+// XXX: c_grammar.lua is order-sensitive.
+  FUNC_API_SINCE(1) FUNC_API_DEPRECATED_SINCE(13) FUNC_API_REMOTE_ONLY
 {
   // Does nothing. `rpcnotify(0,…)` broadcasts to all channels, there are no "subscriptions".
 }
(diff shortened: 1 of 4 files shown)


src/nvim/api/deprecated.c
```diff
@@ -792,7 +792,8 @@ Array nvim_call_atomic(uint64_t channel_id, Array calls, Arena *arena, Error *er
 /// @param channel_id Channel id (passed automatically by the dispatcher)
 /// @param event      Event type string
 void nvim_subscribe(uint64_t channel_id, String event)
-  FUNC_API_SINCE(1) FUNC_API_REMOTE_ONLY
+// XXX: c_grammar.lua is order sensitive? FUNC_API_SINCE FUNC_API_DEPRECATED_SINCE must go first.
+  FUNC_API_SINCE(1) FUNC_API_DEPRECATED_SINCE(13) FUNC_API_REMOTE_ONLY
```
justinmk, line 796: @bfredl is it expected that `FUNC_API_SINCE FUNC_API_DEPRECATED_SINCE ` need to appear before other annotations? I didn't get very trying to make c_grammar order-insensitive.

scripts/gen_eval_files.lua
```diff
@@ -243,7 +243,8 @@ local function get_api_meta()
   end
 
   for _, fun in pairs(functions) do
-    local deprecated = fun.deprecated_since ~= nil
+    -- TODO(justinmk): why isn't `fun.deprecated_since` set for nvim_subscribe?
+    local deprecated = fun.deprecated_since ~= nil or fun.deprecated
```
justinmk, line 247: Without this change, `@deprecated` is not added to the `nvim_subscribe` doc in `runtime/lua/vim/_meta/api.lua`. Why isn't `fun.deprecated_since` set for `nvim_subscribe`?

PR #30458, opened 2024-09-21
build: bump minimum cmake version to 3.16

dundargoc:


clason:
What about the improvements to our scripts that 3.16 unlock?

BUILD.md
```diff
@@ -283,7 +283,7 @@ cmake --build build
 General requirements (see [#1469](https://github.com/neovim/neovim/issues/1469#issuecomment-63058312)):
 
 - Clang or GCC version 4.9+
-- CMake version 3.13+, built with TLS/SSL support
+- CMake version 3.16+, built with TLS/SSL support
   - Optional: Get the latest CMake from an [installer](https://github.com/Kitware/CMake/releases) or the [Python package](https://pypi.org/project/cmake/) (`pip install cmake`)
```
justinmk, line 287: maybe https://cmake.org/download/ is a better link. it also mentions pip, snap, and ubuntu apt repo.

dundargoc: Added to https://github.com/neovim/neovim/pull/30177

dundargoc:
> What about the improvements to our scripts that 3.16 unlock?

I did remove a bit of code. On the other end we already used `--install` in the Makefile, just that it would not have worked for old cmake's so we were already assuming later cmake versions. As for the rest, I figured we'll use them as needed.

dundargoc:
I guess can reduce the log level on local builds? Unsure if that's wanted.

PR #30460, opened 2024-09-21
fix(lsp): handle multiline signature help labels

### LINKED ISSUE #30428: LSP: current parameter in signature help not highlighted if the label spans multiple lines
### Problem

calling signature help in a function whos signature is on more than 1 line, the current parameter is not highlighted. im using zig/zls as an example, im not sure what other servers support signature help
image
image

### Steps to reproduce

```sh
cat >test.zig <<EOF
pub fn main() void {
    foo({}, );
    bar({}, );
}
fn foo(_: void, _: void) void {}
fn bar(
    _: void,
    _: void,
) void {}
EOF
cat >repro.vim <<EOF
autocmd FileType zig lua vim.lsp.start({cmd = {'zls'}})
EOF
nvim --clean -u repro.vim test.zig
```
move to the call sites and run `:lua vim.lsp.buf.signature_help()`

### Expected behavior

i expect the parameters to be highlighted in all cases

### Neovim version (nvim -v)

v0.11.0-dev-794+gf01c764cc

n/a (?)

### Operating system/version

alpine linux kernel 6.11.0

### Terminal name/version

ghostty

xterm-ghostty

### Installation

built from source

MariaSolOs:
Fixes #30428

For reference, this is how it looks on `master`:
image

Versus with this PR:
image

### DIFF
--- a/runtime/lua/vim/lsp/handlers.lua
+++ b/runtime/lua/vim/lsp/handlers.lua
@@ -462,6 +462,8 @@ M[ms.textDocument_typeDefinition] = location_handler
 --- @see # https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_implementation
 M[ms.textDocument_implementation] = location_handler
 
+local sig_help_ns = api.nvim_create_namespace('vim_lsp_signature_help')
+
 --- |lsp-handler| for the method "textDocument/signatureHelp".
 ---
 --- The active parameter is highlighted with |hl-LspSignatureActiveParameter|.
@@ -476,7 +478,7 @@ M[ms.textDocument_implementation] = location_handler
 --- ```
 ---
 ---@param _ lsp.ResponseError?
----@param result lsp.SignatureHelp  Response from the language server
+---@param result lsp.SignatureHelp? Response from the language server
 ---@param ctx lsp.HandlerContext Client context
 ---@param config table Configuration table.
 ---     - border:     (default=nil)
@@ -509,10 +511,15 @@ function M.signature_help(_, result, ctx, config)
     return
   end
   local fbuf, fwin = util.open_floating_preview(lines, 'markdown', config)
+  -- Highlight the active parameter.
   if hl then
-    -- Highlight the second line if the signature is wrapped in a Markdown code block.
-    local line = vim.startswith(lines[1], '```') and 1 or 0
-    api.nvim_buf_add_highlight(fbuf, -1, 'LspSignatureActiveParameter', line, unpack(hl))
+    vim.highlight.range(
+      fbuf,
+      sig_help_ns,
+      'LspSignatureActiveParameter',
+      { hl[1], hl[2] },
+      { hl[3], hl[4] }
+    )
   end
   return fbuf, fwin
 end
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -781,24 +781,37 @@ function M.convert_input_to_markdown_lines(input, contents)
   return contents
 end
 
+--- Returns the line/column-based position in `contents` at the given offset.
+---
+---@param offset integer
+---@param contents string[]
+---@return { [1]: integer, [2]: integer }
+local function get_pos_from_offset(offset, contents)
+  local i = 0
+  for l, line in ipairs(contents) do
+    if offset >= i and offset < i + #line then
+      return { l - 1, offset - i + 1 }
+    else
+      i = i + #line + 1
+    end
+  end
+end
+
 --- Converts `textDocument/signatureHelp` response to markdown lines.
 ---
 ---@param signature_help lsp.SignatureHelp Response of `textDocument/SignatureHelp`
 ---@param ft string|nil filetype that will be use as the `lang` for the label markdown code block
 ---@param triggers table|nil list of trigger characters from the lsp server. used to better determine parameter offsets
 ---@return string[]|nil table list of lines of converted markdown.
----@return number[]|nil table of active hl
+---@return Range4|nil table of active hl
 ---@see https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_signatureHelp
 function M.convert_signature_help_to_markdown_lines(signature_help, ft, triggers)
-  if not signature_help.signatures then
-    return
-  end
   --The active signature. If omitted or the value lies outside the range of
   --`signatures` the value defaults to zero or is ignored if `signatures.length == 0`.
   --Whenever possible implementors should make an active decision about
   --the active signature and shouldn't rely on a default value.
   local contents = {}
-  local active_hl
+  local active_offset ---@type [integer, integer]|nil
   local active_signature = signature_help.activeSignature or 0
   -- If the activeSignature is not inside the valid range, then clip it.
   -- In 3.15 of the protocol, activeSignature was allowed to be negative
@@ -806,9 +819,6 @@ function M.convert_signature_help_to_markdown_lines(signature_help, ft, triggers
     active_signature = 0
   end
   local signature = signature_help.signatures[active_signature + 1]
-  if not signature then
-    return
-  end
   local label = signature.label
   if ft then
     -- wrap inside a code block for proper rendering
@@ -825,6 +835,8 @@ function M.convert_signature_help_to_markdown_lines(signature_help, ft, triggers
     M.convert_input_to_markdown_lines(signature.documentation, contents)
   end
   if signature.parameters and #signature.parameters > 0 then
+    -- First check if the signature has an activeParameter. If it doesn't check if the response
+    -- had that property instead. Else just default to 0.
     local active_parameter = (signature.activeParameter or signature_help.activeParameter or 0)
     if active_parameter < 0 then
       active_parameter = 0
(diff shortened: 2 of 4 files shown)


MariaSolOs:
Hmm, I guess I can use `highlight.range` here...

runtime/lua/vim/lsp/util.lua
```diff
@@ -768,34 +768,42 @@ function M.convert_input_to_markdown_lines(input, contents)
   return contents
 end
 
+---@param offset integer
+---@param contents string[]
+---@return Range2
+local function get_line_and_col_from_offset(offset, contents)
```
justinmk, line 774:
```suggestion
local function get_pos_from_offset(offset, contents)
```

p.s.: another case for https://github.com/neovim/neovim/issues/25509 , probably

runtime/lua/vim/lsp/util.lua
```diff
@@ -843,36 +853,47 @@ function M.convert_signature_help_to_markdown_lines(signature_help, ft, triggers
         documentation?: string | MarkupContent;
       }
       --]=]
```
justinmk, line 855: would the above comment benefit from a "before => after" example to show what's going on here?

justinmk (approved):

mfussenegger:
Would it be possible to add a test case for this?

runtime/lua/vim/lsp/util.lua
MariaSolOs: I'm not sure I understand. What exactly would you like me to clarify further?

runtime/lua/vim/lsp/util.lua
```diff
@@ -781,34 +781,44 @@ function M.convert_input_to_markdown_lines(input, contents)
...
+end
+
 --- Converts `textDocument/signatureHelp` response to markdown lines.
 ---
 ---@param signature_help lsp.SignatureHelp Response of `textDocument/SignatureHelp`
 ---@param ft string|nil filetype that will be use as the `lang` for the label markdown code block
 ---@param triggers table|nil list of trigger characters from the lsp server. used to better determine parameter offsets
 ---@return string[]|nil table list of lines of converted markdown.
----@return number[]|nil table of active hl
+---@return Range4|nil table of active hl
```
wookayin, line 806: is this correct? `active_hl` and `get_pos_from_offset` appear to have two numbers (actually not a Range2), not Range4.

MariaSolOs: It is correct. Notice how `active_hl` is extended twice with 2 `Range2`s here:
https://github.com/neovim/neovim/blob/6e01e0f7b60f5afe802c6feb52b7f7d5fb15cde7/runtime/lua/vim/lsp/util.lua#L908-L912

wookayin: I see, now I get how it contains four elements. It looked like `active_hl` is `Range4[]` at a glance. Though I believe making it explicit (e.g. `row_start, col_start`, `row_end, `col_end`) would make the code a bit easier to read.

runtime/lua/vim/lsp/util.lua
```diff
@@ -781,34 +781,44 @@ function M.convert_input_to_markdown_lines(input, contents)
   return contents
 end
 
+--- Returns the line/column-based position in `contents` at the given offset.
+---
+---@param offset integer
+---@param contents string[]
+---@return Range2
```
wookayin, line 788: Coming from https://github.com/neovim/neovim/pull/30460#discussion_r1798438892, this seems to be fixed:

```suggestion
---@return Range2
+++@return {[1]: integer, [2]: integer} row, col
```

Because Range2 means (start row, start end), not a position.

note: since this is just a local helper function, we can also simply return two integers rather than a table.

runtime/lua/vim/lsp/util.lua
MariaSolOs: I agree that it would make the code easier to read when not inside an editor, but what you described is exactly the definition of `Range4`:
```lua
---@class Range4
---@inlinedoc
---@field [1] integer start row
---@field [2] integer start column
---@field [3] integer end row
---@field [4] integer end column
```

I prefer to use the existing type here, and either way we do want to consolidate the many position/range abstractions we have (see https://github.com/neovim/neovim/issues/25509).

runtime/lua/vim/lsp/util.lua
MariaSolOs: I'll keep the table so that I can still just use `list_extend` without introducing extra variables at the call sites.

runtime/lua/vim/lsp/util.lua
wookayin: Yes, I agree. Just wanted to be clear, I meant using local variables while constructing a table (rather than using list_extend which is not type friendly).

MariaSolOs: I don't get any type errors with the current approach...

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -462,6 +462,8 @@ M[ms.textDocument_typeDefinition] = location_handler
 --- @see # https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_implementation
 M[ms.textDocument_implementation] = location_handler
 
+local sig_help_ns = api.nvim_create_namespace('lsp_signature_help')
```
justinmk, line 465: nit/non-blocker: do we usually prefix our namespaces with `nvim_` ? 

also wondering if we should use dot "." as the separator for namesapces, e.g. `nvim.lsp.foo`.

MariaSolOs: I don't think we have a convention regarding namespace names. E.g. we use [`'vim/snippet'`](https://github.com/neovim/neovim/blob/36890824c7edb7bc5458062c14efbfa0b1d280b5/runtime/lua/vim/snippet.lua#L3) in `vim.snippet`, [`'ts.dev'`](https://github.com/neovim/neovim/blob/36890824c7edb7bc5458062c14efbfa0b1d280b5/runtime/lua/vim/treesitter/dev.lua#L138) in the TS inspector, etc.

We do seem to prefix these with `vim_` in the LSP namespaces though, so I'll stick to that.

MariaSolOs:
@justinmk @mfussenegger I think this is ready to go :)

runtime/lua/vim/lsp/util.lua
```diff
@@ -781,34 +781,44 @@ function M.convert_input_to_markdown_lines(input, contents)
...
-  if not signature_help.signatures then
-    return
-  end
   --The active signature. If omitted or the value lies outside the range of
   --`signatures` the value defaults to zero or is ignored if `signatures.length == 0`.
   --Whenever possible implementors should make an active decision about
   --the active signature and shouldn't rely on a default value.
   local contents = {}
-  local active_hl
+  local active_offset ---@type [number, number]|nil
```
wookayin, line 814: nit:
```suggestion
  local active_offset ---@type [integer, integer]|nil
```

test/functional/plugin/lsp_spec.lua
```diff
@@ -3519,6 +3519,27 @@ describe('LSP', function()
       local expected = { '```cs', 'TestEntity.TestEntity()', '```', 'some doc' }
       eq(expected, result)
     end)
+
+    it('higlights active parameters in multiline signature labels', function()
```
wookayin, line 3523:
```suggestion
    it('highlights active parameters in multiline signature labels', function()
```

test/functional/plugin/lsp_spec.lua
```diff
@@ -3519,6 +3519,27 @@ describe('LSP', function()
...
+              parameters = {
+                { label = '_: void' },
+                { label = '_: void' },
+              },
+            },
+          },
+        }
+        return vim.lsp.util.convert_signature_help_to_markdown_lines(signature_help, 'zig', { '(' })
+      end)
+      local expected = { 3, 4, 3, 11 }
```
wookayin, line 3540: If I understood correctly, this refers to the second parameter (`activeParameter = 1` with zero-based index) => `_: void`; in line 3 (Row 2, with zero-based index). I was confused why it's not `{ 2, 4, 2, 11 }` -- Could you please clarify which index is exactly being used and document it if it would make it easier to understand?

It looks like we're using 1-based index here for line (ro~) or 0-based index for column (with end-exclusive), but `vim.highlight.range` seems to accept 0-based index for line/col (Range4 also implies 0-based row/col).

wookayin: Hmm, maybe I got it; looks like it's because of the additional line introduced as the code is wrapped with a codeblock: "```...```". So it *is* 0-index.

runtime/lua/vim/lsp/util.lua
MariaSolOs: @wookayin Yep! I left a short comment here:
https://github.com/neovim/neovim/blob/21911f2cc556e6b05ad58a92cc7d4539e6a939ae/runtime/lua/vim/lsp/util.lua#L900-L906

But I'll make sure to clarify this in the test too.

test/functional/plugin/lsp_spec.lua
MariaSolOs: @wookayin exactly! I actually left a small comment about that here:
https://github.com/neovim/neovim/blob/21911f2cc556e6b05ad58a92cc7d4539e6a939ae/runtime/lua/vim/lsp/util.lua#L900-L906

But I'll make sure to clarify this in the test too.

MariaSolOs: And thank you for the review @wookayin <3!

PR #30467, opened 2024-09-22
build: use treesitter's CMakeLists.txt

dundargoc:


dundargoc:
I guess we could wait with this PR until treesitter has a releases and then use that instead of bumping the version twice. I don't mind either way, thinking out loud in case anyone has a preference.

justinmk:
we already pin to a version (SHA), so I don't see a problem with this, except that we'll need to be aware when the upstream changes, right?

dundargoc:
> except that we'll need to be aware when the upstream changes, right?

We have clasonbot for that so should be fine.

PR #30468, opened 2024-09-22
build(make): use `git clean -fdx` for `make distclean`

dundargoc:
This also removes all files generated files but without cross-platform
boilerplate. A small downside is this does not work for tarballs, but I
don't believe it is significant enough to block this.


zeertzjq (changes requested):
This will renove config files as well.

dundargoc:
You mean `local.mk`? Yeah, that's true, forgot about that.

dundargoc:
Bleh, people use this apparently. cringe.

PR #30476, opened 2024-09-23
feat(paste): unify cancel and error behavior

### LINKED ISSUE #30462: bracketed paste is not cancelled when vim.paste() returns false
### Problem

bracketed paste is not cancelled when `vim.paste()` returns `false`

### Steps to reproduce

1. Run `nvim --clean`
2. Source the following file:
```lua
vim.paste = (function(overridden)
  return function(lines, phase)
    for i, line in ipairs(lines) do
      if line:find('CANCEL') then
        return false
      end
    end
    return overridden(lines, phase)
  end
end)(vim.paste)
```
3. Use bracketed paste to paste a piece of text that:
   - is long enough to be split into multiple chunks
   - contains `CANCEL` in a non-last chunk
4. The chunk containing `CANCEL` is not pasted, but the chunks that come after are still pasted

In contrast, throwing an error does prevent the following chunks from being pasted:
```lua
vim.paste = (function(overridden)
  return function(lines, phase)
    for i, line in ipairs(lines) do
      if line:find('CANCEL') then
        error(42)
      end
    end
    return overridden(lines, phase)
  end
end)(vim.paste)
```

### Expected behavior

None of the chunks starting from the chunk containing `CANCEL` are pasted

### Nvim version (nvim -v)

v0.11.0-dev-812+ge697c1b43d

N/A

### Operating system/version

Arch Linux

### Terminal name/version

kitty 0.36.2

xterm-kitty

### Installation

build from repo

zeertzjq:
Before this PR, the behavior of nvim_paste is:
- When vim.paste() returns false, return false to the client, but treat
  following chunks normally (i.e. rely on the client cancelling the
  paste as expected).
- When vim.paste() throws an error, still return true to the client, but
  drain the following chunks in the stream without calling vim.paste().

There are two problems with such behavior:
- When vim.paste() errors, the client is still supposed to send the
  remaining chunks of the stream, even though they do nothing.
- Having different code paths for two uncommon but similar situations
  complicates maintenance.

This PR makes both the cancel case and the error case return false to
the client and drain the remaining chunks of the stream, which, apart
from sharing the same code path, is beneficial whether the client checks
the return value of nvim_paste or not:
- If the client checks the return value, it can avoid sending the
  following chunks needlessly after an error.
- If the client doesn't check the return value, chunks following a
  cancelled chunk won't be pasted on the server regardless, which leads
  to less confusing behavior.

Fix #30462


src/nvim/api/vim.c
```diff
@@ -1255,26 +1254,26 @@ Boolean nvim_paste(String data, Boolean crlf, Integer phase, Arena *arena, Error
...
-theend:
-  if (cancel || phase == -1 || phase == 3) {  // End of paste-stream.
-    draining = false;
+  if (phase == 3 || phase == (cancelled ? 2 : -1)) {
     paste_store(kTrue, NULL_STRING, crlf);
   }
-
-  return !cancel;
+theend:
+  ;
```
justinmk, line 1271: unnecessary ; ?

justinmk (commented):
Good idea and implementation :)

src/nvim/api/vim.c
zeertzjq: A declaration can't follow a label

PR #30480, opened 2024-09-23
fix(vim.fs): dirname() returns "." on mingw/msys2

### LINKED ISSUE #30212: vim.fs.dirname() returns "." on MSYS2
### Problem

```
=vim.fs.dirname(vim.fn.stdpath('config'))
.
```

### Steps to reproduce

/

### Expected behavior

```
C:\User\XXX\AppData\Local
```

### Neovim version (nvim -v)

0.10.0

no

### Operating system/version

Windows 10

### Terminal name/version

windows terminal

xterm-256color

### Installation

Msys2

justinmk:
Problem:
`vim.fs.dirname([[C:\User\XXX\AppData\Local]])` returns "." on mingw/msys2.

Solution:
- Check for "mingw" when deciding `iswin`.
- Use `has("win32")` where possible, it works in "fast" contexts since  b02eeb6a7281df0561a021d7ae595c84be9a01be.

fix #30212

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -2,7 +2,9 @@ local uv = vim.uv
 
 local M = {}
 
-local iswin = uv.os_uname().sysname == 'Windows_NT'
+-- Can't use `has('win32')` because the `nvim -ll` test runner doesn't support `vim.fn` yet.
+local sysname = uv.os_uname().sysname:lower()
+local iswin = not not (sysname:find('windows') or sysname:find('mingw'))
 local os_sep = iswin and '\\' or '/'
 
 --- Iterate over all the parents of the given path.
--- a/test/testutil.lua
+++ b/test/testutil.lua
@@ -392,7 +392,7 @@ function M.check_logs()
   )
 end
 
-function M.sysname()
+local function sysname()
   return uv.os_uname().sysname:lower()
 end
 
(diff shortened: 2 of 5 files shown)


test/testutil.lua
```diff
@@ -392,7 +392,7 @@ function M.check_logs()
   )
 end
 
-function M.sysname()
+local function sysname()
   return uv.os_uname().sysname:lower()
```
justinmk, line 396: not used in any existing tests, and they should not really be using this directly.

runtime/lua/vim/fs.lua
```diff
@@ -2,6 +2,7 @@ local uv = vim.uv
 
 local M = {}
 
+-- Can't use `has('win32')` because the `nvim -ll` test runner doesn't support `vim.fn` yet.
 local iswin = uv.os_uname().sysname == 'Windows_NT'
```
zeertzjq, line 6: Also check for MinGW to fix #30212?

runtime/lua/vim/fs.lua
```diff
@@ -2,7 +2,9 @@ local uv = vim.uv
 
 local M = {}
 
-local iswin = uv.os_uname().sysname == 'Windows_NT'
+-- Can't use `has('win32')` because the `nvim -ll` test runner doesn't support `vim.fn` yet.
+local sysname = uv.os_uname().sysname:lower()
+local iswin = sysname:find('windows') or sysname:find('mingw')
```
justinmk, line 7: We should be using has() here, but can't yet because `fs_spec.lua` runs its (many) tests in the `nvim -ll` test runner, which doesn't support `vim.fn` yet.

clason:
Can this be backported? Sounds like a valuable fix.

Issue #28437, opened 2024-04-21
FUNC_API_REMOTE_ONLY in Lua/Vimscript


justinmk:
### Problem

FUNC_API_REMOTE_ONLY marks functions that require a channel id and these functions currently aren't usable from Lua/vimscript.

### Expected behavior

Rename FUNC_API_REMOTE_ONLY to FUNC_API_CHAN_REQUIRED so that Lua/Vimscript can pass an explicit channel id. This allows those functions to be batched via `nvim_exec_lua` , and removes the last remaining use case of nvim_call_atomic. https://github.com/neovim/neovim/pull/28433

zeertzjq:
Some APIs may have different behavior depending on whether it is called via RPC (e.g. `nvim_buf_attach`, `nvim_get_chan_info`), and the script context (for `:verbose`) is also different for Lua vs RPC, so this may be needed for other normal APIs as well.

zeertzjq:
`vim._with()` may be a suitable place to add this, but that's Lua-only.

PR #30491, opened 2024-09-24
fix(paste): only record a paste when it's from RPC

zeertzjq:
Problem:  When using nvim_paste in a mapping during a macro recording,
          both the mapping and the paste are recorded, causing the paste
          to be performed twice when replaying the macro.
Solution: Only record a paste when it is from RPC.

Unfortunately this means there is no way for a script to make a recorded
paste. ~, unless scripts can override the call's channel ID: #28437.~ A way to enable that can be discussed later if there is need.

### DIFF
--- a/test/functional/api/vim_spec.lua
+++ b/test/functional/api/vim_spec.lua
@@ -1358,6 +1358,13 @@ describe('API', function()
         test_paste_repeat_visual_select(true)
       end)
     end)
+    it('in a mapping recorded in a macro', function()
+      command([[nnoremap <F2> <Cmd>call nvim_paste('foo', v:false, -1)<CR>]])
+      feed('qr<F2>$q')
+      expect('foo')
+      feed('@r') -- repeating a macro containing the mapping should only paste once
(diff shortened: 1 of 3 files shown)


justinmk:
Makes sense. This is another case where the lack of `opt` on the rpc function has boxed us into  a corner. We could perhaps abuse the `phase` parameter (e.g. `phase=-2` forces repeat recording).

> Unfortunately this means there is no way for a script to make a recorded paste, unless scripts can override the call's channel ID: #28437

This PR checks `is_internal_call`, would the script need to pass non-zero?

justinmk (approved):

test/functional/api/vim_spec.lua
```diff
@@ -1358,6 +1358,13 @@ describe('API', function()
         test_paste_repeat_visual_select(true)
       end)
     end)
+    it('can be used in a mapping', function()
```
justinmk, line 1361:
```suggestion
    it('in a mapping recorded in a macro', function()
```

Issue #30504, opened 2024-09-25
mouse event for buffer in non-focusable window


errael:
### Problem

I'm looking at implementing `vim`'s popup API in `lua` (actually adding missing pieces to `plenary.nvim/popup`). #30490 is targeted to the same goal and shares the problem with `focusable`/mouse-events. This issue is specifically about how to implement `close = click`.

The idea is to do something like
```
vim.keymap.set({'n', 'i'}, '<LeftRelease>', function()
    mouse_click("buffer")
end, {buffer = popup_buffer})
```
The use of `{buffer = popup_buffer}` is important to avoid interfering with other uses of `<LeftRelease>`. So the idea of examining all clicks in parent window is a non-starter.

One idea is to add a key to `nvim_open_win`, `mousable` or`mouse_enable` or `mouseevent` or ..., when true mouse events are delivered to/through the window. (`mouseevent` is similarly named to `mousemoveevent`). A simple way to define this flag's semantics are `mouseevent` defaults to the same value as `focusable`; `mouseevent` is used, not `focusable`, for deciding on whether or not to deliver mouse events. It interesting that with this, #30490 could be finessed without any changes to the `getmousepos()` API.

You can play with the interaction of `focusable` and click using this code.

```lua
local popup_wid

local function mouse_click(tag)
    local mp = vim.fn.getmousepos()
    vim.print(string.format("CLICK %s: line %d, col %d", tag, mp.screenrow, mp.screencol))
end

local xpos = 30
local ypos = 6
local function Pop1Any()
    local bnr = vim.api.nvim_create_buf(false, true)
    assert(bnr, "Failed to create buffer")
    vim.api.nvim_buf_set_lines(bnr, 0, -1, true, {'simple', '=' .. tostring(ypos) .. '='})

    vim.api.nvim_set_option_value("bufhidden", "wipe", {buf = bnr})
    vim.api.nvim_set_option_value("modifiable", true, {buf = bnr})
    popup_wid = vim.api.nvim_open_win(bnr, false, {
        relative = "editor",
        style = "minimal",
        width = 10,
        height = 2,
        focusable = false, -- XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
        col = xpos,
        row = ypos,
    })
    xpos = xpos + 3
    ypos = ypos + 3
    vim.keymap.set({'n', 'i'}, '<LeftRelease>', function() mouse_click("buffer") end, {buffer = bnr})
end

[...]

Issue #30490, opened 2024-09-23
getmousepos() doesn't find non-focusable windows


errael:
### Problem

vim/neovim difference in `getmousepos()` for floating-windows/popup-windows
In `:help getmousepos`
- `vim`: If the mouse is over a popup window then that window is used.
- `neovim`: If the mouse is over a focusable floating window then that window is used.

I'm looking for a way to get a non-focusable non-hidden window under the mouse. I want to
fill in some of the missing capabilities of plenary.nvim's popup.lua. And popup can not be
focusable. I could go through a list of all windows, filtering as required, and if needed
convert mousepos from screen to floating window. But that seems like a lot of work
to replicate in lua code something that is already mostly in core neovim.

(I noticed that `CTRL-W_CTRL-W` will move to a `hidden` `focusable` window. Is that a bug?)

### Steps to reproduce

Can provide an executable test case if needed.

(https://github.com/neovim/neovim/commit/4e4b1b65903b973cb447b5a70413d0e0363c20f0)

### Expected behavior

Some way to get a non-focusable window from `getmousepos()`.

Is the default, "must be focusable", wrong for `getmousepos()`? Should/can it be changed?

If there is a way to have an option to `getmousepos()`, like ANY_WINDOW, in the following
code, that might take care of this problem. Could get there with optional arg to
getmousepos(). Probably `getmousepos([{dict of options}])`. I'm assuming that
`get_win_by_grid_handle()`/`FOR_ALL_WINDOWS_IN_TAB` is sorted by `zindex`. Also, don't
want a "hidden" window, not sure how that works with FOR_ALL_WINDOWS_IN_TAB.

```
static win_T *mouse_find_grid_win(int *gridp, int *rowp, int *colp)
{
  ...
    if (wp && wp->w_grid_alloc.chars
        && (ANY_WINDOW || !(wp->w_floating && !wp->w_config.focusable))) {
      *rowp = MIN(*rowp - wp->w_grid.row_offset, wp->w_grid.rows - 1);
      *colp = MIN(*colp - wp->w_grid.col_offset, wp->w_grid.cols - 1);
      return wp;
    }
```
or maybe it's something more like
```
&& !(wp->w_floating && (!(ANY_FOCUS || wp->w_config.focusable)))) {
```

Something like `getmousepos({focusable: any, hide: false})` for my case.

I came across nvim__inspect_cell, but it doesn't return the win_id (and ignoring the fact
that it's hidden somewhere in lua metadata).

[...]

PR #30507, opened 2024-09-25
fix(window): respect hide flag of float windows when switching

### LINKED ISSUE #30503: CTRL-W_CTRL-W transfers focus to hidden window
### Problem

If there is a focusable hidden window things get weird.

### Steps to reproduce

1. source the script below
2. enters Ctrl-w_Ctrl-w the focus is transferred to the hidden window
3. enter `:ls`\<CR>
4. enter \<CR>

Observe cursor stays on command line

6. enter \<ESC>

Observe cursor stays in command line

```
local function Pop1Any()
    vim.api.nvim_buf_set_lines(bnr, 0, -1, true, {'simple', '=' .. tostring(ypos) .. '='})

	vim.api.nvim_open_win(bnr, false, {
		relative = "editor",
		style = "minimal",
		width = 10,
		height = 2,
		focusable = true,
		hide = true,
		col = 10,
		row = 10,
	})
end

Pop1Any()
```

### Expected behavior

Could argue that it is the developer's responsibility to modify `focusable` to track `hide`; and in this situation the behavior is undefined and so there is no bug. But getting stuck on the command line is pretty weird.

### Nvim version (nvim -v)

v0.11.0-dev - https://github.com/neovim/neovim/commit/052875b9dc3de3283844a30dce9646f26982542d

NA

### Operating system/version

ubuntu

### Terminal name/version

gnome

xterm-256color

### Installation

make install

glepnir:
Problem: The hide option for floating windows was not respected in some window detection functions, causing them to improperly include hidden floating windows.

Solution: Added checks for the hide property in w_config to ensure hidden floating windows are properly ignored in relevant detection functions.

Fix #30503

### DIFF
--- a/src/nvim/ex_cmds2.c
+++ b/src/nvim/ex_cmds2.c
@@ -590,7 +590,7 @@ void ex_listdo(exarg_T *eap)
           break;
         }
         assert(wp);
-        execute = !wp->w_floating || wp->w_config.focusable;
+        execute = !wp->w_floating || (!wp->w_config.hide && wp->w_config.focusable);
         if (execute) {
           win_goto(wp);
           if (curwin != wp) {
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -359,13 +359,13 @@ newwindow:
             wp = lastwin;  // wrap around
           }
           while (wp != NULL && wp->w_floating
-                 && !wp->w_config.focusable) {
+                 && (wp->w_config.hide || !wp->w_config.focusable)) {
             wp = wp->w_prev;
           }
         } else {  // go to next window
           wp = curwin->w_next;
           while (wp != NULL && wp->w_floating
-                 && !wp->w_config.focusable) {
+                 && (wp->w_config.hide || !wp->w_config.focusable)) {
             wp = wp->w_next;
           }
           if (wp == NULL) {
@@ -2851,7 +2851,7 @@ int win_close(win_T *win, bool free_buf, bool force)
           break;
         }
         if (!wp->w_p_pvw && !bt_quickfix(wp->w_buffer)
-            && !(wp->w_floating && !wp->w_config.focusable)) {
+            && !(wp->w_floating && (wp->w_config.hide || !wp->w_config.focusable))) {
           curwin = wp;
           break;
         }
--- a/test/functional/ui/float_spec.lua
+++ b/test/functional/ui/float_spec.lua
@@ -9068,6 +9068,7 @@ describe('float window', function()
     end)
 
     it('float window with hide option', function()
+      local cwin = api.nvim_get_current_win()
       local buf = api.nvim_create_buf(false,false)
       local win = api.nvim_open_win(buf, false, {relative='editor', width=10, height=2, row=2, col=5, hide = true})
       local expected_pos = {
@@ -9147,6 +9148,22 @@ describe('float window', function()
                                                   |
         ]])
       end
+      -- check window jump with hide
+      feed('<C-W><C-W>')
+      -- should keep on current window
+      eq(cwin, api.nvim_get_current_win())
+      api.nvim_win_set_config(win, {hide=false})
+      api.nvim_set_current_win(win)
+      local win3 = api.nvim_open_win(buf, true, {relative='editor', width=4, height=4, row=2, col=5, hide = false})
+      api.nvim_win_set_config(win, {hide=true})
+      feed('<C-W>w')
+      -- should goto the first window with prev
+      eq(cwin, api.nvim_get_current_win())
+      -- windo
+      command('windo set winheight=6')
+      eq(win3, api.nvim_get_current_win())
+      eq(6, api.nvim_win_get_height(win3))
+      eq(2, api.nvim_win_get_height(win))
     end)
 
     it(':fclose command #9663', function()


zeertzjq:
It may be needed for `:windo` as well:
```diff
diff --git a/src/nvim/ex_cmds2.c b/src/nvim/ex_cmds2.c
index 210539d6a4..275219411a 100644
--- a/src/nvim/ex_cmds2.c
+++ b/src/nvim/ex_cmds2.c
@@ -590,7 +590,7 @@ void ex_listdo(exarg_T *eap)
           break;
         }
         assert(wp);
-        execute = !wp->w_floating || wp->w_config.focusable;
+        execute = !wp->w_floating || (wp->w_config.focusable && !wp->w_config.hide);
         if (execute) {
           win_goto(wp);
           if (curwin != wp) {
```

justinmk:
Is there (or should there be) such a thing as a focusable "hidden" window? If not, then can we set `focusable=false` when `hide` is set? Then we don't need to check both.

What does it mean for a window to be hidden but focusable?

zeertzjq:
Oh, otherwise another check is needed in `win_close()` as well.

glepnir:
root of the problem should be what you @justinmk said. Focusable has no meaning for the hide window, maybe? Because it is invisible, the cursor cannot be focused on this window. It should be checked and an error should be thrown when parsing config. @

glepnir:
Maybe not. In the initial case I'd hide it. Record the hide state. Re-show the window when needed. It could be focusable so I'd only have to record one state instead of two right ?

justinmk:
We need both fields, but the question is, in the code updated in this PR (and similar), why would this code need to check `hide`? When would a window have `hide=true` and `focusable=true` ?

glepnir:
> In the initial case I'd hide it. Record the hide state. Re-show the window when needed. 

initial case create a hide float window first . when some condition is triggered it will be shown by `hide=true`  `focusable` can only set in initial case . because it is hidden it will not be focusable even if focusable is true. Plugin authors do not need to worry about whether certain fields can be set at the same time.

justinmk:
> because it is hidden it will not be focusable even if focusable is true.

Show a Lua code sample where this happens. 

What I'm thinking is, if a plugin sets `focusable=true` then either (1) that should an error, or (2) we use that to also set `hide=false` implicitly.

The question is whether this situation *makes sense*. If it doesn't make sense then we should do one of the above, and then *internally* we don't need to check `(!wp->w_config.hide && wp->w_config.focusable)` everywhere (that makes no sense), we only need to check `focusable`.

zeertzjq:
A plugin usually only changes one field in `nvim_win_set_config`, in which case the other flags remain unchanged.

justinmk:
> A plugin usually only changes one field in `nvim_win_set_config`, in which case the other fields remain unchanged.

But I am suggesting that if a plugin calls `nvim_win_set_config(..., {focusable=true})`, that should automatically also set `hide=false`. Alternatively, I am asking why we shouldn't do that--what is the use-case for `hide=true &&` focusable=true` ?

errael:
> But I am suggesting that if a plugin calls `nvim_win_set_config(..., {focusable=true})`, that should automatically also set `hide=false`.

If `nvim_win_set_config(..., {hide=true})` then `focusable` would be automatically set to `false`?

Later doing`nvim_win_set_config(..., {hide=false})`, what happens to `focusable`?

Maybe `nvim_win_set_config` only allow hide to be set to `true` and ... Never mind, it gets crazy.

-----

PS: As mentioned elsewhere: hide is an action, hidden is a state (and I prefer visible). Can it be changed?

errael:
Mentioning this here because there is a similar problem coming up regarding interacting window flags.

The recently opened #30490 and #30504 can be solved by the addition of a window config flag [mousable](https://wiktionary.org/wiki/mouseable). `mousable` has a similar relationship with `focusable` as `hide`. 

In `nvim_win_open` if `mousable` is not explicitly set, it should be set to the the value of `focusable`. If `focusable` and/or `hide` are changed, and `mousable` is not explicitly changed as well, what should happen? #30504 will have evolving notes on this.

Should it be tracked if an option/flag is, or has been, explicitly set?

Curiously before logging in this morning I was planning to ask if there are any guidelines and/or macros for handling interacting options/flags. Guess the jury is still out.

zeertzjq:
Anyway, in the current implementable, this change is needed for `win_close()`:
```diff
diff --git a/src/nvim/window.c b/src/nvim/window.c
index f61485ccb3..3647bd1b4c 100644
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -2851,7 +2851,7 @@ int win_close(win_T *win, bool free_buf, bool force)
           break;
         }
         if (!wp->w_p_pvw && !bt_quickfix(wp->w_buffer)
-            && !(wp->w_floating && !wp->w_config.focusable)) {
+            && !(wp->w_floating && (wp->w_config.hide || !wp->w_config.focusable))) {
           curwin = wp;
           break;
         }
```

glepnir:
Emulates something like emacs and helix showing diagnostics in the top right corner. I can do this to allow jumping to pasted diagnostic information. Because it won't be focused when hidden. I only need to change hide when showing it. Rather than changing focusable and hide.

```lua
-- just example
local bufnr = api.nvim_create_buf(false, false)
local win = api.nvim_open_win(bufnr, false, {
  relative = 'editor',
  row = 1,
  col = vim.o.columns - 20,
  height = 20,
  width = 20,
  focusable = true,
  hide = true,
})
au('DiagnosticChanged', {
  callback = function()
    local d = vim.diagnostic.get()
    if not vim.tbl_isempty(d) then
      api.nvim_buf_set_lines(bufrn, 0, -1, false, d)
      return api.nvim_win_set_config(win, { hide = false })
    end
    api.nvim_win_set_config(win, { hide = true })
  end,
})
```

errael:
After some experimentation using the patch. Works good with my simple UI usage.

There's an issue about having a visible cursor in a non-focused area. IMHO, it doesn't seem priority. Maybe it should be considered a separate issue.

The following might be considered an artificial situation. It's still possible to enter hidden, focusable windows through scripts (or command line). Even though, in general, a script shouldn't leave the focus in a hidden/focusable window, there's some behavior I'm seeing that might be considered a bug.


Here's a mapping used with the problem code from #30503
```
vim.keymap.set('n', 'Z2', function() vim.api.nvim_set_current_win(popup_wid) end)
```

1. Open the test file from #30503, move the cursor to the middle of the file.
2. `:sou` creating the hidden focusable window
3. Execute the mapping `Z2`. Moves the focus to the hidden window
   Does `vim.api.nvim_set_current_win(popup_wid)` 

Observe: Cursor is visible in the main/**non-focused** window.

4. Scroll the window with the mouse so only last line of buffer is visible in the window.

Observe the cursor doesn't move and ends up far outside the buffer.

5. Execute some ":" commands

Observe the cursor move around in command line.

glepnir:
@errael i will look at it later maybe have a delay..

zeertzjq:
Let's merge this for now.

Issue #25672, opened 2023-10-16
Options refactor roadmap


famiu:
### Problem

The Neovim options code is an incomprehensible mess filled with code duplication, special logic for every option, multiple code paths for the same thing, lack of generalization, etc. This makes Neovim options difficult to maintain, and even more difficult to extend, which has been a blocker towards the addition of new option types (e.g. Callback, Array, Dictionary).

### Expected behavior

This is a roadmap for refactoring Neovim options to be more extensible and maintainable. This is not an absolutely concrete roadmap and each item is subject to change, further additions and deletions may also be made at any time.

- [ ] Unify interfaces for setting options.
  - [x] Implement a single type to represent any option value. This is done in #23850 using `OptVal`.
  - [x] Unify `do_set_option_value` and `set_option_value` codepaths. Done in #25394.
  - [ ] Unify interfaces for setting number, boolean and string options.
    - [x] Unify interfaces for setting number and boolean options. Implemented in #25394.
    - [ ] Unify interfaces for setting number/boolean and string options.
    - [ ] Unify separate interfaces for setting string options (e.g. `set_string_option_direct`, `set_string_option_global`, etc.)
    - [ ] Remove certain functions specific to string options (e.g. `free_string_option`, `clear_string_option`) and replace them with something more general.
  - [ ] Use `OptVal` for option `def_val`.
- [ ] Check if `OPT_FREE` is necessary and remove it if it isn't.
- [ ] Use reference counted strings for string options.
- [ ] Allow certain options to have multiple types. Ref: #24045
- [ ] Use a separate option type (`kOptValTypeUnset`?) to indicate an unset local value for a global-local option, instead of using arbitrary values for each option. This also would mean that boolean options no longer need to be represented using tri-states and instead can be actual booleans.
- [ ] Use `bool` type for boolean option variables, instead of `int`.
- [ ] Fix the behavior of `winbar` and make it possible to unset winbar for a single window.
- [ ] Convert `P_FUNC` to a separate option type for callback options, instead of having them be shoehorned into string options.
- [ ] Add support for Array and Dictionary options.
- [ ] Convert all list-like or map-like string options to Array / Dictionary options.
- [ ] Add a dict-based configuration format for complicated options such as `winbar`, `tabline`, `statusline`, etc.

famiu:
CC: @justinmk @bfredl

justinmk:
[...]

PR #30510, opened 2024-09-25
docs: dev_arch, misc

justinmk:


runtime/doc/options.txt
```diff
@@ -6014,7 +6013,6 @@ A jump table for the options with a short description can be found at |Q_op|.
 						*'statuscolumn'* *'stc'*
 'statuscolumn' 'stc'	string	(default "")
 			local to window
-	EXPERIMENTAL
```
justinmk, line 6017: 'statuscolumn' as been really stable AFAIK (thanks @luukvbaal !), and we don't plan to change its interface. Can probably graduate it?

luukvbaal: I tend to agree, however @bfredl wasn't happy with the fact that the statuscolumn width is estimated, and that a wrong estimation incurs a "panic" redraw. I think he went so far as to say that we couldn't graduate the feature without solving that but I don't really see a way around it (without sacrificing the "grow as needed" property it holds now).

The interface _might_ change the same way 'statusline' might, depending on the ongoing option work ("[Add a dict-based configuration format for complicated options such as winbar, tabline, statusline, etc](https://github.com/neovim/neovim/issues/25672)"). IDK if either of those reasons should actually prevent the feature from being marked stable.

runtime/doc/dev_arch.txt
```diff
@@ -0,0 +1,51 @@
+*dev_internals.txt*          Nvim
+
+
+                            NVIM REFERENCE MANUAL
+
+
+How to develop Nvim, explaination of modules and subsystems    *dev-arch*
```
justinmk, line 7: introducing dev_arch where we can explain the overall system and provide instructions for common dev tasks (such as adding a new UI event)

runtime/doc/options.txt
bfredl: My concern remains, but I am not against removing the `EXPERIMENTAL` label as such. We don't have a "contract" for options the way we do for API functions, and it is reasonable to expect the interface to _evolve_ (with possibly lightly breaking changes, same as any other option) rather than being completely redesigned.

runtime/doc/dev_arch.txt
telemachus: Small typo: s/explaination/explanation/

PR #30513, opened 2024-09-25
build: fix or silence new clang-tidy warnings

dundargoc:


src/nvim/event/loop.h
```diff
@@ -9,8 +9,8 @@
 
 typedef void *WatcherPtr;
 
-#define _NOOP(x)
-KLIST_INIT(WatcherPtr, WatcherPtr, _NOOP)
+#define NOOP(x)
+KLIST_INIT(WatcherPtr, WatcherPtr, NOOP)
```
justinmk, line 13: side note, this is the only place we use klist, could we use kvec instead?

dundargoc: @bfredl is the kväckmaster, he'd know

Issue #29365, opened 2024-06-15
"focusble" is ignored for non-floating windows


cortesi:
### Problem

The "focusable" attribute is explicitly ignored for non-floating windows. For example, here is the code for going to the next/previous window, which ignores focusable==false unless the window is also floating:

https://github.com/neovim/neovim/blob/7e65f3757bdbe41bbf022b05b6869ad6e7febe0d/src/nvim/window.c#L355-L374

I'm working on an addon that adds "companion windows" for certain functionality where non-focusable, non-floating windows would be really useful. I note that it's possible to set the focusable flag on non-floating windows without error, and the fact that focusable only applies to floating windows is not mentioned in the docs. 

- I'd be happy create a PR to change window navigation to respect the focusable flag unambiguously. Can anyone who knows the project well think of a reason why this shouldn't be done?
- If we can't fix the behaviour, I think we should at least fix the docs to be explicit about this. We should also consider returning an error when the focusable flag is set on a non-floating window, as we currently do for other attributes that are floating-only (borders for example). I'd be happy to create a PR for this too.

### Expected behavior

When the focusable flag is false for a non-floating window, I expect window navigation commands to skip the window.

obukhovaa:
Bump, stumble upon the same issue while making similar type of plugin. It is quite inconvenient to do navigation across floating windows while there are non-floating windows in the background, it requires a lot of hustle to ensure cursor doesn't jump there...

justinmk:
related:

- https://github.com/neovim/neovim/issues/30504
- https://github.com/neovim/neovim/issues/30503

PR #30542, opened 2024-09-27
feat(float): support callback for mouse event

### LINKED ISSUE #30504: mouse event for buffer in non-focusable window

zeertzjq:
Close #30504

Early draft. Callback doesn't have arguments yet.

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -3198,10 +3198,15 @@ nvim_open_win({buffer}, {enter}, {config})                   *nvim_open_win()*
                     entered by |nvim_set_current_win()|, or, when the `mouse`
                     field is set to true, by mouse events.
                   • mouse: Specify how this window interacts with mouse
-                    events. Defaults to `focusable` value.
+                    events. May be a boolean or a Lua callback. Defaults to
+                    `focusable` value.
                     • If false, mouse events pass through this window.
                     • If true, mouse events interact with this window
                       normally.
+                    • If a Lua callback, mouse events interact with this
+                      window, but the callback is called in place of the
+                      default handling. The default handling will be used
+                      again if callback returns `true`.
                   • external: GUI should display the window as an external
                     top-level window. Currently accepts no other positioning
                     configuration together with this.
--- a/src/nvim/api/win_config.c
+++ b/src/nvim/api/win_config.c
@@ -1,3 +1,4 @@
+#include <lauxlib.h>
 #include <stdbool.h>
 #include <string.h>
 
@@ -23,6 +24,7 @@
 #include "nvim/globals.h"
 #include "nvim/grid_defs.h"
 #include "nvim/highlight_group.h"
+#include "nvim/lua/executor.h"
 #include "nvim/macros_defs.h"
 #include "nvim/mbyte.h"
 #include "nvim/memory.h"
@@ -132,9 +134,12 @@
 ///       |nvim_set_current_win()|, or, when the `mouse` field is set to true,
 ///       by mouse events.
 ///   - mouse: Specify how this window interacts with mouse events.
-///       Defaults to `focusable` value.
+///       May be a boolean or a Lua callback. Defaults to `focusable` value.
 ///       - If false, mouse events pass through this window.
 ///       - If true, mouse events interact with this window normally.
+///       - If a Lua callback, mouse events interact with this window,
+///         but the callback is called in place of the default handling.
+///         The default handling will be used again if callback returns `true`.
 ///   - external: GUI should display the window as an external
 ///       top-level window. Currently accepts no other positioning
 ///       configuration together with this.
@@ -719,7 +724,18 @@ Dict(win_config) nvim_win_get_config(Window window, Arena *arena, Error *err)
   PUT_KEY_X(rv, focusable, config->focusable);
   PUT_KEY_X(rv, external, config->external);
   PUT_KEY_X(rv, hide, config->hide);
-  PUT_KEY_X(rv, mouse, config->mouse);
+
+  switch (config->mouse) {
+  case kWinMouseIgnore:
+    PUT_KEY_X(rv, mouse, BOOLEAN_OBJ(false));
+    break;
+  case kWinMouseDefault:
+    PUT_KEY_X(rv, mouse, BOOLEAN_OBJ(true));
+    break;
+  case kWinMouseCallback:
+    PUT_KEY_X(rv, mouse, LUAREF_OBJ(api_new_luaref(config->mouse_cb)));
+    break;
+  }
 
   if (wp->w_floating) {
     PUT_KEY_X(rv, width, config->width);
@@ -1208,11 +1224,22 @@ static bool parse_win_config(win_T *wp, Dict(win_config) *config, WinConfig *fco
 
   if (HAS_KEY_X(config, focusable)) {
     fconfig->focusable = config->focusable;
-    fconfig->mouse = config->focusable;
+    if (fconfig->mouse != kWinMouseCallback) {
+      fconfig->mouse = config->focusable ? kWinMouseDefault : kWinMouseIgnore;
+    }
   }
 
   if (HAS_KEY_X(config, mouse)) {
-    fconfig->mouse = config->mouse;
+    if (config->mouse.type == kObjectTypeLuaRef) {
+      fconfig->mouse = kWinMouseCallback;
+      fconfig->mouse_cb = config->mouse.data.luaref;
+      config->mouse.data.luaref = LUA_NOREF;
+    } else if (config->mouse.type == kObjectTypeBoolean) {
+      fconfig->mouse = config->mouse.data.boolean ? kWinMouseDefault : kWinMouseIgnore;
+      fconfig->mouse_cb = LUA_NOREF;
+    } else {
+      api_set_error(err, kErrorTypeValidation, "invalid type for 'mouse'");
+    }
   }
 
   if (HAS_KEY_X(config, zindex)) {
--- a/runtime/lua/vim/_meta/api.lua
+++ b/runtime/lua/vim/_meta/api.lua
@@ -1770,9 +1770,12 @@ function vim.api.nvim_open_term(buffer, opts) end
 ---     `nvim_set_current_win()`, or, when the `mouse` field is set to true,
 ---     by mouse events.
 --- - mouse: Specify how this window interacts with mouse events.
----     Defaults to `focusable` value.
+---     May be a boolean or a Lua callback. Defaults to `focusable` value.
 ---     - If false, mouse events pass through this window.
 ---     - If true, mouse events interact with this window normally.
+---     - If a Lua callback, mouse events interact with this window,
+---       but the callback is called in place of the default handling.
+---       The default handling will be used again if callback returns `true`.
 --- - external: GUI should display the window as an external
 ---     top-level window. Currently accepts no other positioning
 ---     configuration together with this.
--- a/runtime/lua/vim/_meta/api_keysets.lua
+++ b/runtime/lua/vim/_meta/api_keysets.lua
@@ -295,7 +295,7 @@ error('Cannot require a meta file')
 --- @field bufpos? any[]
 --- @field external? boolean
 --- @field focusable? boolean
---- @field mouse? boolean
+--- @field mouse? any
 --- @field vertical? boolean
 --- @field zindex? integer
 --- @field border? any
--- a/src/nvim/api/keysets_defs.h
+++ b/src/nvim/api/keysets_defs.h
@@ -119,7 +119,7 @@ typedef struct {
   Array bufpos;
   Boolean external;
   Boolean focusable;
-  Boolean mouse;
+  Object mouse;
   Boolean vertical;
   Integer zindex;
   Object border;
--- a/src/nvim/buffer_defs.h
+++ b/src/nvim/buffer_defs.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <lauxlib.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -904,6 +905,12 @@ typedef enum {
   kFloatRelativeMouse = 3,
 } FloatRelative;
 
+typedef enum {
+  kWinMouseIgnore = 0,
+  kWinMouseDefault = 1,
+  kWinMouseCallback = 2,
+} WinMouseEvent;
+
 /// Keep in sync with win_split_str[] in nvim_win_get_config() (api/win_config.c)
 typedef enum {
   kWinSplitLeft = 0,
@@ -938,7 +945,8 @@ typedef struct {
   FloatRelative relative;
   bool external;
   bool focusable;
-  bool mouse;
+  WinMouseEvent mouse;
+  LuaRef mouse_cb;
   WinSplit split;
   int zindex;
   WinStyle style;
@@ -965,7 +973,8 @@ typedef struct {
                                       .row = 0, .col = 0, .anchor = 0, \
                                       .relative = 0, .external = false, \
                                       .focusable = true, \
-                                      .mouse = true, \
+                                      .mouse = kWinMouseDefault, \
+                                      .mouse_cb = LUA_NOREF, \
                                       .split = 0, \
                                       .zindex = kZIndexFloatDefault, \
                                       .style = kWinStyleUnused, \
--- a/src/nvim/mouse.c
+++ b/src/nvim/mouse.c
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include "nvim/api/private/defs.h"
+#include "nvim/api/private/helpers.h"
 #include "nvim/ascii_defs.h"
 #include "nvim/buffer.h"
 #include "nvim/buffer_defs.h"
@@ -13,6 +15,7 @@
 #include "nvim/edit.h"
 #include "nvim/eval.h"
 #include "nvim/eval/typval.h"
+#include "nvim/eval/window.h"
 #include "nvim/ex_docmd.h"
 #include "nvim/fold.h"
 #include "nvim/getchar.h"
@@ -20,6 +23,7 @@
 #include "nvim/grid.h"
 #include "nvim/grid_defs.h"
 #include "nvim/keycodes.h"
+#include "nvim/lua/executor.h"
 #include "nvim/macros_defs.h"
 #include "nvim/mark_defs.h"
 #include "nvim/mbyte.h"
@@ -222,6 +226,45 @@ static void call_click_def_func(StlClickDefinition *click_defs, int col, int whi
   got_click = false;
 }
 
+/// When mouse events happen in a window with a mouse callback, invoke it.
+///
+/// @param  button  !is_scroll: return value of get_mouse_button()
+///                 is_scroll: a MSCR_* enum
+///
+/// @return  whether to proceeded with mouse handling.
+static bool handle_mouse_cb(bool is_click, bool is_drag, bool is_scroll, int button)
+{
+  static handle_T last_click_win = 0;
+  int grid = mouse_grid;
+  int row = mouse_row;
+  int col = mouse_col;
+
+  win_T *wp = NULL;
+  if (!is_click && !is_scroll && button != MOUSE_RELEASE) {  // drag or release
+    if (last_click_win > 0) {
+      wp = win_id2wp(last_click_win);
+    }
+  } else {  // click, scroll or move
+    wp = mouse_find_win(&grid, &row, &col);
+    if (is_click) {
+      last_click_win = wp != NULL ? wp->handle : 0;
+    }
+  }
+  if (wp == NULL || wp->w_config.mouse != kWinMouseCallback) {
+    return true;
+  }
+
+  Error err = ERROR_INIT;
+  Object res = nlua_call_ref(wp->w_config.mouse_cb, NULL, (Array)ARRAY_DICT_INIT,
+                             kRetNilBool, NULL, &err);
+  bool retval = LUARET_TRUTHY(res);
+  if (ERROR_SET(&err)) {
+    semsg_multiline("E5108: %s", err.msg);
+    api_clear_error(&err);
+  }
+  return retval;
+}
+
 /// Translate window coordinates to buffer position without any side effects.
 /// Returns IN_BUFFER and sets "mpos->col" to the column when in buffer text.
 /// The column is one for the first column.
@@ -369,6 +412,10 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
       }
     }
     break;
+  }
+
+  if (!handle_mouse_cb(is_click, is_drag, false, which_button)) {
+    return false;
   }
 
   if (c == K_MOUSEMOVE) {
@@ -1063,6 +1110,10 @@ void do_mousescroll(cmdarg_T *cap)
 /// of the MSCR_ values.
 void ins_mousescroll(int dir)
 {
+  if (!handle_mouse_cb(false, false, true, dir)) {
+    return;
+  }
+
   cmdarg_T cap;
   oparg_T oa;
   CLEAR_FIELD(cap);
@@ -1572,6 +1623,10 @@ static bool do_mousescroll_horiz(colnr_T leftcol)
 /// "cap->arg", which is one of the MSCR_ values.
 void nv_mousescroll(cmdarg_T *cap)
 {
+  if (!handle_mouse_cb(false, false, true, cap->arg)) {
+    return;
+  }
+
   win_T *const old_curwin = curwin;
 
   if (mouse_row >= 0 && mouse_col >= 0) {
@@ -1740,7 +1795,7 @@ static win_T *mouse_find_grid_win(int *gridp, int *rowp, int *colp)
   } else if (*gridp > 1) {
     win_T *wp = get_win_by_grid_handle(*gridp);
     if (wp && wp->w_grid_alloc.chars
-        && !(wp->w_floating && !wp->w_config.mouse)) {
+        && !(wp->w_floating && wp->w_config.mouse == kWinMouseIgnore)) {
       *rowp = MIN(*rowp - wp->w_grid.row_offset, wp->w_grid.rows - 1);
       *colp = MIN(*colp - wp->w_grid.col_offset, wp->w_grid.cols - 1);
       return wp;
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -45,6 +45,7 @@
 #include "nvim/grid_defs.h"
 #include "nvim/hashtab.h"
 #include "nvim/keycodes.h"
+#include "nvim/lua/executor.h"
 #include "nvim/macros_defs.h"
 #include "nvim/main.h"
 #include "nvim/map_defs.h"
@@ -802,6 +803,9 @@ int win_fdccol_count(win_T *wp)
 void merge_win_config(WinConfig *dst, const WinConfig src)
   FUNC_ATTR_NONNULL_ALL
 {
+  if (dst->mouse_cb != src.mouse_cb) {
+    api_free_luaref(dst->mouse_cb);
+  }
   if (dst->title_chunks.items != src.title_chunks.items) {
     clear_virttext(&dst->title_chunks);
   }
@@ -857,7 +861,7 @@ void ui_ext_win_position(win_T *wp, bool validate)
       String anchor = cstr_as_string(float_anchor_str[c.anchor]);
       if (!c.hide) {
         ui_call_win_float_pos(wp->w_grid_alloc.handle, wp->handle, anchor,
-                              grid->handle, row, col, c.mouse,
+                              grid->handle, row, col, c.mouse != kWinMouseIgnore,
                               wp->w_grid_alloc.zindex);
       } else {
         ui_call_win_hide(wp->w_grid_alloc.handle);
@@ -889,7 +893,7 @@ void ui_ext_win_position(win_T *wp, bool validate)
         ui_comp_put_grid(&wp->w_grid_alloc, comp_row, comp_col,
                          wp->w_height_outer, wp->w_width_outer, valid, false);
         ui_check_cursor_grid(wp->w_grid_alloc.handle);
-        wp->w_grid_alloc.mouse_enabled = wp->w_config.mouse;
+        wp->w_grid_alloc.mouse_enabled = wp->w_config.mouse != kWinMouseIgnore;
         if (!valid) {
           wp->w_grid_alloc.valid = false;
           redraw_later(wp, UPD_NOT_VALID);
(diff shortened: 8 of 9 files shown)


src/nvim/api/win_config.c
```diff
@@ -130,6 +131,14 @@
 ///   - focusable: Enable focus by user actions (wincmds, mouse events).
 ///       Defaults to true. Non-focusable windows can be entered by
 ///       |nvim_set_current_win()|.
+///   - mouseevent: Specify how this window affects mouse events.
```
justinmk, line 134: Normally callbacks should be `on_xx`, but if `on_mouse` is too awkward here, we can just call it `mouse`.
```suggestion
///   - mouse: Specify how this window affects mouse events. Defaults to `focusable` value.
```

src/nvim/api/win_config.c
```diff
@@ -130,6 +131,14 @@
...
+///       Maybe a boolean or a Lua callback.
+///       - If false, mouse events ignore this window.
+///       - If true, mouse events interact with this window normally.
+///       - If a Lua callback, mouse events interact with this window,
```
justinmk, line 138: Somewhat off-topic but: RPC clients could easily implement their own "callbacks" if our API "callback" fields had a string form invoked as Lua which received a channel-id. 

Example RPC call:

    nvim_open_win(... { mouse = 'local ev = ...; vim.rpcrequest(ev.chan, "my_method")' })

runtime/doc/api.txt
```diff
@@ -3196,6 +3196,16 @@ nvim_open_win({buffer}, {enter}, {config})                   *nvim_open_win()*
...
+                  • mouse: Specify how this window affects mouse events. May
+                    be a boolean or a Lua callback.
+                    • If false, mouse events ignore this window.
+                    • If true, mouse events interact with this window
+                      normally.
+                    • If a Lua callback, mouse events interact with this
+                      window, but the callback is called in place of the
+                      default handling. If this field is not specified but
+                      `focusable` is specified, the value of `focusable` will
+                      be used.
```
justinmk, line 3208: the "If this field..." should go near the top

errael:
Tried a variety of cases, looks like 
```
vim.api.nvim_win_get_config(wid).mouse
```
is alway `nil`.

Wanted to look at how `mouse` changed when `focusable` is changed given different starting conditions.

src/nvim/ui_compositor.c
```diff
@@ -275,7 +275,7 @@ ScreenGrid *ui_comp_mouse_focus(int row, int col)
 {
   for (ssize_t i = (ssize_t)kv_size(layers) - 1; i > 0; i--) {
     ScreenGrid *grid = kv_A(layers, i);
-    if (grid->focusable
+    if (grid->mouse
```
errael, line 278: Wonder if this should be something like
```
grid->mouse != kWinMouseIgnore
```

zeertzjq: `WinMouseEvent` is only used in `w_config`. In `ScreenGrid` a simple boolean is used. I've renamed that to `mouse_enabled` to avoid confusion.

runtime/doc/api.txt
```diff
@@ -3196,6 +3196,15 @@ nvim_open_win({buffer}, {enter}, {config})                   *nvim_open_win()*
                   • focusable: Enable focus by user actions (wincmds, mouse
                     events). Defaults to true. Non-focusable windows can be
                     entered by |nvim_set_current_win()|.
+                  • mouse: Specify how this window affects mouse events. May
+                    be a boolean or a Lua callback. Defaults to `focusable`
+                    value.
```
zeertzjq, line 3201: It may be hard to describe the default value here:
- When `mouse` was `true` or `false`, and `focusable` is specified in `nvim_win_set_config` while `mouse` is not, `mouse` needs to be set to the same value as `focusable` for backward-compatibility.
- However, when `mouse` was a callback, it's better not to change it when `focusable` is specified in `nvim_win_set_config`.

errael: I was originally thinking that it would be good to track if `mouse` had ever been manually set, and if it had been then don't automatically change it to follow `focusable`. But doing as you describe is nice and simple.

It means that if you ever manually set `mouse` to something different than `focusable`, it is probably best to always set it when setting `focusable`. That's reasonable and simple.

I guess it's friendly to not automatically change `mouse` when it's a callback; but it doesn't seem to offer that much and makes things more difficult to describe. Instead of special casing callback, in `nvim_win_set_config` could just say something like:
"If `focusable` is specified and `mouse` is not, then `mouse` is set to the same value as `focusable`."

zeertzjq:
The callback is now invoked for floating windows, although it doesn't have arguments yet.

It isn't hard to enable the callback for non-floating windows, but maybe that should be addressed together with #29365.

errael:
> The callback is now invoked for floating windows, although it doesn't have arguments yet

Looking at the output of `getmousepos()` can guess at the argument :) Mouse press event, mouse release event. Drag event if holding the mouse down and move across floating window. No mouse events if sweep across floater until `set mousemoveevent`.

Implementing window dragging will have problems. If click on left border, then move left, there is no event, drag or otherwise. Could leave an empty cell outside the border and don't move too fast (width not compatible). Something to consider later. (some systems might keep delivering drag event even after the window is left, and there's exit window events)

> (but `getmousepos()` 

Yes, been working for a few pushes

> and `getcharmod()` should work).

Ah, so that's how you get that info. Cool. Yes, that works. Too bad the state of the buttons is not available. Kind of like for `nvim_inpout_mouse`'s `{button}` but a bit mask.
 
> It isn't hard to enable the callback for non-floating windows, but maybe that should be addressed together with #29365.

Popups don't need this AFAIK. 

Interestingly, in simple cases it could be part of a workaround for not getting drag events past the edge or window leave events. Would require temporarily, while dragging a window, swapping into the popup's surrounding window a helper `mouse` function to track movement out of the popup. But in general much more complicated because of overlapping windows.

Maybe a "global" mouse function?

errael:
Here's another possibility. Say the signiture of the mouse callback is something like
```
function mouse_callback(win_id: int, within_bounds: bool, event_id)
```
Currently the callback is only invoked when `within_bounds` is true. If it was always invoked and there was a method `get_event_info(win_id, event_id)` which returned the info like `getmousepos()` and returned stuff like `column_distance` when not within_bounds, positive or negative, giving the distance from the edge. I think if `within_bounds` is false, the event would typically be ignored except when dragging.

Don't know about specific parameters, for example could be only one arg with a single bit for `within_bounds` and win_id could be stuffed in there as well. Don't know what the performance considerations might be. If there's concern about too many event callbacks being invoked, win could have a field for "deliver extra events".

errael:
Was looking at how mouse events work. Wanted to compare regular mappings versus how this PR delivers events to the window through `mouse=callback`. (Had to intuit events for the callback, but it's pretty clear).

Discovered that if there is a regular mapping, like (and similarly for drag events)
```
vim.keymap.set({'n', 'i'}, '<LeftRelease>', function() mouse_mapping("RELEASE") end)
```
Then there there are no left relese events delivered to the window's mouse callback. The mapping events should not be generated if the event is going to window's mouse handler.

errael:
With regualar drag/release mappings, they track mouse movement even when outide the OS window; something similar is needed for `mouse=callback` events. The events should be delived to the window where the press occured. I didn't look for the code that disposes the event before `do_mouse` is called.

Playing around to correctly deliver drag/release events, the following seems to work. The idea is tag each mouse press with a unique id and associate the window that handles the press with the id. Then when processing drag/release events make sure they go to the window where the event was generated.

This is not meant to be a patch, just an example of partially working code. I think it's sufficient for the press to record both current window handle and current_click_id and not save the click_id in `window_S` like in this example. Even the click_id itself may not be needed, only the handle where the last click occured, but it could be used to generate an internal error message if things are out of whack.

```diff
diff --git a/src/nvim/buffer_defs.h b/src/nvim/buffer_defs.h
--- a/src/nvim/buffer_defs.h
+++ b/src/nvim/buffer_defs.h
@@ -1338,4 +1338,6 @@
   StlClickDefinition *w_statuscol_click_defs;
   // Size of the w_statuscol_click_defs array
   size_t w_statuscol_click_defs_size;
+
+  int w_mouse_click_id; // id of click in this window
 };
diff --git a/src/nvim/mouse.c b/src/nvim/mouse.c
--- a/src/nvim/mouse.c
+++ b/src/nvim/mouse.c
@@ -221,20 +221,54 @@
   got_click = false;
 }
 
+static handle_T mouse_click_owner;
+static int current_mouse_click_id;
+
+// Retur the window that has a "mouse" dispatch drag related operation in progress.
+static win_T *drag_mouse(int c)
+{
+  if (c == K_LEFTDRAG || c == K_LEFTRELEASE) {
+    Error err = ERROR_INIT;
+    win_T *wp = find_window_by_handle(mouse_click_owner, &err);
+    if (ERROR_SET(&err)) {
+      DLOG("xyzzy error: %s", err.msg);
+    } else if (wp != NULL && wp->w_mouse_click_id == current_mouse_click_id) {
+      DLOG("xyzzy win_id %d", wp->handle);
+      return wp;
+    }
+  }
+  return NULL;
+}
+
 /// When clicking on a window with a mouse callback, invoke it.
 ///
 /// @return  whether to proceeded with mouse handling.
-static bool invoke_mouse_cb(void)
+static bool invoke_mouse_cb(int c)
 {
   int grid = mouse_grid;
   int row = mouse_row;
   int col = mouse_col;
 
+  if (c == K_LEFTMOUSE) {
+      current_mouse_click_id++;
+  }
+
   if (row < 0 || col < 0) {
     return true;
   }
 
-  win_T *wp = mouse_find_win(&grid, &row, &col);
+  win_T *wp;
+  if ((wp = drag_mouse(c)) == NULL) {
+    wp = mouse_find_win(&grid, &row, &col);
+    if (wp != NULL && c == K_LEFTMOUSE) {
+      mouse_click_owner = wp->handle;
+      wp->w_mouse_click_id = current_mouse_click_id;
+    }
+  }
+  DLOG("xyzzy c = %x, click: %d, win_id %d, has_cb %d",
+       c&0xffff, current_mouse_click_id, wp ? wp->handle : 0,
+       wp->w_config.mouse == kWinMouseCallback);
+
   if (wp == NULL || wp->w_config.mouse != kWinMouseCallback) {
     return true;
   }
@@ -399,7 +433,7 @@
     break;
   }
 
-  if (!invoke_mouse_cb()) {
+  if (!invoke_mouse_cb(c)) {
     return false;
   }
 
@@ -1095,7 +1129,7 @@
 /// of the MSCR_ values.
 void ins_mousescroll(int dir)
 {
-  if (!invoke_mouse_cb()) {
+  if (!invoke_mouse_cb(0)) {
     return;
   }
 
@@ -1608,7 +1642,7 @@
 /// "cap->arg", which is one of the MSCR_ values.
 void nv_mousescroll(cmdarg_T *cap)
 {
-  if (!invoke_mouse_cb()) {
+  if (!invoke_mouse_cb(0)) {
     return;
   }
```

errael:
FYI, see [popup tracking](https://github.com/errael/plenary.nvim/blob/PopupMouse/POPUP.md) for details of the current state of the popup API.

zeertzjq:
> The mapping events should not be generated if the event is going to window's mouse handler.

This can be complicated to achieve, and may require extra restrictions on the callback. Is it needed for a part of the popup API?

errael:
> > The mapping events should not be generated if the event is going to window's mouse handler.
> 
> This can be complicated to achieve, and may require extra restrictions on the callback. Is it needed for a part of the popup API?

If the main app has
```
vim.keymap.set({'n', 'i'}, '<LeftMouse>', function() mouse_mapping("CLICK") end)
```
And a popup is brought up with `close=click`, and `drag=true` (and it's underlying window has `focusable = false, mouse = win_mouse_cb,`). But `win_mouse_cb` is never invoked for `<LeftMouse>`, `<LeftDrag>`, `<LeftRelease>` event because the mapping consumes the event.

So, the answer is yes, it is needed for the popup API.

Could you point out where in the code the mouse event is queued as a key and/or turned into a mapping?

(Maybe more important, is it also needed to have a `nvim_open_win` `mouse` API you can reason about?)

An actual example of this is seen with [Splice9](https://github.com/errael/splice9). All actions have keyboard shortcuts, but the mouse can click on commands displayed in the `HUD` (heads up display). There is use of popups in several situations, a non-modal popup typically has click to close.

As seen in `Splice9`, also note that a clickable command is highlighted when the mouse rolls over it.

HUD-only-partial

errael:
> > The mapping events should not be generated if the event is going to window's mouse handler.
> 
> This can be complicated to achieve, and may require extra restrictions on the callback. 

Could you elaborate on the "extra restrictions"?

zeertzjq:
> > > The mapping events should not be generated if the event is going to window's mouse handler.
> > 
> > 
> > This can be complicated to achieve, and may require extra restrictions on the callback. Is it needed for a part of the popup API?
> 
> If the main app has
> 
> ```
> vim.keymap.set({'n', 'i'}, '<LeftMouse>', function() mouse_mapping("CLICK") end)
> ```
> 
> And a popup is brought up with `close=click`, and `drag=true` (and it's underlying window has `focusable = false, mouse = win_mouse_cb,`). But `win_mouse_cb` is never invoked for `<LeftMouse>`, `<LeftDrag>`, `<LeftRelease>` event because the mapping consumes the event.
> 
> So, the answer is yes, it is needed for the popup API.

That doesn't explain why the mapping shouldn't consume the event. Does the popup API work like that?

> Could you point out where in the code the mouse event is queued as a key and/or turned into a mapping?

Usually, it's `nvim_input` or `nvim_input_mouse` invoked from `state_enter() -> input_get() -> inbuf_poll() -> loop_poll_events() -> loop_uv_run() -> uv_run() -> receive_msgpack() -> parse_msgpack() -> handle_request() -> request_event()` that queue the mouse event as a key. But when Nvim is busy, `loop_poll_events()` can be called by an `os_breakcheck()`.

Mappings are applied in `vgetorpeek() -> handle_mapping()`, which is usually called from `state_enter() -> vpeekc()` or `state_enter() -> safe_vgetc() -> vgetc()`.

Mouse event handling is invoked from `state_enter() -> normal_execute() -> nv_mouse() -> do_mouse()` or `state_enter() -> insert_execute() -> insert_handle_key() -> ins_mouse() -> do_mouse()`.

> Could you elaborate on the "extra restrictions"?

Depends on the implementation.

errael:
> That doesn't explain why the mapping shouldn't consume the event. Does the popup API work like that?

Oops, my bad, the app `splice9` uses a buffer only mapping for the HUD. After adjusting my test... If the focus is not in the HUD, as is typical, then when the popup is clicked, even if the clicked popup spot is over the HUD, the popup closes as expected. But if the focus is in the HUD, then clicking the popup (no matter what's under the mouse under the popup) does not close the popup, and it does seem the event is consumed.

Now that I've got a better idea of what's going on, I have some more things to look at.

Hmm, and I'll have to run some tests around `mousemoveevent` and `<MouseMove>` mapping. Since vim's popup's `mousemoved = "any`` works even with the `<MouseMove>` globally mapped. I see vim's `mouse.c::do_mouse()` has `popup_handle_mouse_moved()`. Probably part of some special casing.

I haven't tried a modal popup, which depends on the popup callback `filter` that looks at all the keystrokes.

`filter` is probably the biggest issue after the mouse.

Is there a way to get, or approximate, that?

-----

Thanks for the tour of mouse input and mapping (I probably should have set a brkpt). But now that I realize that mappings have priority over popup mouse interactions, hopefully I won't want to take a look.

BTW, using `get_special_key_name(c, 0)`, instead of NULL, when doing
```
  Object res = nlua_call_ref(wp->w_config.mouse_cb, get_special_key_name(c, 0),
                             (Array)ARRAY_DICT_INIT, kRetNilBool, NULL, &err);
```
is convenient for dbg printing the event instead of a hex value:)

errael:
For the popup option `mousemoved = "any"`. In `vim` there is a side channel to inform the popup code that the mouse has moved.

In `neovim`, as things are now, the popup code saves the current `&mousemoveevent` and turns it on. When the last popup with `mousemoved = "any"` closes, `&mousemoveevent` is restored. Saving, using, and restoring `<MouseMove>` mapping is also needed. Depending on how `filter = callback` works in `neovim`, it might be possible to avoid changing the mapping.

The same idea should work for the other `mousemoved` options.

In my app the idea is doing a modal dialog/alert; popup a message, close the popup if there is any user action. There are two parts to this
- `mousemoved = "any"`
- `filter = FilterCloseAnyKey` which does `popup_close(wid)`

Note that the filter catches mouse events that come in for mapping; so a mouse event may close the popup either as `mousemoved = "any"` or by the filter.

-----

About `mousemoved = "any"` in vim. Here's how things play out depending on how things are configured. There's an interesting interaction, see (5) (6) below.

```vim
popup_wid = popup_create(['simple', '=stuf='], {
    mousemoved: "any",
    #A# filter: FilterCloseAnyKey,
})
#B# &mousemoveevent = true
#C# nnoremap <special> <MouseMove>   <ScriptCmd>MM()<CR>
```
1. only `mousemoved: "any"`
   move the mouse, popup closes because of the `"any"`
2. Include `#B# &mousemoveevent = true`
   move the mouse, popup closes because of the `"any"`
3. add `<MouseMove>` mapping. Include `#B#` and `#C#`, 
   move the mouse, popup **doesn't** close
   mapped <MouseMove> events are delivered, the "any" trigger doesn't fire
4. add `<MouseMove>` mapping. Include `#A#`, `#B#` and `#C#`, 
   move the mouse, popup closes because of `FilterCloseAnyKey`, not by "any"
   the mouse event is delivered as a key

5. add `<MouseMove>` mapping without `mousemoved: "any"`; without `#B# &mousemoveevent = true`
   no mapping events occur as expected
6. add `<MouseMove>` mapping; add `mousemoved: "any"`
   surprisingly, the mapping events occur; while popup is active with `mousemoved: "any"` it is as though `&mousemoveevent = true`. The popup doesn't close.

errael:
(Off Topic) @zeertzjq I see `vim.on_key()`, which, at first glance, is probably the way to implement `popup filter`. But unlike a `popup filter` there seems to be no way to consume the key or otherwise prevent normal processing.

zeertzjq:
I modified the code so that drag/release events are sent to the callback for the window that received the previous click. However, there is a problem: When the window that received the previous click cannot be found, it's not clear what the behavior should be. It's also hard to decide the arguments for the callback.

I'm also starting to doubt whether the callback is needed at all, as after #30741 is implemented, `vim.on_key()` along with `mouse = true` can kind of already achieve what the callback does. I may move the `mouse = true` and `mouse = false` parts to a separate PR soon.

errael:
> I modified the code so that drag/release events are sent to the callback for
> the window that received the previous click. However, there is a problem: When
> the window that received the previous click cannot be found, it's not clear
> what the behavior should be. 

My inclination is to throw away the events associated with a closed window.
Otherwise there's drag/release events sent to a window that never received a
button press event. Good defensive programming should avoid issues, but still
seems strange.

> It's also hard to decide the arguments for the callback.

The event name and doing `getmousepos()` satisfied my requirements for popup
handling. And passing a string with `nlua_call_ref()` required no effort...

As a related example, I put together a table so I could print/log the params
from `vim.on_key()` It looks like
```
mev_map = {
    ["\128\253\047"] = "<MiddleMouse>",
    ["\128\253\048"] = "<MiddleDrag>",
    ...
}
```
and doing `mev_map[key] or key` to get something printable/decipherable.

I guess it comes down to what/how to expose to the lua code. Passing in the
pieces `is_click`, `is_drag`, `is_scroll`, `button` seems to complicate things,
but then I hadn't considered scroll, maybe the details are needed for
completeness. But it seems like what you want is `<LeftDrag>`, `<RightRelease>`,
and these would have to be reconstructed by the lua code from the pieces.

> I'm also starting to doubt whether the callback is needed at all, as after
> #30741 is implemented, `vim.on_key()` along with `mouse = true` can kind of
> already achieve what the callback does.

I haven't tried out this latest PR and using it in combination with
`vim.on_key()`. Before looking, I'm wondering about resolving interactions;
The key events are global and the mouse events are window. If there's multiple
floating windows, and multiple on_key handlers. All the handlers get all the
events; it's not clear to me how a given on_key handler knows which window the
event is associated with. The floating windows might be from different plugins.

I can't say for sure there's a problem that can't be handled, but...

The mouse window callback seems to get the information to where it's needed
without a lot of extra work.

> I may move the `mouse = true` and `mouse = false` part to a separate PR soon
> (#30844), as there is not as much indecision regarding those.

A note on popup and `vim.on_key()` handling. The popup API about a key filter
says give the keys to the windows in zindex order and stop if the key is
discarded. The popup code has a `vim.on_key()` handler for each popup that has a
filter (simpler code). But considering the zindex ordering and discard behavior
with the popup spec, I'll probably change it to have a single `vim.on_key()`
handler for popups and do the zindex ordering and short circuit in the popup
code. Relates to the semantics of how things work.

zeertzjq:
> My inclination is to throw away the events associated with a closed window.
Otherwise there's drag/release events sent to a window that never received a
button press event.

I also thought so. Updated the PR.

errael:
Using the updated/latest PR, slightly modified to provide the mouse event to the callback
```diff
diff --git a/src/nvim/mouse.c b/src/nvim/mouse.c
--- a/src/nvim/mouse.c
+++ b/src/nvim/mouse.c
@@ -235,1 +235,1 @@
-static bool handle_mouse_cb(bool is_click, bool is_drag, bool is_scroll, int button)
+static bool handle_mouse_cb(bool is_click, bool is_drag, bool is_scroll, int button, int c)
@@ -258,2 +258,2 @@
-  Object res = nlua_call_ref(wp->w_config.mouse_cb, NULL, (Array)ARRAY_DICT_INIT,
-                             kRetNilBool, NULL, &err);
+  Object res = nlua_call_ref(wp->w_config.mouse_cb, get_special_key_name(c, 0),
+                             (Array)ARRAY_DICT_INIT, kRetNilBool, NULL, &err);
@@ -417,1 +417,1 @@
-  if (!handle_mouse_cb(is_click, is_drag, false, which_button)) {
+  if (!handle_mouse_cb(is_click, is_drag, false, which_button, c)) {
@@ -1113,1 +1113,1 @@
-  if (!handle_mouse_cb(false, false, true, dir)) {
+  if (!handle_mouse_cb(false, false, true, dir, 0)) {
@@ -1626,1 +1626,1 @@
-  if (!handle_mouse_cb(false, false, true, cap->arg)) {
+  if (!handle_mouse_cb(false, false, true, cap->arg, 0)) {
```


I think of `neovim` as a window manager when it comes to delivering events to floating windows. Using `ubuntu` with `gnome`, when a button goes down in a window, and the mouse moves outside the window, the events continue to go to the original window; the one where the first click occurred. While outside the window, click the other button, toggle between buttons (while always keeping at least one button down). Events continue to go to the first window.

I'm guessing that other window managers handle events in the same way. Using this common model in `neovim` seems good.

No matter what model is used, the following shows how this PR starts sending drag events to the wrong window.

The trace below was generated by
1. source the lua script below; see 3 floating windows
2. left click in top window
3. drag into the 2nd window
4. right click while still holding down the left button
5. more drag
6. release right
7. more drag
8. release left
9. Enter `:mess` to see the trace
Observe that the click in step 4 is delivered to the wrong callback.
Observe that all the subsequent `<LeftDrag>` events go to the wrong callback.

```
                | window callback that gets the event.
                V
MOUSE WINCB 31 1002 <LeftMouse>: 0x0, line 9, col 50, wid 1002, winrow 3, wincol 20                 
MOUSE WINCB 32 1002 <LeftDrag>: 0x0, line 10, col 50, wid 1002, winrow 4, wincol 20                 
...
MOUSE WINCB 37 1002 <LeftDrag>: 0x0, line 12, col 51, wid 1003, winrow 2, wincol 17                 
MOUSE WINCB 38 1003 <RightMouse>: 0x0, line 12, col 51, wid 1003, winrow 2, wincol 17               
MOUSE WINCB 39 1003 <LeftDrag>: 0x0, line 12, col 52, wid 1003, winrow 2, wincol 18                 
...
MOUSE WINCB 50 1003 <LeftDrag>: 0x0, line 16, col 58, wid 1004, winrow 2, wincol 20                 
MOUSE WINCB 51 1003 <RightRelease>: 0x0, line 16, col 58, wid 1004, winrow 2, wincol 20             
MOUSE WINCB 52 1003 <LeftDrag>: 0x0, line 16, col 57, wid 1004, winrow 2, wincol 19                 
...
MOUSE WINCB 58 1003 <LeftDrag>: 0x0, line 16, col 51, wid 1004, winrow 2, wincol 13                 
MOUSE WINCB 59 1003 <LeftRelease>: 0x0, line 16, col 51, wid 1004, winrow 2, wincol 13              
```


```lua
local count_mouse = 0
local function get_mp(tag)
    count_mouse = count_mouse + 1
    local mp = vim.fn.getmousepos()
    local mod = vim.fn.getcharmod()
    vim.print(string.format("MOUSE %s: 0x%x, line %d, col %d, wid %d, winrow %d, wincol %d",
              tag, mod, mp.screenrow, mp.screencol, mp.winid, mp.winrow, mp.wincol))
    return mp
end

local function create_mouse_cb(wid)
    return function(name)
        get_mp(string.format("WINCB %d %d %s", count_mouse, wid, name))
        return false
    end
end

local xpos = 30
local ypos = 6
local function Pop1Any()

    local popup_wid = vim.api.nvim_open_win(bnr, false, {
        relative = "editor",
        style = "minimal",
        width = 30,
        height = 4,
        focusable = false,
        col = xpos,
        row = ypos,
    })
    vim.api.nvim_buf_set_lines(bnr, 0, -1, true, {'simple', '=' .. tostring(popup_wid) .. '='})
    vim.api.nvim_win_set_config(popup_wid, {mouse = create_mouse_cb(popup_wid)})
    xpos = xpos + 4
    ypos = ypos + 4
end

Pop1Any()
Pop1Any()
Pop1Any()
```

errael:
> It's also hard to decide the arguments for the callback.

Including at least the **event** and the target **win_id** are useful. The target win_id allows one callback to be used for multiple windows in some circumstances.

errael:
I've run a few more tests. Here's one on `ubuntu`/`gnome`. In a `gnome-terminal` running `neovim` do LeftClick; drag out of `neovim` window; RightClick (Left still down) and see popup in `neovim`; RightRelease; drag mouse; do RightClick/RightRelease, see popup in different location; keep the right button down; move the mouse around (still outside of `neovim` window) and LeftClick; observe that the cursor in the `neovim` window moves around.

With the following added at the beginning of `handle_mouse_cb()` the mouse events are locked to the window where the first click occurs. The key to this is detecting a click when before there are no mouse buttons down, what the code calls a `fresh_click`.

```
diff --git a/src/nvim/mouse.c b/src/nvim/mouse.c
--- a/src/nvim/mouse.c
+++ b/src/nvim/mouse.c
@@ -241,0 +242,18 @@
+  // Only tracking three mouse buttons.
+  static int buttons_down = 0;
+  if (button <= MOUSE_RIGHT && !is_drag) {
+    if (is_click) {
+      buttons_down |= (1 << button);
+    } else {
+      buttons_down &= ~(1 << button);
+    }
+  }
+  int n_buttons = 0;
+  for (int mask = 1; mask <= (1 << MOUSE_RIGHT); mask <<= 1) {
+    if (buttons_down & mask) {
+      ++n_buttons;
+    }
+  }
+  bool fresh_click = is_click && n_buttons == 1 && button <= MOUSE_RIGHT;
+  DLOG("%s: n_buttons %d, fresh_click %d", get_special_key_name(c, 0), n_buttons, fresh_click);
+
@@ -243,1 +261,1 @@
-  if (!is_click && !is_scroll && button != MOUSE_RELEASE) {  // drag or release
+  if (!fresh_click && !is_scroll && button != MOUSE_RELEASE) {  // drag or release
@@ -249,1 +267,1 @@
-    if (is_click) {
+    if (fresh_click) {
```

But the bad news is that the popup code needs something similar to find fresh clicks. If there were a "release left no more buttons pressed" event, then that would simplify things a lot. In popup handling, might use `Ctrl-C` to reset mouse state (a just in case to satisfy my paranoia about lost events); this implies that always need an on_key handler active to look for `Ctrl-C`, sigh.

I'm beginning to wonder if `mouse=boolean` and have a stream of events might be OK. It puts a burden on the plugin, but it's beginning to look like that burden is there anyway. I haven't tried anything based on using on_key. The importance of `mouse=true` is that the window shows up in getmousepos(). I haven't looked at what "default mouse handling" does with a window other than letting `getmousepos()` notice the window. Maybe `mouse=true` only affects the window if it is also focusable?

But using `vim.on_key()` seems flawed. IIUC, some mouse events could be lost; if that's true then it becomes impossible to know the mouse state (maybe I'm wrong about events getting lost). Maybe the event stream is best through `mouse=function`. Need that `Ctrl-C`.

An additional observation (which does seem reasonable behavior) is that if LeftRelease (while holding the right button down) then it forgets about a drag (not drag events when the LeftClick again, but about the thing being dragged). But the mouse is still locked to the original window, but no more dragging until after both buttons are released. (I'm not 100% sure of that, but that seems like what's happening).

errael:
@zeertzjq It looks like there were a couple update a few hours ago. Any functional changes?

justinmk:
> > It's also hard to decide the arguments for the callback.
> 
> Including at least the **event** and the target **win_id** are useful.

If the args are not obvious, it's usually a good idea for event handlers to just have an `ev: dict` arg that can be expanded in the future.

zeertzjq:
> @zeertzjq It looks like there were a couple update a few hours ago. Any functional changes?

No, it's a rebase on #30844

errael:
> > @zeertzjq It looks like there were a couple update a few hours ago. Any functional changes?
> 
> No, it's a rebase on #30844

Perfect.

I put a flag in the popup code to switch between `mouse=true` and `mouse=callback`.

errael:
This is no longer needed to implement popup compatibility layer.

PR #30544, opened 2024-09-27
feat(defaults): pretty :help headings

justinmk:
Problem:
Headings in :help do not stand out visually.

Solution:
Define a non-standard `@markup.heading.1.delimiter` group and special-case it in `highlight_group.c`.

FUTURE:
This is a cheap workaround until we have https://github.com/neovim/neovim/issues/25718 which will enable:
- fully driven by `vimdoc/highlights.scm` instead of using highlight tricks (`guibg=bg guifg=bg guisp=fg`)
- better support of "cterm" ('notermguicolors')

Before/After:

image

### DIFF
--- a/runtime/ftplugin/help.lua
+++ b/runtime/ftplugin/help.lua
@@ -1,7 +1,7 @@
 -- use treesitter over syntax (for highlighted code blocks)
 vim.treesitter.start()
 
--- add custom highlights for list in `:h highlight-groups`
+-- Add custom highlights for list in `:h highlight-groups`.
 local bufname = vim.fs.normalize(vim.api.nvim_buf_get_name(0))
 if vim.endswith(bufname, '/doc/syntax.txt') then
   require('vim.vimhelp').highlight_groups({
--- a/src/nvim/highlight_group.c
+++ b/src/nvim/highlight_group.c
@@ -301,6 +301,11 @@ static const char *highlight_init_both[] = {
   "default link @tag         Tag",
   "default link @tag.builtin Special",
 
+  // :help
+  // Higlight "===" and "---" heading delimiters specially.
+  "default @markup.heading.1.delimiter.vimdoc guibg=bg guifg=bg guisp=fg gui=underdouble,nocombine ctermbg=NONE ctermfg=NONE cterm=underdouble,nocombine",
+  "default @markup.heading.2.delimiter.vimdoc guibg=bg guifg=bg guisp=fg gui=underline,nocombine ctermbg=NONE ctermfg=NONE cterm=underline,nocombine",
+
   // LSP semantic tokens
   "default link @lsp.type.class         @type",
   "default link @lsp.type.comment       @comment",
--- a/test/functional/treesitter/highlight_spec.lua
+++ b/test/functional/treesitter/highlight_spec.lua
@@ -894,7 +894,43 @@ describe('treesitter highlighting (help)', function()
       [3] = { bold = true, foreground = Screen.colors.Brown },
       [4] = { foreground = Screen.colors.Cyan4 },
       [5] = { foreground = Screen.colors.Magenta1 },
+      title = { bold = true, foreground = Screen.colors.Magenta1 },
+      h1_delim = { nocombine = true, underdouble = true },
+      h2_delim = { nocombine = true, underline = true },
     }
+  end)
+
+  it('defaults in vimdoc/highlights.scm', function()
+    -- Avoid regressions when syncing upstream vimdoc queries.
+
+    insert [[
+    ==============================================================================
+    NVIM DOCUMENTATION
+
+    ------------------------------------------------------------------------------
+    ABOUT NVIM                                                     *tag-1* *tag-2*
+
+    |news|			News
+    |nvim|			NVim
+    ]]
+
+    feed('gg')
+    exec_lua(function()
+      vim.wo.wrap = false
+      vim.bo.filetype = 'help'
+      vim.treesitter.start()
+    end)
+
+    screen:expect({
+      grid = [[
+        {h1_delim:^========================================}|
+        {title:NVIM DOCUMENTATION}                      |
+                                                |
+        {h2_delim:----------------------------------------}|
+        {title:ABOUT NVIM}                              |
+                                                |
+      ]],
+    })
   end)
 
   it('correctly redraws added/removed injections', function()
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -119,6 +119,9 @@ API
 
 DEFAULTS
 
+• Highlighting:
+  • Improved styling of :checkhealth and :help buffers.
+
 • Mappings:
   • |grn| in Normal mode maps to |vim.lsp.buf.rename()|
   • |grr| in Normal mode maps to |vim.lsp.buf.references()|
--- a/runtime/queries/vimdoc/highlights.scm
+++ b/runtime/queries/vimdoc/highlights.scm
@@ -1,9 +1,9 @@
 (h1
-  (delimiter) @markup.heading.1
+  (delimiter) @markup.heading.1.delimiter
   (heading) @markup.heading.1)
 
 (h2
-  (delimiter) @markup.heading.2
+  (delimiter) @markup.heading.2.delimiter
   (heading) @markup.heading.2)
 
 (h3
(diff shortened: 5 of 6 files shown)


runtime/ftplugin/help.lua
```diff
@@ -1,7 +1,13 @@
 -- use treesitter over syntax (for highlighted code blocks)
 vim.treesitter.start()
 
--- add custom highlights for list in `:h highlight-groups`
+-- Higlight "===" and "---" heading delimiters specially.
+vim.cmd[[
```
lewis6991, line 5: Any reason not to use the Lua highlight API here?

lewis6991 (approved):

clason:
That is not the way to go here; we worked hard on standardizing tree-sitter highlight groups.

Instead, you should customize `@heading.delimiter.1.vimdoc`, which only affects the `@heading.delimiter.1` group in `vimdoc` files.

runtime/ftplugin/help.lua
```diff
@@ -1,7 +1,13 @@
...
+  hi @markup.heading.delimiter1.vimdoc guibg=bg guifg=bg guisp=fg gui=underdouble
```
clason, line 6: Any reason you're not simply using `@markup.heading.delimiter.1.vimdoc`, which would be the documented approach instead of a custom "hack"?

clason: (I used to have the `.delimiter` groups but removed them due to pressure. If we really want to make use of them, we should officially reintroduce and document them (and use them also in markdown etc.; would be a lot of extra work for me in nvim-treesitter, though...) 

As they are part of the fallback hierarchy, they would work gracefully.

runtime/ftplugin/help.lua
clason: Or just define these highlight groups in the default scheme?

clason:
Am I missing something here? Wouldn't that still show the `-` and `=` characters, which seem to be missing from your screenshots?

clason:
Should also add a todo about #25718, which will allow this to work without hacks and custom highlight groups.

justinmk:
> Wouldn't that still show the `-` and `=` characters, which seem to be missing from your screenshots?

No because `guibg=bg guifg=bg` hides those, while `guisp` is used for the underline.

Doesn't seem to work with 'notermguicolors' though.

runtime/ftplugin/help.lua
justinmk: > Or just define these highlight groups in the default scheme?

didn't know about `@markup.heading.1.delimiter.vimdoc` , thanks

> Any reason not to use the Lua highlight API here?

more verbose. someone else can do that if they want

clason:
> No because guibg=bg guifg=bg hides those, while guisp is used for the underline.

Ah, I see. I was expecting a `conceal` here, which...

> Doesn't seem to work with 'notermguicolors' though.

should also work in this case.

src/nvim/highlight_group.c
```diff
@@ -301,6 +301,11 @@ static const char *highlight_init_both[] = {
   "default link @tag         Tag",
   "default link @tag.builtin Special",
 
+  // Higlight "===" and "---" heading delimiters specially.
+  // This is for :help but also presumably useful for markdown etc.
+  "default @markup.heading.1.delimiter guibg=bg guifg=bg guisp=fg gui=underdouble,nocombine ctermbg=NONE ctermfg=NONE cterm=underdouble,nocombine",
+  "default @markup.heading.2.delimiter guibg=bg guifg=bg guisp=fg gui=underline,nocombine ctermbg=NONE ctermfg=NONE cterm=underline,nocombine",
```
justinmk, line 307: Did not specify `.vimdoc` here.

clason: You should.

runtime/queries/markdown/highlights.scm
```diff
@@ -1,11 +1,11 @@
 ;From MDeiml/tree-sitter-markdown & Helix
 (setext_heading
   (paragraph) @markup.heading.1
-  (setext_h1_underline) @markup.heading.1)
+  (setext_h1_underline) @markup.heading.1.delimiter)
 
 (setext_heading
   (paragraph) @markup.heading.2
-  (setext_h2_underline) @markup.heading.2)
+  (setext_h2_underline) @markup.heading.2.delimiter)
```
justinmk, line 8: Also for markdown headings. Can revert this commit if wanted, but maybe let it bake on master to get hate mail first.

image

src/nvim/highlight_group.c
clason: (Sorry, missed the Markdown change here.)

I have to admit I'm not happy about diverging from nvim-treesitter here (especially for Markdown) since this will make syncing changes harder.

Consider introducing these as official highlight groups so I can backport the changes.

clason: (But note that heading delimiters in Markdown could also be leading `#`s.)

justinmk: I'm confused, I thought https://github.com/neovim/neovim/pull/30544#discussion_r1778611134 indicated that `@markup.heading.1.delimiter.vimdoc` was the standard?

What are we diverging from? Is `@markup.heading.2.delimiter` a new group?

clason: Yes.

clason: To be clear, the issues here are:
* We can't include these queries in nvim-treesitter without adding the captures to `:h treesitter-highlight-groups`.
* If we do, we need to use them consistently.
* Adding the (generic) `@markup.heading.X.delimiter` highlight group will not work for some Markdown (and other markup) delimiters.

runtime/queries/markdown/highlights.scm
clason: Looks nice, but I would indeed prefer to keep Markdown out of it (for now); the potential impact is too wide. Keeping this focused on vimdoc would be much more manageable (and a clearer improvement).

src/nvim/highlight_group.c
justinmk: Got it. I'll revert the markdown commit, it's not very high-value (until we have https://github.com/neovim/neovim/issues/25718 ).

I'm happy to update the standard delimiters, but maybe this group will be unnecessary after https://github.com/neovim/neovim/issues/25718 , because we can use queries instead of highlight tricks?

clason: > I'm happy to update the standard delimiters, but maybe this group will be unnecessary after https://github.com/neovim/neovim/issues/25718 , because we can use queries instead of highlight tricks?

Yes, exactly. And if we limit this to vimdoc for now, we can keep it internal. (Maybe add a screentest, though, to make sure I don't accidentally regress when syncing queries...)

test/functional/treesitter/highlight_spec.lua
```diff
@@ -894,9 +894,45 @@ describe('treesitter highlighting (help)', function()
       [3] = { bold = true, foreground = Screen.colors.Brown },
       [4] = { foreground = Screen.colors.Cyan4 },
       [5] = { foreground = Screen.colors.Magenta1 },
+      title = { bold = true, foreground = Screen.colors.Magenta1 },
+      h1_delim = { nocombine = true, underdouble = true },
+      h2_delim = { nocombine = true, underline = true },
     }
   end)
 
+  it('defaults in vimdoc/highlights.scm', function()
```
justinmk, line 903: added test

test/functional/treesitter/highlight_spec.lua
```diff
@@ -894,9 +894,45 @@ describe('treesitter highlighting (help)', function()
...
+    -- Avoid regressions when syncing upstream vimdoc queries.
```
justinmk, line 904: added test

src/nvim/highlight_group.c
```diff
@@ -301,6 +301,11 @@ static const char *highlight_init_both[] = {
   "default link @tag         Tag",
   "default link @tag.builtin Special",
 
+  // Higlight "===" and "---" heading delimiters specially.
+  // This is for :help but also presumably useful for markdown etc.
+  "default @markup.heading.1.delimiter guibg=bg guifg=bg guisp=fg gui=underdouble,nocombine ctermbg=NONE ctermfg=NONE cterm=underdouble,nocombine",
```
clason, line 306: One issue I see with this is that this will not play nicely with other colorschemes since the `guisp` is hard-coded (and likely different from the `@markup.heading.1` group).

(And since it's a "secret" highlight group, colorschemes will not override it.)

clason, line 306:
```suggestion
  "default @markup.heading.1.delimiter.vimdoc guibg=bg guifg=bg guisp=fg gui=underdouble,nocombine ctermbg=NONE ctermfg=NONE cterm=underdouble,nocombine",
```

clason: Now that it's specific to vimdoc, we should be targeted in the group definition.

justinmk: > since the `guisp` is hard-coded

how is it any more "hard coded" than any other `default` highlight attribute?

PR #30555, opened 2024-09-28
fix(float): through an error when set current window to a hidden floating window

glepnir:
Problem: cursor will jump into a hidden floating window when using `nvim_set_current_win` to switch.

Solution: through an error in nvim_set_current_win when window is a hidden floating window

### DIFF
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -949,6 +949,12 @@ void nvim_set_current_win(Window window, Error *err)
     return;
   }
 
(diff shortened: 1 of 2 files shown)


bfredl:
Not sure I agree with this. This API is meant to be "I know what I am doing, let me do it". If a plugin wants to jump to one of it hidden windows, it should be assumed that it has a good reason to doing so (and if it doesn't make sense, the plugin should just not do it).

glepnir:
so we

PR #30556, opened 2024-09-28
fix(column): set signcolumn width after splitting window

### LINKED ISSUE #30547: signcolumn is not updated immediately after split
### Problem

> reproduce steps might be more understandable

I was going to get a full gutter width by `vim.fn.getwininfo(win)[1].textoff`.

But even if I set `number signcolumn=yes`, the value of `textoff right after the `:vsplit` is `4` and not `6`.

I can see it is 6 because of signcolumn width and when I manually retrieve the `textoff` value from command line, it says it is `6`.
But neovim says it is `4` right after the `WinEnter` event.

> [!NOTE]
> this doesn't happen from vim

### Steps to reproduce

**repro.lua**:
```lua
-- to show 'number' option is working
vim.o.number = true
vim.o.signcolumn = "yes"
vim.cmd.wincmd("v")
local infos = vim.fn.getwininfo()
-- new window
vim.print(infos[1].winid, infos[1].textoff)
-- original window
vim.print(infos[2].winid, infos[2].textoff)
```

1. `nvim --clean -u repro.lua`
2. the output is like this:
    ```
    1001
    4
    1000
    6
    ```

> This doesn't happen from vim. Tested with this script:
> **repro.vim**
> ```vim
> set number
> set signcolumn=yes
> 
> wincmd v
> 
> let infos = getwininfo()
> echom infos[0].textoff
> echom infos[1].textoff
> ```
> Running `vim -u repro.vim` prints:
> ```
> 10
> 10
> ```

### Expected behavior

Two windows should have same `textoff` value right after the split.
So output should be like this:
```
1001
6
1000
6
```

### Nvim version (nvim -v)

0.10.1

no, vim 8.2

### Operating system/version

ubuntu 22.04

### Terminal name/version

blink shell

xterm-256-color

### Installation

build from repo

zeertzjq:
Fix #30547

### DIFF
--- a/src/nvim/optionstr.c
+++ b/src/nvim/optionstr.c
@@ -306,7 +306,7 @@ int check_signcolumn(win_T *wp)
       wp->w_minscwidth = 0;
       wp->w_maxscwidth = 1;
     }
-    return OK;
+    goto done;
   }
 
   if (strncmp(val, "auto:", 5) != 0
@@ -326,6 +326,11 @@ int check_signcolumn(win_T *wp)
 
   wp->w_minscwidth = min;
   wp->w_maxscwidth = max;
+
+done:
+  ;
+  int scwidth = wp->w_minscwidth <= 0 ? 0 : MIN(wp->w_maxscwidth, wp->w_scwidth);
+  wp->w_scwidth = MAX(wp->w_minscwidth, scwidth);
   return OK;
 }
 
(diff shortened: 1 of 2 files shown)


lewis6991 (approved):
Is there anyway this can be done without the goto?

zeertzjq:
Yes

PR #30558, opened 2024-09-28
refactor(event): change last use of klist to kvec

bfredl:
`loop->children` might have been a linked list because used to be modified in place while looped over. However the loops that exists rather schedules events to be processed later, outside of the loop, so this can not happen anymore.

When a linked list is otherwise useful it is better to use `lib/queue_defs.h` which defines an _intrusive_ linked list (i e it doesn't need to do allocations for list items like klist ).


src/nvim/event/proc.c
```diff
@@ -123,16 +122,16 @@ int proc_spawn(Proc *proc, bool in, bool out, bool err)
   proc->internal_exit_cb = on_proc_exit;
   proc->internal_close_cb = decref;
   proc->refcount++;
-  kl_push(WatcherPtr, proc->loop->children, proc);
+  kv_push(proc->loop->children, proc);
```
justinmk, line 125: nice improvement, the WatcherPtr indirection was clumsy.

src/nvim/event/proc.c
```diff
@@ -294,15 +293,19 @@ static void decref(Proc *proc)
...
+  size_t i;
+  for (i = 0; i < kv_size(loop->children); i++) {
+    Proc *current = kv_A(loop->children, i);
+    if (current == proc) {
       break;
     }
   }
-  assert(node);
-  kl_shift_at(WatcherPtr, loop->children, node);
+  assert(i < kv_size(loop->children));  // element fournd
```
justinmk, line 303:
```suggestion
  assert(i < kv_size(loop->children));  // element found
```

justinmk (approved):

PR #30559, opened 2024-09-28
feat(treesitter): include capture id in return value of `get_captures_at_pos()`

ribru17:
**Problem:** Currently, it is difficult to get node(s)-level metadata for a capture returned by `get_captures_at_pos()`. This is because it is stored in `metadata[id]` and we do not have access to the value of `id`, so to get this value we have to iterate over the keys of `metadata`. See [this commit](https://github.com/neovim/neovim/commit/d63622930001b39b12f14112fc3abb55b760c447#diff-8bd4742121c2f359d0345f3c6c253a58220f1a28670cc4e1c957992232059a6cR16).

Things would be much simpler if we were given the `id` of the capture so we could use it to just index `metadata` directly.

**Solution:** Include `id` in the data returned by `get_captures_at_pos()`

**Examples:**

Under a normal query like:
```scm
((function_call) @_fn
  (#set! @_fn "test" 789))
```

The returned capture will look like this:
```lua
{                                                                                                                                              
  capture = "_fn",                                                                                                                             
  lang = "lua",                                                                                                                                
  metadata = {                                                                                                                                 
    [39] = {                                                                                                                                   
      test = "789"                                                                                                                             
    }                                                                                                                                          
  }                                                                                                                                            
}
```
only the metadata for this capture (of id 39) is returned, which is good; however, it's still annoying that we don't know the id for the capture, because it means we have to iterate over the keys of `metadata` and access the value that way.

```lua
for _, md in pairs(cap.metadata) do
  vim.print(md)
  break
end
```

Things are even worse if we have a query like the following:

```scm
((number) @_num @_numagain
  (#set! @_num "field1" 15)
  (#set! @_numagain "field2" 88))
```

Now the data returned by `get_captures_at_pos()` is:

```lua
{ {                                                                                                                                           
  capture = "_num",                                                                                                                            
  lang = "lua",                                                                                                                                
  metadata = <1>{                                                                                                                              
    [41] = {                                                                                                                                   
      field1 = "15"                                                                                                                            
    },                                                                                                                                         
    [42] = {                                                                                                                                   
      field2 = "88"                                                                                                                            
    }                                                                                                                                          
  }                                                                                                                                            
}, {                                                                                                                                           
  capture = "_numagain",                                                                                                                       
  lang = "lua",                                                                                                                                
  metadata = <table 1>                                                                                                                         
} }
```

In this case it would be impossible to, say, write a function to get all metadata information at a position (which would be useful for comment metadata & `_get_urls`) without counting duplicate metadata fields. We don't know the capture ids for `_numagain` or `_num` so we have to find them using a `for` loop, yet their metadata tables are identical so we can't properly extract the metadata for just that capture. Of course, this example is a bit convoluted, but we should be able to be robustly extract data in cases like this regardless, which would be possible only if we know the id for each specific capture, which we can then use to index into (potentially duplicate) metadata tables.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -289,6 +289,7 @@ TREESITTER
   false, which allows it to return anonymous nodes as well as named nodes.
 • |treesitter-directive-trim!| can trim all whitespace (not just empty lines)
   from both sides of a node.
+• |vim.treesitter.get_captures_at_pos()| now returns the `id` of each capture
 
 TUI
 
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -902,8 +902,8 @@ get_captures_at_pos({bufnr}, {row}, {col})
     Returns a list of highlight captures at the given position
 
     Each capture is represented by a table containing the capture name as a
-    string as well as a table of metadata (`priority`, `conceal`, ...; empty
-    if none are defined).
+    string, the capture's language, a table of metadata (`priority`,
+    `conceal`, ...; empty if none are defined), and the id of the capture.
 
     Parameters: ~
       • {bufnr}  (`integer`) Buffer number (0 for current buffer)
@@ -911,7 +911,7 @@ get_captures_at_pos({bufnr}, {row}, {col})
       • {col}    (`integer`) Position column
 
     Return: ~
-        (`{capture: string, lang: string, metadata: vim.treesitter.query.TSMetadata}[]`)
+        (`{capture: string, lang: string, metadata: vim.treesitter.query.TSMetadata, id: integer}[]`)
 
 get_node({opts})                                   *vim.treesitter.get_node()*
     Returns the smallest named node at the given position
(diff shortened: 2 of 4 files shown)


clason:
Can you give a before/after example to show what would be easier, and by how much?

ribru17:
Sure, I've updated the PR description

justinmk:
does this simplify anything in `vim.inspect_pos` ? https://github.com/neovim/neovim/blob/e049c6e4c08a141c94218672e770f86f91c27a11/runtime/lua/vim/_inspector.lua#L43

> Now the data returned by `get_captures_at_pos()` is:

is that example supposed to have `id = ` fields?

justinmk (approved):

ribru17:
Nah, both those examples are of before this PR, sry shouldve been clearer

ribru17:
> does this simplify anything in `vim.inspect_pos` ?

I don't think so, looks like it just uses the capture and lang names. I don't think it needs id or metadata at all

PR #24070, opened 2023-06-20
fix(lua)!: remove vim.json global options, "null", "array_mt"

justinmk:
## Problem
- `vim.json` exposes various global options which:
  - affect all Nvim Lua plugins (especially the LSP client)
  - are undocumented and untested
  - can cause confusing problems such as: https://github.com/codota/tabnine-nvim/commit/cc76ae3abe2f129d44b5a8edee2529e0ee0dcf69
- `vim.json` exposes redundant mechanisms:
  - `vim.json.null` is redundant with `vim.NIL`.
  - [`array_mt` is redundant because](https://github.com/neovim/neovim/pull/23229#discussion_r1174536667) Nvim uses a metatable
    (`vim.empty_dict()`) for empty dict instead, which `vim.json` is
    configured to use by default (see `as_empty_dict`).
    Example:
    ```
    :lua vim.print(vim.json.decode('{"bar":[],"foo":{}}'))
    --> { bar = {},  foo = vim.empty_dict() }
    ```
    Thus we don't need to also decorate empty arrays with `array_mt`.

## Solution
Remove the functions from the public vim.json interface.
Comment-out the implementation code to minimize drift from upstream.

TODO:
- Expose the options as arguments to `vim.json.new()`

PR #30561, opened 2024-09-28
feat(json): `escape_slash` encoding option

PolyMeilex:
This exposes cjson's `encode_escape_forward_slash ` option to the user, but without global side-effects.
I want to encode a nice human-readable JSON files in my plugin, but escaping every slash (for example in file paths) makes this impossible. 

The only use case I know for forward slash escaping is preventing HTML injections (eg. injecting `</script>` closing tag), so in context of typical nvim plugin it does not matter. (that's why escaping it in JSON is optional)

Probably best reviewed commit by commit:
- `refactor(json): add json encode context type` 
  - Extracts common encoding state to a `json_encode_t` struct (this matches what `json_parse_t` does)
- `feat(json): add call local version of escape_forward_slash`
  - Makes use of the new context struct to implement new `escape_forward_slash` option

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -811,11 +811,14 @@ vim.json.decode({str}, {opts})                             *vim.json.decode()*
     Return: ~
         (`any`)
 
-vim.json.encode({obj})                                     *vim.json.encode()*
+vim.json.encode({obj}, {opts})                             *vim.json.encode()*
     Encodes (or "packs") Lua object {obj} as JSON in a Lua string.
 
     Parameters: ~
-      • {obj}  (`any`)
+      • {obj}   (`any`)
+      • {opts}  (`table<string,any>?`) Options table with keys:
+                • escape_slash: (boolean) (default false) When true, escapes
+                  `/` character in JSON strings
 
     Return: ~
         (`string`)
--- a/test/functional/lua/json_spec.lua
+++ b/test/functional/lua/json_spec.lua
@@ -152,6 +152,45 @@ describe('vim.json.encode()', function()
     clear()
   end)
 
+  it('dumps strings with & without escaped slash', function()
+    -- With slash
+    eq('"Test\\/"', exec_lua([[return vim.json.encode('Test/', { escape_slash = true })]]))
+    eq(
+      'Test/',
+      exec_lua([[return vim.json.decode(vim.json.encode('Test/', { escape_slash = true }))]])
+    )
+
+    -- Without slash
+    eq('"Test/"', exec_lua([[return vim.json.encode('Test/')]]))
+    eq('"Test/"', exec_lua([[return vim.json.encode('Test/', {})]]))
+    eq('"Test/"', exec_lua([[return vim.json.encode('Test/', { _invalid = true })]]))
+    eq('"Test/"', exec_lua([[return vim.json.encode('Test/', { escape_slash = false })]]))
+    eq(
+      '"Test/"',
+      exec_lua([[return vim.json.encode('Test/', { _invalid = true, escape_slash = false })]])
+    )
+    eq(
+      'Test/',
+      exec_lua([[return vim.json.decode(vim.json.encode('Test/', { escape_slash = false }))]])
+    )
+
+    -- Checks for for global side-effects
+    eq(
+      '"Test/"',
+      exec_lua([[
+        vim.json.encode('Test/', { escape_slash = true })
+        return vim.json.encode('Test/')
+      ]])
+    )
+    eq(
+      '"Test\\/"',
+      exec_lua([[
+        vim.json.encode('Test/', { escape_slash = false })
+        return vim.json.encode('Test/', { escape_slash = true })
+      ]])
+    )
+  end)
+
   it('dumps strings', function()
     eq('"Test"', exec_lua([[return vim.json.encode('Test')]]))
     eq('""', exec_lua([[return vim.json.encode('')]]))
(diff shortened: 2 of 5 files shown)


test/functional/lua/json_spec.lua
```diff
@@ -152,6 +152,23 @@ describe('vim.json.encode()', function()
     clear()
   end)
 
+  it('dumps strings with escaped slash', function()
+    eq('"Test\\/"', exec_lua([[return vim.json.encode('Test/')]]))
+    eq('"Test\\/"', exec_lua([[return vim.json.encode('Test/', {})]]))
+    eq('"Test\\/"', exec_lua([[return vim.json.encode('Test/', { __invalid_key = true })]]))
+    eq('"Test\\/"', exec_lua([[return vim.json.encode('Test/', { escape_forward_slash = true })]]))
```
justinmk, line 159: probably makes sense to also test decode()?

    decode(encode(...))

test/functional/lua/json_spec.lua
```diff
@@ -152,6 +152,23 @@ describe('vim.json.encode()', function()
...
+  end)
+
+  it('dumps strings without escaped slash', function()
+    eq('"Test/"', exec_lua([[return vim.json.encode('Test/', { escape_forward_slash = false })]]))
+    eq(
+      '"Test/"',
+      exec_lua(
+        [[return vim.json.encode('Test/', { __invalid_key = true, escape_forward_slash = false })]]
```
justinmk, line 167: all these tests should be in 1 test, especially to check that the option does not affect anything globally (and there should be an explicit check for that)

bfredl:
can `escape_forward_slash=false` be the new default? I thought we were encoding JSON, not JSON-embedded-inside-html. Would LSP servers break if we just send them JSON?

runtime/doc/lua.txt
```diff
@@ -812,11 +812,14 @@ vim.json.decode({str}, {opts})                             *vim.json.decode()*
...
 
-vim.json.encode({obj})                                     *vim.json.encode()*
+vim.json.encode({obj}, {opts})                             *vim.json.encode()*
     Encodes (or "packs") Lua object {obj} as JSON in a Lua string.
 
     Parameters: ~
-      • {obj}  (`any`)
+      • {obj}   (`any`)
+      • {opts}  (`table<string,any>?`) Options table with keys:
+                • escape_forward_slash: (boolean) When true, escapes `/`
```
justinmk, line 821: unnecessarily verbose

```suggestion
                • escape_slash: (boolean) When true, escapes `/`
```

justinmk:
This seems good, and +1 for making it the default (mention this in news.txt).

As a followup, we also need help with merging upstream changes, e.g.: https://github.com/neovim/neovim/pull/24070#issuecomment-1598366646

PolyMeilex:
> can `escape_forward_slash=false` be the new default? I thought we were encoding JSON, not JSON-embedded-inside-html. Would LSP servers break if we just send them JSON?

I would definitely be in favor of that, I have never seen an implementation of JSON that escapes those (except PHP that does it because it was designed to primarily output to HTML). Even JS does not do that.

So, do we keep the option and just set it to `false` by default? Or do we just drop it, and switch it globally?

justinmk:
> So, do we keep the option and just set it to `false` by default?

👍 I didn't look closely, but if you found a good pattern for supporting options in a non-global way, then that seems useful for possibly supporting other options in the future. For reference, we  [removed the cjson global options](https://github.com/neovim/neovim/pull/24070).

PolyMeilex:
> 👍 I didn't look closely, but if you found a good pattern for supporting options in a non-global way, then that seems useful for possibly supporting other options in the future. For reference, we [removed the cjson global options](https://github.com/neovim/neovim/pull/24070).

Yeah, sure, the only reason for this diff being so long is the addition of `json_encode_options_t` that is available in the whole encode call stack for easy addition of new options in the future.

runtime/doc/lua.txt
PolyMeilex: Done

test/functional/lua/json_spec.lua
PolyMeilex: Merged into 1 test

test/functional/lua/json_spec.lua
PolyMeilex: Added two cases for that, decode seems to be up to standard. It decodes `\\/` as `/`.

PolyMeilex:
I believe the comments are addressed, and no escape is now the default.
The failed CI step seems to be a flake?

justinmk:
@PolyMeilex would you mind rebasing this

runtime/doc/lua.txt
```diff
@@ -812,11 +812,14 @@ vim.json.decode({str}, {opts})                             *vim.json.decode()*
...
 
-vim.json.encode({obj})                                     *vim.json.encode()*
+vim.json.encode({obj}, {opts})                             *vim.json.encode()*
     Encodes (or "packs") Lua object {obj} as JSON in a Lua string.
 
     Parameters: ~
-      • {obj}  (`any`)
+      • {obj}   (`any`)
+      • {opts}  (`table<string,any>?`) Options table with keys:
+                • escape_slash: (boolean) When true, escapes `/` character in
```
justinmk, line 821:
```suggestion
                • escape_slash: (boolean) (default false) When true, escapes `/` character in
```

PolyMeilex: Done

PR #30576, opened 2024-09-29
fix(ui): make sure that the screen is updated before waiting for input

fredizzimo:
Neovim can in some cases wait for new input and block even if there is a pending redraw. 

The easiest way to reproduce this is to `:set scrolloff=9999` and `set cursorline` when using multigrid. In this case the `win_viewport` message will not be received, since there are still pending updates, which in practice leads to very jerky motion in Neovide, see https://github.com/neovide/neovide/issues/2794. 

The added tests are also testing this case, and all of them fail without modifications here.

The fix is in theory quite simple, just update the screen if `must_redraw` is set to true. But in practice it needs some more tests. `redraw_later` has two different meanings. Once of them is, redraw the screen as soon as possible, in order to scroll for example. The other meaning is to ask for a redraw of the screen after the next key press, to reset some state for example. The second case is blocked by `!need_wait_return && (State & MODE_CMDLINE) == 0)`, which seems to be all that is needed, but that is purely based on tests, so there might be some case missing.

### DIFF
--- a/src/nvim/state.c
+++ b/src/nvim/state.c
@@ -66,9 +66,11 @@ getkey:
       // Event was made available after the last multiqueue_process_events call
       key = K_EVENT;
     } else {
-      // Duplicate display updating logic in vgetorpeek()
-      if (((State & MODE_INSERT) != 0 || p_lz) && (State & MODE_CMDLINE) == 0
-          && must_redraw != 0 && !need_wait_return) {
+      // Ensure the screen is fully updated before blocking for input. Because of the duality of
+      // redraw_later, this can't be done in command-line or when waiting for "Press ENTER".
+      // In many of those cases the redraw is expected AFTER the key press, while normally it should
+      // update the screen immediately.
+      if (must_redraw != 0 && !need_wait_return && (State & MODE_CMDLINE) == 0) {
         update_screen();
         setcursor();  // put cursor back where it belongs
       }
--- a/test/functional/ui/multigrid_spec.lua
+++ b/test/functional/ui/multigrid_spec.lua
@@ -2599,4 +2599,257 @@ describe('ext_multigrid', function()
     ]])
     eq(1, api.nvim_get_option_value('cmdheight', {}))
   end)
+
+  describe('centered cursorline', function()
+    before_each(function()
+      -- Force a centered cursorline, this caused some redrawing problems described in #30576.
+      -- Most importantly, win_viewport was not received in time, and sum_scroll_delta did not refresh.
+      command('set cursorline scrolloff=9999')
+    end)
+    it('insert line scrolls correctly', function()
+      for i = 1, 11 do
+        insert('line' .. i .. '\n')
+      end
+      screen:expect({
+        grid = [[
+        ## grid 1
+          [2:-----------------------------------------------------]|*12
+          {11:[No Name] [+]                                        }|
+          [3:-----------------------------------------------------]|
+        ## grid 2
+          line1                                                |
+          line2                                                |
+          line3                                                |
+          line4                                                |
+          line5                                                |
+          line6                                                |
+          line7                                                |
+          line8                                                |
+          line9                                                |
+          line10                                               |
+          line11                                               |
+          {23:^                                                     }|
+        ## grid 3
+                                                               |
+        ]], win_viewport={
+        [2] = {win = 1000, topline = 0, botline = 12, curline = 11, curcol = 0, linecount = 12, sum_scroll_delta = 0};
+      }, win_viewport_margins={
+        [2] = {
+          bottom = 0,
+          left = 0,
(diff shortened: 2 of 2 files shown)


src/nvim/state.c
```diff
@@ -66,9 +66,11 @@ void state_enter(VimState *s)
       // Event was made available after the last multiqueue_process_events call
       key = K_EVENT;
     } else {
-      // Duplicate display updating logic in vgetorpeek()
-      if (((State & MODE_INSERT) != 0 || p_lz) && (State & MODE_CMDLINE) == 0
-          && must_redraw != 0 && !need_wait_return) {
+      // Make sure that the screen is fully updated before blocking for new input.
+      // Due to the duality of redraw_later, this can't be done in command line or when we are
+      // waiting for a return key press. In many of those cases the redraw is actually supposed to
+      // happen after the key press, while normally it's used to update the screen immeditely.
```
justinmk, line 72: typo, also can be tightened a bit.

```suggestion
      // Ensure the screen is fully updated before blocking for input. Because of the duality of
      // redraw_later, this can't be done in command-line or when waiting for "Press ENTER".
      // In many of those cases the redraw is actually supposed to happen after the key press,
      // while normally it should update the screen immediately.
```

src/nvim/state.c
```diff
@@ -66,9 +66,11 @@ void state_enter(VimState *s)
       // Event was made available after the last multiqueue_process_events call
       key = K_EVENT;
     } else {
-      // Duplicate display updating logic in vgetorpeek()
-      if (((State & MODE_INSERT) != 0 || p_lz) && (State & MODE_CMDLINE) == 0
```
justinmk, line 70: Why was it checking 'lazyredraw' before doing an eager update?

test/functional/ui/multigrid_spec.lua
```diff
@@ -2599,4 +2599,256 @@ describe('ext_multigrid', function()
     ]])
     eq(1, api.nvim_get_option_value('cmdheight', {}))
   end)
+
+  describe('centered cursorline', function()
+    before_each(function()
+      command('set cursorline')
+      command('set scrolloff=9999')
```
justinmk, line 2606:
```suggestion
      -- These settings simulate / achieve / force ... #30576
      command('set cursorline scrolloff=9999')
```

justinmk:
> `redraw_later` has two different meanings ...

This would be helpful in the docstring for `redraw_later`. 

Also please include your explanation in the commit message.

justinmk (approved):
Removing special cases and CI passes, that's a good sign.

I compared the test run to master, and didn't find any new `warning: Screen changes were received after the expected state. This indicates indeterminism` warnings.

test/functional/ui/multigrid_spec.lua
```diff
@@ -2599,4 +2599,256 @@ describe('ext_multigrid', function()
     ]])
     eq(1, api.nvim_get_option_value('cmdheight', {}))
   end)
+
+  describe('centered cursorline', function()
+    before_each(function()
+      -- These settings simulate / achieve / force ... #30576
+      command('set cursorline scrolloff=9999')
```
justinmk, line 2606: just to be clear: I was hoping you would fill in the rest of the comment :)

fredizzimo:
I addressed the review comments, but I did not add any comments to `redraw_later`. I think it's better that someone, who actually understands what it does takes care of that.

Better yet, if it was refactored so that no special cases are needed at all. I have no idea how hard that would be though, but it seems like it should be possible to request a redraw after the input is processed rather than before.

test/functional/ui/multigrid_spec.lua
fredizzimo: Ok, I will check that again

fredizzimo: Ok, it should be better now :)

bfredl (approved):
no idea what is going on by 'lazyredraw' either before or after (but that's not important to me, eventually  'lazyredraw' feature should be replaced by "redrawing is just always good enough" feature) , but otherwise looks right to me.

justinmk:
I compared the test run to master, and didn't find any new `warning: Screen changes were received after the expected state. This indicates indeterminism` warnings.

After some bake-time, we could put this in 0.10.3 (in case 0.11 gets delayed).

fredizzimo:
It would be great to get this backported to the 0.10.x branch. This also fixes a lot of problems with the built in terminal refresh. It's very noticeable now after I have used the 0.10.3 version without this fix for some time. And the problems in 0.10.3 appears even worse than I remember them being in older version.

Before that version, I only noticed it when running git status, where the screen would refresh after up to one second after it should. But now I also see it when just typing in the terminal, the cursor moves, but the text is invisible until after a while.

Issue #19918, opened 2022-08-24
How should "introspection" functions be exposed?


justinmk:
## Problem

Where should "core introspection" stuff live? Currently we have "introspection" interfaces in these places:

1. `context.c`
2.  `nvim_get_api`,
3.  specific APIs like `nvim_get_functions` and `nvim_get_commands`
4. `runtime/vim/core/funcs.lua` (does not exist yet)
5. others?

Static things like builtin Vimscript functions, ex cmds, and normal commands, can be shipped as Lua files in the runtime:

- https://github.com/neovim/neovim/pull/12040
- https://github.com/neovim/neovim/issues/8029#issuecomment-1175146743
- https://github.com/neovim/neovim/pull/19667#discussion_r953205326

## Proposal

Current thoughts:

1. `context.c` (`nvim_get_context`) is for mutable state that can be _saved and restored_.
    - variables, regs, jumps, bufs, gvars, …
2.  `nvim_get_api` is for static (immutable) API schema.
3.  Things like `nvim_get_functions` and `nvim_get_commands` are for Vimscript 
4. `runtime/vim/core/*.lua` (does not exist yet) has a dual purpose:
    - it defines the core things (e.g. `eval.lua` defines Vimscript function signatures),
    - and because it's (data-only) Lua, why not ship it so that user plugins can `require()` and inspect them?

Issue #18750, opened 2022-05-25
normal commands structured map ("introspection" / "meta")


andmis:
## Feature description

I need a structured version of `index.txt` which lists the available mappings, by mode, plus documentation, to support [a plugin I'm working on](https://github.com/linty-org/key-menu.nvim).

I've done a first pass JSONifying `index.txt`, the JSON is [here](https://github.com/linty-org/key-menu.nvim/blob/fe5ab88f21c14088476d3dd855d9fbd6b60f30a2/index.json). I can/will just include this in my plugin source for now, but:

I'm wondering, is this something we want to upstream? The structured version of this data could be used to generate the `:help` file, and programmatic access to this data would be useful for plugin authors. For example, it could be used to implement a clone of Emacs's `<C-h>k`. If so, does the core team have thoughts about where it should go?

justinmk:
> The structured version of this data could be used to generate the `:help` file, and programmatic access to this data would be useful for plugin authors. For example, it could be used to implement a clone of Emacs's `<C-h>k`.

Those are definitely goals, but index.txt doesn't really get us there. Instead we want to encode this knowledge in :

```
src/nvim/auevents.lua
src/nvim/eval.lua
src/nvim/ex_cmds.lua
src/nvim/options.lua
```

Previous work in this area:

- https://github.com/neovim/neovim/pull/6288
- https://github.com/neovim/neovim/pull/12040
- `scripts/gen_vimdoc.py`

justinmk:
> I've done a first pass JSONifying `index.txt`, the JSON is [here](https://github.com/linty-org/key-menu.nvim/blob/fe5ab88f21c14088476d3dd855d9fbd6b60f30a2/index.json) ... is this something we want to upstream? The structured version of this data could be used to generate the `:help` file,

@andmis we definitely want that for normal-mode commands, which I just noticed is the total contents of your JSON file.

I updated https://github.com/neovim/neovim/issues/19918 with the overall plan.

We now have [_meta/vimfn.lua](https://github.com/neovim/neovim/blob/1ee905a63aa9f3d6873ec75a0954866d8660699e/runtime/lua/vim/_meta/vimfn.lua) and soon will have `_meta/options.lua` both of which contain "annotations" which can be consumed in a structured way.

Details are in https://github.com/neovim/neovim/pull/24493 and https://github.com/neovim/neovim/pull/24528 , but the current state is:

[...]

PR #30580, opened 2024-09-29
docs: improve luacats support

jamestrew:
Some composite/compound types even as basic as `(string|number)[]` are not currently supported by the luacats LPEG grammar used by gen_vimdoc. It would be parsed & rendered as just `string|number`.

Changeset adds better support for these types.

`make doc` was run. Only a few changes made to the docs itself.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -2303,7 +2303,7 @@ Lua module: vim.lsp.rpc                                              *lsp-rpc*
 *vim.lsp.rpc.PublicClient*
 
     Fields: ~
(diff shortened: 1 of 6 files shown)


lewis6991 (approved):
The diff on the grammar is a little big, but the testcases and changes to the docs look good.

justinmk:
I think it's also time we exposed this as a runtime feature, so that the `_meta/*` files can be consumed as a "database" of info which plugins can use. That would be an answer to https://github.com/neovim/neovim/issues/19918 https://github.com/neovim/neovim/issues/18750

Help is welcome here...

PR #30600, opened 2024-09-30
refactor(lsp): convert str_{utf,bye}index to lua and extend

tris203:
TODO: write this properly

But notes
- approx 15% faster
- intentionally fixes some edge cases around multibyte strings #32048


justinmk:
Do we need this or not? I heard something about a 14% speedup, not sure it's worth the trouble.

PR #30602, opened 2024-10-01
fix(tabline): restore behavior of click after last tabpage

### LINKED ISSUE #30599: double click in tab bar does not open a new tab
### Problem

in the documentation for tab pages, it says that you should be able to double click in the tab bar to open a new tab to the right of the last tab.  this is the behavior in vim9, in neovim you must double click an existing tab, and it will open to the left of that tab.

### Steps to reproduce

in neovim, open multiple tabs so that the tab bar is shown, double click the tab bar itself and observe that no new tab has been opened

in vim9, open multiple tabs so that the tab bar is shown, double click the tab bar itself and observe that a new tab has been opened

the docs for this feature are here... https://neovim.io/doc/user/tabpage.html#tab-page-commands

A double click with the mouse in the non-GUI tab pages line opens a new, empty
tab page.  It is placed left of the position of the click.  The first click
may select another tab page first, causing an extra screen update.

### Expected behavior

a new tab opens when double clicking the tab bar (to the right of the existing open tabs)

### Nvim version (nvim -v)

NVIM v0.10.1 Build type: Release LuaJIT 2.1.1713484068 Run "nvim -V1 -v" for more info

vim opens the tab properly, neovim does not open a new tab

### Operating system/version

debian trixie

### Terminal name/version

gnome-terminal

xterm-256color

### Installation

built from source I think.

zeertzjq:
Also correct the comments about tabpage labels in custom tabline.

Fix #30599

### DIFF
--- a/src/nvim/statusline.c
+++ b/src/nvim/statusline.c
@@ -810,6 +810,16 @@ void draw_tabline(void)
       }
     }
 
+    for (int scol = col; scol < Columns; scol++) {
+      // Use 0 as tabpage number here, so that double-click opens a tabpage
+      // after the last one, and single-click goes to the next tabpage.
+      tab_page_click_defs[scol] = (StlClickDefinition) {
+        .type = kStlClickTabSwitch,
+        .tabnr = 0,
+        .func = NULL,
+      };
+    }
+
     char c = use_sep_chars ? '_' : ' ';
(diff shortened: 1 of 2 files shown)


justinmk (approved):
seems harmless

PR #30623, opened 2024-10-02
ci: bump ubuntu runner version to ubuntu-24.04

dundargoc:
Also bump clang to version 20.


dundargoc:
New clang-analyzer warnings that needs to be fixed. It seems we're using enums incorrectly.

wookayin:
I think this would result in bumping up the minimum glibc version (for the CI-released binaries). I'm not sure we'll want this.

dundargoc:
This only bumps the test runner, not release.

dundargoc:
Also glibc doesn't matter, the releases in neovim-releases repo works on ancient glibc versions anyway.

dundargoc:
https://github.com/neovim/neovim/pull/30652

clason:
praise zig 🙌

Issue #23121, opened 2023-04-16
vim.cmd.foo(2) does not behave like vim.cmd.foo{count=2}


smjonas:
### Problem

The ex-command `:undo {N}` accepts a count to jump to the change after the sequence number `{N}`. However, `vim.cmd.undo` only works with zero arguments and shows an error like
```
E5108: Error executing lua [string ":lua"]:1: Wrong number of arguments                                                                                               
stack traceback:
        [C]: in function 'undo'
        [string ":lua"]:1: in main chunk
```
  when called with an argument. 

A workaround is to use `vim.cmd("undo " .. sequence_number)`.

### Steps to reproduce

- `nvim --clean`
- `:lua vim.cmd.undo(1)`

### Expected behavior

`vim.cmd.undo` should accept a count argument

### Neovim version (nvim -v)

NVIM v0.10.0-dev-2650+g8c6f97bef

### Vim (not Nvim) behaves the same?

No

### Operating system/version

Linux Mint 20.2

### Terminal name/version

Kitty 0.15

### $TERM environment variable

xterm-kitty

### Installation

build from repo

zeertzjq:
`vim.cmd.undo({count = 1})` works

zeertzjq:
Note: https://github.com/neovim/neovim/pull/23243#issuecomment-1651045666

PR #30632, opened 2024-10-02
fix(defaults): properly pass count to quickfix commands

### LINKED ISSUE #30626: `[Q` failed on count with `Wrong number of arguments`
### Problem

With the newly unimpaired keybinding, `[Q` will show error if used with count.

```
E5108: Error executing lua: vim/_defaults.lua:0: Wrong number of arguments
stack traceback:
	[C]: in function 'crewind'
```

### Steps to reproduce

nvim --clean
:grep <something>
1[Q

### Expected behavior

Should behave the same as if `:crewind <count>` is run

### Nvim version (nvim -v)

v0.11.0-dev-876+g5331f87f61

N/A

### Operating system/version

Arch linux 6.11.1-arch1-1

### Terminal name/version

foot v1.18.1

foot

### Installation

AUR

zeertzjq:
Fix #30626

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -225,13 +225,13 @@ do
     })
 
     vim.keymap.set('n', '[Q', function()
-      vim.cmd.crewind({ args = vim.v.count ~= 0 and { vim.v.count } or nil })
+      vim.cmd.crewind({ count = vim.v.count ~= 0 and vim.v.count or nil })
     end, {
       desc = ':crewind',
     })
 
     vim.keymap.set('n', ']Q', function()
-      vim.cmd.clast({ args = vim.v.count ~= 0 and { vim.v.count } or nil })
+      vim.cmd.clast({ count = vim.v.count ~= 0 and vim.v.count or nil })
     end, {
       desc = ':clast',
     })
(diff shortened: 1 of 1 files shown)


justinmk (approved):

runtime/lua/vim/_defaults.lua
```diff
@@ -225,13 +225,13 @@ do
     })
 
     vim.keymap.set('n', '[Q', function()
-      vim.cmd.crewind({ args = vim.v.count ~= 0 and { vim.v.count } or nil })
+      vim.cmd.crewind({ count = vim.v.count ~= 0 and vim.v.count or nil })
```
gpanders, line 228: `:h :crewind` says:

```
:cr[ewind][!] [nr]	Display error [nr].  If [nr] is omitted, the FIRST
			error is displayed.  See |:cc|.
```

so why does `args` not work here? `:crewind 1` works as expected

gpanders: `count` has the same effect (this isn't obvious to me from the help description) so this is fine.

jamestansx: `vim.api.nvim_parse_cmd("crewind 123", {})`  do actually parse the argument to `count` 

```
{                                                                                    
  addr = "?",                                                                                  
  args = {},                                                                                   
  bang = false,                                                                                
  cmd = "crewind",                                                                             
  count = 123,                                                                                 
  magic = {                                                                                    
    bar = true,                                                                                
    file = false                                                                               
  },                                                                                           
  mods = {                                                                                     
    browse = false,                                                                            
    confirm = false,                                                                           
    emsg_silent = false,                                                                       
    filter = {                                                                                 
      force = false,                                                                           
      pattern = ""                                                                             
    },                                                                                         
    hide = false,                                                                              
    horizontal = false,                                                                        
    keepalt = false,                                                                           
    keepjumps = false,                                                                         
    keepmarks = false,                                                                         
    keeppatterns = false,                                                                      
    lockmarks = false,                                                                         
    noautocmd = false,                                                                         
    noswapfile = false,                                                                        
    sandbox = false,                                                                           
    silent = false,                                                                            
    split = "",                                                                                
    tab = -1,                                                                                  
    unsilent = false,                                                                          
    verbose = -1,                                                                              
    vertical = false                                                                           
  },                                                                                           
  nargs = "0",                                                                                 
  nextcmd = "",                                                                                
  range = { 123 }                                                                              
}   
```

zeertzjq: It was #23121

PR #30643, opened 2024-10-03
docs(treesitter): generate TSNode and TSTree docs

ribru17:
**Problem:** The documentation for `TSNode` and `TSTree` methods is incomplete from the LSP perspective. This is because they are written directly to the vimdoc, rather than in Lua and generated to vimdoc.

**Solution:** Migrate the docs to Lua and generate them into the vimdoc. This requires breaking up the `treesitter/_meta.lua` file into a directory with a few different modules.

This commit also makes the vimdoc generator slightly more robust with regard to sections that have multiple help tags (e.g. `*one* *two*`)

Before:
image

After:
image


runtime/lua/vim/treesitter/_meta/tstree.lua
```diff
@@ -0,0 +1,43 @@
...
+---@param end_byte_old integer
+---@param end_byte_new integer
+---@param start_row integer
+---@param start_col integer
+---@param end_row_old integer
+---@param end_col_old integer
+---@param end_row_new integer
+---@param end_col_new integer
+---@nodoc
+function TSTree:edit(start_byte, end_byte_old, end_byte_new, start_row, start_col, end_row_old, end_col_old, end_row_new, end_col_new) end
```
ribru17, line 29: Note: If the function stub is not all on one line, then Lua LS complains of unused parameters. Wat?? :laughing:

runtime/doc/treesitter.txt
```diff
@@ -57,151 +57,6 @@ loaded: >lua
     vim.treesitter.language.add('python', { path = "/path/to/python.wasm" })
 <
 
-==============================================================================
-TREESITTER TREES                                             *treesitter-tree*
```
ribru17, line 61: Note: The ordering of these sections was changed slightly (moved below) because otherwise all sections after them (until the next generated section) were deleted. Is there an easy way to fix this? If not hopefully the reordering is nbd

clason: We need to generate more; then this will work itself out. I am fine with the new order. (Generally, high-level/Neovim-wide sections should come first; low-level API later.)

runtime/lua/vim/treesitter/_meta/tstree.lua
justinmk: That's fine, this is "data" anyway :)

scripts/gen_vimdoc.lua
```diff
@@ -374,7 +385,7 @@ local config = {
       return 'Checkhealth'
     end,
     helptag_fmt = function()
-      return 'vim.health* *health' -- HACK
+      return { 'vim.health', 'health' }
```
justinmk, line 388: Nice, thanks!

justinmk (approved):

PR #28904, opened 2024-05-21
vim-patch:d1d9316: runtime(zig): refactor zig ftplugin, remove auto format

gpanders:
Fix #28878

Refactored zig ftplugin, removed upstream comment, aucmd and
auto formatting support. Updated documentation for zig configuration
settings and added new maintainer.

closes: vim/vim#13803

https://github.com/vim/vim/commit/d1d9316c6621d275593f58664975faf0fc23985f

Co-authored-by: Tiseno <mathias.lindgren@stabelo.se>

PR #30645, opened 2024-10-03
Backport zig refactor to 0.10 branch

tranzystorekk:
This is a backport of https://github.com/neovim/neovim/pull/28904 to 0.10.

The change feels substantial enough to the zig user experience to warrant releasing under the 0.10 version - arguably users who create a zig lsp setup end up confused when nvim automatically does reformatting on save and prints diagnostics to a separate buffer.


clason:
Sorry, but we don't backport vim patches.

tranzystorekk:
Ah, it was worth a shot, but thanks for the heads up!

clason:
Consider it motivation for us to release 0.11 sooner rather than later ;)

Issue #25416, opened 2023-09-29
Docs: Explicitly mark prerelease Lua API


clason:
### Problem

Lua functions added during a development cycle are not considered stable and may change at any time before the release, not just in behavior but in name (or may vanish again altogether). However, it is not obvious from the documentation if a function is still unstable or already part of the public API (in which case renaming or removing have to go through a deprecation process). This means that you can often find them used in plugins, with the obvious effects whenever they are changed. (Prerelease versions should only be used for testing and developing, not productive use, but that's a different topic.)

This is different from the Nvim API, which has `NVIM_API_PRERELEASE` annotations.

### Expected behavior

We should explicitly mark newly-added Lua functions as "unstable" in the documentation. As the documentation is automatically generated from docstrings, so should this be.

The simplest approach would be to add an `--- @unstable` annotation and convert this in `gen_vimdoc` to a friendly (but clear) warning; as part of the release process, these annotations would then have to be removed. 

Alternatively, we could integrate the Lua API into our general API versioning (without the strong contract about not removing functions before 2.0, of course) and use the same `NVIM_API` annotations for them (including `FUNC_API_SINCE`, `FUNC_API_DEPRECATED_SINCE`, etc.?) so we could use the same automation as for the C documentation. This would imply (not necessarily, but in practice), that the API version would be bumped at each minor release instead of the first time an `nvim_foo` API function is added.

This would probably be a bit more work to set up, but versioning the Lua API is something we'll need to do for Nvim 1.0 anyway, and there's no reason we can't start earlier.

justinmk:
Either way sounds good but we definitely want this to be programmatic, not manually maintained in the docs.

But `@since` seems better than `@unstable` because it can be used to indicate both "unstable/experimental" as well as "when was this function first released?"

```
@since unstable
@since 0.1.2
```

clason:
That would be the point of the second proposal: `@since 12` would mark this as "unstable" until API level 12 is released (at which point it would automatically not be marked anymore) -- this would be fully controlled by the `NVIM_API_*` in our top `CMakeLists.txt`. (To be clear: the second proposal -- which was @bfredl's -- is to use our Nvim API version numbering for this, _not_ a separate (Neovim version) counter.)

llllvvuu:
> `--- @since unstable`

Does `--- @since 999999` work for this?

clason:
No, we want to use the `NVIM_API_LEVEL` for this, since we already have tooling for that in our release script. 

[...]

PR #30649, opened 2024-10-03
docs: render `@since` versions, support `@since 0` to mark experimental

justinmk:
ref #25416

One implication of this current approach is that `NVIM_API_LEVEL` should be bumped when a new Lua function is added?

TODO(future): add a lint check which requires `@since` on all new functions.


runtime/doc/lua.txt
```diff
@@ -2461,18 +2469,24 @@ vim.validate({opt})                                           *vim.validate()*
 Lua module: vim.loader                                            *vim.loader*
 
 vim.loader.disable()                                    *vim.loader.disable()*
+    WARNING: This feature is experimental/unstable.
```
justinmk, line 2472: is it time to graduate vim.loader ?

PR #30652, opened 2024-10-04
fix: fix clang-analyzer warnings

dundargoc:


dundargoc:
I changed my mind, I'll just silence these for now. If anyone is motivated to fix them later then go for it.

PR #30653, opened 2024-10-04
fix: handle edge case in `screenpos()`

### LINKED ISSUE #30639: `screenpos()` returns all zeros under specific circumstances
### Problem

`screenpos()` _sometimes_ returns all zeros; ie:

```lua
{
  col = 0,
  curscol = 0,
  endcol = 0,
  row = 0
}
```

instead of the actual screen coordinates, causing plug-ins like [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) to draw their autocomplete pop-up menu in the top-left of the screen instead of at the cursor location:

autocomplete menu appearing in the wrong place

This was originally reported in:

- https://github.com/neovim/neovim/issues/19606

but without a consistent repro. I've been seeing this happen intermittently for years and have never been able to find a reliable repro recipe, until now.

### Steps to reproduce

This one is incredibly sensitive to initial conditions (screen size, font size, path lengths), and rather fiddly to reproduce. Nevertheless, I've found that I can trigger the bug from a clean Neovim install by interacting with the [vim-dirvish](https://github.com/justinmk/vim-dirvish) plug-in and following a very precise set of steps. (I don't think this is a bug with vim-dirvish or with nvim-cmp, but they are useful to illustrate the bug.)

1. Size your terminal window such that it is 137 columns wide. Verify this with:
   ```
   tput cols
   ```
2. Create a new scratch directory in which to reproduce the bug:
   ```
   mkdir /tmp/scratch
   ```
   Note that the name of the path matters here because the path must be of a certain length to trigger the bug.
3. Change into the scratch directory:
   ```
   cd /tmp/scratch
   ```
4. Create directories in which to put the repro files:
   ```
   mkdir -p aspects/nvim/files/.config/nvim/plugin/mappings
   ```
   There's nothing special about the _contents_ of this path (this is just a path from the project I was in when I discovered how to repro the bug) but the _length_ of the path does matter.
5. Create repro files (again, contents don't matter, so `touch` will suffice and, again, the names are just take from the project I was in and its the length that matters):
   ```
   touch aspects/nvim/files/.config/nvim/plugin/mappings/normal.lua
   touch aspects/nvim/files/.config/nvim/plugin/mappings/visual.lua
   ```
6. Grab a copy of vim-dirvish:
   ```
   mkdir -p pack/bundle/opt
   git clone https://github.com/justinmk/vim-dirvish.git pack/bundle/opt/vim-dirvish
   ```
7. Launch Neovim with no config:
   ```
   nvim -u NONE
   ```
8. Load vim-dirvish:
   ```
   set packpath=.
   packadd vim-dirvish
   packloadall
   ```
9. Create a mapping so that you can inspect the result of `screenpos()` (this one taken from #19606):
   ```
   :nmap <c-s> :lua print(vim.inspect(vim.fn.screenpos(0, vim.api.nvim_win_get_cursor(0)[1], vim.api.nvim_win_get_cursor(0)[2]+1)))<cr>
   ```
10. Open the first of the repro files:
    ```
    :e aspects/nvim/files/.config/nvim/plugin/mappings/normal.lua
    ```
11. Create a vertical split:
    ```
    :vs
    ```
12. Move focus to window on right side of the split:
    ```
    <C-w l>
    ```
13. Invoke vim-dirvish to show the contents of the containing directory:
    ```
    :Dirvish
    ```
14. Hit `j` to move down to the sibling file (`visual.lua`).
15. Hit `Enter` to open the file.
16. Hit the `<C-s>` binding you created previously to display the `screenpos()` result; note that it's all zeros:
    ```
    {
      col = 0,
      curscol = 0,
      endcol = 0,
      row = 0
    }
    ```

Additional notes:

- This repros both inside tmux (`TERM=tmux-256color`) and outside (`TERM=xterm-kitty`).
- It also appears to affect the Vim (tested using `-u DEFAULTS` on version 9.1 from Homebrew and 9.0, as provided by Apple with the OS)[^vim].
- As I mentioned above, this whole thing is very sensitive to path lengths and cursor position:
  - In the vim-dirvish listing, move the cursor to the end (`$`) before opening the second file, and the bug repros.
  - Move it to the beginning (`^`) before opening the second file, and it does not repro.
- In common with the original report (#19606), after you've seen the bug, running `:set wrap` (even if it was already set) temporarily fixes the problem and hitting the `<C-s>` mapping shows that `screenpos()` will return sensible results again.
- Font size matters, because it affects the number of characters that appear before the line is too wide to show on the screen. For example, if I make my font small enough so that everything fits, this no longer repros. For reference, this screenshot shows my font size when reproing the issue: font size reference

[^vim]: I posted this here first because Neovim is my daily driver, but I will also report upstream.

### Expected behavior

`screenpos()` should return real screen coordinates, like:

```
{
  col = 70,
  curscol = 70,
  endcol = 70,
  row = 1
}
```

### Nvim version (nvim -v)

- NVIM v0.10.1
- NVIM v0.10.2

yes, 9.1

### Operating system/version

macOS 14.7

### Terminal name/version

kitty 0.36.4

tmux-256color

### Installation

Homebrew

wincent:
As described in the related issue, `screenpos()` can sometimes get confused and return all zeros, causing anything that relies on it (eg. the nvim-cmp completion menu) to draw in the top-left corner of the screen instead of in the right place.

This commit applies the Neovim fix, equivalent to the suggested fix for Vim, seen here:

- https://github.com/vim/vim/issues/15792#issuecomment-2392387779

`validate_cursor_col()` already has an `if` to handle this edge case:

- https://github.com/neovim/neovim/blob/b45c50f3140e7ece593f2126840900f5cc3d39ea/src/nvim/move.c#L758-L759

so it just needs to be copied into `textpos2screenpos()` as well.

Closes: https://github.com/neovim/neovim/issues/30639

### DIFF
--- a/src/nvim/move.c
+++ b/src/nvim/move.c
@@ -1080,7 +1080,9 @@ void textpos2screenpos(win_T *wp, pos_T *pos, int *rowp, int *scolp, int *ccolp,
         row += rowoff;
       }
 
-      col -= wp->w_leftcol;
+      if (col > (int)wp->w_leftcol) {
+        col -= wp->w_leftcol;
+      }
 
       if (col >= 0 && col < wp->w_width_inner && row >= 0 && row < wp->w_height_inner) {
         coloff = col - scol + (local ? 0 : wp->w_wincol + wp->w_wincol_off) + 1;


justinmk:
If vim fixes this soon let's wait for their patch

zeertzjq:
This solution probably doesn't make sense. See https://github.com/vim/vim/pull/15803#pullrequestreview-2349347634

PR #30654, opened 2024-10-04
docs: generate params, returns in builtin.txt

justinmk:
Thanks to @lewis6991 's fantastic work on `gen_*.lua`, we can easily add structured info to builtin.txt, so let's do that.

Notes:
- ~~The parameter list intentionally does not surround each name with "{...}" because this just adds noise.~~
- Currently the types are Lua types. Can revisit this later, or maybe it's fine.

TODO(future):
- remove the "{...}" from param lists in api.txt, lua.txt
- use AI to generate the `returns_desc` for each "return" item, based on the `desc` field.
    - feeding eval.lua into chatgpt/cody hits the free tier limit, can someone do this? :)
    - also for params

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -25,6 +25,12 @@ abs({expr})                                                              *abs()*
 			echo abs(-4)
 <			4
 
+                Parameters: ~
+                  • {expr} (`any`)
+
+                Return: ~
+                  (`number`)
+
 acos({expr})                                                            *acos()*
 		Return the arc cosine of {expr} measured in radians, as a
 		|Float| in the range of [0, pi].
@@ -38,6 +44,12 @@ acos({expr})                                                            *acos()*
 			echo acos(-0.5)
 <			2.094395
 
+                Parameters: ~
+                  • {expr} (`any`)
+
+                Return: ~
+                  (`number`)
+
 add({object}, {expr})                                                    *add()*
 		Append the item {expr} to |List| or |Blob| {object}.  Returns
 		the resulting |List| or |Blob|.  Examples: >vim
@@ -49,6 +61,14 @@ add({object}, {expr})                                                    *add()*
 		Use |insert()| to add an item at another position.
 		Returns 1 if {object} is not a |List| or a |Blob|.
 
+                Parameters: ~
+                  • {object} (`any`)
+                  • {expr} (`any`)
+
+                Return: ~
+                  (`any`) Resulting |List| or |Blob|, or 1 if {object} is not
+                  a |List| or a |Blob|.
+
 and({expr}, {expr})                                                      *and()*
 		Bitwise AND on the two arguments.  The arguments are converted
 		to a number.  A List, Dict or Float argument causes an error.
@@ -57,12 +77,22 @@ and({expr}, {expr})                                                      *and()*
 			let flag = and(bits, 0x80)
 <
 
+                Parameters: ~
+                  • {expr} (`any`)
+                  • {expr1} (`any`)
+
+                Return: ~
+                  (`integer`)
+
 api_info()                                                          *api_info()*
 		Returns Dictionary of |api-metadata|.
 
 		View it in a nice human-readable format: >vim
 		       lua vim.print(vim.fn.api_info())
 <
+
+                Return: ~
+                  (`table`)
 
 append({lnum}, {text})                                                *append()*
 		When {text} is a |List|: Append each item of the |List| as a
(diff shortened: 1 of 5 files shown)


runtime/doc/builtin.txt
```diff
@@ -79,6 +104,12 @@ append({lnum}, {text})                                                *append()*
 			let failed = append(0, ["Chapter 1", "the beginning"])
 <
 
+                Parameters:
+                  • lnum (`integer`)
+                  • text (`any`)
```
lewis6991, line 109: A lot of these types still need refining. Maybe it's worth taking a pass at that during or before this?

lewis6991:
> The parameter list intentionally does not surround each name with "{...}" because this just adds noise.

We've done this pretty much everywhere else so not sure if deviating is justified. Otherwise I would still surround the params with backticks so they are at least highlighted differently.

runtime/doc/builtin.txt
```diff
@@ -181,13 +248,22 @@ assert_equal({expected}, {actual} [, {msg}])                    *assert_equal()*
 <		Will add the following to |v:errors|:
 			test.vim line 12: baz: Expected 'foo' but got 'bar' ~
 
+                Parameters:
+                  • expected (`any`)
+                  • actual (`any`)
+                  • msg (`any?`)
+
+                Return: (`0|1`)
```
lewis6991, line 256: Again for consistency, I would format this as:

```suggestion
                Return: ~
                    (`0|1`)
```

It's a bit wasteful for on vertical space for most of this file, but it's easier to read for returns with a description and has better highlighting.

image

runtime/lua/vim/_meta/vimfn.lua
```diff
@@ -47,7 +47,7 @@ function vim.fn.acos(expr) end
 ---
 --- @param object any
 --- @param expr any
---- @return any
+--- @return any # Resulting |List| or |Blob|, or 1 if {object} is not a |List| or a |Blob|.
```
justinmk, line 50: `returns_desc` field defined in eval.lua. Now just need AI to generate all those `returns_desc` :)

runtime/doc/builtin.txt
```diff
@@ -248,6 +339,12 @@ assert_false({actual} [, {msg}])                                *assert_false()*
 		A value is false when it is zero. When {actual} is not a
 		number the assert fails.
 
+                Parameters:
+                  • actual (`any`)
+                  • msg (`any?`)
```
lewis6991, line 344: In `lua.txt` I was able to align the types as

```suggestion
                Parameters: ~
                  • actual (`any`)
                  • msg    (`any?`)
```

Can we do that here?

Also add a ` ~` at the ends of `Parameters:`

justinmk:
> > The parameter list intentionally does not surround each name with "{...}" because this just adds noise.
> 
> We've done this pretty much everywhere else so not sure if deviating is justified

I was planning to remove the {} from lua.txt/api.txt etc. Any objection to that?

lewis6991:
> I was planning to remove the {} from lua.txt/api.txt etc. Any objection to that?

I kind of like it but not against removing it since it's a weird vimdoc convention.

If you remove it in the listing then I'd strongly suggest also removing it from the signature.

And again maybe wrap in backticks so they get some form of highlighting.

justinmk:
> If you remove it in the listing then I'd strongly suggest also removing it from the signature.
> 
> And again maybe wrap in backticks so they get some form of highlighting.

tree-sitter-vimdoc will soon gain highlighting for the "{...}". So I'll just add them here for now, and revisit the general topic later.

runtime/doc/builtin.txt
justinmk: Does that need to block this? Those types are in the meta files already...

lewis6991: No. Just a light suggestion.

runtime/doc/builtin.txt
justinmk: > In `lua.txt` I was able to align the types as

I tried this:

```lua
      local s = fmt('- x%-14sx (`%s%s`)', fmt('{%s}', pname), ptype, optional)
      write(util.md_to_vimdoc(s, 16, 18, TEXT_WIDTH))
```

but it looks like `md_to_vimdoc` is scrubbing the extra spaces, so the "alignment" is removed.

justinmk:
Updated all param types.

lewis6991 (approved):

src/nvim/eval.lua
```diff
@@ -333,7 +334,7 @@ M.funcs = {
 
     ]=],
     name = 'assert_equalfile',
-    params = {},
+    params = { { 'fname-one', 'string' }, { 'fname-two', 'string' } },
     returns = '0|1',
     signature = 'assert_equalfile({fname-one}, {fname-two})',
```
justinmk, line 339: Future idea: generate the signature docs from the `args` + `params`, to eliminate this `signature` field?

lewis6991: Seems like a relatively simple change we can be iterated by gradually removing `signature` fields for functions with `params`.

PR #30658, opened 2024-10-04
Update close_unresponsive.js

akshitbansal2005:
Identified Issues:
Redundant Object Property Access:

Accessing context.repo.owner and context.repo.repo multiple times. This can be streamlined by destructuring or assigning once for cleaner code. Error Handling:

There is no error handling in case API requests fail. It’s important to ensure that the GitHub API interactions have proper try-catch blocks or equivalent mechanisms. Inefficient Date Difference Calculation:

const diff = now - created_at; const diffDays = diff / (1000 * 60 * 60 * 24); can be replaced with a more readable and efficient method for calculating the difference in days. Hardcoding Strings:

The label "needs:response" and the close_message could be externalized for easier configuration in the future, such as adding them to environment variables or parameters. Missing Async/Await in Loop:

Although the code is using async/await, there are cases where the code could be rewritten for better flow and readability. Specifically, the for-loop can await inside the loop which can cause performance issues when iterating through many issues sequentially. Potential Problem with Empty Events:

If there are no events matching the filter, the array events might be empty, and accessing events[events.length - 1] would lead to an undefined value, causing runtime errors when trying to access created_at. Lack of Logging:

There's no feedback in case of failure or success in processing each issue, which makes it hard to debug. Logging important actions would improve maintainability.

### DIFF
--- a/.github/scripts/close_unresponsive.js
+++ b/.github/scripts/close_unresponsive.js
@@ -2,55 +2,73 @@ function labeledEvent(data) {
   return data.event === "labeled" && data.label.name === "needs:response";
 }
 
+// Externalized constants for flexibility
 const numberOfDaysLimit = 30;
-const close_message = `This has been closed since a request for information has \
-not been answered for ${numberOfDaysLimit} days. It can be reopened when the \
-requested information is provided.`;
+const labelName = "needs:response";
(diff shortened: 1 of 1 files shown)


justinmk:
based on your activity, you are a bot so you are banned from Neovim org.

PR #30659, opened 2024-10-04
fix(lsp): fix cursor position after snippet expansion

### LINKED ISSUE #30656: LSP completion: `CompleteDone`: unwanted cursor jumping
### Problem

I noticed that after expanding `vim._with()` context keys the cursor jumps to the end of line instead to the end of completed word (e.g. `bo`). I found that that case is treated as snippet expansion:

https://github.com/neovim/neovim/blob/305012ea073267492e1d626a304c6993b0dfa145/runtime/lua/vim/lsp/completion.lua#L522-L523

Is it correct in that case?

The text after cursor is cut and saved to `suffix` that is then appended to the text expanded as snippet. 

https://github.com/neovim/neovim/blob/305012ea073267492e1d626a304c6993b0dfa145/runtime/lua/vim/lsp/completion.lua#L535-L545

Assuming that I commit complete `bo` in `vim._with({bo}, f)`, I get `bo}, f)` passed to `vim.snippet.expand`.

https://github.com/neovim/neovim/blob/305012ea073267492e1d626a304c6993b0dfa145/runtime/lua/vim/lsp/completion.lua#L117-L123

Maybe in 
https://github.com/neovim/neovim/blob/305012ea073267492e1d626a304c6993b0dfa145/runtime/lua/vim/lsp/completion.lua#L543

replacing `#line` with `cursor_col` would fix it? 

### Steps to reproduce

init.lua:
```lua
  pattern = {'lua'},
  group = vim.api.nvim_create_augroup('luals', {}),
  callback = function ()
    vim.lsp.start({
      name = 'luals',
      filetype = 'lua',
      cmd = {'lua-language-server'},
      root_dir = vim.uv.cwd(),
      settings = {
        Lua = {
          runtime = {
            path = {
              '?.lua',
              '?/init.lua',
            },
            version = 'LuaJIT',
            pathStrict = true,
          },
          workspace = {
            ignoreDir = {'lua'},
            checkThirdParty = false,
            library = {
              vim.fs.joinpath(vim.env.VIMRUNTIME, 'lua'),
              vim.fs.joinpath('${3rd}/luv/library', 'lua'),
            },
          },
          diagnostics = {
            globals = {'vim'},
          },
        },
      },
    })
  end,
})

group = vim.api.nvim_create_augroup('arctgx.completion', {clear = true}),
  callback = function (args)
    local clientId = args.data.client_id
    local client = assert(vim.lsp.get_client_by_id(clientId))
    vim.notify(('Server %s attached'):format(client.name), vim.log.levels.INFO)
    vim.lsp.completion.enable(true, clientId, args.buf, {autotrigger = true})

    vim.keymap.set({'i'}, '<C-Space>', function ()
      vim.lsp.completion.trigger()
    end, {buffer = args.buf})
  end
})

vim.go.completeopt = 'noinsert,menuone'
```
- `nvim --clean -u init.lua new.lua`
- after server is attached insert `vim._with({bo}, f)` and try to confirm `bo` completion with `<C-y>` and see the cursor on the end of line. 

### Expected behavior

Cursor stays at the end of `bo`. 

### Nvim version (nvim -v)

v0.11.0-dev-897+g305012ea0

no

### Operating system/version

Debian Sid

### Terminal name/version

alacritty 0.14.0-dev (6585d604), tmux 3.5

tmux-256color

### Installation

from repo

przepompownia:
Problem: on `CompleteDone` cursor can jump to the end of line instead of the end of completed word.

Solution: trim the part of the line that was cut off and pasted again.

Fixes #30656

Co-authored-by: @mfussenegger


przepompownia:
Still a draft while here is no test for #30656.

przepompownia:
Probably I need help with creating test. I'm not sure if I can check the cursor position in `test/functional/plugin/lsp/completion_spec.lua` I tried 
```lua
  it('should put cursor at the end of completed word', function()
    local completion_list = {
      isIncomplete = false,
      items = {
        {
          insertText = "bo",
          insertTextFormat = 2,
          kind = 10,
          label = "bo?",
          sortText = "0001",
        },
      },
    }
    local expected = {
      abbr = "bo?",
      dup = 1,
      empty = 1,
      hl_group = "",
      icase = 1,
      info = "",
      kind = "Property",
      menu = "",
      word = "bo",
    }
    local line = 'vim._with({bo|}, f)'
    local cursor_col_before = line:find('|') - 1
    local result = complete(line, completion_list)

    local cursor_col_after = exec_lua(function ()
      return vim.fn.col(".")
      -- local row, col = vim.api.nvim_win_get_cursor(0)
      -- return col
    end)
    eq(cursor_col_before, cursor_col_after)
  end)
```
in `vim.lsp.completion: protocol` set.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -540,7 +540,7 @@ local function on_complete_done()
     -- Remove the already inserted word.
     local start_char = cursor_col - #completed_item.word
     local line = api.nvim_buf_get_lines(bufnr, cursor_row, cursor_row + 1, true)[1]
-    api.nvim_buf_set_text(bufnr, cursor_row, start_char, cursor_row, #line, { '' })
+    api.nvim_buf_set_text(bufnr, cursor_row, start_char, cursor_col, #line, { '' })
```
justinmk, line 543: If this "works", I would guess it's only by accident. The parameter is a row, not a column.

The `#line` arg looks suspicious, is that intentional @MariaSolOs ?

justinmk:
> Probably I need help with creating test. I'm not sure if I can check the cursor position in `test/functional/plugin/lsp/completion_spec.lua` I

A new test probably isn't needed, just assert the cursor position in an existing test (may need to add some extra text to the initial buffer, to cover this scenario).

What problem did you have with checking `vim.fn.col(".")` ?

runtime/lua/vim/lsp/completion.lua
MariaSolOs: I agree that changing the `row` parameter to `col` might just be making the tests pass but not be a correct change.

Regarding `#line`, what do you find suspicious about it @justinmk?

przepompownia: Ah, I see. Being focused on creating the correct test I left the wrong untested change. The original change I tested manually was placed inside installed lua files and lost with next `make install`. Without checking I see that `#line` (not `cursor_row` should be probably change to `cursor_col`).

przepompownia:
> What problem did you have with checking `nvim_win_get_cursor` or `vim.fn.col(".")` ?


With `return vim.fn.col(".")` (as above) I get
```
FAILED   ./test/functional/plugin/lsp/completion_spec.lua @ 542: vim.lsp.completion: protocol should put cursor at the end of completed word
./test/functional/plugin/lsp/completion_spec.lua:575: Expected objects to be the same.
Passed in:
(number) 1
Expected:
(number) 13
```

runtime/lua/vim/lsp/completion.lua
przepompownia: Even after changing the argument there is an incorrect "suffix" that is pasted after the completed word.

przepompownia: I see that returning empty string can avoid that but not aware possibly side effect (and all tests from `./test/functional/plugin/lsp/completion_spec.lua` passed for me).

MariaSolOs:
> > What problem did you have with checking `nvim_win_get_cursor` or `vim.fn.col(".")` ?
> 
> With `return vim.fn.col(".")` (as above) I get
> 
> ```
> FAILED   ./test/functional/plugin/lsp/completion_spec.lua @ 542: vim.lsp.completion: protocol should put cursor at the end of completed word
> ./test/functional/plugin/lsp/completion_spec.lua:575: Expected objects to be the same.
> Passed in:
> (number) 1
> Expected:
> (number) 13
> ```

In the snippet tests I use [this utility function](https://github.com/neovim/neovim/blob/988482d9422b40e86225935326127063d881b090/test/functional/lua/snippet_spec.lua#L115) to verify the cursor position. Maybe try something similar?

runtime/lua/vim/lsp/completion.lua
mfussenegger: > Even after changing the argument there is an incorrect "suffix" that is pasted after the completed word.

You'd also have to adapt the `suffix` logic.
`apply_snippet` no longer needs to have a `suffix` parameter and `clear_word` doesn't need to have a return value

```diff
diff --git a/runtime/lua/vim/lsp/completion.lua b/runtime/lua/vim/lsp/completion.lua
index 71ea2df10..fa9cf3e88 100644
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -111,16 +111,15 @@ local function parse_snippet(input)
   end)
   return ok and tostring(parsed) or input
 end
 
 --- @param item lsp.CompletionItem
---- @param suffix? string
-local function apply_snippet(item, suffix)
+local function apply_snippet(item)
   if item.textEdit then
-    vim.snippet.expand(item.textEdit.newText .. suffix)
+    vim.snippet.expand(item.textEdit.newText)
   elseif item.insertText then
-    vim.snippet.expand(item.insertText .. suffix)
+    vim.snippet.expand(item.insertText)
   end
 end
 
 --- Returns text that should be inserted when a selecting completion item. The
 --- precedence is as follows: textEdit.newText > insertText > label
@@ -537,19 +536,16 @@ local function on_complete_done()
       return nil
     end
 
     -- Remove the already inserted word.
     local start_char = cursor_col - #completed_item.word
-    local line = api.nvim_buf_get_lines(bufnr, cursor_row, cursor_row + 1, true)[1]
-    api.nvim_buf_set_text(bufnr, cursor_row, start_char, cursor_row, #line, { '' })
-    return line:sub(cursor_col + 1)
+    api.nvim_buf_set_text(bufnr, cursor_row, start_char, cursor_row, cursor_col, { '' })
   end
 
-  --- @param suffix? string
-  local function apply_snippet_and_command(suffix)
+  local function apply_snippet_and_command()
     if expand_snippet then
-      apply_snippet(completion_item, suffix)
+      apply_snippet(completion_item)
     end
 
     local command = completion_item.command
     if command then
       client:_exec_cmd(command, { bufnr = bufnr }, nil, function()
@@ -563,37 +559,37 @@ local function on_complete_done()
       end)
     end
   end
 
   if completion_item.additionalTextEdits and next(completion_item.additionalTextEdits) then
-    local suffix = clear_word()
+    clear_word()
     lsp.util.apply_text_edits(completion_item.additionalTextEdits, bufnr, offset_encoding)
-    apply_snippet_and_command(suffix)
+    apply_snippet_and_command()
   elseif resolve_provider and type(completion_item) == 'table' then
     local changedtick = vim.b[bufnr].changedtick
 
     --- @param result lsp.CompletionItem
     client.request(ms.completionItem_resolve, completion_item, function(err, result)
       if changedtick ~= vim.b[bufnr].changedtick then
         return
       end
 
-      local suffix = clear_word()
+      clear_word()
       if err then
         vim.notify_once(err.message, vim.log.levels.WARN)
       elseif result and result.additionalTextEdits then
         lsp.util.apply_text_edits(result.additionalTextEdits, bufnr, offset_encoding)
         if result.command then
           completion_item.command = result.command
         end
       end
 
-      apply_snippet_and_command(suffix)
+      apply_snippet_and_command()
     end, bufnr)
   else
-    local suffix = clear_word()
-    apply_snippet_and_command(suffix)
+    clear_word()
+    apply_snippet_and_command()
   end
 end
 
 --- @class vim.lsp.completion.BufferOpts
 --- @field autotrigger? boolean Whether to trigger completion automatically. Default: false
```

przepompownia: > Regarding `#line`, what do you find suspicious about it @justinmk?

Like in #30656 - the part from the end of the word to the end of line is appended to the string that is then expanded as a snippet.

przepompownia:
> > > What problem did you have with checking `nvim_win_get_cursor` or `vim.fn.col(".")` ?
> > 
> > 
> > With `return vim.fn.col(".")` (as above) I get
> > ```
> > FAILED   ./test/functional/plugin/lsp/completion_spec.lua @ 542: vim.lsp.completion: protocol should put cursor at the end of completed word
> > ./test/functional/plugin/lsp/completion_spec.lua:575: Expected objects to be the same.
> > Passed in:
> > (number) 1
> > Expected:
> > (number) 13
> > ```
> 
> In the snippet tests I use [this utility function](https://github.com/neovim/neovim/blob/988482d9422b40e86225935326127063d881b090/test/functional/lua/snippet_spec.lua#L115) to verify the cursor position. Maybe try something similar?

You can see I tried it but forgot to `unpack` the result. 
```lua
    local cursor_col_after = exec_lua(function ()
      -- return vim.fn.col(".")
      local row, col = unpack(vim.api.nvim_win_get_cursor(0))
      return col
    end)
```
The effect is the same (0 as the first column).

przepompownia:
```lua
    local line = 'vim._with({bo|}, f)'
    local cursor_col_before = line:find('|') - 1
    local result = complete(line, completion_list)
    local cursor_col_after = exec_lua(function ()
      -- return vim.fn.col(".")
      local row, col = unpack(vim.api.nvim_win_get_cursor(0))
      return col
    end)
    eq(cursor_col_before, cursor_col_after)

```
Is it possible that `complete` executes something asynchronously and `cursor_col_after` can be checked too early?

przepompownia:
We need to test the callback attached to `CompleteDone` that is not executed in `completion_spec.lua` and is a local function. I need a different approach than what I originally proposed.

przepompownia:
```lua
  it('should put cursor at the end of completed word', function()
    local line = 'vim._with({n|}, f)'
    local cursor_col = line:find('|') - 1
    line = line:gsub('|', '')
    local line_to_cursor = line:sub(1, cursor_col)
    local client_start_boundary = exec_lua(function ()
      return vim.fn.match(line_to_cursor, '\\k*$')
    end)
    local completed_word = 'noautocmd'

    local expected_cursor_column = client_start_boundary + #completed_word

    local cursor_column = exec_lua(function()
      local buf = vim.api.nvim_get_current_buf()
      vim.api.nvim_buf_set_lines(buf, 0, 1, true, {line})
      vim.api.nvim_win_set_cursor(0, {1, cursor_col})
      vim.cmd.startinsert()
      vim.api.nvim_buf_set_text(0, 0, client_start_boundary, 0, cursor_col, {completed_word})
      vim.api.nvim_set_vvar('completed_item', {
        abbr = 'noautocmd?',
        info = '',
        kind = 'Property',
        menu = '',
        user_data = {
          nvim = {
            lsp = {
              client_id = 1,
              completion_item = {
                data = {
                  id = 1,
                  uri = 'file:///tmp/test.lua'
                },
                insertText = 'noautocmd',
                insertTextFormat = 2,
                kind = 10,
                label = 'noautocmd?',
                sortText = '0001'
              }
            }
          }
        },
        word = 'noautocmd',
      })

      vim.api.nvim_exec_autocmds('CompleteDone', { buffer = buf, modeline = false })

      local _, col = unpack(vim.api.nvim_win_get_cursor(0))

      return col
    end)

    eq(expected_cursor_column, cursor_column)
  end)
```
It seems to be closer to what we want but still needs to run the autocommand (probably run `vim.lsp.completion.enable`) and (how to do?) simulate `vim.v.event.reason == 'accept'`.

przepompownia:
Now the problem is that **in the test** the result word is appended to the existing part instead of replacing it. For example completion `vim._with({no|}, f)` results with `vim._with({nonoautocmd}, f)`.

justinmk:
> (how to do?) simulate `vim.v.event.reason == 'accept'`.

looks like you figured that out, right? (`feed('<c-y>')`).


> Now the problem is that **in the test** the result word is appended to the existing part instead of replacing it. For example completion `vim._with({no|}, f)` results with `vim._with({nonoautocmd}, f)`.

but it works interactively? are you certain `trigger_at_pos` is triggering at the right position?

przepompownia:
> but it works interactively? are you certain `trigger_at_pos` is triggering at the right position?

I have no problem with interactive use unless someone points out some untested use case.

> looks like you figured that out, right? (`feed('<c-y>')`).
Yes, `on_complete_done` is executed also in the test case, but I found that `clear_word` isn't. I suspect we need to add a resolve provider to the server.

przepompownia:
Adding 
```diff
diff --git a/test/functional/plugin/lsp/completion_spec.lua b/test/functional/plugin/lsp/completion_spec.lua
index 73961506f..3b78f8ca9 100644
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -495,6 +495,7 @@ describe('vim.lsp.completion: protocol', function()
         capabilities = {
           completionProvider = {
             triggerCharacters = { '.' },
+            resolveProvider = true,
           },
         },
```
is probably needed like above (but then it seems to affect `vim.lsp.completion: protocol executes commands` case) on in another way, but it's not enough: we need some response from `completionItem/resolve` like 
````lua
{
  data = {
    id = 1,
    uri = "file:///.../examples/with.lua"
  },
  detail = "boolean?",
  documentation = {
    kind = "markdown",
    value = "```lua\n(field) vim.context.mods.noautocmd: boolean?\n```"
  },
  insertText = "noautocmd",
  insertTextFormat = 2,
  kind = 10,
  label = "noautocmd?",
  sortText = "0001"
}
````

**Comment by przepompownia** [comment]
Now I see that `vim.snippet.expand('noautocmd')` is executed in the test case, but the cursor still isn't placed as expected.

**Comment by justinmk** [comment]
if there is potential for a race while waiting for the server request/response, may be worth trying `t.retry()` https://github.com/neovim/neovim/blob/056009f74146b349c66790fbcba3130e85a6c6da/test/testutil.lua#L96

**Comment by przepompownia** [comment]
The problem has place before `CompleteDone`. To reduce complexity I put `vim.api.nvim_clear_autocmds({group = ('vim/lsp/completion-%d'):format(vim.api.nvim_get_current_buf())})` before `vim.v.completed_item = ...` and see that the buffer is not updated. Probably it was not necessary to any previous test. For example, the case `executes commands` behaves the same - we can see it adding 
```lua
    local col_line = exec_lua(function()
      return { vim.api.nvim_win_get_cursor(0)[2], vim.api.nvim_buf_get_lines(0, 0, 1, true)[1]}
    end)
    eq({5, 'hello'}, col_line)
```
at the end (the result is `{1, 'h'}`). 

I suspect that overriding `vim.fn.complete()` is not enough for this test, but still not understand why. After commenting it (with dependants) I get the correct result, both with and without `CompleteDone` autocommand.

justinmk:
> Probably it was not necessary to any previous test

sounds good. is this unblocked then?

przepompownia:
Before the change I tried to catch the new cursor position using `retry` but with no success. 

Finally I created a separate test set with as much shared local functions as possible with original `vim.fn.complete`.

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -726,3 +731,68 @@ describe('vim.lsp.completion: protocol', function()
...
+
+describe('vim.lsp.completion: integration', function()
+  before_each(function()
+    clear()
+    exec_lua(create_server_definition)
+  end)
+
+  after_each(clear)
+
+  it('should put cursor at the end of completed word', function()
```
justinmk, line 743:
```suggestion
  it('puts cursor at the end of completed word', function()
```

przepompownia:
After switching to use original `vim.fn.complete` setting `vim.v.completed_item` stopped working. Now this assignment works under `CompleteDone` autocommand created before `vim.lsp.completion.enable` creates its own analogue.

mfussenegger:
> is probably needed like above (but then it seems to affect vim.lsp.completion: protocol executes commands case) on in another way, but it's not enough: we need some response from completionItem/resolve like

`resolve` shouldn't be needed to expand snippets, that's entirely optional.
It looks like there was just a `clear_word()` call missing in one of the branches.


That `feed('<c-x><c-o>')` didn't trigger completion correctly seems to be related to a missing `vim.schedule` call - `vim.fn.complete()` wasn't called in the mainloop but in a luv callback.

I'm not entirely sure why that's only in the testcase a problem. Maybe due to the test-case using the in-process server.

With these two things fixed, it's possible to simplify the test-case:


```diff
diff --git a/runtime/lua/vim/lsp/completion.lua b/runtime/lua/vim/lsp/completion.lua
index e399a6a829..2e88672e58 100644
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -459,11 +459,13 @@ local function trigger(bufnr, clients)
         )
         vim.list_extend(matches, client_matches)
       end
     end
     local start_col = (server_start_boundary or word_boundary) + 1
-    vim.fn.complete(start_col, matches)
+    vim.schedule(function()
+      vim.fn.complete(start_col, matches)
+    end)
   end)
 
   table.insert(Context.pending_requests, cancel_request)
 end
 
@@ -584,10 +586,11 @@ local function on_complete_done()
       end
 
       apply_snippet_and_command()
     end, bufnr)
   else
+    clear_word()
     apply_snippet_and_command()
   end
 end
 
 --- @class vim.lsp.completion.BufferOpts
diff --git a/test/functional/plugin/lsp/completion_spec.lua b/test/functional/plugin/lsp/completion_spec.lua
index 579c4a7acf..dc1f8a19f3 100644
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -739,70 +739,36 @@ describe('vim.lsp.completion: integration', function()
   end)
 
   after_each(clear)
 
   it('puts cursor at the end of completed word', function()
-    local line = ('vim._with({%s|}, f)'):format('noau')
-    local completed_word = 'noautocmd'
-
-    local completion_item = {
-      insertText = completed_word,
-      insertTextFormat = 2, -- vim.lsp.protocol.InsertTextFormat.Snippet
-      label = completed_word,
-    }
-
     local completion_list = {
       isIncomplete = false,
-      items = { completion_item },
+      items = {
+        {
+          label = 'hello',
+          insertText = "${1:hello} friends",
+          insertTextFormat = 2,
+        },
+      },
     }
-
-    local initial_cursor_col = line:find('|') - 1
-    line = line:gsub('|', '')
-    local line_to_cursor = line:sub(1, initial_cursor_col)
-    local word_start = exec_lua(function()
-      return vim.fn.match(line_to_cursor, '\\k*$')
-    end)
-
     exec_lua(function()
-      vim.api.nvim_create_autocmd('CompleteDone', {
-        callback = function()
-          vim.v.completed_item = {
-            user_data = {
-              nvim = {
-                lsp = {
-                  client_id = vim.tbl_keys(vim.lsp.get_clients({ name = 'dummy' }))[1],
-                  completion_item = completion_item,
-                },
-              },
-            },
-            word = completed_word,
-          }
-        end,
-      })
+      vim.o.completeopt = 'menuone,noselect'
     end)
-
-    local expected_cursor_column = word_start + #completed_word
-
-    create_server(completion_list, completion_item)
-    eq(
-      1,
-      exec_lua(function()
-        return #vim.lsp.get_clients()
-      end)
-    )
-
-    feed('i' .. line)
-
-    exec_lua(function()
-      vim.api.nvim_win_set_cursor(0, { 1, initial_cursor_col })
-      vim.lsp.completion.trigger()
+    create_server(completion_list)
+    feed('i world<esc>0ih<c-x><c-o>')
+    retry(nil, nil, function()
+      eq(1, exec_lua(function() return vim.fn.pumvisible() end))
     end)
-
-    feed('<C-x><C-o><C-y>')
-
-    local cursor_column = exec_lua(function()
+    feed('<C-n><C-y>')
+    eq({true, {"hello friends world"}}, exec_lua(function()
+      return {
+        vim.snippet.active({ direction = 1}),
+        vim.api.nvim_buf_get_lines(0, 0, -1, true)
+      }
+    end))
+    feed('<tab>')
+    eq(#"hello friends", exec_lua(function()
       return vim.api.nvim_win_get_cursor(0)[2]
-    end)
-
-    eq(expected_cursor_column, cursor_column)
+    end))
   end)
 end)
```

przepompownia:
Indeed, `clear_word` was present on master when no resolve provider. Now it is again there and does not affect the current test case.

@mfussenegger can you reproduce the bug with cleared `CompleteDone` autocommand (e.g. `vim.api.nvim_clear_autocmds({group = string.format('vim/lsp/completion-%d', args.buf)})` in LspAttach)? I cannot (also with scheduled `complete()`). 

It would be nice to have simpler test case and it seems for me that `clear_word()` in _no resolve_ case should cause the same bug but I still cannot see it on master. Let's wait for finding the explanation before we simplify the test case.

przepompownia:
I confused disabling the whole `CompleteDone` callback with only resolve provider. It probably will allow me see what happens.

mfussenegger:
I'm not sure what you mean. My posted test-case will fail without the the `suffix`/`#line` change and passes with the fixes from this PR

przepompownia:
After adding `client.server_capabilities.completionProvider.resolveProvider = nil` (in `LspAttah`) to the real repro I see the bug so we can simplify the test case probably as you suggest :)

przepompownia:
Thank you all for the hints. Writing this test turned out to be not as simple as the fix itself.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -462,7 +461,9 @@ local function trigger(bufnr, clients)
       end
     end
     local start_col = (server_start_boundary or word_boundary) + 1
-    vim.fn.complete(start_col, matches)
+    vim.schedule(function()
+      vim.fn.complete(start_col, matches)
```
justinmk, line 465: is there a 1-line comment that can give a hint about why the schedule() is needed?

justinmk (approved):

przepompownia:
The current test fails without `schedule`d `complete`. I have no idea of an example that shows the need of `schedule` in real environment.

przepompownia:
While not sure we can redefine
```lua
    exec_lua(function()
      vim.fn.complete = vim.schedule_wrap(vim.fn.complete)
    end)
```
only for the test case.

przepompownia:
Even if we need such `schedule` in real env, it's probably a separate issue. @mfussenegger what do you think?

runtime/lua/vim/lsp/completion.lua
mfussenegger: It raises a `Vim:E565: Not allowed to change text or change window`  without it.

I suspect this doesn't happen outside of test-cases because the rpc layer already adds a schedule to escape the luv callback.

We could probably also do something like this:

```diff
diff --git a/test/functional/plugin/lsp/completion_spec.lua b/test/functional/plugin/lsp/completion_spec.lua
index 7e2dd0fbe3..a25bc738c9 100644
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -729,13 +729,10 @@ end)
 
 describe('vim.lsp.completion: integration', function()
   before_each(function()
     clear()
     exec_lua(create_server_definition)
-    exec_lua(function()
-      vim.fn.complete = vim.schedule_wrap(vim.fn.complete)
-    end)
   end)
 
   after_each(clear)
 
   it('puts cursor at the end of completed word', function()
diff --git a/test/functional/plugin/lsp/testutil.lua b/test/functional/plugin/lsp/testutil.lua
index a36cbac568..ca00805a16 100644
--- a/test/functional/plugin/lsp/testutil.lua
+++ b/test/functional/plugin/lsp/testutil.lua
@@ -62,10 +62,11 @@ M.create_server_definition = function()
       function srv.request(method, params, callback)
         table.insert(server.messages, {
           method = method,
           params = params,
         })
+        callback = vim.schedule_wrap(callback)
         local handler = handlers[method]
         if handler then
           handler(method, params, callback)
         elseif method == 'initialize' then
           callback(nil, {
```

mfussenegger: I'd tend to, alone for the reason to bring the in-process test server closer to the behavior of a cmd-stdio/tcp server.

And for future people + test-cases - debugging that this was necessary took quite some time as it's non-obvious and the errror wasn't propagated.

mfussenegger: Looks like making it general would break a couple of other tests due to the changed timing.
Given the impact I'd rather change this separately so I'm going ahead and merge this as is.

PR #30660, opened 2024-10-04
docs: more `@since` annotations

justinmk:


CMakeLists.txt
```diff
@@ -146,7 +146,7 @@ set(NVIM_VERSION_PATCH 0)
 set(NVIM_VERSION_PRERELEASE "-dev") # for package maintainers
 
 # API level
-set(NVIM_API_LEVEL 12)        # Bump this after any API change.
+set(NVIM_API_LEVEL 13)        # Bump this after any API/stdlib change.
```
justinmk, line 149: "api level" applies to all API surface area, not only the RPC API.

runtime/doc/lua.txt
```diff
@@ -3074,6 +3095,8 @@ vim.fs.parents({start})                                     *vim.fs.parents()*
         (`string?`)
 
 vim.fs.rm({path}, {opts})                                        *vim.fs.rm()*
+    WARNING: This feature is experimental/unstable.
```
justinmk, line 3098: Unreleased APIs are marked as experimental. This means `make doc` is needed when doing a release...

zeertzjq: `release.sh` already does that.

PR #30661, opened 2024-10-04
docs: update LspAttach and LspDetach examples

gpanders:
The current LspAttach example shows setting options which are already set by default. We should expect that users are going to copy-paste these examples, so we shouldn't use examples that are superfluous and unnecessary.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -87,8 +87,13 @@ To override or delete any of the above defaults, set or unset the options on
 
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(args)
+        -- Unset 'formatexpr'
         vim.bo[args.buf].formatexpr = nil
+
+        -- Unset 'omnifunc'
         vim.bo[args.buf].omnifunc = nil
+
+        -- Unmap K
         vim.keymap.del('n', 'K', { buffer = args.buf })
       end,
     })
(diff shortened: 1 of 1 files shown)


runtime/doc/lsp.txt
```diff
@@ -87,8 +87,13 @@ To override or delete any of the above defaults, set or unset the options on
 
     vim.api.nvim_create_autocmd('LspAttach', {
       callback = function(args)
+        -- Unset 'formatexpr'
         vim.bo[args.buf].formatexpr = nil
+
+        -- Unset 'omnifunc'
         vim.bo[args.buf].omnifunc = nil
+
+        -- Unmap K
```
justinmk, line 96:
```suggestion
        -- Unset 'omnifunc'
        vim.bo[args.buf].omnifunc = nil
        -- Unmap K
```

justinmk (approved):

MariaSolOs (approved):

runtime/doc/lsp.txt
```diff
@@ -523,32 +527,25 @@ EVENTS                                                            *lsp-events*
...
         local client = vim.lsp.get_client_by_id(args.data.client_id)
-        -- Do something with the client
-        vim.cmd("setlocal tagfunc< omnifunc<")
+
+        -- Remove the autocommand to format the buffer on save, if it exists
+        if client.supports_method('textDocument/formatting') then
+            vim.api.nvim_clear_autocmds({
+                event = 'BufWritePre',
+                buffer = args.buf,
+            })
```
zeertzjq, line 547: Why 4-space indent here?

gpanders: https://github.com/neovim/neovim/pull/30709

PR #30665, opened 2024-10-05
fix(coverity/497355): shada_read_when_writing out of bounds read

devon-g:
Problem:
There appears to be an intentional array out of bounds read when indexing global and numbered marks since they are adjacent in the struct that holds them.

Solution:
Explicitly index numeric marks array to avoid reading out of bounds from global marks array.

### DIFF
--- a/src/nvim/shada.c
+++ b/src/nvim/shada.c
@@ -1886,13 +1886,18 @@ static inline ShaDaWriteResult shada_read_when_writing(FileDescriptor *const sd_
           shada_free_shada_entry(&entry);
           break;
         }
-        if (wms->global_marks[idx].data.type == kSDItemMissing) {
+
+        // Global or numbered mark.
+        PossiblyFreedShadaEntry *mark
+          = idx < 26 ? &wms->global_marks[idx] : &wms->numbered_marks[idx];
+
+        if (mark->data.type == kSDItemMissing) {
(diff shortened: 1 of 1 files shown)


src/nvim/shada.c
```diff
@@ -1886,13 +1886,21 @@ static inline ShaDaWriteResult shada_read_when_writing(FileDescriptor *const sd_
...
           break;
         }
-        if (wms->global_marks[idx].data.type == kSDItemMissing) {
+
+        PossiblyFreedShadaEntry *global_or_numbered_mark;
+        if (idx < 26) {
+          global_or_numbered_mark = &wms->global_marks[idx];
+        } else {
+          global_or_numbered_mark = &wms->numbered_marks[idx];
+        }
```
justinmk, line 1895: This is a very short-lived variable so doesn't need an elaborate name.

```suggestion
        // Global or numbered mark.
        PossiblyFreedShadaEntry *mark = idx < 26 ? &wms->global_marks[idx] : &wms->numbered_marks[idx];
```

justinmk (approved):

src/nvim/shada.c
devon-g: Appended your suggestion, thanks!

devon-g: Oh I didn't see that you already applied them, my mistake.

Issue #22980, opened 2023-04-09
Tracking issue: spring-cleaning the runtime


clason:
### Problem

We have inherited (and still actively maintain by porting updates to) Vim's runtime files, a lot of which are very useful and part of Neovim's value proposition. However, some of them have aged less well and are obsolete, broken, or otherwise useless (either due to divergent Neovim features or upstream neglect). Keeping these files around makes it harder to focus our limited maintenance budget and sets wrong expectations (that these functions are working and useful in Neovim). We should remove them or, if necessary, replace them with new, working, implementations:
* [x] Vimballs: obsolete with packages #22402
* [x] `:CheckHealth`: obsolete with :checkhealth #23031
* [x] `bugreport.vim`: obsolete with :checkhealth #23053
* [ ] `macros`: compatibility shims for `packadd`, or scripts that replicate standard Neovim features (`editexisting.vim`, `less.vim`)
* [ ] `tools`: `check_colors.vim` (broken on Neovim), ~`emoji_list.vim`~ (actually a Lua script, move to `lua`?)
* [ ] `macmap.vim` `mswin.vim` (different defaults)
* [ ] `pack/dist/opt`: `swapmouse` (really?), `shellmenu` (targets gVim?)
* [ ] ...

### Expected behavior

Eternal sunshine of the spotless runtime.

---

## Proposal

Consolidate all "one-off" scripts we want to keep to a new `runtime/scripts` directory (making sure to update the docs accordingly). Err on the side of keeping stuff for now.

craigmac:
How about our own ftplugin/lua.lua, or an after/ftplugin/lua.lua (I'd prefer this) so we can tailor lua experience with neovim specific things in mind, things that upstream won't accept in their ftplugin/lua.vim, like making K work better with vim.api.nvim_

dundargoc:
You can already put lua files in `ftplugin` and `after/ftplugin` right now I think? Maybe I misunderstood something.

zeertzjq:
- When removing a runtime file, also remove the reference to it from docs.
- Please don't remove `emoji_list.vim`. What's the cost of keeping it?

clason:
> What's the cost of keeping it?

Gonna flip that on you: why do you care?

zeertzjq:
> > What's the cost of keeping it?
> 
> Gonna flip that on you: why do you care?

It can be used to check if Nvim and host terminal's character width assumptions match.

clason:
Shouldn't that be a health check then?

[...]

PR #30668, opened 2024-10-05
fix(runtime): fully port emoji_list to Lua

clason:
Problem: `runtime/tools/emoji_list.vim` is a Lua script masquerading as Vimscript, which is unnecessary now that `:source` works for Lua files.

Solution: Remove Vimscript wrapper.

Question: Is there a better place for this script? E.g., `runtime/lua/emoji_list.lua`? Or is this too niche to pollute the `require` namespace with?


justinmk:
> Is there a better place for this script? E.g., `runtime/lua/emoji_list.lua`? Or is this too niche to pollute the `require` namespace with?

It could live in `runtime/pack/dist/opt/` . The "tools" subdir should be eliminated, either those should live in scripts/ or if they are wanted by users then `runtime/pack/dist/opt/` is the place for them.

clason:
> It could live in runtime/pack/dist/opt/ .

Hmmm... This feels a bit wrong; those are for "proper" plugins (i.e., plugins that set stuff up and then react to commands or mappings), while this is a single-shot script. It's most similar to stuff directly under `runtime/` like `macmap.vim` (although those are hateful, too).

Maybe a separate `runtime/scripts` folder, which can collect all those "one-off" scripts (that we want to keep!) from `runtime/`, `tools/`, and `pack/dist/opt` (like `swapmouse` and `shellmenu` -- which I'd put on the chopping block, personally)?

justinmk:
> Maybe a separate `runtime/scripts` folder

sure. the stuff in runtime/macros/ could also go in there.

clason:
> sure. the stuff in runtime/macros/ could also go in there.

If we want to keep it... How do we decide? Do you want to make an executive decision?

clason:
I any case, this need not block this PR and is better tackled in the context of #22980. I'll start chipping away on that in a follow-up PR.

justinmk:
> Do you want to make an executive decision?

no 😂 

I think moving these various scripts to one place is a good improvement to start with. There might be one or two scripts that obviously should be dropped entirely, but anything marginally useful can stay.

clason:
Even `swapmouse.vim`?

PR #30672, opened 2024-10-05
Update filetype.lua to cover missing apache_dotconf names

nisbet-hubbard:
Currently, detection patterns only match Apache config files supplied in distros. As a result, the following dotconf files [provided by the upstream source tarball](https://github.com/apache/httpd/tree/trunk/docs/conf/extra) aren’t correctly highlighted as `apache` filetype by Neovim

> httpd-autoindex.conf
httpd-dav.conf
httpd-default.conf
httpd-info.conf
httpd-languages.conf
httpd-manual.conf
httpd-mpm.conf
httpd-multilang-errordoc.conf
httpd-ssl.conf
httpd-userdir.conf
httpd-vhosts.conf
proxy-html.conf

This PR fixes the issue.


clason:
Thank you, but the filetype detection logic is taken from Vim (even if the implementation is different). Please contribute any changes there (unless you can demonstrate that this is a regression, i.e., it works in Vim).

PR #30674, opened 2024-10-05

clason:
- **vim-patch:af449f6: runtime(compiler): add cppcheck linter compiler plugin**
- **vim-patch:7c3f9af: runtime(misc): unset compiler in various ftplugins**

### DIFF
--- a/runtime/ftplugin/context.vim
+++ b/runtime/ftplugin/context.vim
@@ -12,10 +12,6 @@ let b:did_ftplugin = 1
 let s:cpo_save = &cpo
 set cpo&vim
 
-if !exists('current_compiler')
-  compiler context
-endif
-
 let b:undo_ftplugin = "setl com< cms< def< inc< sua< fo< ofu<"
 
 setlocal comments=b:%D,b:%C,b:%M,:% commentstring=%\ %s formatoptions+=tjcroql2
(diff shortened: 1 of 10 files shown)


runtime/ftplugin/context.vim
```diff
@@ -108,6 +104,11 @@ if get(g:, 'context_mappings', 1)
     endif
 endif
 
+if !exists('current_compiler')
+  b:undo_ftplugin ..= "| compiler make"
+  compiler context
+endif
```
zeertzjq, line 110:
```suggestion
endif

```

runtime/ftplugin/context.vim
```diff
@@ -108,6 +104,11 @@ if get(g:, 'context_mappings', 1)
...
+b:undo_ftplugin ..= "| sil! delc -buffer ConTeXt | sil! delc -buffer ConTeXtLog | sil! delc -buffer ConTeXtJobStatus | sil! delc -buffer ConTeXtStopJobs"
```
zeertzjq, line 111:
```suggestion
  let b:undo_ftplugin ..= "| compiler make"
  compiler context
endif

let b:undo_ftplugin ..= "| sil! delc -buffer ConTeXt | sil! delc -buffer ConTeXtLog | sil! delc -buffer ConTeXtJobStatus | sil! delc -buffer ConTeXtStopJobs"
```

clason: Sorry, forgot that this was vim9script. Should I drop the change entirely? I think that would be more consistent.

zeertzjq: If you drop the change than it's inconsistent with other runtime files :)

zeertzjq (approved):

PR #30677, opened 2024-10-05
ci: bump macos runner version to macos-15

dundargoc:


clason:
Ruby oldtest fails due to different (expected) error message; probably just a newer Ruby version on the runner. Let's wait for Vim to hit and fix that.

zeertzjq:
It's already fixed in Vim. See #30683.

PR #30685, opened 2024-10-06
fix(coverity/508289): vterm lookup_colour dead code

devon-g:
argcount is compared to zero when it cannot be zero

return only valid branch instead of checking


devon-g:
Forgot I made this same pull request a month ago my bad.

PR #30687, opened 2024-10-06
fix(coverity/510275): linematch out of bounds access

devon-g:
Problem:
Int pointer cast to unsigned long pointer causes potential memory corruption.

Solution:
Cast and store value first, then pass the new pointer.

### DIFF
--- a/src/nvim/linematch.c
+++ b/src/nvim/linematch.c
@@ -150,7 +150,9 @@ static int count_n_matched_chars(mmfile_t **sp, const size_t n, bool iwhite)
 mmfile_t fastforward_buf_to_lnum(mmfile_t s, linenr_T lnum)
 {
   for (int i = 0; i < lnum - 1; i++) {
-    s.ptr = strnchr(s.ptr, (size_t *)&s.size, '\n');
+    size_t n = (size_t)s.size;
+    s.ptr = strnchr(s.ptr, &n, '\n');
+    s.size = (int)n;
     if (!s.ptr) {
       break;
     }


src/nvim/linematch.c
```diff
@@ -150,7 +150,8 @@ static int count_n_matched_chars(mmfile_t **sp, const size_t n, bool iwhite)
 mmfile_t fastforward_buf_to_lnum(mmfile_t s, linenr_T lnum)
 {
   for (int i = 0; i < lnum - 1; i++) {
-    s.ptr = strnchr(s.ptr, (size_t *)&s.size, '\n');
+    size_t n = (size_t)s.size;
+    s.ptr = strnchr(s.ptr, &n, '\n');
```
zeertzjq, line 154: You need to assign the new value back to `s.size`

devon-g: How embarrassing... it has been fixed, thanks!

src/nvim/linematch.c
```diff
@@ -150,7 +150,9 @@ static int count_n_matched_chars(mmfile_t **sp, const size_t n, bool iwhite)
...
+    s.size = n;
```
zeertzjq, line 155:
```suggestion
    s.size = (int)n;
```

devon-g: Yeah I see that it failed the checks because of that but I'm a bit confused why my machine isn't failing to compile. I tried make lint and compiling with gcc, clang, and zig cc and they all allow it for some reason but the checks here fail.

zeertzjq: Because `-Werror` is only used on CI

devon-g: Ah I see that makes sense, thanks!

PR #30690, opened 2024-10-06
fix(api): adjust Visual position after nvim_buf_set_text

### LINKED ISSUE #29558: The "other end" of a visual selection does not move when using `nvim_buf_set_text`
### Problem

When using `nvim_buf_set_text` to insert text in front of a visual selection, the "other end" does not move, it stays on the same column. The cursor, on the other hand, does move in a reasonable way.

This is causing problems in our [collaborative editing project](https://github.com/ethersync/ethersync). We could probably fix it on our end, but this seems like a Neovim bug.

### Steps to reproduce

- `nvim --clean`
- `i1234<Esc>`
- `:lua vim.defer_fn(function() vim.api.nvim_buf_set_text(0,0,0,0,0,{"0"}) end, 3000)` (This will insert a 0 in front of the line in 3 seconds.)
- While the timeout is running, `hhvl` to create a visual selection containing the "23".

Result: The cursor moves one column to the right (and stays on the 3), but the "other end" of the visual selection stays on the second column (and thus moves from the 2 to the 1).

### Expected behavior

The "other end" should also be pushed to the right, like the cursor.

### Neovim version (nvim -v)

0.10.0

Vim doesn't have nvim_buf_set_text

### Operating system/version

NixOS unstable

### Terminal name/version

kitty 0.35.1

xterm-kitty

### Installation

nixpkgs

glepnir:
Problem:
Visual selection could end up in the wrong place after
nvim_buf_set_text or nvim_buf_set_lines. In some delete cases,
Visual.lnum was already clamped before the line shift happened, so the
adjustment got skipped.

Solution:
Split fix_cursor_cols into reusable fix_pos_col logic and reuse it
for Visual updates. Also adjust Visual.lnum before changed_lines so
the shift uses the original position before final clamping.

Fix #29558

### DIFF
--- a/src/nvim/api/buffer.c
+++ b/src/nvim/api/buffer.c
@@ -439,6 +439,13 @@ void nvim_buf_set_lines(uint64_t channel_id, Buffer buf, Integer start, Integer 
     mark_adjust_buf(b, (linenr_T)start, (linenr_T)(end - 1), adjust, (linenr_T)extra,
                     true, kMarkAdjustApi, kExtmarkNOOP);
 
+    if (VIsual_active && b == curbuf && VIsual.lnum >= (linenr_T)start) {
+      if (VIsual.lnum >= (linenr_T)end) {
+        VIsual.lnum += (linenr_T)extra;
+      }
+      check_visual_pos();
+    }
+
     extmark_splice(b, (int)start - 1, 0, (int)(end - start), 0,
                    deleted_bytes, (int)new_len, 0, inserted_bytes,
                    kExtmarkUndo);
@@ -665,6 +672,12 @@ void nvim_buf_set_text(uint64_t channel_id, Buffer buf, Integer start_row, Integ
     linenr_T adjust = end_row >= start_row ? MAXLNUM : 0;
     mark_adjust_buf(b, (linenr_T)start_row, (linenr_T)end_row - 1, adjust, (linenr_T)extra,
                     true, kMarkAdjustApi, kExtmarkNOOP);
+
+    if (VIsual_active && b == curbuf && VIsual_mode != Ctrl_V) {
+      fix_pos_col(b, &VIsual, (linenr_T)start_row, (colnr_T)start_col, (linenr_T)end_row,
+                  (colnr_T)end_col, (linenr_T)new_len, (colnr_T)last_item.size, 1);
+      check_visual_pos();
+    }
 
     extmark_splice(b, (int)start_row - 1, (colnr_T)start_col,
                    (int)(end_row - start_row), col_extent, old_byte,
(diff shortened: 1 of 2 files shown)


blinry:
I just re-encountered this issue (#29558), when collaboratively editing in [Ethersync](https://github.com/ethersync/ethersync).

@glepnir, what's the status of this PR? :)

glepnir:
Oh, this PR has been around for quite a while. I'll take a look at it in the next few days.

glepnir:
@blinry Hello  I’ve updated the PR. I used your reproduction steps to write a test case and it passes. Could you give this PR a try?

blinry:
Thanks to getting back to this so quickly, @glepnir! I tried the PR – this solves the issue in many cases!

But I still found a case where the selection shifts around: When inserting or removing lines before the selection. Is it possible that the columns are shifted now, but the rows are not (yet)?

## To reproduce

Start with the buffer content

```
1
2
3
```

Run

```
:lua vim.defer_fn(function() vim.api.nvim_buf_set_text(0,0,0,0,0,{"0",""}) end, 3000)
```

to insert a line "0" at the front in 3 seconds. While the timeout is running, visually select the line "2" with `V`.

Result: Lines "1" and "2" are selected.
Expected result: Only line "2" is selected.

glepnir:
I have update. now should works. because before just modified the column logic in fix_cursor_cols. the new bug you mentioned need a smiliar logic for line in fix_cursor.

blinry:
This is a big improvement! I found two more edge case related to insertions in the first line of the selection, maybe you have an idea? Maybe both have the same solution. (This is kind of fun, I hope it is for you, as well!)

# Case 1

Start with buffer content

```
1
2
```

Run this to insert a "0" in front of the "1":

```
:lua vim.defer_fn(function() vim.api.nvim_buf_set_text(0,0,0,0,0,{"0"}) end, 3000)
```

Select both lines while the timeout is running.

Result: The first line is selected
Expected result: Both lines are still selected

# Case 2

Start with buffer content

```
123
```

Run this to insert a newline in front of the "1":

```
:lua vim.defer_fn(function() vim.api.nvim_buf_set_text(0,0,0,0,0,{"", ""}) end, 3000)
```

Select the 2 visually while the timeout is running.

Result: The newline, as well as the 1 and 2 are selected
Expected result: The 2 is still selected

glepnir:
case1  the visual start should still at `1` not `0` right?

glepnir:
another case when using visual-block..still need adjust..and there also need a similar change in nvim_buf_set_lines.

blinry:
I tested the current state of this branch:

- line-wise selection: works perfectly! :sparkles:
- block selection: as you said, seems to need adjustment of the "other end"
- regular selection: if text is inserted in the same line but *after* the column of the "other end", its position is (falsely) shifted to the right

## To reproduce

Start with `123`, visually select the 2, then insert a 4 at the end of the line.

Expected result: 2 is still selected
Actual result: 23 is selected

glepnir:
Here are the edge cases that need to be handled. What confuses me is whether the visual start or end should follow the selected cell or remain fixed at the mark (row, col). If it's fixed, when the row/col changes (e.g., the original line is replaced with an empty line), the visual start or end column may no longer be valid. The case with Visual-block is different because it always maintains a rectangle. There are many edge cases to consider here. Vim only has setline for full line replacement. When the visual start becomes invalid, Vim exits Visual mode. Alternatively, a simpler approach could be to prevent text insertion while Visual mode is active, keeping it in sync. @justinmk  any idea...

blinry:
- Preventing buffer modifications when Visual mode is active would break our use-case of collaborative text editing, so I would not go in that direction.
- For block-wise selections, it's indeed not entirely clear what the "expected" behavior should be. But to me, it feels like the block is spanned by the cursor and the "other end", and both should be pushed around in the same way by edits. That would feel consistent with the way I'd expect regular selections to behave.
- When setting a line that contained the cursor or "other end" to be empty, my favorite solution would be to let the position "fall back" to the first column. That's what's currently happening to the cursor position under edits already.

Do you have an idea for what needs to be done to not shift the "other end" in case of an insertion after its column, in the case of a regular selection? That's the last "To reproduce" case I described in my last comment. I think if we could fix that, this would be a big improvement to the status quo! We could tackle `nvim_buf_set_lines` in a follow-up PR.

glepnir:
> Do you have an idea for what needs to be done to not shift the "other end" in case of an insertion after its column, in the case of a regular selection? 

Adding some start and end col checks in the current PR should be able to handle it. But this will add some extra code... Maybe not what Justin wants because he wants to keep it simple. I am currently just confused about how to adjust the mark in different visual modes. Visual and Visual block cannot be processed synchronously. the visual-block mode is skipped in the current PR...

blinry:
Yeah, maybe we can get some input from @justinmk on the most elegant solution. :)

And thanks for your work so far, @glepnir, this will solve a real problem for us!

blinry:
We'd still love to see this solved, for our "Collaborative Editing Protocol" project of Teamtype! Let us know if we can provide any more input to this.

glepnir:
@blinry Could you try out this PR now ?

src/nvim/api/buffer.c
```diff
@@ -1277,6 +1284,36 @@ static void fix_cursor(win_T *win, linenr_T lo, linenr_T hi, linenr_T extra)
   }
 }
 
+/// Adjust a position's col/lnum after text replacement between
+/// (start_row, start_col) and (end_row, end_col).
+static void fix_pos_col(pos_T *pos, linenr_T start_row, colnr_T start_col, linenr_T end_row,
+                        colnr_T end_col, linenr_T new_rows, colnr_T new_cols_at_end_row,
```
justinmk: why doesn't `fix_cursor_cols` do this already? what are the differences? how many functions do we have like this? can we consolidate them or do they have any common code that could be shared?

src/nvim/api/buffer.c
justinmk, line 687: off topic for this PR, I wonder if we can extract this `mark_adjust_buf` + fix_pos_col + extmark_splice + changed_lines to a util function. This pattern is done in many places in the codebase.

glepnir: yeah I tried this, the line mode and text mode paths look similar but they actually diverge in a few annoying ways (end_row inclusive vs exclusive, VIsual fixup timing, the col args being all zero vs real values). any single helper ends up either with a misleading end_row param that means different things in each mode, or with like 11 parameters and a bool flag..

blinry:
This is a huge improvement, thank both of you so much for your work on this! \o/

PR #30691, opened 2024-10-06
fix(runtime): add comment string for gleam lang

lampewebdev:
This adds `commentstring` for the gleam lang so that `gcc` and other comment key bindings work in gleam files.


echasnovski:
I believe this type of changes are better to first try to get merged into 'vim/vim'.

clason:
Not if the default `/*%s*/` commentstring is valid for Gleam as well.

clason:
Looks like the answer is "no": https://tour.gleam.run/functions/documentation-comments/

Then the change indeed has to go through Vim first.

PR #25073, opened 2023-09-09
feat(lsp): use treesitter for markdown floating windows

MariaSolOs:
Closes #20857, #20246 (?)

This PR includes the following changes:
- A new private function `_normalize_markdown` has been added, which transforms the input contents as follows:
  1. Trim out empty lines before and after the content. For simplicity I'm also removing the `pad_top` and `pad_bottom` options (open to discussion about this).
  2. Remove carriage returns.
  3. Consecutive blank lines are collapsed into a single one.
  4. Separator Markdown lines (e.g. `---`) will be replaced by a horizontal divider spanning the entire floating window and blank lines before and after the separator are removed.

Note that this means that this function won't be performing any additional highlighting or configure the buffer.
- `open_floating_preview` will now display the input content as is OR use Treesitter for displaying a Markdown buffer if the response received by the LSP request indicated that the content had a `'markdown'` `MarkupKind`. If the server indicates that the `MarkupKind` is `'plaintext'`, we'll respect that and display the contents as is.

- `convert_input_to_markdown_lines` will now simply return the unprocessed input when it is `MarkupContent` with kind '`plaintext`' (previously it would wrap it around `<text>` tags and display it as Markdown).

- `trim_empty_lines` is now deprecated since its functionality can be replaced by `vim.split(..., { trimempty = true })`

PR #30695, opened 2024-10-06
feat(lsp): improve LSP doc hover rendering

justinmk:
# Problem:
- Some servers like LuaLS add unwanted blank lines after multiline `@param` description.
    - LuaLS bug: https://github.com/LuaLS/lua-language-server/issues/2333
- List items do not wrap nicely.

# Solution:
- When rendering the LSP doc hover, remove blank lines (except in the main description).
- Set 'breakindent'.

## before/after

image

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -99,15 +99,21 @@ local function get_border_size(opts)
   return { height = height, width = width }
 end
 
----@param s string Multline string
----@param no_blank boolean? Drop blank lines after main description.
---- Workaround for https://github.com/LuaLS/lua-language-server/issues/2333
+--- Splits string at newlines, optionally removing unwanted blank lines.
+---
+--- @param s string Multiline string
+--- @param no_blank boolean? Drop blank lines for each @param/@return (except one empty line
+--- separating each). Workaround for https://github.com/LuaLS/lua-language-server/issues/2333
 local function split_lines(s, no_blank)
   s = string.gsub(s, '\r\n?', '\n')
   local lines = {}
   local in_desc = true -- Main description block, before seeing any @foo.
   for line in vim.gsplit(s, '\n', { plain = true, trimempty = true }) do
-    in_desc = (not line:find('^ ?%@.?param')) and in_desc or false
+    local start_annotation = not not line:find('^ ?%@.?[pr]')
+    in_desc = (not start_annotation) and in_desc or false
+    if start_annotation and no_blank and not (lines[#lines] or ''):find('^%s*$') then
+      table.insert(lines, '') -- Separate each @foo with a blank line.
+    end
     if in_desc or not no_blank or not line:find('^%s*$') then
       table.insert(lines, line)
     end
@@ -127,7 +133,7 @@ end
 --- Convenience wrapper around vim.str_utfindex
 ---@param line string line to be indexed
 ---@param index integer|nil byte index (utf-8), or `nil` for length
----@param encoding string|nil utf-8|utf-16|utf-32|nil defaults to utf-16
+---@param encoding 'utf-8'|'utf-16'|'utf-32'|nil defaults to utf-16
 ---@return integer `encoding` index of `index` in `line`
 function M._str_utfindex_enc(line, index, encoding)
   local len32, len16 = vim.str_utfindex(line)
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -99,9 +99,20 @@ local function get_border_size(opts)
   return { height = height, width = width }
 end
 
-local function split_lines(value)
-  value = string.gsub(value, '\r\n?', '\n')
-  return split(value, '\n', { plain = true, trimempty = true })
+---@param s string Multline string
+---@param no_blank boolean? Drop blank lines after main description.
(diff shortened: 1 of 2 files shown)


wookayin:
Removing unwanted blank lines is nice. But personally, I somehow like empty lines before `@param`, `@return`, etc. so they are clearly separated: e.g.
```diff
 @*param* `buffer` — Buffer handle, or 0 for current buffer 
+                                                                            
 @*param* `send_buffer` — True if the initial notification should contain the
  whole buffer: first notification will be `nvim_buf_lines_event`.
  Else the first notification will be `nvim_buf_changedtick_event`.
  Not for Lua callbacks.
+                                                                            
 @*param* `opts` — Optional parameters.
 - on_lines: Lua callback invoked on change.
 ...
 - preview: also attach to command preview (i.e. 'inccommand')
   events.
+                                                                            
 @*return* — False if attach failed (invalid parameter, or buffer isn't loaded);
  otherwise True. 
```
Would that be possible without making `split_lines` too complicated?

justinmk:
> personally, I somehow like empty lines before `@param`, `@return`, etc. so they are clearly separated: e.g.

That is a quirk of LuaLS. Clangd doesn't do this. If we want to change the presentation, we can always choose to do that (in a separate PR), but it's unrelated to this PR which is simply cleaning up junk in the server response.

mfussenegger (commented):
Isn't this overfitting towards luals? 

There might be other servers where this results in worse hover doc rendering

justinmk:
@mfussenegger I will revert/adjust if we find any such cases, I tested clangd/tsserver/lua. This is also temporary until https://github.com/LuaLS/lua-language-server/issues/2333 is fixed.

This logic should be harmless for non-LuaLS cases.

MariaSolOs:
Hmm I personally dislike hacks like this one that accommodate a misbehaving server. I also thought that one of the goals of #25073 was to simplify the markdown rendering of these floating windows and preventing this kind of modifications to the markup content.

justinmk:
> one of the goals of #25073 was to simplify the markdown rendering of these floating windows and preventing this kind of modifications to the markup content.

This doesn't conflict with that though.

The fact is, different servers add (luals) or remove (clangd) blank lines in random places. We're the presentation layer, why shouldn't we have a consistent display of the `@param` fields?

I don't like hacks, but,
- we've been suffering from this bug for 1+ years
- this hack is generalized, it's not specific to a server (except the assumption of `@x` annotations)
- this is easy to revert
- it's covered by tests (this behavior had no coverage before now)

Do we want to live with this bug? Is there an alternative?

runtime/lua/vim/lsp/util.lua
```diff
@@ -99,9 +99,26 @@ local function get_border_size(opts)
   return { height = height, width = width }
 end
 
-local function split_lines(value)
-  value = string.gsub(value, '\r\n?', '\n')
-  return split(value, '\n', { plain = true, trimempty = true })
```
justinmk, line 104: Why is this `trimempty` acceptable, but the new `split_lines` logic is not acceptable?

clason: Isn't that stuff we're trying to get rid of?

MariaSolOs:
> * we've been suffering from this bug for 1+ years

Well then maybe let's contribute the fix upstream? As mentioned in https://github.com/LuaLS/lua-language-server/issues/2333, the bug is reproducible outside neovim. Do we want all editors to create a fix like this one? Why use LSP at that point?

> * it's covered by tests (this behavior had no coverage before now)

I don't see why this is an argument in favor of the change. Of course I won't complain about adding a test, but I don't see why we need to test this particular broken case.

> Do we want to live with this bug? Is there an alternative? Why is the old split_lines acceptable but the new one isn't (specifically, which parts)?

I understand where you're coming from, but I try to maintain the general goal of "let's stick to the LSP and only the LSP" that we have all been trying to follow (you have even corrected me in previous PRs when our roles were reversed!). If a language server sends us broken markdown, I believe that we should not do anything to it and display it as is.

justinmk:
> contribute the fix upstream? As mentioned in [LuaLS/lua-language-server#2333](https://github.com/LuaLS/lua-language-server/issues/2333), the bug is reproducible outside neovim.

Even if that bug is fixed, "normalizing" empty/blank lines is something that will forever be needed, it's a general issue. This is a *presentation layer* decision, there is nothing in the LSP spec or markdown, spiritually or technically, that says we must show identical whitespace in our hover. Especially for TUIs, whitespace is how we do layout.

mfussenegger:
> , "normalizing" empty/blank lines is something that will forever be needed, it's a general issue. 

Normalizing blank linkes to fit the markdown definition of a paragraph would be fine, what's to me not fine is the assumption of a specific layout for `hover` and the `^ ?%@.?[pr]` handling. That's language specific and has no special meaning in markdown

justinmk:
> specific layout for `hover` and the `^ ?%@.?[pr]` handling

Yes, my rationalizations were flimsy.

Avoiding server-specific hacks in our client is our strategy for maintainability, so that we don't have an unhappy, buggy mess. Delegating to markdown as in #25073 is high-leverage. 

But if we've exhausted that path and can optimize presentation/UX with a low-risk, test-covered, isolated hack, let's be open to that.

I'm still open to reverting this, but can we wait to see if it causes a problem?

PR #30700, opened 2024-10-06
fix(lsp): tagfunc fails in unusual buffer

justinmk:
Problem:
tagfunc failed in a weird buffer (either a directory or some other non-file buffer, I don't remember):

    E987: Invalid return value from tagfunc
    E5108: Error executing lua …/runtime/lua/vim/lsp/util.lua:311: EISDIR: illegal operation on a directory
    stack traceback:

at this line:

    local data = assert(uv.fs_read(fd, stat.size, 0))

Solution:
Check for directory.

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -296,6 +296,9 @@ local function get_lines(bufnr, rows)
   end
 
   local filename = api.nvim_buf_get_name(bufnr)
+  if vim.fn.isdirectory(filename) ~= 0 then
+    return {}
+  end
 
   -- get the data from the file
   local fd = uv.fs_open(filename, 'r', 438)


runtime/lua/vim/lsp/util.lua
```diff
@@ -297,6 +297,11 @@ local function get_lines(bufnr, rows)
 
   local filename = api.nvim_buf_get_name(bufnr)
 
+  if vim.fn.isdirectory(filename) then
+    vim.fn.bufload(bufnr)
+    return buf_lines()
+  end
```
justinmk, line 303: Alternatively, could handle the `fs_read` result below. Should it silently return empty then? I don't fully understand what this function does.

runtime/lua/vim/lsp/util.lua
```diff
@@ -297,6 +297,11 @@ local function get_lines(bufnr, rows)
 
   local filename = api.nvim_buf_get_name(bufnr)
 
+  if vim.fn.isdirectory(filename) then
```
wookayin, line 300:
```suggestion
  if vim.fn.isdirectory(filename) == 1 then
```
`vim.fn.isdirectory` returns NUMBER (1 or 0).

mfussenegger (commented):
Wouldn't this just be masking bugs?

Directories have no `lines`/content in the sense that LSP could make sense of it because the representation is entirely arbitrary and up to a client.
So if you got references pointing to a directory there must have been something else wrong?

The non-file buffer case should already be handled via the `uri` `file` check further above.

runtime/lua/vim/lsp/util.lua
justinmk: oops, late night PR, thanks :)

justinmk:
> So if you got references pointing to a directory there must have been something else wrong?

I figured as much. I wrote down this error a ~week ago and didn't have time to investigate, so this PR is more of a discussion point.

> The non-file buffer case should already be handled via the `uri` `file` check further above.

Directories don't yield a file: URI?

mfussenegger:
> Directories don't yield a file: URI?

At least with netrw they seem to get a file uri. But I was also referring to your "some other non-file buffer" remark. 

I think if we handle directories here it should return an empty list. Because directories have no lines in the sense of a document. The buffer content would depend on a plugin like netrw - and that representation is certainly not something language servers will account for

mfussenegger (approved):

runtime/lua/vim/lsp/util.lua
```diff
@@ -296,6 +296,9 @@ local function get_lines(bufnr, rows)
   end
 
   local filename = api.nvim_buf_get_name(bufnr)
+  if vim.fn.isdirectory(filename) ~= 0 then
```
MariaSolOs, line 299: For my own learning: It feels odd for a function to be named `isX` without returning a boolean. Is this naming something that we have control over or is part of our vim legacy?

dundargoc: It's because in vimscript boolean `1` is true and `0` is false, but in lua this doesn't work as both of these are truthy. I've suggested adding a layer for vimscript boolean functions to return true and false instead of 1 and 0 but it was shot down due to adding extra magic between the vimscript<->lua layer (which I can't disagree with completely). It's not an optimal situation, I agree. I wonder if we couldn't supplement `vim.fs` with a function like this, since I feel like "is directory" is a useful and common enough check to get it's own dedicated function. Something like `vim.fs.isdir`?

justinmk: > Something like `vim.fs.isdir`?

Yeah we need that, though maybe `vim.fs.ftype() => 'file'|'dir'|...` would be better.

dundargoc: I would like to make a case against `vim.fs.ftype()`. I understand the API surface would be reduced, but I have a suspicion separate `isdir` and `isfile` functions would be more useful. My reasoning is that we already have something like `vim.fs.ftype` which is `vim.uv.fs_stat`. Despite this, both [neovim](https://github.com/neovim/neovim/blob/master/test/testutil.lua#L34-L43) and [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/util.lua#L117-L132) have created isdir/isfile helper functions. I think this points to there being a value in having separate isdir/isfile functions.

PR #30707, opened 2024-10-07
fix(runtime): Set comments and commentstring for RPM spec files

cryptomilk:
Set comments and commentstring for RPM spec files.

This allows to use `gc` in spec files.


glepnir:
for common files, please create a PR in upstream vim. then port to neovim via vim-patch

clason:
Yes, these files are maintained outside this repo. The maintainer is listed at the top of the file, but usually a PR to vim (with a ping) is fine.

PR #30721, opened 2024-10-08
feat(statusline): nvim_eval_statusline('') returns default statusline

### LINKED ISSUE #28444: nvim_eval_statusline() can return the default statusline
### Problem

An old limitation of N/Vim is that the default internal logic for empty 'statusline' option, is not exposed as a 'statusline' *expression*. We added `nvim_eval_statusline()` which is very nice, but still can't use it to get the default statusline result.

### Expected behavior

`nvim_eval_statusline()` should be able to yield the default statusline. It could look like any of : 

- Passing empty string: `nvim_eval_statusline('')`
- or, a new opts flag: `nvim_eval_statusline('',{default=true, ...})`

msaher:
Fixes #28444. This requires moving the default statusline implementation into the format DSL as opposed to C


src/nvim/statusline.c
```diff
@@ -382,7 +315,13 @@ static void win_redr_custom(win_T *wp, bool draw_winbar, bool draw_ruler)
       }
     } else {
       opt_idx = kOptStatusline;
+
       stl = ((*wp->w_p_stl != NUL) ? wp->w_p_stl : p_stl);
+      // statusline has a fallback when its NUL (unlike winbar)
```
justinmk, line 320:
```suggestion
      // For back-compat, statusline has a fallback when empty (unlike 'winbar').
```

src/nvim/statusline.c
```diff
@@ -62,6 +62,34 @@ typedef enum {
   kNumBaseHexadecimal = 16,
 } NumberBase;
 
+/// @return  fallback fmt string when 'statusline' is empty. Changes if 'rule' is set or if 'showcmdloc=statusline'
```
justinmk, line 65: ruler?
```suggestion
/// @return  fallback fmt string when 'statusline' is empty. Changes if 'ruler' is set or if 'showcmdloc=statusline'
```

justinmk:
> the C default status line implementation is a little tricky to translate into a format string

Might need to define a lua/vimscript function(s) which are part of the new default fmt. Or depending on the exact behavior, maybe a new 'statusline' flag is needed.

msaher:
I'll try to debug the tests to understand what's happening

justinmk:
FYI: to [run a single test](https://github.com/neovim/neovim/blob/master/test/README.md#filter-by-file) you can use TEST_FILE:

    TEST_FILE=test/functional/example_spec.lua make functionaltest

msaher:
Replicating the **exact** behavior of the C implementation is non-trivial. This is because it has quirks. Some are:

1. The exact placement of `'showcmd'`, `'ruler`' and `'keymap'` depend on the global variable `ru_col`. The Lua API does not expose `ru_col`
2. When the window size is small, the content of `'showcmd'` may write on top `'keymap'` (bug?). See screenshot below where I had `:set arabic` (adds `<arab>`) and the number of columns is was 17

c-statusline

3. The amount of padding required for items "on the right" depends on the width of "items on the left", so using `"%{minwid}.{maxwid}"` isn't straight forward.

Please note I'm new to neovim core so what I mentioned above may not 100% be accurate.

--- 

We could re-implement the default statusline as a vimscript/lua function, but trying to preserve the C implementation quirks makes it rather awkward. Here's a dirty inefficient attempt at translating the C logic into lua to illustrate what I mean (not exactly easy to maintain). `vim.api.nvim_get_ru_col()` is not part of the API, I created it myself to expose `ru_col`. 


```lua
-- strings are not immutable in lua. To avoid too many allocations consider using a string buffer
function statusline()
  local winid = vim.g.statusline_winid
  local bufid = vim.api.nvim_win_get_buf(winid)
  -- start with the filename
  local left = vim.api.nvim_eval_statusline("%f", {winid = winid}).str
  local result 

  local is_modified = vim.api.nvim_get_option_value("modified", { buf = bufid })
  local is_readonly = vim.api.nvim_get_option_value("readonly", { buf = bufid })
  local is_help = vim.api.nvim_get_option_value("buftype", { buf = bufid }) == "help"
  local is_preview = vim.api.nvim_get_option_value("previewwindow", { win = winid })

  -- if the buffer/window is special then prepend a space and add format specifiers
  -- we can't stick %h%w%m%r directly because an extra space is only included if the buffer is special
  if is_modified or is_readonly or is_help or is_preview then
    left = left .. " "
  end

  if is_help then
    left = left .. "[Help]"
  end
  if is_preview then
    left = left .. "[Preview]"
  end
  if is_modified then
    left = left .. "[+]"
  end
  if is_readonly then
    left = left .. "[RO]"
  end

  local is_stl_global = vim.o.statusline == 3
  local stl_width
  if is_stl_global then
    stl_width = vim.o.columns
  else
    stl_width = vim.api.nvim_win_get_width(winid)
  end

  local ru_col = vim.api.nvim_get_ru_col() -- TODO: not actually part of the API. Just a hacky function I added temporarily for demonstration purposes
  local this_ru_col = math.max(ru_col - (vim.o.columns - stl_width), (stl_width + 1) / 2);

  if this_ru_col <= 1 then
    left = "<" -- no room for file name!
  else
    local len = string.len(left)
    if len >= this_ru_col then
      first_char_idx = len-this_ru_col-1
      left = string.sub(left, first_char_idx)
      left = "<" .. left
    end
  end

  local fillchar
  if winid == vim.api.nvim_get_current_win() then
    fillchar = vim.opt.fillchars:get("stl")
  else
    fillchar = vim.opt.fillchars:get("stlnc")
  end
  -- fallback to " " if empty
  if vim.tbl_isempty(fillchar) then
    fillchar = " "
  end

  if vim.o.showcmd and vim.o.showcmdloc == "statusline" then
    local showcmd_width = math.min(10, this_ru_col - string.len(left) - 2);
    if showcmd_width > 0 then
      local showcmd_str = vim.api.nvim_eval_statusline("%S", {}).str
      local showcmd_at = this_ru_col-showcmd_width-1
      local padding_needed = showcmd_at-string.len(left) 
      result = left .. string.rep(fillchar, padding_needed) .. showcmd_str
    else
      reuslt = left
    end
  else
    result = left
  end

  -- 'keymap'
  -- TODO: handle case where showcmd_str is allowed to overwrite keymap_name
  local keymap_name = nil
  if vim.b.keymap_name ~= nil then
    keymap_name = vim.api.nvim_eval_statusline("%k", {}).str
    if this_ru_col-string.len(result) > string.len(keymap_name) + 1 then
      local showkeymap_at = this_ru_col - string.len(keymap_name) - 1
      -- padding_needed = -(this_ru_col-padding_needed - showkeymap_at)
      pdding_needed = showkeymap_at-string.len(result)
      result = result .. string.rep(fillchar, padding_needed) .. keymap_name
    end
  end

  if vim.o.ruler then
    local ruler
    local position = vim.api.nvim_eval_statusline("%(%l,%c%V%)", { winid = windid}).str
    local percentage = vim.api.nvim_eval_statusline("%P%%", { winid = winid }).str

    -- compute space between ruler and percetnage
    if this_ru_col + string.len(position)+string.len(percentage)+1 < stl_width then -- +1 because at least we need one space
      local padding_needed = stl_width-this_ru_col-string.len(position)-string.len(percentage)+1
      ruler = position .. string.rep(fillchar, padding_needed) .. percentage
    else
      ruler = position
    end

    local padding_needed = this_ru_col-string.len(result)
    result = result .. string.rep(fillchar, padding_needed) .. ruler

  end

  return result

end
```

The `ru_col` api:

```C
Integer nvim_get_ru_col(void)
  FUNC_API_SINCE(11)
{
  return (Integer) ru_col;
}
```

--- 

How should we proceed given this information?

I think that adding a relatively simple (unlike the above) vimscript/lua function that mimics the C statusline without preserving the same quirks is a sound idea since it will avoid the extra complexity and most users will likely not notice the difference.

justinmk:
> I think that adding a relatively simple (unlike the above) vimscript/lua function that mimics the C statusline without preserving the same quirks

Yeah, that is what we want to evaluate. If there are minor differences then it may be acceptable. Moving the C code to a similar or bigger amount of Lua code isn't really useful.

Does `v:echospace` help with positioning?

msaher:
> Does v:echospace help with positioning?

I don't think so. `v:echospace` is equivalent to `sc_col-1`, but `sc_col` is independent of `ru_col` when there's a status line.

---

Where should a hypothetical lua statusline live? Right now I have something like this

```lua
function default_statusline()
  local buffer = "%<%f %h%m%r%="

  local has_showcmd = vim.o.showcmd and vim.o.showcmdloc == "statusline"
  local has_keymap = vim.b.keymap_name ~= nil

  local showcmd = has_showcmd and (has_keymap and "%-4.S ") or "%-11.S" or ""
  local keymap = has_keymap and "%k " or ""
  local ruler = vim.o.ruler and "%-14.(%l,%c%V%) %P" or ""

  return buffer .. showcmd .. keymap .. ruler
end
```

If we want to maintain the convention that `vim.o.statusline = ""` means "use the default statusline", then `win_redr_custom()` must call this *lua* function when it finds out that `'statusline'` is empty, but I'm unsure if this is currently possible.

justinmk:
> Where should a hypothetical lua statusline live?

I guess in `runtime/lua/vim/_defaults.lua`. But the more I think about this, the more I'm leaning towards avoiding that entirely. Instead, can you describe what currently isn't possible with the `%x` flags in 'statusline' ? And then we need to decide if we want to add new flags to support that, or just drop those "features".

msaher:
>The keymap_name ? %k : '' expression can be inlined in a 'statusline' expression

I needed a trailing space `"%k "` to mimic the C implementation, otherwise it would "hug" the ruler part. (Should've been a trailing `fillchar` instead of space now that I think about it).

> What is blocking us from a self-contained 'statusline' expression instead of a function? That is the goal.

1) In the C implementation, `'showcmd'` is placed independently of `'keymap'`. See screenshots below. What this means is that the minimum width taken by `"%S"` changes depending on the existence of a `'keymap'`.

keymap

Furthermore, the *exact* space between the `'showcmd'` part and `'keymap'` part depends on the *length* of the `b:keymap_name`. See screenshot below.

image


Suggestion: maaaybe swap the order of and `'showcmd'` and `'keymap'`. That way the width taken by `'showcmd'` won't be dynamic.

2) In the C implementation, the `'keymap'` part may be omitted if there's not enough "room" for it. See snippet below from `win_redr_status()`. A 'statusline' expression can't conditionally hide it.

```c
if (get_keymap_str(wp, "<%s>", NameBuff, MAXPATHL)
    && this_ru_col - len > (int)strlen(NameBuff) + 1) {
  grid_line_puts(off + this_ru_col - (int)strlen(NameBuff) - 1, NameBuff, -1, attr);
}
```

Suggestion: Drop "feature". Only affects small windows

3) In the C implementation, the `showcmd` part may be omitted as per the snippet below from `win_redr_status()`.

```c
// Draw the 'showcmd' information if 'showcmdloc' == "statusline".
if (p_sc && *p_sloc == 's') {
  const int sc_width = MIN(10, this_ru_col - len - 2);

  if (sc_width > 0) {
    grid_line_puts(off + this_ru_col - sc_width - 1, showcmd_buf, sc_width, attr);
  }
}
```

Suggestion: Drop "feature". Only affects small windows

4) The "ruler" part (`"%-14.(%l,%c%V%) %P"`) is always shown regardless of the value of `'ruler'`, but the C implementation only adds it if `'ruler'` is set.

5) The C implementation may call `win_redr_custom()` if the user has set a custom `'rulerformat'`. A 'statusline' expression hard codes the "ruler" part.

6) The `"%P"` part (for percentage) is always shown. But in the C implementation, it's only added if there's "room" for it. See the condition from `win_redr_ruler()` below.

7) The padding between `"%l,%c%V%"` and `" %P"` is normally `14`, but in the C implementation it's dynamically calculated to ensure it doesn't take too much space. See the loop from `win_redr_ruler()` below. [^1].

[^1]: Here `i` is the length of the `%l,%c%V%` part, and `o` is the length of `%l,%c%V% %P`. We add a `fillchar` every iteration as padding

```c
int i = (int)strlen(buffer);
get_rel_pos(wp, buffer + i + 1, RULER_BUF_LEN - i - 1);
int o = i + vim_strsize(buffer + i + 1);

....

// Never use more than half the window/screen width, leave the other half
// for the filename.
int this_ru_col = MAX(ru_col - (Columns - width), (width + 1) / 2);
if (this_ru_col + o < width) {
  // Need at least 3 chars left for get_rel_pos() + NUL.
  while (this_ru_col + o < width && RULER_BUF_LEN > i + 4) {
    i += (int)schar_get(buffer + i, fillchar);
    o++;
  }
  get_rel_pos(wp, buffer + i, RULER_BUF_LEN - i);
}
```

---

I do agree that a 'statusline' expression is ideal. It will make it easier for users to customize it.

justinmk:
Great analysis, thanks. Those compromises don't sound bad to me. Let's start with what can be expressed as a self-contained 'statusline' expression and see how it behaves. If there is a truly needed feature, we can track that as an enhancement to the 'statusline' expressions.

Note that meanwhile this might break a lot of tests, we'll have to see how many...

msaher:
I added ` "%<%f %h%m%r%=%-10.S%k %-14.(%l,%c%V%) %P"` as a start.

> Those compromises don't sound bad to me

Same. They mostly impact small windows which I don't think are very common.

>If there is a truly needed feature, we can track that as an enhancement to the 'statusline' expressions.

I think the main enhancement we might need involves `'ruler'`. As of now, a self-contained 'statusline' expression will break `'ruler'` (doing `:set noruler` won't actually disable the `"%(%l,%c%V%) %P"` part). It also doesn't respect a custom `'rulerformat'`.

---

Suggestion: Consider adding a `'ruler'` flag `"%u"` (`"%r"` is already taken). Maybe something like this: 

1) when "`ruler"` is on, `vim.o.statusline = "%f %u"` is equivalent to `vim.o.statusline = "%f " .. vim.o.rulerformat`.
2) when "`ruler`" is off,  `vim.o.statusline = "%f %u"` is equivalent to `vim.o.statusline = "%f "`. 

Additionally, If `rulerformat` is empty then fallback to something that resembles the C ruler like `"%-14.(%l,%c%V%) %P"`. This will allow us to simplify `win_redr_ruler()` as well since we're essentially getting rid of the C implementation of `ruler`.

Still, this doesn't solve point 6 and 7 from the above post.

Side note: I'm not sure how useful is the concept of distinguishing`'rulerformat'` from `'statusline'` but I suppose we have to preserve it for backward compatibility.

msaher:
> Suggestion: Consider adding a `'ruler'` flag `"%u"` (`"%r"` is already taken). Maybe something like this: 
> 1) when "`ruler"` is on, `vim.o.statusline = "%f %u"` is equivalent to `vim.o.statusline = "%f " .. vim.o.rulerformat`.

I did something similar except that it respects `ru_col`. This has the desirable side effect that the space taken by 'ruler' is dynamic. With `vim.o.rulerformat = "%l,%c%V%= %P"`:

image

image

I believe this fixes point 7 from the analysis post above.

>7. The padding between "%l,%c%V%" and " %P" is normally 14, but in the C implementation it's dynamically calculated to ensure it doesn't take too much space. See the loop from win_redr_ruler() below.

---

Implementation-wise, this adds a little complexity because  "%u" recursively calls `build_stl_str_hl()`, but since that function uses static variables, we have to add an offset to ensure the called `build_stl_str_hl()` doesn't overwrite the caller `build_stl_str_hl()`'s static memory.

It would be easier to just allocate new memory, but given how frequently this function is called I decided its better to re-use existing memory instead

justinmk:
Can you rebase this? Try `git fetch --all && git merge --squash origin/master`

PR #26507, opened 2023-12-10
fix(lsp): lsp.buf.definition() handles responses from multiple servers

antonk52:
Resolves #22318 when vim.lsp.buf.definition() subsequently jumps to multiple locations if more than one server responds.

Motivation: In case of multiple servers supporting the same lsp method, neovim should merge the results and handle them the same way it currently handles multiple results from a single server. This can reduce confusion by removing jumping to multiple locations without the user's action. In comparison vscode merges responses from its plugins/lsp servers and presents them to the user in a single picker. This is also partially solved in [lspsaga](https://nvimdev.github.io/lspsaga/definition/), where in `peek_definition` every result in opened in a separate float. 

I am uncertain whether this is the best place to handle the entire thing and open for feedback. As of opening this PR is in incomplete state as there are no tests or documentation. I did not go any further as I am not sure what would be a good way to pass multiple contexts in case the option `on_list` is set.

In the future the same approach can be applied to other methods such as `type_definition`, `declaration`, `implementation`, and the general idea of merging results from multiple servers for `hover`.

PR #30722, opened 2024-10-08
feat(lsp): support multiple clients in lsp.buf.references

mfussenegger:
Relates to:

- https://github.com/neovim/neovim/issues/17712
- https://github.com/neovim/neovim/issues/30034

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -77,6 +77,9 @@ LSP
   customizing the transformation of an LSP CompletionItem to |complete-items|.
 • |vim.lsp.diagnostic.from()| can be used to convert a list of
   |vim.Diagnostic| objects into their LSP diagnostic representation.
+• |vim.lsp.buf.references()| now handles multiple clients but no longer
+  triggers the global `textDocument/references` handler from
+  `vim.lsp.handlers`
 
 LUA
 
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -438,11 +438,59 @@ end
 ---@param opts? vim.lsp.ListOpts
 function M.references(context, opts)
   validate('context', context, 'table', true)
-  local params = util.make_position_params()
-  params.context = context or {
-    includeDeclaration = true,
-  }
-  request_with_opts(ms.textDocument_references, params, opts)
+  local clients = vim.lsp.get_clients({ method = ms.textDocument_references })
+  if not next(clients) then
+    return
+  end
+  local win = api.nvim_get_current_win()
+  local bufnr = api.nvim_get_current_buf()
+  opts = opts or {}
+
+  local all_items = {}
+  local title = 'References'
+
+  local function on_done()
+    if not next(all_items) then
+      vim.notify('No references found')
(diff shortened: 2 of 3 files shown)


runtime/doc/news.txt
```diff
@@ -75,6 +75,9 @@ LSP
   customizing the transformation of an LSP CompletionItem to |complete-items|.
 • |vim.lsp.diagnostic.from()| can be used to convert a list of
   |vim.Diagnostic| objects into their LSP diagnostic representation.
+• |vim.lsp.buf.references()| now handles multiple clients but no longer
+  triggers the global `textDocument/references` handler from
```
MariaSolOs, line 79: Why does this affect the handler? And on that note, might this change affect plugins like Telescope/fzf-lua that "wrap" these methods to display them in the respective pickers?

wookayin: I agree, removal of the handler is technically a breaking change.

It looks like it's because the design of `|lsp-handler|` assumes there is only one LSP client involved in a single LSP request (`ctx.client_id` is not plural). As this implementation tries to fire multiple LSP requests, there would be multiple response and thus multiple LSP handler invocations (called asynchronously); however, it doesn't look very straightforward how to combine and merge the results into a single qflist.

antonk52: There is a similar effort in #26507 for lsp.buf.definitions. Ideally we'd want to migrate all lsp.buf.* calls to be able to handle results from multiple servers

mfussenegger: I don't mind marking this as breaking but it's imho generally grey area because another point of view is that this would be a fix, given that currently using references with multiple clients is broken in that each client overrides the results from the previous client.

Also it's not breaking in the sense that users would get errors - at worst they'll get the stock neovim behavior, instead of one customized by a plugin. (We also went with a `on_list` argument in anticipation of this change)

https://github.com/neovim/neovim/pull/22598#issuecomment-1463367606 also explains a bit more why this is necessary.
The global-handler model is not suited for any actions where results need to be merged from multiple clients and we'll have to do this for most of the lsp.buf. methods. Hover, signature help, all the go-to variants, etc.

Plugins which currently override the handler and rely on that are similarly broken for multiple-clients.
It could be an option that we preserve that behavior if there is only one client for the buffer - but that would imho just be more confusing and drag out getting it fixed properly

justinmk:
It any extra test coverage warranted here?

przepompownia:
This change leaves me confused which offset encoding inside `on_list` option callback to use if there are more than one client.

```lua
local function onList(params)
  local client = assert(lsp.get_client_by_id(params.context.client_id)) -- fails since it became ambiguous
  if #params.items == 1 then
    lsp.util.show_document(
      params.items[1].user_data,
      client.offset_encoding, -- what to use instead?
      {reuse_win = true, focus = true}
    )
    return
  end
  vim.fn.setqflist({}, ' ', {title = 'References', items = params.items, context = params.context})
  api.nvim_command('botright copen')
end
```

geril2207:
Seems that it also no longer triggers `textDocument/definition` for `vim.lsp.buf.definition`

PR #30747, opened 2024-10-10
fix(lsp): str_byteindex_enc bounds checking

### LINKED ISSUE #30675: LSP: Semantic token highlighting error (Tinymist)
### Problem

Semantic token highlighting fails when using tinymist with error
```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:304: /usr/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:149: index
 out of range                                                                                                                                                             
stack traceback:                                                                                                                                                          
        [builtin#36]: at 0x7a4e388484e0                                                                                                                                   
        /usr/share/nvim/runtime/lua/vim/lsp/semantic_tokens.lua:304: in function 'handler'                                                                                
        /usr/share/nvim/runtime/lua/vim/lsp/client.lua:687: in function ''                                                                                                
```
The errors was introduced by the new 0.10.2 version, likely an off by one error
image
on 0.11.0-dev it has highlighting, but they ain't coming from the LSP

```lua
vim.lsp.set_log_level("debug")

local pattern = 'typst'
local cmd = { 'tinymist' }
local root_markers = { '.git', '.editorconfig', 'typst.toml' }

	pattern = pattern,
	callback = function(args)
		vim.lsp.start({
			name = 'bugged-ls',
			cmd = cmd,
			root_dir = root_dir,
			settings = settings,
		})
	end,
})
```

### Expected behavior

The text should be highlighted with the proper LSP tokens, for keywords, strings, functions and so on
image

### Nvim version (nvim -v)

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1713484068 Run "nvim -V1 -v" for more info

tinymist v0.11.28

### Operating system/version

Arch Linux x86_64

### Log file

https://gist.github.com/C0deBr3ak3r/129eb416ab046490b1a529d7a50b6142

### LINKED ISSUE #30832: LSP: Error with LSP Semantic Tokens when Zig code contains non-ASCII characters preceded by "\\"
### Problem

When opening a Zig file containing multi-line strings that start with a backslash followed by non-ASCII characters(e.g., `\\µ`), Neovim encounters an error related to semantic tokens. This issue appears to be related to the implementation of `util._str_byteindex_enc` in `runtime/lua/vim/lsp/util.lua`.

1. Ensure Neovim is set up with a Zig LSP (e.g., zls).
2. Create a new Zig file with the following content:

```zig
pub fn main() !void {
    const std = @import("std");
    const a =
        \\µs
        \\
    ;
    std.debug.print("{s}", .{a});
}
```

3. Open the file in Neovim.

### Expected behavior

The Zig code should be parsed correctly, and no error should occur when strings contain `\\` followed by non-ASCII characters. The LSP semantic tokens should be handled properly, without causing an "index out of range" error.

### Nvim version (nvim -v)

NVIM v0.10.2

zls 0.13.0

### Operating system/version

Ubuntu 16.04.6 LTS

### Log file

_No response_

tris203:
Problem:
Previously the index was only checked against the UTF8 length. This could cause unexpected behaviours for strings containing multibyte chars

Solution:
Check indicies correctly against their max value before returning the fallback length

Fixes: #30675

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -171,25 +171,24 @@ end
 ---@param encoding string utf-8|utf-16|utf-32| defaults to utf-16
 ---@return integer byte (utf-8) index of `encoding` index `index` in `line`
 function M._str_byteindex_enc(line, index, encoding)
-  local len = #line
-  if index > len then
-    -- LSP spec: if character > line length, default to the line length.
-    -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
-    return len
-  end
+  -- LSP spec: if character > line length, default to the line length.
+  -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
+  local len8 = #line
   if not encoding then
     encoding = 'utf-16'
   end
   if encoding == 'utf-8' then
-    if index then
+    if index and index <= len8 then
       return index
     else
-      return len
+      return len8
     end
-  elseif encoding == 'utf-16' then
-    return vim.str_byteindex(line, index, true)
+  end
+  local len32, len16 = vim.str_utfindex(line)
+  if encoding == 'utf-16' then
+    return index <= len16 and vim.str_byteindex(line, index, true) or len8
   elseif encoding == 'utf-32' then
-    return vim.str_byteindex(line, index)
+    return index <= len32 and vim.str_byteindex(line, index) or len8
   else
     error('Invalid encoding: ' .. vim.inspect(encoding))
   end


tris203:
```
vim.str_utfindex({str}, {index})                          *vim.str_utfindex()*
    Convert byte index to UTF-32 and UTF-16 indices. If {index} is not
    supplied, the length of the string is used. All indices are zero-based.

    Embedded NUL bytes are treated as terminating the string. Invalid UTF-8
    bytes, and embedded surrogates are counted as one code point each. An
    {index} in the middle of a UTF-8 sequence is rounded upwards to the end of
    that sequence.
```
PS: This note about embedded NUL bytes appears to not be correct

runtime/lua/vim/lsp/util.lua
```diff
@@ -171,25 +171,23 @@ end
...
 ---@return integer byte (utf-8) index of `encoding` index `index` in `line`
 function M._str_byteindex_enc(line, index, encoding)
-  local len = #line
-  if index > len then
-    -- LSP spec: if character > line length, default to the line length.
-    -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
-    return len
-  end
+  local len8 = #line
+  local len32, len16 = vim.str_utfindex(line)
```
mfussenegger, line 175: len32 and len16 are not used in the utf-8 case so this could be moved down to avoid the costs.


```diff
diff --git a/runtime/lua/vim/lsp/util.lua b/runtime/lua/vim/lsp/util.lua
index 1e47fffc40..1652ca9889 100644
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -170,11 +170,10 @@ end
 ---@param index integer UTF index
 ---@param encoding string utf-8|utf-16|utf-32| defaults to utf-16
 ---@return integer byte (utf-8) index of `encoding` index `index` in `line`
 function M._str_byteindex_enc(line, index, encoding)
   local len8 = #line
-  local len32, len16 = vim.str_utfindex(line)
   -- LSP spec: if character > line length, default to the line length.
   -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
   if not encoding then
     encoding = 'utf-16'
   end
@@ -182,11 +181,13 @@ function M._str_byteindex_enc(line, index, encoding)
     if index and index <= len8 then
       return index
     else
       return len8
     end
-  elseif encoding == 'utf-16' then
+  end
+  local len32, len16 = vim.str_utfindex(line)
+  if encoding == 'utf-16' then
     return index <= len16 and vim.str_byteindex(line, index, true) or len8
   elseif encoding == 'utf-32' then
     return index <= len32 and vim.str_byteindex(line, index) or len8
   else
     error('Invalid encoding: ' .. vim.inspect(encoding))
```

runtime/lua/vim/lsp/util.lua
```diff
@@ -171,25 +171,24 @@ end
...
 function M._str_byteindex_enc(line, index, encoding)
-  local len = #line
-  if index > len then
-    -- LSP spec: if character > line length, default to the line length.
-    -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
-    return len
-  end
+  local len8 = #line
+  -- LSP spec: if character > line length, default to the line length.
+  -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
```
justinmk, line 176:
```suggestion
  -- LSP spec: if character > line length, default to the line length.
  -- https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
  local len8 = #line
```

runtime/lua/vim/lsp/util.lua
```diff
@@ -171,25 +171,24 @@ end
 ---@param encoding string utf-8|utf-16|utf-32| defaults to utf-16
 ---@return integer byte (utf-8) index of `encoding` index `index` in `line`
 function M._str_byteindex_enc(line, index, encoding)
```
justinmk, line 173: looks like we currently don't have a direct test of this , but worth adding to `test/functional/plugin/lsp/utils_spec.lua` ?

justinmk (approved):

runtime/lua/vim/lsp/util.lua
tris203: #30735 will remove the need for this wrapper function

tris203:
@justinmk do we want to backport this?

clason:
Depends, was this a regression from switching to utf8proc?

clason:
I guess CI will tell us.

tris203:
> I guess CI will tell us. 

It was caused by me in #30288

This is the correct fix

clason:
Then we should backport indeed.

PR #30749, opened 2024-10-10
build(lpeg): lpeg.so is missed due to looking for lpeg.dylib on MacOS

trws:
build(lpeg): lpeg.so is missed due to looking for lpeg.dylib

Problem: Currently Neovim only searches for lpeg.dylib, liblpeg.dylib, etc. on MacOS, but a normal installation of lpeg will produce lpeg.so instead.  There are explicit workarounds for this by modifying lpeg's package in Homebrew, and will shortly be another in spack.

Solution: also search for lpeg.so even if that's not the platform default, because Lua and luarocks use it anyway.

### DIFF
--- a/cmake/FindLpeg.cmake
+++ b/cmake/FindLpeg.cmake
@@ -1,4 +1,13 @@
-find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} PATH_SUFFIXES lua/5.1)
+find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg.so lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} PATH_SUFFIXES lua/5.1)
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND LPEG_LIBRARY MATCHES ".so$")
+  execute_process(
+    COMMAND otool -hv "${LPEG_LIBRARY}"
+    OUTPUT_VARIABLE LPEG_HEADER
+    )
+  if(LPEG_HEADER MATCHES ".* BUNDLE .*")
+    message(FATAL_ERROR "lpeg library found at ${LPEG_LIBRARY} but built as a bundle rather than a dylib, please rebuild with `-dynamiclib` rather than `-bundle`")
+  endif()
+endif()
 
 find_package_handle_standard_args(Lpeg DEFAULT_MSG LPEG_LIBRARY)
 mark_as_advanced(LPEG_LIBRARY)


dundargoc:
Duplicate of https://github.com/neovim/neovim/pull/30627. I still would like some input on my [comment](https://github.com/neovim/neovim/pull/30627#issuecomment-2399488738) as it seems looking for `.so` won't always work if it's not built as a shared library, which may increase confusion. @carlocab do you have any input on this?

carlocab:
I suggest accepting `lpeg.so` on macOS, _but_ also checking that the LPeg library can be explicitly linked to (instead of merely `dlopen`ed) so that you can error out with an informative message at configure time if that's not the case.

trws:
We could do that. The check wouldn’t be all that difficult.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

trws:
Ok, here's a version that does the check using `otool -hv` to dump the Mach-O header, then checks to see if that contains "BUNDLE" in which case it prints an informative error message.

cmake/FindLpeg.cmake
```diff
@@ -1,4 +1,13 @@
-find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} PATH_SUFFIXES lua/5.1)
+find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg.so lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} PATH_SUFFIXES lua/5.1)
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND LPEG_LIBRARY MATCHES ".so$")
+  execute_process(
+    COMMAND otool -hv "${LPEG_LIBRARY}"
+    OUTPUT_VARIABLE LPEG_HEADER
+    )
+  if(LPEG_HEADER MATCHES ".*BUNDLE.*")
```
carlocab, line 7:
```suggestion
  if(LPEG_HEADER MATCHES ".* BUNDLE .*")
```
Maybe safer? Doesn't matter hugely though.

cmake/FindLpeg.cmake
```diff
@@ -1,4 +1,13 @@
...
+    message(FATAL_ERROR "lpeg library found at ${LPEG_LIBRARY} but built as a bundle rather than a dylib, please rebuild with `-shared -fPIC` rather than `-bundle`")
```
carlocab, line 8:
```suggestion
    message(FATAL_ERROR "lpeg library found at ${LPEG_LIBRARY} but built as a bundle rather than a dylib, please rebuild with `-dynamiclib` rather than `-bundle`")
```
This is slightly more idiomatic on macOS. (`-fPIC` is a no-op on modern macOS -- all code is position-independent.)

carlocab (commented):
Looks fine to me, but a few suggestions! (I don't have commit access here, though, so @dundargoc will likely have to ✅ this)

cmake/FindLpeg.cmake
trws: Fair point, I'll take both of these.

dundargoc (approved):

casr:
This commit is preventing the build from working in MacPorts due to:

```
CMake Error at cmake/FindLpeg.cmake:8 (message):
  lpeg library found at /opt/local/lib/lua/5.1/lpeg.so but built as a bundle
  rather than a dylib, please rebuild with `-dynamiclib` rather than
  `-bundle`
```

Two libs are installed in the MacPorts' prefix of `/opt/local`:

```
/opt/local/lib/lua/5.1/lpeg.dylib
/opt/local/lib/lua/5.1/lpeg.so
```

The output of `otool` for these two files is:

```
$ otool -hv "/opt/local/lib/lua/5.1/lpeg.dylib"
/opt/local/lib/lua/5.1/lpeg.dylib:
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64    ARM64        ALL  0x00       DYLIB    15       1400   NOUNDEFS DYLDLINK TWOLEVEL NO_REEXPORTED_DYLIBS

$ otool -hv "/opt/local/lib/lua/5.1/lpeg.so"
/opt/local/lib/lua/5.1/lpeg.so:
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64    ARM64        ALL  0x00      BUNDLE    14       1336   NOUNDEFS DYLDLINK TWOLEVEL
```

My question is would changing the order of the `find_library2` paths be sufficient to correct this so that searching for `lpeg.so` would come after searching for `lpeg${CMAKE_SHARED_LIBRARY_SUFFIX}`?

```diff
-find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg.so lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} PATH_SUFFIXES lua/5.1)
+find_library2(LPEG_LIBRARY NAMES lpeg_a lpeg liblpeg_a lpeg${CMAKE_SHARED_LIBRARY_SUFFIX} lpeg.so PATH_SUFFIXES lua/5.1)
```

dundargoc:
> My question is would changing the order of the find_library2 paths be sufficient to correct

Theoretically no, but probably in practice probably yes (it's undefined, so it can be different for each platform). The proper way to enforce order is to just make two calls to `find_library` with the first call being the higher priority.

PR #30751, opened 2024-10-11
test(tui_spec): better prevent race between paste and input

zeertzjq:
Problem:
Using a single RPC request to child server isn't enough to prevent race
between nvim_paste and nvim_input.

Solution:
Ensure both child client and child server have processed pending events
by sending an empty DCS response to the child client and waiting for
TermResponse autocommand on the child server.


justinmk:
For "waiting" on nvim_input to be processed, what do you think about https://github.com/neovim/neovim/issues/30760 ?

PR #30753, opened 2024-10-11
fix(lua): avoid recursive vim.on_key() callback

### LINKED ISSUE #30752: Using `vim.on_key()` stuck at `Press ENTER or...`.
### Problem

The only way out is to `kill -INT <pid>`. Other kills work as well.

### Steps to reproduce

Using `nvim --clean`, edit and source the following, get the following for any keystroke.
```
Press ENTER or type command to continue                                                             
Press ENTER or type command to continue                                                             
...
```

```
local ns_id
local function key_cb(key, typed)
    vim.print("key_cb")
    vim.print(string.format("KEYCB: key '%s', typed '%s'", key, typed))
end

ns_id = vim.on_key(key_cb)


key_cb("foo", "bar")

-- The defered function doesn't work. Guess it's never executed.
--vim.defer_fn(function() vim.on_key(nil, ns_id) end, 9000)
```

### Expected behavior

Expect `Ctrl-C` to get out of loop.

### Nvim version (nvim -v)

https://github.com/neovim/neovim/commit/61f1b091ea97793f9b644cebf6c84cf6bbb4f0bc

Can't reproduce in vim

### Operating system/version

ubuntu

### Terminal name/version

gnone

xterm-256color

### Installation

make install

zeertzjq:
Fix #30752

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -1678,6 +1678,8 @@ vim.on_key({fn}, {ns_id})                                       *vim.on_key()*
 
     Note: ~
       • {fn} will be removed on error.
+      • {fn} won't be invoked recursively, i.e. if {fn} itself consumes input,
+        it won't be invoked for those keys.
       • {fn} will not be cleared by |nvim_buf_clear_namespace()|
 
     Parameters: ~
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -258,6 +258,9 @@ These existing features changed their behavior.
   more emoji characters than before, including those encoded with multiple
   emoji codepoints combined with ZWJ (zero width joiner) codepoints.
 
+• |vim.on_key()| callbacks won't be invoked recursively when a callback itself
+  consumes input.
+
 ==============================================================================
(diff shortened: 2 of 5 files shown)


justinmk (approved):

errael:
Tried this. :+1:

After first reading `avoid recursive vim.on_key()`, I thought some calls would be lost. Oh wait, some are lost; curiously doing `j` with map `j -> gj` nothing's lost is a simple test but `type` not as expected (as mentioned elsewhere) . 

Should be using a scheduled function, or other alternative, anyway. If deferred, need  to make sure all debug output is deferred so that order is preserved.

errael:
Opened #30770 which is loosely related.

zeertzjq:
"recursive callback" only happens when the callback itself consumes input, which isn't common. Maybe I should make it clearer in the docs.

errael:
> "recursive callback" only happens when the callback itself consumes input, which isn't common. Maybe I should make it clearer in the docs.

Don't know, seems difficult to make perfectly clear. In this case, I guess the recursion is because the `Press enter` is indirectly called through `on_key` and the `<CR>` key is queued...

PR #30764, opened 2024-10-11
feat(defaults): map gri to vim.lsp.buf.implementation()

gpanders:
Continuing the default LSP maps under the "gr" prefix. Mnemonic: "i" for "implementation".

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -71,11 +71,12 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                                *grr* *gra* *grn* *i_CTRL-S*
+                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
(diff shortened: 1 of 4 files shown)


lewis6991 (approved):

justinmk (approved):

ofseed:
Sorry I missed the discussion on this RP, but I think it would be better to map `vim.lsp.buf.implementation` to another key? I would like to reserve `gri` and `gro` for `vim.lsp.buf.incoming_calls` and `vim.lsp.buf.outgoing_calls`, respectively. The appropriateness of this method lies not only in `i` for `incoming` and `o` for `outgoing`, but vim also uses it for pairs of mappings that can complement each other, like `gi` and `go`. I remember the previous idea was to use `gI`?

justinmk:
> reserve `gri` and `gro` for `vim.lsp.buf.incoming_calls` and `vim.lsp.buf.outgoing_calls`, respectively. ... for pairs of mappings that can complement each other

That's a good callout. I would guess navigating to an implementation is far more common than listing incoming/outgoing calls, so perhaps `grI` / `grO` are appropriate there.

PR #30772, opened 2024-10-11
docs(api): adjust `nvim_echo` comment to make `api.txt` more understandable

errael:
The comment gets confusing (at least to me) when translated to docs.
```
      • {opts}     Optional parameters.
                   • verbose: Message was printed as a result of 'verbose'
                     option if Nvim was invoked with -V3log_file, the message
```
But I may be misunderstanding...

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -662,8 +662,8 @@ nvim_echo({chunks}, {history}, {opts})                           *nvim_echo()*
                    be omitted for no highlight.
       • {history}  if true, add to |message-history|.
       • {opts}     Optional parameters.
-                   • verbose: Message was printed as a result of 'verbose'
-                     option if Nvim was invoked with -V3log_file, the message
+                   • verbose: Message is printed as a result of 'verbose'
+                     option. If Nvim was invoked with -V3log_file, the message
                      will be redirected to the log_file and suppressed from
                      direct output.
 
--- a/runtime/lua/vim/_meta/api.lua
+++ b/runtime/lua/vim/_meta/api.lua
@@ -1110,8 +1110,8 @@ function vim.api.nvim_del_var(name) end
 --- can be omitted for no highlight.
 --- @param history boolean if true, add to `message-history`.
 --- @param opts vim.api.keyset.echo_opts Optional parameters.
---- - verbose: Message was printed as a result of 'verbose' option
----   if Nvim was invoked with -V3log_file, the message will be
+--- - verbose: Message is printed as a result of 'verbose' option.
+---   If Nvim was invoked with -V3log_file, the message will be
 ---   redirected to the log_file and suppressed from direct output.
 function vim.api.nvim_echo(chunks, history, opts) end
 
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -781,8 +781,8 @@ void nvim_set_vvar(String name, Object value, Error *err)
 ///                can be omitted for no highlight.
 /// @param history  if true, add to |message-history|.
 /// @param opts  Optional parameters.
-///          - verbose: Message was printed as a result of 'verbose' option
-///            if Nvim was invoked with -V3log_file, the message will be
+///          - verbose: Message is printed as a result of 'verbose' option.
+///            If Nvim was invoked with -V3log_file, the message will be
 ///            redirected to the log_file and suppressed from direct output.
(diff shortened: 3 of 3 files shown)


zeertzjq:
I'm not sure if this is correct, and I can't understand this flag either. @bfredl?

errael:
> I'm not sure if this is correct ...

This PR is also about dipping my toe in the water. The `docs/docs` test fails, and suggests doing `make doc`. But...
```

FAILED: touches/doc-vim /src/tools/neovim/neovim/build/touches/doc-vim 
cd /src/tools/neovim/neovim && /usr/bin/cmake -E touch /src/tools/neovim/neovim/build/touches/doc-vim && /usr/bin/cmake -E env VIMRUNTIME=/src/tools/neovim/neovim/runtime /src/tools/neovim/neovim/build/bin/nvim -u NONE -l scripts/gen_vimdoc.lua
Target: api.txt
E5113: Error while calling lua chunk: scripts/gen_vimdoc.lua:971: assertion failed!
stack traceback:
	[C]: in function 'assert'
	scripts/gen_vimdoc.lua:971: in function 'gen_target'
	scripts/gen_vimdoc.lua:1026: in function 'run'
	scripts/gen_vimdoc.lua:1030: in main chunk
```

bfredl:
Maybe it is more understandable if we don't cut of the last sentence halfway:

>         if Nvim was invoked with -V3log_file, the message will be
>         redirected to the log_file and suppressed from direct output.

And it only makes sense to do this if the message is printed for a sufficiently high 'verbose' level. Maybe the order of the two sentences should be flipped.

errael:
> Maybe it is more understandable if we don't cut of the last sentence halfway:
> 
> > ```
> >     if Nvim was invoked with -V3log_file, the message will be
> >     redirected to the log_file and suppressed from direct output.
> > ```

Other than the "was" to "is", this PR only terminated the first sentence properly. But you're right should have had some "..." a few words into the 2nd sentence. I kind of understood the last sentence.

> And it only makes sense to do this if the message is printed for a sufficiently high 'verbose' level. 

I don't get that.

After some playing around, starting up with
```
nvim -Vlog_file ...
```
and doing

```lua
vim.api.nvim_echo({{"message\n"}}, false, {verbose = true})
```

is very handy. While most of the stuff in `log_file` goes out buffered. The text from `nvim_echo` is **unbuffered**, (hooray) and works well with `tail -f log_file` while debugging.

> Maybe the order of the two sentences should be flipped.

That might be more clear.

If someone want to provide updated text this PR can be used as a vehicle.

Or I can close this, which might be best considering `make doc` doesn't work for me.

wookayin:
The change looks good, but please try to make docgen `make doc` come through. Actually in my machine gen_vimdoc works fine,

The assertion error at Line 971 could be when the `ext` is neither `lua`, `c`, nor `h`. You can check what's wrong there:
```diff
-   local parser = assert(parsers[ext])
+   local parser = assert(parsers[ext], vim.inspect({f = f, ext = ext})
```

Yes, it's something gen_vimdoc.lua could (maybe) improve in error messages sometime in the future.

errael:
> The change looks good, but please try to make docgen `make doc` come through. Actually in my machine gen_vimdoc works fine,
> 
> The assertion error at Line 971 could be when the `ext` is neither `lua`, `c`, nor `h`. You can check what's wrong there:
> 
> ```diff
> -   local parser = assert(parsers[ext])
> +   local parser = assert(parsers[ext], vim.inspect({f = f, ext = ext})
> ```
> 
> (Maybe it's something gen_vimdoc.lua could (maybe) improve in error messages sometime in the future...)

Thanks @wookayin. The problem was
```
E5113: Error while calling lua chunk: scripts/gen_vimdoc.lua:971: {
  ext = "c~",
  f = "src/nvim/api/vim.c~"
}
```
Should both
```
runtime/doc/api.txt
runtime/lua/vim/_meta/api.lua
```
be added to the PR?

BTW, there was discussion about improving`nvim_echo` doc in general; but, considering the comment at https://github.com/neovim/neovim/discussions/30770#discussioncomment-10923308, I guess that can happen when the function is reconsidered.

justinmk:
> Should both
> 
> ```
> runtime/doc/api.txt
> runtime/lua/vim/_meta/api.lua
> ```
> 
> be added to the PR?

yes, those are generated files

justinmk (approved):

errael:
ping (I was cleaning up branches and noticed this is still around) Something I need to do?

dundargoc:

PR #30773, opened 2024-10-11
fix(coverity/497375): f_strpart cast overflow

devon-g:
Problem:
Casting long to int introduces risk of overflow.

Solution:
Work with all longs rather than casting between int and long.


src/nvim/strings.c
```diff
@@ -2838,10 +2838,10 @@ void f_strpart(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
   }
 
   if (argvars[2].v_type != VAR_UNKNOWN && argvars[3].v_type != VAR_UNKNOWN) {
-    int off;
+    long off;
```
zeertzjq, line 2841: Please use `int64_t` instead of `long`

devon-g: Yes sir o7

zeertzjq (approved):

PR #30781, opened 2024-10-12
feat(defaults): map gO to document symbol

nik-rev:
Adds 2 new LSP mappings under gr prefix.

gO: vim.lsp.buf.document_symbol()

mnemonic: "s" for "symbol"

gO was planned to show document symbols

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -71,7 +71,7 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                              *grr* *gra* *grn* *gri* *grs* *grS* *i_CTRL-S*
+                              *grr* *gra* *grn* *gri* *grs* *gO* *i_CTRL-S*
 Some keymaps are created unconditionally when Nvim starts:
 - "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -71,7 +71,7 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                   *grr* *gra* *grn* *gri* *i_CTRL-S*
+                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
 Some keymaps are created unconditionally when Nvim starts:
 - "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -71,12 +71,14 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
+                              *grr* *gra* *grn* *gri* *grs* *grS* *i_CTRL-S*
 Some keymaps are created unconditionally when Nvim starts:
 - "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
 - "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
 - "gri" is mapped in Normal mode to |vim.lsp.buf.implementation()|
+- "grs" is mapped in Normal mode to |vim.lsp.buf.workspace_symbol()|
+- "grS" is mapped in Normal mode to |vim.lsp.buf.document_symbol()|
 - CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
 
 If not wanted, these keymaps can be removed at any time using
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -77,8 +77,8 @@ Some keymaps are created unconditionally when Nvim starts:
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
 - "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
 - "gri" is mapped in Normal mode to |vim.lsp.buf.implementation()|
-- "grs" is mapped in Normal mode to |vim.lsp.buf.workspace_symbol()|
-- "grS" is mapped in Normal mode to |vim.lsp.buf.document_symbol()|
+- "grS" is mapped in Normal mode to |vim.lsp.buf.workspace_symbol()|
+- "grs" is mapped in Normal mode to |vim.lsp.buf.document_symbol()|
 - CTRL-S is mapped in Insert mode to |vim.lsp.buf.signature_help()|
 
 If not wanted, these keymaps can be removed at any time using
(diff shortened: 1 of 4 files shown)


max397574:
Imo the other way around would make more sense
since workspace is a bigger scope than document

nikitarevenco:
> Imo the other way around would make more sense since workspace is a bigger scope than document

I can see that, it does makes more sense semantically. However, what about in practice?

`grs` takes 3 keystrokes, whereas `grS` takes 4 and a bit of stretching.
So I think it would make sense to use `grs` for the workspace, because from my personal experience it's a lot more common for me to search by symbols in workspace rather than in document.

justinmk (changes requested):
`gO` should show current document symbols, that was always the plan for `gO`.

nikitarevenco:
> `gO` should show current document symbols, that was always the plan for `gO`.

Ok, that makes sense actually. I have changed it so that `grs` is workspace symbols, `gO` is document symbols

runtime/doc/lsp.txt
```diff
@@ -71,12 +71,14 @@ options are not restored when the LSP client is stopped or detached.
...
   a custom keymap for `K` exists.
 
-                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
+                              *grr* *gra* *grn* *gri* *grs* *grS* *i_CTRL-S*
 Some keymaps are created unconditionally when Nvim starts:
 - "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
 - "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
 - "gri" is mapped in Normal mode to |vim.lsp.buf.implementation()|
+- "grs" is mapped in Normal mode to |vim.lsp.buf.workspace_symbol()|
```
justinmk, line 80: would this be a good use of `gr/` or is that conceivably useful for something else (e.g. some sort of LSP-driven file navigation)?

runtime/lua/vim/_defaults.lua
```diff
@@ -157,7 +157,7 @@ do
   --- client is attached. If no client is attached, or if a server does not support a capability, an
   --- error message is displayed rather than exhibiting different behavior.
   ---
-  --- See |grr|, |grn|, |gra|, |gri|, |i_CTRL-S|.
+  --- See |grr|, |grn|, |gra|, |gri|, |grs|, |grS|, |i_CTRL-S|.
```
justinmk, line 160:
```suggestion
  --- See |grr|, |grn|, |gra|, |gri|, |grs|, |gO|, |i_CTRL-S|.
```

runtime/doc/lsp.txt
nikitarevenco: > would this be a good use of `gr/` or is that conceivably useful for something else (e.g. some sort of LSP-driven file navigation)?

I think "s" for symbol will make it easier to remember, "/" is associated with searching for something. `vim.lsp.buf.list_workspace_folders` seems like it has to do with some sort of navigation, (i haven't used it though). So maybe reserve `/` for that?

justinmk: >  "/" is associated with searching for something.

So we should reserve `gr/` for builtin (or delegated) "freceny"/fuzzy-finder (https://github.com/neovim/neovim/issues/660).

nikitarevenco: Yes, solid idea

mfussenegger:
workspace symbol search has overlap with `:tjump`. Given that you typically use it with a query I'm not sure if a dedicated keymap is warranted?

nikitarevenco:
> workspace symbol search has overlap with `:tjump`. Given that you typically use it with a query I'm not sure if a dedicated keymap is warranted?

I still feel like having a keymap for it would be useful due to how common this action is used.

I didn't find anything for `:h grs` so I don't understand how it overlaps. If it does and I am missing something, then we maybe can use `grS`?

mfussenegger:
Sorry I phrased that poorly. What I meant to say is that `:tselect`/`:tjump` also use workspace symbols (via the lsp tagfunc) and typing `:ts <query>` has about the same number of keystrokes as would typing `grs <query>` has.

The UI is currently different, but I think we should improve that instead of offering more ways to do the same thing.

justinmk:
so let's move forward with gO but drop the other one for now.

nikitarevenco:
Ok, so I can see where you are coming from. I've removed grs

runtime/doc/lsp.txt
```diff
@@ -71,12 +71,14 @@ options are not restored when the LSP client is stopped or detached.
...
   a custom keymap for `K` exists.
 
-                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
+                                 *grr* *gra* *grn* *gri* *gO* *i_CTRL-S*
 Some keymaps are created unconditionally when Nvim starts:
 - "grn" is mapped in Normal mode to |vim.lsp.buf.rename()|
 - "gra" is mapped in Normal and Visual mode to |vim.lsp.buf.code_action()|
 - "grr" is mapped in Normal mode to |vim.lsp.buf.references()|
 - "gri" is mapped in Normal mode to |vim.lsp.buf.implementation()|
+- "grs" is mapped in Normal mode to |vim.lsp.buf.workspace_symbol()|
```
justinmk, line 80: This line needs to be removed. Please check your PRs so that reviewers don't have to spend time on this.

runtime/doc/lsp.txt
```diff
@@ -71,12 +71,14 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
+                                 *grr* *gra* *grn* *gri* *gO* *i_CTRL-S*
```
justinmk, line 74: The `*gO*` tag should be removed here. Keep the gO mention  below, but also mention the LSP default at the existing `*gO*` tag.

And then check that CI passes, and fixe any problems if not.

runtime/doc/lsp.txt
nikitarevenco: Sorry, I'll make sure to up my communication for the next time. 

I only have access to a phone for the next few days, but it's easier for me to make sure the PR is ready for review when I have access to my computer 

I will let you know when I do (I know its just 1 line change, but there could be some other problem I might not notice, so I don't want to waste your time)

ofseed:
I'm not particularly sure that the current status of `vim.lsp.buf.document_symbols` is sufficient to be the default; I've at least found two potential issues with it.

1. `gO` is used in `help` and `:Man` to display the outline, which is the meaning of `O`. However, our current `vim.lsp.buf.document_symbols` ignores the hierarchy and directly returns all symbols. Ideally, it should only return the first or second level of scope.
2. Even if we use it in a way that maps to `vim.lsp.buf.references`, achieving something similar to `grr` but displaying all symbols, there are still shortcomings. This is because the current implementation still ignores the scope of the symbols. An improvement has been made for `vim.lsp.buf.references` have not been made for this yet, see https://github.com/neovim/neovim/pull/29164.

I think it’s best to find solutions to the above two issues before promoting it to the default mapping.

justinmk:
> `gO` is used in `help` and `:Man` to display the outline, which is the meaning of `O`. However, our current `vim.lsp.buf.document_symbols` ignores the hierarchy and directly returns all symbols. Ideally, it should only return the first or second level of scope.

It could be improved, but I don't see why that would block this. We have `document_symbols` which does something useful, and is the better approximation of a "document outline". We can always update the mapping if we get something better in the future.

runtime/doc/lsp.txt
nikitarevenco: hey, can you please elaborate what you mean by:

> but also mention the LSP default at the existing *gO* tag.

Wouldn't this line already do that:

```
 "gO" is mapped in Normal mode to |vim.lsp.buf.document_symbol()|
```

I removed the `*gO*` tag

nikitarevenco: I'm confused why removing the tag was necessary though

runtime/doc/lsp.txt
```diff
@@ -71,12 +71,13 @@ options are not restored when the LSP client is stopped or detached.
 - |K| is mapped to |vim.lsp.buf.hover()| unless |'keywordprg'| is customized or
   a custom keymap for `K` exists.
 
-                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
+                                   *grr* *gra* *grn* *gri* *i_CTRL-S*
```
justinmk, line 74:
```suggestion
                                          *grr* *gra* *grn* *gri* *i_CTRL-S*
```

justinmk:
I will update the docs for `:help gO` in another PR.

PR #30777, opened 2024-10-12
feat(runtime): cleanup one-off scripts

clason:
Problem: Some runtime files no longer spark joy.

Solution: Kondo that folder.

Still sparks joy (moved to new `runtime/scripts` folder):
* `macros/less.*`
* `mswin.vim`
* `tools/emoji_list.lua`

No longer sparks joy (removed):
* `macmap.vim` (gvimrc file; not useful in Nvim)
* `tools/check_colors.vim` (no longer useful with new default colorscheme and treesitter)
* `macros/editexisting.vim` (throws error on current Nvim)
* `macros/justify.vim` (obsolete shim for `packadd! justify`)
* `macros/matchit.vim` (same)
* `macros/shellmenu.vim` (same)
* `macros/swapmous.vim` (same)

Closes #22980 

---

Separate commits for reviewing purposes; will squash and add `news.txt`/`:h vim-differences` when done.

PR #30791, opened 2024-10-13
docs: fix path to check_colors.vim

whitebyte:
The file was moved in 567c0e6cd706fd814d081fae49bcc20a83dbc8b1 but docs weren't updated


clason:
Thanks, but the file will actually be removed: #30777

PR #30793, opened 2024-10-13
fix(tui): avoid flushing buffer halfway an OSC 2 sequence

zeertzjq:
Problem:  Setting title while TUI buffer is almost full may cause the
          end of a flush to be treated as a part of an OSC 2 or OSC 0
          sequence, leading to problems like invisible cursor.
Solution: Make the whole sequence to set title a unibi_ext string.

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -134,11 +134,12 @@ struct TUIData {
     int resize_screen;
     int reset_scroll_region;
     int set_cursor_style, reset_cursor_style;
-    int save_title, restore_title;
+    int save_title, restore_title, set_title;
     int set_underline_style;
     int set_underline_color;
     int sync;
   } unibi_ext;
+  char *set_title;
   char *space_buf;
   size_t space_buf_len;
   bool stopped;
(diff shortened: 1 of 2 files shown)


src/nvim/tui/tui.c
```diff
@@ -1807,8 +1808,10 @@ retry: \
       unibi_format(vars, vars + 26, str, params, out, tui, pad, tui); \
       if (tui->overflow) { \
         tui->bufpos = orig_pos; \
-        flush_buf(tui); \
-        goto retry; \
+        if (orig_pos > 0) { \
```
zeertzjq, line 1811: If `orig_pos` is 0, `flush_buf(tui)` won't do anything and `goto retry;` will lead to an infinite loop. This situation can't happen when setting title as the maximum title length is much smaller than the TUI buffer size, but it can happen with a very long URL which has no size limit. #30794

justinmk: That seems worth a 1-line comment. Or more than 1 line, all of that is useful insight.

```suggestion
        /* If orig_pos is 0, `flush_buf` won't do anything and `goto retry` will loop. */ \
        if (orig_pos > 0) { \
```

justinmk (approved):

PR #30795, opened 2024-10-13
Update vim.lua to interpret `ctermfg=3` consistently 

nisbet-hubbard:
Currently, colour no. 3 in the 8-colour scheme is interpreted as DarkYellow
https://github.com/neovim/neovim/blob/fb74fd2954ee4eeee566bb2fcfd480a4281f1e4a/runtime/syntax/dircolors.vim#L52

Thus, `hi Statement term=bold ctermfg=3` in Vim’s old default colorscheme for light background (ie peachpuff-legacy) is shown as bold DarkYellow in Vim/Neovim: https://github.com/mohvn/peachpuff-legacy/blob/eee3101fd33d58350274a16e60df7452080d2c56/colors/peachpuff.vim#L54

The same colour, however, is interpreted as Brown in vim.lua because `3	    Brown, DarkYellow`. As a result, although vim.lua is supposed to revert to the old Vim defaults, Statement actually sports a different colour under `:colo vim` (non-bold Brown) than with `:colo peachpuff-legacy` (bold DarkYellow) in a terminal context.

This PR fixes the inconsistency.

### DIFF
--- a/runtime/colors/vim.lua
+++ b/runtime/colors/vim.lua
@@ -245,7 +245,7 @@ if vim.o.background == 'light' then
   hi('Constant',         { fg = 'Magenta',                              ctermfg = 'DarkRed' })
   hi('Special',          { fg = '#6a5acd',                              ctermfg = 'DarkMagenta' })
   hi('Identifier',       { fg = 'DarkCyan',                             ctermfg = 'DarkCyan' })
-  hi('Statement',        { fg = 'Brown', bold = true,                   ctermfg = 'Brown' })
+  hi('Statement',        { fg = 'DarkYellow', bold = true,              ctermfg = 'DarkYellow', cterm = { bold = true } })
   hi('PreProc',          { fg = '#6a0dad',                              ctermfg = 'DarkMagenta' })
   hi('Type',             { fg = 'SeaGreen', bold = true,                ctermfg = 'DarkGreen' })
   hi('Underlined',       { fg = 'SlateBlue', underline = true,          ctermfg = 'DarkMagenta', cterm = { underline = true } })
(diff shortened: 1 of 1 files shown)


zeertzjq:
Where does `ctermfg=3` come from? Vim doesn't use `ctermfg=3` for `Statement` in its default colorscheme.

echasnovski (commented):
As far as I can tell, Vim [uses 'syntax/syncolor.vim'](https://github.com/vim/vim/blob/5dcee3c723b817fcd14c9abc7435e5cc4c0fc862/src/highlight.c#L554) when `vim --clean` is done.

That in turn has [plain (non bold) `Brown`](https://github.com/vim/vim/blob/5dcee3c723b817fcd14c9abc7435e5cc4c0fc862/runtime/syntax/syncolor.vim#L53) color for `Statement` with light background and 'notermguicolors' (the `cterm=NONE ctermfg=Brown ctermbg=NONE` part). @zeertzjq, do these conclusions about how Vim works look correct?

So the current 'colors/vim.lua' definitions look correct to me as they should be the same as the default Vim color scheme (i.e. what is shown in `vim --clean`).

zeertzjq:
Why should the color from 8 colors be used here, instead of the one from 256 colors?

zeertzjq:
> So, yeah, the root cause is that with `--clean` Vim defaults to `t_Co=8`.

No it doesn't. It only defaults to `t_Co=8` under specific condition. Therefore I'm going to close this PR.

echasnovski:
One thing I'd also like to clarify is that `:colors vim` was intended as a more or less frozen fallback to the state of default color scheme before it was changed in 0.10.0.

However, following its `vim` name, it is also useful to keep it up to date to upstream `vim --clean` *as much as reasonably possible*. For example, accounting for new highlight groups (dc4037f612) or adjusting some existing ones following upstream change (31b3c62845) is reasonable, while accounting for all values of `t_Co` does not look like it.

PR #30796, opened 2024-10-13
docs: update autotrigger description of vim.lsp.compleiton.BufferOpts

glepnir:
Currently the behavior of autotrigger is not well documented.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -596,7 +596,7 @@ local function on_complete_done()
 end
 
 --- @class vim.lsp.completion.BufferOpts
---- @field autotrigger? boolean Whether to trigger completion automatically. Default: false
+--- @field autotrigger? boolean  Default: false When true, completion triggers automatically based on the server's `triggerCharacters`.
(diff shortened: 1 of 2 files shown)


glepnir:
maybe need open a new docs: misc collection PR.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -596,7 +596,7 @@ local function on_complete_done()
 end
 
 --- @class vim.lsp.completion.BufferOpts
---- @field autotrigger? boolean Whether to trigger completion automatically. Default: false
+--- @field autotrigger? boolean  Default: false When true completion will only be triggered automatically if |v:char| is contained within the client's `triggerCharacters`.
```
mfussenegger, line 599:
```suggestion
--- @field autotrigger? boolean  Default: false When true, completion triggers automatically based on the server's `triggerCharacters`.
```

Just a suggestion.

- Removes the "only" as it somehow implies that there's a way to have it auto-triggered in more cases.
- Uses `server` instead of `client` to make it clearer that the trigger characters are defined by the server. (Not sure about this, as technically we save the servers definition on the client - so this could be confusing?)

glepnir: let's go

PR #30806, opened 2024-10-14
build: remove enforcement of cmake build type

dundargoc:
Forcing the build type to be one of the default types was intended to
prevent usage of custom build types which risks unpredictability.
Unfortunately, many distributions patch this check out so they can use
a custom type anyway, making the check moot in the first place. Instead,
allow custom build types but don't try to accomodate them in the code:
we're making the assumption that anyone who wants to use a custom build
type will be able to set any additional necessary flags needed for their
specific system.

### DIFF
--- a/cmake/Util.cmake
+++ b/cmake/Util.cmake
@@ -175,8 +175,7 @@ function(add_target)
   add_custom_target(${target} DEPENDS ${touch_file})
 endfunction()
 
-# Set default build type to BUILD_TYPE. Also limit the list of allowable build
-# types to the ones defined in variable allowableBuildTypes.
+# Set default build type to BUILD_TYPE.
 #
 # The correct way to specify build type (for example Release) for
 # single-configuration generators (Make and Ninja) is to run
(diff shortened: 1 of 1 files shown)


cmake/Util.cmake
```diff
@@ -193,28 +192,26 @@ endfunction()
 # Passing CMAKE_BUILD_TYPE for multi-config generators will not only not be
 # used, but also generate a warning for the user.
 function(set_default_buildtype BUILD_TYPE)
-  set(allowableBuildTypes Debug Release MinSizeRel RelWithDebInfo)
-  if(NOT BUILD_TYPE IN_LIST allowableBuildTypes)
+  set(defaultBuildTypes Debug Release MinSizeRel RelWithDebInfo)
+  if(NOT BUILD_TYPE IN_LIST defaultBuildTypes)
     message(FATAL_ERROR "Invalid build type: ${BUILD_TYPE}")
```
jamessan, line 197: Won't this still fail?

dundargoc: Nah, this is only when calling the function eith a non-default build type. Tbh unsure if this is still needed though. We can probably remkve this

jamessan: Isn't `None` a non-default build type? :) Or am I missing something?

jamessan: Ah, I was. This is about the default build type (that the project opts in to) not the specific build type the user is selecting.

dundargoc: Yeah, that's what I tried to say lol. I don't think this is needed, so will remove this as well.

jamessan (approved):

PR #30813, opened 2024-10-14
docs: update msgpack specification reference to master

d-itu:
Updated the msgpack specification reference from commit 0b8f5ac to the latest master branch. This change resolves formatting issues in the previous reference and ensures consistency with the msgpack-rpc specification reference.


dundargoc:
Hold up. Doesn't it need to be a commit hash since the master ref changes?

clason:
Also, if there were changes, we need to make sure that we still conform with the new version. I always took this as a historical record ("this is what we implemented").

justinmk:
yeah, the current URL points to the spec that we satisfy. If Nvim satisfies a later version of the spec we can update to that (or a tag, ideally), but it is never appropriate to link to a branch.

Issue #25259, opened 2023-09-20
feat: Add a callback option to `vim.lsp.buf.code_action()`


luckasRanarison:
### Problem

It's currently impossible to get information about the selected code action using `vim.lsp.buf.code_action`.

### Expected behavior

Why would that be useful? 
It would allow us to play around with code actions, for example: get the last code action context and apply the same code action to all diagnostics with the same context in the current workspace, or get the last code action edits and reuse it by implementing it as a dot repeat.

justinmk:
As discussed in https://github.com/neovim/neovim/pull/25260, we want a generic solution for this, not special-purpose callbacks sprinkled in numerous places.

- https://github.com/neovim/neovim/pull/22598
- https://github.com/neovim/neovim/pull/22598#issuecomment-1463367606

luckasRanarison:
I agree, having a global hook system would be better. Looking forward to it.

PR #30830, opened 2024-10-16
feat(lsp): show server name in code actions

### LINKED ISSUE #30710: LSP: vim.lsp.buf.code_action() option to show name of the language server providing the action
### Problem

I checked the docs and searched online, but as far as I know, `vim.lsp.buf.code_action()` does not currently offer a way to customize the actions shown in the menu. Allowing customization could be useful in cases where there are many available actions or actions with similar wordings, and adding the name of the language server that is offering the action can help select the correct option.

Sorry in advance if this feature already exists but I was not able to find anything related to it.

### Expected behavior

Allow a way to customize `vim.lsp.buf.code_action()` to show the name of the language servers

JordanllHarper:
Problem:

Outlined in #30710, currently users need to implement their own solution for showing the language server in the code actions list.

Solution:

This fixes #30710 by adding the language server name next to each
action.

Edit: Update to solution block message for clarity

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -182,6 +182,8 @@ LSP
 • |vim.lsp.buf.format()| now supports passing a list of ranges
   via the `range` parameter (this requires support for the
   `textDocument/rangesFormatting` request).
+• |vim.lsp.buf.code_action()| actions show client name when there are multiple
+  clients.
 
 LUA
 
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -827,11 +827,19 @@ local function on_code_action_results(results, opts)
     return
   end
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx: lsp.HandlerContext}
   local function format_item(item)
-    local title = item.action.title:gsub('\r\n', '\\r\\n')
-    return title:gsub('\n', '\\n')
+    local clients = vim.lsp.get_clients({ bufnr = item.ctx.bufnr })
+    local title = item.action.title:gsub('\r\n', '\\r\\n'):gsub('\n', '\\n')
+
+    if #clients == 1 then
+      return title
+    end
+
+    local source = vim.lsp.get_client_by_id(item.ctx.client_id).name
+    return ('%s [%s]'):format(title, source)
   end
+
   local select_opts = {
     prompt = 'Code actions:',
     kind = 'codeaction',


justinmk:
Why is this an option? When would it be unacceptable to show this?

We can and will try to give full control over presentation through other mechanisms (besides the existing mechanism of overriding the handler). But these minor options meanwhile should be avoided.

JordanllHarper:
First PR so just learning the development and design process here. Will fix that formatting issue.

Thought process behind this was the information overload when initiating code actions when it came to presentation. 

If we don't want this optional, I'm happy to change it. However I'm wondering how the user might customise this behaviour if they don't want to see the server name, given that the action text and the server text are joined? Would this require a user to fully override `vim.ui.select`to remove this?

justinmk:
> information overload initiating code actions

a screenshot would help

JordanllHarper:
> A screenshot would help

Here you are! I actually think I'm changing my mind now I'm looking at this.

image

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +830,16 @@ local function on_code_action_results(results, opts)
...
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx : lsp.HandlerContext}
   local function format_item(item)
-    local title = item.action.title:gsub('\r\n', '\\r\\n')
-    return title:gsub('\n', '\\n')
+    local title = item.action.title
+    if opts and opts.show_server then
+      local source = vim.lsp.get_client_by_id(item.ctx.client_id).name or "unknown"
+      title = title .. "[" .. source .. "]"
```
justinmk, line 838:
```suggestion
      title = title .. " [" .. source .. "]"
```

justinmk:
screenshot looks fine to me. Add a bit of whitespace to visually separate it.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,13 @@ local function on_code_action_results(results, opts)
     return
   end
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx : lsp.HandlerContext}
   local function format_item(item)
-    local title = item.action.title:gsub('\r\n', '\\r\\n')
-    return title:gsub('\n', '\\n')
+    local source = vim.lsp.get_client_by_id(item.ctx.client_id).name or 'unknown'
+    local title = item.action.title .. ' [' .. source .. ']'
```
justinmk, line 833: nit: use format(), it is generally more readable

justinmk (approved):

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,15 @@ local function on_code_action_results(results, opts)
     return
   end
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx : lsp.HandlerContext}
   local function format_item(item)
-    local title = item.action.title:gsub('\r\n', '\\r\\n')
-    return title:gsub('\n', '\\n')
+    local source = vim.lsp.get_client_by_id(item.ctx.client_id).name or 'unknown'
+    return string
```
justinmk, line 833: arguably redundant if there is only 1 client attached, can we make it conditional?

```lua
local clients = vim.lsp.get_clients({bufnr=ctx.bufnr})
if #clients > 1 then
  ...
```

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,15 @@ local function on_code_action_results(results, opts)
...
+      .format([[%s [%s] ]], item.action.title, source)
```
justinmk, line 834:
```suggestion
      ([[%s [%s] ]]):format(item.action.title, source)
```

runtime/lua/vim/lsp/buf.lua
```diff
@@ -211,7 +211,7 @@ end
 --- Table must contain `start` and `end` keys with {row,col} tuples using
 --- (1,0) indexing.
 --- Can also be a list of tables that contain `start` and `end` keys as described above,
---- in which case `textDocument/rangesFormatting` support is required.
+--- in which case `textDocument/rangesFormatting` support is required.lspbuf
```
wookayin, line 214: is this an artifact?

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,15 @@ local function on_code_action_results(results, opts)
     return
   end
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx : lsp.HandlerContext}
```
wookayin, line 830:
```suggestion
  ---@param item {action: lsp.Command|lsp.CodeAction, ctx: lsp.HandlerContext}
```

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,14 @@ local function on_code_action_results(results, opts)
...
   local function format_item(item)
-    local title = item.action.title:gsub('\r\n', '\\r\\n')
-    return title:gsub('\n', '\\n')
+    local source = vim.lsp.get_client_by_id(item.ctx.client_id).name or 'unknown'
+    return ([[%s [%s] ]]):format(item.action.title, source)
```
wookayin, line 833: Not sure if we really need this fallback, and if 'unknown' is a good one to display. Do we have any particular path where `vim.lsp.Client.name` can be nil? (it's a non-nil field).

Also please prefer `''` instead of `[[ ]]` here because it reads (slightly) better. Q: why do we need a trailing whitespace here?

wookayin (changes requested):

runtime/lua/vim/lsp/buf.lua
wookayin: Also, please update the commit message / title accordingly as we no longer have an boolean option.

runtime/lua/vim/lsp/buf.lua
JordanllHarper: Trailing whitespace was due to previous impl. `[[%s [%s]]]` is wanted, but trying to do that gave me errors. Using `''` seems to fix this though.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,19 @@ local function on_code_action_results(results, opts)
     return
   end
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx: lsp.HandlerContext}
   local function format_item(item)
-    local title = item.action.title:gsub('\r\n', '\\r\\n')
-    return title:gsub('\n', '\\n')
+    local clients = vim.lsp.get_clients({ bufnr = item.ctx.bufnr })
```
wookayin, line 832: `format_item` will be called many times, one for each of the CodeAction entries so this will be inefficient. `clients` can be moved up to the parent scope as a closure.

wookayin (approved):
Looks good, thank you. (other than news.txt)

JordanllHarper:
Would news.txt need to be updated for this kind of change? (sorry if that's a stupid question)

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,19 @@ local function on_code_action_results(results, opts)
...
+    local title = item.action.title
```
justinmk, line 833: is the crlf replacement needed here?
```suggestion
    local title = item.action.title:gsub('\r\n', '\\r\\n'):gsub('\n', '\\n')
```

justinmk (approved):

runtime/lua/vim/lsp/buf.lua
```diff
@@ -827,11 +827,19 @@ local function on_code_action_results(results, opts)
     return
   end
 
-  ---@param item {action: lsp.Command|lsp.CodeAction}
+  ---@param item {action: lsp.Command|lsp.CodeAction, ctx: lsp.HandlerContext}
   local function format_item(item)
```
MariaSolOs, line 831: Thinking ahead I wonder if it would be more flexible to provide a `format_item` callback setting to let users customize each action as their heart desire instead of having a bunch of boolean settings.

justinmk: Definitely, or even a more generalized solution for overriding handlers. But this PR doesn't add a new boolean flag.

runtime/doc/news.txt
```diff
@@ -75,6 +75,8 @@ LSP
   customizing the transformation of an LSP CompletionItem to |complete-items|.
 • |vim.lsp.diagnostic.from()| can be used to convert a list of
   |vim.Diagnostic| objects into their LSP diagnostic representation.
+• |vim.lsp.buf.code_action()| actions show client name when there are multiple
+  clients.
```
justinmk, line 79: this isn't a breaking change, please move it to the "features" LSP section below

runtime/lua/vim/lsp/buf.lua
JordanllHarper: I would be happy to implement this in the current or a future PR (whatever is most appropriate).

mfussenegger (approved):

runtime/lua/vim/lsp/buf.lua
justinmk: it's a bigger topic, see https://github.com/neovim/neovim/issues/25259#issuecomment-1727336789

PR #30838, opened 2024-10-17
fix(clipboard-osc52): use tmux passthrough sequence when running in tmux

holio0:
Problem:

A nvim instance behind the tmux session cannot send osc52 sequence to the host terminal.

Solution:

Wraps the query in the tmux passthrough sequence when running inside tmux to ensure the query arrives at the "host" terminal.


zeertzjq:
> tmux has a set-clipboard option which, when set to 'on', allows
applications to set the system clipboard using the usual OSC 52 escape
sequence.

holio0:
Got it, thanks!

PR #30842, opened 2024-10-17
feat(highlight): make `PmenuMatch` and `PmenuMatchSel` bold

echasnovski:
Problem: both `PmenuMatch` and `PmenuMatchSel` can provide helpful
  information about characters which actually match query in completion
  candidates. This is not as useful with default regular match, but much
  more useful with 'completopt+=fuzzy'.

Solution: make both highlight groups bold. This will also affect
  (i.e. benefit) other color schemes which do not define groups
  separately.
  This is possible since the recently merged changes to `PmenuMatch` and
  `PmenuMatchSel` combine attributes with underlying "base" groups.
  See PR 29980.

### DIFF
--- a/src/nvim/highlight_group.c
+++ b/src/nvim/highlight_group.c
@@ -144,6 +144,8 @@ static const char e_missing_argument_str[]
 static const char *highlight_init_both[] = {
   "Cursor            guifg=bg      guibg=fg",
   "CursorLineNr      gui=bold      cterm=bold",
+  "PmenuMatch        gui=bold      cterm=bold",
+  "PmenuMatchSel     gui=bold      cterm=bold",
   "PmenuSel          gui=reverse   cterm=reverse,underline blend=0",
   "RedrawDebugNormal gui=reverse   cterm=reverse",
   "TabLineSel        gui=bold      cterm=bold",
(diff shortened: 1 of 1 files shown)


echasnovski:
Here are couple of screenshots:

screenshot_2024-10-17_14:09:29
screenshot_2024-10-17_14:10:30

This works with custom `Pmenu` and `PmenuSel`:
screenshot_2024-10-17_14:13:29

justinmk (approved):

echasnovski:
> Underline would be more visually helpful, no?

I thought about that and did try it. This didn't look like the best approach for these reasons (from most to least impactful for me):
- Underline looks too subtle to be visually useful for me. Tested this in several terminal emulators.
- Default `PmenuSel` has underline as `cterm` effect, so underline in `PmenuMatchSel` will not be useful at all.
- As default color scheme already has the "bold for important stuff (keywords)" approach, it seems to be more coherent.
- Several other editors seem to prefer using bold for this kind of visual feedback. So it might be a good idea to also adopt this.

echasnovski:
@justinmk, is there any concern against merging this?

clason:
It's been approved, no? Just needs a rebase to make CI happy.

PR #30843, opened 2024-10-17
refactor(options): impl default 'titlestring' by format flags

### LINKED ISSUE #1248: refactor(options): impl default 'titlestring' by format flags
At the moment, when `titlestring` is not set, the default value is generated by accessing `curbuf->b_fname`, `bufIsChanged(curbuf)`, `curbuf->b_ffname` & `append_arg_number()` (an extracted version of this is in [`src/nvim/title.c:113`](https://github.com/josephholsten/neovim/blob/rework-titles/src/nvim/title.c#L113)).

Unfortunately, it is not possible to reproduce this behaviour using a statusline format string. The closest might be:

```
:set titlestring=%t%(\ %M%)%(\ \(%F\)%)%a\ -\ VIM
```

where those represent the flags found in `option_defs.h`:

```
#define STL_FILENAME    't'             /* last part (tail) of file path */
#define STL_MODIFIED_ALT 'M'            /* modified flag - other display */
#define STL_FULLPATH    'F'             /* full path of file in buffer */
#define STL_ARGLISTSTAT 'a'             /* argument list status as (x of y) */
```

The only part which differs from the default `titlestring` is `%F`, which shows the full path of the file, including the last part (tail). The default uses the full path of the file with the part in `%t` and the final seperator removed.

The best way to allow for this would be to add a new statusline flag, say `%d` which shows the full path of the directory containing the current file.

Then, since statusline support is on all the time (according to [`src/nvim/version.c:140`](https://github.com/neovim/neovim/blob/master/src/nvim/version.c#L140)), we can simplify the implementation of `maketitle()` to always use the standard statusline processing.

Thoughts? Objections?

zshuzh:
Fixes #1248. This PR implements the default titlestring using the format DSL instead of C, with the help of a suggestion by @justinmk ([link](https://github.com/neovim/neovim/issues/1248#issuecomment-1185487263)).

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -275,9 +275,10 @@ These existing features changed their behavior.
   current window, and it no longer throws |E444| when there is only one window
   on the screen. Global variable `vim.g.pager` is removed.
 
-• 'titlestring' now uses printf-style '%' items (see: 'statusline') to
-  implement the default behaviour. The implementation is equivalent to setting
-  'titlestring' to `%t%(\ %M%)%(\ \(%{expand(\"%:~:h\")}\)%)%a\ -\ Nvim`.
+• Default 'titlestring' is now implemented with 'statusline' "%" format items.
+  This means the default, empty value is essentially an alias to:
+  `%t%(\ %M%)%(\ \(%{expand(\"%:~:h\")}\)%)%a\ -\ Nvim`. This is only an
+  implementation simplification, not a behavior change.
 
 ==============================================================================
 REMOVED FEATURES                                                 *news-removed*
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -275,6 +275,10 @@ These existing features changed their behavior.
   current window, and it no longer throws |E444| when there is only one window
   on the screen. Global variable `vim.g.pager` is removed.
 
+• 'titlestring' now uses printf-style '%' items (see: 'statusline') to
+  implement the default behaviour. The implementation is equivalent to setting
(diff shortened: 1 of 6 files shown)


justinmk:
I'm glad someone finally got around to this! Wow, all tests are passing 😂 Can you confirm this has test coverage? 

And need a mention in `:help news-changed`. And `vim_diff.txt`.

zshuzh:
The functionality has existing test coverage in [test/functional/ui/title_spec.lua](https://github.com/neovim/neovim/blob/master/test/functional/ui/title_spec.lua). Specifically, it tests the default title with an unnamed file and a named file.

I've now also documented the changes in `:h news-changed`, `vim_diff.txt`, and `:h 'titlestring'`.

zshuzh:
@justinmk Any update on getting this merged? I've got some free time this weekend so I can make changes if required.

runtime/doc/news.txt
```diff
@@ -275,6 +275,10 @@ These existing features changed their behavior.
   current window, and it no longer throws |E444| when there is only one window
   on the screen. Global variable `vim.g.pager` is removed.
 
+• 'titlestring' now uses printf-style '%' items (see: 'statusline') to
+  implement the default behaviour. The implementation is equivalent to setting
+  'titlestring' to `%t%(\ %M%)%(\ \(%{expand(\"%:~:h\")}\)%)%a\ -\ Nvim`.
```
justinmk, line 280:
```suggestion
• Default (empty) 'titlestring' is now implemented with 'statusline' "%" format
  items, equivalent to: `%t%(\ %M%)%(\ \(%{expand(\"%:~:h\")}\)%)%a\ -\ Nvim`.
  This is only an implementation simplification, not a behavior change.
```

Issue #30741, opened 2024-10-09
Allow `vim.on_key()` to consume the key and prevent mapping.


errael:
### Problem

This is about a implementing a compatibility layer to provide `vim`'s `popup` API.

Without the option to consume the keys, `vim.on_key()` is not suitable for implementing a `popup`'s  `filter`.

With `vim.on_key()` there's also
```
{key} is the key after mappings have been applied, and {typed} is the
key(s) before mappings are applied
```
Which also interferes with correct operation of `popup`'s  `filter`, since it may need to see each key as it's entered, not multiple keys.  This is likely related to the following.

There's a `popup` option `mapping` 
```
If the filter consumes all the keys, set the "mapping" property to zero
so that mappings do not get in the way
```
In the `vim` code, in `vgetc()` there's a check for `popup_no_mapping` which is missing in the following `neovim` code
```
  if (mod_mask) {           // no mapping after modifier has been read
    no_mapping++;
```
(But my analysis could be flawed, since `no_mapping` can be set in various ways)

-----

This is related to #30504 and #30542.

### Expected behavior

Some things that comes to mind around `vim.on_key()`. Change the signature to
```
on_key(fn: function?, ns_id: integer? [, {options}]): bool?
```

1. Use the return value to indicate if the key should be discarded and/or has been handled
2. Add `allow_mapping` option to `vim.on_key()`, default `true`
   For when the `mapping` property on the `popup` is active.
3. Add `enable_return_value_controls_discard` option to `vim.on_key()`
   Only needed if there's compatibility concerns around (1)?
   Could also be useful to insure that `on_key` callbacks that do not have this set are invoked first.

There's some weirdness possible since, in `vgetc()`, the mappings are done before the callback is executed. So, depending on implementation, when `allow_mapping` is `false` then other uses of `vim.on_key()` might see different arguments than they would otherwise have seen.

zeertzjq:
Is using `getchar()` or `getcharstr()` no feasible?

errael:
> Is using `getchar()` or `getcharstr()` not feasible for this specific case?

When I first looked at this, `getchar()` seemed a poor fit. Could either hang waiting for a char (not sure where it hangs and what would be prevented). Could have a timer loop checking for an available character, then get it when available.

But after getting the character, is the idea to use `feedkeys()` to put it back if it is not one you want? Wouldn't this potentially re-order the keys?

`getchar` feels very messy.

errael:
[...]

PR #30844, opened 2024-10-17
feat(float): allow enabling mouse for non-focusable window

zeertzjq:
Problem:  Cannot allow mouse interaction for non-focusable float window.
Solution: Add a "mouse" field to float window config.

Extracted from #30542, removing the callback part.

### DIFF
--- a/src/nvim/api/win_config.c
+++ b/src/nvim/api/win_config.c
@@ -129,7 +129,12 @@
 ///            fractional.
 ///   - focusable: Enable focus by user actions (wincmds, mouse events).
 ///       Defaults to true. Non-focusable windows can be entered by
-///       |nvim_set_current_win()|.
+///       |nvim_set_current_win()|, or, when the `mouse` field is set to true,
+///       by mouse events.
+///   - mouse: Specify how this window interacts with mouse events.
+///       Defaults to `focusable` value.
+///       - If false, mouse events pass through this window.
+///       - If true, mouse events interact with this window normally.
 ///   - external: GUI should display the window as an external
 ///       top-level window. Currently accepts no other positioning
 ///       configuration together with this.
@@ -714,6 +719,7 @@ Dict(win_config) nvim_win_get_config(Window window, Arena *arena, Error *err)
   PUT_KEY_X(rv, focusable, config->focusable);
   PUT_KEY_X(rv, external, config->external);
   PUT_KEY_X(rv, hide, config->hide);
+  PUT_KEY_X(rv, mouse, config->mouse);
 
   if (wp->w_floating) {
     PUT_KEY_X(rv, width, config->width);
@@ -1202,6 +1208,11 @@ static bool parse_win_config(win_T *wp, Dict(win_config) *config, WinConfig *fco
 
   if (HAS_KEY_X(config, focusable)) {
     fconfig->focusable = config->focusable;
+    fconfig->mouse = config->focusable;
+  }
+
+  if (HAS_KEY_X(config, mouse)) {
+    fconfig->mouse = config->mouse;
   }
 
   if (HAS_KEY_X(config, zindex)) {
--- a/src/nvim/grid_defs.h
+++ b/src/nvim/grid_defs.h
@@ -80,8 +80,8 @@ struct ScreenGrid {
   // whether the compositor should blend the grid with the background grid
   bool blending;
 
-  // whether the grid can be focused with mouse clicks.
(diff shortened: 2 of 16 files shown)


src/nvim/grid_defs.h
```diff
@@ -80,8 +80,8 @@ struct ScreenGrid {
   // whether the compositor should blend the grid with the background grid
   bool blending;
 
-  // whether the grid can be focused with mouse clicks.
-  bool focusable;
+  // whether the grid interacts with mouse events.
+  bool mouse_enabled;
```
justinmk, line 84:
```suggestion
  bool mouse;
```

src/nvim/api/win_config.c
```diff
@@ -129,7 +129,12 @@
 ///            fractional.
 ///   - focusable: Enable focus by user actions (wincmds, mouse events).
 ///       Defaults to true. Non-focusable windows can be entered by
-///       |nvim_set_current_win()|.
+///       |nvim_set_current_win()|, or, when the `mouse` field is set to true,
+///       by mouse events.
+///   - mouse: Specify how this window affects mouse events.
```
justinmk, line 134:
```suggestion
///   - mouse: Specify how this window treats mouse events.
```

src/nvim/api/win_config.c
```diff
@@ -129,7 +129,12 @@
...
+///       Defaults to `focusable` value.
+///       - If false, mouse events ignore this window.
+///       - If true, mouse events interact with this window normally.
```
justinmk, line 137:
```suggestion
///       - If false, window ignores mouse events.
///       - If true, mouse interacts with this window normally.
```

justinmk:
https://github.com/neovim/neovim/pull/30542#issuecomment-2419095045 answers my initial question:

> I'm also starting to doubt whether the callback is needed at all, as after #30741 is implemented, `vim.on_key()` along with `mouse = true` can kind of already achieve what the callback does.

Somewhat of a one-way door, unless we define `mouse: Object` in this PR, so that we can extend it later?

zeertzjq:
Dict fields are not part of the API contract.

src/nvim/api/win_config.c
zeertzjq: > window ignores mouse events.

I don't think this is correct.

src/nvim/api/win_config.c
zeertzjq: No

src/nvim/grid_defs.h
zeertzjq: No

src/nvim/api/win_config.c
justinmk: That's a bit pedantic though. Being technically correct leads to awkward documentation without any benefit AFAICT.

zeertzjq:
Hmm, I mean, they are not a part of the signature. If a Dict field that accepts a boolean is extended to also accept a callback, it's a backward-compatible change.

justinmk:
> If a Dict field that accepts a boolean is extended to also accept a callback, it's a backward-compatible change.

I think we should formalize that and also extend it to the function parameters/signature. They are equivalent in practice.

justinmk (approved):

errael:
FYI, On `nvim_open_win()` with `mouse = true, focusable = false`; if you left click on the floating window, the floating window gets focus, can edit/select. Isn't this wrong since `focusable` is `false`? You can use `vim.fn.win_gotoid()` to focus the window.

BTW, this is is not an issue for the popup window compatibility code since it consumes/discards these keys in a `vim.on_key()` callback. (Assuming that #30741 gets a fix.)

zeertzjq:
> FYI, On `nvim_open_win()` with `mouse = true, focusable = false`; if you left click on the floating window, the floating window gets focus, can edit/select. Isn't this wrong since `focusable` is `false`? You can use `vim.fn.win_gotoid()` to focus the window.
> 
> BTW, this is is not an issue for the popup window compatibility code since it consumes/discards these keys in a `vim.on_key()` callback. (Assuming that #30741 gets a fix.)

Hmm, I initially intended that for the callback case, not the `true` case, so maybe the callback is needed after all.

errael:
> initially intended that for the callback case

I can verify that returning `false` from the callback prevents a click from focusing. Instead of a `boolean` could have `mouse=` take a three value enum, for example: `ignore`, `click_will_focus`, `click_will_not_focus`.

PR #30845, opened 2024-10-17
feat(vim.ui.open): add support for `lemonade`

### LINKED ISSUE #30837: Add lemonade option for `vim.ui.open`
### Problem

`vim.ui.open` should support [lemonade](https://github.com/lemonade-command/lemonade) as an option, similar to how `lemonade` is an option for the clipboard.

Currently neovim uses `lemonade copy` and `lemonade paste` to support copy and pasting through an ssh connection (see `:h clipboard-tool`), but `lemonade` also supports a `lemonade open` to open url's from within an ssh connection in the host machine.

This should simply just be a change above https://github.com/neovim/neovim/blob/fae4d3fa1b65b7632d048cad2a1beb7afaa61c21/runtime/lua/vim/ui.lua#L168 to add
```lua
elseif vim.fn.executable('lemonade') == 1 then
    cmd = { 'lemonade', 'open', path }
```

I'm willing to make this PR if given approval for the idea, but open to others implementing as well

### Expected behavior

This would allow users to use `vim.ui.open` (through `gx` or other means) to open files/links through an ssh connection

uthmanmoh:
`lemonade` reference: https://github.com/lemonade-command/lemonade

Closes #30837

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -240,6 +240,10 @@ UI
   which controls the tool used to open the given path or URL. If you want to
   globally set this, you can override vim.ui.open using the same approach
   described at |vim.paste()|.
+- `vim.ui.open()` now supports
+  [lemonade](https://github.com/lemonade-command/lemonade) as an option for
+  opening urls/files. This is handy if you are in an ssh connection and use
+  `lemonade`.
 • The |ins-completion-menu| now supports cascading highlight styles.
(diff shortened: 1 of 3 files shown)


uthmanmoh:
Appreciate review on this one @justinmk. Wasn't able to get all tests passing locally (with and without this change) but looks like the CI tests passed so that should be good right?

PR #30859, opened 2024-10-19
feat(terminal)!: make 'belloff' and 'visualbell' apply to terminal bell

zeertzjq:
vim-patch:8.2.4744: a terminal window can't use the bell
vim-patch:8.2.4745: using wrong flag for using bell in the terminal

BREAKING CHANGE: Bells from :terminal are now silent by default, unless
'belloff' option doesn't contain "term" or "all".


justinmk (approved):
This is a good change.

PR #30860, opened 2024-10-19
feat(lsp): include `end_col` and `end_lnum` in `vim.lsp.buf.symbols_t…

ofseed:
Similar to https://github.com/neovim/neovim/pull/29164.

PR #30861, opened 2024-10-19
build(deps): bump libuv to v1.49.2

clason:
https://github.com/libuv/libuv/releases/tag/v1.49.2


dundargoc (approved):

clason:
Failure is from https://github.com/libuv/libuv/pull/4578

ofseed:
Is this caused by libuv? The CI test for my PR https://github.com/neovim/neovim/pull/30860 which only involves runtime files has the same failure. Shouldn't the CI environment there still be on the previous version?

clason:
Hmm, indeed. But the failure is consistent here and only goes away if I revert to the commit before the mentioned PR.

PR #30863, opened 2024-10-19
ci: bump Intel macOS runners to 13

clason:
Problem: macos-12 GH runners are deprecated and will be removed soon.

Solution: use macos-13 runners instead.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -81,9 +81,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner: [ macos-12, macos-14 ]
+        runner: [ macos-13, macos-14 ]
         include:
-          - runner: macos-12
+          - runner: macos-13
             arch: x86_64
           - runner: macos-14
             arch: arm64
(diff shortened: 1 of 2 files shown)


clason:
required tests need adapting @bfredl

dundargoc:
> Should we rename the flavors (intel and chocolate) instead of hardcoding version names?

Don't think it's needed. It looks like macos-13 is the last free intel tier anyway.

clason:
>  It looks like macos-13 is the last free intel tier anyway.

No, it's only the last Intel runner _currently_ available on the free tier. That may change in the future, when macos-13 is deprecated -- they could, for instance, provide an additional `macos-14-intel` runner. (Or, indeed, not, but it'll be awhile until the last Intel mac sold is obsolete and can be dropped like a hot potato.)

clason:
Still, how do we want to handle this? Maybe simplest to un-require the intel checks (at least for now) so we don't block the currently open PRs?

dundargoc:
I'm convinced, I pushed a fix that should make the names version independent (opted for `arm` and `intel` just because, no actual preference except I will be vetoing "chocolate").

> Still, how do we want to handle this? Maybe simplest to un-require the intel checks (at least for now) so we don't block the currently open PRs?

Don't think it really matters as long as they're done around the same time.

clason:
> Don't think it really matters as long as they're done around the same time.

Well, that forces every currently open PR to rebase on master.

dundargoc:
What's the alternative? Don't we need to rebase all PRs regardless once we add checks with new names?

clason:
The alternative is removing the macos-12 checks from the required list but _not_ adding the macos-intel ones (immediately); this would allow both new and old PRs to pass.

dundargoc:
Hmm, true. Although we'd need to remember to activate the checks in the future, so there's risk of us forgetting it. Aight, go for it.

tris203:
Backported in #31124

PR #30865, opened 2024-10-19
test(rpc): retry flaky 'vim.rpcrequest and vim.rpcnotify' test

clason:
Problem: 'vim.rpcrequest and vim.rpcnotify' is flaky on Windows.

Solution: retry it.

### DIFF
--- a/test/functional/lua/vim_spec.lua
+++ b/test/functional/lua/vim_spec.lua
@@ -1319,7 +1319,9 @@ describe('lua stdlib', function()
     ]],
     }
     feed('<cr>')
-    eq({ 3, NIL }, api.nvim_get_var('yy'))
+    retry(10, nil, function()
+      eq({ 3, NIL }, api.nvim_get_var('yy'))
+    end)
 
     exec_lua([[timer:close()]])
   end)


test/functional/lua/vim_spec.lua
```diff
@@ -1326,7 +1326,9 @@ describe('lua stdlib', function()
     ]],
     }
     feed('<cr>')
-    eq({ 3, NIL }, api.nvim_get_var('yy'))
+    retry(nil, 200, function()
```
dundargoc, line 1329: I think the 200 should be the first argument since that is the number of tries. The second argument seems to be `max_ms`, which I assume means "rerun as much as possible until time runs out".

clason: Yes, that is what bfredl suggested.

dundargoc: Well, it seems weird to me since the default value of `max_ms` is 10000 which is much bigger than 200. If it was intentional then it's fine.

clason: @bfredl ?

clason: Changed it to `10, nil` to match the retry of the test before; let's see if that works too.

clason: Seems to work as well, so let's go with that.

PR #30866, opened 2024-10-19
fix(float): cursor visible in "hidden" floating window

### LINKED ISSUE #30553: Unexpected cursor artifacts when hidden window is focused
### Problem

This is about having a visible cursor in a non-focused area; it is a visual artifact. This happens with both a text window and the command line.

This is a follow on issue after #30503 is fixed. IMHO, it seems low priority.

This might be considered an artificial situation. It's possible to enter hidden windows through scripts (or command line). Even though, in general, a script shouldn't leave the focus in a hidden window, there's some behavior I'm seeing that might be considered a bug.

Note: this issue is independent of whether or not the hidden window is focusable.

### Steps to reproduce

1. Edit the test file below, move the cursor to the middle of the file.
2. `:sou` creating the hidden focusable window
3. Execute the mapping `Z2`. Moves the focus to the hidden window
   Does `vim.api.nvim_set_current_win(popup_wid)` 

Observe: Cursor is visible in the main/**non-focused** window.

4. Scroll the window with the mouse so only last line of buffer is visible in the window.

Observe the cursor doesn't move and ends up away from the buffer's text lines.

5. Execute some ":" commands

Observe the cursor move around and get stuck in the command line.

```lua
local popup_wid
local function Pop1Any()
    vim.api.nvim_buf_set_lines(bnr, 0, -1, true, {'simple', 'win'})

	popup_wid = vim.api.nvim_open_win(bnr, false, {
		relative = "editor",
		style = "minimal",
		width = 10,
		height = 2,
		focusable = false,
		hide = true,
		col = 10,
		row = 10,
	})
end

Pop1Any()

vim.keymap.set('n', 'Z1', function() vim.print(vim.inspect(vim.api.nvim_list_wins())) end)
vim.keymap.set('n', 'Z2', function() vim.api.nvim_set_current_win(popup_wid) end)
vim.keymap.set('n', 'Z3', ":call nvim_set_current_win(" .. popup_wid .. ")<CR>")
vim.keymap.set('n', 'Z4', function() vim.api.nvim_win_set_config(popup_wid, {hide=false}) end)
vim.keymap.set('n', 'Z5', function() vim.api.nvim_win_set_config(popup_wid, {hide=true}) end)
vim.o.tm = 5000
```

### Expected behavior

Not sure. Maybe when hidden is focused, probably no visible cursor in regular; and for command line, only visible while entering a command.

### Nvim version (nvim -v)

NVIM v0.11.0-dev - https://github.com/neovim/neovim/commit/7b71fdbc1e9fcb71e642e67e0ac9a2711dd67df0

NA

### Operating system/version

ubuntu

### Terminal name/version

gnome

xterm-256color

### Installation

make install

glepnir:
Problem: Currently when change curwin to a floating window the cursor is missing visual effect is presented in a hidden floating window.

Solution: hide cursor when jump into a hidden floating window. and show cursor after jump to a normal window.

Fix #30553

### DIFF
--- a/src/nvim/ui.c
+++ b/src/nvim/ui.c
@@ -537,7 +537,21 @@ void ui_flush(void)
   if (!ui_active()) {
     return;
   }
+
+  static bool was_busy = false;
+
   cmdline_ui_flush();
+
+  if (State != MODE_CMDLINE && curwin->w_floating && curwin->w_config.hide) {
+    if (!was_busy) {
+      ui_call_busy_start();
+      was_busy = true;
+    }
+  } else if (was_busy) {
+    ui_call_busy_stop();
+    was_busy = false;
+  }
+
   win_ui_flush(false);
   msg_ext_ui_flush();
   msg_scroll_flush();
--- a/test/functional/ui/float_spec.lua
+++ b/test/functional/ui/float_spec.lua
@@ -9585,6 +9585,211 @@ describe('float window', function()
                                                   |
         ]])
       end
+
+      --
+      -- Cursor visibility:
+      --
+      -- Cursor is not visible in a hide=true floating window.
+      api.nvim_set_current_win(win)
+      if multigrid then
+        screen:expect({
+          grid = [[
+          ## grid 1
+            [2:----------------------------------------]|*6
+            [3:----------------------------------------]|
+          ## grid 2
+                                                    |
+            {0:~                                       }|*5
+          ## grid 3
+                                                    |
+          ## grid 4 (hidden)
+            {1:          }|
+            {2:~         }|
+          ]], win_viewport={
+          [2] = {win = 1000, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+          [4] = {win = 1001, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+        }, win_viewport_margins={
+          [2] = {
+            bottom = 0,
+            left = 0,
+            right = 0,
+            top = 0,
+            win = 1000
+          },
+          [4] = {
+            bottom = 0,
+            left = 0,
+            right = 0,
+            top = 0,
+            win = 1001
+          }
+        }
+        })
+      else
+        screen:expect({
+          grid = [[
+                                                    |
+            {0:~                                       }|*5
+                                                    |
+          ]]
+        })
+      end
+
+      -- Show cursor if cmdline is entered while curwin is a hide=true floating window.
+      feed(':')
+      if multigrid then
+        screen:expect({
+          grid = [[
+          ## grid 1
+            [2:----------------------------------------]|*6
+            [3:----------------------------------------]|
+          ## grid 2
+                                                    |
+            {0:~                                       }|*5
+          ## grid 3
+            :^                                       |
+          ## grid 4 (hidden)
+            {1:          }|
+            {2:~         }|
+          ]], win_viewport={
+          [2] = {win = 1000, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+          [4] = {win = 1001, topline = 0, botline = 2, curline = 0, curcol = 0, linecount = 1, sum_scroll_delta = 0};
+        }, win_viewport_margins={
+          [2] = {
+            bottom = 0,
+            left = 0,
+            right = 0,
+            top = 0,
+            win = 1000
+          },
+          [4] = {
+            bottom = 0,
+            left = 0,
+            right = 0,
+            top = 0,
+            win = 1001
+          }
+        }
+        })
+      else
+        screen:expect({
+          grid = [[
+                                                    |
+            {0:~                                       }|*5
+            :^                                       |
+          ]]
+        })
+      end
+      feed('<ESC>')
+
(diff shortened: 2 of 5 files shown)


src/nvim/api/ui_events.in.h
```diff
@@ -41,6 +41,10 @@ void option_set(String name, Object value)
   FUNC_API_SINCE(4);
 void chdir(String path)
   FUNC_API_SINCE(12);
+void cursor_on(void)
+  FUNC_API_SINCE(13);
+void cursor_off(void)
+  FUNC_API_SINCE(13);
```
zeertzjq, line 47: Why are these two needed? Don't `busy_stop` and `busy_start` do the same thing?

zeertzjq, line 47: Why are these two needed? Don't `busy_stop` and `busy_start` do the same thing?

zeertzjq (dismissed):

src/nvim/api/ui_events.in.h
glepnir: Yes, after a try it works.  I'm still not familiar with the tui part 😢. So I tried sending invisible/normal. will fix

glepnir: oh i found it `flush_buf_end` it will check the `tui->is_invisible && !should_invisible(tui)` to show or hide cursor

zeertzjq:
This doesn't handle the case when using `nvim_win_set_config` on current window

glepnir:
and into cmdline from a hidden floating window the cursor need shown on cmdline or keep hide ? if need shown `command_line_enter` also need process right ?

zeertzjq:
Hmm, maybe the change should be done in `ui_comp_grid_cursor_goto()` instead?

src/nvim/ui.c
```diff
@@ -70,6 +70,7 @@ bool ui_cb_ext[kUIExtCount];  ///< Internalized UI capabilities.
 static bool has_mouse = false;
 static int pending_has_mouse = -1;
 static bool pending_default_colors = false;
+static bool was_busy = false;
```
justinmk, line 73: only `ui_flush` uses this so it should be defined in that scope.

test/functional/ui/float_spec.lua
```diff
@@ -9195,6 +9195,146 @@ describe('float window', function()
...
+          grid = [[
+                                                    |
+            {0:~                                       }|*5
+            :^                                       |
+          ]]
+        })
+      end
+      feed('<ESC>')
+
+      -- show cursor after swith to a normal window
```
justinmk, line 9292:
```suggestion
      -- show cursor after switch to a normal window
```

src/nvim/ui.c
glepnir: yes. I wrote it like this at first. but I was afraid of breaking any rules so I moved it to the top

test/functional/ui/float_spec.lua
```diff
@@ -9195,6 +9195,146 @@ describe('float window', function()
                                                   |
         ]])
       end
+
+      -- check cursor visible
```
justinmk, line 9199:
```suggestion
      -- 
      -- Cursor visibility:
      --
      -- Cursor is not visible in a hide=true floating window.
```

test/functional/ui/float_spec.lua
```diff
@@ -9195,6 +9195,146 @@ describe('float window', function()
...
+        screen:expect({
+          grid = [[
+                                                    |
+            {0:~                                       }|*5
+                                                    |
+          ]]
+        })
+      end
+
+      -- show cursor in cmdline when jump into cmdline from a hidden floating window
```
justinmk, line 9245:
```suggestion
      -- Show cursor if cmdline is entered while curwin is a hide=true floating window.
```

test/functional/ui/float_spec.lua
```diff
@@ -9195,6 +9195,146 @@ describe('float window', function()
...
+          grid = [[
+                                                    |
+            {0:~                                       }|*5
+            :^                                       |
+          ]]
+        })
+      end
+      feed('<ESC>')
+
+      -- show cursor after switch to a normal window
```
justinmk, line 9292:
```suggestion
      -- Show cursor after switching to a normal window (hide=false).
```

src/nvim/ui.c
```diff
@@ -536,7 +536,21 @@ void ui_flush(void)
...
+  static bool was_busy = false;
+
   cmdline_ui_flush();
+
+  if (State != MODE_CMDLINE && curwin->w_floating) {
+    if (curwin->w_config.hide) {
+      ui_call_busy_start();
+      was_busy = true;
+    }
+  } else if (was_busy) {
```
zeertzjq, line 549: The branching here doesn't seem correct. When switching from a hidden floating window to a non-hidden floating window, the cursor isn't remade visible.

src/nvim/ui.c
```diff
@@ -536,7 +536,27 @@ void ui_flush(void)
   if (!ui_active()) {
     return;
   }
+
+  static bool was_busy = false;
+
   cmdline_ui_flush();
+
+  if (State != MODE_CMDLINE && curwin->w_floating) {
+    if (curwin->w_config.hide) {
```
zeertzjq, line 545: Why don't you merge these two conditions?

glepnir: mistake sry.

PR #30868, opened 2024-10-19
`vim.diagnostic.setqflist` improvements

jfly:
I made 2 changes here:

1. We use the new `'u'` action to update the quickfix list so we don't lose our position in the quickfix list when updating it.
2. Rather than creating a new quickfix list each time, we update the exiting one if we've already created one.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -2,6 +2,8 @@ local api, if_nil = vim.api, vim.F.if_nil
 
 local M = {}
 
+local _qf_id = nil
+
 --- [diagnostic-structure]()
 ---
 --- Diagnostics use the same indexing as the rest of the Nvim API (i.e. 0-based
@@ -848,9 +850,24 @@ local function set_list(loclist, opts)
   local diagnostics = get_diagnostics(bufnr, opts --[[@as vim.diagnostic.GetOpts]], false)
   local items = M.toqflist(diagnostics)
   if loclist then
-    vim.fn.setloclist(winnr, {}, ' ', { title = title, items = items })
+    vim.fn.setloclist(winnr, {}, 'u', { title = title, items = items })
   else
-    vim.fn.setqflist({}, ' ', { title = title, items = items })
+    -- Check if the diagnostics quickfix list no longer exists.
+    if _qf_id and vim.fn.getqflist({ id = _qf_id }).id == 0 then
+      _qf_id = nil
+    end
+
+    -- If we already have a diagnostics quickfix, update it rather than creating a new one.
+    -- This avoids polluting the finite set of quickfix lists, and preserves the currently selected
+    -- entry.
+    vim.fn.setqflist({}, _qf_id and 'u' or ' ', {
+      title = title,
+      items = items,
+      id = _qf_id,
+    })
+
+    -- Get the id of the newly created quickfix list.
+    _qf_id = vim.fn.getqflist({ id = 0 }).id
   end
   if open then
     api.nvim_command(loclist and 'lwindow' or 'botright cwindow')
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -1,6 +1,8 @@
 local api, if_nil = vim.api, vim.F.if_nil
 
-local M = {}
+local M = {
+  _qf_id = nil,
+}
```
justinmk, line 5: this could be a local variable in the module, no need to quasi-expose it in the public interface.

```suggestion
local M = {}

_qf_id = nil
```

runtime/lua/vim/diagnostic.lua
```diff
@@ -827,9 +829,32 @@ local function set_list(loclist, opts)
...
+      vim.fn.setqflist({}, 'u', {
+        title = title,
+        items = items,
+        id = M._qf_id,
+      })
+    else
+      vim.fn.setqflist({}, ' ', {
+        title = title,
+        items = items,
+        id = M._qf_id,
```
justinmk, line 852: here it would be `nil` ?
```suggestion
```

runtime/lua/vim/diagnostic.lua
```diff
@@ -827,9 +829,32 @@ local function set_list(loclist, opts)
...
+      })
+
+      -- Get the id of the newly created quickfix list.
+      M._qf_id = vim.fn.getqflist({ id = 0 }).id
+    end
```
justinmk, line 857:
```suggestion
    vim.fn.setqflist({}, M._qf_idwand 'u' or ' ', {
      title = title,
      items = items,
      id = M._qf_id,
    })
    M._qf_id = vim.fn.getqflist({ id = 0 }).id
```

justinmk (approved):
Worth a try IMO. Let's wait a few days for comments, then ping me if there are none.

runtime/lua/vim/diagnostic.lua
jfly: Done!

runtime/lua/vim/diagnostic.lua
jfly: Done! Note: this does mean we re-query for the quickfix list id even when we're just updating it. I don't see a problem with that.

(Also, my old code was kind of weird that it passed a `nil` value for `id` when creating a new quickfix list. Thanks to your suggestion, that's gone now.)

jfly (commented):
> Let's wait a few days for comments, then ping me if there are none.

Thanks @justinmk! I've responded to your feedback and tested out the latest version of the code.

This is ready to merge from my perspective.

mfussenegger (commented):
Maybe worth to distinguish by title/context or something.

E.g. I frequently use these mappings:

```
command -nargs=0 -bar Errors :lua vim.diagnostic.setqflist { title = "Errors", severity = vim.diagnostic.severity.ERROR }
command -nargs=0 -bar Warnings :lua vim.diagnostic.setqflist { title = "Warnings", severity = vim.diagnostic.severity.WARN }
```

and I'd kinda expect that Warnings/Errors create separate lists.

justinmk:
> Maybe worth to distinguish by title/context or something.

Can this be done here?

jfly:
> Can this be done here?

Yes, sorry for the silence here. I'm busy this week, though. I can do this next week. Or, @yochem, feel free to take this over if you've got time this week!

justinmk:
Doesn't need to block this. Would you mind rebasing this though?

yochem:
Sorry, I'm also a bit busy these upcoming weeks. @jfly if you could work on this next week, that'll be great! My current implementation idea was:

Instead of a variable that only stores an id of the used qfwindow, have a table that maps context to the ids. Context would be a table (or serialized version) of the following:

- title
- bufnr (if provided)
- namespace (if provided)

nNeD5:
Face a bug probably related to this.
My issue about it: https://github.com/neovim/neovim/issues/31540

PR #30869, opened 2024-10-19
perf(decor): improve performance for long lines

vanaigr:
# Problem

Currently, to calculate the final syntax highlighting for a given column in a line, `decor_redraw_col()` has to iterate through all `DecorRange`s, and combine their highlighting. This causes performance problems with long lines, since syntax highlighting uses ephemeral extmarks to specify highlighting ranges, and all extmarks are added before drawing a line and have to be iterated, even though most of them are very far away from the current column.

Additionally, since `DecorRange` have to be sorted by priority and insertion order, inserting a new `DecorRange` requires moving other `DecorRange`s around, which can also cause performance issues since `DecorRange` is quite big.  Also, `decor_redraw_col()` removes `DecorRange`s that end before the current column, which means all `DecorRange`s are moved to a different place in the list on every call.

# Solution

This PR makes it so that `DecorRange`s don't have to be moved around when adding/deleting them. This is achieved with a second list that stores indices into the first list. Insertion order is stored in a field in `DecorRange`. 
 
The indices list is divided into 2 parts. The first part stores indices to currently active ranges (which start before the current position and end after it). The second part stores indices to future ranges (which start after the current position).

The second part of the array is sorted by position instead of priority+insertion order. This makes it possible to only check a few `DecorRange`s on each column when making the ranges active. The first part is still sorted by priority+insertion order.

Last, the space from a removed `DecorRange` is used to store an index to the next removed `DecorRange` in the list, so nothing has to move when removing a range from the middle of the array. New ranges are added in the holes from previously removed ranges.

# Results

The measurements were done with `nvim-treesitter` highlighting:

|Time, ms|Before|After|
|-|-|-|
|Redraw|411|1|
|Adding extmarks|41|41|
|Total|452|42|

File used for measurements:

<details>

<summary>file.json</summary>

```json
{"a":{"a":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"b":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"c":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"d":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"e":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"f":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}}},"b":{"a":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"b":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"c":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"d":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"e":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"f":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}}},"c":{"a":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"b":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"c":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"d":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"e":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"f":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}}},"d":{"a":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"b":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"c":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"d":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"e":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"f":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}}},"e":{"a":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"b":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"c":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"d":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"e":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"f":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}}},"f":{"a":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"b":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"c":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"d":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"e":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}},"f":{"a":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"b":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"c":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"d":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"e":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}},"f":{"a":{},"b":{},"c":{},"d":{},"e":{},"f":{}}}}}
```

</details>

Command used to measure the total time:
```lua
=(function() local a = vim.uv.hrtime(); vim.cmd('redraw!'); local b = vim.uv.hrtime(); return (b - a) / 1000000 end)()
```

"Adding marks" row is the time taken by [`on_line_impl()`](https://github.com/neovim/neovim/blob/4833a63929a83c11379036ead33913711134015b/runtime/lua/vim/treesitter/highlighter.lua#L293) in highlighter.lua.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -247,7 +247,8 @@ OPTIONS
 
 PERFORMANCE
 
-• TODO
+• Significantly reduced redraw time for long lines with treesitter
+  highlighting.
 
 PLUGINS
 
--- a/src/nvim/decoration.h
+++ b/src/nvim/decoration.h
@@ -27,17 +27,19 @@ typedef enum {
   kDecorKindVirtText,
   kDecorKindVirtLines,
   kDecorKindUIWatched,
-} DecorRangeKind;
+} DecorRangeKindEnum;
+
+typedef uint8_t DecorRangeKind;
 
 typedef struct {
   int start_row;
   int start_col;
   int end_row;
   int end_col;
-  int ordering;
+  int ordering;  ///< range insertion order
   DecorPriority priority;
   bool owned;  ///< ephemeral decoration, free memory immediately
-  DecorRangeKind kind: 8;
+  DecorRangeKind kind;
   // next pointers MUST NOT be used, these are separate ranges
   // vt->next could be pointing to freelist memory at this point
   union {
--- a/src/nvim/decoration.h
+++ b/src/nvim/decoration.h
@@ -34,6 +34,10 @@ typedef struct {
   int start_col;
   int end_row;
   int end_col;
+  int ordering;
+  DecorPriority priority;
+  bool owned;  ///< ephemeral decoration, free memory immediately
+  DecorRangeKind kind: 8;
   // next pointers MUST NOT be used, these are separate ranges
   // vt->next could be pointing to freelist memory at this point
   union {
@@ -46,9 +50,6 @@ typedef struct {
     } ui;
   } data;
   int attr_id;  ///< cached lookup of inl.hl_id if it was a highlight
-  bool owned;   ///< ephemeral decoration, free memory immediately
-  DecorPriority priority;
-  DecorRangeKind kind;
   /// Screen column to draw the virtual text.
   /// When -1, it should be drawn on the current screen line after deciding where.
   /// When -3, it may be drawn at a position yet to be assigned.
@@ -57,9 +58,23 @@ typedef struct {
   int draw_col;
 } DecorRange;
 
+typedef union {
+  DecorRange range;
+  int next_free_i;
+} DecorRangeSlot;
+
 typedef struct {
   MarkTreeIter itr[1];
-  kvec_t(DecorRange) active;
+  kvec_t(DecorRangeSlot) slots;
+  kvec_t(int) ranges_i;
+  /// Indices in [0; current_end) range of `ranges_i` point to ranges that start before the
+  /// current position and are sorted by priority and order of insertion.
+  int current_end;
+  /// Indices in [future_begin, kv_size(ranges_i)) range of `ranges_i` point to ranges that
+  /// start after the current position and are sorted by starting position.
+  int future_begin;
+  int free_slot_i;  ///< Index of the last freed slot.
+  int new_range_ordering;  ///< Index for keeping track of range insertion order.
   win_T *win;
   int top_row;
   int row;
@@ -76,11 +91,21 @@ typedef struct {
   bool running_decor_provider;
 } DecorState;
 
-EXTERN DecorState decor_state INIT( = { 0 });
+EXTERN DecorState decor_state INIT( = { .free_slot_i = -1 });
 // TODO(bfredl): These should maybe be per-buffer, so that all resources
 // associated with a buffer can be freed when the buffer is unloaded.
 EXTERN kvec_t(DecorSignHighlight) decor_items INIT( = KV_INITIAL_VALUE);
 
 #ifdef INCLUDE_GENERATED_DECLARATIONS
 # include "decoration.h.generated.h"
+# include "decoration.h.inline.generated.h"
 #endif
+
+static inline int decor_redraw_col(win_T *wp, int col, int win_col, bool hidden, DecorState *state)
+  FUNC_ATTR_ALWAYS_INLINE
+{
+  if (col <= state->col_until) {
+    return state->current;
+  }
+  return decor_redraw_col_impl(wp, col, win_col, hidden, state);
+}
--- a/src/nvim/decoration.h
+++ b/src/nvim/decoration.h
@@ -58,6 +58,9 @@ typedef struct {
   int draw_col;
 } DecorRange;
 
+/// DecorRange can be removed from `DecorState` list in any order,
+/// so we track available slots using a freelist (with `next_free_i`).
+/// The list head is in `DecorState.free_slot_i`.
 typedef union {
   DecorRange range;
   int next_free_i;
@@ -67,14 +70,16 @@ typedef struct {
   MarkTreeIter itr[1];
   kvec_t(DecorRangeSlot) slots;
   kvec_t(int) ranges_i;
-  /// Indices in [0; current_end) range of `ranges_i` point to ranges that start before the
-  /// current position and are sorted by priority and order of insertion.
+  /// Indices in [0; current_end) of `ranges_i` point to ranges that start
+  /// before current position. Sorted by priority and order of insertion.
   int current_end;
-  /// Indices in [future_begin, kv_size(ranges_i)) range of `ranges_i` point to ranges that
-  /// start after the current position and are sorted by starting position.
+  /// Indices in [future_begin, kv_size(ranges_i)) of `ranges_i` point to
+  /// ranges that start after current position. Sorted by starting position.
   int future_begin;
-  int free_slot_i;  ///< Index of the last freed slot.
-  int new_range_ordering;  ///< Index for keeping track of range insertion order.
+  /// Head of DecorRangeSlot freelist. -1 if none are freed.
+  int free_slot_i;
+  /// Index for keeping track of range insertion order.
+  int new_range_ordering;
   win_T *win;
   int top_row;
   int row;
@@ -91,7 +96,7 @@ typedef struct {
   bool running_decor_provider;
 } DecorState;
 
-EXTERN DecorState decor_state INIT( = { .free_slot_i = -1 });
+EXTERN DecorState decor_state INIT( = { 0 });
 // TODO(bfredl): These should maybe be per-buffer, so that all resources
 // associated with a buffer can be freed when the buffer is unloaded.
 EXTERN kvec_t(DecorSignHighlight) decor_items INIT( = KV_INITIAL_VALUE);
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -473,6 +473,8 @@ bool decor_redraw_line(win_T *wp, int row, DecorState *state)
   int const cur_end = state->current_end;
   int fut_beg = state->future_begin;
 
+  // Move future ranges to start right after current ranges.
+  // Otherwise future ranges will grow forward indefinitely.
   if (fut_beg == count) {
     fut_beg = count = cur_end;
   } else if (fut_beg != cur_end) {
@@ -528,6 +530,7 @@ static void decor_range_insert(DecorState *state, DecorRange *range)
   range->ordering = state->new_range_ordering++;
 
   int index;
+  // Get space for a new `DecorRange` from the freelist or allocate.
   if (state->free_slot_i >= 0) {
     index = state->free_slot_i;
     DecorRangeSlot *slot = &kv_A(state->slots, index);
@@ -635,11 +638,11 @@ void decor_init_draw_col(int win_col, bool hidden, DecorRange *item)
 
 void decor_recheck_draw_col(int win_col, bool hidden, DecorState *state)
 {
-  int const count = state->current_end;
+  int const end = state->current_end;
   int *const indices = state->ranges_i.items;
   DecorRangeSlot *const slots = state->slots.items;
 
-  for (int i = 0; i < count; i++) {
+  for (int i = 0; i < end; i++) {
     DecorRange *const r = &slots[indices[i]].range;
     if (r->draw_col == -3) {
       decor_init_draw_col(win_col, hidden, r);
@@ -683,6 +686,7 @@ next_mark:
   int cur_end = state->current_end;
   int fut_beg = state->future_begin;
 
+  // Promote future ranges before the cursor to active.
   for (; fut_beg < count; fut_beg++) {
     int const index = indices[fut_beg];
     DecorRange *const r = &slots[index].range;
@@ -711,7 +715,7 @@ next_mark:
     cur_end++;
   }
 
-  if (fut_beg != count) {
+  if (fut_beg < count) {
     DecorRange *r = &slots[indices[fut_beg]].range;
     if (r->start_row == row) {
       col_until = MIN(col_until, r->start_col - 1);
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -314,7 +314,8 @@ void decor_check_to_be_deleted(void)
 
 void decor_state_free(DecorState *state)
 {
-  kv_destroy(state->active);
+  kv_destroy(state->slots);
+  kv_destroy(state->ranges_i);
 }
 
 void clear_virttext(VirtText *text)
@@ -399,14 +400,30 @@ bool decor_redraw_reset(win_T *wp, DecorState *state)
 {
   state->row = -1;
   state->win = wp;
-  for (size_t i = 0; i < kv_size(state->active); i++) {
-    DecorRange item = kv_A(state->active, i);
-    if (item.owned && item.kind == kDecorKindVirtText) {
-      clear_virttext(&item.data.vt->data.virt_text);
-      xfree(item.data.vt);
+
+  int *const indices = state->ranges_i.items;
+  DecorRangeSlot *const slots = state->slots.items;
+
+  int const beg_pos[] = { 0, state->future_begin };
+  int const end_pos[] = { state->current_end, (int)kv_size(state->ranges_i) };
+
+  for (int pos_i = 0; pos_i < 2; pos_i++) {
+    for (int i = beg_pos[pos_i]; i < end_pos[pos_i]; i++) {
+      DecorRange *const r = &slots[indices[i]].range;
+      if (r->owned && r->kind == kDecorKindVirtText) {
+        clear_virttext(&r->data.vt->data.virt_text);
+        xfree(r->data.vt);
+      }
     }
   }
-  kv_size(state->active) = 0;
+
+  kv_size(state->slots) = 0;
+  kv_size(state->ranges_i) = 0;
+  state->free_slot_i = -1;
+  state->current_end = 0;
+  state->future_begin = 0;
+  state->new_range_ordering = 0;
+
   return wp->w_buffer->b_marktree->n_keys;
 }
 
@@ -452,6 +469,23 @@ bool decor_redraw_start(win_T *wp, int top_row, DecorState *state)
 
 bool decor_redraw_line(win_T *wp, int row, DecorState *state)
 {
+  int count = (int)kv_size(state->ranges_i);
+  int const cur_end = state->current_end;
+  int fut_beg = state->future_begin;
+
+  if (fut_beg == count) {
+    fut_beg = count = cur_end;
+  } else if (fut_beg != cur_end) {
+    int *const indices = state->ranges_i.items;
+    memmove(indices + cur_end, indices + fut_beg, (size_t)(count - fut_beg) * sizeof(indices[0]));
+
+    count = cur_end + (count - fut_beg);
+    fut_beg = cur_end;
+  }
+
+  kv_size(state->ranges_i) = (size_t)count;
+  state->future_begin = fut_beg;
+
   if (state->row == -1) {
     decor_redraw_start(wp, row, state);
   }
@@ -459,7 +493,7 @@ bool decor_redraw_line(win_T *wp, int row, DecorState *state)
   state->col_until = -1;
   state->eol_col = -1;
 
-  if (kv_size(state->active)) {
+  if (cur_end != 0 || fut_beg != count) {
     return true;
   }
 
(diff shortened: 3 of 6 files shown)


vanaigr:
`draw_virt_text()` and `handle_inline_virtual_text()` now iterate only through active ranges. They don't check the range position to know if it's active, but they check for `draw_col` being a value other than -10, and it seems that all inactive ranges have `draw_col` == -10 and it is set to a different value only when the range becomes active:
https://github.com/neovim/neovim/blob/4833a63929a83c11379036ead33913711134015b/src/nvim/decoration.c#L661-L664

vanaigr:
The `test / ubuntu asan clang oldtest (pull_request)` failed probably because I took address of the first element of an empty array:
```c
#define kv_A(v, i) ((v).items[(i)])
...
DecorRangeSlot *const slots = &kv_A(state->slots, 0);
```

Can I directly access the `items` of the `kvect_t` structs?

zeertzjq:
> The `test / ubuntu asan clang oldtest (pull_request)` failed likely because I took address of the first element of an empty array:
> 
> ```c
> #define kv_A(v, i) ((v).items[(i)])
> ...
> DecorRangeSlot *const slots = &kv_A(state->slots, 0);
> ```
> 
> Can I directly access the `items` of the `kvect_t` structs?

`items` of empty `kvec_t` is `NULL`

vanaigr:
> > The `test / ubuntu asan clang oldtest (pull_request)` failed likely because I took address of the first element of an empty array:
> > ```c
> > #define kv_A(v, i) ((v).items[(i)])
> > ...
> > DecorRangeSlot *const slots = &kv_A(state->slots, 0);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Can I directly access the `items` of the `kvect_t` structs?
> 
> `items` of empty `kvec_t` is `NULL`

Yes, but it reads nothing from the array if it is empty.

zeertzjq:
Yes, but ASAN doesn't allow you to use `[0]` on a NULL pointer. You can assign `->items` to the pointer directly instead.

src/nvim/decoration.h
```diff
@@ -76,11 +91,22 @@ typedef struct {
...
 EXTERN kvec_t(DecorSignHighlight) decor_items INIT( = KV_INITIAL_VALUE);
 
 #ifdef INCLUDE_GENERATED_DECLARATIONS
 # include "decoration.h.generated.h"
 #endif
+
+inline int decor_redraw_col(win_T *wp, int col, int win_col, bool hidden, DecorState *state)
+  REAL_FATTR_ALWAYS_INLINE;
+
+inline int decor_redraw_col(win_T *wp, int col, int win_col, bool hidden, DecorState *state)
```
zeertzjq, line 106:
```suggestion
# include "decoration.h.inline.generated.h"
#endif

inline int decor_redraw_col(win_T *wp, int col, int win_col, bool hidden, DecorState *state)
  FUNC_ATTR_ALWAYS_INLINE
```

src/nvim/decoration.h
```diff
@@ -57,9 +58,23 @@ typedef struct {
...
 typedef struct {
   MarkTreeIter itr[1];
-  kvec_t(DecorRange) active;
+  kvec_t(DecorRangeSlot) slots;
+  kvec_t(int) ranges_i;
+  /// Indices in [0; current_end) range of `ranges_i` point to ranges that start before the
+  /// current position and are sorted by priority and order of insertion.
+  int current_end;
+  /// Indices in [future_begin, kv_size(ranges_i)) range of `ranges_i` point to ranges that
+  /// Start after the current position and are sorted by starting position.
```
zeertzjq, line 74:
```suggestion
  /// start after the current position and are sorted by starting position.
```

justinmk:
If you have benchmark code available, you may want to add it somewhere in `test/benchmark/`.

Please add a sentence in `:help news-features`, there currently a lonely "Performance" section there :) Something like `Redraw time for ... reduced by 400x` is something users love to see.

vanaigr:
I added the benchmarks and news.
image
image

src/nvim/decoration.h
vanaigr: I made the function `static inline` because otherwise "decoration.h.inline.generated.h" is empty and linking fails with unresolved symbol error.

lewis6991 (commented):
Can you add some comment around the blocks explaining their purpose, as the whole slot system isn't particularly straight forward.

vanaigr:
I noticed that the original behavior was not always consistent. Normal extmarks can be added both before and after ephemeral extmarks. For example, with this lua highlights file: 
```
(function_declaration) @testing1
```

and this test file:
```lua
----
function ignore() end
function test()
local a = 5
  if a > 10 then
    print('something')
    a = a + 10
  end
end

vim.api.nvim_set_hl(0, '@testing1', { fg = 'green' })
vim.api.nvim_set_hl(0, 'testing2', { fg = 'blue', italic = true })
local ns = vim.api.nvim_create_namespace('testing')
vim.api.nvim_buf_clear_namespace(0, ns, 0, -1)
vim.api.nvim_buf_set_extmark(0, ns, 0, 0, {
  end_row = 5,
  end_col = 1,
  priority = 100,
  hl_group = 'testing2',
})
```

After sourcing the file, it looks like this:
image
But after updating line 3 (for example by jumping with `cursorline`), it changes the highlight:
image

And it can be reverted back by updating the first line at the same time (for example with `:redraw!` or visual selection):
image
image

vanaigr:
Rebased and added a comment that info about freed `DecorRange`s is stored in a freelist

src/nvim/decoration.h
```diff
@@ -57,9 +58,28 @@ typedef struct {
...
+  /// Indices in [0; current_end) of `ranges_i` point to ranges that start
+  /// before current position. Sorted by priority and order of insertion.
+  int current_end;
+  /// Indices in [future_begin, kv_size(ranges_i)) of `ranges_i` point to
+  /// ranges that start after current position. Sorted by starting position.
+  int future_begin;
+  /// Head of DecorRangeSlot freelist. -1 if none are freed.
+  int free_slot_i;
+  /// Index for keeping track of range insertion order.
+  int new_range_ordering;
```
justinmk, line 82: Looks like 4+ new fields were added, and a lot of related logic. Is there any way to extract parts of the logic to common functions, to help with understanding it in the future? Or some other way to simplify the code?

vanaigr:
@bfredl

bfredl:
The decor provider concept is going to need to be re-engineered to support long lines "natively". The rationale for the current `on_line` entry point and the quite naive `kvec_t(DecorRange)` is that there is a limited amount of work and resulting marks per line, which of course flies out of the window if the rest of the editor is refactored to allow really really long lines without exploding (think opening a 10MB log.json file without linebreaks directly). The new API for 0.12 is going to need to replace it with `on_range` for arbitrary ranges of text (which both can be multi-line when we have trivially short lines, and multi-callback per line for looong lines) to be prepared for this possible future.

The other end of the design spectrum would be to replace `kvec_t(DecorState)` with a separate `MarkTree(extmark)` for ephemeral decoration. I don't think that should be done just yet, but that would be the natural end point of making DecorState "smarter" as a solution, if we end up concluding that will be necessary despite these planned API improvements.   

That being said, if this micro-optimization does make end-user experience noticeably better already, we _should_ merge this PR for 0.11 pretty much as is. The fact it contains a benchmark is a huge plus and something which will remain even if the implementation is changed later.

vanaigr:
Rebased.

I couldn't find anything that I could factor out since only thing that is repeated multiple times in iteration bounds.

lewis6991:
It's not obvious to me what a lot of the new blocks are doing without studying them thoroughly. Could you at least add some comments on what's going on and the rationale for it.

vanaigr:
Anything else?

runtime/doc/news.txt
```diff
@@ -247,7 +247,7 @@ OPTIONS
 
 PERFORMANCE
 
-• TODO
+• Significantly reduced redraw time for long lines with syntax highlighting.
```
bfredl, line 250: "syntax" -> "treesitter"

src/nvim/decoration.h
```diff
@@ -34,6 +34,10 @@ typedef struct {
   int start_col;
   int end_row;
   int end_col;
+  int ordering;
```
bfredl, line 37: `///< range insertion order`

src/nvim/decoration.h
```diff
@@ -34,6 +34,10 @@ typedef struct {
...
+  DecorPriority priority;
+  bool owned;  ///< ephemeral decoration, free memory immediately
+  DecorRangeKind kind: 8;
```
bfredl, line 40: We don't use bitfields in our owned code, better to change the typedef to `typedef uint8_t DecorRangeKind`

bfredl (approved):

runtime/doc/news.txt
vanaigr: Are ephemeral decorations primarily used by treesitter?

clason: We are pretty consistent in our docs in distinguishing (legacy) "syntax highlighting" from "treesitter highlighting". So this is more about consistent user documentation than technical correctness.

Issue #8587, opened 2018-06-18
"E138: main.shada.tmp.X files exist, cannot write ShaDa file!" on close.


jcjolley:
- `nvim --version`:
```
NVIM v0.3.0
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: C:/msys64/mingw64/bin/gcc.exe -Wconversion -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -O2 -g -DMI
N_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough
 -D__USE_MINGW_ANSI_STDIO -D_WIN32_WINNT=0x0600 -Wvla -fdiagnostics-color=auto -Wno-array-bounds -DINCLUDE_GENERATED_DEC
LARATIONS -IC:/projects/neovim/build/config -IC:/projects/neovim/src -IC:/projects/neovim/.deps/usr/include -IC:/msys64/
mingw64/include -IC:/projects/neovim/build/src/nvim/auto -IC:/projects/neovim/build/include
Compiled by appveyor@APPVYR-WIN


Features: -acl +iconv -jemalloc +tui
See ":help feature-compile"

   system vimrc file: "$VIM\sysinit.vim"
  fall-back for $VIM: "C:/Program Files (x86)/nvim/share/nvim"

Run :checkhealth for more info
```
- Operating system/version: Windows 10 Pro 1703 15063.1115
- Terminal name/version:
- `$TERM`:

### Steps to reproduce using `nvim -u NORC`

```
nvim -u NORC
:q
```

### Actual behaviour
![nvime138repro](https://user-images.githubusercontent.com/2086176/41547037-614b1a5a-72dc-11e8-8d70-e1ed595e3d9e.gif)

### Expected behaviour
Close without error.

ZyX-I:
This is the expected behaviour if you somehow generated all temporary files. You should’ve had a bunch of errors explaining why those temporary files could not be renamed to replace the `main.shada` file before having this error. On linux that would be most likely some permission issue (e.g. `main.shada` was created by root so Neovim could not overwrite it), not sure about Windows. You must remove `main.shada`, rename most recent of the temporary files to `main.shada`, remove other temporary files and solve the issue that caused generation of temporary files.

janlazo:
[...]

PR #30872, opened 2024-10-20
fix(exit): close memfiles after processing events

### LINKED ISSUE #30867: `Internal error: qf_fill_buffer()` when updating quickfix list inside of `DiagnosticChanged` callback
### Problem

I'm trying to keep my quickfix list in sync with Neovim's diagnostics through a `DiagnosticChanged` autocmd that calls `vim.diagnostic.setqflist`. This mostly works, but when closing Neovim,I get the following error:

```
Error executing vim.schedule lua callback: Vim:E685: Internal error: qf_fill_buffer()
stack traceback:
        [C]: in function 'setqflist'
        ...wrapped-0.10.2/share/nvim/runtime/lua/vim/diagnostic.lua:808: in function 'set_list'
        ...wrapped-0.10.2/share/nvim/runtime/lua/vim/diagnostic.lua:1963: in function 'setqflist'
        ...eremy/sync/scratch/jfly/2024-10-18-vim-crash/minimal.lua:43: in function <...eremy/sync/scratch/jfly/2024-10-18-vim-crash/minimal.lua:41>
```

It feels like this callback is getting invoked sometime during Vim's shutdown lifecycle when the quickfix list can no longer be written to?

### Steps to reproduce

<details>
<summary>minimal.lua</summary>

```lua
-- Run this file as `nvim --clean -u ./minimal.lua hgttg.py

for name, url in pairs {
  nonels = 'https://github.com/nvimtools/none-ls.nvim',
  plenary = 'https://github.com/nvim-lua/plenary.nvim.git', -- none-ls depends on this
  
  lspconfig = 'https://github.com/neovim/nvim-lspconfig',
} do
  end
end

-- ADD INIT.LUA SETTINGS _NECESSARY_ FOR REPRODUCING THE ISSUE

vim.lsp.set_log_level("debug")

require('lspconfig').ruff.setup {}

local null_ls = require("null-ls")
null_ls.setup({
    debug = true,
    sources = {
      null_ls.builtins.formatting.black,
    },
})

vim.api.nvim_create_autocmd('DiagnosticChanged', {
  callback = function(args)
    local window_height = vim.api.nvim_win_get_height(0)
    for i = 1, window_height do
        vim.print(i .. "/" .. window_height .. " filling up the screen")
    end
    vim.print(vim.inspect(args))

    vim.schedule(function()
      vim.diagnostic.setqflist({ open = false })
    end)
  end,
})

vim.cmd "copen"
```
</details>

<details>
<summary>hgttg.py</summary>

```
print(f"don't panic")
print(f"don't panic")
print(f"don't panic"  )
print(f"don't panic")
print(f"don't panic")

```
</details>

Run with both `ruff` and `black` on your PATH:

```
nvim --clean -u ./minimal.lua hgttg.py
```

This will open Neovim with a quickfix list that looks something like this:

Image

Then, close Neovim with `:qa`. Scroll down through the output until you see the error:

Image

<hr>

Sorry for such a complicated "minimal" repo. I've spent many hours trying to simplify this, and just don't have a good enough mental model of Neovim's exit lifecycle to simplify this. Happy to investigate more if folks point me in the right direction.

### Expected behavior

I would expect Neovim to not crash here. In other words: ideally the quickfix buffer would be writeable whenever `DiagnosticChanged` events fire. If that's not feasible, I'd love to know what the right way is to check if the quickfix list is writeable, I haven't been able to find anything that works.

### Nvim version (nvim -v)

I can repro this issue on both `master` (4833a63929a83c11379036ead33913711134015b at time of writing) and Neovim 0.10.2

N/A, this is LSP specific

### Operating system/version

NixOS

### Terminal name/version

Alacritty

alacritty

### Installation

both build from repo and nixpkgs

zeertzjq:
Fix #30867

Problem:  When exiting, processed events may still use memfiles after
          they are closed.
Solution: Close memfiles after processing events.

### DIFF
--- a/src/nvim/main.c
+++ b/src/nvim/main.c
@@ -678,11 +678,13 @@ void os_exit(int r)
   } else {
     ui_flush();
     ui_call_stop();
-    ml_close_all(true);           // remove all memfiles
   }
 
   if (!event_teardown() && r == 0) {
     r = 1;  // Exit with error if main_loop did not teardown gracefully.
+  }
+  if (!ui_client_channel_id) {
+    ml_close_all(true);  // remove all memfiles
   }
   if (used_stdin) {
     stream_set_blocking(STDIN_FILENO, true);  // normalize stream (#2598)


zeertzjq:
I tried to add a test for the swapfile case, but even though I can reproduce leaving swapfiles interactively, I can't reproduce it in a test.

justinmk:
> But maybe it's worth doing this anyway, as this prevents leaving swapfiles when events processed during exit replace or add lines in a buffer.

I'm wondering if something like that is the root cause of https://github.com/neovim/neovim/issues/8587#issuecomment-2321938404

zeertzjq:
I tried to add a test for the quickfix internal error, but it didn't work either.

src/nvim/main.c
```diff
@@ -678,12 +678,14 @@ void os_exit(int r)
...
     ui_flush();
     ui_call_stop();
-    ml_close_all(true);           // remove all memfiles
   }
 
   if (!event_teardown() && r == 0) {
     r = 1;  // Exit with error if main_loop did not teardown gracefully.
   }
+  if (!ui_client_channel_id) {
+    ml_close_all(true);           // remove all memfiles
```
justinmk, line 687: Is the `ui_client_channel_id` check because we never initialized memfiles for that case?

justinmk, line 687:
```suggestion
    ml_close_all(true);  // remove all memfiles
```

justinmk (approved):

GitMurf:
Just commenting here as an fyi as it was something I finally tracked down earlier today (potential hunch) and it also seemed to relate to swap files (and I saw other reports of similar hunches online). That is that sometimes when I quit neovim with `:qa` it exits to my terminal but the screen stays "painted" with the neovim buffers I had on screen last and I have to do a "clear" in my terminal.

Today I finally got some ideas on it and in my testing it happened most often when:

1. Had shada tmp files full.
2. Interestingly enough my main shada file got pretty large (500kb) and stopped updating even after I cleared the temp files out.
3. This is more subjective, but it seemed to happen more often when using a loaded mini.session (plugin) saved session which has a "auto write / save session on exit" feature.

It seems like this could be related as when I deleted my main shada file that was growing larger and seemed "stuck", it stopped happening (nvim staying rendered / painted after quit) for the time being.

For mini.sessions it could make sense based on the details in this PR that something is happening "weirdly" during quit when the sessions are getting auto saves. I presume it is hooking into events during exit, and from the sounds of this PR, that could definitely be relevant here and explain some of my "problems".

justinmk:
@GitMurf I see that you also commented on https://github.com/neovim/neovim/issues/8587#issuecomment-2321938404 , maybe those details are relevant there.

GitMurf:
> @GitMurf I see that you also commented on https://github.com/neovim/neovim/issues/8587#issuecomment-2321938404 , maybe those details are relevant there.

I had figured I commented enough there with my issues with shada historically ;) but this was more of an observation in relation to the specifics about nvim closing and conflicts in the timing of events and memfiles.

But I agree it doesn't hurt to mention the information in the issue that is still active :) thanks. I'll paste it over there.

PR #30876, opened 2024-10-20
refactor(lsp): create `vim.g/b.inlay_hint_enabled`

ofseed:
## Problem
In https://github.com/neovim/neovim/pull/28543#discussion_r1587440644, we discussed that we should make more use of vim variables like `vim.b` to store variables related to the lifecycle of buffers, but using it to store tables may have potential risks (https://github.com/neovim/neovim/issues/12544). However, there is no problem using `enable`(the variable presents whether the inlay hint is enabled) in `globalstate/bufstate`, as it is just a boolean value, and doing so has two advantages.
1. This provides a configuration method for global/local variables that aligns with the editor options style and does not disrupt our current `enable` strategy (related: https://github.com/neovim/neovim/issues/28603).
2. `enable` is a user-controlled state, while other variables managed by `bufstate` (such as `client_hints` and `version`) are controlled by LSP. They have different lifecycles, and separating them will better facilitate the next step of refactoring.
## Solution
Create two new variables `vim.g.inlay_hint_enabled` and `vim.b.inlay_hint_enabled`

### DIFF
--- a/runtime/lua/vim/lsp/inlay_hint.lua
+++ b/runtime/lua/vim/lsp/inlay_hint.lua
@@ -69,21 +69,10 @@ function M.on_inlayhint(err, result, ctx, _)
     return
   end
 
-  local lines = api.nvim_buf_get_lines(bufnr, 0, -1, false)
-  ---@param position lsp.Position
-  ---@return integer
-  local function pos_to_byte(position)
-    local col = position.character
-    if col > 0 then
-      local line = lines[position.line + 1] or ''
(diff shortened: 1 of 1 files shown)


ofseed:
cc @justinmk I would like to know if you think this solution is appropriate.

justinmk:
Single-purpose variables aren't the right direction. Using `vim.g` / `vim.b` for the full "config" of a topic could definitely makes sense, but requires https://github.com/neovim/neovim/issues/12544#issuecomment-1116794687 first...

PR #30917, opened 2024-10-23
refactor(options): always allocate option values

famiu:
Instead of keeping `P_ALLOCED` and `P_DEF_ALLOCED` flags to check if an option value is allocated, always allocate option values to simplify the logic.

Ref: #25672

### DIFF
--- a/src/nvim/option.c
+++ b/src/nvim/option.c
@@ -291,8 +291,7 @@ static void set_init_default_cdpath(void)
     }
   }
   buf[j] = NUL;
-  options[kOptCdpath].def_val = CSTR_AS_OPTVAL(buf);
-  options[kOptCdpath].flags |= P_DEF_ALLOCED;
+  change_option_default(kOptCdpath, CSTR_AS_OPTVAL(buf));
 
   xfree(cdpath);
 }
@@ -302,8 +301,6 @@ static void set_init_default_cdpath(void)
 /// only happen for non-indirect options.
 /// Also set the default to the expanded value, so ":set" does not list
 /// them.
-/// Don't set the P_ALLOCED flag, because we don't want to free the
-/// default.
 static void set_init_expand_env(void)
 {
   for (OptIndex opt_idx = 0; opt_idx < kOptIndexCount; opt_idx++) {
@@ -318,14 +315,8 @@ static void set_init_expand_env(void)
       p = option_expand(opt_idx, NULL);
     }
     if (p != NULL) {
-      set_option_varp(opt_idx, opt->var, CSTR_TO_OPTVAL(p), opt->flags & P_ALLOCED);
-      opt->flags |= P_ALLOCED;
-
-      if (opt->flags & P_DEF_ALLOCED) {
-        optval_free(opt->def_val);
-      }
-      opt->def_val = CSTR_TO_OPTVAL(p);
-      opt->flags |= P_DEF_ALLOCED;
+      set_option_varp(opt_idx, opt->var, CSTR_TO_OPTVAL(p), true);
+      change_option_default(opt_idx, CSTR_TO_OPTVAL(p));
     }
   }
 }
@@ -355,6 +346,9 @@ static void set_init_fenc_default(void)
 void set_init_1(bool clean_arg)
 {
   langmap_init();
+
+  // Allocate the default option values.
+  alloc_options_default();
 
   set_init_default_shell();
   set_init_default_backupskip();
@@ -463,6 +457,25 @@ static OptVal get_option_default(const OptIndex opt_idx, int opt_flags)
   }
 }
 
+/// Allocate the default values for all options by copying them from the stack.
+/// This ensures that we don't need to always check if the option default is allocated or not.
+static void alloc_options_default(void)
+{
+  for (OptIndex opt_idx = 0; opt_idx < kOptIndexCount; opt_idx++) {
+    options[opt_idx].def_val = optval_copy(options[opt_idx].def_val);
+  }
+}
+
+/// Change the default value for an option.
+///
+/// @param  opt_idx  Option index in options[] table.
+/// @param  value    New default value. Must be allocated.
+static void change_option_default(const OptIndex opt_idx, OptVal value)
+{
+  optval_free(options[opt_idx].def_val);
+  options[opt_idx].def_val = value;
+}
+
 /// Set an option to its default value.
 /// This does not take care of side effects!
 ///
@@ -512,17 +525,8 @@ static void set_options_default(int opt_flags)
 static void set_string_default(OptIndex opt_idx, char *val, bool allocated)
   FUNC_ATTR_NONNULL_ALL
 {
-  if (opt_idx == kOptInvalid) {
-    return;
-  }
-
-  vimoption_T *opt = &options[opt_idx];
-  if (opt->flags & P_DEF_ALLOCED) {
-    optval_free(opt->def_val);
-  }
-
-  opt->def_val = CSTR_AS_OPTVAL(allocated ? val : xstrdup(val));
-  opt->flags |= P_DEF_ALLOCED;
+  assert(opt_idx != kOptInvalid);
+  change_option_default(opt_idx, CSTR_AS_OPTVAL(allocated ? val : xstrdup(val)));
 }
 
 /// For an option value that contains comma separated items, find "newval" in
@@ -563,16 +567,14 @@ void free_all_options(void)
   for (OptIndex opt_idx = 0; opt_idx < kOptIndexCount; opt_idx++) {
     if (options[opt_idx].indir == PV_NONE) {
       // global option: free value and default value.
-      if ((options[opt_idx].flags & P_ALLOCED) && options[opt_idx].var != NULL) {
+      if (options[opt_idx].var != NULL) {
         optval_free(optval_from_varp(opt_idx, options[opt_idx].var));
-      }
-      if (options[opt_idx].flags & P_DEF_ALLOCED) {
-        optval_free(options[opt_idx].def_val);
       }
     } else if (options[opt_idx].var != VAR_WIN) {
       // buffer-local option: free global value
       optval_free(optval_from_varp(opt_idx, options[opt_idx].var));
     }
+    optval_free(options[opt_idx].def_val);
   }
   free_operatorfunc_option();
   free_tagfunc_option();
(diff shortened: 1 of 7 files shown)


src/nvim/option.c
```diff
@@ -463,12 +457,31 @@ static OptVal get_option_default(const OptIndex opt_idx, int opt_flags)
...
+  options[opt_idx].def_val = value;
+}
+
 /// Set an option to its default value.
 /// This does not take care of side effects!
 ///
 /// @param  opt_idx    Option index in options[] table.
 /// @param  opt_flags  Option flags.
-static void set_option_default(const OptIndex opt_idx, int opt_flags)
+static void set_option_to_default(const OptIndex opt_idx, int opt_flags)
```
justinmk, line 484: may be confusing for little gain to redefine (line 473 above) the meaning of `set_option_default` (compared to Vim). 

instead we could keep that name here, and use a name like `def_option_default` ("define").

src/nvim/option.c
```diff
@@ -512,17 +525,8 @@ static void set_options_default(int opt_flags)
 static void set_string_default(OptIndex opt_idx, char *val, bool allocated)
```
justinmk, line 525: is this defining/initializing the default or "setting the option to its default" ?

src/nvim/option_vars.h
```diff
@@ -8,49 +8,43 @@
 // option_vars.h: definition of global variables for settable options
 
 // Option Flags
-#define P_ALLOCED      0x01U        ///< the option is in allocated memory,
-                                    ///< must use free_string_option() when
-                                    ///< assigning new value. Not set if default is
-                                    ///< the same.
-#define P_EXPAND       0x02U        ///< environment expansion.  NOTE: P_EXPAND can
+#define P_EXPAND       0x01U        ///< environment expansion.  NOTE: P_EXPAND can
```
justinmk, line 11: is it really required to bump all these flags or could we keep them as-is and just comment-out the P_ALLOCED line and explain that it's unused

justinmk (approved):

src/nvim/option.c
famiu: It's changing the default value of an option, so the former

src/nvim/option_vars.h
famiu: It's not required but it can be useful to free up flag space as we can only have up to 32 flags
Let me know if I should revert this change

src/nvim/option.c
famiu: On second thought, `modify_option_default` seems like a better name

src/nvim/option_vars.h
justinmk: > plan to rework the whole option flag system in the future anyway to no longer use macros

it's probably worth avoiding the churn here, until then.

famiu: Should be fine now

src/nvim/option_vars.h
```diff
@@ -8,16 +8,18 @@
...
 // Option Flags
-#define P_ALLOCED      0x01U        ///< the option is in allocated memory,
-                                    ///< must use free_string_option() when
-                                    ///< assigning new value. Not set if default is
-                                    ///< the same.
+// #define P_ALLOCED      0x01U        ///< the option is in allocated memory,
+//                                     ///< must use free_string_option() when
+//                                     ///< assigning new value. Not set if default is
+//                                     ///< the same.
+//                                     ///< No longer used since option value is always allocated.
```
justinmk, line 15:
```suggestion
// #define P_ALLOCED      0x01U        ///< Not used. #30917
```

src/nvim/option.c
```diff
@@ -3518,11 +3517,13 @@ static const char *did_set_option(OptIndex opt_idx, void *varp, OptVal old_value
   }
 
   // Free options that are in allocated memory.
-  // Use "free_oldval", because recursiveness may change the flags (esp. init_highlight()).
-  if (free_oldval) {
+  // NOTE: When initializing an option for the first time, don't try to free its value as it has
+  // not been allocated yet.
+  if (opt_initialized[opt_idx]) {
```
zeertzjq, line 3522: Is this check really needed? If an option hasn't been allocated, freeing it is no-op.

famiu: Yes this check is needed. Using `free` on unallocated memory is in fact not no-op and will cause a segfault

zeertzjq: What unallocated memory?

zeertzjq: Oh, there is a single option: `p_vfile`

zeertzjq: It works if you use `empty_string_option` for that:
```diff
diff --git a/src/nvim/globals.h b/src/nvim/globals.h
index 45a31ce6b0..1cdb38f5f8 100644
--- a/src/nvim/globals.h
+++ b/src/nvim/globals.h
@@ -713,12 +713,6 @@ EXTERN char *escape_chars INIT( = " \t\\\"|");  // need backslash in cmd line
 
 EXTERN bool keep_help_flag INIT( = false);  // doing :ta from help file
 
-// When a string option is NULL (which only happens in out-of-memory situations), it is set to
-// empty_string_option, to avoid having to check for NULL everywhere.
-//
-// TODO(famiu): Remove this when refcounted strings are used for string options.
-EXTERN char *empty_string_option INIT( = "");
-
 EXTERN bool redir_off INIT( = false);        // no redirection for a moment
 EXTERN FILE *redir_fd INIT( = NULL);         // message redirection file
 EXTERN int redir_reg INIT( = 0);             // message redirection register
diff --git a/src/nvim/option.c b/src/nvim/option.c
index a0f58eaa16..72f8a5d2f3 100644
--- a/src/nvim/option.c
+++ b/src/nvim/option.c
@@ -3442,7 +3442,6 @@ static const char *did_set_option(OptIndex opt_idx, void *varp, OptVal old_value
   bool restore_chartab = false;
   bool value_changed = false;
   bool value_checked = false;
-  static bool opt_initialized[kOptIndexCount] = { 0 };
 
   optset_T did_set_cb_args = {
     .os_varp = varp,
@@ -3516,14 +3515,7 @@ static const char *did_set_option(OptIndex opt_idx, void *varp, OptVal old_value
     set_option_sctx(opt_idx, opt_flags, script_ctx);
   }
 
-  // Free options that are in allocated memory.
-  // NOTE: When initializing an option for the first time, don't try to free its value as it has
-  // not been allocated yet.
-  if (opt_initialized[opt_idx]) {
-    optval_free(old_value);
-  } else {
-    opt_initialized[opt_idx] = true;
-  }
+  optval_free(old_value);
 
   const bool scope_both = (opt_flags & (OPT_LOCAL | OPT_GLOBAL)) == 0;
 
diff --git a/src/nvim/option_vars.h b/src/nvim/option_vars.h
index bf72d4a18a..a4a7b8aaa8 100644
--- a/src/nvim/option_vars.h
+++ b/src/nvim/option_vars.h
@@ -350,6 +350,12 @@ enum {
 #define LISPWORD_VALUE \
   "defun,define,defmacro,set!,lambda,if,case,let,flet,let*,letrec,do,do*,define-syntax,let-syntax,letrec-syntax,destructuring-bind,defpackage,defparameter,defstruct,deftype,defvar,do-all-symbols,do-external-symbols,do-symbols,dolist,dotimes,ecase,etypecase,eval-when,labels,macrolet,multiple-value-bind,multiple-value-call,multiple-value-prog1,multiple-value-setq,prog1,progv,typecase,unless,unwind-protect,when,with-input-from-string,with-open-file,with-open-stream,with-output-to-string,with-package-iterator,define-condition,handler-bind,handler-case,restart-bind,restart-case,with-simple-restart,store-value,use-value,muffle-warning,abort,continue,with-slots,with-slots*,with-accessors,with-accessors*,defclass,defmethod,print-unreadable-object"
 
+// When a string option is NULL, it is set to empty_string_option,
+// to avoid having to check for NULL everywhere.
+//
+// TODO(famiu): Remove this when refcounted strings are used for string options.
+EXTERN char empty_string_option[] INIT( = "");
+
 // The following are actual variables for the options
 
 EXTERN char *p_ambw;             ///< 'ambiwidth'
@@ -772,7 +778,7 @@ EXTERN unsigned ve_flags;
 #define VE_NONEU       32U      // "NONE"
 EXTERN OptInt p_verbose;        ///< 'verbose'
 #ifdef IN_OPTION_C
-char *p_vfile = "";             ///< used before options are initialized
+char *p_vfile = empty_string_option;  ///< used before options are initialized
 #else
 extern char *p_vfile;           ///< 'verbosefile'
 #endif
```

famiu: @zeertzjq  I wonder how future-proof this is, like whether this would work if we add Callback, Array and Dictionary types for options in the future

justinmk: surely those would be initialized to null?

zeertzjq (approved):

justinmk:
coverity:

```
*** CID 511434:  Memory - corruptions  (BAD_FREE)
/src/nvim/option.c: 5369 in buf_copy_options()
5363               tabstop_set(p_vts, &buf->b_p_vts_array);
5364             } else {
5365               buf->b_p_vts_array = NULL;
5366             }
5367             buf->b_help = false;
5368             if (buf->b_p_bt[0] == 'h') {
>>>     CID 511434:  Memory - corruptions  (BAD_FREE)
>>>     "clear_string_option" frees incorrect pointer "buf->b_p_bt".
5369               clear_string_option(&buf->b_p_bt);
5370             }
5371             buf->b_p_ma = p_ma;
5372             COPY_OPT_SCTX(buf, BV_MA);
5373           }
5374         }

** CID 511433:  Memory - illegal accesses  (OVERRUN)
/src/nvim/optionstr.c: 322 in check_signcolumn()


________________________________________________________________________________________________________
*** CID 511433:  Memory - illegal accesses  (OVERRUN)
/src/nvim/optionstr.c: 322 in check_signcolumn()
316           wp->w_maxscwidth = val[5] - '0';
317         } else {  // auto
318           wp->w_minscwidth = 0;
319           wp->w_maxscwidth = 1;
320         }
321       } else {
>>>     CID 511433:  Memory - illegal accesses  (OVERRUN)
>>>     Overrunning array of 1 bytes at byte offset 5 by dereferencing pointer "val + 5".
322         if (strncmp(val, "auto:", 5) != 0
323             || strlen(val) != 8
324             || !ascii_isdigit(val[5])
325             || val[6] != '-'
326             || !ascii_isdigit(val[7])) {
327           return FAIL;
```

PR #30877, opened 2024-10-20
feat(lsp)!: support multiple clients in goto methods

mfussenegger:
Relates to:

- https://github.com/neovim/neovim/issues/30034
- https://github.com/neovim/neovim/issues/17712
- https://github.com/neovim/neovim/issues/16363

Closes:

- https://github.com/neovim/neovim/issues/26936 (but only provides bufnr
  and method)
- https://github.com/neovim/neovim/issues/22318

Might fix: https://github.com/neovim/neovim/issues/30737

Issue #26782, opened 2023-12-28
Lua: Interface for moving the cursor to a position


gpanders:
# Problem

Moving the cursor is a fundamental action in Nvim, and there are many different ways to do so, depending on context:

- `vim.api.nvim_win_set_cursor()` or `vim.fn.cursor()` are the low level "primitives"
- `vim.diagnostic.goto_next()` and `vim.diagnostic.goto_prev()` move between diagnostics
- https://github.com/neovim/neovim/pull/24499 proposed a method for moving between tree-sitter nodes
- `vim.lsp.util.jump_to_location` moves the cursor to a given position within a specific buffer

https://github.com/neovim/neovim/pull/24499#issuecomment-1658752117 mentions an abstract `vim.goto()` or `vim.cursor()` function that accepts any arbitrary "position" and moves the cursor to that location.

Related: https://github.com/neovim/neovim/issues/25509
Related: https://github.com/neovim/neovim/pull/26745

# Proposal

We cannot name the function `vim.goto()` because `goto` is a reserved keyword in LuaJIT. IMO `vim.cursor()` is not a great name because "cursor" is not a verb. So I propose `vim.jump()`: "jump" is a standard and common idiom in Vim terminology and is therefore already familiar with users.

`vim.jump()` accepts a [Position](https://github.com/neovim/neovim/issues/25509) as an argument and moves the cursor to that location (as well as an optional buffer number to support jumping between buffers, to support the LSP use case? Or is the buffer number included in the `Position`?)

`vim.jump()` can then be used for all of the use cases covered in the problem description:

- As the go-to "primitive" for basic cursor movement:

  ```lua
  -- Jump to the beginning of a buffer
  vim.jump({ 0, 0 })

  -- Jump to the last line of a buffer
  vim.jump({ -1, 0 })
  ```

- For use with `vim.diagnostic` and `vim.treesitter` APIs:

  ```lua
  -- Jump to next diagnostic, replaces vim.diagnostic.goto_next()
  vim.jump(vim.diagnostic.next(1))

  -- Jump to previous diagnostic, replaces vim.diagnostic.goto_prev()
  vim.jump(vim.diagnostic.next(-1))

  -- Jump to Tree-sitter node
  local node = vim.treesitter.get_node()
  vim.jump(node:pos())
  ```

- Could possibly be used to replace `vim.lsp.util.jump_to_location`, in service of https://github.com/neovim/neovim/issues/25272

clason:
[...]

PR #30927, opened 2024-10-24
fix(lsp): set tagstack on jump via goto methods

### LINKED ISSUE #30926: LSP: `vim.lsp.buf.definition` does not set tagstack
### Problem

Since commit 0083e03d6, `vim.lsp.buf.definition` does not seem to set the tag stack. This results in things like `C-o` being completely broken and jumping back to random places.

Seems this is also a known issue as mentioned in #30877 as well. 

### Steps to reproduce

With an LSP server like `clangd` or `rust-analyzer`, jump to a function definition with `vim.lsp.buf.definition` and then press `C-o`. Cursor does not come back to the initial location from where `vim.lsp.buf.definition` was triggered.

### Expected behavior

Jumping with `vim.lsp.buf.definition` and then pressing `C-o` should jump back to the initial location.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1035+g50e63c817

NA

### Operating system/version

Arch Linux

### Terminal name/version

alacritty 0.14.0 (22a44757)

tmux-256color

### Installation

Build from repo

mfussenegger:
Follow up to https://github.com/neovim/neovim/pull/30877
Fixes https://github.com/neovim/neovim/issues/30926

### DIFF
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -61,6 +61,9 @@ local function get_locations(method, opts)
     return
   end
   local win = api.nvim_get_current_win()
+  local from = vim.fn.getpos('.')
+  from[1] = bufnr
+  local tagname = vim.fn.expand('<cword>')
   local remaining = #clients
 
   ---@type vim.quickfix.entry[]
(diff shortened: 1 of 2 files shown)


SanchayanMaity:
Tested this and works now!

runtime/lua/vim/lsp/buf.lua
```diff
@@ -96,6 +99,13 @@ local function get_locations(method, opts)
       if #all_items == 1 then
         local item = all_items[1]
         local b = item.bufnr or vim.fn.bufadd(item.filename)
+
+        -- Save position in jumplist
+        vim.cmd("normal! m'")
+        -- Push a new item into tagstack
+        local tagstack = { { tagname = tagname, from = from } }
+        vim.fn.settagstack(vim.fn.win_getid(win), { items = tagstack }, 't')
+
```
mfussenegger, line 108: Note all this logic might be a candidate for vim.jump (https://github.com/neovim/neovim/issues/26782)

PR #27339, opened 2024-02-04
feat(lsp): completion side effects

MariaSolOs:
Closes #25714

This PR introduces `vim.lsp.completion` with the goal of supporting the completion behaviours (snippet expansion, execution of commands, application of text edits) specified by LSP.

To enable completions for a given buffer and client, users should call `vim.lsp.completion.enable(true, client_id, bufnr, opts)`. Currently the options parameter includes the `autotrigger` property, which when set will request completions after a trigger character has been typed.

This PR also sets new default keymaps for the snippet navigation:
- `<Tab>` will jump to the next tabstop if the snippet is active and jumpable forwards.
- `<S-Tab>` will jump to the previous tabstop if the snippet is active and jumpable backwards.

If these keys are being used in other keymaps, they will be restored once the snippet session ends.

PR #30936, opened 2024-10-24
feat(lsp)!: support filter and on_result opts for lsp.completion.get

mfussenegger:
Allows to implement custom completion lookups like the following, which
shows only snippets, and auto-expands if there is only a single match:

    ---@param item lsp.CompletionItem
    local function filter(_, item)
      local kind = vim.lsp.protocol.CompletionItemKind[item.kind] or ''
      return kind == 'Snippet'
    end

    local function on_result(startcol, matches)
      if #matches == 0 then
        return exit()
      else
        vim.fn.complete(startcol, matches)
        if #matches == 1 then
          api.nvim_feedkeys(
            api.nvim_replace_termcodes("<C-n>", true, false, true), 'n', true)
          api.nvim_feedkeys(
            api.nvim_replace_termcodes("<CR>", true, false, true), 'm', true)
        end
      end
    end
    vim.lsp.completion.get({ filter = filter, on_result = vim.schedule_wrap(on_result) })

---

I'm not sure about the rename

TODO:

- [ ] update remaining trigger references or revert rename
- [ ] maybe some additional tests?

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1637,6 +1637,14 @@ Lua module: vim.lsp.completion                                *lsp-completion*
       • {convert}?      (`fun(item: lsp.CompletionItem): table`) Transforms an
                         LSP CompletionItem to |complete-items|.
 
+*vim.lsp.completion.get.opts*
+
+    Fields: ~
+      • {convert}?    (`fun(item: lsp.CompletionItem): table`)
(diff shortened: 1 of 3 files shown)


justinmk:
For reference: [previous discussion](https://github.com/neovim/neovim/pull/27339#discussion_r1493820152) (not sure if it's exactly relevant here)

przepompownia:
It would be easier for me to read it if it was up to date.

PR #30061, opened 2024-08-15
KeyInputPre event (vim-patch: 9.1.0563, 9.1.0597, 890f97ce57e)

Shougo:

PR #30939, opened 2024-10-24
feat(lua): vim.on_key() can consume the key

errael:
Partial fix for #30741; option to prevent mapping not implemented.

`vim.on_key()` 3rd arg is opts. {may_discard=true} for `vim.on_key()` return to control discard.

The option to prevent mapping can be implemented. It is desirable that when `vim.on_key()` is called, the lua function invokes "C" code to set some global variable like `on_keys_disables_mappings` for use in getchar.c; this avoids a lua call to get the value of the boolean for each key.

### DIFF
--- a/runtime/lua/vim/_editor.lua
+++ b/runtime/lua/vim/_editor.lua
@@ -651,7 +651,7 @@ do
   end
 end
 
-local on_key_cbs = {} --- @type table<integer,function>
+local on_key_cbs = {} --- @type table<integer,[function, table]>
 
 --- Adds Lua function {fn} with namespace id {ns_id} as a listener to every,
 --- yes every, input key.
@@ -664,34 +664,37 @@ local on_key_cbs = {} --- @type table<integer,function>
 ---           it won't be invoked for those keys.
 ---@note {fn} will not be cleared by |nvim_buf_clear_namespace()|
 ---
----@param fn fun(key: string, typed: string)? Function invoked for every input key,
+---@param fn nil|fun(key: string, typed: string): string? Function invoked for every input key,
 ---          after mappings have been applied but before further processing. Arguments
 ---          {key} and {typed} are raw keycodes, where {key} is the key after mappings
 ---          are applied, and {typed} is the key(s) before mappings are applied.
 ---          {typed} may be empty if {key} is produced by non-typed key(s) or by the
 ---          same typed key(s) that produced a previous {key}.
----          When {fn} is `nil` and {ns_id} is specified, the callback associated with
----          namespace {ns_id} is removed.
+---          If {fn} returns an empty string, {key} is discarded/ignored.
+---          When {fn} is `nil`, the callback associated with namespace {ns_id} is removed.
 ---@param ns_id integer? Namespace ID. If nil or 0, generates and returns a
 ---                      new |nvim_create_namespace()| id.
+---@param opts table? Optional parameters
 ---
 ---@see |keytrans()|
 ---
 ---@return integer Namespace id associated with {fn}. Or count of all callbacks
 ---if on_key() is called without arguments.
-function vim.on_key(fn, ns_id)
+function vim.on_key(fn, ns_id, opts)
   if fn == nil and ns_id == nil then
     return vim.tbl_count(on_key_cbs)
   end
 
   vim.validate('fn', fn, 'callable', true)
   vim.validate('ns_id', ns_id, 'number', true)
+  vim.validate('opts', opts, 'table', true)
+  opts = opts or {}
 
   if ns_id == nil or ns_id == 0 then
     ns_id = vim.api.nvim_create_namespace('')
   end
 
-  on_key_cbs[ns_id] = fn
+  on_key_cbs[ns_id] = fn and { fn, opts }
   return ns_id
 end
 
@@ -700,12 +703,23 @@ end
 function vim._on_key(buf, typed_buf)
   local failed_ns_ids = {}
   local failed_messages = {}
+  local discard = false
   for k, v in pairs(on_key_cbs) do
-    local ok, err_msg = pcall(v, buf, typed_buf)
+    local ok, rv = pcall(v[1], buf, typed_buf)
+    if ok and rv ~= nil then
+      if type(rv) == 'string' and #rv == 0 then
+        discard = true
+        -- break   -- Without break deliver to all callbacks even when it eventually discards.
+        -- "break" does not make sense unless callbacks are sorted by ???.
+      else
+        ok = false
+        rv = 'return string must be empty'
+      end
+    end
     if not ok then
       vim.on_key(nil, k)
       table.insert(failed_ns_ids, k)
-      table.insert(failed_messages, err_msg)
+      table.insert(failed_messages, rv)
     end
   end
 
@@ -718,6 +732,7 @@ function vim._on_key(buf, typed_buf)
       )
     )
   end
+  return discard
 end
 
 --- Convert UTF-32, UTF-16 or UTF-8 {index} to byte index.
--- a/src/nvim/lua/executor.c
+++ b/src/nvim/lua/executor.c
@@ -2063,12 +2063,13 @@ char *nlua_register_table_as_callable(const typval_T *const arg)
   return name;
 }
 
-void nlua_execute_on_key(int c, char *typed_buf)
+/// @return true to discard the key
+bool nlua_execute_on_key(int c, char *typed_buf)
 {
   static bool recursive = false;
 
   if (recursive) {
-    return;
+    return false;
   }
   recursive = true;
 
@@ -2097,9 +2098,15 @@ void nlua_execute_on_key(int c, char *typed_buf)
 
   int save_got_int = got_int;
   got_int = false;  // avoid interrupts when the key typed is Ctrl-C
-  if (nlua_pcall(lstate, 2, 0)) {
+  bool discard = false;
+  if (nlua_pcall(lstate, 2, 1)) {
     nlua_error(lstate,
                _("Error executing  vim.on_key Lua callback: %.*s"));
+  } else {
+    if (lua_isboolean(lstate, -1)) {
+      discard = lua_toboolean(lstate, -1);
+    }
+    lua_pop(lstate, 1);
   }
   got_int |= save_got_int;
 
@@ -2112,6 +2119,7 @@ void nlua_execute_on_key(int c, char *typed_buf)
 #endif
 
   recursive = false;
+  return discard;
 }
 
 // Sets the editor "script context" during Lua execution. Used by :verbose.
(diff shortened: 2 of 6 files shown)


errael:
There is no test for `may_discard=true`. I've looked at the test vim_spec.lua; need to feed a key that has a side effect that can be detected. Any thoughts?

BTW, no need for the test if there's no chance this PR is in general acceptable.

errael:
After looking around in test/functional/lua, and at `describe('vim.on_key', ...`, as well as the goodies in `test.functional.testnvim` and friends, I think I can put together a test. But how to test the test? In particular, how to run something less than `make functionaltest`? I didn't see any hints in `neovim/test/README.md`.

errael:
Aha, something _was_ in there,
```
TEST_FILE=test/functional/lua/vim_spec.lua TEST_FILTER='vim.on_key' make functionaltest
```

errael:
@zeertzjq @justinmk Added a test. 

An open question is whether the `{allow_mapping=false}` option should be added to this PR. If so, I need some implementation guidance
- where to set up core global `on_key_no_mapping` for use in `vgetc()`
- how to set the "C" global from the `lua` code `vim.on_key()`

Note: while any `vim.on_key()` callback has `{allow_mapping=false}` then no mappings would occur.

runtime/lua/vim/_editor.lua
```diff
@@ -670,28 +670,33 @@ local on_key_cbs = {} --- @type table<integer,function>
...
 ---          {typed} may be empty if {key} is produced by non-typed key(s) or by the
 ---          same typed key(s) that produced a previous {key}.
+---          If {opts.may_discard} then when {fn} returns
+---          true, {key} is ignored.
 ---          When {fn} is `nil` and {ns_id} is specified, the callback associated with
 ---          namespace {ns_id} is removed.
 ---@param ns_id integer? Namespace ID. If nil or 0, generates and returns a
 ---                      new |nvim_create_namespace()| id.
+---@param opts table|nil Optional parameters
+---              - may_discard {fn} controls discard (default false)
```
justinmk, line 680: Why is this option needed? Previously, the spec for `fn` was that its return type is "void". If we now treat a `true` return value specially, we don't need to make that optional.

runtime/lua/vim/_editor.lua
```diff
@@ -670,28 +670,33 @@ local on_key_cbs = {} --- @type table<integer,function>
 ---          are applied, and {typed} is the key(s) before mappings are applied.
 ---          {typed} may be empty if {key} is produced by non-typed key(s) or by the
 ---          same typed key(s) that produced a previous {key}.
+---          If {opts.may_discard} then when {fn} returns
+---          true, {key} is ignored.
 ---          When {fn} is `nil` and {ns_id} is specified, the callback associated with
 ---          namespace {ns_id} is removed.
```
justinmk, line 676: nit: while we're here we can simplify this


```suggestion
---          When {fn} is `nil`, the callback associated with namespace {ns_id} is removed.
```

runtime/lua/vim/_editor.lua
```diff
@@ -700,12 +705,19 @@ end
 function vim._on_key(buf, typed_buf)
   local failed_ns_ids = {}
   local failed_messages = {}
+  local discard = false
   for k, v in pairs(on_key_cbs) do
-    local ok, err_msg = pcall(v, buf, typed_buf)
+    local ok, discard_or_err_msg = pcall(v[1], buf, typed_buf)
```
justinmk, line 710:
```suggestion
    local ok, rv = pcall(v[1], buf, typed_buf)
```

justinmk:
> It is desirable that when `vim.on_key()` is called, the lua function invokes "C" code to set some global variable like `on_keys_disables_mappings` for use in getchar.c; this avoids a lua call to get the value of the boolean for each key.

Why is that desirable? The current impl in this PR seems much more obvious, and avoids another global...

src/nvim/lua/executor.c
```diff
@@ -2063,12 +2063,13 @@ char *nlua_register_table_as_callable(const typval_T *const arg)
...
 
-void nlua_execute_on_key(int c, char *typed_buf)
+// @return true to discard the key
+bool nlua_execute_on_key(int c, char *typed_buf)
 {
   static bool recursive = false;
 
   if (recursive) {
-    return;
+    return false;
```
justinmk, line 2072: would `TriState` be useful here?

test/functional/lua/vim_spec.lua
```diff
@@ -3416,6 +3417,32 @@ describe('lua stdlib', function()
                                                                     |
       ]])
     end)
+
+    it('option {may_discard=true}', function()
```
justinmk, line 3421:
```suggestion
    it('can discard input', function()
```

justinmk:
In order to future-proof this (and remove the need for `KeyInputPre` https://github.com/neovim/neovim/pull/30061), why not make the return value a string? Empty string means "discard", and non-empty can be used in the future to replace the char with one (or more) chars.

runtime/lua/vim/_editor.lua
errael: > Why is this option needed? Previously, the spec for `fn` was that its return type is "void". If we now treat a `true` return value specially, we don't need to make that optional.

It's not needed. It's there as a compatibility thing, even then it seemed a stretch. I'll remove it unless there's further comment to the contrary.

errael:
> > It is desirable that when `vim.on_key()` is called, the lua function invokes "C" code to set some global variable like `on_keys_disables_mappings` for use in getchar.c; this avoids a lua call to get the value of the boolean for each key.
> 
> Why is that desirable? The current impl in this PR seems much more obvious, and avoids another global...

This is not part of "can consume"; this is for "can prevent mapping", see #30741. The current `getchar.c::vgetc()` has
```
      if (mod_mask) {           // no mapping after modifier has been read
        no_mapping++;
```
and the idea is `if (mod_mask || on_key_disable_mapping)`. A global avoids doing another lua function invocation in `vget()`; that may not be an issue.

errael:
@justinmk I did "resolve conversation" for the changes I made in response to your comment. It always seemed more appropriate for the originator of the comment to resolve it; but I've never seen it used that way (if it's used at all).

Is there a preferred protocol around that in this project?

errael:
@justinmk @zeertzjq I've updated this PR so that the `vim.on_key()` callback returns a string rather than a boolean as discussed in https://github.com/neovim/neovim/pull/30939#issuecomment-2442238348.

The question of "prevent mapping" (https://github.com/neovim/neovim/pull/30939#issue-2612617742, https://github.com/neovim/neovim/pull/30939#issuecomment-2442227420) is still open. This PR is OK standalone without the "prevent mapping" or the "prevent mapping" can be included with some guidance.

max397574:
if we implemented the functionality that if a string is returned that gets used to replace the type char then the "prevent mappings" thing would be nothing more than just returning the second parameter passed to the function wouldn't it?

zeertzjq:
No

src/nvim/lua/executor.c
```diff
@@ -2063,12 +2063,13 @@ char *nlua_register_table_as_callable(const typval_T *const arg)
   return name;
 }
 
-void nlua_execute_on_key(int c, char *typed_buf)
+// @return true to discard the key
```
zeertzjq, line 2066:
```suggestion
/// @return true to discard the key
```

runtime/lua/vim/_editor.lua
```diff
@@ -651,7 +651,7 @@ do
   end
 end
 
-local on_key_cbs = {} --- @type table<integer,function>
+local on_key_cbs = {} --- @type table<integer,[function, {}?]>
```
zeertzjq, line 654: This annotation may be inaccurate. Currently `vim.on_key()` only checks if `opts` is a table. It doesn't check the content of `opts`, so `opts` is not necessarily an empty table.

src/nvim/lua/executor.c
errael: Right. Since this is private, I probably thought `///` might have a bad interaction with the public documentation generation...

runtime/lua/vim/_editor.lua
errael: I'd like to keep the _private implementation_ a list assuming there will eventually be `vim.on_key()` options; or at least until it's decided there will be no options.  The code that works with a list is not that complex, so taking it out and putting it back soon is not that big of a deal. But in any event probably
```
local on_key_cbs = {} --- @type table<integer,[function, table?]>
```
makes more sense for the private data structure.

I can take the `{opts}` out of the public facing
```
vim.on_key({fn}, {ns_id}, {opts})
```
But I'm hoping that soon there's a PR with `{mapping = boolean?}`, (default true). BTW, I believe that would be the last missing piece for a full (pretty full) popup implementation.

Any thoughts on adding `{mapping = boolean?}` and getting the information from `lua` to `C`.

zeertzjq: This is not big deal, but there is another thing: It may be better to use `opts = opts or {}` in `vim.on_key()`, so that `opts` is never `nil` in the table here.

zeertzjq: > Any thoughts on adding {mapping = boolean?} and getting the information from lua to C.

I think putting this in `vim.on_key()` may be a bit confusing, considering that disabling mappings is a global flag. But I have no idea where else to put it either.

errael: The vim doc only recommends using it in limited situations when all keys are consumed by the filter. If this guidance is used, there's no issue.
```
							*popup-mapping*
Normally the key is what results after any mapping, since the keys pass on as
normal input if the filter does not use it.  If the filter consumes all the
keys, set the "mapping" property to zero so that mappings do not get in the
way.  This is default for |popup_menu()| and |popup_dialog()|.
```

errael: > This is not big deal, but there is another thing: It may be better to use `opts = opts or {}` in `vim.on_key()` (like other Lua APIs do), so that `opts` is never `nil` in the table here, and then you can remove the `?`.

Cool. Good to know.

errael: > I think putting this in `vim.on_key()` may be a bit confusing, considering that disabling mappings is a global flag,

Yeah, the docs should say something like `typically only false when all keys are consumed`.

src/nvim/lua/executor.c
errael: DONE

runtime/lua/vim/_editor.lua
errael: DONE (small edits, nothing about mapping)

errael:
@zeertzjq I've made the small cleanup edits. Haven't done `news.txt`, should that be part of this PR?

It seems best to handle `{mapping = boolean}` option in a separate PR since it might be controversial.

zeertzjq (approved):

PR #30945, opened 2024-10-25
fix(completion): determine LSP completion matching by completeopt

### LINKED ISSUE #30416: [lsp,completion] Do fuzzy matching only if `completeopt` has `fuzzy` in it
### Problem

Triggering the new built-in LSP autocompletion always does fuzzy matching on the completion item [here](https://github.com/neovim/neovim/blob/0ba388847417b8aed469976acf94f9633cb03f15/runtime/lua/vim/lsp/completion.lua?plain=1#L249).
In case a user is not using `completeopt+=fuzzy`, this yields different result when the completion is automatically filtered and when user triggers it. Here are some examples:

With `completeopt+=fuzzy`:

Having an array in JS, and doing `arr.re`, autotrigger completion gives this:

screenshot_2024_09_18_15_14_14

When I do `<C-e>` to cancel completion, and then do a manual `.trigger()`, I get this:

screenshot_2024_09_18_15_14_24

"As you type" filtering seems generally smarter than doing `matchfuzzy`, but I understand why `entries` appears.
Ideally, we would have things sorted by the fuzzy matching score, but that is probably a discussion for another time.

Now, Without the `fuzzy` in the `completeopt`:

This is what autotrigger gives:
screenshot_2024_09_18_15_18_35

And this is if I cancel and trigger manually:
screenshot_2024_09_18_15_19_05

Both `entries` and `forEach` should not appear in this case.

### Expected behavior

In case `completeopt` doesn't have `fuzzy`, filtering should be done strictly, from the start of the string. Basically, this:

```lua
    matches = function(item)
      local text = item.filterText or item.label
      return vim.startswith(text, prefix)
    end
```

This change still gives a bit of different results in the manual completion, but it is more correct:

screenshot_2024_09_18_15_23_39

I'm not sure how `smartcase`, `ignorecase` and other options should play with this, but the general idea is to do more strict matching without `fuzzy` in autocompletion.

kristijanhusak:
Fixes #30416

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -244,8 +244,24 @@ function M._lsp_to_complete_items(result, prefix, client_id)
   else
     ---@param item lsp.CompletionItem
     matches = function(item)
-      local text = item.filterText or item.label
-      return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
+      if item.filterText then
+        return next(vim.fn.matchfuzzy({ item.filterText }, prefix)) ~= nil
+      end
+
+      if item.textEdit then
+        -- server took care of filtering
+        return true
+      end
+
+      local text = item.label
+      if vim.o.completeopt:find('fuzzy') ~= nil then
+        return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
+      end
+
+      if vim.o.ignorecase and (not vim.o.smartcase or not prefix:find('%u')) then
+        return vim.startswith(text:lower(), prefix:lower())
+      end
+      return vim.startswith(text, prefix)
     end
   end
 
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -220,6 +220,20 @@ local function get_doc(item)
   return ''
 end
 
+---@param value string
+---@param prefix string
+---@return boolean
+local function match_item_by_value(value, prefix)
+  if vim.o.completeopt:find('fuzzy') ~= nil then
+    return next(vim.fn.matchfuzzy({ value }, prefix)) ~= nil
+  end
+
+  if vim.o.ignorecase and (not vim.o.smartcase or not prefix:find('%u')) then
+    return vim.startswith(value:lower(), prefix:lower())
+  end
+  return vim.startswith(value, prefix)
+end
+
 --- Turns the result of a `textDocument/completion` request into vim-compatible
 --- |complete-items|.
 ---
@@ -245,7 +259,7 @@ function M._lsp_to_complete_items(result, prefix, client_id)
     ---@param item lsp.CompletionItem
     matches = function(item)
       if item.filterText then
-        return next(vim.fn.matchfuzzy({ item.filterText }, prefix)) ~= nil
+        return match_item_by_value(item.filterText, prefix)
       end
 
       if item.textEdit then
@@ -253,15 +267,7 @@ function M._lsp_to_complete_items(result, prefix, client_id)
         return true
       end
 
-      local text = item.label
-      if vim.o.completeopt:find('fuzzy') ~= nil then
-        return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
-      end
-
-      if vim.o.ignorecase and (not vim.o.smartcase or not prefix:find('%u')) then
-        return vim.startswith(text:lower(), prefix:lower())
-      end
-      return vim.startswith(text, prefix)
+      return match_item_by_value(item.label, prefix)
     end
   end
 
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
I don't think non-fuzzy match is going to work on `item.filterText` because some servers do imply the client can do fuzzy matching. See https://github.com/ansible/ansible-language-server/issues/556

Also see this part of the specification:


> A completion item provides additional means to influence filtering and sorting. They are expressed by either creating a CompletionItem with a insertText or with a textEdit. The two modes differ as follows:

> Completion item provides an insertText / label without a text edit: in the model the client should filter against what the user has already typed using the word boundary rules of the language (e.g. resolving the word under the cursor position). The reason for this mode is that it makes it extremely easy for a server to implement a basic completion list and get it filtered on the client.

> Completion Item with text edits: in this mode the server tells the client that it actually knows what it is doing. If you create a completion item with a text edit at the current cursor position no word guessing takes place and no filtering should happen. This mode can be combined with a sort text and filter text to customize two things. If the text edit is a replace edit then the range denotes the word used for filtering. If the replace changes the text it most likely makes sense to specify a filter text to be used.


With these changes we'd further diverge from the specification.


I think what could be an option is to:

- ~Always fuzzy match if filterText is present~ Actually - this might not be needed if we never pre-filter if there is a textEdit.
- If item has insertText/label without a text edit: Fuzzy match if `fuzzy` is set, otherwise prefix match
- If item has textEdit: Never filter

kristijanhusak:
Ok, so if I understand correctly, I just need to do an early check if there is a `textEdit` and always match?

```lua
    matches = function(item)
      -- LSP server took care of filtering
      if item.textEdit then
        return true
      end

      local text = item.filterText or item.label
      return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
      if fuzzy then
        return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
      end

      local case_insensitive = ignorecase and (not smartcase or not prefix:find('%u'))

      if case_insensitive then
        return vim.startswith(text:lower(), prefix:lower())
      end

      return vim.startswith(text, prefix)
    end
```

I'm not sure if I need to do something about this part of specification:

> If the text edit is a replace edit then the range denotes the word used for filtering. If the replace changes the text it most likely makes sense to specify a filter text to be used

mfussenegger:
Re-reading the passage again I think it would need to be something like:

```diff
diff --git a/runtime/lua/vim/lsp/completion.lua b/runtime/lua/vim/lsp/completion.lua
index e36d329dc5..2d208c90fb 100644
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -242,12 +242,25 @@ function M._lsp_to_complete_items(result, prefix, client_id)
       return true
     end
   else
     ---@param item lsp.CompletionItem
     matches = function(item)
-      local text = item.filterText or item.label
-      return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
+      if item.filterText then
+        return next(vim.fn.matchfuzzy({ item.filterText }, prefix)) ~= nil
+      end
+      if item.textEdit then
+        -- server took care of filtering
+        return true
+      end
+      local text = item.label
+      if vim.o.completeopt:find('fuzzy') ~= nil then
+        return next(vim.fn.matchfuzzy({ text }, prefix)) ~= nil
+      end
+      if vim.o.ignorecase and (not vim.o.smartcase or not prefix:find('%u')) then
+        return vim.startswith(text:lower(), prefix:lower())
+      end
+      return vim.startswith(text, prefix)
     end
   end
 
   local candidates = {}
   local bufnr = api.nvim_get_current_buf()
```

kristijanhusak:
That sounds about right to me also. Updated the logic and added few more tests.

mfussenegger (approved):

kristijanhusak:
@mfussenegger with your suggested change, we are assuming that the `filterText` should always be fuzzy matched.

Since the LSP specification doesn't mention anything about fuzzy matching, could we apply the same logic to `filterText` as we do for `label`? Since not all of the LSPs provide `filterText`, a user who disables fuzzy matching could get different results for the same typed words in different languages, depending if LSP provides `filterText` or not. I believe we should be consistent with the functionality.

mfussenegger:
> @mfussenegger with your suggested change, we are assuming that the `filterText` should always be fuzzy matched.
> 
> Since the LSP specification doesn't mention anything about fuzzy matching, could we apply the same logic to `filterText` as we do for `label`? Since not all of the LSPs provide `filterText`, a user who disables fuzzy matching could get different results for the same typed words in different languages, depending if LSP provides `filterText` or not. I believe we should be consistent with the functionality.

Fine for me to change it to also do prefix match on filterText if fuzzy is missing from completeopt.

PR #30958, opened 2024-10-27
refactor(options): make 'shellslash' and 'completeslash' no-op on POSIX

famiu:
Problem: `'shellslash'` and `'completeslash'` are the only platform-specific options and require extra logic just to accomodate them.

Solution: Make `'shellslash'` and `'completeslash'` no-op on POSIX systems instead of disabling them altogether.

### DIFF
--- a/src/nvim/buffer.c
+++ b/src/nvim/buffer.c
@@ -2105,6 +2105,7 @@ void free_buf_options(buf_T *buf, bool free_p_ff)
   clear_string_option(&buf->b_p_lw);
   clear_string_option(&buf->b_p_bkc);
   clear_string_option(&buf->b_p_menc);
+  clear_string_option(&buf->b_p_csl);
 }
 
 /// Get alternate file "n".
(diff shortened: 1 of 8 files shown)


zeertzjq:
Did you forget my other comment:
> `exists('+shellslash')` is commonly used to check if setting 'shellslash' has an effect. If `exists('+shellslash')` always returns 1, it will cause even more breakage.

I searched on GitHub. `exists('+shellslash')` is also used to check if `\` should be escape in a file path, so if it always returns 1 it will also cause escaping problems.

justinmk:
> `exists('+shellslash')` is also used to check if `\` should be escaped in a file path, so if it always returns 1 it will also cause escaping problems.

This is a valid point. Can we special-case `exists('+shellslash')` to return FALSE on non-Windows? I realize that this is somewhat re-inventing the "hidden options" concept, but I'm hoping that this will be limited to 1 or 2 special-cases and thus we still have a net benefit from removing the "hidden options" concept.

famiu:
Superseded by #28400

PR #30966, opened 2024-10-27
fix(lsp): list all workspace folders in healthcheck

### LINKED ISSUE #30648: `:checkhealth lsp` should show workspace folders
This client has 2 workspace folders configured, but only shows the root directory. We should also show all workspace folders if there are any (or perhaps if there are >1)?

```
vim.lsp: Active Clients ~
- gopls (id: 1)
    Root directory: ~/src/some/project
    Command: gopls
    Settings: {
      gopls = {
        analyses = {
          nilness = true,
          unusedparams = true,
          unusedwrite = true
        }
      }
    }
    Attached buffers: 1
```

cc @mfussenegger

MariaSolOs:
Closes #30648 

Note that when workspace folders is defined we can just print those, since they should include the `root_dir`.

Example of what it looks like:
image

### DIFF
--- a/runtime/lua/vim/lsp/health.lua
+++ b/runtime/lua/vim/lsp/health.lua
@@ -39,12 +39,27 @@ local function check_active_clients()
       elseif type(client.config.cmd) == 'function' then
         cmd = tostring(client.config.cmd)
       end
+      local dirs_info ---@type string
+      if client.workspace_folders and #client.workspace_folders > 1 then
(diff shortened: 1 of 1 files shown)


justinmk:
Thanks! Also looking for help with upstreaming as much of `:checkhealth lspconfig` as possible. https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/health.lua

PR #30967, opened 2024-10-28
fix(pum): don't select item when clicking to the left/right

zeertzjq:
Problem:  Selecting an item in the right-click menu when clicking to the
          left/right of it is confusing, especially in a UI that doesn't
          support 'mousemoveevent'.
Solution: Don't select an item when clicking to the left/right of the
          right-click menu.


justinmk:
"To the left/right" here means "outside of the menu"?

zeertzjq:
Yes

PR #30969, opened 2024-10-28
docs(options): add NonInteractive and plaintext rendering to powershell 'shellcmdflag'

kmoschcau:
This at least somewhat hints to pwsh/powershell, that shell sessions created from :! are not interactive.
Secondly `$PSStyle.OutputRendering` = 'plaintext' causes pwsh/powershell to omit ANSI escape sequences in its output.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -5130,7 +5130,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 							*shell-powershell*
 	To use PowerShell: >vim
 		let &shell = executable('pwsh') ? 'pwsh' : 'powershell'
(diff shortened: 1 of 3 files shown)


justinmk:
Thanks! Please edit the item in `src/nvim/options.lua` then run `make doc` and commit the results. That is what the failing CI is hinting at.

I wonder why these flags were not already mentioned, are these relatively new features of powershell?

kmoschcau:
I don't have an authoritative answer when these features were added to powershell. But most discussions around it happened in 2021.

Also I had problems getting the build to succeed. On Windows, building fails in the `luajit` step, because the `[` executable can't be found, which makes sense. On Linux it complained because `gettext` was missing. This should be documented as build dependencies.

I made the requested changes.

PR #30973, opened 2024-10-28
docs: fix some line wrappings

MrChozo:
Problem:
Paragraph below bulleted list was joined to last list item instead of dropping below as expected. Same with "Notes:" paragraph below it.

Solution:
Added necessary newlines.

### DIFF
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -41,15 +41,19 @@ thus you cannot edit beyond that.
 The command-lines that you enter are remembered in a history table.  You can
 recall them with the up and down cursor keys.  There are actually five
 history tables:
+
 - one for ':' commands
 - one for search strings
 - one for expressions
 - one for input lines, typed for the |input()| function.
 - one for debug mode commands
+
 These are completely separate.  Each history can only be accessed when
 entering the same type of line.
 Use the 'history' option to set the number of lines that are remembered.
+
 Notes:
+
 - When you enter a command-line that is exactly the same as an older one, the
   old one is removed (to avoid repeated commands moving older commands out of
   the history).


MrChozo:
Huh. No idea what made "test / ubuntu asan clang unittest (pull_request)" fail.

zeertzjq:
Are blank lines before bulleted lists necessary?

MrChozo:
It's very likely that my assumption about the HTML version of the docs using Markdown is incorrect. I haven't found any resources about this topic, but it's not the only spot in the HTML version that hasn't translated well:

Screenshot 2024-10-30 150217

runtime/doc/cmdline.txt
```diff
@@ -41,15 +41,19 @@ thus you cannot edit beyond that.
...
 - one for search strings
 - one for expressions
 - one for input lines, typed for the |input()| function.
 - one for debug mode commands
+
 These are completely separate.  Each history can only be accessed when
 entering the same type of line.
 Use the 'history' option to set the number of lines that are remembered.
+
 Notes:
```
justinmk, line 55: This blank is definitely necessary, otherwise the parser has no way of knowing when a "block" starts and ends.

runtime/doc/cmdline.txt
```diff
@@ -41,15 +41,19 @@ thus you cannot edit beyond that.
...
 recall them with the up and down cursor keys.  There are actually five
 history tables:
+
 - one for ':' commands
 - one for search strings
 - one for expressions
 - one for input lines, typed for the |input()| function.
 - one for debug mode commands
+
 These are completely separate.  Each history can only be accessed when
```
justinmk, line 51: This one is also necessary for the same reason. Because of the inconsistency of vimdoc, once a list item starts, it doesn't end until a blank line.

zeertzjq:
So how about the remaining blank lines?

runtime/doc/cmdline.txt
```diff
@@ -41,15 +41,19 @@ thus you cannot edit beyond that.
 The command-lines that you enter are remembered in a history table.  You can
 recall them with the up and down cursor keys.  There are actually five
 history tables:
+
 - one for ':' commands
```
justinmk, line 45: Technically this blank line isn't needed.

runtime/doc/cmdline.txt
```diff
@@ -41,15 +41,19 @@ thus you cannot edit beyond that.
...
 - one for input lines, typed for the |input()| function.
 - one for debug mode commands
+
 These are completely separate.  Each history can only be accessed when
 entering the same type of line.
 Use the 'history' option to set the number of lines that are remembered.
+
 Notes:
+
 - When you enter a command-line that is exactly the same as an older one, the
```
justinmk, line 57: Technically this blank line isn't needed.

PR #30914, opened 2024-10-23
docs: misc

dundargoc:

PR #30981, opened 2024-10-29
docs(eval): use table in virtcol function

glepnir:
when is a list-like table first is line number second can be string like `$`


zeertzjq:
The same change show be applied to `col()` and `charcol()`.

dundargoc:

PR #30983, opened 2024-10-29
fix(defaults): omit extraneous info from unimpaired mapping errors

### LINKED ISSUE #30625: Improve error handling for new unimpaired mappings

gpanders:
Resolves https://github.com/neovim/neovim/issues/30625.

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -219,132 +219,150 @@ do
 
   --- vim-unimpaired style mappings. See: https://github.com/tpope/vim-unimpaired
   do
-    ---@param lhs string
-    ---@param rhs function
-    ---@param desc string
-    local function create_unimpaired_mapping(lhs, rhs, desc)
-      vim.keymap.set('n', lhs, function()
-        local _, err = pcall(rhs) ---@type any, string
-        if err then
-          vim.api.nvim_err_writeln(err)
-        end
-      end, { desc = desc })
+    --- Execute a command and print errors without a stacktrace.
+    --- @param opts table Arguments to |nvim_cmd()|
+    local function cmd(opts)
+      local _, err = pcall(vim.api.nvim_cmd, opts, {})
+      if err then
+        vim.api.nvim_err_writeln(err:sub(#'Vim:' + 1))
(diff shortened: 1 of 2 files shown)


test/functional/editor/defaults_spec.lua
```diff
@@ -95,6 +95,60 @@ describe('default', function()
     end)
   end)
 
-  -- describe('key mappings', function()
-  -- end)
+  describe('key mappings', function()
+    describe('unimpaired-style mappings', function()
+      it('do not show a full stack trace #30625', function()
+        n.clear({ args_rm = { '--cmd', 'comclear | mapclear | mapclear!' } })
```
zeertzjq, line 101:
```suggestion
        n.clear({ args_rm = { '--cmd' } })
```

runtime/lua/vim/_defaults.lua
```diff
@@ -219,132 +219,150 @@ do
...
-    local function create_unimpaired_mapping(lhs, rhs, desc)
-      vim.keymap.set('n', lhs, function()
-        local _, err = pcall(rhs) ---@type any, string
-        if err then
-          vim.api.nvim_err_writeln(err)
-        end
-      end, { desc = desc })
+    --- Execute a command and print errors without a stacktrace.
+    --- @param opts table Arguments to |nvim_cmd()|
+    local function cmd(opts)
```
justinmk, line 224: This might be an example of the kind of thing to put in an internal-only `_util.lua` module, where useful stuff can live and potentially (but not always) graduate later to a public interface. (This particular thing might "graduate" to a flag on the existing `vim.cmd`.)

test/functional/editor/defaults_spec.lua
gpanders: Why? Won't that remove all `--cmd` arguments? I only want to remove the `:mapclear` ones.

PR #30984, opened 2024-10-29
feat(defaults): add unimpaired empty line below/above cursor mapping

### LINKED ISSUE #30779: normal command to add blank line above/below cursor
### Problem

For some people it is quite a common task to add empty space between functions / statements in code. Usually, this is done by using `ok` or `Oj` and then moving to the previous line. But this isn't ideal because:
- it doesn't maintain exact cursor position. if we use a mark, the command becomes a lot longer
- we can't create multiple lines this way by prefixing it with a count 

### Expected behavior

I propose that there will be added 2 default keymappings to add add empty line above and below:

```lua
vim.keymap.set("n", "gL", "<cmd>call append(line('.') - 1, repeat([''], v:count1))<CR>", { desc = "empty line above" })

vim.keymap.set("n", "gl", "<cmd>call append(line('.'), repeat([''], v:count1))<CR>", { desc = "add empty line below" })
```

the keymappings are just a suggestion from me , since `go` and `gO` are both taken and `gl` & `gL` are both available

mnemonic: "L" for "**L**ine"

yochem:
## Problem

No default way to add an empty line below/above cursor in normal mode, an edit that is done fairly often by programmers and text writers. Currently There is no way to do this without creating marks or manually navigating back to previous cursor position.

## Solution

Like tpope's unimpaired, adds `[<Space>` and `]<Space` to add an empty line above and below cursor respectively.

---

Idea based on tpope's unimpaired, implementation based on [tummetott/unimpaired.nvim](https://github.com/tummetott/unimpaired.nvim).

Closes #30779.


justinmk:
I don't see a reason to deviate from unimpaired's `]e`.

yochem:
I need some help with writing the test for this. I have the following test:

```lua
it('should add an empty line above the cursors current line', function()
  n.clear()
  n.insert([[first line]])
  n.feed(']<Space>')
  n.expect([[
    first line
    ]])
end)
```

Which results in:
```
Passed in:
(string) 'first line'
Expected:
(string) 'first line
'
```

Using `n.feed('o')` instead (which should have the same behaviour, but leave cursor in insert mode) passes the test.

Running `VIMRUNTIME=./runtime ./build/bin/nvim` and pressing `[<Space>` works as expected: an empty line is inserted above the cursor.

I've pushed the failing tests for ease of checking out.

gpanders:
Default mappings are disabled in tests by default. Look at the test added in https://github.com/neovim/neovim/pull/30983 for an example of how to test this.

yochem:
Thank you! this did the trick:

```diff
it('#tagged should add an empty line above the cursors current line', function()
- n.clear()
+ n.clear({ args_rm = { '--cmd' } })
  n.insert([[first line]])
  n.feed('[<Space>')
  n.expect([[

  first line]])
end)
```

yochem:
I'm not sure in which help file to put the documentation for this. And should I maybe mention this in the help section for `o` and `O`? E.g. "|]<Space>| inserts empty lines without leaving normal mode."

telemachus:
> And should I maybe mention this in the help section for o and O? E.g. "|]| inserts empty lines without leaving normal mode."

Yes, please. As a user, I think it would help to cross-reference these in `:help o` and `:help O`.

yochem:
Saw this in the `insert.txt` doc:

> 'formatoptions' can be set to copy the comment leader when opening a new line.

It's not something that I use, but probably the expected behavior of this keymap (at least, I conceptually think of this keymap as `o` + `<Esc>`.

So I will change the implementation to the following:

```lua
vim.keymap.set('n', ']<Space>', function ()
  local cursor = vim.api.nvim_win_get_cursor(0)
  vim.cmd.norm(vim.v.count1 .. "o")
  vim.api.nvim_win_set_cursor(0, cursor)
end)
```

This _actually_ uses `o` under the hood and thus respects [`:h fo-o`](https://neovim.io/doc/user/change.html#fo-o).

gpanders:
> Saw this in the `insert.txt` doc:
> 
> > 'formatoptions' can be set to copy the comment leader when opening a new line.
> 
> It's not something that I use, but probably the expected behavior of this keymap (at least, I conceptually think of this keymap as `o` + `<Esc>`.
> 
> So I will change the implementation to the following:
> 
> ```lua
> vim.keymap.set('n', ']<Space>', function ()
>   local cursor = vim.api.nvim_win_get_cursor(0)
>   vim.cmd.norm(vim.v.count1 .. "o")
>   vim.api.nvim_win_set_cursor(0, cursor)
> end)
> ```
> 

No, please keep the behavior identical to vim-unimpaired. It should insert an empty line without adding a comment leader.

yochem:
Sure! In that case, this PR is ready for review.

yochem:
And just out of curiosity: why is identical behavior with a vim plugin more important than sharing behavior with a similar vim command? Is it because vim-unimpaired is so well known and used?

Btw, this is the reason tpope did not add support for `fo-o`:

> This behavior is very deliberate, as the reason I bothered adding these to begin with was o<Esc> was forcing comment characters where I didn't want them. I'm not sure why anyone want want a blank comment line.

Which is a fair point, wouldn't it be that this is _his_ preference and this is already something that can already be configured using `set formatoptions-=o`.

But that are my 2 cents. I don't care that much about which approach is chosen here.

zeertzjq:
1. When one uses `o` to open a new line they can use `Ctrl-U` to remove the comment leader. When one uses a Normal mode command it's harder to remove the comment leader.
2. It's more convenient to have two different behaviors to be available at the same time, instead of having to toggle an option to switch between them.

gpanders:
>And just out of curiosity: why is identical behavior with a vim plugin more important than sharing behavior with a similar vim command?

The identical behavior with vim-unimpaired is not really the point (my comment was a bit misleading here), but rather that the behavior that vim-unimpaired implements is the behavior that we actually want (for the reasons tpope himself gave in your comment as well as zeertzjq's comments).

gpanders (approved):

telemachus:
Unless I'm confused, this version of `]<Space>` and `[<Space>` is not dot-repeatable (unlike both vim-unimpaired and unimpaired.nvim). I think it would be worth looking into updating this so that it is repeatable.

gpanders:
> Unless I'm confused, this version of `]<Space>` and `[<Space>` is not dot-repeatable (unlike both vim-unimpaired and unimpaired.nvim). I think it would be worth looking into updating this so that it is repeatable.

vim-unimpaired's version is not dot-repeatable either: it relies on tpope's vim-repeat https://github.com/tpope/vim-unimpaired/blob/6d44a6dc2ec34607c41ec78acf81657248580bf1/plugin/unimpaired.vim#L234

I'm not sure what unimpaired.nvim does.

telemachus:
> vim-unimpaired's version is not dot-repeatable either: it relies on tpope's vim-repeat https://github.com/tpope/vim-unimpaired/blob/6d44a6dc2ec34607c41ec78acf81657248580bf1/plugin/unimpaired.vim#L234

Ah, right. Linewise mappings in unimpaired rely on vim-repeat.

> I'm not sure what unimpaired.nvim does.

It seems to use [this code](https://github.com/tummetott/unimpaired.nvim/blob/main/lua/unimpaired/functions.lua#L3-L13):

```lua
-- Dot repetition of a custom mapping breaks as soon as there is a dot repeatable normal
-- mode command inside the mapping. This function restores the dot repetition of
-- the mapping while preserving the [count] when called as last statement inside
-- the custom mapping
local restore_dot_repetition = function(count)
    count = count or ''
    local callback = vim.go.operatorfunc
    vim.go.operatorfunc = ''
    vim.cmd('silent! normal ' .. count .. 'g@l')
    vim.go.operatorfunc = callback
end
```

yochem:
> I'm not sure what unimpaired.nvim does.

As far as I am aware, this is the same implementation as unimpaired.nvim ([see here](https://github.com/tummetott/unimpaired.nvim/blob/8e504ba95dd10a687f4e4dacd5e19db221b88534/lua/unimpaired/functions.lua#L188-L198)).

telemachus:
The implementation of the core feature is the same, but when unimapired.nvim creates the mappings, I think that it wraps (some of) them with code to make them dot-repeatable.

See here: https://github.com/tummetott/unimpaired.nvim/blob/main/lua/unimpaired/keymaps.lua.

PR #30992, opened 2024-10-30
fix(netrw): added `\M` to avoid matching against `~` in the directory…

### LINKED ISSUE #30991: Netrw E874 when navigating on remote directory with `~` when using netrw_liststyle = 3
### Problem

With netrw_liststyle = 3, navigating in a remote directory that contains `~` shortcut for home directory, the following error message occurs :
```
Error detected while processing function <SNR>33_NetrwBrowse[252]..<SNR>33_PerformListing[241]..<SNR>33_NetrwTreeListing:
line   13:
E874: (NFA) Could not pop the stack!
E874: (NFA) Could not pop the stack!
```

### Steps to reproduce

- With netrw_liststyle = 3 :
  - `nvim scp://<HOST>/~/`
  - Navigate to another directory

### Expected behavior

Navigate without an error message

### Nvim version (nvim -v)

0.10.2

yes, vim 9.1

### Operating system/version

Ubuntu 24.04 LTS

### Terminal name/version

Alacritty 0.13.2

alacritty

### Installation

snap

tomben13:
Fixes #30991.

The regexp was matching against the `~` character in the directory name, which causes the regexp engine to pop the latest substitute string which didn't exist.

`\M` disables the `~` special character (and others) for the rest of the string, which is the directory name that may contain `~` in our case.

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -9550,7 +9550,7 @@ fun! s:NetrwTreeListing(dirname)
     let w:netrw_treetop= a:dirname
     let s:netrw_treetop= w:netrw_treetop
 "    call Decho("w:netrw_treetop<".w:netrw_treetop."> (reusing)",'~'.expand("<slnum>"))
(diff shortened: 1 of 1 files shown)


clason:
We don't maintain `netrw`; please contribute any changes to vim/vim.

PR #30993, opened 2024-10-30
refactor(options): unify `put_set(num|bool|string)`

famiu:
Problem: There are three separate functions for printing the `:set` command for an setting an option to a file descriptor. These functions are used when creating the session file for an option. Having a function for each type increase code duplication and also makes it harder to add logic for new option types.

Solution: Replace `put_set(num|bool|string)` with a single `put_set` function which works for all option types, this reduces code duplication and also makes it trivial to add support for more option types in the future.

### DIFF
--- a/src/nvim/option.c
+++ b/src/nvim/option.c
@@ -4368,34 +4368,23 @@ int makeset(FILE *fd, int opt_flags, int local_only)
             cmd = "setlocal";
           }
 
-          if (option_has_type(opt_idx, kOptValTypeBoolean)) {
-            if (put_setbool(fd, cmd, opt->fullname, *(int *)varp) == FAIL) {
+          bool do_endif = false;
+          // Don't set 'syntax' and 'filetype' again if the value is
+          // already right, avoids reloading the syntax file.
+          if (opt->indir == PV_SYN || opt->indir == PV_FT) {
+            if (fprintf(fd, "if &%s != '%s'", opt->fullname,
+                        *(char **)(varp)) < 0
+                || put_eol(fd) < 0) {
               return FAIL;
             }
-          } else if (option_has_type(opt_idx, kOptValTypeNumber)) {
-            if (put_setnum(fd, cmd, opt->fullname, (OptInt *)varp) == FAIL) {
+            do_endif = true;
+          }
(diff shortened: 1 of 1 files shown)


src/nvim/option.c
```diff
@@ -4368,35 +4368,8 @@ int makeset(FILE *fd, int opt_flags, int local_only)
...
-              return FAIL;
-            }
-          } else {    // string
-            bool do_endif = false;
-
-            // Don't set 'syntax' and 'filetype' again if the value is
-            // already right, avoids reloading the syntax file.
-            if (opt->indir == PV_SYN || opt->indir == PV_FT) {
-              if (fprintf(fd, "if &%s != '%s'", opt->fullname,
-                          *(char **)(varp)) < 0
```
zeertzjq, line 4386: Where did this check go?

famiu: @echasnovski How necessary is this check? Removing it removes a fair amount of complexity, and tests seem tp pass just fine with it. Seems like it's just for performance? If so, unless keeping that check is _really_ worth the added complexity, I'm in favor of removing it.

famiu: I'll add it back for now

src/nvim/option.c
```diff
@@ -4409,110 +4398,133 @@ int makeset(FILE *fd, int opt_flags, int local_only)
...
+    break;
+  }
+  case kOptValTypeNumber: {
+    if (fprintf(fd, "%s %s=", cmd, name) < 0) {
+      return FAIL;
+    }
+
+    OptInt value_num = value.data.number;
+    OptInt wc;
+
```
zeertzjq, line 4455: Why is there now a blank line between `OptInt wc` and `wc_use_keyname(varp, &wc)`?

famiu: Was just a stylistic choice to make it easier to read for myself. Will remove it since it's problematic

zeertzjq: Then it's inconsistent with the boolean and string branches. Actually, I was suggesting moving the blank line before the `OptInt wc`.

famiu: Should be good now

zeertzjq (approved):

PR #30994, opened 2024-10-30
fix(float): missing error handle in win_alloc_aucmd_win

glepnir:
Problem: missng error handling of win_new_float in win_alloc_aucmd_win. and this function only used in autocmd.c so move it to autocmd.c with static.

Solution: handler err after calling win_new_float.

PS:  Is this a design that no error handling is needed here?  we assuming that the creation will always succeed then..not sure returning directly is okay. maybe need some data clean up..

### DIFF
--- a/src/nvim/autocmd.c
+++ b/src/nvim/autocmd.c
@@ -1252,6 +1252,27 @@ bool check_nomodeline(char **argp)
   return true;
 }
 
+// Init `aucmd_win[idx]`. This can only be done after the first window
+// is fully initialized, thus it can't be in win_alloc_first().
+static bool aucmd_win_alloc(int idx)
+{
+  Error err = ERROR_INIT;
+  WinConfig fconfig = WIN_CONFIG_INIT;
+  fconfig.width = Columns;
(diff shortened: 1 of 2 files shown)


zeertzjq (changes requested):
No error can happen, and even if error can happen, it's much more complicated than a simple `return`, so it's better to just `abort()` on failure.

src/nvim/autocmd.c
```diff
@@ -1252,6 +1252,27 @@ bool check_nomodeline(char **argp)
   return true;
 }
 
+// Init `aucmd_win[idx]`. This can only be done after the first window
+// is fully initialized, thus it can't be in win_alloc_first().
+static inline bool aucmd_win_alloc(int idx)
```
zeertzjq, line 1257: Why `inline`?

glepnir: ops forget undo here.

zeertzjq (approved):

zeertzjq:
Thinking again, I don't think this should be merged.

PR #30996, opened 2024-10-30
refactor(options): move option callbacks from option.c to option_cb.c

famiu:
Problem: `option.c` is too big

Solution: Move option callbacks from `option.c` to `option_cb.c`.
A possible follow-up PR would be to move option callbacks from `optionstr.c` to `option_funcs.c` as well.


famiu:
Question: Should I do `optionstr.c` in this PR too or should I leave it for a follow-up.

IMO it is better to leave it for a follow-up PR to avoid having too many changes in one PR.

zeertzjq:
I thinks the `option_funcs.c` name is a bit strange. What source file doesn't mainly consist of "funcs"?

famiu:
> I thinks the `option_funcs.c` name is a bit strange. The "funcs" part is confusing and meaningless. What source file doesn't mainly consist of "funcs"?

How about `option_cb.c`

famiu:
> I thinks the `option_funcs.c` name is a bit strange. The "funcs" part is confusing and meaningless. What source file doesn't mainly consist of "funcs"?

Renamed it to `option_cb.c`. Lmk if that's fine, or if you can think of a better name

justinmk:
This doesn't really seem like a useful way to slice the module.

famiu:
> This doesn't really seem like an intuitive way to slice the module.

Yeah, I've made several attempts to salvage this and I agree. Need to find a better way to separate the components

PR #30997, opened 2024-10-30
docs(build): how to build deb

spacey-sooty:
Using a package manager isn't really viable on Ubuntu LTS cause important plugins like nvim-treesitter require 0.10 or greater and the packaged version is 0.95, but I still want system files to be removed cleanly when I uninstall or upgrade

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -12,8 +12,7 @@
     - To build on Windows, see the [Building on Windows](#building-on-windows) section. _MSVC (Visual Studio) is recommended._
 4. `sudo make install`
     - Default install location is `/usr/local`
-    - On Debian/Ubuntu, instead of `sudo make install`, you can try `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This helps ensure clean removal of installed files.
-        - Note: This is an unsupported, "best-effort" feature of the Nvim build.
+    - On Debian/Ubuntu, instead of `sudo make install`, you can try `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This helps ensure clean removal of installed files. Note: This is an unsupported, "best-effort" feature of the Nvim build.
 
 **Notes**:
 - From the repository's root directory, running `make` will download and build all the needed dependencies and put the `nvim` executable in `build/bin`.
--- a/BUILD.md
+++ b/BUILD.md
@@ -12,7 +12,8 @@
     - To build on Windows, see the [Building on Windows](#building-on-windows) section. _MSVC (Visual Studio) is recommended._
 4. `sudo make install`
     - Default install location is `/usr/local`
-    - On Debian/Ubuntu, instead of installing files directly with `sudo make install`, you can run `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This should help ensuring the clean removal of installed files.
+    - On Debian/Ubuntu, instead of `sudo make install`, you can try `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This helps ensure clean removal of installed files.
+        - Note: This is an unsupported, "best-effort" feature of the Nvim build.
 
 **Notes**:
 - From the repository's root directory, running `make` will download and build all the needed dependencies and put the `nvim` executable in `build/bin`.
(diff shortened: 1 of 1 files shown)


dundargoc:
The thing is that I don't see the benefit of doing this over just using the appimage. You can't update neovim via apt, nor does the package make use of external dependencies like an actual package manager.

przepompownia:
Good to know about `cmake -G generator`. It would be good to add some general note about it instead of for DEB only. 

Appimages cannot be simply used within containers created by `systemd-run` (for example like in `:h $NVIM_APPNAME`). 

As a Debian user I stopped using neovim deb packages (also for rare root usages) since I build it from source (and install as a normal user with `CMAKE_EXTRA_FLAGS="-DCMAKE_INSTALL_PREFIX=${HOME}/tmp/nvim/builds/from-src"`). I see one benefit (for single user desktop instances): deb package is easy to uninstall, unlike files distributed in `/usr/local`.

dundargoc:
> It would be good to add some general note about it instead of for DEB only. 

Nah, this only works for DEB. No one has configured the other formats to work (and I don't want think neovim should maintain code for 10 different package formats). 

> I see one benefit (for single user desktop instances): deb package is easy to uninstall 

Holdup. Didn't you say the problem with appimages were that they can't be used in containers, and that the benefit of deb packages is that they are easy to uninstall. But containers are also easy to "uninstall". So the one drawback you mentioned for appimages is where deb packages won't help either IIUC.

dundargoc:
Honestly this whole thing sounds anti-patterny to me but I don't care enough to argue about it if people have such strong stockholm syndrome about it. The cpack cmake code has been stable so it hasn't been a maintenance burden so far. If that changes we might need ro revisit. For now reintroducing this garbage to the build docs is fine for me.

dundargoc (approved):

dundargoc:
@justinmk @jamessan

przepompownia:
> So the one drawback you mentioned for appimages is where deb packages won't help either IIUC.

From non-appimage ways, deb is one easy uninstallable, like extracted tarball or `make install`-ed structure in some not shared directory. 

> Nah, this only works for DEB. No one has configured the other formats to work

I tried 
```sh
(cd build && cpack -G TZST)
tar x --zstd -f build/nvim-linux64.tar.zst -C /tmp/
/tmp/nvim-linux64/bin/nvim
```
and it works for me.

dundargoc:
> From non-appimage ways, deb is one easy uninstallable, like extracted tarball or make install-ed structure in some not shared directory.

Nice dodge Neo, I thought we were talking about benefits of deb over appimage specifically.

> I tried ... and it works for me.

Hmm, I see. The problem is that I don't think it's guaranteed to work for all package formats since there's a lot of code needed to support deb builds. If we want to add a note on this in general we should probably mention either that it might not work for all fornats, or just specify which formats does work.

BUILD.md
```diff
@@ -12,6 +12,7 @@
     - To build on Windows, see the [Building on Windows](#building-on-windows) section. _MSVC (Visual Studio) is recommended._
 4. `sudo make install`
     - Default install location is `/usr/local`
+    - On Debian/Ubuntu, instead of installing files directly with `sudo make install`, you can run `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This should help ensuring the clean removal of installed files.
```
justinmk, line 15:
```suggestion
    - On Debian/Ubuntu, instead of `sudo make install`, you can try `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This helps ensure clean removal of installed files.
        - Note: This is an unsupported, "best-effort" feature of the Nvim build.
```

BUILD.md
```diff
@@ -12,6 +12,8 @@
     - To build on Windows, see the [Building on Windows](#building-on-windows) section. _MSVC (Visual Studio) is recommended._
 4. `sudo make install`
     - Default install location is `/usr/local`
+    - On Debian/Ubuntu, instead of `sudo make install`, you can try `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This helps ensure clean removal of installed files.
+        - Note: This is an unsupported, "best-effort" feature of the Nvim build.
```
justinmk, line 16:
```suggestion
    - On Debian/Ubuntu, instead of `sudo make install`, you can try `cd build && cpack -G DEB && sudo dpkg -i nvim-linux64.deb` to build DEB-package and install it. This helps ensure clean removal of installed files. Note: This is an unsupported, "best-effort" feature of the Nvim build.
```

justinmk (approved):

dundargoc:

PR #30998, opened 2024-10-30
Ignore test files in root directory

yochem:
Problem: Running make test in the root folder produces:

- `/.swp`
- `/Xtest_logging`
- `/Xviminfo`

Solution: Ignore the test files for Git.


dundargoc:
IMO a better solution would be to not generate the files in the root in the first place. This might be a decent temporary solution though.

PR #30999, opened 2024-10-30
fix(lsp): Ignore ServerCancelled error when pulling lsp diagnostics

### LINKED ISSUE #30985: LSP: rust_analyzer: -32802: server cancelled the request
### Problem

editor is being interrupted with message `rust_analyzer: -32802: server cancelled the request` while typing on every keystroke after the latest updates of rust-analyzer

Just start typing in `main.rs` inside `fn main() {|}` of `cargo new --bin test_project`

```lua
local pattern = 'rust'
local cmd = { '/opt/homebrew/bin/rust-analyzer' }
local root_markers = { 'Cargo.toml' }

    pattern = pattern,
    callback = function(args)
        vim.lsp.start({
            name = 'rust-analyzer',
            cmd = cmd,
            root_dir = root_dir,
            settings = settings
        })
    end
})
```

### Expected behavior

There is no this message with interrupt

### Nvim version (nvim -v)

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1727870382

rust-analyzer 0.0.0 (3b3a87fe9 2024-10-27)

### Operating system/version

macos 15.1

### Log file

_No response_

Jesse-Bakker:
Fix #30985

### DIFF
--- a/runtime/lua/vim/lsp/diagnostic.lua
+++ b/runtime/lua/vim/lsp/diagnostic.lua
@@ -306,11 +306,20 @@ end
 --- )
 --- ```
 ---
----@param _ lsp.ResponseError?
+---@param error lsp.ResponseError?
 ---@param result lsp.DocumentDiagnosticReport
 ---@param ctx lsp.HandlerContext
 ---@param config vim.diagnostic.Opts Configuration table (see |vim.diagnostic.config()|).
-function M.on_diagnostic(_, result, ctx, config)
+function M.on_diagnostic(error, result, ctx, config)
+  if error ~= nil and error.code == protocol.ErrorCodes.ServerCancelled then
+    if error.data == nil or error.data.retriggerRequest == true then
+      local client = vim.lsp.get_client_by_id(ctx.client_id)
+      if client ~= nil then
+        client.request(ctx.method, ctx.params)
+      end
+    end
+    return
+  end
   if result == nil or result.kind == 'unchanged' then
     return
   end
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1514,7 +1514,7 @@ get_namespace({client_id}, {is_pull})
                      client. Defaults to push
 
                                           *vim.lsp.diagnostic.on_diagnostic()*
-on_diagnostic({_}, {result}, {ctx}, {config})
+on_diagnostic({error}, {result}, {ctx}, {config})
     |lsp-handler| for the method "textDocument/diagnostic"
 
     See |vim.diagnostic.config()| for configuration options. Handler-specific
@@ -1539,6 +1539,7 @@ on_diagnostic({_}, {result}, {ctx}, {config})
 <
 
     Parameters: ~
+      • {error}   (`lsp.ResponseError?`)
       • {result}  (`lsp.DocumentDiagnosticReport`)
       • {ctx}     (`lsp.HandlerContext`)
       • {config}  (`vim.diagnostic.Opts`) Configuration table (see
--- a/runtime/lua/vim/lsp/protocol.lua
+++ b/runtime/lua/vim/lsp/protocol.lua
@@ -174,6 +174,7 @@ local constants = {
     -- Defined by the protocol.
     RequestCancelled = -32800,
     ContentModified = -32801,
+    ServerCancelled = -32802,
   },
 
   -- Describes the content type that a client supports in various


runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -306,11 +306,15 @@ end
...
 ---
----@param _ lsp.ResponseError?
+---@param error lsp.ResponseError?
 ---@param result lsp.DocumentDiagnosticReport
 ---@param ctx lsp.HandlerContext
 ---@param config vim.diagnostic.Opts Configuration table (see |vim.diagnostic.config()|).
-function M.on_diagnostic(_, result, ctx, config)
+function M.on_diagnostic(error, result, ctx, config)
+  -- Ignore ServerCancelled error
+  if error ~= nil and error.code == -32802 then
```
MariaSolOs, line 315: Instead of hardcoding the code here you should be adding it to protocol's error codes: https://github.com/neovim/neovim/blob/b4599acbf8e30aa1234bc4041546f960dcc3a238/runtime/lua/vim/lsp/protocol.lua#L163

MariaSolOs:
I don't think this is the right fix. The LSP says that if the server sends this response we should look at the cancellation data and possibly retrigger the request: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticServerCancellationData. Kind of surprised that this isn't behind a client capability but this PR would be against what the protocol says we should do.

But even if we ignore the retriggering for now, I don't think we should check the code in the request handler but instead handle it here where we ignore `RequestCancelled` errors:
https://github.com/neovim/neovim/blob/b4599acbf8e30aa1234bc4041546f960dcc3a238/runtime/lua/vim/lsp/rpc.lua#L471-L490

MariaSolOs (changes requested):

Jesse-Bakker:
I went with handling the error in the request handler, because server-side cancellation is only defined for pull diagnostics (and semantic token requests, but that's behind a capability). The `RequestCancelled` error is for client-side cancellation, which is defined for all requests.

(It's not behind a client capability, because the cancellation was introduced in the same version as pull diagnostics, which means there should be no clients that implement pull diagnostics but not server-side cancellation)

mfussenegger (commented):
Could you also add a test case?

MariaSolOs:
> The LSP says that if the server sends this response we should look at the cancellation data and possibly retrigger the request: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticServerCancellationData. 

@mfussenegger any thoughts on this?

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -306,11 +306,20 @@ end
...
+---@param error lsp.ResponseError?
 ---@param result lsp.DocumentDiagnosticReport
 ---@param ctx lsp.HandlerContext
 ---@param config vim.diagnostic.Opts Configuration table (see |vim.diagnostic.config()|).
-function M.on_diagnostic(_, result, ctx, config)
+function M.on_diagnostic(error, result, ctx, config)
+  if error ~= nil and error.code == protocol.ErrorCodes.ServerCancelled then
+    if error.data == nil or error.data.retriggerRequest == true then
+      local client = vim.lsp.get_client_by_id(ctx.client_id)
+      if client ~= nil then
```
MariaSolOs, line 317: Maybe an assertion that `client` isn’t nil is more appropriate?

MariaSolOs:
Hey @Jesse-Bakker, it seems like a lot of people are still struggling with #30985. Are you having issues with this PR? Let us know and we can help :)

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -306,11 +306,20 @@ end
...
+        client.request(ctx.method, ctx.params)
```
lithammer, line 318: Not 100% sure but I think this should be:

```suggestion
        client:request(ctx.method, ctx.params)
```

gpanders:
>Kind of surprised that this isn't behind a client capability but this PR would be against what the protocol says we should do.

Extra surprising since there is a `serverCancelSupport` field in [`SemanticTokensClientCapabilities`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensClientCapabilities) (which we set to `false`!)

runtime/lua/vim/lsp/diagnostic.lua
gpanders: Is it possible that between the first request being sent and the server cancellation being received that the client has exited? If so, silently failing when `client` is nil seems like the better approach I think.

gpanders:
Will continue this in https://github.com/neovim/neovim/pull/31345. Thank you @Jesse-Bakker for your contribution!

MariaSolOs:
> > Kind of surprised that this isn't behind a client capability but this PR would be against what the protocol says we should do.
> 
> Extra surprising since there is a `serverCancelSupport` field in [`SemanticTokensClientCapabilities`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensClientCapabilities) (which we set to `false`!)

Yeah this feels like a gap in the protocol.

PR #31013, opened 2024-10-31
fix(lsp): hover border type can be string

nikolightsaber:
Border type can also be a string as defined in `api-win_config`


MariaSolOs (approved):

MariaSolOs:
For reference, this type matches what [`get_border_size`](https://github.com/MariaSolOs/neovim/blob/e4a74e986c40ca60ad49ea89020d6a9be36e928b/runtime/lua/vim/lsp/util.lua#L45) expects.

nikolightsaber:
> For reference, this type matches what [`get_border_size`](https://github.com/MariaSolOs/neovim/blob/e4a74e986c40ca60ad49ea89020d6a9be36e928b/runtime/lua/vim/lsp/util.lua#L45) expects.
Yes noticed that thanks for merging

PR #31016, opened 2024-10-31
ci: remove all reviewer related CI jobs

dundargoc:
It seems it's not really useful based on preliminary feedback.


gpanders:
I like having these. I have notifications for the repo disabled because I don't have time or bandwidth to triage every single issue or PR, but I'd like to be notified about PRs which touch certain areas.

If people don't like label based reviewers, maybe we could look at using CODEOWNERS instead?

dundargoc:
Codeowners only work for people with merge rights last time I checked, but it might be a decent tradeoff?

dundargoc:
OK, seems like this is still useful for people who aren't chronically online and enable all notifications for neovim. I think CODEOWNERS in its current form is a disaster and unacceptable, but perhaps there will show up a more useful mechanism that we can use in the future (users subscribing to a label perhaps?)

PR #19290, opened 2022-07-08
feat(defaults): enable mouse support

matveyt:
Closes #15521

Issue #15521, opened 2021-08-29
defaults: enable mouse support


matveyt:
### Feature already in Vim?

No

### Feature description

Set default to `set mouse=ar`

gpanders:
I don't think this should be a default. See the discussion [here](https://github.com/tpope/vim-sensible/issues/152) for some background.

justinmk:
We had this enabled but later reverted it. We can address most of the objections by:

- mapping `<A-leftmouse>` to "escape" mouse-mode
- mapping `<rightmouse>` to open a menu with copy/paste/etc (things people expect from their terminal)
    - ideally this would be driven by `:menu`. https://github.com/neovim/neovim/issues/12727#issuecomment-907880785
    - `mousemodel=popup` https://github.com/neovim/neovim/issues/14921

justinmk:
With https://github.com/neovim/neovim/pull/19170 there is now zero reason not to enable mouse by default. (see above)

justinmk:
# Resolution

- Oceania has always been at war with Eastasia. Nvim has always enabled 'mouse' by default.
- https://github.com/neovim/neovim/pull/6022 temporarily disabled 'mouse' with the intention of re-enabling it after we had a full answer for users who were used to the old behavior.
- The full answer is now provided by https://github.com/neovim/neovim/pull/19290 , specifically:
    - `:set mouse=nvi`
    - Pressing ":" (to enter cmdline-mode) temporarily _disables mouse_. Try it.
    - Right-click shows a popup menu with: `Cut`, `Copy`, `Paste`, `Disable mouse`
        - The `Disable mouse` item shows `:help default-mouse` which explains how to disable mouse permanently.

---
_Locked to keep the summary visible. You can always [chat](https://neovim.io/#chat) or open an issue if you have new information/topics to discuss._

PR #31038, opened 2024-11-02
feat(mouse)!: default 'mouse' to "a"

errael:
Problem: By default button2 does not paste to command line with a gui.
Solution: change 'mouse' default so it's consistent with vim.
BREAKING CHANGE: 'mouse' default to "a" (was "nvi").

-----

It currently works with `nvim TUI` since it seems to ignore the `'mouse'` `"c"` flag (a bug); the `gui`s honor this flag. In any event the current default `'mouse'` setting is contrary to the vim behavior.

`vim` doc
```
'mouse'			string	(default "", "a" for GUI and Win32,
					set to "a" or "nvi" in |defaults.vim|)
```

Running `vim` or `gvim` 
```
:verbose set mouse?
  mouse=a
Last set from /junk/install/vim/share/vim/vim91/defaults.vim line 84
```
At least on my system, the default is `'mouse'` set to `"a"`. That setting seems most consistent with `neovim`'s philosophy.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -4145,7 +4145,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 	listing continues until finished.
 
 							*'mouse'*
-'mouse'			string	(default "nvi")
+'mouse'			string	(default "a")
 			global
 	Enables mouse support. For example, to enable the mouse in Normal mode
 	and Visual mode: >vim
--- a/runtime/lua/vim/_meta/options.lua
+++ b/runtime/lua/vim/_meta/options.lua
@@ -4250,7 +4250,7 @@ vim.go.more = vim.o.more
 --- 'selectmode'	whether to start Select mode or Visual mode
 ---
 --- @type string
-vim.o.mouse = "nvi"
+vim.o.mouse = "a"
 vim.go.mouse = vim.o.mouse
 
 --- The window that the mouse pointer is on is automatically activated.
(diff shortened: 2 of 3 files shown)


errael:
ping @justinmk @zeertzjq

zeertzjq:
Please search existing discussions

zeertzjq (changes requested):
Please search and read existing discussions.

errael:
> Please search and read existing discussions.

Searching for `mouse` in discussions doesn't seem to bring up anything relevant. Other than #31018

zeertzjq:
- #15521
- #19290

errael:
Thanks, I took discussion too literally.

zeertzjq:
When mouse handling is disabled in Nvim, most terminal emulators still have their own mouse handling, which is why middle-click pasting still works. I guess most GUIs, on the other hand, don't do any mouse handling when Nvim has disabled it.

errael:
I'm confused, I'll need to read the references more clearly. Specifically the difference between terminal mouse handling and nvim mouse handling.

So the idea that button2 paste to command-line in the `gui`s us a bug in each of the `gui`s. I vaguely remember some of the discussion about mouse settings from the `vim` mailing list. It seemed like the "experts" had most of the complaints. I kind of thought that the idea with `neovim` was reasonable default. But then we all have our own definition. Anyway it's easy enough to set in my init.

PR #31040, opened 2024-11-02
fix(inccommand): ensure cursor is where it belongs

### LINKED ISSUE #30696: inccommand: nvim_buf_set_text causes cursor flickering
### Problem

When typing an argument of a previewable command in command line mode (only on some key presses), the cursor is briefly moved to the end of the text. This only occurs when the buffer text is changed using `vim.api.nvim_buf_set_text` in the preview callback.

Initially reported here (the recording might be useful): https://github.com/smjonas/inc-rename.nvim/issues/63

### Steps to reproduce

- `nvim --clean`
- `:e test.lua`

Paste the following Lua code:
```lua
local function replace_first_line_preview(opts, preview_ns, preview_buf)
  local replacement = opts.args
  vim.api.nvim_buf_set_text(0, 0, 0, 1, -1, { replacement })
  return 2
end

vim.api.nvim_create_user_command(
  "ReplaceFirstLine",
  function() end,
  { nargs = 1, preview = replace_first_line_preview }
)
```
- `:so %`
- start typing `:ReplaceFirstLine a`
- move cursor to the left so that it is no longer at the end of the command argument in the command line
- type multiple `a`'s (hold the `a` key if the flicker is not immediately visible)

Observe that while typing some `a`'s, the cursor is briefly moved to the right of the last `a` before being reset to the current position. position.

### Expected behavior

Cursor should not flicker.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-911+ga2008118a

n/a

### Operating system/version

Linux Mint 21.2

### Terminal name/version

Kitty 0.21.2

xterm-kitty

### Installation

appimage

luukvbaal:
Problem:  Inccommand preview callback may flush inaccurate cmdline cursor position.
Solution: Ensure cursor is where it belongs when doing command preview.

Fix #30696

### DIFF
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -2550,6 +2550,9 @@ static bool cmdpreview_may_show(CommandLineState *s)
     goto end;
   }
 
+  // Cursor may be at the end of the message grid rather than at cmdspos.
+  // Place it there in case preview callback flushes it. #30696
+  cursorcmd();
   // Flush now: external cmdline may itself wish to update the screen which is
   // currently disallowed during cmdpreview(no longer needed in case that changes).
   cmdline_ui_flush();
--- a/test/functional/ui/inccommand_user_spec.lua
+++ b/test/functional/ui/inccommand_user_spec.lua
@@ -508,6 +508,39 @@ describe("'inccommand' for user commands", function()
     feed(':Test')
     eq('nosplit', api.nvim_get_option_value('inccommand', {}))
   end)
+
+  it('does not flush intermediate cursor position at end of message grid', function()
+    exec_lua([[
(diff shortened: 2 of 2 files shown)


luukvbaal:
Not sure we can cover this with a test. I was only able to reproduce the problem in 1 out of 4 terminals I tried in the first place.

src/nvim/ex_getln.c
```diff
@@ -2550,6 +2550,7 @@ static bool cmdpreview_may_show(CommandLineState *s)
     goto end;
   }
 
+  cursorcmd();
```
famiu, line 2553: Should probably add a comment above referencing this PR just in case someone wonders why this line exists in the future

famiu (approved):

Issue #28772, opened 2024-05-16
`SafeState` or something similar could be really useful for `vim.ui_attach`


folke:
### Problem

I just read up about `SafeState` and something like this (`SafeUIState`) could be really useful for `vim.ui_attach` when triggered:

* right before `getchar`
* right before `cmdpreview_may_show`
* right before `redraw`

Especially the second one would be great, since the current work-around for this in Noice is not ideal.

When `cmdpreview` is active, `redraw`s are ignored, but with `vim.ui_attach` this is not ideal.

User enters `:%s/foo/bar`. The last `r` will correctly trigger `ui_attach` and will then show the correct preview.
However, it's not possible for the cmdline ui to redraw itself (since redraws are disabled when `cmdpreview` is active).

My current work-around is to use `nvim_feedkeys` with `<space><bs>`.

### Related issues

- [ ] https://github.com/neovim/neovim/issues/20311
- [ ] https://github.com/neovim/neovim/issues/20463

folke:
@luukvbaal maybe you have some ideas?

With all your recent changes, I was able to finally just queue the messages received in the `vim.ui_attach` callback and no longer need to do screen updates or `redraws` during.

zeertzjq:
What does "trigger `vim.ui_attach`" mean?

folke:
The last `r` will send an ext_cmdline message with the new cmdline

zeertzjq:
I'm confused. What does this have to do with `SafeState`?

luukvbaal:
I'm not sure I understand why a dedicated event is needed, won't flushing the desired state in the appropriate places be sufficient; like in #27950?

And for cmdpreview in particular, removing the redraw restriction is the way forward: #28510.

> right before redraw

> I was able to finally just queue the messages received in the `vim.ui_attach` callback and no longer need to do screen updates or `redraws` during.

Care to elaborate on these? I deliberately did not look at noice's implementation and it's workarounds when working on the default ext UI and instead aimed to resolve issues in the C core as I ran into them.

folke:
I'm not sure what #27950 achieves regarding my bug report?

> And for cmdpreview in particular, removing the redraw restriction is the way forward: https://github.com/neovim/neovim/issues/28510.

That would be great

> Care to elaborate on these?

[...]

PR #31041, opened 2024-11-02
fix(treesitter): remove redundant on_bytes callback

luukvbaal:
Problem:  Treesitter highlighter implements an on_bytes callback that
          just re-marks a buffer range for redraw. The edit that
          prompted the callback will already have done that.
Solution: Remove redundant on_bytes callback from the treesitter
          highlighter module.

### DIFF
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -1605,8 +1605,7 @@ LanguageTree:register_cbs({cbs}, {recursive})
       • {cbs}        (`table<TSCallbackNameOn,function>`) An
                      |nvim_buf_attach()|-like table argument with the
                      following handlers:
-                     • `on_bytes` : see |nvim_buf_attach()|, but this will be
-                       called after the parsers callback.
+                     • `on_bytes` : see |nvim_buf_attach()|.
                      • `on_changedtree` : a callback that will be called every
                        time the tree has syntactical changes. It will be
                        passed two arguments: a table of the ranges (as node
--- a/runtime/lua/vim/treesitter/highlighter.lua
+++ b/runtime/lua/vim/treesitter/highlighter.lua
@@ -93,9 +93,6 @@ function TSHighlighter.new(tree, opts)
   opts = opts or {} ---@type { queries: table<string,string> }
   self.tree = tree
   tree:register_cbs({
-    on_bytes = function(...)
-      self:on_bytes(...)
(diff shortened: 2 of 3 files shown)


clason:
> The edit that prompted the callback will already have done that.

Is this guaranteed for _any_ possible change? I wouldn't be surprised if some edits (like, say, creating a buffer with `set_lines`) won't...

luukvbaal:
Could be I'm missing context from why it was originally added but I think the splice/changed methods are called quite diligently throughout the code base, so I would be surprised. Happy to be proven wrong but the tests currently don't.

bfredl:
I think this is fine. `on_changedtree` will already mark the correct region for redraw (which might be larger than the raw edit event) anyway.

luukvbaal:
> I think this is fine. `on_changedtree` will already mark the correct region for redraw (which might be larger than the raw edit event) anyway.

Yes that was my understanding as well. Do you remember or have a clue why it was originally added?

justinmk (approved):

justinmk:
> Do you remember or have a clue why the `on_bytes` callback was originally added?

is there no discussion in the git history / prs ? if it was part of a big change, it's probably just an oversight.

do we need to mention this in breaking changes or no?

luukvbaal:
> is there no discussion in the git history / prs?

I couldn't find relevant discussion in the PR that added it.

I am not aware of it breaking anything. Should update the help on `on_bytes` in `treesitter.txt` though.

On the other hand, I think this callback, in conjuction with the `nvim__buf_redraw_range` to `nvim__redraw -> ui_flush` change is the main (only?) source of the `textlock` problems @folke mentioned in https://github.com/neovim/neovim/issues/28772#issuecomment-2119302411.

justinmk:
I looked through the history. This looks fine, let's try it.

folke:
> > is there no discussion in the git history / prs?
> 
> I couldn't find relevant discussion in the PR that added it.
> 
> I am not aware of it breaking anything. Should update the help on `on_bytes` in `treesitter.txt` though.
> 
> On the other hand, I think this callback, in conjuction with the `nvim__buf_redraw_range` to `nvim__redraw -> ui_flush` change is the main (only?) source of the `textlock` problems @folke mentioned in https://github.com/neovim/neovim/issues/28772#issuecomment-2119302411.

Will check tomorrow!

PR #31051, opened 2024-11-02
docs: use American English instead of British English

dundargoc:
This makes docs more consistent and more greppable.

This is primarily focused on words ending on "our" in British English.


clason:
Maybe add to the [cringelist](https://github.com/neovim/neovim/blob/86e54734bf9e05605d8b7146e7b0e79025138ba2/scripts/gen_help_html.lua#L37-L47)?

dundargoc:
Yeah, we might need a language police CI to ban british if we want to prevent regressions but maybe that's too much lol

bfredl (approved):
image

clason:
Do `neighbour` next?

justinmk:
biscuits => cookies, chips => fries, crumpets => english muffins

clason:
you forgot trousers => pants 😤

scripts/gen_help_html.lua
```diff
@@ -44,6 +44,14 @@ local spell_dict = {
   viml = 'Vimscript',
   ['tree-sitter'] = 'treesitter',
   ['Tree-sitter'] = 'Treesitter',
+  colour = 'color',
```
dundargoc, line 47: TODO: Add a title-case invariant check/regular word dict to avoid duplicating every word

craigmac:
makes sense,`:colourscheme` and some others were never a thing!

dundargoc:
I can't come up with a non-biased way to say "using 'ou' is super ugly and morally wrong" so I guess I'll just leave that out of the commit message.

zeertzjq:
These changes will make merging documentation in Vim patches harder

zeertzjq:
And... Why are you changing the oldtests as well?

clason:
...unless someone were to make the same changes in Vim. Presumably, they care about consistent documentation as well.

dundargoc:
I don't think this is a change vim would be interested in.

PR #31060, opened 2024-11-03
refactor(options): use os_win/os_buf for local options

zeertzjq:
Conversely, don't use them for global options.

### DIFF
--- a/src/nvim/option.c
+++ b/src/nvim/option.c
@@ -1886,7 +1886,7 @@ static const char *did_set_arabic(optset_T *args)
       // set rightleft mode
       if (!win->w_p_rl) {
         win->w_p_rl = true;
-        changed_window_setting(curwin);
+        changed_window_setting(win);
       }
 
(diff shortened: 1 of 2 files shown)


zeertzjq:
It's not currently clear whether `curwin` or `os_win` should be used for these three options:
- `'helpheight'`
- `'winheight'`
- `'winwidth'`

As these three are global options.

famiu:
`:help 'winheight'` and `:help 'winwidth'` talk about the current window, so perhaps `curwin` should be used?

famiu:
`'helpheight'` is a complex case, imo it should use `os_win`.

One thing worth nothing, `did_set_option` does set `os_win` to `curwin` (and `os_buf` to `curbuf`), so in most cases they're probably going to be the same value. In fact I can't think of many cases where it'd change.

zeertzjq:
Considering the `ONE_WINDOW` condition, I'm going to use `curwin` for all three, as `ONE_WINDOW` is for current tabpage.

PR #31062, opened 2024-11-03
refactor(options): remove fileformat macros 

famiu:


dundargoc:
Unsure if out of scope for this PR, but is "mac" fileformat even a thing anymore? Mac uses the same line endings as linux since a long time. I wonder if we have any plans to remove it.

famiu:
> Unsure if out of scope for this PR, but is "mac" fileformat even a thing anymore? Mac uses the same line endings as linux since a long time. I wonder if we have any plans to remove it.

I think `"mac"` just uses `<CR>` as line ending, or at least that's how it is in the codebase, so not exactly the same. And yes, it is out of scope for this PR

clason:
Yes, but no current macOS filesystem uses "mac" line endings. The question is whether we want to drop support for those (pre-Mac OS X) entirely.

PR #31063, opened 2024-11-03
ci: work around flaky python tests

dundargoc:
It's the same workaround as 88ed9ffcd1c022811ebc8818e153fe94e4741671,
which was later removed in f707ce76acb86e064d549dc762f5636af072d3c5
after it turned out to be stable after a while.


.github/workflows/test_windows.yml
```diff
@@ -31,6 +31,13 @@ jobs:
           cmake --preset ci -D CMAKE_BUILD_TYPE='RelWithDebInfo' ${{ inputs.build_flags }}
           cmake --build build
 
+      # FIXME(dundargoc): this workaround is needed for macos as the python3
```
zeertzjq, line 34: not macos

PR #31068, opened 2024-11-04
idris2: add filetype

srghma:
added as was suggested here https://github.com/nvim-treesitter/nvim-treesitter/pull/7277#issuecomment-2453391849

why use `vim.g.filetype_idr`? bc maybe the idris3 will be released in future, but idris2 rn is a most used version


zeertzjq:
Read the comment more carefully

PR #31070, opened 2024-11-04
fix(snippet): use indent function to generate base indent

glepnir:
Problem: we have indent function get space counts and respect expandtab

Solution: using indent to get count and generate base_indent

### DIFF
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -477,7 +477,9 @@ end
 function M.expand(input)
   local snippet = G.parse(input)
   local snippet_text = {}
-  local base_indent = vim.api.nvim_get_current_line():match('^%s*') or ''
+  local curbuf = vim.api.nvim_get_current_buf()
+  local expandtab = vim.bo[curbuf].expandtab
+  local base_indent = expandtab and (' '):rep(vim.fn.indent('.')) or ''
 
   -- Get the placeholders we should use for each tabstop index.
   --- @type table<integer, string>
@@ -518,8 +520,6 @@ function M.expand(input)
     end
 
     local shiftwidth = vim.fn.shiftwidth()
-    local curbuf = vim.api.nvim_get_current_buf()
-    local expandtab = vim.bo[curbuf].expandtab
 
     local lines = {} --- @type string[]
     for i, line in ipairs(text_to_lines(text)) do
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/snippet.lua
```diff
@@ -477,7 +477,9 @@ end
 function M.expand(input)
   local snippet = G.parse(input)
   local snippet_text = {}
-  local base_indent = vim.api.nvim_get_current_line():match('^%s*') or ''
+  local curbuf = vim.api.nvim_get_current_buf()
```
MariaSolOs, line 480: Instead of adding a new variable could you move the declaration of `bufnr` here?

https://github.com/neovim/neovim/blob/f330eaaeb35557521b608c41fd95a2bb2a76972b/runtime/lua/vim/snippet.lua#L586

runtime/lua/vim/snippet.lua
```diff
@@ -477,7 +477,9 @@ end
...
+  local expandtab = vim.bo[curbuf].expandtab
```
MariaSolOs, line 481: Nit: To keep the convention of using snake_case variable names
```suggestion
  local expand_tab = vim.bo[curbuf].expandtab
```

test/functional/lua/snippet_spec.lua
```diff
@@ -52,7 +52,7 @@ describe('vim.snippet', function()
     test_expand_success(
       { 'function $1($2)', '  $0', 'end' },
       { '  function ()', '    ', '  end' },
-      '',
+      'vim.opt.expandtab = true',
```
MariaSolOs, line 55: This isn't enough to check that the fix works is it?

runtime/lua/vim/snippet.lua
zeertzjq: But isn't that inconsistent with the option name?

runtime/lua/vim/snippet.lua
```diff
@@ -477,7 +477,9 @@ end
...
+  local base_indent = expandtab and (' '):rep(vim.fn.indent('.')) or ''
```
zsugabubus, line 482: Why is it empty with `noexpandtab`? And instead of manually constructing the indent wouldn't it be better to use `:left` that supports all kind of indent (e.g. `shiftwidth=4 tabstop=8`, `vartabstop`)?

glepnir: yep left make more sense insetead generate a indent

PR #31074, opened 2024-11-04
feat(float): add winborder  option

### LINKED ISSUE #20202: global option to set default floating window border
### Feature already in Vim?

No

### Feature description

**Why?**
Currently unifying the look and feel of neovim ui is kind of hard to deal with because of absence of such options.

From user perspective if you want something more unique then `single` or `rounded` borders you would need for every plugin that is using floating windows at some point to add few lines of code for controlling (if plugin allows this) its look.

From plugin author perspective you need to make it public such an option (complicate plugin config) instead of something like `nvim_open_win(.... { border = vim.get_float_border() }` and forget about it.

---
Because of `:h api-contract` and some limitation of vim options (can't be real lists or dicts) it's hard to deal with such problem.

<details><summary><b>What I suggest</b></summary>
Not sure if it is the best thing (maybe community/devs can find a better one solution).
Lets introduce two new options and one new function (`vim` or `vim.api`):
- first option is `floatborder` (or you name it) which can be one of already provided options of the border for `nvim_open_win` such as *none, single, double, rounded, solid, shadow* or any others plus one new option named *borderchars*.
- second option is `borderchars` itself with is a comma-separated characters (or maybe something like *fillchars* for more conrol) for the border the same as array for `nvim_open_win()` border option.
- our new function should just from `floatborder` and `borderchars` options generate string or array required for `nvim_open_win()` border option.

So at the end users will just need to change such two options to have the same borders. And plugin authors just to call the function to generate border option when they add `nvim_open_win()` in their plugin.
</details>

Any ideas how to improve this suggestion? What do you think overall about this feature request?

---
**EDIT** as **bfredl** suggested just some lua option like `vim.ui.border_style` seems even less work to implement this and for users and plugin authors to use.

glepnir:
Problem:
There is currently no global option to define the default border style for floating windows. This leads to repetitive code when developers need consistent styling across multiple floating windows.

Solution:
Introduce a global option winborder to specify the default border style for floating windows. When a floating window is created without explicitly specifying a border style, the value of the winborder option will be used. This simplifies configuration and ensures consistency in floating window appearance.

Fix #20202


famiu:
Just a minor nit about the title: There's no terminology like "global-window option", I think you probably meant global-local window options (yeah, I hate the terminology as well). I think just putting "feat(float): add `'winborder'` option" in the title is fine.

bfredl:
If we add this as an option I think it should actually be global-only, and be available as a default style for plugins to use if they want when opening the window.  

the _per window_ option value is redundant with the value in `win_config` which is already per window and has better semantics than window options for window layout related things (The semantics of window options are more for how a buffer is being displayed in a window).

seblj:
Are there plans to eventually extend this to support custom border characters as well?

runtime/lua/vim/_meta/options.lua
```diff
@@ -7759,6 +7759,15 @@ vim.o.winbl = vim.o.winblend
 vim.wo.winblend = vim.o.winblend
 vim.wo.winbl = vim.wo.winblend
 
+--- Configures the border style of floating windows, border values only
+--- support string type which defined in `nvim_open_win()`.
+---
+--- @type string
```
max397574, line 7765: if it only support certain values why not directly make this an enum?

glepnir: we will have a dict type of option value in future that support custom character .this is in plan.

max397574: ig this is meant https://github.com/neovim/neovim/pull/31084

lewis6991 (approved):

test/old/testdir/gen_opt_test.vim
```diff
@@ -359,6 +359,7 @@ let test_values = {
       "\ default behaviours
       \ 'othernum': [[-1, 0, 100], ['']],
       \ 'otherstring': [['', 'xxx'], []],
+      \ 'winborder': [['rounded', 'none', 'single', 'solid'], ['xxx']],
```
zeertzjq, line 362: Please add this at the `Nvim-only options` section instead of here.

runtime/doc/options.txt
```diff
@@ -7090,6 +7090,12 @@ A jump table for the options with a short description can be found at |Q_op|.
 
 	UI-dependent. Works best with RGB colors. 'termguicolors'
 
+						*'winborder'* *'winbd'*
+'winborder' 'winbd'	string	(default "")
+			global
+	Configures the border style of floating windows, border values only
+	support string type which defined in |nvim_open_win()|.
```
justinmk, line 7097:
```suggestion
    Defines the default border style of floating windows. See |nvim_open_win()|
    for valid values.
```

justinmk: We should probably move the list of border names here, and link to this from |nvim_open_win()|

src/nvim/option_vars.h
```diff
@@ -762,6 +762,7 @@ EXTERN int p_wa;                ///< 'writeany'
 EXTERN int p_wb;                ///< 'writebackup'
 EXTERN OptInt p_wd;             ///< 'writedelay'
 EXTERN int p_cdh;               ///< 'cdhome'
+EXTERN char *p_winbd;           ///< 'winborder'
```
justinmk, line 765: seems like many of these are alphabetical order, which also helps avoid merge conflicts.

this could go on line

src/nvim/options.lua
```diff
@@ -9915,6 +9915,20 @@ return {
       short_desc = N_('Controls transparency level for floating windows'),
       type = 'number',
     },
+    {
+      abbreviation = 'winbd',
```
justinmk, line 9919: if the abbreviations are hard to remember they are pretty useless, aren't they? do we actually need an abbreviation for this?

test/functional/ui/float_spec.lua
```diff
@@ -9518,6 +9518,116 @@ describe('float window', function()
         })
       end
     end)
+
```
justinmk, line 9521: also need a single "validation" tests that runs through all the valid values.

This also somewhat mitigates the fact that the values are repeated at `p_winbd_values` and [`parse_border_style`](https://github.com/neovim/neovim/blob/ac7e0ff32ff18be69e2469ff7fc66370be20a6e1/src/nvim/api/win_config.c#L933-L945)

```lua
    it('#winborder validation', function()
      command('set winborder=single')
      command('set winborder=double')
      command('set winborder=...')
    end)

File: src/nvim/options.lua (hunk 5 above)
Comment (glepnir) [reply src/nvim/options.lua:9919]: `wbd` should be sounds good ?

Comment (zeertzjq) [reply src/nvim/options.lua]: How is that less hard to remember than `winbd`?

File: test/old/testdir/gen_opt_test.vim [hunk 7]
```diff
@@ -139,6 +139,7 @@ let test_values = {
       \ 'wildcharm': [[-1, 0, 100, 'x', '^Y', '^@', '<Esc>', '<', '^'],
       \		['', 'xxx', '<xxx>', '<t_xxx>', '<Esc', '<t_xx', '<C-C>',
       \		'<NL>', '<CR>', K_KENTER]],
+      \ 'winborder': [['rounded', 'none', 'single', 'solid'], ['xxx']],
```
zeertzjq, line 142: Still wrong place. It should be in the `Nvim-only options` section.

glepnir: Ops I searched wrong and overlooked it. now should be right.

runtime/doc/api.txt
```diff
@@ -3228,28 +3228,20 @@ nvim_open_win({buffer}, {enter}, {config})                   *nvim_open_win()*
...
-                    • "shadow": A drop shadow effect by blending with the
-                      background.
-                    • If it is an array, it should have a length of eight or
-                      any divisor of eight. The array will specify the eight
-                      chars building up the border in a clockwise fashion
-                      starting with the top-left corner. As an example, the
-                      double box style could be specified as: >
-                      [ "╔", "═" ,"╗", "║", "╝", "═", "╚", "║" ].
+                    be a string or an array. The string values are the same as
+                    those described in |winborder|. If it is an array, it
```
justinmk, line 3232:
```suggestion
                    those described in 'winborder'. If it is an array, it
```

runtime/doc/api.txt
```diff
@@ -3228,28 +3228,20 @@ nvim_open_win({buffer}, {enter}, {config})                   *nvim_open_win()*
...
+                    should have a length of eight or any divisor of eight. The
+                    array will specify the eight chars building up the border
+                    in a clockwise fashion starting with the top-left corner.
+                    As an example, the double box style could be specified as: >
+                     [ "╔", "═" ,"╗", "║", "╝", "═", "╚", "║" ].
```
justinmk, line 3237: 'winborder' probably should also support this array form, but that can wait until @famiu's work

famiu: > 'winborder' probably should also support this array form, but that can wait until @famiu's work. Or maybe it can be a comma-separated option?

We could do something similar to what `'fillchars'` does

famiu: > comma-separated should be fine `set winborder=╔, ═ ,╗, ║, ╝, ═, ╚, ║ `. Similar to fillchars, it means that the user must declare the field name for each position like `set winborder=topleft:╔, top:═...` or i understand wrong 🤔

Yes that is what I suggested, since a list order may seem arbitrary. The `fillchars` syntax has the disadvantage of being unnecessarily verbose, but it follows an existing precedent so that's a plus imo.

glepnir: I think this should be implemented after we support `array/dict` option vlaues. when border is an array it has too many extension possibilities in `nvim_oepn_win`.. currently it is also possible to support all the formats described in `nvim_open_win`, but this may make the code a bit complicated and not very meaningful. If we have an explicit markup in the option value such as =[xx] it may be much easier.  or other smart way which i missed ?

justinmk:
@glepnir rebase? This PR is perhaps the most urgent of your existing PRs, imo.

gpanders:
I think we should add this to the 0.11 milestone to make the transition from https://github.com/neovim/neovim/pull/30935 a bit easier (users who override the global hover handler to add borders will suddenly find their hover windows have no borders when they upgrade. It would be great to add to the news entry that they can simply add `set winborder=rounded` to their config and move on).

bfredl (approved):

src/nvim/options.lua
```diff
@@ -10186,6 +10186,26 @@ local options = {
       short_desc = N_('Controls transparency level for floating windows'),
       type = 'number',
     },
+    {
+      defaults = { if_true = '' },
+      cb = 'did_set_winborder',
+      values = { 'double', 'single', 'shadow', 'rounded', 'solid', 'none' },
+      desc = [=[
+        Defines the default border style of floating windows. The default value
+        is empty, which is equivalent to "none". Valid values include:
```
gpanders, line 10195: Why make the default empty rather than `none`?

test/functional/ui/float_spec.lua
```diff
@@ -9909,6 +9909,253 @@ describe('float window', function()
...
+          {0:~ }{5:│}{1:der }{5:│}{0:                                }|
+          {0:~ }{5:│}{2:~   }{5:│}{0:                                }|
+          {0:~ }{5:└────┘}{0:                                }|
+                                                  |
+        ]])
+      end
+
+      -- it is currently not supported.
+      eq('Vim(set):E474: Invalid argument: winborder=custom', pcall_err(command, 'set winborder=custom'))
+    end)
```
przepompownia, line 10158: `nvim_win_set_config` has not been tested here (#32974).

src/nvim/api/win_config.c
```diff
@@ -1279,12 +1273,13 @@ static bool parse_win_config(win_T *wp, Dict(win_config) *config, WinConfig *fco
     }
   }
 
-  if (HAS_KEY_X(config, border)) {
+  if (HAS_KEY_X(config, border) || *p_winbd != NUL) {
```
zeertzjq, line 1276: This'll prevent creating split windows using `nvim_open_win` whenever `'winborder'` is set.

src/nvim/optionstr.c
```diff
@@ -2005,6 +2005,15 @@ const char *did_set_winhighlight(optset_T *args)
   return NULL;
 }
 
+/// The 'winborder' option is changed.
+const char *did_set_winborder(optset_T *args)
+{
+  if (opt_strings_flags(p_winbd, opt_winborder_values, NULL, true) != OK) {
```
zeertzjq, line 2011: Why is the last argument `true`? This is not a list option.

Actually the `did_set_winborder()` function is not needed anymore. `did_set_str_generic()` is automatically used when `values` is present.

zeertzjq: Fix in #32978

justinmk:
https://github.com/vim/vim/pull/17091 has code for the comma-separated form.

PR #31078, opened 2024-11-04
fix(editorconfig): rename augroup

tomtomjhj:
Problem:
Builtin editorconfig plugin's augroup name conflicts with https://github.com/editorconfig/editorconfig-vim, overriding that plugin.

Solution:
Rename the augroup.

### DIFF
--- a/runtime/lua/editorconfig.lua
+++ b/runtime/lua/editorconfig.lua
@@ -151,7 +151,7 @@ function properties.trim_trailing_whitespace(bufnr, val)
   )
   if val == 'true' then
     vim.api.nvim_create_autocmd('BufWritePre', {
-      group = 'editorconfig',
+      group = 'nvim_editorconfig',
       buffer = bufnr,
       callback = function()
         local view = vim.fn.winsaveview()
@@ -163,7 +163,7 @@ function properties.trim_trailing_whitespace(bufnr, val)
   else
     vim.api.nvim_clear_autocmds({
       event = 'BufWritePre',
-      group = 'editorconfig',
+      group = 'nvim_editorconfig',
       buffer = bufnr,
     })
   end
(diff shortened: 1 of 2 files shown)


clason:
What does that plugin offer over the builtin support? Is having it installed something we should seriously support?

max397574:
the readme of the plugin even says that you don't need to install it with the more recent neovim versions

image

tomtomjhj:
nvim's is missing automatic colorcolumn setup. https://github.com/editorconfig/editorconfig-vim/blob/ba2ce027c5b0e523e658d24657ce3ae3306c9fe0/doc/editorconfig.txt#L100-L120

dundargoc:
Is that even an official editorconfig option? I took a quick glance and couldn't find anything

gpanders:
> nvim's is missing automatic colorcolumn setup. https://github.com/editorconfig/editorconfig-vim/blob/ba2ce027c5b0e523e658d24657ce3ae3306c9fe0/doc/editorconfig.txt#L100-L120

`max_line_indicator` is a non-standard property, so we will not support it by default unless it's added to the specification. But you can add any arbitrary property yourself, if you like (see `:h editorconfig-custom-properties`).

gpanders:
That said, we really should use the `nvim_` prefix for the augroup anyway, as that's standard for other builtin autocommands.

gpanders (approved):

gpanders:
This is technically a breaking change so can you please update your commit message and PR title [appropriately](https://github.com/neovim/neovim/blob/3e855d533f4477ffddfe94ccea48255979c8b7fb/CONTRIBUTING.md?plain=1#L110-L114) and add an entry to the breaking changes section of news.txt?

justinmk (approved):

tomtomjhj:
subsumed by #32001

PR #31084, opened 2024-11-05
[WIP] refactor(options): autogenerate option variables

famiu:
Depends on #31317


famiu:
Seems like mixing integer keys and string keys for a table in `options.lua` causes `luaeval()` to fail for `test_options_all.vim`.
Are there any solutions to that?

PR #31086, opened 2024-11-05
feat(health): support show health in floating window

glepnir:
Problem: health can not shown in a floating window
Solution: add g:health variable

QQ_1730807070970

### DIFF
--- a/runtime/doc/health.txt
+++ b/runtime/doc/health.txt
@@ -21,6 +21,17 @@ To run all healthchecks, use: >vim
 <
 Plugin authors are encouraged to write new healthchecks. |health-dev|
 
+                                                                    *g:health*
+g:health  This global variable controls the behavior and appearance of the
+          `health` floating window. It should be a dictionary containing the
+          following optional keys:
+          - `style`: string? Determines the display style of the `health` window.
+                            Set to `'float'` to enable a floating window. Other
+                            styles are not currently supported.
+
+          Example: >lua
+            vim.g.health = { style = 'float' }
+
 Commands                                                     *health-commands*
 
                                                            *:che* *:checkhealth*
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -2034,7 +2034,7 @@ Lua module: vim.lsp.util                                            *lsp-util*
       • {zindex}?        (`integer`) override `zindex`, defaults to 50
       • {title}?         (`string`)
       • {title_pos}?     (`'left'|'center'|'right'`)
-      • {relative}?      (`'mouse'|'cursor'`) (default: `'cursor'`)
+      • {relative}?      (`'mouse'|'cursor'|'editor'`) (default: `'cursor'`)
       • {anchor_bias}?   (`'auto'|'above'|'below'`, default: `'auto'`) -
                          "auto": place window based on which side of the
                          cursor has more lines
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -348,6 +348,8 @@ UI
 • |vim.diagnostic.setqflist()| updates an existing quickfix list with the
   given title if found
 • |ui-messages| content chunks now also contain the highlight group ID.
+• |:checkhealth| can be display in a floating window and controlled by
+  the |g:health| variable.
 
 ==============================================================================
 CHANGED FEATURES                                                 *news-changed*
--- a/runtime/lua/vim/health.lua
+++ b/runtime/lua/vim/health.lua
@@ -11,6 +11,17 @@
 --- <
 --- Plugin authors are encouraged to write new healthchecks. |health-dev|
 ---
+---                                                              *g:health*
+--- g:health  This global variable controls the behavior and appearance of the
+---           `health` floating window. It should be a dictionary containing the
+---           following optional keys:
+---           - `style`: string? Determines the display style of the `health` window.
+---                             Set to `'float'` to enable a floating window. Other
+---                             styles are not currently supported.
+---
+---           Example: >lua
+---             vim.g.health = { style = 'float' }
+---
 --- Commands                                *health-commands*
 ---
 ---                                                              *:che* *:checkhealth*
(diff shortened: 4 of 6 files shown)


runtime/lua/vim/health.lua
```diff
@@ -327,17 +327,38 @@ end
 ---                            `:checkhealth vim.* nvim` will healthcheck `vim.lsp`, `vim.treesitter`
 ---                            and `nvim` modules.
 function M._check(mods, plugin_names)
+  local arg = ''
+  local t = vim.split(plugin_names, '%s')
```
glepnir, line 331: This is a simple implementation that does not change ex_checkhealth at the moment. This makes the meaning of the parameter a bit strange.  currently we don't have float modifier so ++flag is used. It might be better to modify ex_checkhelath and then add a new parameter here like`mods, plugin_names, arg`.  Is this better or add a float modifier?

src/nvim/ex_docmd.c
```diff
@@ -7626,8 +7626,25 @@ static void ex_checkhealth(exarg_T *eap)
     mods_len = add_win_cmd_modifiers(mods, &cmdmod, &multi_mods);
     assert(mods_len < sizeof(mods));
   }
+
+  char *plugin_name  = NULL;
+  char *e = strstr(eap->arg, "++f");
```
justinmk, line 7631: I think we've discussed a `:float` or `:popup` command that works like `:split` and `:tab`. Don't want a health-specific special case here.

Meanwhile to unblock this pr, for now we could support a `vim.g.health = { style = 'float' }` option.

runtime/lua/vim/health.lua
```diff
@@ -331,13 +331,24 @@ function M._check(mods, plugin_names)
...
   local emptybuf = vim.fn.bufnr('$') == 1 and vim.fn.getline(1) == '' and 1 == vim.fn.line('$')
 
-  -- When no command modifiers are used:
-  -- - If the current buffer is empty, open healthcheck directly.
-  -- - If not specified otherwise open healthcheck in a tab.
-  local buf_cmd = #mods > 0 and (mods .. ' sbuffer') or emptybuf and 'buffer' or 'tab sbuffer'
-
   local bufnr = vim.api.nvim_create_buf(true, true)
-  vim.cmd(buf_cmd .. ' ' .. bufnr)
+  if #mods > 0 and mods == 'float' then
```
justinmk, line 335: can we re-use the lsp floating window util? maybe extract it to a private `vim._util` module for now.

src/nvim/ex_docmd.c
glepnir: i have also consider a modifier like `float` but unsure how this modifier work bause position can be also configed?

runtime/lua/vim/health.lua
glepnir: `vim.lsp.open_floating_preview` take content and create float seems like not fit here

justinmk: `contents` is just used to guess the window size. we can just pass the healthcheck contents, what's the problem?

runtime/lua/vim/health.lua
```diff
@@ -331,13 +349,30 @@ function M._check(mods, plugin_names)
...
+  if
+    vim.g.health
+    and type(vim.g.health) == 'table'
+    and vim.tbl_get(vim.g.health, 'style') == 'float'
+  then
+    local syntax = vim.tbl_get(vim.g.health, 'syntax') or ''
+    local float_bufnr, float_winid = vim.lsp.util.open_floating_preview({}, syntax, {
+      height = math.floor(vim.o.lines * 0.8),
+      width = math.floor(vim.o.columns * 0.8),
+      border = vim.tbl_get(vim.g.health, 'border'),
```
glepnir, line 362: redundant ? vim.lsp.util.open_floating_preview.Opts have it, but we will have border option so can instead there i guess..

runtime/doc/health.txt
```diff
@@ -21,6 +21,23 @@ To run all healthchecks, use: >vim
...
 Plugin authors are encouraged to write new healthchecks. |health-dev|
 
+                                                                    *g:health*
+g:health  This global variable controls the behavior and appearance of the
+          `health` floating window. It should be a dictionary containing the
+          following optional keys:
+          - `style`: string? Determines the display style of the `health` window.
+                            Set to `'float'` to enable a floating window. Other
+                            styles are not currently supported.
+          - `syntax`: string? Specifies the syntax highlighting to use
```
justinmk, line 31: why would this be overridable? checkhealth has its own filetype, people can extend that in the usual ways.

runtime/doc/health.txt
```diff
@@ -21,6 +21,23 @@ To run all healthchecks, use: >vim
...
+                            for the content of the `health` window.
+                            Defaults to an empty string if not provided.
+          - `border`: (string|table) Configures the border style of the floating window.
```
justinmk, line 34: Don't need this if we get https://github.com/neovim/neovim/pull/31074 , right?

glepnir: yes

runtime/doc/news.txt
```diff
@@ -348,6 +348,8 @@ UI
 • |vim.diagnostic.setqflist()| updates an existing quickfix list with the
   given title if found
 • |ui-messages| content chunks now also contain the highlight group ID.
+• |health| can be displayed in a floating window and controlled by
```
justinmk, line 351:
```suggestion
• |:checkhealth| can display in a floating window, controlled by
```

runtime/doc/health.txt
```diff
@@ -21,6 +21,16 @@ To run all healthchecks, use: >vim
 <
 Plugin authors are encouraged to write new healthchecks. |health-dev|
 
+                                                                    *g:health*
```
justinmk, line 24: this is similar to `g:editorconfig` 👍

runtime/doc/lsp.txt
```diff
@@ -2034,7 +2034,8 @@ Lua module: vim.lsp.util                                            *lsp-util*
       • {zindex}?        (`integer`) override `zindex`, defaults to 50
       • {title}?         (`string`)
       • {title_pos}?     (`'left'|'center'|'right'`)
-      • {relative}?      (`'mouse'|'cursor'`) (default: `'cursor'`)
+      • {pos}?           (`'center'`)
+      • {relative}?      (`'mouse'|'cursor'|'editor'`) (default: `'cursor'`)
```
justinmk, line 2038: doesn't `pos` conflict with `relative`? it can't be relative to cursor but also centered...

let's add `center` to `relative` and drop `pos`?

justinmk (approved):
Looks quite minimal 👍 

Add a small test to `health_spec.lua` ?

runtime/doc/lsp.txt
glepnir: I think relative in there is same as relative in nvim_open_win. in nvim_open_win there is no center

glepnir: removed.  just use offset to make it center

runtime/lua/vim/lsp/util.lua
```diff
@@ -1492,9 +1494,10 @@ end
 --- @field zindex? integer override `zindex`, defaults to 50
 --- @field title? string
 --- @field title_pos? 'left'|'center'|'right'
+--- @field pos? 'center'
```
justinmk, line 1497: is this still needed/used?

runtime/doc/health.txt
```diff
@@ -21,6 +21,19 @@ To run all healthchecks, use: >vim
...
+g:health  This global variable controls the behavior and appearance of the
+          `health` floating window. It should be a dictionary containing the
+          following optional keys:
+          - `style`: string? Determines the display style of the `health` window.
+                            Set to `'float'` to enable a floating window. Other
+                            styles are not currently supported.
+          - `max_height`: integer? Max height of floating window
+          - `max_width`: integer? Max width of floating window
```
justinmk, line 32: Want to avoid surface area where possible. max_height/width should be dropped for now. Just choose something reasonable.

glepnir: Just thought some people might have some fixed settings for height and width.  cleaned up

dundargoc:
@justinmk worth documenting that `q` closes the checkhealth buffer for the float version but not regular version or do we consider this implied?

clason:
That was already changed...

dundargoc:
Except this PR unchanges it so that `q` doesn't close the non-float version of checkhealth, but does close the float version.

clason:
Ah, that's a a merge fail then. I don't think that's intended.

dundargoc:
You don't think? I assumed this is intended, since it treats the non-float checkhealth as a regular buffer but the float version as a "float" buffer that can be closed with `q` as with lsp hovers. I think this change makes sense. I'll go ahead and approve it lol

dundargoc (approved):

clason:
> You don't think?

Given the timing (float PR from November, buffer change this week) I don't. Could be wrong, of course -- @justinmk will set things straight.

runtime/doc/health.txt
```diff
@@ -21,6 +21,17 @@ To run all healthchecks, use: >vim
 <
 Plugin authors are encouraged to write new healthchecks. |health-dev|
 
+                                                                    *g:health*
+g:health  This global variable controls the behavior and appearance of the
+          `health` floating window. It should be a dictionary containing the
+          following optional keys:
+          - `style`: string? Determines the display style of the `health` window.
```
justinmk, line 28: This is a temporary hack until we have `:help preview-popup` or possibly a `:float` modifier. https://github.com/neovim/neovim/issues/31902

justinmk (approved):
I will tweak the docs in my docs PR

PR #31089, opened 2024-11-05
refactor(options): autogenerate valid values and flag enums for options

famiu:
Problem: Option metadata like list of valid values for an option and option flags are not listed in the `options.lua` file and are instead manually defined in C, which means option metadata is split between several places.

Solution: Put metadata such as list of valid values for an option and option flags in `options.lua`, and autogenerate the corresponding C variables and enums.

Supersedes #28659


famiu:
@zeertzjq 
Seems like mixing integer keys and string keys for a table in `options.lua` causes `luaeval()` to fail for `test_options_all.vim`.
Are there any solutions to that?

zeertzjq:
Change `[key] = value` to `{ key, value }`?

src/nvim/options.lua
```diff
@@ -9529,6 +9733,15 @@ return {
...
       cb = 'did_set_virtualedit',
       defaults = { if_true = '' },
+      values = { 'block', 'insert', 'all', 'onemore', 'none', 'NONE' },
+      flags = {
+        block = 5,
+        insert = 6,
+        all = 4,
+        onemore = 8,
+        none = 16,
+        noneu = 32,
```
zeertzjq, line 9743:
```suggestion
        Block = 5,
        Insert = 6,
        All = 4,
        Onemore = 8,
        None = 16,
        NoneU = 32,
```

src/clint.py
```diff
@@ -899,7 +899,8 @@ def CheckIncludes(filename, lines, error):
                     not name.endswith('_defs.h') and
                     not name.endswith('h.inline.generated.h') and
                     not name.endswith('_defs.generated.h') and
-                    not name.endswith('_enum.generated.h')):
+                    not name.endswith('_enum.generated.h') and
+                    not name.endswith('_vars.generated.h')):
```
zeertzjq, line 903: No

zeertzjq, line 903: `option_vars.generated.h` not equivalent to a `_defs` header, as it contains `EXTERN` variables.

src/nvim/option_vars.h
```diff
@@ -5,6 +5,10 @@
 #include "nvim/sign_defs.h"
 #include "nvim/types_defs.h"
 
+#ifdef INCLUDE_GENERATED_DECLARATIONS
+# include "option_vars.generated.h"
```
zeertzjq, line 9:
```suggestion
# include "option_vars.generated.h"  // NOLINT(build/include_defs)
```

famiu:
`NOLINT(build/include_defs)` doesn't seem to work. I remember this being the case the last time I tried to do that as well

zeertzjq:
Hmm, let me try to fix that

zeertzjq:
Fixed in #31821

src/nvim/drawline.c
```diff
@@ -40,6 +40,7 @@
 #include "nvim/memory.h"
 #include "nvim/move.h"
 #include "nvim/option.h"
+#include "nvim/option_defs.h"
```
zeertzjq, line 43: Not needed anymore

zeertzjq, line 43:
```suggestion
```

src/nvim/generators/gen_options.lua
```diff
@@ -138,6 +144,91 @@ map_w('static ' .. hashfun)
 
 opt_map_fd:close()
 
+vars_w('// IWYU pragma: private, include "nvim/option_defs.h"')
```
zeertzjq, line 147:
```suggestion
vars_w('// IWYU pragma: private, include "nvim/option_vars.h"')
```

src/nvim/generators/gen_options.lua
```diff
@@ -138,6 +144,91 @@ map_w('static ' .. hashfun)
...
+  end
+end
+
+-- Generate valid values for each option.
+for _, option in ipairs(options_meta) do
+  --- @type function
+  local preorder_traversal
+  --- @param prefix string
+  --- @param values vim.option_valid_values
+  preorder_traversal = function(prefix, values)
```
zeertzjq, line 196: Why not just use `local function preorder_traversal(prefix, values)`?

src/nvim/options.lua
```diff
@@ -2077,6 +2140,25 @@ return {
...
+        'horizontal',
+        'vertical',
+        'closeoff',
+        'hiddenoff',
+        'foldcolumn:',
+        'followwrap',
+        'internal',
+        'indent-heuristic',
+        'linematch:',
+        { 'algorithm', { 'myers', 'minimal', 'patience', 'histogram' } },
```
zeertzjq, line 2160: The colon after `algorithm` is now missing in the cmdline completion of `:set diffopt=`. The colon needs to be added here, and then something needs to be done in `gen_options.lua` to avoid syntax error in the generated file.

famiu: Good catch, should be fixed now

src/nvim/optionstr.c
```diff
@@ -68,84 +68,6 @@ static const char e_wrong_number_of_characters_for_field_str[]
...
-static char *(p_cb_values[]) = { "unnamed", "unnamedplus", NULL };
-static char *(p_cmp_values[]) = { "internal", "keepascii", NULL };
-// Note: Keep this in sync with fill_culopt_flags()
-static char *(p_culopt_values[]) = { "line", "screenline", "number", "both", NULL };
-static char *(p_dy_values[]) = { "lastline", "truncate", "uhex", "msgsep", NULL };
-static char *(p_fdo_values[]) = { "all", "block", "hor", "mark", "percent", "quickfix", "search",
-                                  "tag", "insert", "undo", "jump", NULL };
-// Note: Keep this in sync with spell_check_sps()
-static char *(p_sps_values[]) = { "best", "fast", "double", "expr:", "file:", "timeout:", NULL };
-/// Also used for 'viewoptions'!  Keep in sync with SSOP_ flags.
```
zeertzjq, line 97: Maybe this `Also used for 'viewoptions'` comment should also be moved to `options.lua`?

famiu:
Noticed that `culopt` and `wim` flags weren't moved yet, so moved those as well

src/nvim/ui.c
```diff
@@ -325,7 +325,7 @@ void ui_busy_stop(void)
 
 /// Emit a bell or visualbell as a warning
 ///
-/// val is one of the BO_ values, e.g., BO_OPER
+/// val is one of the OptBoFlag values, e.g., kOptBoFlagOperator
```
zeertzjq, line 328:
```suggestion
/// val is one of the OptBoFlags values, e.g., kOptBoFlagOperator
```

zeertzjq (approved):

PR #31093, opened 2024-11-06
feat(lua): vim.on_key() can disable mapping

errael:
Problem: emulating vim popup, for filter, requires temporarily disabling mapping. https://github.com/neovim/neovim/issues/30741#issuecomment-2451195906

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -1398,6 +1398,9 @@ vim.on_key({fn}, {ns_id}, {opts})                               *vim.on_key()*
       • {ns_id}  (`integer?`) Namespace ID. If nil or 0, generates and returns
                  a new |nvim_create_namespace()| id.
       • {opts}   (`table?`) Optional parameters
+                 • {mapping}? (boolean, default true) When false, key mapping
+                   is disabled; use with caution. If all the keys are
+                   discarded, set this and mappings do not get in the way.
 
     Return: ~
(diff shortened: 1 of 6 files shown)


errael:
@zeertzjq @justinmk Any takers?

justinmk:
Haven't digested this yet, but keeping it rebased and continuing to simplify it, will help. Else we have https://github.com/neovim/neovim/issues/30741 to track it.

errael:
> Haven't digested this yet, but keeping it rebased and continuing to simplify it, will help. Else we have #30741 to track it.

Looks like `runtime/lua/vim/_editor.lua` has moved to `runtime/lua/vim/_core/editor.lua`.

Updating this PR will take a while; it's been over a year since I've looked at neovim and related stuff. Initial examination feels like the change to `editor.lua` will just plug in. But I want to review all the popup stuff.

errael:
It's now simple to use the nvim popup API which in turn uses this feature, see #38185 and [popup.nvim](https://github.com/errael/popup.nvim).

> ... keeping it rebased and continuing to simplify it, will help. ...

Is there some area where it looks like simplification is needed?

errael:
As this is for vim's popup_* API, it no longer seems relevant. The vim docs specifically say that disabling mapping should only be done in limited circumstances, in particular when a popup window eats all keys. In `vim`, `:help popup-mapping` for more discussion.

PR #31098, opened 2024-11-06
docs!: standardize doc for ChanInfo/ChanOpen autocmds/v:event

errael:
Problem:
ChanInfo/ChanOpen autocmd doc not standard; their v:event key not documented in :he v:event.
The name of the v:event key for ChanInfo/ChanOpen is too ambiguous.

Solution:
This PR is in two commits. Update the docs. Change key name to "chan_info" from "info".

BREAKING CHANGE: Change ChanInfo/ChanOpen v:event key name to "chan_info" from "info".


clason:
`:h api-contract`

PR #31103, opened 2024-11-07
perf(mouse): only generate a <MouseMove> for a new character positon

### LINKED ISSUE #30965: too many <MouseMove> events when using `&mousemoveevent`
### Problem

When `'mousemoveevent'` is true `neovim` can generate dozens of mouse events without changing the character position.

Sample output excerpt, using the test below, showing 30 events at line 13, col 27
```
Move 199 line 13, col 27                                                                            
Move 200 line 14, col 27                                                                            
...
Move 229 line 14, col 27                                                                            
Move 230 line 15, col 27                                                                            
```

### Steps to reproduce

To see the problem, source the following, move the mouse around, examine `:messages`
```lua
local ev = 0
vim.keymap.set('n', '<MouseMove>', function()
    local mp = vim.fn.getmousepos()
    ev = ev + 1
    vim.print(string.format("Move %d line %d, col %d", ev, mp.screenrow, mp.screencol)) 
end)
vim.o.mousemoveevent = true
```

### Expected behavior

Expect only one event per change of mouse character position

In `vim`'s `gui.c::gui_send_mouse_event()` there's
```
// Don't report a mouse move unless moved to a
// different character position.
if (button == MOUSE_MOVE)
```
There's something similar to this in `neovim`'s `mouse.c::jump_to_mouse()` but that seems unrelated to the mouse move event.

I vaguely recall seeing that fractional cells are a thing; but I don't see a way to get the fractional mouse position if this is the idea.

### Nvim version (nvim -v)

https://github.com/neovim/neovim/commit/54249d051c3e0b97321939e6a3ae3d2e83971ce7

No. Only one event per character pos change since feature introduction: https://github.com/vim/vim/commit/c4cb544cd5beaa864b3893e4b8d0085393c7dbce

### Operating system/version

ubuntu

### Terminal name/version

gnome-terminal

xterm-256color

### Installation

make install

errael:
Fix #30965

Problem: Can receive dozens of <MouseMove> events for same character position. #30965

Solution: Leverage check_multiclick to detect position unchanged.


src/nvim/os/input.c
```diff
@@ -367,12 +371,14 @@ static uint8_t check_multiclick(int code, int grid, int row, int col)
   orig_mouse_row = row;
 
   uint8_t modifiers = 0;
-  if (orig_num_clicks == 2) {
-    modifiers |= MOD_MASK_2CLICK;
-  } else if (orig_num_clicks == 3) {
-    modifiers |= MOD_MASK_3CLICK;
-  } else if (orig_num_clicks == 4) {
-    modifiers |= MOD_MASK_4CLICK;
+  if (code != KE_MOUSEMOVE) {
```
zeertzjq, line 374: This condition is not needed. When `code == KE_MOUSEMOVE` the function has already returned above.

errael: It only returns above when there has been no move. When there has been a move it continues on since the current position must be saved.

After saving the current position there could be
```
if (code == KE_MOUSEMOVE) { return 0; }
```
but that looked messier.

zeertzjq: Oops

test/functional/ui/mouse_spec.lua
```diff
@@ -1913,14 +1913,28 @@ describe('ui/mouse/input', function()
...
     command('nnoremap <MouseMove> <Cmd>let g:mouse_move += 1<CR>')
     command('nnoremap <2-MouseMove> <Cmd>let g:mouse_move2 += 1<CR>')
-    feed('<MouseMove><0,0>')
-    feed('<MouseMove><0,0>')
-    api.nvim_input_mouse('move', '', '', 0, 0, 0)
-    api.nvim_input_mouse('move', '', '', 0, 0, 0)
+    feed('<MouseMove><1,0>')
+    feed('<MouseMove><2,0>')
+    api.nvim_input_mouse('move', '', '', 0, 0, 3)
+    api.nvim_input_mouse('move', '', '', 0, 0, 4)
```
zeertzjq, line 1919: The description of this test should be changed. Maybe to this:
```
<MouseMove> to different locations can be mapped
```

errael: DONE

errael:
There's a failure 
```
FAIL: /home/runner/work/neovim/neovim/src/nvim/os/input.c (File size changed from 16846 to 16853)
```
Where the from size is what I see with "ls -l". Either I can't reproduce this at home, I'm using the wrong command or I'm looking in the wrong spot for the error file.

Suggestions?

zeertzjq:
Did you run `make format`?

errael:
> Did you run `make format`?

And of course it's documented in `CONTRIBUTING.md`. Thanks.

src/nvim/os/input.c
```diff
@@ -289,8 +289,7 @@ size_t input_enqueue(String keys)
     unsigned new_size
       = trans_special(&ptr, (size_t)(end - ptr), (char *)buf, FSK_KEYCODE, true, NULL);
 
-    if (new_size) {
-      new_size = handle_mouse_event(&ptr, buf, new_size);
+    if (new_size && (new_size = handle_mouse_event(&ptr, buf, new_size))) {
```
zeertzjq, line 292: The `continue;` should still be executed if `handle_mouse_event` returns `0`, otherwise if a key notation follows `<MouseMove>` (doesn't happen in Nvim TUI, but still possible) it won't be interpreted.

errael: Oh crap. DONE.

zeertzjq (approved):

PR #31105, opened 2024-11-07
fix(api): trigger buffer events for quickfix buffer updates

### LINKED ISSUE #31085: for quickfix, `vim.api.nvim_buf_attach()` `on_bytes` doesn't get called on `:caddexpr`
### Problem

I'm trying to use a plugin to colorize ANSI escape sequences in the quickfix window. The plugin needs to get a callback when the quickfix window updates. However `vim.api.nvim_buf_attach()` `on_bytes` doesn't get called on `:caddexpr`.

(Plugin issue at <https://github.com/m00qek/baleia.nvim/issues/27>).

### Steps to reproduce

The `minimal.lua`:

```lua
vim.cmd('copen')

for _, buffer in ipairs(vim.api.nvim_list_bufs()) do
  if vim.api.nvim_get_option_value('filetype', { buf = buffer }) == 'qf' then
    vim.api.nvim_buf_attach(buffer, false, {
      on_bytes = function()
        print('on_bytes')
      end,
    })
  end
end
```

- First run `/usr/bin/nvim --clean -u minimal.lua`.
- Inside *Neovim*, run `:caddexpr "foo"`.

### Expected behavior

The `on_bytes` callback runs.

### Nvim version (nvim -v)

0.10.2

not relevant

### Operating system/version

Arch Linux kernel 6.11.6

### Terminal name/version

wezterm 20240203-110809-5046fc22

wezterm

### Installation

system package manager

glepnir:
Problem: Buffer events (specifically on_bytes callbacks) weren't triggered when the 
quickfix list was modified, preventing buffer change notifications.

Solution: Add code to send both bytes and lines change notifications after 
quickfix buffer updates to properly trigger all attached callbacks.

Fix #31085

### DIFF
--- a/src/nvim/quickfix.c
+++ b/src/nvim/quickfix.c
@@ -16,6 +16,7 @@
 #include "nvim/autocmd_defs.h"
 #include "nvim/buffer.h"
 #include "nvim/buffer_defs.h"
+#include "nvim/change.h"
 #include "nvim/charset.h"
 #include "nvim/cursor.h"
 #include "nvim/drawscreen.h"
@@ -31,6 +32,7 @@
 #include "nvim/ex_eval.h"
 #include "nvim/ex_eval_defs.h"
 #include "nvim/ex_getln.h"
+#include "nvim/extmark.h"
 #include "nvim/fileio.h"
 #include "nvim/fold.h"
 #include "nvim/garray.h"
@@ -50,6 +52,7 @@
 #include "nvim/message.h"
 #include "nvim/move.h"
 #include "nvim/normal.h"
+#include "nvim/ops.h"
 #include "nvim/option.h"
 #include "nvim/option_defs.h"
 #include "nvim/option_vars.h"
@@ -4149,6 +4152,8 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
   }
 
   linenr_T old_line_count = buf->b_ml.ml_line_count;
+  colnr_T old_endcol = ml_get_buf_len(buf, old_line_count);
+  bcount_t old_bytecount = get_region_bytecount(buf, 1, old_line_count, 0, old_endcol);
   int qf_winid = 0;
 
   win_T *win;
@@ -4182,7 +4187,25 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
   qf_update_win_titlevar(qi);
 
   qf_fill_buffer(qf_get_curlist(qi), buf, old_last, qf_winid);
-  buf_inc_changedtick(buf);
+
+  linenr_T new_line_count = buf->b_ml.ml_line_count;
+  colnr_T new_endcol = ml_get_buf_len(buf, new_line_count);
+  bcount_t new_byte_count;
+
+  linenr_T delta = new_line_count - old_line_count;
+  if (old_last == NULL) {
+    new_byte_count = get_region_bytecount(buf, 1, new_line_count, 0, new_endcol);
+    extmark_splice(buf, 0, 0, old_line_count - 1, 0, old_bytecount, new_line_count - 1, new_endcol,
+                   new_byte_count, kExtmarkNoUndo);
+    changed_lines(buf, 1, 0, old_line_count > 0 ? old_line_count + 1 : 1, delta, true);
+  } else {
+    linenr_T start_lnum = old_line_count + 1;
+    new_byte_count = get_region_bytecount(buf, start_lnum, new_line_count, 0, new_endcol);
+    extmark_splice(buf, old_line_count - 1, old_endcol, 0, 0, 0, delta, new_endcol, new_byte_count,
+                   kExtmarkNoUndo);
+    changed_lines(buf, start_lnum, 0, new_line_count + 1, delta + 1, true);
+  }
+  buf->b_changed = false;
 
   if (old_last == NULL) {
     qf_win_pos_update(qi, 0);
--- a/test/functional/lua/buffer_updates_spec.lua
+++ b/test/functional/lua/buffer_updates_spec.lua
@@ -420,6 +420,35 @@ describe('lua buffer event callbacks: on_lines', function()
     feed('i<Tab>')
     eq({ '\ta' }, exec_lua('return _G.res[#_G.res]'))
   end)
+
+  it('quickfix buffer send change', function()
+    command('copen')
+    exec_lua(function()
+      vim.api.nvim_buf_attach(vim.api.nvim_get_current_buf(), false, {
+        on_lines = function(...)
+          vim.g.qf_on_lines = { ... }
+        end,
+        on_bytes = function(...)
+          vim.g.qf_on_bytes = { ... }
+        end,
+      })
+    end)
+    command('caddexpr "foo"')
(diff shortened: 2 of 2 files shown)


src/nvim/quickfix.c
```diff
@@ -4184,6 +4189,33 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
   qf_fill_buffer(qf_get_curlist(qi), buf, old_last, qf_winid);
   buf_inc_changedtick(buf);
 
+  if (buf_updates_active(buf)) {
```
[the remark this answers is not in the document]
glepnir: updated. but I am not familiar with the code related to buffer events. not sure if it is correct.😮‍💨

luukvbaal (commented):
Still rather have @bfredl look at and sign off on this.

src/nvim/quickfix.c
```diff
@@ -4182,7 +4187,30 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
   qf_update_win_titlevar(qi);
 
   qf_fill_buffer(qf_get_curlist(qi), buf, old_last, qf_winid);
-  buf_inc_changedtick(buf);
+
+  linenr_T new_line_count = buf->b_ml.ml_line_count;
```
justinmk, line 4191: Instead of all the new code (below), could we maybe change `qf_fill_buffer` to use `nvim_buf_set_lines` or `nvim_buf_set_text`, which already call `extmark_splice` + `changed_lines` in the expected way?

I guess not, because it looks like `qf_fill_buffer` is doing extra behavior per-line...

src/nvim/quickfix.c
```diff
@@ -4149,6 +4152,8 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
   }
 
   linenr_T old_line_count = buf->b_ml.ml_line_count;
+  colnr_T old_endcol = ml_get_buf_len(buf, old_line_count);
+  bcount_t old_byte_count = get_region_bytecount(buf, 1, old_line_count, 0, old_endcol);
```
justinmk:
```suggestion
  bcount_t old_bytecount = get_region_bytecount(buf, 1, old_line_count, 0, old_endcol);
```

src/nvim/quickfix.c
```diff
@@ -4182,7 +4187,30 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
...
   qf_fill_buffer(qf_get_curlist(qi), buf, old_last, qf_winid);
-  buf_inc_changedtick(buf);
+
+  linenr_T new_line_count = buf->b_ml.ml_line_count;
+  colnr_T new_endcol = ml_get_buf_len(buf, new_line_count);
+  bcount_t new_byte_count;
+
+  if (old_last == NULL) {
+    new_byte_count = get_region_bytecount(buf, 1, new_line_count, 0, new_endcol);
+    extmark_splice(buf, 0, 0, old_line_count - 1, 0, old_byte_count, new_line_count - 1,
```
justinmk: is `old_line_count` guaranteed to be >=1 ?

test/functional/lua/buffer_updates_spec.lua
```diff
@@ -420,6 +420,35 @@ describe('lua buffer event callbacks: on_lines', function()
     feed('i<Tab>')
     eq({ '\ta' }, exec_lua('return _G.res[#_G.res]'))
   end)
+
+  it('quickfix buffer send change', function()
```
justinmk, line 424: this test file is for the lua interface to bufupdates, whereas the fix is a low-level change, so probably fits better in `api/buffer_updates_spec`. perhaps after the diff-mode test here: https://github.com/neovim/neovim/blob/76d213efbee8e6857b3aa397827fa8756a327430/test/functional/api/buffer_updates_spec.lua#L584

#31085 mentions `on_bytes`, but the fix in this PR is simply to enable buffer-updates, so there is no need to test all the "lua callback" functionality. Instead we just need to verify that buffer updates are working in quickfix bufs (using `expectn('nvim_buf_lines_event',` or whatever common practice is in `api/buffer_updates_spec`).

src/nvim/quickfix.c
```diff
@@ -4182,7 +4187,30 @@ static void qf_update_buffer(qf_info_T *qi, qfline_T *old_last)
...
+
+    changed_lines(buf, 1, 0, old_line_count > 0 ? old_line_count + 1 : 1,
+                  new_line_count - old_line_count, true);
+  } else {
+    linenr_T start_lnum = old_line_count + 1;
+    new_byte_count = get_region_bytecount(buf, start_lnum, new_line_count, 0, new_endcol);
+    extmark_splice(buf, old_line_count - 1, old_endcol,
+                   0, 0, 0,
+                   new_line_count - old_line_count, new_endcol,
+                   new_byte_count, kExtmarkNoUndo);
```
justinmk: style nit: all the linebreaks here aren't really helping much. and line-length limit is 100, not 80.

```c
  linenr_T delta = new_line_count - old_line_count;
  if (old_last == NULL) {
    new_byte_count = get_region_bytecount(buf, 1, new_line_count, 0, new_endcol);
    extmark_splice(buf, 0, 0, old_line_count - 1, 0, old_byte_count, new_line_count - 1, new_endcol,
                   new_byte_count, kExtmarkNoUndo);
    changed_lines(buf, 1, 0, old_line_count > 0 ? old_line_count + 1 : 1, delta, true);
  } else {
    linenr_T start_lnum = old_line_count + 1;
    new_byte_count = get_region_bytecount(buf, start_lnum, new_line_count, 0, new_endcol);
    extmark_splice(buf, old_line_count - 1, old_endcol, 0, 0, 0, delta, new_endcol, new_byte_count,
                   kExtmarkNoUndo);
    changed_lines(buf, start_lnum, 0, new_line_count + 1, delta + 1, true);
  }
```

src/nvim/quickfix.c
glepnir: initial value of `ml_line_count` is 1 in `ml_open`, It is only set to 0 when buf_freeall is called.

test/functional/lua/buffer_updates_spec.lua
glepnir: extmark_splice called `buf_updates_send_splice` which trigger `on_bytes`

justinmk, line 424:
```suggestion
  it('in quickfix buffer', function()
```

test/functional/lua/buffer_updates_spec.lua
```diff
@@ -420,6 +420,35 @@ describe('lua buffer event callbacks: on_lines', function()
...
+    command('copen')
+    exec_lua(function()
+      vim.api.nvim_buf_attach(vim.api.nvim_get_current_buf(), false, {
+        on_lines = function(...)
+          vim.g.qf_on_lines = { ... }
+        end,
+        on_bytes = function(...)
+          vim.g.qf_on_bytes = { ... }
```
justinmk, line 432: The other tests use `check_events`. Using a different approach makes the tests harder to understand. OTOH, this test is testing both "lines" and "bytes".

PR #31110, opened 2024-11-08
feat(lsp): option to highlight hover range

ribru17:
**Problem:** Despite the LSP providing the option for language servers to specify a range with a hover response (for highlighting), Neovim does not give the option to highlight this range.

**Solution:** Add an option to `buf.hover()` which causes this range to be highlighted.

### Before
image

### After
image

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -338,13 +338,16 @@ Highlight groups that are meant to be used by |vim.lsp.buf.document_highlight()|
 You can see more about the differences in types here:
 https://microsoft.github.io/language-server-protocol/specification#textDocument_documentHighlight
 
-                                                           *hl-LspReferenceText*
+                                                         *hl-LspReferenceText*
 LspReferenceText          used for highlighting "text" references
-                                                           *hl-LspReferenceRead*
+                                                         *hl-LspReferenceRead*
 LspReferenceRead          used for highlighting "read" references
-                                                          *hl-LspReferenceWrite*
+                                                        *hl-LspReferenceWrite*
 LspReferenceWrite         used for highlighting "write" references
-                                                          *hl-LspInlayHint*
+                                                       *hl-LspReferenceTarget*
+LspReferenceTarget        used for highlighting reference targets (e.g. in a
+                          hover range)
+                                                             *hl-LspInlayHint*
 LspInlayHint              used for highlighting inlay hints
 
 
@@ -1334,6 +1337,14 @@ hover({config})                                          *vim.lsp.buf.hover()*
     the hover window and focus it). In the floating window, all commands and
     mappings are available as usual, except that "q" dismisses the window. You
     can scroll the contents the same as you would any other buffer.
+
+    Note: to disable hover highlights, add the following to your config: >lua
+        vim.api.nvim_create_autocmd('ColorScheme', {
+          callback = function()
+            vim.api.nvim_set_hl(0, 'LspReferenceTarget', {})
+          end,
+        })
+<
 
     Parameters: ~
       • {config}  (`vim.lsp.buf.hover.Opts?`) See |vim.lsp.buf.hover.Opts|.
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -20,6 +20,8 @@ local function client_positional_params(params)
   end
 end
 
+local hover_ns = api.nvim_create_namespace('vim_lsp_hover_range')
+
 --- @class vim.lsp.buf.hover.Opts : vim.lsp.util.open_floating_preview.Opts
 --- @field silent? boolean
 
@@ -30,13 +32,24 @@ end
 --- In the floating window, all commands and mappings are available as usual,
 --- except that "q" dismisses the window.
 --- You can scroll the contents the same as you would any other buffer.
+---
+--- Note: to disable hover highlights, add the following to your config:
+---
+--- ```lua
+--- vim.api.nvim_create_autocmd('ColorScheme', {
+---   callback = function()
+---     vim.api.nvim_set_hl(0, 'LspReferenceTarget', {})
+---   end,
+--- })
+--- ```
 --- @param config? vim.lsp.buf.hover.Opts
 function M.hover(config)
   config = config or {}
   config.focus_id = ms.textDocument_hover
 
   lsp.buf_request_all(0, ms.textDocument_hover, client_positional_params(), function(results, ctx)
-    if api.nvim_get_current_buf() ~= ctx.bufnr then
+    local bufnr = assert(ctx.bufnr)
+    if api.nvim_get_current_buf() ~= bufnr then
       -- Ignore result since buffer changed. This happens for slow language servers.
       return
     end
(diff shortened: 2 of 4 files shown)


runtime/lua/vim/lsp/buf.lua
```diff
@@ -30,13 +33,17 @@ end
 --- In the floating window, all commands and mappings are available as usual,
 --- except that "q" dismisses the window.
 --- You can scroll the contents the same as you would any other buffer.
+---
+--- Note: when using `apply_highlight`, ensure that the |hl-LspHoverRange| highlight is defined or
```
MariaSolOs, line 37: What if instead of requiring this we just link it to `Visual` by default?

MariaSolOs (approved):

runtime/lua/vim/lsp/buf.lua
ribru17: Now at least it is linked to `LspReferenceText` which links to `Visual`; would you prefer I link it directly to `Visual`?

MariaSolOs: Oops my bad I missed that. It's fine to link it to `LspReferenceText`, but I would then change this note to "If using `apply_highlight`, the hovered range will be highlighted with |hl-LspHoverRange|". That's enough to let users know of how they can customize it.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -87,6 +95,33 @@ function M.hover(config)
       else
         vim.list_extend(contents, util.convert_input_to_markdown_lines(result.contents))
       end
+      local range = result.range
+      if range and config.apply_highlight then
```
MariaSolOs, line 99: Hmmm given that we now support multiple responses for a single hover request (cc @lewis6991 @mfussenegger), this could highlight several ranges.

What should we do if a client doesn't include a highlight range but another one does? What if the ranges are different?

ribru17: Yeah I was wondering about that... right now all provided ranges will be highlighted, so they will appear combined if they overlap at all. This seemed like the most obvious solution but maybe there are cases I didn't consider where this would be unintuitive?

MariaSolOs: If we end up using cycling for handling multiple hover responses (see https://github.com/neovim/neovim/pull/31029#issuecomment-2453280867 for more info on the idea) then we could potentially match the highlighted range with the current hover. I don't think there's an issue with the current approach of "over-highlighting" but extmarks aren't my area of expertise.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -20,8 +20,11 @@ local function client_positional_params(params)
   end
 end
 
+local hover_ns = api.nvim_create_namespace('vim_lsp_hover_range')
+
 --- @class vim.lsp.buf.hover.Opts : vim.lsp.util.open_floating_preview.Opts
 --- @field silent? boolean
+--- @field apply_highlight? boolean Whether to highlight the range associated with the hover result. Defaults to `false`.
```
mfussenegger, line 27: I'm not sure this warrants a new option. Could be done by default. If users don't want the highlight they could configure the `LspHoverRange` hl group accordingly.

runtime/lua/vim/lsp/buf.lua
```diff
@@ -100,7 +135,17 @@ function M.hover(config)
       return
     end
 
-    lsp.util.open_floating_preview(contents, format, config)
+    local _, winid = lsp.util.open_floating_preview(contents, format, config)
+
+    if config.apply_highlight then
+      api.nvim_create_autocmd('WinClosed', {
```
mfussenegger, line 141: Is this necessary? The buffer has `bufhidden = 'wipe'` set. Once the window closes the buffer disappears, and I'd think with it the highlights.

runtime/lua/vim/lsp/buf.lua
ribru17: This makes sense. The only downside that I can think of is that this override would have to go in a `ColorScheme` autocmd, but I don't think that's too unreasonable. Can mention this in the doc

runtime/lua/vim/lsp/buf.lua
ribru17: I might be misunderstanding, but wouldn't this only apply to highlights within the hover buffer itself? This autocmd is for the regular buffer the user is on, i.e. whenever the hover window closes, remove the hover highlight from the other, regular code buffer

runtime/lua/vim/lsp/buf.lua
```diff
@@ -87,6 +101,33 @@ function M.hover(config)
...
+
+        vim.hl.range(
+          bufnr,
+          hover_ns,
+          'LspHoverRange',
+          { start_line, start_idx },
+          { end_line, end_idx },
+          { priority = vim.hl.priorities.user }
+        )
+      end
```
mfussenegger, line 130:
```suggestion
      if range then
        local start = range.start
        local end_ = range["end"]
        local start_idx = util._get_line_byte_from_position(
          bufnr,
          start,
          client.offset_encoding
        )
        local end_idx = util._get_line_byte_from_position(
          bufnr,
          end_,
          client.offset_encoding
        )

        vim.hl.range(
          bufnr,
          hover_ns,
          'LspHoverRange',
          { start.line, start_idx },
          { end_.line, end_idx },
          { priority = vim.hl.priorities.user }
        )
      end
```

No need to unpack the start/end position only to create a new table in the same format.

mfussenegger (approved):

runtime/doc/lsp.txt
```diff
@@ -344,6 +344,8 @@ LspReferenceText          used for highlighting "text" references
 LspReferenceRead          used for highlighting "read" references
                                                           *hl-LspReferenceWrite*
 LspReferenceWrite         used for highlighting "write" references
+                                                              *hl-LspHoverRange*
+LspHoverRange             used for highlighting hover ranges
```
justinmk, line 348: alignment of the tag looks off

ribru17: Thanks- should they all end at col 80?

src/nvim/highlight_group.c
```diff
@@ -215,6 +215,7 @@ static const char *highlight_init_both[] = {
   "default link LspReferenceRead            LspReferenceText",
   "default link LspReferenceText            Visual",
   "default link LspReferenceWrite           LspReferenceText",
+  "default link LspHoverRange               LspReferenceText",
```
justinmk, line 218: Would `LspHoverTarget` be slightly more generic / meaningful for potentially other cases? The "range" part seems over-specific.

Maybe even something more generic like `LspReferenceTarget`

justinmk (approved):
This is a great idea. Needs a rebase

runtime/doc/lsp.txt
justinmk: oh I was thinking the tags should left-align on the same column. doesn't matter too much guess :P

src/nvim/highlight_group.c
ribru17: Nice, yeah I'll do that

src/nvim/highlight_group.c
```diff
@@ -215,6 +215,7 @@ static const char *highlight_init_both[] = {
   "default link LspReferenceRead            LspReferenceText",
   "default link LspReferenceText            Visual",
   "default link LspReferenceWrite           LspReferenceText",
+  "default link LspReferenceTarget          LspReferenceText",
```
justinmk, line 218: Reviewing the LSP spec and search results, I'm wondering why we need a new highlight. It  seems like `DocumentHighlightKind.Text` perfectly fits the use-case here. @MariaSolOs @mfussenegger any preference?

justinmk (approved):

src/nvim/highlight_group.c
ribru17: The only downside to this I can think of is that it would not be possible to disable highlights specifically for hover actions, since it is tied to hl group, but I don't know if anyone would even want to do this

MariaSolOs: I prefer having the separate highlight group. I don't set `LspReferenceText` in my config because I find them to be too noisy, but I would like to have the highlight for hover range.

PR #31127, opened 2024-11-08
fix(treesitter): show proper node name error messages

ribru17:
**Problem:** Currently node names with non-alphanumeric, non underscore/hyphen characters (only possible with anonymous nodes) are not given a proper error message. See tree-sitter/tree-sitter#3892 for more details.

**Solution:** Apply a different scanning logic to anonymous nodes to correctly identify the entire node name (i.e., up until the final double quote)


ribru17:
cc @clason

PR #31138, opened 2024-11-09
docs: misc

dundargoc:


dundargoc:
Removed the "Last change:" lines from the neovim-specific runtime files as they seem to be perpetually outdated/incorrect.

PR #31158, opened 2024-11-10
feat(defaults): add new `[` and `]` pending operators

### LINKED ISSUE #30729: directional text motions (like vim-ninja-feet)
### Problem

Vim and Neovim have concise mappings for acting on text objects. Example: `vab` selects-and-includes a pair of `()`s, `di[` deletes the contents of `[]`s. 

However there aren't many text motions available that incorporate the user's current cursor position. There are horizontal options like `dfB` to mean "delete from the cursor to the first letter B" but a user cannot as easily express "delete all text starting from this cursor position up to the end of this paragraph" as a mapping in a generic way.

### Expected behavior

As it happens, there's a plugin that does this, https://github.com/tommcdo/vim-ninja-feet

In short: It adds "operator-pending" mappings that consider the cursor's current position.

A typical text motion + object pair like `vap`, now has 2 more options `v[ap` and `v]ap`. `[` means "from the start of the text object to the current cursor" and `]` means "from the cursor to the end of the text object"

With built-in text objects or other plugins, examples like the following become easily possible:

`v[ic` means "select all lines that are comments from the current cursor position to the last comment"
`d]ip` means "delete from the current cursor line to the end of the paragraph"
`gc[ii` means "comment all of the lines from the cursor down to the last line of same indent"

Neovim has recently been added more meaningful text operations such as [built-in commenting](https://github.com/neovim/neovim/pull/28176), [unimpaired-style mappings](https://github.com/neovim/neovim/commit/bb7604eddafb31cd38261a220243762ee013273a), and there's a GitHub issue for [adding `gs` to sort lines](https://github.com/neovim/neovim/issues/19354#issue-1303638463). Please consider adding directional text motions to improve manipulating text. What do you think?

ColinKennedy:
Closes: https://github.com/neovim/neovim/issues/30729

This PR introduces two new pending operators: `[` and `]`.

In Neovim/Vim, `dip` deletes a paragraph under the current cursor. This PR now supports two new mappings, `d[ip` and `d]ip`.

- `d[ip` means "delete from the start of the paragraph to the current cursor position"
- `d]ip` means "delete from the current cursor position to the bottom of the paragraph"

Most `{operator}{object}` pairs, for example `dap`, `yip`, `cas`, `daw`, etc etc now have a `{operator}[{object}` and `{operator}]{object}` equivalent.

And of course, this works for custom text objects / plugins. These new `[` and `]` pending operators greatly improve the flexibility and expressiveness of Vim text objects!

## How It Works
When a user types `d]ap`, we ...

1. Save the user's current cursor, `v:operator`, `operatorfunc`, and other details as needed
2. Replace `operatorfunc` with our own.
3. Immediately execute the replaced `operatorfunc`.

And once executing...

1. Restore the previous cursor position, previous `v:operator`, and `operatorfunc`
2. If `[` or `]` was pressed, nudge the Vim `'[` or `']` marks if needed
3. Execute the operation

For anyone familiar, this PR is a lua version of [vim-ninja-feet](https://github.com/tommcdo/vim-ninja-feet). Props goes to Tom for the awesome idea!

## Considerations For Reviewers
Here's some extra things that are not in the PR but could be useful / importing to consider.

### Question 1 - Should the operator include / exclude the current line (for line-wise text objects)?

<details>
<summary>Click to expand</summary>

When operating over whole-lines, `[` can be thought of "searching up" for lines and is `]` "searching down". When that happens, should the current cursor's line be included in the search?

For example take this paragraph of text

```
aaa
bbbb  <-- NOTE: The user's cursor is here
cc
```

If we `d]ip` should the `bbbb` line also be deleted? What about if we use `d[ip`? Should the `bbbb` line be deleted or ignored also?

Currently this PR does "always delete the line, whether we call `[` or `]`".

I was thinking maybe this could be solved in one of two ways.

#### Idea 1 - 4 mappings
Instead of providing `[` and `]` we provide `[/[[` and `]/]]`. So `d[ap` could include the cursor line and `d[[ap` could exclude the cursor line. A user could decide while typing if they care to include the line or not.

(Note: I think Idea 2 is better, see below)

#### Idea 2 - Configuration
If we anticipate that a user would always want to include the cursor line / exclude it then there isn't a need to provide a separate mapping for it. Instead we could provide configuration values such as.

```
g:inclusive_pending_cursor = nil / "both" (default, this PR's behavior)
g:inclusive_pending_cursor = "up" (only include the cursor's line if using `[`)
g:inclusive_pending_cursor = "down" (only include the cursor's line if using `]`)
g:inclusive_pending_cursor = "none" (never include the cursor's line)
```

And if the user actually did want the "4 mappings" solution they could define their own omap that temporarily sets `g:inclusive_pending_cursor` before running the regular omap. Which I think makes this idea more flexible than the previous idea.
</details>

### Question 2 - Do we want to consider column info even if the text object is a line?

<details>
<summary>Click to expand</summary>

Text objects like `ip` (e.g. `dip`) operate on whole-lines. Currently, the PR respects this. When a user calls `d[ip`, it respects only the cursor's row and ignores the cursor's column position.

In visual terms imagine you have text like this

```
aaa
bb|your cursor here|tt
cc
```

Typing `d[ip` would result in this:

```
cc
```

If we considered the cursor's column position, the result instead would be

```
tt
cc
```

If a user did want to delete the whole row they could still do it by calling `dd` after calling `d[ip`. But I admit it is a bit wordy. What do you all think?
</details>

### DIFF
--- a/runtime/lua/vim/_cursor.lua
+++ b/runtime/lua/vim/_cursor.lua
@@ -1,0 +1,111 @@
+--- All function(s) that can be called externally by other Lua modules.
+---
+--- If a function's signature here changes in some incompatible way, this
+--- package must get a new **major** version.
+---
+---@module 'vim._cursor'
+---
+
+local M = {}
+
+local _Direction = { down = 'down', up = 'up' }
+
+---@class vim._cursor.Options
+---    Describe which way to adjust the cursor and how.
+---@field count number
+---    The number of times to call the command. Usually is just `1`.
+---@field direction "down" | "up"
+---    Which way to crop the text object. "up" means "operate from the cursor's
+---    position to the up/left-most position" and "down" means "operate from
+---    the cursor's position (including the cursor's current line, if
+---    applicable) to the down/right-most position".
+
+local _CURSOR
+---@type string
+local _DIRECTION
+---@type string
+local _OPERATOR
+---@type string
+local _OPERATOR_FUNCTION
+
+--- Check if the operatorfunc that is running will run on a whole-line.
+---
+--- See Also:
+---     :help g@
+---
+---@param mode "block" | "char" | "line"
+---@return boolean # If `mode` is meant to run on the full-line (ignores column data).
+---
+local function _is_line_mode(mode)
+  return mode == 'line'
+end
+
+--- Execute the original operatorfunc but crop it based on the cursor position.
+---
+--- Important:
+---     You must call `prepare()` once before this function can be called.
+---
+---@param mode "block" | "char" | "line"
+---    The caller context. See `:help :map-operator` for details.
+---
+function M.operatorfunc(mode)
+  vim.fn.setpos('.', _CURSOR)
+  vim.o.operatorfunc = _OPERATOR_FUNCTION
+  ---@type string
+  local mode_character
+  local is_line = _is_line_mode(mode)
+
+  if is_line then
+    mode_character = "'" -- Includes only line information
+  else
+    mode_character = '`' -- Includes column information
+  end
+
+  ---@type string
+  local direction
+  local inclusive_toggle = ''
+
+  if _DIRECTION == _Direction.up then
+    direction = '['
+
+    if not is_line then
+      local buffer, row, column, offset = unpack(vim.fn.getpos("'" .. direction))
+      ---@cast row number
+
+      if column > #vim.fn.getline(row) then
+        row = row + 1
+        column = 0
+      end
+
+      vim.fn.setpos("'" .. direction, { buffer, row, column, offset })
+    end
+  else
+    direction = ']'
+
+    local buffer, row, column, offset = unpack(vim.fn.getpos("'" .. direction))
+    ---@cast row number
+
+    if not is_line and column == #vim.fn.getline(row) then
+      -- NOTE: Move the mark past the current cursor column
+      inclusive_toggle = 'v'
+    else
+      vim.fn.setpos("'" .. direction, { buffer, row, column + 1, offset })
+    end
+  end
+
+  vim.fn.feedkeys(_OPERATOR .. inclusive_toggle .. mode_character .. direction)
+end
+
+--- Remember anything that we will need to recall once we execute `operatorfunc`.
+---
+---@param options vim._cursor.Options
+---    Mapping data that must be remembered before `operatorfunc` is called.
+---
+function M.prepare(options)
+  _DIRECTION = options.direction
+  _OPERATOR = vim.v.operator
+  _CURSOR = vim.fn.getpos('.')
+  _OPERATOR_FUNCTION = vim.o.operatorfunc
+end
+
+return M
--- a/test/functional/lua/cursor_pending_operator_spec.lua
+++ b/test/functional/lua/cursor_pending_operator_spec.lua
@@ -1,0 +1,1428 @@
+--- Make sure the basic functionality of cursor-text-object works.
+
+local t = require('test.testutil')
+local n = require('test.functional.testnvim')()
+
+local api = n.api
+local clear = n.clear
+local eq = t.eq
+local exec_lua = n.exec_lua
+local fn = n.fn
+
+--- Get the lines of `buffer`.
+---
+---@param buffer number # A 1-or-more identifier for the Vim buffer.
+---@return string # The text in `buffer`.
+---
+local function get_lines(buffer)
+  return fn.join(api.nvim_buf_get_lines(buffer, 0, -1, false), '\n')
+end
+
+--- Run `keys` from NORMAL mode.
+---
+---@param keys string Some command to run. e.g. `d]ap`.
+---
+local function call_command(keys)
+  exec_lua(function()
+    vim.cmd('normal ' .. keys)
+  end)
+end
+
+--- Create a new Vim buffer with `text` contents.
+---
+---@param text string All of the text to add into the buffer.
+---@param file_type string? Apply a type to the newly created buffer.
+---@return number # A 1-or-more identifier for the Vim buffer.
+---@return number # A 1-or-more identifier for the Vim window.
+---
+local function make_buffer(text, file_type)
+  local buffer = api.nvim_create_buf(false, false)
+  api.nvim_set_current_buf(buffer)
+
+  if file_type then
+    api.nvim_set_option_value('filetype', file_type, {})
+  end
+
+  api.nvim_buf_set_lines(buffer, 0, -1, false, fn.split(text, '\n'))
+
+  return buffer, api.nvim_get_current_win()
+end
+
+--- Make sure `input` becomes `expected` when `keys` are called.
+---
+---@param cursor {[1]: number, [2]: number} The row & column position. (row=1-or-more, column=0-or-more).
+---@param keys string Some command to run. e.g. `d]ap`.
+---@param input string The buffer's original text.
+---@param expected string The text that we expect to get after calling `keys`.
+---
+local function run_simple_test(cursor, keys, input, expected)
+  local buffer, window = make_buffer(input)
+  api.nvim_win_set_cursor(window, cursor)
+
+  call_command(keys)
+
+  eq(expected, get_lines(buffer))
+end
+
+--- Initialize 'commentstring' so `:help gc` related tests work as expected.
+---
+---@param text string The template for creating comments. e.g. `"# %s"`.
+---@param buffer number A 0-or-more Vim buffer ID.
+---
+local function set_commentstring(text, buffer)
+  api.nvim_set_option_value('commentstring', text, { buf = buffer })
+end
+
+before_each(function()
+  clear({ args_rm = { '--cmd' }, args = { '--clean' } })
+end)
+
+describe('basic', function()
+  it('works with inner count', function()
+    run_simple_test(
+      { 2, 0 },
+      'd]2ap',
+      [[
+      some text
+          more text <-- NOTE: The cursor will be set here
+          even more lines!
+      still part of the paragraph
+
+      another paragraph
+      with text in it
+
+      last paragraph
+      ]],
+      [[
+      some text
+      last paragraph
+      ]]
+    )
+  end)
+
+  it('works with outer count', function()
+    run_simple_test(
+      { 2, 0 },
+      '2d]ap',
+      [[
+      some text
+          more text <-- NOTE: The cursor will be set here
+          even more lines!
+      still part of the paragraph
+
+      another paragraph
+      with text in it
+
+      last paragraph
+      ]],
+      [[
+      some text
+      last paragraph
+      ]]
+    )
+  end)
+end)
+
+describe(':help c', function()
+  describe('down', function()
+    it('works cap', function()
+      run_simple_test(
+        { 2, 0 },
+        'c]ap',
+        [[
+        some text
+            more text <-- NOTE: The cursor will be set here
+            even more lines!
+        still part of the paragraph
+
+        another paragraph
+        with text in it
+        ]],
+        [[
+        some text
+
+        another paragraph
+        with text in it
+        ]]
+      )
+    end)
+
+    it('works ca}', function()
+      run_simple_test(
+        { 3, 0 },
+        'c]a}',
+        [[
+        {
+            some text
+                more text  <-- NOTE: The cursor will be set here
+                even more lines!
+            still part of the paragraph
+
+            more lines
+        }
+
+        {
+            another paragraph
+            with text in it
+        }
+        ]],
+        [[
+        {
+            some text
+
+
+        {
+            another paragraph
+            with text in it
+        }
+        ]]
+      )
+    end)
+  end)
+
+  describe('up', function()
+    it('works cap', function()
+      run_simple_test(
+        { 7, 0 },
+        'c[ap',
+        [[
+        some text
+            more text
+            even more lines!
+        still part of the paragraph
+
+        first line, second paragraph
+        another paragraph  <-- NOTE: The cursor will be set here
+        with text in it
+        ]],
+        [[
+        some text
+            more text
+            even more lines!
+        still part of the paragraph
+
+
+        with text in it
+        ]]
+      )
+    end)
+
+    it('works ca}', function()
+      run_simple_test(
+        { 3, 0 },
+        'c[a}',
+        [[
+        {
+            some text
+                more text  <-- NOTE: The cursor will be set here
+                even more lines!
+            still part of the paragraph
+
+            more lines
+        }
+
+        {
+            another paragraph
+            with text in it
+        }
+        ]],
+        [[
+
+                more text  <-- NOTE: The cursor will be set here
+                even more lines!
+            still part of the paragraph
+
+            more lines
+        }
+
+        {
+            another paragraph
+            with text in it
+        }
+        ]]
+      )
+    end)
+  end)
+end)
+
+describe(':help d', function()
+  describe('down', function()
+    it('works with da)', function()
+      run_simple_test(
+        { 3, 0 },
+        'd]a)',
+        [[
+        (
+            some text
+                more text  <-- NOTE: The cursor will be set here
+                even more lines!
+            still part of the paragraph
+
+            more lines
+        )
+
+        (
+            another paragraph
+            with text in it
+        )
+        ]],
+        [[
+        (
+            some text
+
+
+        (
+            another paragraph
+            with text in it
+        )
+        ]]
+      )
+    end)
+
+    it('works with da]', function()
+      run_simple_test(
+        { 3, 0 },
+        'd]a]',
+        [[
+        [
+            some text
+                more text  <-- NOTE: The cursor will be set here
+                even more lines!
+            still part of the paragraph
+
+            more lines
+        ]
+
+        [
+            another paragraph
+            with text in it
+        ]
+        ]],
+        [[
+        [
+            some text
+
+
+        [
+            another paragraph
+            with text in it
+        ]
+        ]]
+      )
+    end)
+
+    it('works with da}', function()
+      run_simple_test(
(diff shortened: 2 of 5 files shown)


clason:
Hmm, this is useful but I feel that this is a bit heavy for including by default (not every plugin should be upstreamed; there's a relatively high bar). Especially the new module looks to be high-cost -- _unless_ it allows refactoring to remove similar code in (multiple) other places. Can you comment on that?

ColinKennedy:
> Unless it allows refactoring to remove similar code in (multiple) other places. Can you comment on that?

As far as I'm aware the only way to achieve what this PR does is to set a mark, use `v` to begin a range, provide a text object, then jump to the previous mark and then type the text operator. e.g. `mbvap'mgc`. Versus `gc]ap`. There might be a smaller way without this PR but the expressiveness and flexibility shouldn't be discounted. Vim's strengths are its concise expressiveness and this feature would x3 every mapping (Neo)vim has available + custom text operators / objects.

> Especially the new module looks to be high-cost

I'd like to talk about that for a bit. if by new module you mean the `_cursor.lua` file, that's 49 lines of source code. Most of the lines are handling a couple edge cases. I'm assuming by cost you don't actually mean the file's contents but something else that I'm not seeing. Would you please elaborate on that?

clason:
I'm talking about design, maintenance, and documentation cost, which does not (just) scale linearly with loc; there's an up-front cost, too. (For example, this will "squat" the `vim.cursor` namespace -- will this conflict with other plans?)

ColinKennedy:
> (For example, this will "squat" the vim.cursor namespace -- will this conflict with other plans?)

I can rename the file. Originally I called it `_directional_pending_operator.lua` but I thought that that was a mouthful :)

The code is small enough that it could also just be inlined. I was following another PR as a guide while deciding where to place things here.

> I'm talking about design, maintenance, and documentation cost, which does not (just) scale linearly with loc

That's fair I just wanted to make sure we're talking about the same thing. I figured you didn't literally mean the code as it is in the PR.

When I made the original issue it was tagged with "needs-owner" which I saw is defined as "We don't plan to work on this, but will accept high quality contributions from someone who will own the feature and follow up on bug reports.". FWIW I'm offering to own this feature and its associated costs if that is allowed. That said, I've been using this equivalent (current state) code for 8 years without issue. It is simply heavenly!

echasnovski:
Here are a bit more points to consider here:
- My main concern about this is that it will not play nicely with various `[x`/`]x` plugin mappings which implement custom movements. For example, `[i`/`]i` is a fairly obvious choice for plugins which work with indent levels; `[a`/`]a` is the same for navigating arguments, etc. Those are the obvious choices, but there are more `[`/`]` motions implemented by plugins which might start with the same letter as some textobject.
- Having these `[` and `]` be available only in Operator-pending mode is not really consistent with how textobjects work, which are (usually) available in both Visual and Operator-pending mode. Similar, `[`/`]` (if accepted at all) should be available in both Visual and Operator-pending mode.

All in all, I personally think that this is indeed a bit heavy custom mappings to be included by default. By that I mostly mean that it maps a heavily used pair of single characters (in what should be two heavily used modes) which is bound to introduce user troubles in the long run. The feature is interesting in on itself, but adding it should be a conscious choice by the user.

echasnovski:
> To piggyback on what echasnovski says here, neovim itself [recently added a bunch of default mappings inspired by vim-unimpaired—e.g., [here](https://github.com/neovim/neovim/commit/bb7604eddafb31cd38261a220243762ee013273a) and [here](https://github.com/neovim/neovim/pull/30984), and all of them use `[` or `]` as a mapping prefix.

I also thought about it, but they don't really apply here as they are created in Normal mode (no Visual and no Operator-pending mode). And it is reasonable, as they might change buffer.

Same goes for `[d`/`]d`/`[D`/`]D`. Although I'd argue that it is reasonable to also have them in Visual and Operator-pending mode as they don't change buffer.

ColinKennedy:
> My main concern about this is that it will not play nicely with various [x/]x plugin mappings which implement custom movements.

So good news on this front, there is no conflict with this PR and those mappings. As in these pending operators would not prevent users from defining their own with `[` / `]` for their own purpose. If by conflict you mean not a "mapping conflict" but something more semantic - `d]ap` "delete from the current position to the next paragraph" is still consistent with `]a` "move from the current position to the next arglist entry". If you're still unsure then please suggest an alternative mapping.

> (usually) available in both Visual and Operator-pending mode

My bad I forgot about that, adding support for `v]ap` and others would be a good addition to this PR. I can add that.

> By that I mostly mean that it maps a heavily used pair of single characters

Neovim recently [added 28 directional (pairwise) mappings](https://github.com/neovim/neovim/blob/079e5f4f9b67a5aa2c1b481ce78711bf8c76caea/runtime/lua/vim/_defaults.lua) and this PR plays well with all of them. The pending operators listed here also do not prevent users from defining their own. If you'd like say there's a potential conflict then can I see a practical example of what you're describing? I see no conflict in any example given so far. And moreover they're actually fairly in-line with a user's expectations.

telemachus:
My concern was about *cognitive* overload more than programmatic clashes. That said, there are limited keys on our keyboards, so maybe (a lot of) overloading is inevitable.

ColinKennedy:
The cognitive overload of 2 new mappings vs the 28 that were just merged?

ColinKennedy:
If I can make a suggestion - I think value-proposition conversations like are better suited for a chatroom / github issue rather than the PR. I'd much rather this conversation thread be about the contents of the PR rather than "is this valuable / useful". Especially since we've been going back and forth for a while. Would you be up for chatting about this on the Neovim Matrix? If not, the GitHub issue? I'd like to keep this PR conversation thread about the technical contents of the PR if possible.

clason:
No, this is exactly where these discussions should happen (and they _are_ part of PR review): in the open and, more importantly, findable in the future.

ColinKennedy:
I wasn't suggesting that as a way of hiding the chat but keeping conversation points separate. Since GitHub conversations are one flat list rather than categorized / threaded as they often should be.

Anyway I'm happy to continue chatting here if that's what everyone wants.

telemachus:
> The cognitive overload of 2 new mappings vs the 28 that were just merged?

Point taken. I suppose I meant more the cognitive overload of "Wait, what do `]` and `[` do in this particular {neo,}vim incantation?" The more each key is used, the more you have to think about how the key affects *this* particular use. 

That said, after reading this thread again from top to bottom, you've convinced me that the uses are separate enough and the added value is worth it.

echasnovski:
> > My main concern about this is that it will not play nicely with various [x/]x plugin mappings which implement custom movements.
> 
> So good news on this front, there is no conflict with this PR and those mappings. As in these pending operators would not prevent users from defining their own with `[` / `]` for their own purpose. If by conflict you mean not a "mapping conflict" but something more semantic - `d]ap` "delete from the current position to the next paragraph" is still consistent with `]a` "move from the current position to the next arglist entry".

Ok, let's not call it a conflict, but definitely an issue. If this PR gets merged, both `[` and `]` are expected to *always* work as "move from cursor position to start/end of textobject". This introduces following situations:
- User needs to review/know *all* directional `[` and `]` in order to not have conflicts with textobjects *they* use (so that they can still use standalone `[` and `]` with *any* textobject they please). As textobjects can start with any letter (although most common are `a`, `i`, `g`). This is an issue of comparing many (textobjects) with many (direction mappings).
    Instead, if not merging this PR, the decision boils down to "whether I want this type of standalone `[`/`]` mappings which bring *these* possible mappings issues".
- Same situation as the first one, but for Neovim built-ins. Having `[` and `]` built-in mappings will basically mean that in order to have them work 100% there can not be any new built-in directional `[x`/`]x` mappings with `x` matching how some built-in textobject starts. So there can not be `[a` / `[i` / `[g` built-in mappings (and maybe more if I missed some built-in textobjects or there will be in the future).
    I'd much rather leave the door open for any new directional mappings without any this type of deep dive concerns.

> If you're still unsure then please suggest an alternative mapping.

The `g[` and `g]` might be good alternatives, but they are already present with useful functionality (at least for someone, I'd presume).
My main solution here is to not make these default mappings and leave a choice of adding them to the user as a standalone plugin.

> If you'd like say there's a potential conflict then okay, can I see a practical example of what you're describing?

Apart from situations described above, there is no one exact practical example. Mostly an additional instinct that introducing this type of mappings will cause *unexpected* problems and future limitations.

ColinKennedy:
> Point taken. I suppose I meant more the cognitive overload of "Wait, what do ] and [ do in this particular {neo,}vim incantation?" The more each key is used, the more you have to think about how the key affects this particular use.

Alright now I understand what you meant. That said I think this is all consistent though. `]b` means "from the current position (in the buffers list), go to the right buffer" and `d]as` means "from the current position, delete to the right of the sentence". A user is learning 1 concept, `]` = "right", and applying it to multiple places. IMO that reinforces a person's understanding of both mappings rather than being a cognitive overload. But again, just my two cents.

> That said, after reading this thread again from top to bottom, you've convinced me that the uses are separate enough and the added value is worth it.

I'm really glad to hear you say! If you have other concerns I'd be happy to help.

> Ok, let's not call it a conflict, but definitely an issue. If this PR gets merged, both [ and ] are expected to always work as "move from cursor position to start/end of textobject". This introduces following situations: ... Apart from situations described above, there is no one exact practical example. Mostly an additional instinct that introducing this type of mappings will cause unexpected problems and future limitations.

@echasnovski I understand you've got an instinct, gut feeling about this and I want to alleviate those potential concerns. I think the disconnect here is that I haven't done a good enough job at measuring, with actual code, the possible impact this PR could have. Partly the reason why I didn't is because I've been using this for years (and I use a > 30 text object plugins) and none of the conflicts folks have raised here has actually come up in practice. So I assumed that whatever cost / risk is minimal.

So how about this suggestion? I'm out of time for today but I'd like to reply back with various, minimal code scenarios (e.g. what happens if someone does map `[x` / `]x`) and let's talk about those findings. Let's get an accurate, empirical idea about cost.

echasnovski:
> So how about this suggestion? I'm out of time for today but I'd like to reply back with various, minimal code scenarios (e.g. what happens if someone does map `[x` / `]x`) and let's talk about those findings. Let's get an accurate, empirical idea about cost.

The empirical idea of this being built-in is already too high in my opinion at least for this reason:

> * Same situation as the first one, but for Neovim built-ins. Having `[` and `]` built-in mappings will basically mean that in order to have them work 100% there can not be any new built-in directional `[x`/`]x` mappings with `x` matching how some built-in textobject starts. So there can not be `[a` / `[i` / `[g` built-in mappings (and maybe more if I missed some built-in textobjects or there will be in the future).
>       I'd much rather leave the door open for any new directional mappings without any this type of deep dive concerns.

In other words it is not a very well isolated case of adding two mappings: it affects any future considerations about new built-in textobjects and new built-in direction mappings. This is true for all new built-in mappings but it is especially true for these fundamental one character `[` and `]` ones. These are fairly hard to predict which will pop up in the future.

The case for user side is even greater as they are not limited to built-in textobjects and built-in directional mappings.

------

Here are some examples form *current* ecosystem:
- ['mini.indentscope' defines `[i` and `]i`](https://github.com/echasnovski/mini.nvim/blob/7b4d5d48b6b5a75009d63f8f3e4ef4819b7e8139/doc/mini-indentscope.txt#L147-L148) movements by default. This means that this PR's `d]ip`/`d]i}`/etc. will (unexpectedly for the common user, I'd say) not work. 
- ['LazyVim/LazyVim'](https://github.com/LazyVim/LazyVim/blob/0819f9396e94a4188522b6739ade3d6024d2b8a7/lua/lazyvim/plugins/treesitter.lua#L79-L82) defines `[a` and `]a` as mappings. This means that this PR's `d]ap`/`d]a}` won't work.
- ['chrisgrieser/nvim-various-textobjs'](https://github.com/chrisgrieser/nvim-various-textobjs) has a bunch of one letter textobjects. Most of them are not really useful with this PR's `[`/`]` (as they are in itself "from cursor to ..." type of textobjects), but this is an example that there can be a lot of things

craigmac:
My main concern is that these mappings don't buy the user much from my point of view, I do understand they are omaps, so stepping on other bindings shouldn't be an issue.

Questions to OP, that I think would help in selling it: do you think this is as generally as useful as the unimpaired bindings are for about 80% of the userbase? I read through the code and I understand what it's doing, but I generally can't see myself ever using this sort of thing, I would say it's useful but niche. Counterpoint: there are many other text object plug-ins, my favourite being https://github.com/kana/vim-textobj-entire - why include this one and not include them? 

TLDR: great job, but not convinced it belongs in core at the moment

ColinKennedy:
> So there can not be [a / [i / [g built-in mappings (and maybe more if I missed some built-in textobjects or there will be in the future).

I think it's actually the opposite. A quick test showed `d[aap` calls `[a`, not `[`, when both mappings are present. Also your example of indentscope do not conflict as `[i` / `]i` are normal mode mappings. I would need more time to show concrete code examples.

> do you think this is as generally as useful as the unimpaired bindings are for about 80% of the userbase? ... I generally can't see myself ever using this sort of thing

The short answer is yes. Because these pending operators are augmenting existing mappings, they can be used in any case where a user is using text-objects. The problem that is PR solves is discoverability. People don't think to try something so they do not think to try it.

Ever needed to uncomment the top of a block of some commented text? `gc[ip`. Or sort parts of a dict/list because you have inline comments that you don't want to mess up? `gs]ii`. Copy just part of some text? `y]ap`. Or delete the lower part of a function? Well could be `d]ii` / `d]if` depending on your setup.

I'll speak personally for a moment. I never use the built-in `:help gs`. But I do sort stuff sometimes. In the past I had to sort by doing `vip` + `:sort`. Now it's just `gsip`. Before adding that text operator I rarely ever sorted, now...I do sort stuff all the time! I believe you when you say you're not sure how often you'd use this but I think that you might surprise yourself. People don't miss what they've never had. Maybe over the next couple days you'll comment back saying "ever since I saw this PR I'm seeing situations where it could applied everywhere".

> Counterpoint: there are many other text object plug-ins, my favourite being https://github.com/kana/vim-textobj-entire - why include this one and not include them?

Everyone has a favorite text object that they use all the time but not everyone will agree with which is best / most useful. And that's why not every text object will make it into Neovim's core.

But everyone **does** have a favorite text object or text operator.

The reason why you may want to merge this PR is because these pending operators work with your favorite. And your 2nd favorite. And your 3rd favorite. And 4th. All the down to your least favorite, niche text object. And remember, your 4th favorite might be someone else's favorite. These pending operators work for you, it works for them, it works for everyone. That's why you should consider merging this PR.

echasnovski:
> > So there can not be [a / [i / [g built-in mappings (and maybe more if I missed some built-in textobjects or there will be in the future).
> 
> I think it's actually the opposite. A quick test showed `d[aap` calls `[a`, not `[`, when both mappings are present.

Yes, and that is the point: with `[a` present, using `d[ap`/`d[a}` is now impossible. Which makes it a conflict with `[` from this PR's mappings.

> Also your example of indentscope do not conflict as `[i` / `]i` are normal mode mappings.

They are [Normal](https://github.com/echasnovski/mini.nvim/blob/7b4d5d48b6b5a75009d63f8f3e4ef4819b7e8139/lua/mini/indentscope.lua#L644-L645), [Visual](https://github.com/echasnovski/mini.nvim/blob/7b4d5d48b6b5a75009d63f8f3e4ef4819b7e8139/lua/mini/indentscope.lua#L647-L648), and [Operator-pending](https://github.com/echasnovski/mini.nvim/blob/7b4d5d48b6b5a75009d63f8f3e4ef4819b7e8139/lua/mini/indentscope.lua#L655-L656) mappings. As any motion within current buffer ideally should be.

craigmac:
> People don't think to try something so they do not think to try it.

I don't think "if we add it, people will use it" is a good measure of what should be included in core. I can't find any objective metric that shows users are using this plugin _en masse_, like vim-unimpaired levels of usage, which to me indicates it's not solving a problem many people have. Sure, it's nice, and I will try it out for sure, but as others have mentioned the cost may be too high. And if this becomes default, why not [vim-targets](https://github.com/wellle/targets.vim) style next/last targets as well?

test/functional/lua/cursor_pending_operator_spec.lua
```diff
@@ -0,0 +1,1428 @@
+--- Make sure the basic functionality of cursor-text-object works.
```
justinmk, line 1: these tests should live in `test/functional/editor` . the lua/ subdir is for the Lua "stdlib"

runtime/lua/vim/_cursor.lua
```diff
@@ -0,0 +1,111 @@
...
+---    Describe which way to adjust the cursor and how.
+---@field count number
+---    The number of times to call the command. Usually is just `1`.
+---@field direction "down" | "up"
+---    Which way to crop the text object. "up" means "operate from the cursor's
+---    position to the up/left-most position" and "down" means "operate from
+---    the cursor's position (including the cursor's current line, if
+---    applicable) to the down/right-most position".
+
+local _CURSOR
```
justinmk, line 23: no need for uppercase

justinmk:
Not clear that there's strong demand for this feature, so I agree it's premature to put this in core. Especially if it works as a plugin without changes to core.

If any changes are needed in core to make the plugin work better, let's focus on those specific changes first.

ColinKennedy:
> I don't think "if we add it, people will use it" is a good measure of what should be included in core. 

You made a personal assertion and I responded in kind with a personal assertion, that response is not part of the arguments in favor of this PR, it was a reply to you.

Anyway, though I find it regrettable that this PR didn't merge, I took a few minutes to make a test - https://gist.github.com/ColinKennedy/7e632f88570be89762b0ca8372769b72. There are 15 tests of various mapping combinations. In no instance did `[` or `]` shadow any another mapping. As far as I can tell that's evidence that this PR never prevented anyone from setting their own mappings.

PR #31160, opened 2024-11-11
feat(defaults): add ]p and [p linewise mappings

telemachus:
Paging @gpanders

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -376,6 +376,21 @@ do
       local linenr = vim.api.nvim_win_get_cursor(0)[1]
       vim.api.nvim_buf_set_lines(0, linenr, linenr, true, repeated)
     end, { desc = 'Add empty line below cursor' })
+
+    -- Paste linewise unconditionally
+    vim.keymap.set('n', '[p', function()
+      local body = vim.fn.getreg(vim.v.register)
+      body = vim.fn['repeat']({ body }, vim.v.count1)
+      local linenr = vim.api.nvim_win_get_cursor(0)[1]
+      vim.api.nvim_buf_set_lines(0, linenr - 1, linenr - 1, true, body)
+    end, { desc = 'Paste before linewise' })
+
+    vim.keymap.set('n', ']p', function()
+      local body = vim.fn.getreg(vim.v.register)
+      body = vim.fn['repeat']({ body }, vim.v.count1)
+      local linenr = vim.api.nvim_win_get_cursor(0)[1]
+      vim.api.nvim_buf_set_lines(0, linenr, linenr, true, body)
+    end, { desc = 'Paste after linewise' })
   end
 end
 
--- a/test/functional/editor/defaults_spec.lua
+++ b/test/functional/editor/defaults_spec.lua
@@ -200,6 +200,49 @@ describe('default', function()
           ]])
         end)
       end)
+
+      describe('[p', function()
+        it('pastes linewise before the cursor', function()
+          n.clear({ args_rm = { '--cmd' } })
+          n.insert([[first line]])
+          n.feed('^yiw[p')
+          n.expect([[
+          first
+          first line]])
+        end)
+
+        it('works with a count', function()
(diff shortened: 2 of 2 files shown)


telemachus:
Closes #31160 (hopefully)

telemachus:
This pull request has a Lua version of what seems to be the standard recommendation for how to handle this in Vim. After thinking about it, however, I wonder whether something like the following would be better.

```lua
vim.keymap.set('n', '[p', function()
  local body = vim.fn.getreg(vim.v.register)
  body = vim.fn["repeat"]({ body }, vim.v.count1)
  local linenr = vim.api.nvim_win_get_cursor(0)[1]
  vim.api.nvim_buf_set_lines(0, linenr - 1, linenr - 1, true, body)
end, { desc = 'Paste before linewise' })

vim.keymap.set('n', ']p', function()
  local body = vim.fn.getreg(vim.v.register)
  body = vim.fn["repeat"]({ body }, vim.v.count1)
  local linenr = vim.api.nvim_win_get_cursor(0)[1]
  vim.api.nvim_buf_set_lines(0, linenr, linenr, true, body)
end, { desc = 'Paste before linewise' })
```

gpanders:
> This pull request has a Lua version of what seems to be the standard way to handle this in Vim. After thinking about it, however, I wonder whether something like the following would be better.

I find this _much_ easier to understand what's happening, so if this implementation meets the requirements then let's go with this.

telemachus:
Done. The tests still pass, and I cannot think of an obvious way that the simpler implementation is different.

telemachus:
I think I need to add something to news.txt and vim_diff.txt, but I wanted to see about the implementation first.

telemachus:
Sorry, but after a little more thinking, I realized one difference: the more complex method correctly handles blockwise yanks. Because `nvim_buf_set_lines` does not allow newlines within the replacement argument, the simpler method errors on blockwise yanks.

I suspect many people use `[p` and `]p` to paste charwise yanks in linewise fashion, so maybe the limitation is one we can live with. But it may be better to go back to the more complex code.

gpanders:
> the more complex method correctly handles blockwise yanks.

Can you please add a test case for this? Would be good to capture this in case this is ever refactored by someone else in the future. 

Does using nvim_buf_set_text work?

telemachus:
> > the more complex method correctly handles blockwise yanks.
> 
> Can you please add a test case for this? Would be good to capture this in case this is ever refactored by someone else in the future.

Yes, I'll add a test for blockwise yanks. I also want to add a test that whatever method we settle on properly handles indentation. Unfortunately, I don't know how to get the functional tests to create a specific indentation context. (I'm guessing I have to set a filetype and maybe other variables. I'll look into it.)

> Does using nvim_buf_set_text work?

Maybe. But I think it would be far more complicated since I'd have to calculate the correct column to enter the text at—again, in order to handle proper indentation. vim-unimpaired gets indentation for free by using [`:help [p`](https://neovim.io/doc/user/change.html#%5Dp) (or `[P`). (Here's the vim-unimpaired code, if you want to look at it: https://github.com/tpope/vim-unimpaired/blob/master/plugin/unimpaired.vim#L409-L419.) In my latest version, I've been using the same technique. It works, but it's very much Vimscript in Lua clothing. I can't see a straightforward way to replicate these commands in Lua-like Lua.

Anyhow, I'll put this into draft for now. I may bug you in Matrix chat if I have questions about the testing framework.

aktau:
> I am going to propose over at Vim that they add variants for :put and :put! that handle indentation too. If that happens, Neovim will get those commands for free.

I suppose that once this is in, re-defining `]p`, `[p` to do what I'd want (meaning the vim-unimpaired versions) would become fairly short. Which is good enough for me. Can you link back the upstream issue once you've made it? Thanks for all the effort so far!

telemachus:
@aktau Good news: vim now has an `iput[!]` command that works like `put` but maintains indentation. You can map that to whatever you like. It was just merged, but I assume that vim will pick it up soon. https://github.com/vim/vim/pull/16886

aktau:
That's  excellent news. Thanks a lot @telemachus.

PR #31168, opened 2024-11-11
fix(tests): needing two calls to setup a screen is cringe

bfredl:
Before calling "attach" a screen object is just a dummy container for (row, col) values whose purpose is to be sent as part of the "attach" function call anyway. 

Just create the screen in an attached state directly. Keep the _complete_ `(row, col, options)` config together as the unit it is on the neovim side. It is still completely valid to later detach and re-attach as needed, including to another session.


test/functional/ui/screen.lua
```diff
@@ -180,14 +155,30 @@ local function _init_colors()
...
+--- @field ext_newgrid? boolean
+--- @field ext_popupmenu? boolean
+--- @field ext_wildmenu? boolean
+--- @field rgb? boolean
+--- @field _debug_float? boolean
+
 --- @param width? integer
 --- @param height? integer
+--- @param options? test.functional.ui.screen.Opts
+--- @param session? test.Session
```
zeertzjq, line 170:
```suggestion
--- @param session? test.Session|false
```

PR #31177, opened 2024-11-12
fix(api): nvim__complete_set requires completeopt=popup

glepnir:
Problem: even completeopt does not include popup using this api will auto create a floating preview window.

Solution: check cot flags include popup

### DIFF
--- a/test/functional/editor/completion_spec.lua
+++ b/test/functional/editor/completion_spec.lua
@@ -10,6 +10,7 @@ local fn = n.fn
 local command = n.command
 local api = n.api
 local poke_eventloop = n.poke_eventloop
+local exec_lua = n.exec_lua
 
 describe('completion', function()
   local screen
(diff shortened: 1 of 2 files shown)


test/functional/api/vim_spec.lua
```diff
@@ -5361,4 +5361,27 @@ describe('API', function()
     api.nvim__redraw({ win = 0, range = { 0, -1 } })
     n.assert_alive()
   end)
+
+  it('nvim__complete_set', function()
```
justinmk, line 5365: this doesn't seem like the right place for this test.

how about `test/functional/ui/popupmenu_spec.lua` ?

glepnir: make sense

test/functional/editor/completion_spec.lua
```diff
@@ -1326,4 +1327,29 @@ describe('completion', function()
       ]],
     })
   end)
+
+  describe('nvim__complete_set', function()
+    it('through an error when completeopt does not include popup', function()
```
justinmk, line 1332:
```suggestion
    it("fails when 'completeopt' does not include popup", function()
```

PR #26799, opened 2023-12-30
feat(float): 'previewpopup' option

glepnir:
Problem:
There is no support for Vim's 'previewpopup' option.

Solution:
Add 'previewpopup' to show :pedit, :ptag, and :psearch previews in a
floating window instead of a split. The value is
"height:N,width:N,border:style"; "border" accepts a named 'winborder'
style and follows 'winborder' when omitted.

Fix #9920

PR #25541, opened 2023-10-08
feat(ui): add pumborder

glepnir:
Problem: missing option 'pumborder'.

Solution: support 'pumborder' option. and refactor `win_redr_border` to `grid_redr_border`,
which redraw border for window grid and pum grid.

image

Fix #15551

PR #31180, opened 2024-11-12
fix(eval): add preview window relate argument in complete_info

glepnir:
Problem: what list missing preview relate flags.

Soltion: add preview_winid and preview_bufnr in coplete_info what list

### DIFF
--- a/src/nvim/insexpand.c
+++ b/src/nvim/insexpand.c
@@ -2851,6 +2851,8 @@ static void get_complete_info(list_T *what_list, dict_T *retdict)
 #define CI_WHAT_ITEMS           0x04
 #define CI_WHAT_SELECTED        0x08
 #define CI_WHAT_INSERTED        0x10
+#define CI_WHAT_PREVIEWWIN      0x20
+#define CI_WHAT_PREVIEWBUF      0x40
 #define CI_WHAT_ALL             0xff
   int what_flag;
 
@@ -2873,6 +2875,10 @@ static void get_complete_info(list_T *what_list, dict_T *retdict)
         what_flag |= CI_WHAT_SELECTED;
       } else if (strcmp(what, "inserted") == 0) {
         what_flag |= CI_WHAT_INSERTED;
+      } else if (strcmp(what, "preview_winid") == 0) {
+        what_flag |= CI_WHAT_PREVIEWWIN;
+      } else if (strcmp(what, "preview_bufnr") == 0) {
+        what_flag |= CI_WHAT_PREVIEWBUF;
       }
     }
   }
@@ -2929,10 +2935,19 @@ static void get_complete_info(list_T *what_list, dict_T *retdict)
     }
     if (ret == OK && (what_flag & CI_WHAT_SELECTED)) {
       ret = tv_dict_add_nr(retdict, S_LEN("selected"), selected_idx);
-      win_T *wp = win_float_find_preview();
-      if (wp != NULL) {
-        tv_dict_add_nr(retdict, S_LEN("preview_winid"), wp->handle);
-        tv_dict_add_nr(retdict, S_LEN("preview_bufnr"), wp->w_buffer->handle);
+    }
+  }
+
+  // Support for preview window and preview buffer information
+  if (ret == OK && (what_flag & (CI_WHAT_PREVIEWWIN | CI_WHAT_PREVIEWBUF))) {
+    win_T *wp = win_float_find_preview();
+    if (wp != NULL) {
+      if (what_flag & CI_WHAT_PREVIEWWIN) {
+        ret = tv_dict_add_nr(retdict, S_LEN("preview_winid"), wp->handle);
+      }
+
+      if (ret == OK && (what_flag & CI_WHAT_PREVIEWBUF)) {
+        ret = tv_dict_add_nr(retdict, S_LEN("preview_bufnr"), wp->w_buffer->handle);
       }
     }
   }
--- a/test/functional/ui/popupmenu_spec.lua
+++ b/test/functional/ui/popupmenu_spec.lua
@@ -1603,7 +1603,7 @@ describe('builtin popupmenu', function()
       end)
     end
 
-    describe('floating window preview #popup', function()
+    describe('floating window preview popup', function()
       it('pum popup preview', function()
         --row must > 10
         screen:try_resize(40, 11)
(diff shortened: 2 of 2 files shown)


justinmk:
Somewhat related: `cmdline` flag discussed in https://github.com/neovim/neovim/pull/31269#discussion_r1880449995 .

But in this case, `preview_winid` is a return value giving information about placement, whereas in https://github.com/neovim/neovim/pull/31269#discussion_r1880449995 , `cmdline` is used to control/decide a window.

glepnir:
Is there anything blocking here?

justinmk:
Looks like these fields were added for `nvim__complete_set` https://github.com/neovim/neovim/commit/e38027ef69f75653ee953b16ebf4a8652a3fb748

That would have been useful to mention the PR description.

The interface of `nvim__complete_set` is still pretty strange and needs more thought.

The help doc for `nvim__complete_set` doesn't mention `preview_bufnr` / `preview_winid` at all, nor does it link to `complete_info()` where the fields are documented. Let's do that here.

Are these fields related to `:help preview-window` and `:help 'previewpopup'`? It seems like we should implement 'previewpopup' before continuing this interface.

glepnir:
> The interface of nvim__complete_set is still pretty strange and needs more thought.

After I finish the other completion work..I will rethink this and provide a PR in the hope that you can give more guidance where to go.

> Are these fields related to :help preview-window and :help 'previewpopup'? It seems like we should implement 'previewpopup' (after this PR) before continuing to develop this interface--or explain why not.

Different. vim uses completepopup(i impleent in https://github.com/neovim/neovim/pull/25541) to set the maximum height and width of info popup. Lack of adaptivity. vim uses popup_findinfo function to get info popup winid/bufnr and use it with CompleteChanged. I add fields to complete_info here because our float interface lacks design compatibility with vim's popup. about previewpopup (https://github.com/neovim/neovim/pull/26799)  our current focus is not on these. So these PRs have been put on hold.

justinmk:
> Different. vim uses completepopup(

Reading Vim's help I see that both 'previewpopup' and 'completepopup' exist. 

Why are the fields in this PR called `preview_winid` and `preview_winid` if they aren't related to the previewpopup/previewwindow?

src/nvim/insexpand.c
```diff
@@ -2929,10 +2935,19 @@ static void get_complete_info(list_T *what_list, dict_T *retdict)
...
+  // Support for preview window and preview buffer information
+  if (ret == OK && (what_flag & (CI_WHAT_PREVIEWWIN | CI_WHAT_PREVIEWBUF))) {
+    win_T *wp = win_float_find_preview();
+    if (wp != NULL) {
+      if (what_flag & CI_WHAT_PREVIEWWIN) {
+        ret = tv_dict_add_nr(retdict, S_LEN("preview_winid"), wp->handle);
+      }
+
+      if (ret == OK && (what_flag & CI_WHAT_PREVIEWBUF)) {
+        ret = tv_dict_add_nr(retdict, S_LEN("preview_bufnr"), wp->w_buffer->handle);
```
justinmk, line 2950: Why do we need both of these fields? The buffer shown by `preview_winid` is implicitly the "preview bufnr".

src/nvim/insexpand.c
```diff
@@ -2929,10 +2935,19 @@ static void get_complete_info(list_T *what_list, dict_T *retdict)
...
-        tv_dict_add_nr(retdict, S_LEN("preview_bufnr"), wp->w_buffer->handle);
+    }
+  }
+
+  // Support for preview window and preview buffer information
+  if (ret == OK && (what_flag & (CI_WHAT_PREVIEWWIN | CI_WHAT_PREVIEWBUF))) {
+    win_T *wp = win_float_find_preview();
+    if (wp != NULL) {
+      if (what_flag & CI_WHAT_PREVIEWWIN) {
+        ret = tv_dict_add_nr(retdict, S_LEN("preview_winid"), wp->handle);
```
justinmk, line 2946: What is the actual use-case for this field? The docs are unclear. Is it for showing info about the selected item? Then it should be named something like `info_winid`.

The name `preview_winid` only makes sense if it's coupled to the the `:help preview-window` concept.

justinmk:
Let's focus on this PR and complete it while we have the context in our heads...

src/nvim/insexpand.c
glepnir: yes .. It is just a convenience. users do not need nvim_win_get_buf or winbufnr

src/nvim/insexpand.c
glepnir: Explain how it works now When the `info` of `complete-item` is set and `cot+=popup` We use a floating window here `(preview_winid, preview_bufnr)`, these two fields can be obtained in `complete_info()` with `ComplateChanged`, the main purpose is to set the filetype of buffer. lsp usually returns markdown or plaintext. 

The second is that lsp supports [completionitem_resolve request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem_resolve). Info may be obtained through a second request. Call nvim__complete_set in the asynchronous request callback of lsp to create a floating for the selected item and display info. eg

```lua
vim.api.nvim_create_autocmd('CompleteChanged', {
     callback = function()
                  -- it's better use a debounce timer here
                  local data = vim.fn.complete_info({'selected'})
                  client:requiest('completionitem_resolve', complete_item, function(err, result,...)
                            vim.api.nvim__complete_set(data.selected, {info = result })
                   end, bufnr) 
     end
})
```

Problems noticed. The first selected field of nvim__complete_set_info may be a bit redundant. This should be controlled by the user. Because if C-N/C-P is frequent, a timer is needed here for anti-shake. So the user should know clearly whether the currently selected item and the item before sending the request are the same item when calling nvim_complete_set_info. We do not need this selected parameter for verification.

```lua
                  local data = vim.fn.complete_info({'selected'})
                  client:requiest('completionitem_resolve', complete_item, function(err, result,...)
                           local curdata =  vim.fn.complete_info({'selected'})
                            if curdata.selected ~= data.selected then return end
                           -- a bit redundant ???????????????
                            vim.api.nvim__complete_set(curdata.selected, {info = result })
                   end, bufnr) 
 ````

src/nvim/insexpand.c
justinmk: Let's drop it for now. That is not a very strong use-case for increasing surface area here.

Keeping the surface-area minimal will help us focus on the important parts of this interface (which we are still figuring out).

src/nvim/insexpand.c
justinmk: > When the `info` of `complete-item` is set and `cot+=popup` We use a floating window here `(preview_winid, preview_bufnr)`, these two fields can be obtained in `complete_info()`

So is `info_winid` a good name for the field?

glepnir: Yes. This needs to be distinguished from the legacy preview.  just describes the how current works since you missed that PR

src/nvim/insexpand.c
glepnir: if we want drop the `nvim__complete_set` also need remove bufnr from result dict right ?

glepnir:
> i think we could add a filter to nvim_list_wins() . API allows optional parameters now. nvim_list_wins({kind='preview'}) (ideally it would be nvim_get_win(), but don't want to deal with that rn)

From Martix. This looks pretty promising instead of vim's `popup_findinfo`.

Issue #30986, opened 2024-10-29
InspectTree: exiting source buffer raises E855


xudyang1:
### Problem

Exiting source buffer where `:InspectTree` is called raises `E855: Autocommands caused command to abort`

### Steps to reproduce

1. nvim --clean repro.lua
2. `:InspectTree`
3. `:wincmd p | quit`
4. `E855: Autocommands caused command to abort` is reported

Or:

- `nvim --clean  -c 'InspectTree' -c 'wincmd p | quit' repro.lua`

### Expected behavior

No error raised

### Nvim version (nvim -v)

0.10.2

### Vim (not Nvim) behaves the same?

no

### Operating system/version

ubuntu 22.04

### Terminal name/version

windows terminal

### $TERM environment variable

xterm-256color

### Installation

app image

zeertzjq:
There seems to be another problem:
- Normally, running `:enew` in the only window showing the source buffer will close the `:InspectTree` window.
- However, if `bufhidden=unload` or `bufhidden=delete` or `bufhidden=wipe` is set in the source buffer, running `:enew` in the only window showing it doesn't close the `:InsepctTree` window.

PR #31181, opened 2024-11-12
fix(treesitter): add 'QuitPre' event to autocommands in inspect_tree

hyyker:
Problem: Quitting source buffer for ```:InspectTree``` command raises ```E855``` when source buffer and tree views are the only open buffers as addressed in #30986.

Solution: Add ```QuitPre``` event to autocmd handling closing/hiding the source buffer to close all open tree views. This allows nvim to quit when source and tree buffers are the only open windows.

### DIFF
--- a/runtime/lua/vim/treesitter/dev.lua
+++ b/runtime/lua/vim/treesitter/dev.lua
@@ -527,7 +527,7 @@ function M.inspect_tree(opts)
     end,
   })
 
-  api.nvim_create_autocmd({ 'BufHidden', 'BufUnload' }, {
+  api.nvim_create_autocmd({ 'BufHidden', 'BufUnload', 'QuitPre' }, {
     group = group,
     buffer = buf,
     once = true,


dundargoc:
> This workflow requires approval from a maintainer 

Bruh, not this again. @justinmk @bfredl @jamessan push whatever buttons needed to get rid of this for the love of god.

dundargoc:
Wait, this might not be possible for people's first PR ever.

@hykerr is this your first PR to any repository on github?

hykerr:
Yeah it's my first, @dundargoc. Had a stupid typo in the commit subject so now it has to be approved again. That's just mb

dundargoc:
No worries, I just had a brainfart. Thabks for the PR :)

dundargoc:
Does this close https://github.com/neovim/neovim/issues/30986?

guilherme-batalheiro:
Hey @hykerr, just out of curiosity and to learn more - how does adding the QuitPre event fix the problem? It doesn’t seem to be doing much at first look, since the window is already closed, or am I mistaken? I was working on this issue all day.

hykerr:
@guilherme-batalheiro, I'm not entirely sure how it works but I'll try my best to explain my understanding. Taking a look at the help pages for the events in the autocmd:

```
BufHidden
Before a buffer becomes hidden: when there are
no longer windows that show the buffer, but
the buffer is not unloaded or deleted.

Not used for ":qa" or ":q" when exiting Vim.
```

```
BufUnload
Before unloading a buffer, when the text in
the buffer is going to be freed.
```

```
QuitPre
When using :quit, :wq or :qall, before
deciding whether it closes the current window
or quits Vim.  For :wq the buffer is written
before QuitPre is triggered.  Can be used to
close any non-essential window if the current
window is the last ordinary window.
```

When ```BufHidden``` is triggered, nvim checks if any associated autocmd made the current buffer the only window. This happens here because we close all open tree windows and the bug only happens if the tree window(s) and the source window are the only open windows. Because of that, E855 is thrown.

```c
// buffer.c: 596
if (abort_if_last && one_window(win)) {
    // Autocommands made this the only window.
    emsg(_(e_auabort));
    return false;
}
```

When doing ```:q```, there is no such check and ```QuitPre``` is called quite early:
```c
// ex_docmd.c: 4748
// Trigger QuitPre and maybe ExitPre
if (before_quit_autocmds(wp, false, eap->forceit)) {
    return;
}
                                                                                                                                          
// If there is only one relevant window we will exit. 
if (check_more(false, eap->forceit) == OK && only_one_window()) { 
    exiting = true;
}
```

I hope this makes it more clear :)

I'm realizing now that my fix for the original bug introduces another bug:
When the source buffer is open multiple times (e.g. when split) and you quit one of these windows, it will also close the tree view which shouldn't happen (that's probably why the code didn't use QuitPre to begin with). The easiest fix would probably be to check if there are multiple source windows open like so:

```lua
api.nvim_create_autocmd({ 'BufHidden', 'BufUnload', 'QuitPre' }, {
    group = group,
    buffer = buf,
    -- remove this line because E855 would happen again
    -- once = true, 
    callback = function()
      -- don't close inpector window if source buffer
      -- has more than one open window
      if #vim.fn.win_findbuf(buf) > 1 then
          return
      end

      -- close all tree windows
        close_win(window)
      end
    end,
  })
```

I'll open an issue (for the sake of completeness) and a pr with the above fix. But maybe there might be a more elegant way to work around this?

guilherme-batalheiro:
@hykerr tank you so much for the explanation! I didn't know that Neovim made that check for BufHidden.

PR #31182, opened 2024-11-12
test(treesitter): update inspect_tree tests

guilhas07:
To prevent #30986 and #31198 regression update inspect_tree tests.

### DIFF
--- a/test/functional/treesitter/inspect_tree_spec.lua
+++ b/test/functional/treesitter/inspect_tree_spec.lua
@@ -120,14 +120,17 @@ describe('vim.treesitter.inspect_tree', function()
   end)
 
   it('updates source and tree buffer windows and closes them correctly', function()
+    local name = t.tmpname()
+    n.command('edit ' .. name)
     insert([[
       print()
       ]])
+    n.command('set filetype=lua | write')
 
     -- setup two windows for the source buffer
     exec_lua(function()
       _G.source_win = vim.api.nvim_get_current_win()
-      vim.api.nvim_open_win(0, false, {
+      _G.source_win2 = vim.api.nvim_open_win(0, false, {
         win = 0,
         split = 'left',
       })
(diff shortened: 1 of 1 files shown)


zeertzjq:
This isn't correct. It leads to a `No write since last change error`.

guilhas07:
> This isn't correct. It leads to a `No write since last change` error.

I know that occurs. It's because the content on the buffer isn't saved. 

https://github.com/neovim/neovim/blob/c6ba0607384ee5609add4676b1e86e03601ce230/test/functional/treesitter/inspect_tree_spec.lua#L123-L125

I thought of saving the buffer contents, and after removing the file but if felt wrong. Another approach would be to split this into a simpler test without buffer contents. What do you think?

zeertzjq:
You can use `n.expect_exit(n.command, 'q!')`

guilhas07:
> You can use `n.expect_exit(n.command, 'q!')`

Wouldn't that ignore the error `E855: Autocommands caused command to abort` ?

guilhas07:
@zeertzjq Sorry for the ping, but what do you think? The tests now cover the previously mentioned fixes and the error isn't masked like it previously was.

test/functional/treesitter/inspect_tree_spec.lua
```diff
@@ -120,22 +120,24 @@ describe('vim.treesitter.inspect_tree', function()
   end)
 
   it('updates source and tree buffer windows and closes them correctly', function()
+    local name = t.tmpname()
+    n.command('e ' .. name)
```
justinmk, line 124: use full command names in code. helps when searching for usages. same for the rest of this file
```suggestion
    n.command('edit ' .. name)
```

test/functional/treesitter/inspect_tree_spec.lua
```diff
@@ -120,22 +120,24 @@ describe('vim.treesitter.inspect_tree', function()
...
     insert([[
       print()
       ]])
+    n.command('set filetype=lua | w')
 
     -- setup two windows for the source buffer
     exec_lua(function()
       _G.source_win = vim.api.nvim_get_current_win()
-      vim.api.nvim_open_win(0, false, {
+      _G.source_win_copy = vim.api.nvim_open_win(0, false, {
```
justinmk, line 133: what is this a "copy" of?

justinmk:
```suggestion
      _G.source_win = vim.api.nvim_open_win(0, false, {
```

justinmk (approved):

test/functional/treesitter/inspect_tree_spec.lua
guilhas07: Of the window above `source_win`. Maybe `copy` isn't the best name, because it is simply another window displaying the source buffer.

justinmk: `source_win2` or something like that seems less confusing

PR #31186, opened 2024-11-13
feat(defaults): dot-repeat [<Space>

telemachus:
Problem: `[<Space>` and `]<Space>` do not support repetition.

Solution: use `operatorfunc` and `g@l` to make these mappings dot repeatable.

### DIFF
--- a/runtime/lua/vim/_buf.lua
+++ b/runtime/lua/vim/_buf.lua
@@ -1,0 +1,23 @@
+local M = {}
+
+--- Adds one or more blank lines above or below the cursor.
+-- TODO: move to _defaults.lua once it is possible to assign a Lua function to options #25672
+--- @param above? boolean Place blank line(s) above the cursor
+local function add_blank(above)
+  local offset = above and 1 or 0
+  local repeated = vim.fn['repeat']({ '' }, vim.v.count1)
+  local linenr = vim.api.nvim_win_get_cursor(0)[1]
+  vim.api.nvim_buf_set_lines(0, linenr - offset, linenr - offset, true, repeated)
+end
+
+-- TODO: move to _defaults.lua once it is possible to assign a Lua function to options #25672
+function M.space_above()
+  add_blank(true)
+end
+
+-- TODO: move to _defaults.lua once it is possible to assign a Lua function to options #25672
+function M.space_below()
+  add_blank()
+end
+
+return M
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -366,16 +366,16 @@ do
 
     -- Add empty lines
     vim.keymap.set('n', '[<Space>', function()
-      local repeated = vim.fn['repeat']({ '' }, vim.v.count1)
-      local linenr = vim.api.nvim_win_get_cursor(0)[1]
-      vim.api.nvim_buf_set_lines(0, linenr - 1, linenr - 1, true, repeated)
-    end, { desc = 'Add empty line above cursor' })
+      -- TODO: update once it is possible to assign a Lua function to options #25672
+      vim.go.operatorfunc = "v:lua.require'vim._buf'.space_above"
+      return 'g@l'
+    end, { expr = true, desc = 'Add empty line above cursor' })
 
     vim.keymap.set('n', ']<Space>', function()
-      local repeated = vim.fn['repeat']({ '' }, vim.v.count1)
-      local linenr = vim.api.nvim_win_get_cursor(0)[1]
-      vim.api.nvim_buf_set_lines(0, linenr, linenr, true, repeated)
-    end, { desc = 'Add empty line below cursor' })
+      -- TODO: update once it is possible to assign a Lua function to options #25672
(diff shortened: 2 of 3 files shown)


telemachus:
@gpanders I wasn't sure whether this deserved to go into news.txt or the documentation. (Users may expect these mappings to support dot repetition since they change the buffer.) If you want me to add explicit mention of this somewhere, just let me know.

gpanders:
> @gpanders I wasn't sure whether this deserved to go into news.txt or the documentation. (Users may expect these mappings to support dot repetition since they change the buffer.) If you want me to add explicit mention of this somewhere, just let me know.

I think it's fine to skip news for this.

runtime/lua/vim/_defaults.lua
```diff
@@ -366,16 +366,14 @@ do
 
     -- Add empty lines
     vim.keymap.set('n', '[<Space>', function()
-      local repeated = vim.fn['repeat']({ '' }, vim.v.count1)
-      local linenr = vim.api.nvim_win_get_cursor(0)[1]
-      vim.api.nvim_buf_set_lines(0, linenr - 1, linenr - 1, true, repeated)
-    end, { desc = 'Add empty line above cursor' })
+      vim.go.operatorfunc = "v:lua.require'vim._unimpaired'.space_above"
```
gpanders, line 369: @famiu @bfredl Do we have a tracking issue to allow assigning Lua functions directly to options? If so would be good to mention it here so that we can update this in the future.

famiu: Not a tracking issue for this specifically, but it is one of the items in #25672

telemachus: If it makes sense, I am happy to add a TODO (for myself or just in general) linking to that issue.

gpanders: Yes please do

runtime/lua/vim/_unimpaired.lua
justinmk, line 1: if we can't think of a long-term term for the module then put the functions in _defaults.lua.

maybe `_buf.lua` is a long-term name, in anticipation of some future `vim.buf` module.

gpanders: >if we can't think of a long-term term for the module then put the functions in _defaults.lua.

We can't because `operatorfunc` requires `require`, so these functions should be in their own module (`_defaults.lua` is not meant to be `require`d).

The motivation behind my [earlier comment](https://github.com/neovim/neovim/pull/31186/files#r1844127334) was because of this. Whenever we are able to assign Lua functions to an option directly, then we _can_ remove the separate `_unimpaired` module and put the functions in `_defaults.lua` directly.

Not opposed to `_buf.lua` or whatever as a separate name, but long term these functions won't be in a separate module at all so I'm not too concerned about the naming.

runtime/lua/vim/_defaults.lua
telemachus: Done

runtime/lua/vim/_buf.lua
```diff
@@ -0,0 +1,20 @@
...
+--- Adds one or more blank lines above or below the cursor.
+--- @param above? boolean Place blank line(s) above the cursor
+local function add_blank(above)
+  local offset = above and 1 or 0
+  local repeated = vim.fn['repeat']({ '' }, vim.v.count1)
+  local linenr = vim.api.nvim_win_get_cursor(0)[1]
+  vim.api.nvim_buf_set_lines(0, linenr - offset, linenr - offset, true, repeated)
+end
+
+function M.space_above()
```
gpanders, line 12: Add the TODO comments to this file too. Once the TODO is resolved these should be inlined into `_defaults.lua` so let's make that clear to future readers.

runtime/lua/vim/_unimpaired.lua
telemachus: Done: moved to `_buf.lua`. (I didn't put them in `_defaults.lua` since (I think) the assignment only works if there is an exported function, and I didn't think it made sense to have `_defaults.lua` export anything.) Let me know if you disagree.

runtime/lua/vim/_buf.lua
telemachus: Yes, sorry, that makes sense. Done.

gpanders (approved):

runtime/lua/vim/_unimpaired.lua
justinmk: > Whenever we are able to assign Lua functions to an option directly, then we _can_ remove the separate `_unimpaired` module and put the functions in `_defaults.lua` directly.

ok should we mention that in a comment on the functions?

runtime/lua/vim/_buf.lua
```diff
@@ -0,0 +1,21 @@
+-- TODO: move to _defaults.lua once it is possible to assign a Lua function to options #25672
```
justinmk, line 1: let's add this todo on each of the functions. since `_buf.lua` might be extended for other purposes later, having this top-level TODO would be confusing

telemachus: Done

Issue #22534, opened 2023-03-05
windows: drop support for MSVCRT and get rid of WCHAR workarounds


bfredl:
Background: There are two common libc implementations for windows supplied by microsoft https://www.msys2.org/docs/environments/#msvcrt-vs-ucrt

Importantly, the newer runtime UCRT has native support for UTF-8, including setting it as the _active codepage_ even before `main()` gets called. On the other hand, msvcrt, which was the runtime when we initially implemented windows support, does not support UTF-8 at all, and requires the use of `wmain()` entry point (#7060) as well as using WCHAR variants of basic functions, such as opening a file (#18122) . Also, we use libuv for a lot of basic file operations even in fully sync contexts, as libuv hides the difference between incompatible windows runtimes for us (and does UTF-8 from/to WCHAR conversion internally)

This is a proposal for neovim to **only support the UCRT runtime on windows** or at least formally drop support for msvcrt, realistically for the 0.10 development cycle. Then we can use standard libc functions for manipulating arguments and filenames, and remove/simplify a lot of the existing workarounds. Our own MSVC binaries now use UCRT with UTF-8 fully configured (#22384). 

Furthermore, there is a code path to compile neovim using MINGW/MSYS2 (earlier discussion: #22416 ). This would need to be adjusted to

1. supply the same manifest information when compiled with UCRT (as in #22384)
2. fail the build when attempting to build with MSVCRT.

bfredl:
For 0.10 on windows it will be **strongly** recommended to only use UCRT builds with. We don't actively prevent other builds but they are deprecated. Any bug which only occurs on MSVCRT will have the recommended fix: switch to a UCRT build.

For 0.11 we will actively require UCRT on windows, and remove more support code which is no longer needed for a UCRT-only windows app.

justinmk:
Can `:checkhealth` detect MSVCRT build and show a warning?

axelhj:
I got the error `#  error UCRT is the only supported C runtime on windows` when trying to build from master today. Does this mean this issue can be closed? Or it includes additional refactorings?

PR #31193, opened 2024-11-13
refactor(windows)!: only support UCRT, even for mingw

bfredl:
work towards #22534 .

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -45,6 +45,10 @@ API
 • Renamed `nvim__id_dictionary` (unsupported/experimental API) to
   `nvim__id_dict`.
 
+BUILD
+
+On Windows, only building with the UCRT runtime is supported.
+
 DEFAULTS
 
 • |]d-default| and |[d-default| accept a count.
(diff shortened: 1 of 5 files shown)


dundargoc (approved):
Looks good. There is a `target_link_libraries(test_lib INTERFACE -municode)` in `test/functional/fixtures/CMakeLists.txt`, can that be removed as well?

runtime/doc/news.txt
```diff
@@ -148,6 +148,10 @@ VIMSCRIPT
   treats BIN, STR and FIXSTR as separate types. Any of these is returned as a
   string if possible, or a |blob| if the value contained embedded NUL:s.
 
+BUILD SYSTEM
```
justinmk, line 151: The sections are in alphabetical order, this should go after "API". Could also just name it `BUILD` or maybe `CORE` or `INTERNAL`

justinmk:
This blurb would be helpful in the commit message for each of these planned changes:

    The newer UCRT runtime has native support for UTF-8, including setting
    it as the _active codepage_ even before `main()` gets called. Whereas
    MSVCRT does not support UTF-8 and requires `wmain()`.

runtime/doc/news.txt
bfredl: it is not `INTERNAL` . This note is for third-party distros (one of the main consumers of release notes)

justinmk (approved):

PR #31195, opened 2024-11-13
fix(api): consistent usage/documentation of hl_group

luukvbaal:
Problem:  Documentation for "hl_group" in nvim_buf_set_extmark() is
          unclear. "hl_group" in nvim_echo() does not accept
          highlight group id.
Solution: Move documentation for highlight group name/id to first
          mention of hl_group. Update nvim_echo() to accept highlight
          group id.

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -658,8 +658,8 @@ nvim_echo({chunks}, {history}, {opts})                           *nvim_echo()*
 
     Parameters: ~
       • {chunks}   A list of `[text, hl_group]` arrays, each representing a
-                   text chunk with specified highlight. `hl_group` element can
-                   be omitted for no highlight.
+                   text chunk with specified highlight group name or ID.
+                   `hl_group` element can be omitted for no highlight.
       • {history}  if true, add to |message-history|.
       • {opts}     Optional parameters.
                    • verbose: Message is printed as a result of 'verbose'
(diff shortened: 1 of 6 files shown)


runtime/doc/api.txt
```diff
@@ -2750,15 +2750,15 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
...
-                  • sign_hl_group: name of the highlight group used to
-                    highlight the sign column text.
-                  • number_hl_group: name of the highlight group used to
-                    highlight the number column.
-                  • line_hl_group: name of the highlight group used to
-                    highlight the whole line.
-                  • cursorline_hl_group: name of the highlight group used to
-                    highlight the sign column text when the cursor is on the
-                    same line as the mark and 'cursorline' is enabled.
+                  • sign_hl_group: highlight group used to highlight the sign
```
bfredl, line 2753: repeating `highlight` sounds a bit unnattural, maybe just ` highlight group used for the sign column text`  (and so on for the others)

test/functional/api/vim_spec.lua
```diff
@@ -3620,6 +3629,12 @@ describe('API', function()
         msg_a{15:msg_b}{16:msg_c}                         |
       ]],
       }
+      async_meths.nvim_echo({
+        { 'msg_a' },
+        { 'msg_b', api.nvim_get_hl_id_by_name('Statement') },
+        { 'msg_c', api.nvim_get_hl_id_by_name('Special') },
+      }, true, {})
+      screen:expect_unchanged()
```
bfredl, line 3637: nit: using `screen:expect_unchanged()` when a `screen:expect()` works fine (just needing some minor change) is considered an anti-pattern, as is it less robust than expecting a different state. 

In this case, better to change the messages slightly (like using `msg_d` or whatever), and then assert this distinct state.

bfredl: I e `screen:unchanged()` is more for asserting that nothing happened at all UI-wise, which is not the case here. we are actively interested in a _different_ message.

luukvbaal: Hmm yeah I knew it was intended to be used like that but I've also used it (here and in statuslcolumn/sign tests) as a way to test that different commands yield the same screen state. Adding new `screen:expect()` in such cases usually leads to warnings.

justinmk (approved):

clason:
ASAN test was failing (and now fails on all PRs, preventing automerge 😞 )

dundargoc:
I hate all of you. Stop merging PRs with failures lol.

clason:
Good thing you already recorded your talk 😄

luukvbaal:
Oopsie woopsie

PR #31205, opened 2024-11-14
fix(messages): proper multiline Lua print() messages

### LINKED ISSUE #21044: `print` should only send one `msg_show` event with `ext_messages`
### Feature already in Vim?

_No response_

### Feature description

Right now if you do something like the below, a separate `msg_show` event will be created for every line in the `print` output.

It would be better to get all the content in one `msg_show` event. This is already the case for most commands like `:ls`.

```lua
print(vim.inspect({a=1, b=2}))
```

luukvbaal:
Problem:  Separate message emitted for each newline present in Lua
          print() arguments.
Solution: Make msg_multiline() handle NUL bytes. Refactor print() to use
          msg_multiline(). Refactor vim.print() to use print().

Fix #21044

### DIFF
--- a/src/nvim/message.c
+++ b/src/nvim/message.c
@@ -249,35 +249,33 @@ bool msg(const char *s, const int hl_id)
   return msg_hl_keep(s, hl_id, false, false);
 }
 
-/// Similar to msg_outtrans, but support newlines and tabs.
-void msg_multiline(const char *s, int hl_id, bool check_int, bool hist, bool *need_clear)
+/// Similar to msg_outtrans_len, but support newlines and tabs.
+void msg_multiline(String str, int hl_id, bool check_int, bool hist, bool *need_clear)
   FUNC_ATTR_NONNULL_ALL
 {
-  const char *next_spec = s;
-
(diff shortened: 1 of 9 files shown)


src/nvim/message.c
```diff
@@ -249,35 +249,37 @@ bool msg(const char *s, const int hl_id)
   return msg_hl_keep(s, hl_id, false, false);
 }
 
-/// Similar to msg_outtrans, but support newlines and tabs.
-void msg_multiline(const char *s, int hl_id, bool check_int, bool hist, bool *need_clear)
+/// Similar to msg_outtrans_len, but support newlines and tabs.
+void msg_multiline(const char *str, int hl_id, bool check_int, bool hist, bool *need_clear, int len)
```
bfredl, line 253: Could change it to `String str`. Callers with unknown length can use `cstr_as_string`  (works without copy)

przepompownia:
With this change I can remove the workaround used to reassemble multi-line output split into separate events - one per line.

justinmk (approved):
Beautiful

PR #31208, opened 2024-11-14
fix(lsp): hover keymap

snelling-a:
the default `K` keymap for `vim.lsp.buf.hover()` is implemented differently to the other default lsp keymaps

e.g.

```lua
-- https://github.com/neovim/neovim/blob/7d771c3eeef5b4dca9ebc5ed6f7ca03f2b26b6bc/runtime/lua/vim/_defaults.lua#L170-L172
vim.keymap.set('n', 'grr', function()
    vim.lsp.buf.references()
end, { desc = 'vim.lsp.buf.references()' })
```

vs

```lua
-- https://github.com/neovim/neovim/blob/7d771c3eeef5b4dca9ebc5ed6f7ca03f2b26b6bc/runtime/lua/vim/lsp.lua#L368
vim.keymap.set('n', 'K', vim.lsp.buf.hover, { buffer = bufnr, desc = 'vim.lsp.buf.hover()' })
```

this makes it impossible to override the options of `hover` but still keep the default keymap

i.e.

```lua
-- this works
local signature_help = vim.lsp.buf.signature_help
---@param config vim.lsp.buf.signature_help.Opts
---@diagnostic disable-next-line: duplicate-set-field
vim.lsp.buf.signature_help = function(config)
    signature_help(vim.tbl_extend("force", config or {}, { border = "rounded" }))
end

-- this does not
local hover = vim.lsp.buf.hover
---@param config vim.lsp.buf.hover.Opts
---@diagnostic disable-next-line: duplicate-set-field
vim.lsp.buf.hover = function(config)
    hover(vim.tbl_extend("force", config or {}, {
        border = "rounded",
        title = client.name,
        title_pos = "left",
    }))
end

```

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -365,7 +365,9 @@ function lsp._set_defaults(client, bufnr)
       and is_empty_or_default(bufnr, 'keywordprg')
       and vim.fn.maparg('K', 'n', false, false) == ''
     then
-      vim.keymap.set('n', 'K', vim.lsp.buf.hover, { buffer = bufnr, desc = 'vim.lsp.buf.hover()' })
+      vim.keymap.set('n', 'K', function()
+        vim.lsp.buf.hover()
+      end, { buffer = bufnr, desc = 'vim.lsp.buf.hover()' })
     end
   end)
   if client.supports_method(ms.textDocument_diagnostic) then
--- a/test/functional/plugin/lsp_spec.lua
+++ b/test/functional/plugin/lsp_spec.lua
@@ -466,10 +466,17 @@ describe('LSP', function()
               true,
               exec_lua(function()
                 local keymap --- @type table<string,any>
+                local called = false
+                local origin = vim.lsp.buf.hover
+                vim.lsp.buf.hover = function()
+                  called = true
+                end
                 vim._with({ buf = _G.BUFFER }, function()
                   keymap = vim.fn.maparg('K', 'n', false, true)
                 end)
-                return keymap.callback == vim.lsp.buf.hover
+                keymap.callback()
+                vim.lsp.buf.hover = origin
+                return called
               end)
             )
             client.stop()
@@ -480,13 +487,13 @@ describe('LSP', function()
           eq('', get_buf_option('omnifunc'))
           eq('', get_buf_option('formatexpr'))
           eq(
-            '',
+            true,
             exec_lua(function()
               local keymap --- @type string
               vim._with({ buf = _G.BUFFER }, function()
                 keymap = vim.fn.maparg('K', 'n', false, false)
               end)
-              return keymap
+              return keymap:match('<Lua %d+: .+/runtime/lua/vim/lsp%.lua:%d+>') ~= nil
             end)
           )
         end,


MariaSolOs:
@snelling-a I don't think this change is needed. The issue is that you should return the result of the original function in your override:
```lua
local hover = vim.lsp.buf.hover
---@param config vim.lsp.buf.hover.Opts
---@diagnostic disable-next-line: duplicate-set-field
vim.lsp.buf.hover = function(config)
    return hover(vim.tbl_extend("force", config or {}, { // Add the return!
        border = "rounded",
        title = client.name,
        title_pos = "left",
    }))
end
```

Maverickx0:
@MariaSolOs Cannot take affect.

snelling-a:
> @snelling-a I don't think this change is needed. The issue is that you should return the result of the original function in your override:
> 
> ```lua
> local hover = vim.lsp.buf.hover
> ---@param config vim.lsp.buf.hover.Opts
> ---@diagnostic disable-next-line: duplicate-set-field
> vim.lsp.buf.hover = function(config)
>     return hover(vim.tbl_extend("force", config or {}, { // Add the return!
>         border = "rounded",
>         title = client.name,
>         title_pos = "left",
>     }))
> end
> ```

@MariaSolOs, thank you for your feedback. your suggestion does make sense, unfortunately, it doesn't work. 
i'm not exactly sure why, i would guess it has something to do with the hover keymap using the definition of the hover function vs. the `g**` keymaps using a function call. but when i get down into to the c code i get a bit lost 😅 

either way, the keymaps are implemented inconsistently 🤔 

---

on a side note, thank you for all you've done for the project and community. your talk last year at neovimconf was really cool and got me looking for opportunities to contribute

mfussenegger:
>  I don't think this change is needed. The issue is that you should return the result of the original function in your override:

If the default keymap is setup before the function was monkey patched, the keymap will forever contain the original function reference and won't see changes to it.
By making it lazy it will pick up the current `hover` function whenever the keymap is used.

Change seems good to me but the test failures look related

MariaSolOs:
> If the default keymap is setup before the function was monkey patched, the keymap will forever contain the original function reference and won't see changes to it. By making it lazy it will pick up the current `hover` function whenever the keymap is used.

Oh I see, that makes sense. I was using my own configuration as a reference and I override hover as shown above and it works, but it's indeed better if this change makes it work regardless of the order!

Sorry for the confusion :)

snelling-a:
> @snelling-a I don't think this change is needed. The issue is that you should return the result of the original function in your override:
> 
> ```lua
> local hover = vim.lsp.buf.hover
> ---@param config vim.lsp.buf.hover.Opts
> ---@diagnostic disable-next-line: duplicate-set-field
> vim.lsp.buf.hover = function(config)
>     return hover(vim.tbl_extend("force", config or {}, { // Add the return!
>         border = "rounded",
>         title = client.name,
>         title_pos = "left",
>     }))
> end
> ```

@MariaSolOs on second look, something else was causing the issue. it was wrapped in another function that wasn't working as intended. if i implement it as you described, it works perfectly 👍 

@mfussenegger the failing tests were related to my change. I'm not entirely sure how to fix it :diagonally

MariaSolOs:
Sounds like this PR is not needed anymore then? @snelling-a @mfussenegger

ddad431:
> Sounds like this PR is not needed anymore then? @snelling-a @mfussenegger

@MariaSolOs  Is this question has been solved? Now, I can not  change the hover config without the remap. 

Form above you reply, I thought maybe you haven't understand other people's demands. The neovim had given the deafult keymap. So this indicated  that user can config specify lsp action function without remap. It should work the same as signature_help. 

For the current behavior is like above maintainer said, the keymap will forever contain the original function reference and won't see changes to it if you don't remap.

But if I'm wrong, please tell me.

MariaSolOs:
> @MariaSolOs Is this question has been solved? Now, I can not change the hover config without the remap.
> 
> Form above you reply, I thought maybe you haven't understand other people's demands. The neovim had given the deafult keymap. So this indicated that user can config specify lsp action function without remap. It should work the same as signature_help.
> 
> For the current behavior is like above maintainer said, the keymap will forever contain the original function reference and won't see changes to it if you don't remap.
> 
> But if I'm wrong, please tell me.

I do understand people's demands and why the current keymap definition is problematic. My comment comes from the fact that the implementer of the PR seems to have found a workaround though.

Feel free to collaborate here and update the tests to fix CI and merge this change! :)

snelling-a:
> > @MariaSolOs Is this question has been solved? Now, I can not change the hover config without the remap.
> > Form above you reply, I thought maybe you haven't understand other people's demands. The neovim had given the deafult keymap. So this indicated that user can config specify lsp action function without remap. It should work the same as signature_help.
> > For the current behavior is like above maintainer said, the keymap will forever contain the original function reference and won't see changes to it if you don't remap.
> > But if I'm wrong, please tell me.
> 
> I do understand people's demands and why the current keymap definition is problematic. My comment comes from the fact that the implementer of the PR seems to have found a workaround though.
> 
> Feel free to collaborate here and update the tests to fix CI and merge this change! :)


@mar

> > @MariaSolOs Is this question has been solved? Now, I can not change the hover config without the remap.
> > Form above you reply, I thought maybe you haven't understand other people's demands. The neovim had given the deafult keymap. So this indicated that user can config specify lsp action function without remap. It should work the same as signature_help.
> > For the current behavior is like above maintainer said, the keymap will forever contain the original function reference and won't see changes to it if you don't remap.
> > But if I'm wrong, please tell me.
> 
> I do understand people's demands and why the current keymap definition is problematic. My comment comes from the fact that the implementer of the PR seems to have found a workaround though.
> 
> Feel free to collaborate here and update the tests to fix CI and merge this change! :)



> I have looked into the failures. Maybe check the call for callback instead of reference for callback.
> 
> ```diff
> --- a/test/functional/plugin/lsp_spec.lua
> +++ b/test/functional/plugin/lsp_spec.lua
> @@ -466,10 +466,16 @@ describe('LSP', function()
>                true,
>                exec_lua(function()
>                  local keymap --- @type table<string,any>
> +                local called = false
> +                local origin = vim.lsp.buf.hover
> +                vim.lsp.buf.hover = function()
> +                  called = true
> +                end
>                  vim._with({ buf = _G.BUFFER }, function()
>                    keymap = vim.fn.maparg('K', 'n', false, true)
>                  end)
> -                return keymap.callback == vim.lsp.buf.hover
> +                keymap.callback()
> +                vim.lsp.buf.hover = origin
> +                return called
>                end)
>              )
>              client:stop()
> ```

thanks for the assist @ddad431

snelling-a:
@MariaSolOs @mfussenegger 

WDYT?

justinmk:
When merging, please include the problem description in the commit message.

MariaSolOs:
> @MariaSolOs @mfussenegger
> 
> WDYT?

LGTM! Thank you for the fix!

PR #31220, opened 2024-11-15

clason:
- **vim-patch:460799d: runtime(netrw): small fixes to netrw#BrowseX**
- **vim-patch:0acd3ab: runtime(sh): better function support for bash/zsh in indent script**
- **vim-patch:7c3b65e: runtime(cfg): only consider leading // as starting a comment**


zeertzjq:
Missing `runtime/indent/testdir/` changes

zeertzjq (approved):

PR #31224, opened 2024-11-15
fix(ui): no fast context for prompt message kinds

luukvbaal:
Problem:  No longer able to show prompt messages with vim.ui_attach(). 
Solution: Do not execute callback in fast context for prompt message
          kinds. These events must be safe to show the incoming message
          so the event itself serves to indicate that the message
          should be shown immediately.

### DIFF
--- a/src/nvim/ui.c
+++ b/src/nvim/ui.c
@@ -717,6 +717,13 @@ void ui_call_event(char *name, bool fast, Array args)
 {
   bool handled = false;
   UIEventCallback *event_cb;
+
+  // Prompt messages should be shown immediately so must be safe
+  if (strcmp(name, "msg_show") == 0) {
+    char *kind = args.items[0].data.string.data;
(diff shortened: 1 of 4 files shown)


folke:
Great!

folke:
Was just investigating that `return_prompt` and see that you also took care of that.

src/nvim/ui.c
```diff
@@ -717,6 +717,13 @@ void ui_call_event(char *name, bool fast, Array args)
 {
   bool handled = false;
   UIEventCallback *event_cb;
+
+  // Prompt messages should be shown immediately so must be safe
+  if (strcmp(name, "msg_show") == 0) {
+    char *kind = args.items[0].data.string.data;
+    fast = !kind || (strncmp(kind, "confirm", 7) != 0 && strcmp(kind, "return_prompt") != 0);
```
justinmk, line 724: fwiw we have `strequal`

justinmk (approved):

PR #31245, opened 2024-11-17
fix(checkhealth): disable 'listchars'

### LINKED ISSUE #31145: Health headings highlight breaks with listchars multispace
### Problem

Image

- listchars multispace breaks the visual heading of `:checkhealth`. 
- `:set listchars-=multispace:••` restores the heading. 
- Running `:Inspect` on the space part in the heading only shows the hl group of the heading, so no clue how this multispace gets highlighted.
- `:hi clear Whitespace` (which should highlight listchars) doesn't change anything.

### Steps to reproduce

nvim --clean
:set list listchars+=multispace:++
:checkhealth

### Expected behavior

A continuous header, with or without the listchars char (I don't care if it shows ++ or nothing, but the background should be the same).

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1096+g079e5f4f9b-Homebrew

Does not apply

### Operating system/version

MacOS v14.6.1

### Terminal name/version

iTerm2 3.5.5

tmux-256color

### Installation

homebrew install --HEAD neovim

yochem:
Problem:
'listchars' (in particular multispace) breaks visual heading due to
`Whitespace` highlight group.

Solution:
Disable 'list' (and thus all listchars) by default for `checkhealth`
files.

Fixes #31145

### DIFF
--- a/runtime/ftplugin/checkhealth.vim
+++ b/runtime/ftplugin/checkhealth.vim
@@ -8,11 +8,11 @@ endif
 
 runtime! ftplugin/help.vim
 
-setlocal wrap breakindent linebreak
+setlocal wrap breakindent linebreak nolist
 let &l:iskeyword='!-~,^*,^|,^",192-255'
 
 if exists("b:undo_ftplugin")
-  let b:undo_ftplugin .= "|setl wrap< bri< lbr< kp< isk<"
+  let b:undo_ftplugin .= "|setl wrap< bri< lbr< kp< isk< list<"
 else
-  let b:undo_ftplugin = "setl wrap< bri< lbr< kp< isk<"
+  let b:undo_ftplugin = "setl wrap< bri< lbr< kp< isk< list<"
 endif


runtime/ftplugin/checkhealth.vim
```diff
@@ -8,7 +8,7 @@ endif
 
 runtime! ftplugin/help.vim
 
-setlocal wrap breakindent linebreak
```
clason, line 11: Needs to be added to the `undo_ftplugin` dance below

runtime/ftplugin/checkhealth.vim
```diff
@@ -8,7 +8,7 @@ endif
...
+setlocal wrap breakindent linebreak nolist
```
clason, line 11: Needs to be added to the `undo_ftplugin` dance below

yochem: Sorry, of course. Added `list<` now to both.

PR #31253, opened 2024-11-18
ci: match files instead of directories

ofseed:
The pattern match (which tries to match directories directly) in the `hashFiles()` function is incorrect, see https://github.com/actions/toolkit/blob/bb2278e5cfbb40afc20890c415e9ffa836631cd5/packages/glob/src/internal-hash-files.ts#L28

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -28,6 +28,6 @@ runs:
     - uses: actions/cache@v4
       with:
         path: .deps
-        key: ${{ env.CACHE_KEY }}-${{ steps.image.outputs.version }}-${{ hashFiles('cmake**',
-          '.github/**', 'CMakeLists.txt',
+        key: ${{ env.CACHE_KEY }}-${{ steps.image.outputs.version }}-${{ hashFiles('cmake*/**/*',
+          '.github/**/*', 'CMakeLists.txt',
           'runtime/CMakeLists.txt', 'src/nvim/**/CMakeLists.txt') }}
--- a/scripts/lintcommit.lua
+++ b/scripts/lintcommit.lua
@@ -102,6 +102,10 @@ local function validate_commit(commit_message)
       return [[Scope can't be empty]]
     end
 
+    if scope == 'ci' then
+      return [[Scope can't be "ci", use it as a type instead]]
+    end
+
     if vim.startswith(scope, 'nvim_') then
       return [[Scope should be "api" instead of "nvim_..."]]
     end


clason:
Nit: type should be `ci`, not  `fix` (distinction is important for release note generation).

dundargoc:
I don't really understand the problem. I quickly checked the linked code but I don't k ow TS that well. Does the caching not work currently?

justinmk:
This looks right to me. node projects tend to use https://github.com/isaacs/node-glob which wants `/**/` to recurse. Whereas `cmake**` won't do what we expect, I think.

ofseed:
> Nit: type should be `ci`, not `fix` (distinction is important for release note generation).

Corrected it and added it as a commit lint rule.

> this GHA isn't using that though

It should use [`@actions/glob`](https://www.npmjs.com/package/@actions/glob#patterns).

ofseed:
> I don't really understand the problem. I quickly checked the linked code but I don't know TS that well. Does the caching not work currently?

I noticed that there might be this issue in https://github.com/neovim/neovim/pull/30861#issuecomment-2423749315, where the version of libuv was changed, but my PR (including subsequent PRs during that time) has the same error as https://github.com/neovim/neovim/pull/30861.

dundargoc:
You could be right, but I think you're assuming too much. The only way to know for sure is it doesn't work is to check whether the hash number changes when a relevant file changes. I can check later to make sure.

Regardless, perhaps this is better/more readable even if the current syntax works?

clason:
Even if they're valid, they might behave differently.

ofseed:
My point is that regardless of whether `cmake**` and `.github/**` are valid, we should not rely on this usage as long as it is not documented. And I do *assume* they are invalid (by reading the source code, not by practice).

dundargoc:
I confirmed caching works currently, whatever failure you saw was unrelated to that.

> And I do assume they are invalid (by reading the source code, not by practice).

lmao, next time I suggest you just test it out instead of failing to read typescript code.

Anyway, I don't see any harm in this and the new glob syntax is probably easier to read, so let's go with that.

dundargoc (approved):

ofseed:
Sorry to trouble you with confirmation, and thank you for your suggestion. Next time, I will test it out myself. I rushed to create this PR directly because I didn't see such syntax in the documentation.

justinmk:
Forgot about this, is it not needed ?

ofseed:
> Forgot about this, is it not needed ?

Yes, this PR was created when I tried to solve a CI issue, but after testing by @dundargoc, it was found that it did not solve any issue, although it also did no harm.

PR #31268, opened 2024-11-19
clipboard: remove !empty($TMUX)

Freed-Wu:
$TMUX is unnecessary:

tmux new -d
echo foobar | tmux load-buffer -
tmux save-buffer -

It is like lemonade:

lemonade server &
lemonade copy foobar
lemonade paste

### DIFF
--- a/runtime/autoload/provider/clipboard.vim
+++ b/runtime/autoload/provider/clipboard.vim
@@ -158,7 +158,7 @@ function! provider#clipboard#Executable() abort
     let s:copy['*'] = s:copy['+']
     let s:paste['*'] = s:paste['+']
     return 'termux-clipboard'
-  elseif !empty($TMUX) && executable('tmux')
+  elseif executable('tmux') && (!empty($TMUX) || 0 == jobwait([jobstart(['tmux', 'list-buffers'])], 2000)[0])
     let tmux_v = v:lua.vim.version.parse(system(['tmux', '-V']))
     if !empty(tmux_v) && !v:lua.vim.version.lt(tmux_v, [3,2,0])
       let s:copy['+'] = ['tmux', 'load-buffer', '-w', '-']


gpanders:
I'm not sure about this. Using tmux outside of an active tmux session seems like an uncommon use case. And it introduces the possibility of Nvim selecting a non-functional clipboard provider: just because someone has `tmux` installed doesn't mean there is an active session running (in fact I'd wager that in the vast majority of cases, if `$TMUX` is undefined then there _isn't_ a tmux server running).

Freed-Wu:
TMUX is not defined doesn't mean tmux is not run in a daemon. In fact, lemonade also need user to run `lemonade server &` before `lemonade copy` and `lemonade paste`. We don't have any reason to treat tmux specially. run a tmux session in a daemon, then use its clipboard in all neovim session it useful in a ssh environment, esepecially if user have run a tmux in the machine which run ssh, they maybe don't want to run a tmux in the machine which be `ssh`ed to face two tmux to avoid confusion, but run a tmux sessin in a daemon to enjoy its clipboard.

However, I also think it is not a best solution. We can run `tmux list-buffer` before we set tmux clipboard. If it return true, means `tmux save-buffer -` and `tmux load-buffer -` can work.

gpanders:
> We don't have any reason to treat tmux specially.

We do though, because unlike lemonade the dominant use case of tmux is that the user is "inside" the tmux session where `$TMUX` is defined.

> We can run tmux list-buffer before we set tmux clipboard. If it return true, means tmux save-buffer - and tmux load-buffer - can work.

I agree this would be better. Hopefully it can be done without making the code too complicated.

justinmk:
Please provide a clear (but non-verbose) Problem/Solution summary in the commit message.

"oldtest" CI jobs are failing:

```
Caught exception in Test_filter_display(): Vim(display):clipboard: provider returned invalid data @ command line..script /home/runner/work/neovim/neovim/test/old/testdir/runtest.vim[577]..function RunTheTest[55]..Test_filter_display, line 7
```

runtime/autoload/provider/clipboard.vim
```diff
@@ -158,16 +158,21 @@ function! provider#clipboard#Executable() abort
...
-    let tmux_v = v:lua.vim.version.parse(system(['tmux', '-V']))
-    if !empty(tmux_v) && !v:lua.vim.version.lt(tmux_v, [3,2,0])
-      let s:copy['+'] = ['tmux', 'load-buffer', '-w', '-']
-    else
-      let s:copy['+'] = ['tmux', 'load-buffer', '-']
+  elseif executable('tmux')
+    if empty($TMUX)
+      call system(['tmux', 'list-buffers'])
+    endif
+    if !empty($TMUX) || v:shell_error == 0
```
justinmk, line 165: we can make this a bit clearer by using jobwait+jobstart:
```suggestion
    let tmux = !empty($TMUX) || 0 == jobwait([jobstart(['tmux', 'list-buffers'])], 1000)[0]
    if tmux
```

gou4shi1:
This PR breaks me:
I's using neovim over SSH and I have tmux buffers running for background tasks.
My kitty support OSC52, but this PR makes neovim use tmux which can not copy&paste over SSH.

justinmk:
@gou4shi1 Nvim 0.11.1 (unreleased) and 0.12 (unreleased), you can force osc52 via:

    let g:clipboard = 'osc52'

In older Nvim, you can force it the "long way" by setting `g:clipboard` to a dict.

Autodetection can't magically do what everyone wants. In your case, Nvim detected tmux correctly, but you prefer osc52, so overriding is the correct approach.

gou4shi1:
> let g:clipboard = 'osc52'

Thanks for you reply, I fixed my issue with

```
if vim.env.SSH_TTY ~= nil then
  vim.g.clipboard = "osc52"
end
```

Issue #21150, opened 2022-11-21
vim.ui_attach: document the effect of return value of callback


zeertzjq:
In the current implementation, if `vim.ui_attach()` callback returns `true`, the event is not passed to UI, otherwise it is still passed to UI. For `ext_popupmenu` this difference is unnoticeable in TUI because TUI doesn't not support `ext_popupmenu`, but it is noticeable in a UI that supports `ext_popupmenu`. Anyway, this behavior is neither documented nor tested.

PR #31269, opened 2024-11-19
feat(ui): use builtin completion popupmenu with ext_cmdline

luukvbaal:
Problem:  UIs implementing ext_cmdline/message must also implement
          ext_popupmenu in order to get cmdline completion with
          wildoptions+=pum.
Solution: Allow marking a window as the ext_cmdline window through
          nvim_open_win(), including prompt offset. Anchor the cmdline-
          completion popupmenu to this window.

### DIFF
--- a/src/nvim/api/win_config.c
+++ b/src/nvim/api/win_config.c
@@ -124,8 +124,7 @@
 ///       - `row=0` and `col=0` if `anchor` is "SW" or "SE"
 ///         (thus like a tooltip near the buffer text).
 ///   - row: Row position in units of "screen cell height", may be fractional.
-///   - col: Column position in units of "screen cell width", may be
-///            fractional.
+///   - col: Column position in units of screen cell width, may be fractional.
 ///   - focusable: Enable focus by user actions (wincmds, mouse events).
 ///       Defaults to true. Non-focusable windows can be entered by
 ///       |nvim_set_current_win()|, or, when the `mouse` field is set to true,
@@ -205,6 +204,8 @@
 ///           focused on it.
 ///   - vertical: Split vertically |:vertical|.
 ///   - split: Split direction: "left", "right", "above", "below".
+///   - _cmdline_offset: (EXPERIMENTAL) When provided, anchor the |cmdline-completion|
+///     popupmenu to this window, with an offset in screen cell width.
 ///
 /// @param[out] err Error details, if any
 ///
@@ -286,6 +287,10 @@ Window nvim_open_win(Buffer buffer, Boolean enter, Dict(win_config) *config, Err
       api_set_error(err, kErrorTypeException, "Failed to create window");
     }
     goto cleanup;
+  }
+
+  if (fconfig._cmdline_offset < INT_MAX) {
+    cmdline_win = wp;
   }
 
   // Autocommands may close `wp` or move it to another tabpage, so update and check `tp` after each
@@ -636,6 +641,11 @@ restore_curwin:
       didset_window_options(win, true);
     }
   }
+  if (fconfig._cmdline_offset < INT_MAX) {
+    cmdline_win = win;
+  } else if (win == cmdline_win && fconfig._cmdline_offset == INT_MAX) {
+    cmdline_win = NULL;
+  }
 #undef HAS_KEY_X
 }
 
@@ -768,6 +778,9 @@ Dict(win_config) nvim_win_get_config(Window window, Arena *arena, Error *err)
   const char *rel = (wp->w_floating && !config->external
                      ? float_relative_str[config->relative] : "");
   PUT_KEY_X(rv, relative, cstr_as_string(rel));
+  if (config->_cmdline_offset < INT_MAX) {
+    PUT_KEY_X(rv, _cmdline_offset, config->_cmdline_offset);
+  }
 
   return rv;
 }
@@ -1316,6 +1329,10 @@ static bool parse_win_config(win_T *wp, Dict(win_config) *config, WinConfig *fco
 
   if (HAS_KEY_X(config, hide)) {
     fconfig->hide = config->hide;
+  }
+
+  if (HAS_KEY_X(config, _cmdline_offset)) {
+    fconfig->_cmdline_offset = (int)config->_cmdline_offset;
   }
 
   return true;
--- a/src/nvim/buffer_defs.h
+++ b/src/nvim/buffer_defs.h
@@ -995,6 +995,7 @@ typedef struct {
   bool noautocmd;
   bool fixed;
   bool hide;
+  int _cmdline_offset;
 } WinConfig;
 
 #define WIN_CONFIG_INIT ((WinConfig){ .height = 0, .width = 0, \
@@ -1008,7 +1009,8 @@ typedef struct {
                                       .style = kWinStyleUnused, \
                                       .noautocmd = false, \
                                       .hide = false, \
-                                      .fixed = false })
+                                      .fixed = false, \
+                                      ._cmdline_offset = INT_MAX })
 
 // Structure to store last cursor position and topline.  Used by check_lnums()
 // and reset_lnums().
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -2999,10 +2999,9 @@ static int cmd_startcol(void)
 int cmd_screencol(int bytepos)
 {
   int m;  // maximum column
-
   int col = cmd_startcol();
   if (KeyTyped) {
-    m = Columns * Rows;
+    m = cmdline_win ? cmdline_win->w_view_width * cmdline_win->w_view_height : Columns * Rows;
     if (m < 0) {        // overflow, Columns or Rows at weird value
       m = MAXCOL;
     }
--- a/test/functional/lua/ui_event_spec.lua
+++ b/test/functional/lua/ui_event_spec.lua
@@ -320,6 +320,101 @@ describe('vim.ui_attach', function()
       },
     })
   end)
+
+  it('ext_cmdline completion popupmenu', function()
+    screen:try_resize(screen._width, 10)
+    screen:add_extra_attr_ids { [100] = { background = Screen.colors.Black } }
+    exec_lua([[
+      vim.o.wildoptions = 'pum'
+      local buf = vim.api.nvim_create_buf(false, true)
+      vim.cmd('call setline(1, range(1, 10))')
+      _G.win = vim.api.nvim_open_win(buf, false, {
+        relative = 'editor',
+        col = 3,
+        row = 3,
+        width = 20,
+        height = 1,
+        style = 'minimal',
+        focusable = false,
+        zindex = 300,
+        _cmdline_offset = 0,
+      })
+      vim.ui_attach(ns, { ext_cmdline = true }, function(event, content, _, firstc)
+        if event == 'cmdline_show' then
+          local prompt = vim.api.nvim_win_get_config(_G.win)._cmdline_offset == 0
+          prompt = (prompt and firstc or 'Excommand:') .. content[1][2]
+          vim.api.nvim_buf_set_lines(buf, -2, -1, false, { prompt })
+          vim.api.nvim_win_set_cursor(_G.win, { 1, #prompt })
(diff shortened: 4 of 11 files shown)


src/nvim/api/win_config.c
```diff
@@ -212,6 +212,10 @@
 ///   - hide: If true the floating window will be hidden.
 ///   - vertical: Split vertically |:vertical|.
 ///   - split: Split direction: "left", "right", "above", "below".
+///   - cmdline: If true, indicate that this window is used for ext_cmdline.
+///     Nvim will anchor the 'wildoptions' popupmenu to this window.
+///   - cmdline_offset: Offset in units of "screen cell width" indicating how much longer
+///     (negative for shorter) the displayed prompt is from the prompt emitted by "cmdline_show".
```
luukvbaal, line 218: Could keep these private for use with the default ext_cmdline (`_cmdline/_cmdline_offset`) but after testing this seems useful enough to expose.

test/functional/lua/ui_event_spec.lua
```diff
@@ -306,6 +306,92 @@ describe('vim.ui_attach', function()
...
+        cmdline_offset = -1, -- not inserting "firstc" below
+      })
+      vim.ui_attach(ns, { ext_cmdline = true }, function(event, ...)
+        if event == 'cmdline_show' then
+          local content = select(1, ...)
+          vim.api.nvim_buf_set_lines(buf, -2, -1, false, {content[1][2]})
+          vim.api.nvim_win_set_cursor(_G.win, {1, #content[1][2]})
+          vim.api.nvim__redraw( { win = _G.win, cursor = true, flush = true })
+        end
+        return true
```
przepompownia, line 336: `:help vim.ui_attach()` doesn't seem to mention the meaning of the return value.

luukvbaal: Yeah, it makes Nvim not send the event to remote UIs. I'm not sure what the reason behind it is. Here is a related issue: #21150.

przepompownia: Indeed (if I understand correctly): the test case uses remote UI and cannot pass without it. By comparison, at TUI it doesn't matter.

luukvbaal: This test does not test the remote UI state. In this case specifically returning `true` is used to hide the ext_ state from the screen:expect output, the screen class is the remote UI.

src/nvim/api/win_config.c
justinmk: If we expose these, they should be options (for reference, the existing 'previewwindow' option makes a similar kind of decision). Something like a new, global 'popupwindow' option (or 'popupopt', 'pmenuopt', ...).

Setting a global option from an unrelated "create"/"open" API is strange.

If it helps unblock this PR, for now let's just not expose these.

src/nvim/ex_getln.c
```diff
@@ -2942,10 +2942,9 @@ static int cmd_startcol(void)
 int cmd_screencol(int bytepos)
 {
   int m;  // maximum column
-
   int col = cmd_startcol();
   if (KeyTyped) {
-    m = Columns * Rows;
+    m = cmdline_win ? cmdline_win->w_width_inner * cmdline_win->w_height_inner : Columns * Rows;
```
justinmk, line 2947: Would it make sense to *always* define `cmdline_win`? The default impl would anchor its dimensions to `Columns * Rows`, but that impl would be contained in the default `cmdline_win` handler, so code never needs to check if `cmdline_win` exists:

```suggestion
    m = cmdline_win->w_width_inner * cmdline_win->w_height_inner;
```

I don't know if this is a net gain, so this is suggestion is also a question :)

src/nvim/api/win_config.c
luukvbaal: Still awaiting @bfredl's review, who originally suggested adding this to `nvim_open_win` :) From matrix:

>> I was also wondering if some API to "set" the cmdline window would be better for some internal work Nvim can do. But perhaps documenting that the implementation should move the cursor there to this end is sufficient?

> I think this can be done, but being future-proof for more "roles" being added later. perhaps a field in nvim_win_set_config/win_open

src/nvim/api/win_config.c
```diff
@@ -212,6 +212,10 @@
 ///   - hide: If true the floating window will be hidden.
 ///   - vertical: Split vertically |:vertical|.
 ///   - split: Split direction: "left", "right", "above", "below".
+///   - cmdline: If true, indicate that this window is used for ext_cmdline.
+///     Nvim will anchor the 'wildoptions' popupmenu to this window.
```
justinmk, line 216: Not clear from this description what the window shows. Is it only for the popup menu? `ext_cmdline` presents the entire cmdline, not only the popup menu.

src/nvim/api/win_config.c
justinmk: `nvim_win_set_config` would fit better, since it is essentially window-local properties, which is very similar to window-local options. Though a global option seems to make more sense than a window-local option, since a global option is easier to "query".

luukvbaal: `nvim_win_set_config` and `nvim_open_win` share the same `opts` dictionary, so adding it to one implies the other. I'm not sure how a global option would work, do you mean that option would hold the window handle?

justinmk: > do you mean that option would hold the window handle?

Yes. That seems natural, no?

src/nvim/api/win_config.c
luukvbaal: What the window contains is entirely up to the cmdline implementation. This just allows the implementation to "mark" the window that is used to display the cmdline as such, to allow Nvim to do some work with it internally.

For now that just means anchoring the completion popupmenu, not sure if anything else could be done. Originally also thought this could be used to redraw and move the cursor to it, but since we can do that with `nvim__redraw` this is no longer necessary.

src/nvim/api/win_config.c
luukvbaal: Yes if it's a global option it would have to be set/unset whenever the window is deleted (which the user could do, at which point the handler would have to create a new window). So unless we disallow deleting a window marked as such, attaching the flag to the window itself seems to make more sense to me.

luukvbaal: There is something to say for symmetry with `previewwindow` though, so should probably promote it to an actual window option if @bfredl agrees. This does bring into question the entire `WinConfig` struct.

justinmk: > This does bring into question the entire `WinConfig` struct.

Let's save that for https://github.com/neovim/neovim/issues/30453 , out of scope here.

bfredl: I'd say keep it a `_` field in winconfig for now.

src/nvim/ex_getln.c
luukvbaal: Don't think so; it would have to be a grid not a window (only the default grid relates to `Columns/Rows`). But it's not straightforward to get the inner dimensions from the grid, and we can't query the cmdline row/column from the default grid, so would need ternaries anyways.

src/nvim/buffer_defs.h
```diff
@@ -958,6 +958,11 @@ typedef struct {
   bool noautocmd;
   bool fixed;
   bool hide;
+/// cmdline_offset: Offset in units of "screen cell width" indicating how much
+/// longer (negative for shorter) the displayed prompt is from the prompt
+/// emitted by "cmdline_show". Used to position the |cmdline-completion| popup
+/// window with ext_cmdline.
+  int _cmdline_offset;
```
justinmk, line 965: comment indent looks wrong

test/functional/lua/ui_event_spec.lua
```diff
@@ -306,6 +306,90 @@ describe('vim.ui_attach', function()
...
+      _G.win = vim.api.nvim_open_win(buf, false, {
+        relative = 'editor',
+        col = 3,
+        row = 3,
+        width = 20,
+        height = 1,
+        style = 'minimal',
+        focusable = false,
+        zindex = 300,
+        _cmdline_offset = -1, -- not inserting "firstc" below
```
justinmk, line 326: do we need a test for some other offset? maybe using `nvim_win_set_config`?

src/nvim/buffer_defs.h
```diff
@@ -958,6 +958,11 @@ typedef struct {
   bool noautocmd;
   bool fixed;
   bool hide;
+/// cmdline_offset: Offset in units of "screen cell width" indicating how much
+/// longer (negative for shorter) the displayed prompt is from the prompt
+/// emitted by "cmdline_show". Used to position the |cmdline-completion| popup
+/// window with ext_cmdline.
```
justinmk, line 964: "Setting this activates ..."

src/nvim/buffer_defs.h
```diff
@@ -958,6 +958,11 @@ typedef struct {
   bool noautocmd;
   bool fixed;
   bool hide;
+/// cmdline_offset: Offset in units of "screen cell width" indicating how much
```
bfredl, line 961: no need for the quotes, cell width is a well known concept used everywhere already.

luukvbaal: Sure, but IIRC I copied this verbatim from nvim_open_win.

test/functional/ui/cmdline_spec.lua
```diff
@@ -623,8 +623,7 @@ local function test_cmdline(linegrid)
       { 'unplace', '', '', '' },
     }
 
-    command('set wildmode=full')
-    command('set wildmenu')
+    command('set wildmode=full wildmenu wildoptions=pum')
```
bfredl, line 626: Is this just a clean up of the tests to make them more explicit, or is this necessary due to a change of behavior?

luukvbaal: Includes a behavior change. In this PR ext_popupmenu doesn't (I thought it shouldn't) receive events unless the user sets wildoptions=pum. ext_wildmenu still does regardless of the option.

bfredl: ext_popupmenu also includes wildmenu (regardless of options):

> - `ext_popupmenu`	Externalize |popupmenu-completion| and
>			'wildmenu'. |ui-popupmenu|

Adding a separate `ext_wildmenu` format  was a mistake and this option is now deprecated.

luukvbaal: Yes but if the user indicates they don't want a popupmenu for the cmdline through wildoptions, why would we still emit the event?

src/nvim/api/win_config.c
```diff
@@ -295,6 +295,10 @@ Window nvim_open_win(Buffer buffer, Boolean enter, Dict(win_config) *config, Err
     goto cleanup;
   }
 
+  if (fconfig._cmdline_offset < INT_MAX) {
```
justinmk, line 298: In the `nvim_open_win` we should mention the flag and mention "EXPERIMENTAL". We don't have a better way of documenting experiemental stuff currently. But not documenting it at all means it could be forgotten and/or lead to confusion in the future.

test/functional/ui/cmdline_spec.lua
justinmk: Behavior changes LGTM, ok w/ @bfredl ?

przepompownia:
I started using this and haven't noticed any special issues for now.

przepompownia:
Resolving the conflict would make local merging easier.

test/functional/ui/cmdline_spec.lua
bfredl: No, this should not be changed. what `ext_popupmenu` actually means is forward **the data** related to multiple-choice completion. `ext_popupmenu` does emphatically not refer to the one particular visual style of a menu (which TUI vim has called "the popupmenu" in big singularis) Thus whether it is used is irrelevant of `wildoptions+=pum` is set or not, which is a _newer option_ meant to control legacy TUI only.

przepompownia:
@luukvbaal let me not guess how to resolve (at local merge) the conflict with recent multigrid info changes :grin:

PR #31272, opened 2024-11-19
fix(messages): no message kind for search pattern

### LINKED ISSUE #31244: ui-messages: add kind for search pattern messages
### Problem

I am reporting this issue on`:help ui-messages` suggestion, but in practice I found that empty `kind` can cause problems with distinguishing message types. #31205 significantly reduces the issue for me, but it would be good to have named kind for such messages. I don't know how many other cases we'll get an empty kind.

`empty-kind.lua`:
```lua
local api = vim.api

local function log(msg)
  vim.schedule(function ()
    vim.cmd.normal 'G'
  end)
end

api.nvim_create_autocmd('CmdlineEnter', {
  callback = function ()
    vim.ui_detach(ns)
    vim.schedule(function ()
      api.nvim__redraw({statusline = true})
    end)
  end
})

api.nvim_create_autocmd({'UIEnter', 'CmdlineLeave'}, {
  callback = function ()
    vim.ui_attach(ns, {ext_messages = true, ext_cmdline = false}, function (event, kind, content, _)
      if event == 'msg_show' and kind == '' then
        log(vim.inspect(content))
      end
    end)
  end
})
```
- `nvim --clean -u empty-kind.lua empty-kind.lua`
- press `n` and see appended lines

Related to https://github.com/neovim/neovim/issues/27811

@luukvbaal maybe in the wave of your other UI fixes it will be easier to find?

### Expected behavior

Not empty kind for such messages

przepompownia:
Closes #31244

### DIFF
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -800,6 +800,7 @@ must handle.
 		"rpc_error"	Error response from |rpcrequest()|
 		"return_prompt"	|press-enter| prompt after a multiple messages
 		"quickfix"	Quickfix navigation message
+		"search_echo"	Search echo message
 		"search_count"	Search count message ("S" flag of 'shortmess')
 		"wmsg"		Warning ("search hit BOTTOM", |W10|, …)
 	    New kinds may be added in the future; clients should treat unknown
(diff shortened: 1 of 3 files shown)


runtime/doc/ui.txt
```diff
@@ -800,6 +800,7 @@ must handle.
 		"rpc_error"	Error response from |rpcrequest()|
 		"return_prompt"	|press-enter| prompt after a multiple messages
 		"quickfix"	Quickfix navigation message
+		"search_echo"	Search echo message
```
luukvbaal, line 803: "search_cmd" seems more accurate, `SEARCH_ECHO` is just the name of the internal flag.
```suggestion
		"search_cmd"	Entered search command
```

PR #31279, opened 2024-11-20
fix(messages): more ext_messages kinds

### LINKED ISSUE #31248: ui-messages: add kind for `:hi` output messages
### Problem

Similarly to #31244: the `:highlight` command output event has empty kind in `ui_attach` callback.

### Expected behavior

Allow to recognize such output by a special kind or if it makes sense, use an existing kind.

luukvbaal:
Add kinds for various commands that output a list, the 'wildmode'
    list, and  number prompts.

Close #31248

### DIFF
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -795,13 +795,16 @@ must handle.
 		"echo"		|:echo| message
 		"echomsg"	|:echomsg| message
 		"echoerr"	|:echoerr| message
+		"list_cmd"	List output for various commands (|:ls|, |:set|, …)
 		"lua_error"	Error in |:lua| code
 		"lua_print"	|print()| from |:lua| code
 		"rpc_error"	Error response from |rpcrequest()|
+		"number_prompt"	Number input prompt (|inputlist()|, |z=|, …)
 		"return_prompt"	|press-enter| prompt after a multiple messages
 		"quickfix"	Quickfix navigation message
 		"search_cmd"	Entered search command
 		"search_count"	Search count message ("S" flag of 'shortmess')
+		"wildlist"	'wildmode' "list" message
 		"wmsg"		Warning ("search hit BOTTOM", |W10|, …)
 	    New kinds may be added in the future; clients should treat unknown
 	    kinds as the empty kind.
--- a/test/functional/ui/messages_spec.lua
+++ b/test/functional/ui/messages_spec.lua
@@ -64,7 +64,7 @@ describe('ui/ext_messages', function()
     }
   end)
 
-  it('msg_show kind=confirm,confirm_sub,emsg,wmsg,quickfix', function()
+  it('msg_show kinds', function()
     feed('iline 1\nline 2<esc>')
 
     -- kind=confirm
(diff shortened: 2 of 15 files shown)


zeertzjq:
Tabs and spaces are used inconsistently in docs

runtime/doc/ui.txt
```diff
@@ -789,19 +789,28 @@ must handle.
...
 		"" (empty)	Unknown (consider a feature-request: |bugs|)
+		"buflist"	|:buffers| list
 		"confirm"	|confirm()| or |:confirm| dialog
 		"confirm_sub"	|:substitute| confirm dialog |:s_c|
 		"emsg"		Error (|errors|, internal error, |:throw|, …)
 		"echo"		|:echo| message
 		"echomsg"	|:echomsg| message
 		"echoerr"	|:echoerr| message
+		"highlight"	|:highlight| message
+		"inputlist"	|inputlist()| message
```
justinmk, line 800: Maybe a `_cmd` suffix is a useful convention (like `search_cmd`).

```suggestion
		"highlight_cmd"	|:highlight| message
		"inputlist"	|inputlist()| message
```

runtime/doc/ui.txt
```diff
@@ -789,19 +789,28 @@ must handle.
...
 		"lua_print"	|print()| from |:lua| code
+		"map"		|:map| mapping list
 		"rpc_error"	Error response from |rpcrequest()|
+		"number_prompt"	Number input prompt (|inputlist()|, |z=|, …)
 		"return_prompt"	|press-enter| prompt after a multiple messages
 		"quickfix"	Quickfix navigation message
 		"search_cmd"	Entered search command
 		"search_count"	Search count message ("S" flag of 'shortmess')
+		"set"		|:set| option list
+		"spellsugest"	Spell suggestion message |z=|
```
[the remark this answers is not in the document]
luukvbaal: How ironic😅

test/functional/ui/messages_spec.lua
```diff
@@ -64,7 +64,7 @@ describe('ui/ext_messages', function()
     }
   end)
 
-  it('msg_show kind=confirm,confirm_sub,emsg,wmsg,quickfix', function()
+  it('msg_show different kinds', function()
```
justinmk, line 67:
```suggestion
  it('msg_show kinds', function()
```

justinmk (approved):

runtime/doc/ui.txt
luukvbaal: > Not clear to me why all these things need their own special kind

Yeah they really don't. If we can come up with an appropriate umbrella kind for some of these that would be fine too. The only new kind that actually fixes an issue here is `number_prompt`; which needs to be non-fast.

luukvbaal: I think a `list_cmd` might suffice for buflist, highlight, let, map, set etc.

justinmk (approved):

przepompownia:
Thank you! I need some time to go through the kinds added here.

runtime/doc/ui.txt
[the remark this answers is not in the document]
przepompownia: Finally I cannot find this change in master and force-pushed commits does not allow reproduce what happened later.

Issue #31270, opened 2024-11-19
Pattern of vim.uri_encode() for RFC2732 has unescaped magic characters


troiganto:
### Problem

For `vim.uri_encode()`, there is the following definition:
```lua
local PATTERNS = {
  ...
  -- RFC 2732
  -- https://tools.ietf.org/html/rfc2732
  rfc2732 = "^A-Za-z0-9%-_.!~*'()[]",
  ...
}
```
Some of the characters here have special meaning within Lua patterns and should be escaped.

Especially, `uri_encode()` is defined as follows:
```lua
function M.uri_encode(str, rfc)
  local pattern = PATTERNS[rfc] or PATTERNS.rfc3986
  return (str:gsub('([' .. pattern .. '])', percent_encode_char)) -- clamped to 1 retval with ()
end
```

For RFC2732, this gives the following pattern:
```lua
return (str:gsub('([^A-Za-z0-9%-_.!~*'()[]])', percent_encode_char))
```

This makes a bit clearer that the pattern now has two unescaped closing brackets. The matcher interprets this as a character group followed by a literal closing bracket.

The solution would be to escape all magic characters in `PATTERNS` with percents.

### Steps to reproduce

The following command:
```
:=vim.uri_encode('[:]', 'rfc2732')
```
produces the output:
```
[%3a
```

### Expected behavior

It should produce the output:
```
[%3a]
```
(Mind the closing bracket.)

### Nvim version (nvim -v)

0.10.2

### Vim (not Nvim) behaves the same?

No, this is a Lua-exclusive bug

### Operating system/version

Fedora Silverblue 41

### Terminal name/version

kitty 0.36.4

### $TERM environment variable

xterm-kitty

### Installation

AppImage

PR #31284, opened 2024-11-21
fix(uri): add escape char for RFC2732 in `uri_encode()` (#31270)

JonnyKong:
**Problem:**
The brackets in the RFC2732 regular expression are currently unescaped, causing them to be misinterpreted as special characters denoting character groups rather than as literal characters. See #31270 for detailed examples.

**Solution:**
Escape the brackets.

### DIFF
--- a/test/functional/lua/uri_spec.lua
+++ b/test/functional/lua/uri_spec.lua
@@ -252,4 +252,12 @@ describe('URI methods', function()
       end
     )
   end)
+
+  describe('encode to uri', function()
+    it('rfc2732 including brackets', function()
+      exec_lua("str = '[:]'")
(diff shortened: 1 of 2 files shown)


test/functional/lua/uri_spec.lua
```diff
@@ -252,4 +252,12 @@ describe('URI methods', function()
       end
     )
   end)
+
+  describe('encode to uri', function()
+    it('rfc2732 including brackets', function()
+      exec_lua("str = '[:]'")
+      exec_lua("rfc = 'rfc2732'")
+      eq('[%3a]', exec_lua('return vim.uri_encode(str, rfc)'))
```
justinmk, line 260: Wow, we really didn't have a test for `uri_encode` :O

PR #31286, opened 2024-11-21
refactor(lsp): rename `offset_encoding` to `position_encoding`

ofseed:
There was a discussion earlier at https://github.com/neovim/neovim/pull/31249#discussion_r1849409391. I am not very supportive of renaming `vim.lsp.Client.offset_encoding` to `vim.lsp.Client.position_encoding`(which would be a breaking change unlike the others), but I have completed it and placed it in a separate commit for easier review and revert.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -904,8 +904,9 @@ Lua module: vim.lsp.client                                        *lsp-client*
       • {rpc}                   (`vim.lsp.rpc.PublicClient`) RPC client
                                 object, for low level interaction with the
                                 client. See |vim.lsp.rpc.start()|.
-      • {offset_encoding}       (`string`) The encoding used for communicating
-                                with the server. You can modify this in the
+      • {offset_encoding}       (`string`) Called "position encoding" in LSP
+                                spec, the encoding used for communicating with
+                                the server. You can modify this in the
                                 `config`'s `on_init` method before text is
                                 sent to the server.
       • {handlers}              (`table<string,lsp.Handler>`) The handlers
@@ -1062,9 +1063,10 @@ Lua module: vim.lsp.client                                        *lsp-client*
       • {get_language_id}?    (`fun(bufnr: integer, filetype: string): string`)
                               Language ID as string. Defaults to the buffer
                               filetype.
-      • {offset_encoding}?    (`'utf-8'|'utf-16'|'utf-32'`) The encoding that
-                              the LSP server expects. Client does not verify
-                              this is correct.
+      • {offset_encoding}?    (`'utf-8'|'utf-16'|'utf-32'`) Called "position
+                              encoding" in LSP spec, the encoding that the LSP
+                              server expects. Client does not verify this is
+                              correct.
       • {on_error}?           (`fun(code: integer, err: string)`) Callback
                               invoked when the client operation throws an
                               error. `code` is a number describing the error.
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1840,7 +1840,7 @@ Lua module: vim.lsp.util                                            *lsp-util*
 
 
                                      *vim.lsp.util.apply_text_document_edit()*
-apply_text_document_edit({text_document_edit}, {index}, {offset_encoding})
+apply_text_document_edit({text_document_edit}, {index}, {position_encoding})
     Applies a `TextDocumentEdit`, which is a list of changes to a single
     document.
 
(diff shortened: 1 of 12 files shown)


MariaSolOs:
Thoughts on deprecating `offset_encoding` before fully removing it?

ofseed:
Considering that `offset_encoding` mostly appears in positional parameters, I think it's more of a documentation change? The real refactoring and breaking change are the fields in `vim.lsp.Client` (or `vim.lsp.CTGroup`, but it is an internal module and can be changed safely). I did not completely remove it but used `vim.deprecate`, and whether this change should be completed is still up for discussion. It is merely a renaming, yet it is a breaking change.

MariaSolOs:
> Considering that `offset_encoding` mostly appears in positional parameters, I think it's more of a documentation change? The real refactoring and breaking change are the fields in `vim.lsp.Client` (or `vim.lsp.CTGroup`, but it is an internal module and can be changed safely). I did not completely remove it but used `vim.deprecate`, and whether this change should be completed is still up for discussion. It is merely a renaming, yet it is a breaking change.

I see. I agree with @justinmk that consistency with the spec terminology is nice, but I want to better understand @mfussenegger's [comment](https://github.com/neovim/neovim/pull/31249#discussion_r1851051599). @mfussenegger: Are you also in favour of this rename?

mfussenegger:
To me it's not really worth it.

If some feel strongly about this I won't object but it would need to bump the deprecation removal to `0.13` instead of `0.12`.
Otherwise users will get a deprecation on nightly once this is merged - and plugins are forced to add a compat layer on every use if they want to stay warning-free with both 0.10 and nightly - that's tedious.

runtime/lua/vim/lsp/sync.lua
```diff
@@ -326,10 +326,10 @@ end
 ---@param lines string[]
 ---@param start_range vim.lsp.sync.Range
 ---@param end_range vim.lsp.sync.Range
----@param offset_encoding string
+---@param position_encoding string
 ---@param line_ending string
 ---@return integer
-local function compute_range_length(lines, start_range, end_range, offset_encoding, line_ending)
+local function compute_range_length(lines, start_range, end_range, position_encoding, line_ending)
```
justinmk, line 332: Let's do non-breaking changes such as this.

runtime/lua/vim/lsp/util.lua
```diff
@@ -904,17 +904,20 @@ end
...
   -- location may be Location or LocationLink
   local uri = location.uri or location.targetUri
   if uri == nil then
     return false
   end
-  if offset_encoding == nil then
-    vim.notify_once('show_document must be called with valid offset encoding', vim.log.levels.WARN)
+  if position_encoding == nil then
+    vim.notify_once(
+      'show_document must be called with valid position encoding',
```
justinmk, line 918: Keep this change also, especially since it's a description and not a literal parameter name. But even here, the parameter name can be renamed because it's non-breaking.

runtime/doc/lsp.txt
```diff
@@ -2138,17 +2139,17 @@ rename({old_fname}, {new_fname}, {opts})               *vim.lsp.util.rename()*
...
     Parameters: ~
-      • {location}         (`lsp.Location|lsp.LocationLink`)
-      • {offset_encoding}  (`'utf-8'|'utf-16'|'utf-32'?`)
-      • {opts}             (`table?`) A table with the following fields:
-                           • {reuse_win}? (`boolean`) Jump to existing window
-                             if buffer is already open.
-                           • {focus}? (`boolean`) Whether to focus/jump to
-                             location if possible. (defaults: true)
+      • {location}           (`lsp.Location|lsp.LocationLink`)
+      • {position_encoding}  (`'utf-8'|'utf-16'|'utf-32'?`)
```
justinmk, line 2147: This is non-breaking. We can keep this change. Similar for the others.

runtime/doc/lsp.txt
```diff
@@ -904,7 +904,7 @@ Lua module: vim.lsp.client                                        *lsp-client*
       • {rpc}                   (`vim.lsp.rpc.PublicClient`) RPC client
                                 object, for low level interaction with the
                                 client. See |vim.lsp.rpc.start()|.
-      • {offset_encoding}       (`string`) The encoding used for communicating
+      • {position_encoding}     (`string`) The encoding used for communicating
```
justinmk, line 907: Since this is breaking, we don't need to make this change. Instead, as a bridge, let's just mention "position encoding" in the docs so that it can be found by searching, and to help users understand that `offset_encoding` is an alias to the spec's "position encoding".
```suggestion
      • {offset_encoding}     (`string`) Called "position encoding" in LSP spec. The encoding used for communicating
```

runtime/lua/vim/lsp/client.lua
```diff
@@ -148,8 +149,10 @@ local validate = vim.validate
 --- See |vim.lsp.rpc.start()|.
 --- @field rpc vim.lsp.rpc.PublicClient
 ---
---- The encoding used for communicating with the server. You can modify this in
---- the `config`'s `on_init` method before text is sent to the server.
+--- Called "position encoding" in LSP spec,
+--- the encoding used for communicating with the server.
+--- You can modify this in the `config`'s `on_init` method
+--- before text is sent to the server.
```
justinmk, line 155: nit: line length limit is 100 chars... why is this crunched up?

justinmk (approved):
LGTM. No breaking changes. So we have 95% aligned with LSP spec, except for one structure where we added a hint to its docs and avoided a breaking change.

mfussenegger:
I kinda feel as if having mixed naming is more confusing than just sticking to the different name. 

Are we sure this is an improvement?

justinmk:
> I kinda feel as if having mixed naming is more confusing than just sticking to the different name.

There is only one place AFAIK where the old name is still used, and its doc mentions "position encoding". That seems pretty limited to me.

PR #31292, opened 2024-11-21
fix(grid): double grid_line_start() with ext_messages

luukvbaal:
Problem:  Hit double grid_line_start() assert when redrawing from
          ext_messages msg_ruler event.
Solution: Do not start() batched grid calls when win_redr_ruler() will not
          puts() anything.

### DIFF
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -1092,9 +1092,13 @@ int showmode(void)
   win_T *ruler_win = curwin->w_status_height == 0 ? curwin : lastwin_nofloating();
   if (redrawing() && ruler_win->w_status_height == 0 && global_stl_height() == 0
       && !(p_ch == 0 && !ui_has(kUIMessages))) {
-    grid_line_start(&msg_grid_adj, Rows - 1);
+    if (!ui_has(kUIMessages)) {
+      grid_line_start(&msg_grid_adj, Rows - 1);
+    }
     win_redr_ruler(ruler_win);
-    grid_line_flush();
+    if (!ui_has(kUIMessages)) {
+      grid_line_flush();
+    }
   }
 
   redraw_cmdline = false;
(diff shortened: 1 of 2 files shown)


src/nvim/drawscreen.c
```diff
@@ -1092,9 +1092,13 @@ int showmode(void)
   win_T *ruler_win = curwin->w_status_height == 0 ? curwin : lastwin_nofloating();
   if (redrawing() && ruler_win->w_status_height == 0 && global_stl_height() == 0
       && !(p_ch == 0 && !ui_has(kUIMessages))) {
-    grid_line_start(&msg_grid_adj, Rows - 1);
+    if (!ui_has(kUIMessages)) {
+      grid_line_start(&msg_grid_adj, Rows - 1);
```
luukvbaal, line 1096: Alternative smaller diff that achieves the same thing at the cost of still calling start() and flush() here:
```diff
@@ -588,6 +588,7 @@ void win_redr_ruler(win_T *wp)
     ADD_C(chunk, INTEGER_OBJ(HLF_MSG));
     assert(attr == HL_ATTR(HLF_MSG));
     ADD_C(content, ARRAY_OBJ(chunk));
+    grid_line_flush();
     ui_call_msg_ruler(content);
     did_show_ext_ruler = true;
   } else {
```

test/functional/ui/messages_spec.lua
```diff
@@ -1159,6 +1159,21 @@ stack traceback:
     exec_lua([[vim.print({ foo = "bar" })]])
     screen:expect_unchanged()
   end)
+
+  it('ruler redraw does not crash due to double grid_line_start()', function()
+    exec_lua([[
+      local ns = vim.api.nvim_create_namespace('')
+      vim.ui_attach(ns, { ext_messages = true }, function(event, ...)
+        if event == 'msg_ruler' then
+          vim.api.nvim__redraw( { flush = true })
```
zeertzjq, line 1168:
```suggestion
          vim.api.nvim__redraw({ flush = true })
```

justinmk:
cc @fredizzimo

PR #31302, opened 2024-11-22
fix(api): don't try to get/set option for invalid option name

famiu:
Problem: `validate_option_value_args()` returns `OK` even if option name is invalid or if option doesn't have the supported scope, which leads to Neovim still trying to erroneously get/set the option in those cases, which can lead to an assertion failure when `option_has_scope()` is invoked. This issue miraculously doesn't exist in release builds since the assertion is skipped and `(get/set)_option_value_for` returns if there is an error set, but that is not the intended location for that error to be caught.

Solution: Make `validate_option_value_args()` return `FAIL` if there is an error set, which causes the API option functions to return early instead of trying to get/set an invalid option.

### DIFF
--- a/test/functional/api/vim_spec.lua
+++ b/test/functional/api/vim_spec.lua
@@ -1770,6 +1770,11 @@ describe('API', function()
     end)
 
     it('validation', function()
+      eq("Unknown option 'foobar'", pcall_err(api.nvim_set_option_value, 'foobar', 'baz', {}))
+      eq(
+        "Unknown option 'foobar'",
+        pcall_err(api.nvim_set_option_value, 'foobar', 'baz', { win = api.nvim_get_current_win() })
+      )
       eq(
         "Invalid 'scope': expected 'local' or 'global'",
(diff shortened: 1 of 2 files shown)


zeertzjq:
> I'm not sure if I should try to add test coverage for this, or if it's even possible to add test coverage since this only crashes in Debug builds.

Assertions are also enabled in RelWithDebInfo builds.

test/functional/api/vim_spec.lua
```diff
@@ -1770,6 +1770,11 @@ describe('API', function()
     end)
 
     it('validation', function()
+      eq("Unknown option 'foobar'", pcall_err(api.nvim_get_option_value, 'foobar', {}))
+      eq(
+        "Unknown option 'foobar'",
+        pcall_err(api.nvim_get_option_value, 'foobar', { win = api.nvim_get_current_win() })
+      )
```
zeertzjq, line 1777: The assertion is only triggered for `nvim_set_option_value`, not for `nvim_get_option_value`.

zeertzjq (approved):

PR #31311, opened 2024-11-22
feat(lsp): support `textDocument/foldingRange`

### LINKED ISSUE #12836: Support LSP folding
Some Language Servers support textDocument/foldingRange

If the capability is present, Neovim should support folding provided by the LSP.

Here is a starter how it's done in Vim (according to my understanding): https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/ui/vim/folding.vim

ofseed:
Closes https://github.com/neovim/neovim/issues/12836.

Implemented the [`textDocument/foldingRange`](https://microsoft.github.io/language-server-protocol/specification/#textDocument_foldingRange) of LSP.

* Provides `vim.lsp.foldexpr()` which can be used to enable LSP folding.
* Provides `vim.lsp.foldtext()` which can be used to display the `collapsedText` returned by LSP servers.
* Provides `vim.lsp.foldclose(kind, winid)`, which can be used to fold a certain kind, and it can be used to fold imports automatically.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -204,6 +204,7 @@ won't run if your server doesn't support them.
 - `'textDocument/diagnostic'`
 - `'textDocument/documentHighlight'`
 - `'textDocument/documentSymbol'`
+- `'textDocument/foldingRange'`
 - `'textDocument/formatting'`
 - `'textDocument/hover'`
 - `'textDocument/implementation'`
@@ -696,6 +697,13 @@ commands                                                    *vim.lsp.commands*
     String command: String arguments?: any[]
 
     The second argument is the `ctx` of |lsp-handler|
+
+foldexpr({lnum})                                          *vim.lsp.foldexpr()*
+    Provides an interface between the built-in client and a `foldexpr`
+    function.
+
+    Parameters: ~
+      • {lnum}  (`integer`) line number
 
 formatexpr({opts})                                      *vim.lsp.formatexpr()*
     Provides an interface between the built-in client and a `formatexpr`
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1662,8 +1662,19 @@ Lua module: vim.lsp.folding_range                          *lsp-folding_range*
 foldclose({kind}, {winid})                 *vim.lsp.folding_range.foldclose()*
     Close all {kind} of folds in the the window with {winid}.
 
+    To automatically fold imports when opening a file, you can use an autocmd: >lua
+        vim.api.nvim_create_autocmd('LspNotify', {
+          callback = function(args)
+            if args.data.method == 'textDocument/didOpen' then
+              vim.lsp.folding_range.foldclose('imports', vim.fn.bufwinid(args.buf))
+            end
+          end,
+        })
+<
+
     Parameters: ~
-      • {kind}   (`lsp.FoldingRangeKind`) 'comment' | 'imports' | 'region'
+      • {kind}   (`lsp.FoldingRangeKind`) Kind to close, one of "comment",
+                 "imports" or "region".
       • {winid}  (`integer?`) Defaults to the current window.
 
 
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -705,6 +705,10 @@ foldexpr({lnum})                                          *vim.lsp.foldexpr()*
     Parameters: ~
       • {lnum}  (`integer`) line number
 
+foldtext()                                                *vim.lsp.foldtext()*
+    Provides a `foldtext` function that shows the `collapsedText` retrieved,
+    defaults to the first folded line if `collapsedText` is not provided.
+
 formatexpr({opts})                                      *vim.lsp.formatexpr()*
     Provides an interface between the built-in client and a `formatexpr`
     function.
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1657,6 +1657,17 @@ trigger()                                       *vim.lsp.completion.trigger()*
 
 
 ==============================================================================
+Lua module: vim.lsp.folding_range                          *lsp-folding_range*
+
+foldclose({kind}, {winid})                 *vim.lsp.folding_range.foldclose()*
+    Close all {kind} of folds in the the window with {winid}.
+
+    Parameters: ~
+      • {kind}   (`lsp.FoldingRangeKind`) 'comment' | 'imports' | 'region'
+      • {winid}  (`integer?`) Defaults to the current window.
+
+
+==============================================================================
 Lua module: vim.lsp.inlay_hint                                *lsp-inlay_hint*
 
 enable({enable}, {filter})                       *vim.lsp.inlay_hint.enable()*
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -698,6 +698,24 @@ commands                                                    *vim.lsp.commands*
 
     The second argument is the `ctx` of |lsp-handler|
 
+foldclose({kind}, {winid})                               *vim.lsp.foldclose()*
+    Close all {kind} of folds in the the window with {winid}.
+
+    To automatically fold imports when opening a file, you can use an autocmd: >lua
+        vim.api.nvim_create_autocmd('LspNotify', {
+          callback = function(args)
+            if args.data.method == 'textDocument/didOpen' then
+              vim.lsp.foldclose('imports', vim.fn.bufwinid(args.buf))
+            end
+          end,
+        })
+<
+
+    Parameters: ~
+      • {kind}   (`lsp.FoldingRangeKind`) Kind to close, one of "comment",
+                 "imports" or "region".
+      • {winid}  (`integer?`) Defaults to the current window.
+
 foldexpr({lnum})                                          *vim.lsp.foldexpr()*
     Provides an interface between the built-in client and a `foldexpr`
     function.
@@ -1658,28 +1676,6 @@ enable({enable}, {client_id}, {bufnr}, {opts})
 
 trigger()                                       *vim.lsp.completion.trigger()*
     Trigger LSP completion in the current buffer.
-
-
-==============================================================================
-Lua module: vim.lsp.folding_range                          *lsp-folding_range*
-
-foldclose({kind}, {winid})                 *vim.lsp.folding_range.foldclose()*
-    Close all {kind} of folds in the the window with {winid}.
-
-    To automatically fold imports when opening a file, you can use an autocmd: >lua
-        vim.api.nvim_create_autocmd('LspNotify', {
-          callback = function(args)
-            if args.data.method == 'textDocument/didOpen' then
-              vim.lsp.folding_range.foldclose('imports', vim.fn.bufwinid(args.buf))
-            end
-          end,
-        })
-<
-
-    Parameters: ~
-      • {kind}   (`lsp.FoldingRangeKind`) Kind to close, one of "comment",
-                 "imports" or "region".
-      • {winid}  (`integer?`) Defaults to the current window.
 
 
 ==============================================================================
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -230,6 +230,8 @@ LSP
 • |vim.lsp.buf.hover()| now highlights hover ranges using the
   |hl-LspReferenceTarget| highlight group.
 • Functions in |vim.lsp.Client| can now be called as methods.
+• Implemented LSP folding: |vim.lsp.foldexpr()|
+  https://microsoft.github.io/language-server-protocol/specification/#textDocument_foldingRange
 
 LUA
 
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -11,6 +11,7 @@ local lsp = vim._defer_require('vim.lsp', {
   codelens = ..., --- @module 'vim.lsp.codelens'
   completion = ..., --- @module 'vim.lsp.completion'
   diagnostic = ..., --- @module 'vim.lsp.diagnostic'
+  folding_range = ..., --- @module 'vim.lsp.folding_range'
   handlers = ..., --- @module 'vim.lsp.handlers'
   inlay_hint = ..., --- @module 'vim.lsp.inlay_hint'
   log = ..., --- @module 'vim.lsp.log'
@@ -57,6 +58,7 @@ lsp._request_name_to_capability = {
   [ms.textDocument_documentHighlight] = { 'documentHighlightProvider' },
   [ms.textDocument_documentLink] = { 'documentLinkProvider' },
   [ms.textDocument_documentSymbol] = { 'documentSymbolProvider' },
+  [ms.textDocument_foldingRange] = { 'foldingRangeProvider' },
   [ms.textDocument_formatting] = { 'documentFormattingProvider' },
   [ms.textDocument_hover] = { 'hoverProvider' },
   [ms.textDocument_implementation] = { 'implementationProvider' },
@@ -1092,6 +1094,12 @@ end
 ---@return table[] tags A list of matching tags
 function lsp.tagfunc(pattern, flags)
   return vim.lsp._tagfunc(pattern, flags)
+end
+
+--- Provides an interface between the built-in client and a `foldexpr` function.
+---@param lnum integer line number
+function lsp.foldexpr(lnum)
+  return vim.lsp.folding_range._foldexpr(lnum)
 end
 
 ---Checks whether a client is stopped.
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -1102,6 +1102,12 @@ function lsp.foldexpr(lnum)
   return vim.lsp.folding_range._foldexpr(lnum)
 end
 
+--- Provides a `foldtext` function that shows the `collapsedText` retrieved,
+--- defaults to the first folded line if `collapsedText` is not provided.
+function lsp.foldtext()
+  return vim.lsp.folding_range._foldtext()
+end
+
 ---Checks whether a client is stopped.
 ---
 ---@param client_id (integer)
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -3,6 +3,7 @@ local validate = vim.validate
 
 local lsp = vim._defer_require('vim.lsp', {
   _changetracking = ..., --- @module 'vim.lsp._changetracking'
+  _folding_range = ..., --- @module 'vim.lsp._folding_range'
   _snippet_grammar = ..., --- @module 'vim.lsp._snippet_grammar'
   _tagfunc = ..., --- @module 'vim.lsp._tagfunc'
   _watchfiles = ..., --- @module 'vim.lsp._watchfiles'
@@ -11,7 +12,6 @@ local lsp = vim._defer_require('vim.lsp', {
   codelens = ..., --- @module 'vim.lsp.codelens'
   completion = ..., --- @module 'vim.lsp.completion'
   diagnostic = ..., --- @module 'vim.lsp.diagnostic'
-  folding_range = ..., --- @module 'vim.lsp.folding_range'
   handlers = ..., --- @module 'vim.lsp.handlers'
   inlay_hint = ..., --- @module 'vim.lsp.inlay_hint'
   log = ..., --- @module 'vim.lsp.log'
@@ -1099,13 +1099,33 @@ end
 --- Provides an interface between the built-in client and a `foldexpr` function.
 ---@param lnum integer line number
 function lsp.foldexpr(lnum)
-  return vim.lsp.folding_range._foldexpr(lnum)
+  return vim.lsp._folding_range.foldexpr(lnum)
+end
+
+--- Close all {kind} of folds in the the window with {winid}.
+---
+--- To automatically fold imports when opening a file, you can use an autocmd:
+---
+--- ```lua
+--- vim.api.nvim_create_autocmd('LspNotify', {
+---   callback = function(args)
+---     if args.data.method == 'textDocument/didOpen' then
+---       vim.lsp.foldclose('imports', vim.fn.bufwinid(args.buf))
+---     end
+---   end,
+--- })
+--- ```
+---
+---@param kind lsp.FoldingRangeKind Kind to close, one of "comment", "imports" or "region".
+---@param winid? integer Defaults to the current window.
+function lsp.foldclose(kind, winid)
+  return vim.lsp._folding_range.foldclose(kind, winid)
 end
 
 --- Provides a `foldtext` function that shows the `collapsedText` retrieved,
 --- defaults to the first folded line if `collapsedText` is not provided.
 function lsp.foldtext()
-  return vim.lsp.folding_range._foldtext()
+  return vim.lsp._folding_range.foldtext()
 end
 
 ---Checks whether a client is stopped.
--- a/runtime/lua/vim/lsp/_folding_range.lua
+++ b/runtime/lua/vim/lsp/_folding_range.lua
@@ -305,22 +305,8 @@ local function foldclose(kind, winid)
   end)
 end
 
---- Close all {kind} of folds in the the window with {winid}.
----
---- To automatically fold imports when opening a file, you can use an autocmd:
----
---- ```lua
---- vim.api.nvim_create_autocmd('LspNotify', {
----   callback = function(args)
----     if args.data.method == 'textDocument/didOpen' then
----       vim.lsp.folding_range.foldclose('imports', vim.fn.bufwinid(args.buf))
----     end
----   end,
---- })
---- ```
----
----@param kind lsp.FoldingRangeKind Kind to close, one of "comment", "imports" or "region".
----@param winid? integer Defaults to the current window.
+---@param kind lsp.FoldingRangeKind
+---@param winid? integer
 function M.foldclose(kind, winid)
   vim.validate('kind', kind, 'string')
   vim.validate('winid', winid, 'number', true)
@@ -368,7 +354,7 @@ function M.foldclose(kind, winid)
 end
 
 ---@return string
-function M._foldtext()
+function M.foldtext()
   local bufnr = api.nvim_get_current_buf()
   local lnum = vim.v.foldstart
   local row = lnum - 1
@@ -381,7 +367,7 @@ end
 
 ---@param lnum? integer
 ---@return string level
-function M._foldexpr(lnum)
+function M.foldexpr(lnum)
   local bufnr = api.nvim_get_current_buf()
   local bufstate = bufstates[bufnr] or setup(bufnr)
   if not bufstate then
--- a/runtime/lua/vim/lsp/_folding_range.lua
+++ b/runtime/lua/vim/lsp/_folding_range.lua
@@ -107,14 +107,9 @@ local function schedule_foldupdate(bufnr)
   end
 end
 
----@param result lsp.FoldingRange[]?
----@type lsp.Handler
-local function handler(err, result, ctx)
-  if err then
-    log.error(err)
-    return
-  end
-
+---@param results table<integer,{err: lsp.ResponseError?, result: lsp.FoldingRange[]?}>
+---@type lsp.MultiHandler
+local function multi_handler(results, ctx)
   local bufnr = assert(ctx.bufnr)
   -- Handling responses from outdated buffer only causes performance overhead.
   if util.buf_versions[bufnr] ~= ctx.version then
@@ -122,7 +117,13 @@ local function handler(err, result, ctx)
   end
 
   local bufstate = assert(bufstates[bufnr])
-  bufstate.client_ranges[ctx.client_id] = result
+  for client_id, result in pairs(results) do
+    if result.err then
+      log.error(result.err)
+    else
+      bufstate.client_ranges[client_id] = result.result
+    end
+  end
   bufstate.version = ctx.version
 
   if api.nvim_get_mode().mode:match('^i') then
@@ -133,16 +134,30 @@ local function handler(err, result, ctx)
   end
 end
 
+---@param result lsp.FoldingRange[]?
+---@type lsp.Handler
+local function handler(err, result, ctx)
+  multi_handler({ [ctx.client_id] = { err = err, result = result } }, ctx)
+end
+
 --- Request `textDocument/foldingRange` from the server.
 --- `foldupdate()` is scheduled once after the request is completed.
 ---@param bufnr integer
----@param client vim.lsp.Client The client whose server supports `foldingRange`.
----@return integer? request_id
+---@param client? vim.lsp.Client The client whose server supports `foldingRange`.
 local function request(bufnr, client)
   ---@type lsp.FoldingRangeParams
   local params = { textDocument = util.make_text_document_params(bufnr) }
-  local _, request_id = client:request(ms.textDocument_foldingRange, params, handler, bufnr)
-  return request_id
+
+  if client then
+    client:request(ms.textDocument_foldingRange, params, handler, bufnr)
+    return
+  end
+
+  if not next(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange })) then
+    return
+  end
+
+  vim.lsp.buf_request_all(bufnr, ms.textDocument_foldingRange, params, multi_handler)
 end
 
 -- NOTE:
@@ -193,11 +208,7 @@ local function setup(bufnr)
         row_kinds = {},
         row_text = {},
       }
-      for _, client in
-        ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
-      do
-        request(bufnr, client)
-      end
+      request(bufnr)
     end,
     --- Sync changed rows with their previous foldlevels before applying new ones.
     on_bytes = function(_, _, _, start_row, _, _, old_row, _, _, new_row, _, _)
@@ -280,11 +291,7 @@ local function setup(bufnr)
     end,
   })
 
-  for _, client in
-    ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
-  do
-    request(bufnr, client)
-  end
+  request(bufnr)
 
   return bufstates[bufnr]
 end
@@ -324,33 +331,15 @@ function M.foldclose(kind, winid)
   end
   -- Schedule `foldclose()` if the buffer is not up-to-date.
 
-  --- Index in the form of request_id -> true?
-  ---@type table<integer, true?>
-  local scheduled_request = {}
-  for _, client in
-    ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
-  do
-    local request_id = request(bufnr, client)
-    if request_id then
-      scheduled_request[request_id] = true
-    end
+  if not next(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange })) then
+    return
   end
-
-  api.nvim_create_autocmd('LspRequest', {
-    buffer = bufnr,
-    callback = function(args)
-      ---@type integer
-      local request_id = args.data.request_id
-      if scheduled_request[request_id] and args.data.request.type == 'complete' then
-        scheduled_request[request_id] = nil
-      end
-      -- Do `foldclose()` if all the requests is completed.
-      if next(scheduled_request) == nil then
-        foldclose(kind, winid)
-        return true
-      end
-    end,
-  })
+  ---@type lsp.FoldingRangeParams
+  local params = { textDocument = util.make_text_document_params(bufnr) }
+  vim.lsp.buf_request_all(bufnr, ms.textDocument_foldingRange, params, function(...)
+    multi_handler(...)
+    foldclose(kind, winid)
+  end)
 end
 
 ---@return string
--- a/runtime/lua/vim/lsp/folding_range.lua
+++ b/runtime/lua/vim/lsp/folding_range.lua
@@ -120,10 +120,12 @@ end
 --- `foldupdate()` is scheduled once after the request is completed.
 ---@param bufnr integer
 ---@param client vim.lsp.Client The client whose server supports `foldingRange`.
+---@return integer? request_id
 local function request(bufnr, client)
   ---@type lsp.FoldingRangeParams
   local params = { textDocument = util.make_text_document_params(bufnr) }
-  client:request(ms.textDocument_foldingRange, params, handler, bufnr)
+  local _, request_id = client:request(ms.textDocument_foldingRange, params, handler, bufnr)
+  return request_id
 end
 
 -- NOTE:
@@ -279,7 +281,19 @@ end
 
 --- Close all {kind} of folds in the the window with {winid}.
 ---
----@param kind lsp.FoldingRangeKind 'comment' | 'imports' | 'region'
+--- To automatically fold imports when opening a file, you can use an autocmd:
+---
+--- ```lua
+--- vim.api.nvim_create_autocmd('LspNotify', {
+---   callback = function(args)
+---     if args.data.method == 'textDocument/didOpen' then
+---       vim.lsp.folding_range.foldclose('imports', vim.fn.bufwinid(args.buf))
+---     end
+---   end,
+--- })
+--- ```
+---
+---@param kind lsp.FoldingRangeKind Kind to close, one of "comment", "imports" or "region".
 ---@param winid? integer Defaults to the current window.
 function M.foldclose(kind, winid)
   vim.validate('kind', kind, 'string')
@@ -292,7 +306,39 @@ function M.foldclose(kind, winid)
     return
   end
 
-  foldclose(kind, winid)
+  if bufstate.version == util.buf_versions[bufnr] then
+    foldclose(kind, winid)
+    return
+  end
+  -- Schedule `foldclose()` if the buffer is not up-to-date.
+
+  --- Index in the form of request_id -> true?
+  ---@type table<integer, true?>
+  local scheduled_request = {}
+  for _, client in
+    ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
+  do
+    local request_id = request(bufnr, client)
+    if request_id then
+      scheduled_request[request_id] = true
+    end
+  end
+
+  api.nvim_create_autocmd('LspRequest', {
+    buffer = bufnr,
+    callback = function(args)
+      ---@type integer
+      local request_id = args.data.request_id
+      if scheduled_request[request_id] and args.data.request.type == 'complete' then
+        scheduled_request[request_id] = nil
+      end
+      -- Do `foldclose()` if all the requests is completed.
+      if next(scheduled_request) == nil then
+        foldclose(kind, winid)
+        return true
+      end
+    end,
+  })
 end
 
 ---@param lnum? integer
--- a/runtime/lua/vim/lsp/folding_range.lua
+++ b/runtime/lua/vim/lsp/folding_range.lua
@@ -11,6 +11,9 @@ local M = {}
 ---
 --- Index in the form of row -> [foldlevel, mark]
 ---@field row_level table<integer, [integer, ">" | "<"?]?>
+---
+--- Index in the form of start_row -> kinds
+---@field row_kinds table<integer, table<lsp.FoldingRangeKind, boolean>?>>
 
 ---@type table<integer, vim.lsp.folding_range.BufState?>
 local bufstates = {}
@@ -20,12 +23,20 @@ local bufstates = {}
 ---@param ranges lsp.FoldingRange[]
 local function rangeadd(bufstate, ranges)
   local row_level = bufstate.row_level
+  local row_kinds = bufstate.row_kinds
 
   for _, range in ipairs(ranges) do
     local start_row = range.startLine
     local end_row = range.endLine
     -- Adding folds within a single line is not supported by Nvim.
     if start_row ~= end_row then
+      local kind = range.kind
+      if kind then
+        local kinds = row_kinds[start_row] or {}
+        kinds[kind] = true
+        row_kinds[start_row] = kinds
+      end
+
       for row = start_row, end_row do
         local level = row_level[row] or { 0 }
         level[1] = level[1] + 1
@@ -92,6 +103,7 @@ local function handler(err, result, ctx)
 
   local bufstate = assert(bufstates[bufnr])
   bufstate.row_level = {}
+  bufstate.row_kinds = {}
 
   rangeadd(bufstate, result)
   bufstate.version = ctx.version
@@ -139,6 +151,7 @@ local function setup(bufnr)
   -- Register the new `bufstate`.
   bufstates[bufnr] = {
     row_level = {},
+    row_kinds = {},
   }
 
   -- Event hooks from `buf_attach` can't be removed externally.
@@ -155,6 +168,7 @@ local function setup(bufnr)
     on_reload = function()
       bufstates[bufnr] = {
         row_level = {},
+        row_kinds = {},
       }
       for _, client in
         ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
@@ -202,6 +216,7 @@ local function setup(bufnr)
         if api.nvim_buf_is_loaded(bufnr) then
           bufstates[bufnr] = {
             row_level = {},
+            row_kinds = {},
           }
           foldupdate(bufnr)
         end
@@ -244,6 +259,40 @@ local function setup(bufnr)
   end
 
   return bufstates[bufnr]
+end
+
+---@param kind lsp.FoldingRangeKind
+---@param winid integer
+local function foldclose(kind, winid)
+  vim._with({ win = winid }, function()
+    local bufnr = api.nvim_win_get_buf(winid)
+    local row_kinds = bufstates[bufnr].row_kinds
+    -- Reverse traverse to ensure that the smallest ranges are closed first.
+    for row = api.nvim_buf_line_count(bufnr) - 1, 0, -1 do
+      local kinds = row_kinds[row]
+      if kinds and kinds[kind] then
+        vim.cmd(row + 1 .. 'foldclose')
+      end
+    end
+  end)
+end
+
+--- Close all {kind} of folds in the the window with {winid}.
+---
+---@param kind lsp.FoldingRangeKind 'comment' | 'imports' | 'region'
+---@param winid? integer Defaults to the current window.
+function M.foldclose(kind, winid)
+  vim.validate('kind', kind, 'string')
+  vim.validate('winid', winid, 'number', true)
+
+  winid = winid or api.nvim_get_current_win()
+  local bufnr = api.nvim_win_get_buf(winid)
+  local bufstate = bufstates[bufnr]
+  if not bufstate then
+    return
+  end
+
+  foldclose(kind, winid)
 end
 
 ---@param lnum? integer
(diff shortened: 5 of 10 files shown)


runtime/doc/lua.txt
```diff
@@ -2138,6 +2138,37 @@ vim.list_extend({dst}, {src}, {start}, {finish})           *vim.list_extend()*
...
+    https://stackoverflow.com/questions/12394841/safely-remove-items-from-an-array-table-while-iterating/53038524#53038524
+
+    Parameters: ~
+      • {t}      (`any[]`)
+      • {first}  (`integer`)
+      • {last}   (`integer`)
+      • {v}      (`any`)
+
+vim.list_remove({t}, {first}, {last})                      *vim.list_remove()*
+    Efficiently remove items from middle of a list a list.
```
Julian, line 2158:
```suggestion
    Efficiently remove items from middle of a list.
```

runtime/doc/news.txt
```diff
@@ -230,6 +232,8 @@ LUA
   is more performant and easier to read.
 • |vim.str_byteindex()| and |vim.str_utfindex()| gained overload signatures
   supporting two new parameters, `encoding` and `strict_indexing`.
+• |vim.list_insert()| and |vim.list_remove()| were added to efficiently
+  insert/remove items from middle of a list a list.
```
Julian, line 236:
```suggestion
  insert/remove items from middle of a list.
```

runtime/doc/news.txt
```diff
@@ -222,6 +222,8 @@ LSP
 • |vim.lsp.buf.hover()| now highlights hover ranges using the
   |hl-LspReferenceTarget| highlight group.
 • Functions in |vim.lsp.Client| can now be called as methods.
+• Implemented LSP inlay hints: |vim.lsp.foldexpr()|
```
Julian, line 225:
```suggestion
• Implemented LSP folding ranges: |vim.lsp.foldexpr()|
```

ofseed: Thanks.

runtime/lua/vim/shared.lua
```diff
@@ -737,6 +737,51 @@ function vim.list_slice(list, start, finish)
...
+--- every iteration, instead this function will re-index  the table exactly
+--- once.
+---
+--- Based on https://stackoverflow.com/questions/12394841/safely-remove-items-from-an-array-table-while-iterating/53038524#53038524
+---
+---@param t any[]
+---@param first integer
+---@param last integer
+---@param v any
+function vim.list_insert(t, first, last, v)
```
lewis6991, line 752: Can you not expose this by prefixing it with an `_`.

runtime/lua/vim/shared.lua
```diff
@@ -737,6 +737,51 @@ function vim.list_slice(list, start, finish)
...
+--- Calling table.remove() in a loop will re-index the tail of the table on
+--- every iteration, instead this function will re-index  the table exactly
+--- once.
+---
+--- Based on https://stackoverflow.com/questions/12394841/safely-remove-items-from-an-array-table-while-iterating/53038524#53038524
+---
+---@param t any[]
+---@param first integer
+---@param last integer
+function vim.list_remove(t, first, last)
```
lewis6991, line 777: Same here.

runtime/lua/vim/shared.lua
ofseed: Do you mean to consider it as an internal API? Then I will also remove it from the commit messages.

lewis6991: Yes please.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,297 @@
...
+    on_detach = function()
+      bufstates[bufnr] = nil
+    end,
+    on_reload = function()
+      bufstate = { row_level = {}, row_text = {}, kind_ranges = {} }
+      bufstates[bufnr] = bufstate
+      for _, client in
+        ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
+      do
+        if client.server_capabilities.foldingRangeProvider then
```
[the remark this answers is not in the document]
ofseed: Oh, I know the reason now, `foldingRange` was not added to the supported methods of `client:supports_method` 😆. These checks should not be necessary because there is `get_clients()`

ofseed:
I have added tests for all the new APIs in this PR. This PR is now ready to be merged (or reviewed, of course).

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    return
+  end
+
+  local ranges = bufstate.kind_ranges[kind] or {}
+  foldclose(winid, ranges)
+end
+
+---@return string
+function M._foldtext()
+  local bufnr = api.nvim_get_current_buf()
```
lewis6991, line 296: Is this really doing anything useful? It just seems to be returning the text of the top of the folded region?

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -217,6 +217,11 @@ RCS[ms.textDocument_diagnostic] = function(...)
   return vim.lsp.diagnostic.on_diagnostic(...)
 end
 
+--- @private
+RCS[ms.textDocument_foldingRange] = function(...)
+  return vim.lsp.folding_range.on_folding_range(...)
+end
```
lewis6991, line 223: We don't add handlers here anymore. Instead this should be inlined into the fold module.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    })
+  end
+end
+
+--- |lsp-handler| for the method `textDocument/foldingRange`
+--- Store folding range then update folds for a specific buffer.
+---@private
+---@param result lsp.FoldingRange[]?
+---@param ctx lsp.HandlerContext
+function M.on_folding_range(err, result, ctx)
```
lewis6991, line 62: Should this collect the results of all servers and process them in one loop so we then only schedule one update at the end?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+---@type table<integer, vim.lsp.folding_range.BufState?>
+local bufstates = {}
+
+--- Update visible folds in the buffer.
+---@param bufnr integer
+local function foldupdate(bufnr)
+  for _, winid in ipairs(vim.fn.win_findbuf(bufnr)) do
+    local wininfo = vim.fn.getwininfo(winid)[1]
+    if wininfo and wininfo.tabnr == vim.fn.tabpagenr() then
+      if vim.wo[winid].foldmethod == 'expr' then
```
lewis6991, line 30: Is this condition really necessary, especially since the expr could be something different anyway?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+
+local augroup_update = api.nvim_create_augroup('vim_lsp_folding_range/update', {})
+
+--- Schedule `foldupdate()` after leaving insert mode.
+---@param bufnr integer
+local function schedule_foldupdate(bufnr)
+  if #(api.nvim_get_autocmds({
+    group = augroup_update,
+    buffer = bufnr,
+  })) == 0 then
```
lewis6991, line 45: Instead of collecting autocmds, would it be better to send an id and check for that instead?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+        vim._list_remove(bufstate.row_level, start_row, start_row + math.abs(row) - 1)
+      end
+    end,
+  })
+  api.nvim_create_autocmd('LspDetach', {
+    group = augroup_refresh,
+    buffer = bufnr,
+    callback = function()
+      local clients = vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange })
+      if vim.tbl_isempty(clients) then
```
lewis6991, line 177:
```suggestion
      if next(clients) then
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    schedule_foldupdate(bufnr)
+  else
+    foldupdate(bufnr)
+  end
+end
+
+local augroup_refresh = vim.api.nvim_create_augroup('vim_lsp_folding_range/refresh', {})
+
+---@param bufnr integer
+---@param client_id integer The ID of the client whose server supports `foldingRange`.
```
lewis6991, line 122:
```suggestion
---@param client vim.lsp.Client The client whose server supports `foldingRange`.
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    callback = function(args)
+      local client = assert(vim.lsp.get_client_by_id(args.data.client_id))
+      if
+        client:supports_method(ms.textDocument_foldingRange, bufnr)
+        and (
+          args.data.method == ms.textDocument_didChange
+          or args.data.method == ms.textDocument_didOpen
+        )
+      then
+        refresh(bufnr, args.data.client_id)
```
lewis6991, line 202:
```suggestion
        refresh(bufnr, client)
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    end,
+  })
+  api.nvim_create_autocmd('LspDetach', {
+    group = augroup_refresh,
+    buffer = bufnr,
+    callback = function()
+      local clients = vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange })
+      if vim.tbl_isempty(clients) then
+        for _, client in ipairs(clients) do
+          refresh(bufnr, client.id)
```
lewis6991, line 179:
```suggestion
          refresh(bufnr, client)
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    on_detach = function()
+      bufstates[bufnr] = nil
+    end,
+    on_reload = function()
+      bufstate = { row_level = {}, row_text = {}, kind_ranges = {} }
+      bufstates[bufnr] = bufstate
+      for _, client in
+        ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
+      do
+        refresh(bufnr, client.id)
```
lewis6991, line 150:
```suggestion
        refresh(bufnr, client)
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    foldupdate(bufnr)
+  end
+end
+
+local augroup_refresh = vim.api.nvim_create_augroup('vim_lsp_folding_range/refresh', {})
+
+---@param bufnr integer
+---@param client_id integer The ID of the client whose server supports `foldingRange`.
+---@return integer? request_id
+local function refresh(bufnr, client_id)
```
lewis6991, line 124:
```suggestion
local function refresh(bufnr, client)
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+  ---@type lsp.FoldingRangeParams
+  local params = { textDocument = util.make_text_document_params(bufnr) }
+  local client = assert(vim.lsp.get_client_by_id(client_id))
```
lewis6991, line 127:
```suggestion
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+      then
+        refresh(bufnr, args.data.client_id)
+      end
+    end,
+  })
+
+  for _, client in
+    ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
+  do
+    refresh(bufnr, client.id)
```
lewis6991, line 210:
```suggestion
    refresh(bufnr, client)
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+    return
+  end
+  ---@cast bufstate vim.lsp.folding_range.BufState
+
+  -- Defer `foldclose()` if the buffer is not up-to-date.
+  if bufstate.version ~= util.buf_versions[bufnr] then
+    for _, client in
+      ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
+    do
+      local request_id = refresh(bufnr, client.id)
```
lewis6991, line 274:
```suggestion
      local request_id = refresh(bufnr, client)
```

runtime/lua/vim/lsp/folding_range.lua
ofseed: It is to support `collapsedText` in `lsp.foldingRange` (proposed). To be honest, I haven't found a server that supports it, but I have a friend who is trying to implement LSP for Verilog and said this feature is somewhat useful. Considering it is very simple, I added it.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+      api.nvim_create_autocmd('LspRequest', {
```
lewis6991, line 275: This autocmd seems a bit redundant, isn't the callback unconditional?

runtime/lua/vim/lsp/folding_range.lua
ofseed: I am currently not considering support for multiple servers. This is because such functionality is generally provided by a single server, and I feel that vim's own fold feature is unlikely to support this (correct me if I'm wrong), as multiple sources of `foldingRange` may have intersections, making it impossible to calculate the `foldlevel` in that case.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,320 @@
...
+
+--- Setup event hooks for folding range.
+---@param bufnr integer
+---@return vim.lsp.folding_range.BufState
+local function setup(bufnr)
+  ---@type vim.lsp.folding_range.BufState
+  local bufstate = { row_level = {}, row_text = {}, kind_ranges = {} }
+  bufstates[bufnr] = bufstate
+
+  api.nvim_buf_attach(bufnr, false, {
```
lewis6991, line 140: Need to also detach the buffer when all clients detach too.

runtime/lua/vim/lsp/folding_range.lua
clason: Let's wait until it's actually part of the spec then; things might change too much until then.

(It's better to leave something as a follow-up than to risk adding an API we have to do the deprecation dance around.)

runtime/lua/vim/lsp/folding_range.lua
ofseed: `foldupdate` to be called is possible only after the `vim.lsp.foldexpr` setup, which means that when it is called, `foldexpr` has already been ensured to be `vim.lsp.foldexpr`. This is to prevent the situation where `expr` is set but `foldmethod` is not set.

runtime/lua/vim/lsp/folding_range.lua
ofseed: This is to avoid executing `foldclose` when the buffer has already been modified, but the bufstate has not yet been updated. This ensures that the result of `foldclose` corresponds to the modified buffer.

More specifically, there is a situation where `foldclose` is executed immediately when opening a file (common in autocmd, such as the automatic folding import example given above). At this point, the bufstate is likely not updated yet, but what you actually want is the result after it has been updated, corresponding to the content you see in the buffer.

runtime/lua/vim/lsp/folding_range.lua
ofseed: Is [this](https://github.com/ofseed/neovim/blob/e2a24c75c34a0c3adb6caf318d7a289f0c27c874/runtime/lua/vim/lsp/folding_range.lua#L172-L189) not enough?

Detaching the buffer should be unnecessary, because if this buffer has already been setup, then even if all clients detach at a certain moment (or any client has not yet attached), it is possible that it will be attached by new clients in the future.

runtime/lua/vim/lsp/folding_range.lua
ofseed: It has already become part of the current specification. In [3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#version_3_17_0). It's just that the server I commonly use does not support this feature.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: Supporting multiple providers is valuable as it opens the door for plugins to implement custom folding via the LSP interface.

Supporting multiple servers is something we've put a lot of effort into recently and want to support it for every method.

Looking for intersections would be a problem, but this implementation just needs to make sure they are handled gracefully, and not necessarily correctly. I think the fold backend will already do something graceful so you may not need to do anything here, we just need to make sure the plumping is in place.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: So you basically want to defer this code until all lines in a redraw have been calculated?

runtime/lua/vim/lsp/folding_range.lua
lewis6991: If there are no clients, then the buf attach callbacks should not be invoked.

runtime/lua/vim/lsp/folding_range.lua
ofseed: It is not for redrawing, but to wait for the `foldingRange` request to get a result.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: The fact it doesn't provide highlighting already makes it behind other foldtexts, so I'd omit this for now.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: Oh, because `LspRequest` is triggered on the handler. Ok can you just add some comments explaining this. I'd probably want to re-implement this in the future when some other things have been added.

runtime/lua/vim/lsp/folding_range.lua
ofseed: If you are referring to `on_bytes`, then it indeed will not be invoked, because when there are no clients, [here](https://github.com/ofseed/neovim/blob/e2a24c75c34a0c3adb6caf318d7a289f0c27c874/runtime/lua/vim/lsp/folding_range.lua#L172-L189) the row_levels will be cleared. This is intentional, because `row_level` being empty covers a wider range of possibilities.

runtime/lua/vim/lsp/folding_range.lua
MariaSolOs: `collapsed_text` is often used for folding regions (delimited by `#region/#endregion`).

Example from TypeScript:
image

I think that we can support it, I doubt that the spec will change a lot.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I can make some more attempts in this area, but I hope that will be in another PR. Currently (only supporting one server), it can handle most needs. Even implementing custom folding through the LSP interface often covers situations that LSP server does not support folding on a specific filetype. I anticipate that such functionality may bring a lot of complexity, and it would be better to make modifications when support for a single server is already mature, as this proves that providing LSP folding through `foldexpr` is feasible.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: Because it is standard Lua. No need to use a redundant wrapper.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: `on_bytes` will still be invoked until you detach from the buffer.

runtime/lua/vim/lsp/folding_range.lua
clason: (I just went by the "(proposed)" in the second comment. If it's already part of the spec, then that's different of course.)

runtime/lua/vim/lsp/folding_range.lua
ofseed: Looking back, I think `#client == 0` should be sufficient, and its meaning should be clearer.

runtime/lua/vim/lsp/folding_range.lua
MariaSolOs: > (I just went by the "(proposed)" in the second comment. If it's already part of the spec, then that's different of course.)

That's fair! I just wanted to give an example of how it is used :)

runtime/lua/vim/lsp/folding_range.lua
lewis6991: Then this PR at the very least needs to make sure that multiple clients are handled gracefully by not processing multiple responses and probably only make a request to a single server.

runtime/lua/vim/lsp/folding_range.lua
ofseed: Are you referring to directly removing the callback set by this buf attach? Then we would need to find a way to add it back, which would significantly increase complexity. I think it's sufficient to return after it finds the table is empty. Callbacks created by `buf_attach` will only be triggered on the buffer associated with a window whose `vim.wo.foldexpr = "v:lua.vim.lsp.foldexpr"`, so it makes sense in the vast majority of cases. I'm not quite sure what your suggested approach is.

runtime/lua/vim/lsp/folding_range.lua
ofseed: After weighing the options, considering that this PR already includes some code for handling multiple clients, I plan to add preliminary support within this PR. This might be better than asking the user to choose which client to use when encountering multiple clients. However, this shouldn’t be the main goal, especially since I can’t find any use case in practical scenarios that would allow me to perform manual testing.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: You reattach when you set up. You just need to return true in on_bytes to detach. It isn't that hard, it is just basic hygiene to clean up after all clients have detached.

ofseed: I think clearing bufstate is not an option. Because bufstate represents a buffer's window's `foldexpr` is `vim.lsp.foldexpr`, if it is cleared, next time `vim.lsp.foldexpr` is calculated, setup will be called again, and the callback will be created.

This is a design based on the callbacks and the bufstate has the same lifecycle. If the logic of the code is unclear, I can add more explanations later. However, it may be necessary to create an autocmd for `OptionSet` to clear bufstate and these callbacks when `foldexpr` is not `vim.lsp.foldexpr`.

ofseed: According to my understanding, you mean that for a buffer with `foldexpr` set to `vim.lsp.foldexpr`, it can be attached by LSP before or after (it's a goal), but during the time it is not attached (even if `foldexpr` is still `vim.lsp.foldexpr`), all callback executions need to be stopped.

This can be done through `LspAttach` and `LspDetach`, but I don't think it brings any real benefits. Especially during this time period, these callbacks return almost immediately

ofseed:
For the somewhat controversial bufstate and callback cleanup mentioned above, I pushed an update to try to control them more granularly, and I detailed the expected behavior in the comments.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,365 @@
...
+
+--- Setup `bufstate` and its event hooks.
+---@param bufnr integer
+---@return vim.lsp.folding_range.BufState
+local function setup(bufnr)
+  ---@type vim.lsp.folding_range.BufState
+  local bufstate = { row_level = {}, row_text = {}, kind_ranges = {} }
+  if not api.nvim_buf_is_loaded(bufnr) then
+    return bufstate
+  end
```
lewis6991, line 155: Can't you check for available clients here and return if there are none, this avoids setting up attaching to the buffer and setting up autocmds, and I think allows you to remove the OptionSet autocmd.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,365 @@
...
+    end,
+    --- When modifying the buffer, new foldinfo is scheduled to be updated,
+    --- synchronize the changed rows with their previous foldlevels as possible,
+    --- before new foldlevels is updated.
+    ---@param start_row integer
+    ---@param old_row integer
+    ---@param new_row integer
+    on_bytes = function(_, _, _, start_row, _, _, old_row, _, _, new_row, _, _)
+      if bufstate == nil then
+        return true
```
lewis6991, line 182: I think you can remove the autocmds heres.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,365 @@
...
+    buffer = bufnr,
+    callback = function()
+      local clients = vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange })
+      if #clients == 0 then
+        for _, client in ipairs(clients) do
+          refresh(bufnr, client)
+        end
+      else
+        bufstate = { row_level = {}, row_text = {}, kind_ranges = {} }
+        bufstates[bufnr] = bufstate
```
lewis6991, line 207: I don't think you should be registering bufstate here.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,365 @@
...
+  if bufstate and bufstate.row_text[row] then
+    return bufstate.row_text[row]
+  end
+  return vim.fn.getline(lnum)
+end
+
+---@param lnum? integer
+---@return string level
+function M._foldexpr(lnum)
+  vim.validate('lnum', lnum, 'number', true)
```
lewis6991, line 355: This function is pretty hot and is only called internally by vim, so I think this `vim.validate` can be removed.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,365 @@
...
+  local row_level = {}
+  ---@type table<integer, string>
+  local row_text = {}
+  ---@type table<lsp.FoldingRangeKind, lsp.FoldingRange[]?>>
+  local kind_ranges = {}
+
+  for _, range in ipairs(result) do
+    local kind = range.kind
+    if kind then
+      ---@cast kind lsp.FoldingRangeKind
```
lewis6991, line 85: Is LuaLS not able to infer this?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,365 @@
...
+api.nvim_create_autocmd('OptionSet', {
+  group = augroup_setup,
+  pattern = 'foldexpr',
+  callback = function()
+    if vim.v.option_old == 'v:lua.vim.lsp.foldexpr()' then
+      for bufnr in pairs(bufstates) do
+        if
+          ---@param winid integer
+          vim.iter(vim.fn.win_findbuf(bufnr)):all(function(winid)
+            return vim.wo[winid].foldexpr ~= 'v:lua.vim.lsp.foldexpr()'
```
lewis6991, line 251: This won't work if #31280 gets merged.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I remember it wasn't before, but I am about to change the data structure of `kind_ranges` (for multi-client support, and to ensure it collapses from the smallest place, this issue will not exist).

runtime/lua/vim/lsp/folding_range.lua
ofseed: This is not registering but resetting bufstate. The assignment `bufstates[bufnr] = bufstate` needs to be done because `bufstates[bufnr]` still points to the previously constructed `bufstate`, not the new one.

runtime/lua/vim/lsp/folding_range.lua
ofseed: If garbage collection is needed, `OptionSet` is necessary. Because the user may have changed `foldexpr` themselves, there is no way to know that the user has changed it other than using `OptionSet`. At the same time, LSP clients will not make any changes in this case.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I have already subscribed to this PR. If it is merged, this place will become more efficient because we can check if the expr functions are equal instead of checking strings.

runtime/lua/vim/lsp/folding_range.lua
ofseed: As I noted in the code above, the lifecycle of bufstate and these callbacks should be synchronized. So if we do not need to destroy bufstate here, then we also do not need to destroy the callback. However, callbacks of `nvim_buf_attach` can only be destroyed if it returns true, which is the reason for adding the branch here.

ofseed:
I pushed a new update that, in addition to changing the code marked as `resolved` above, includes an additional change that adjusts the data structure required for `foldclose` to support multiple clients while also ensuring that the smallest range is prioritized for closing. Prioritizing the closing of the smallest range makes sense because after closing a larger range, vim will not close the smaller ranges within it. This broadens the applicability of `foldclose`.

ofseed:
The reason for having the same lifecycle for bufstate and its related callbacks is that they are interdependent. If their lifecycles are not the same, then any error on one side will lead to issues on the other side. If the code on one side is changed, it may cause errors in the code on the other side. In reality, this would result in an indirect recursion of continuously reporting errors; we might just happen to be at a certain moment, but new errors will arise in the future. I believe we should consider their lifecycle management based on ensuring that their lifecycles are the same, as finding a way that perfectly covers all situations is very risky.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: The point is that there is no reliable way of checking this, nor do I think it is necessary.

lewis6991: Fine, revert everything back to the stage you are happy with and then someone else can review and merge this. If this gets merged I'll take a look at resolving the memory leaks in a follow up.

ofseed: Perhaps I'm overthinking it, and I hope that's the case. Later, I will add some test cases that I'm concerned about so that the code can be changed more safely.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I have changed it to a method very similar to your suggestion, and now I will mark this thread as resolved.

ofseed:
@lewis6991 Thank you so much for your patient review. I apologize if I’ve wasted your time. I’ve revisited the management of these lifecycles and added tests to cover them, ensuring that callbacks are destroyed when the buffer is unloaded and avoiding duplicate creation on the same buffer. I’d appreciate any suggestions if you have the time to review it again.

I am currently working on implementing support for multiple clients.

lewis6991 (approved):
Would it make sense for `setup()` to early return when there are no supported clients, and additionally set `bufstate=nil` in `LspDetach` so we detach from the buffer on the next `on_bytes` call. Then every subsequent call to `foldexpr()` will early return until a supported client becomes available which we then we re-attach to the buffer?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,396 @@
...
+    on_reload = function()
+      if bufstate == nil then
+        return true
+      end
+      bufstate = {
+        row_level = {},
+        row_kinds = {},
+        row_text = {},
+      }
+      bufstates[bufnr] = bufstate
```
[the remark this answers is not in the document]
ofseed: Oh right, there’s no reason to use `bufstate` in `setup` instead of `bufstates[bufnr]`.

ofseed:
It works in most cases, but there are still some edge cases. The main issue is that we cannot guarantee that `foldexpr()` will be executed after a new LSP attach. For example, when a newly started nvim opens its first buffer, `foldexpr()` will be executed immediately, but it has not been LSP attached yet, so bufstate will not be. After a very short time, the LSP attach completes, but at this point, `foldexpr()` is unlikely to be executed again, the folds in the buffer keep outdated unless modifying the buffer or using something like `zx` after that. This issue can be resolved with an additional `LspAttach` autocmd, but I'm not sure if there are other edge cases because if a new autocmd is created, we also need to consider how to destroy it.

ofseed:
I pushed support for multiple clients, so I marked the related reviews above as resolved.

runtime/doc/lsp.txt
```diff
@@ -1647,6 +1659,27 @@ trigger()                                       *vim.lsp.completion.trigger()*
     Trigger LSP completion in the current buffer.
 
 
+==============================================================================
+Lua module: vim.lsp.folding_range                          *lsp-folding_range*
+
+foldclose({kind}, {winid})                 *vim.lsp.folding_range.foldclose()*
```
justinmk, line 1665: The logic is maybe complex enough to live in its own module, but we probably don't need to expose it as such? Should this live at `vim.lsp.foldclose`, or do we plan to have many more fold-related functions in the future? E.g. `folding_range.enable()`, `folding_range.is_enabled`, etc.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+---@field row_level table<integer, [integer, ">" | "<"?]?>
+---
+--- Index in the form of start_row -> kinds
+---@field row_kinds table<integer, table<lsp.FoldingRangeKind, true?>?>>
+---
+--- Index in the form of start_row -> collapsed_text
+---@field row_text table<integer, string?>
+
+---@type table<integer, vim.lsp.folding_range.BufState?>
+local bufstates = {}
```
justinmk, line 28: Just for reference: this is yet another case of a pattern I think we discussed generalizing, right? (can't find the PR/issue ...)

runtime/doc/lsp.txt
ofseed: This is mainly about naming convention. I found that the functions under `vim.lsp.<name>`, apart from those directly related to LSP (such as `vim.lsp.buf_request`), only have functions related to specific methods when there are concrete interfaces in nvim. For example, `vim.lsp.tagfunc`, `vim.lsp.formatexpr`. The others are either in `vim.lsp.buf` or have their own module.

It is also possible to add additional functions. I hope to provide as few interfaces as possible initially to cover the functionality that LSP can offer. However, for convenience, it may also be possible to add a `foldopen()` in the future to correspond with `foldclose()`, or to add a function/hook to get `foldlevel` corresponding to vim's `set foldlevel`.

runtime/lua/vim/shared.lua
```diff
@@ -737,6 +737,51 @@ function vim.list_slice(list, start, finish)
...
+--- Calling table.remove() in a loop will re-index the tail of the table on
+--- every iteration, instead this function will re-index  the table exactly
+--- once.
+---
+--- Based on https://stackoverflow.com/questions/12394841/safely-remove-items-from-an-array-table-while-iterating/53038524#53038524
+---
+---@param t any[]
+---@param first integer
+---@param last integer
+function vim._list_remove(t, first, last)
```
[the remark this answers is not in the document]
lewis6991: `remove` was initially chosen as this function is supposed to be a replacement for `table.remove`. Same for `vim.list_insert` and `table.insert`.

runtime/doc/lsp.txt
justinmk: If more functions are planned then I guess this is fine.

runtime/lua/vim/shared.lua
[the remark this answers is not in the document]
justinmk: Not a blocker for now

runtime/lua/vim/lsp/folding_range.lua
ofseed: Yes, there was [a draft PR](https://github.com/neovim/neovim/pull/30841) before. Many of lewis's earlier reviews focused on ensuring that these bufstates can be correctly destroyed to avoid potential memory leaks.

I still haven't given up on this idea, so I kept the code related to managing bufstates in a single `setup` function there so that it can truly generalize to other modules in the future. Currently, it should ensure that nothing remains on the unloaded buffer. It would be better if we could delete it when there is no corresponding method supported on the buffer, as lewis suggested. Still, for now, the actual effect is not very significant, and the risk is relatively high.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+--- vim.api.nvim_create_autocmd('LspNotify', {
+---   callback = function(args)
+---     if args.data.method == 'textDocument/didOpen' then
+---       vim.lsp.folding_range.foldclose('imports', vim.fn.bufwinid(args.buf))
+---     end
+---   end,
+--- })
+--- ```
+---
+---@param kind lsp.FoldingRangeKind "comment" | "imports" | "region"
```
lewis6991, line 345: For consistency with other docs:

```suggestion
---@param kind lsp.FoldingRangeKind 'comment'|'imports'|'region'
```

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+  vim.validate('kind', kind, 'string')
+  vim.validate('winid', winid, 'number', true)
+
+  winid = winid or api.nvim_get_current_win()
+  local bufnr = api.nvim_win_get_buf(winid)
+  local bufstate = bufstates[bufnr]
+  if not bufstate then
+    return
+  end
+  ---@cast bufstate -nil
```
lewis6991, line 357: I'm pretty sure LuaLS correctly infers this already.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+
+  -- Schedule `foldclose()` if the buffer is not up-to-date.
+  if bufstate.version ~= util.buf_versions[bufnr] then
```
lewis6991, line 360:
```suggestion
  if bufstate.version == util.buf_versions[bufnr] then
    foldclose(kind, winid)
    return
  end
```

To reduce code nesting.

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+    --- Index in the form of request_id -> true?
+    ---@type table<integer, true?>
+    local scheduled_request = {}
+    for _, client in
+      ipairs(vim.lsp.get_clients({ bufnr = bufnr, method = ms.textDocument_foldingRange }))
+    do
+      local request_id = request(bufnr, client)
+      if request_id then
+        scheduled_request[request_id] = true
+        api.nvim_create_autocmd('LspRequest', {
```
lewis6991, line 370: Can this be flattened into one autocmd?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+local function foldclose(kind, winid)
+  vim._with({ win = winid }, function()
+    local bufnr = api.nvim_win_get_buf(winid)
+    local row_kinds = bufstates[bufnr].row_kinds
+    -- Reverse traverse to ensure that the smallest ranges are closed first.
+    for row = api.nvim_buf_line_count(bufnr) - 1, 0, -1 do
+      local kinds = row_kinds[row]
+      if kinds and kinds[kind] then
+        vim.cmd(row + 1 .. 'foldclose')
+      end
```
lewis6991, line 326: This looks quite costly. Doesn't `vim.cmd.normal('zM')` do this?

runtime/lua/vim/lsp/folding_range.lua
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  -- Event hooks from `buf_attach` can't be removed externally.
+  -- Hooks and `bufstate` share the same lifecycle;
+  -- they should self-destroy if `bufstate == nil`.
+  api.nvim_buf_attach(bufnr, false, {
+    -- `on_detach` also runs on buffer reload (`:e`).
+    -- Ensure `bufstate` and hooks are cleared to avoid duplication or leftover states.
+    on_detach = function()
+      ---@diagnostic disable-next-line: cast-local-type
```
lewis6991, line 180: What's this doing?

lewis6991 (approved):

runtime/lua/vim/lsp/folding_range.lua
ofseed: Thanks, I didn't know there were performance differences in this. But I think it is still necessary to use `:foldclose`, because `zc` affects `foldenable` and `'zm'` affects `'foldlevel'`, we should choose one that has no side effects.

runtime/lua/vim/lsp/folding_range.lua
ofseed: Right, I may have left unnecessary code during the rebase process.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I've deleted it now.

runtime/lua/vim/lsp/folding_range.lua
ofseed: Done

runtime/lua/vim/lsp/folding_range.lua
ofseed: It seems that changing it to the form of `'` will cause lintdoc to error, possibly because it is confused with type unions.

lewis6991: Oh the type is `lsp.FoldRangeKind`. These can be fully omitted.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: Running a command on every single line in the buffer will exhibit issues on large files.

I thought I saw `foldclose` called in other places in the `foldexpr` path, but I can't see any now, so `foldclose` looks like it can only be triggered directly by the user, so will be ok for now.

runtime/lua/vim/lsp/folding_range.lua
lewis6991, line 1: I feel like this module should be called `fold.lua`. `folding_range` is just the name of the current method the spec provides, but more methods may be added in the future.

ofseed: Could other methods be unrelated to this method? Actually, I thought about using semantic tokens to provide foldtext, but this doesn't have much impact since foldtext is already under `vim.lsp`.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I slightly added content, keeping it in single quotes instead of double quotes.

ofseed: > Oh the type is `lsp.FoldRangeKind`. These can be fully omitted.

The purpose of this comment is to generate documentation, as `lsp.FoldingRangeKind` will not be automatically generated, so I mentioned their possible values.

runtime/lua/vim/lsp/folding_range.lua
ofseed: I have precautions regarding this. `:foldclose` will not be called on every line, but rather at the beginning of a certain `kind`. If it is folding imports, it will only be called once; if it is folding comments, it will also only be called on the first line so even if a file has 100 functions, each with a docstring, it will only be called 100 times. This should already be quite large.

runtime/lua/vim/lsp/folding_range.lua
lewis6991: Folding is a general concept. I don't think it's a good idea to name modules directly from method names. Can we rename this to `fold.lua`. Preferably I would keep this module hidden and expose everything at the `vim.lsp.*` or `vim.lsp.buf.*` level.

lewis6991 (approved):
LGTM. @mfussenegger can you review too?

runtime/lua/vim/lsp/_folding_range.lua
```diff
@@ -0,0 +1,404 @@
...
+--- Index in the form of row -> [foldlevel, mark]
+---@field row_level table<integer, [integer, ">" | "<"?]?>
+---
+--- Index in the form of start_row -> kinds
+---@field row_kinds table<integer, table<lsp.FoldingRangeKind, true?>?>>
+---
+--- Index in the form of start_row -> collapsed_text
+---@field row_text table<integer, string?>
+
+---@type table<integer, vim.lsp.folding_range.BufState?>
```
mfussenegger, line 27: Is there any reason for the `?` suffixes for the table values? It's kinda implicit that entries can be missing and that you get a `nil` so I'm wondering if I'm missing some aspect to this.

runtime/lua/vim/lsp/_folding_range.lua
```diff
@@ -0,0 +1,404 @@
...
+            row_level = {},
+            row_kinds = {},
+            row_text = {},
+          }
+          foldupdate(bufnr)
+        end
+      else
+        for _, client in ipairs(clients) do
+          request(bufnr, client)
+        end
```
mfussenegger, line 250: I don't quite understand why it's necessary to re-request the fold information for other clients if one client detaches from the buffer. The state should still be up to date, no?

Shouldn't this instead only clear out the `bufstate.client_ranges` for the given client id?

runtime/lua/vim/lsp/_folding_range.lua
```diff
@@ -0,0 +1,404 @@
...
+      if scheduled_request[request_id] and args.data.request.type == 'complete' then
+        scheduled_request[request_id] = nil
+      end
+      -- Do `foldclose()` if all the requests is completed.
+      if next(scheduled_request) == nil then
+        foldclose(kind, winid)
+        return true
+      end
+    end,
+  })
```
mfussenegger, line 375: A bit an unusual pattern to use `LspRequest` for a countdown latch to react once all responses were received.
Typical pattern is to use a counter like in `buf_request_all` or in some of the methods in lsp.buf.

But no strong opinion on this.

mfussenegger (commented):
Looks good to me - left a couple of minor comments

runtime/lua/vim/lsp/_folding_range.lua
lewis6991: Nah, they are not implicit, though they should be. Adding them gives you more strict type checking.

runtime/lua/vim/lsp/_folding_range.lua
lewis6991: Hmm, yeah I think that might be better.

runtime/lua/vim/lsp/_folding_range.lua
ofseed: Yes, this can be done; I didn't add `client_ranges` to handle multiple clients before.

runtime/lua/vim/lsp/_folding_range.lua
ofseed: Okay, I will add another `lsp.MultiHandler` and reduce the number of calls to `foldupdate`.

ofseed:
I pushed the changes for the comments above. I also changed a test because that depends on the `LspRequest` command and will not be triggered in the test environment.

runtime/doc/lsp.txt
```diff
@@ -697,6 +698,35 @@ commands                                                    *vim.lsp.commands*
...
     The second argument is the `ctx` of |lsp-handler|
 
+foldclose({kind}, {winid})                               *vim.lsp.foldclose()*
+    Close all {kind} of folds in the the window with {winid}.
+
+    To automatically fold imports when opening a file, you can use an autocmd: >lua
+        vim.api.nvim_create_autocmd('LspNotify', {
+          callback = function(args)
+            if args.data.method == 'textDocument/didOpen' then
+              vim.lsp.folding_range.foldclose('imports', vim.fn.bufwinid(args.buf))
```
lewis6991, line 708: Needs updating.

runtime/lua/vim/lsp/_folding_range.lua
```diff
@@ -0,0 +1,400 @@
...
+        foldupdate(bufnr)
+        scheduled_foldupdate[bufnr] = nil
+      end,
+    })
+  end
+end
+
+---@param result lsp.FoldingRange[]?
+---@type lsp.Handler
+local function handler(err, result, ctx)
```
lewis6991, line 112: Can you implement this so it calls `multi_handler` to eliminate the code duplication? In fact you could probably inline that directly since it should only be a few lines.

runtime/lua/vim/lsp/_folding_range.lua
```diff
@@ -0,0 +1,400 @@
...
+--             return vim.wo[winid].foldexpr ~= 'v:lua.vim.lsp.foldexpr()'
+--           end)
+--         then
+--           bufstates[bufnr] = nil
+--           api.nvim_clear_autocmds({ buffer = bufnr, group = augroup_setup })
+--         end
+--       end
+--     end
+--   end,
+-- })
```
lewis6991, line 329: This block can be removed for now. If we add it then it shouldn't try and compare the value of foldexpr.

runtime/lua/vim/lsp/_folding_range.lua
ofseed: Indeed, code redundancy should be reduced. It is for a general purpose so I did not inline it.

ofseed:
I revisited the code and found that `vim.lsp.buf_request_all` notifies an error when there are no supported clients in the buffer, which is not the behavior we expect. Therefore, I exposed the `on_unsupported` parameter for `vim.lsp.buf_request_all`, partly to keep it in sync with the parameters of `vim.lsp.buf_request`. I kept this change in a separate commit, so please note if the commits in this PR need to be squashed.

runtime/lua/vim/lsp.lua
```diff
@@ -928,10 +930,13 @@ end
 ---               Can also be passed as a function that returns the params table for cases where
 ---               parameters are specific to the client.
 ---@param handler lsp.MultiHandler (function)
---- Handler called after all requests are completed. Server results are passed as
---- a `client_id:result` map.
+---       Handler called after all requests are completed. Server results are passed as
+---       a `client_id:result` map.
+---@param on_unsupported? fun()
```
lewis6991, line 935: Can we avoid doing this and just use `get_clients` to check for unsupported before calling `buf_request_all`?

lewis6991: Also note, many of the functions in `lsp/buf.lua` don't use `buf_request_all` and simply call all the clients manually:

```lua
function M.references(context, opts)
  validate('context', context, 'table', true)
  local clients = lsp.get_clients({ method = ms.textDocument_references, bufnr = bufnr })
  if not next(clients) then
    return
  end

  local all_items = {}

  local function on_done()
    if not next(all_items) then
      vim.notify('No references found')
    else
      ...
    end
  end

  local remaining = #clients
  for _, client in ipairs(clients) do
    local params = ...
    client:request(ms.textDocument_references, params, function(_, result)
      local items = util.locations_to_items(result or {}, client.offset_encoding)
      vim.list_extend(all_items, items)
      remaining = remaining - 1
      if remaining == 0 then
        on_done()
      end
    end)
  end
end
```

So pick your poison. Eventually I plan to refactor this pattern out anyway,  so don't overthink it, just avoid extending any API.

ofseed: Okay, sorry I don't know about these plans.

lewis6991: Yes `vim.lsp.buf_request_all` and `lsp.MultiHandler` will remain, but any changes to them is TBD.

ofseed:
Feel free to ping me if there are any issue related!

clason:
Please open a  new issue with minimal reproducing steps.

PR #31317, opened 2024-11-23
refactor(options): fix confusing naming of `scope` and `req_scope`

famiu:
Problem: The name `scope` is often used to refer to option flags because `OPT_LOCAL` and `OPT_GLOBAL` are often used to determine the option scope. This leads to the name `req_scope` being used for actual option scopes instead.

Solution: Since the end-goal is to remove `OPT_LOCAL` and `OPT_GLOBAL` entirely and replace them with `OptScope`, rename `OptScope` variables to `scope` and the old scope flag variables to `opt_flags`.

### DIFF
--- a/src/nvim/api/deprecated.c
+++ b/src/nvim/api/deprecated.c
@@ -636,12 +636,12 @@ void nvim_win_set_option(uint64_t channel_id, Window window, String name, Object
 /// Gets the value of a global or local (buffer, window) option.
 ///
 /// @param[in]   from       Pointer to buffer or window for local option value.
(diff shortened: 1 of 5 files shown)


zeertzjq (commented):
This is inconsistent with the function names `get_varp_scope()` etc.

famiu:
> This is inconsistent with the function names `get_varp_scope()` etc.

For now, yes, but this is to be addressed by #31084 

Unless you want me to rename those functions temporarily

zeertzjq:
Ah, OK

zeertzjq (approved):

PR #26827, opened 2023-12-31
perf(treesitter): partial injection and incremental invalidation

tomtomjhj:
The two main bottlenecks in editing big files with many injections are (1) parsing and (2) processing the injection queries. This PR resolves (2) by applying injection query to only visible regions.

Benchmark:
1. download <https://raw.githubusercontent.com/torvalds/linux/master/drivers/gpu/drm/amd/include/asic_reg/dcn/dcn_3_2_0_sh_mask.h>
1. setup nvim-treesitter for `cpp` (`.h` file is recognized as `cpp` filetype by default)
1. `:let g:__ts_debug = 1`
1. `:e dcn_3_2_0_sh_mask.h`, `:lua vim.treesitter.start()`
1. go to the last comment in the file (`G?//<CR>`), then append a space (`A<Space><Esc>`)
1. go to the next line (`j`) and append 0 (`A0<Esc>`)
1. see log at `~/.local/state/nvim/treesitter.log`

<details>
<summary>before</summary>

```
1:cpp:(nvim) parse:450: (#regions=1)  { parse_time = 2923.453483, query_time = 0, regions_parsed = 1 }
1:cpp:(nvim) parse:450: (#regions=1)  { parse_time = 0, query_time = 0, regions_parsed = 0 }
1:cpp:(nvim) parse:450: (#regions=1)  { parse_time = 0, query_time = 0, regions_parsed = 0 }
1:comment:(nvim) new:145: (#regions=1)  START
1:cpp:(nvim) new:145: (#regions=1)  START
1:cpp:(nvim) parse:450: (#regions=1)  { parse_time = 0, query_time = 1004.06346, range = { 0, 56 }, regions_parsed = 0 }
1:cpp:(nvim) parse:450: (#regions=194402)  { parse_time = 1.506477, query_time = 0.17726, range = { 0, 56 }, regions_parsed = 25 }
1:comment:(nvim) parse:450: (#regions=27859)  { parse_time = 1.215564, query_time = 0.001373, range = { 0, 56 }, regions_parsed = 5 }
1:cpp:(nvim) parse:450: (#regions=1)  { parse_time = 0, query_time = 0, range = { 222837, 222893 }, regions_parsed = 0 }
1:cpp:(nvim) parse:450: (#regions=194402)  { parse_time = 2.845539, query_time = 0.450294, range = { 222837, 222893 }, regions_parsed = 50 }
[...]

PR #31324, opened 2024-11-23
feat(marks): add conceal_lines to nvim_buf_set_extmark()

luukvbaal:
**feat(marks): add conceal_lines to nvim_buf_set_extmark()**

Implement an extmark property that conceals lines vertically.

**feat(treesitter): vertical conceal for conceal nodes**

TSHighlighter now places marks for conceal_lines metadata. A new
internal decor provider callback _on_conceal_line was added that
instructs the highlighter to place conceal_lines marks whenever the
editor needs to know whether a line is concealed. The bundled markdown
queries use `conceal_lines` metadata to conceal code block fence lines.

Work on #25718

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -2698,6 +2698,10 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
                     When a character is supplied it is used as |:syn-cchar|.
                     "hl_group" is used as highlight for the cchar if provided,
                     otherwise it defaults to |hl-Conceal|.
+                  • conceal_lines: string which should be empty. When
+                    provided, lines in the range are not drawn at all
+                    (according to 'conceallevel'); the next unconcealed line
+                    is drawn instead.
                   • spell: boolean indicating that spell checking should be
                     performed within this extmark
                   • ui_watched: boolean that indicates the mark should be
--- a/runtime/doc/treesitter.txt
+++ b/runtime/doc/treesitter.txt
@@ -492,10 +492,10 @@ capture marks comments as to be checked: >query
 There is also `@nospell` which disables spellchecking regions with `@spell`.
 
                                                 *treesitter-highlight-conceal*
-Treesitter highlighting supports |conceal| via the `conceal` metadata. By
-convention, nodes to be concealed are captured as `@conceal`, but any capture
-can be used. For example, the following query can be used to hide code block
-delimiters in Markdown: >query
+Treesitter highlighting supports |conceal| via the `conceal` and `conceal_lines`
+metadata. By convention, nodes to be concealed are captured as `@conceal`, but
+any capture can be used. For example, the following query can be used to hide
+code block delimiters in Markdown: >query
 
     ((fenced_code_block_delimiter) @conceal (#set! conceal ""))
 <
@@ -506,7 +506,13 @@ still highlighted the same as other operators: >query
 
     "!=" @operator (#set! conceal "≠")
 <
-Conceals specified in this way respect 'conceallevel'.
+To conceal an entire line (do not draw it at all), a query with `conceal_lines`
+metadata can be used: >query
+
+    ((comment) @comment @spell
+      (#set! conceal_lines ""))
+<
+Conceals specified in this way respect 'conceallevel' and 'concealcursor'.
 
 Note that although you can use any string for `conceal`, only the first
 character will be used: >query
--- a/runtime/lua/vim/treesitter/highlighter.lua
+++ b/runtime/lua/vim/treesitter/highlighter.lua
@@ -67,6 +67,8 @@ end
 --- This state is kept during rendering across each line update.
 ---@field private _highlight_states vim.treesitter.highlighter.State[]
 ---@field private _queries table<string,vim.treesitter.highlighter.Query>
+---@field  _conceal_line boolean?
+---@field  _conceal_checked table<integer, boolean>
 ---@field tree vim.treesitter.LanguageTree
 ---@field private redraw_count integer
 ---@field parsing boolean true if we are parsing asynchronously
@@ -108,6 +110,11 @@ function TSHighlighter.new(tree, opts)
         self:on_changedtree(t:included_ranges(true))
       end)
     end,
+    on_child_added = function(child)
+      child:for_each_tree(function(t)
+        self._conceal_line = self._conceal_line or self:get_query(t:lang()):query().has_conceal_line
+      end)
+    end,
   }, true)
 
   local source = tree:source()
@@ -115,7 +122,7 @@ function TSHighlighter.new(tree, opts)
 
   self.bufnr = source
   self.redraw_count = 0
-  self._highlight_states = {}
+  self._conceal_checked = {}
   self._queries = {}
 
   -- Queries for a specific language can be overridden by a custom
@@ -123,8 +130,10 @@ function TSHighlighter.new(tree, opts)
   if opts.queries then
     for lang, query_string in pairs(opts.queries) do
       self._queries[lang] = TSHighlighterQuery.new(lang, query_string)
+      self._conceal_line = self._conceal_line or self._queries[lang]:query().has_conceal_line
     end
   end
+  self._conceal_line = self._conceal_line or self:get_query(tree:lang()):query().has_conceal_line
 
   self.orig_spelloptions = vim.bo[self.bufnr].spelloptions
 
@@ -141,7 +150,7 @@ function TSHighlighter.new(tree, opts)
   -- immediately afterwards will not error.
   if vim.g.syntax_on ~= 1 then
     vim.cmd.runtime({ 'syntax/synload.vim', bang = true })
-    vim.api.nvim_create_augroup('syntaxset', { clear = false })
+    api.nvim_create_augroup('syntaxset', { clear = false })
   end
 
   vim._with({ buf = self.bufnr }, function()
@@ -159,6 +168,7 @@ function TSHighlighter:destroy()
   if api.nvim_buf_is_loaded(self.bufnr) then
     vim.bo[self.bufnr].spelloptions = self.orig_spelloptions
     vim.b[self.bufnr].ts_highlight = nil
+    api.nvim_buf_clear_namespace(self.bufnr, ns, 0, -1)
     if vim.g.syntax_on == 1 then
       api.nvim_exec_autocmds(
         'FileType',
@@ -187,10 +197,9 @@ function TSHighlighter:prepare_highlight_states(srow, erow)
       return
     end
 
-    local highlighter_query = self:get_query(tree:lang())
-
+    local hl_query = self:get_query(tree:lang())
     -- Some injected languages may not have highlight queries.
-    if not highlighter_query:query() then
+    if not hl_query:query() then
       return
     end
 
@@ -200,7 +209,7 @@ function TSHighlighter:prepare_highlight_states(srow, erow)
       tstree = tstree,
       next_row = 0,
       iter = nil,
-      highlighter_query = highlighter_query,
+      highlighter_query = hl_query,
     })
   end)
 end
@@ -222,7 +231,15 @@ end
 ---@param changes Range6[]
 function TSHighlighter:on_changedtree(changes)
   for _, ch in ipairs(changes) do
-    api.nvim__redraw({ buf = self.bufnr, range = { ch[1], ch[4] + 1 }, flush = false })
+    api.nvim__redraw({ buf = self.bufnr, range = { ch[1], ch[4] }, flush = false })
+    -- Only invalidate the _conceal_checked range if _conceal_line is set and
+    -- ch[4] is not UINT32_MAX (empty range on first changedtree).
+    if ch[4] == 2 ^ 32 - 1 then
+      self._conceal_checked = {}
+    end
+    for i = ch[1], self._conceal_line and ch[4] ~= 2 ^ 32 - 1 and ch[4] or 0 do
+      self._conceal_checked[i] = false
+    end
   end
 end
 
@@ -286,8 +303,10 @@ end
 ---@param self vim.treesitter.highlighter
 ---@param buf integer
 ---@param line integer
----@param is_spell_nav boolean
-local function on_line_impl(self, buf, line, is_spell_nav)
+---@param on_spell boolean
+---@param on_conceal boolean
+local function on_line_impl(self, buf, line, on_spell, on_conceal)
+  self._conceal_checked[line] = true
   self:for_each_highlight_state(function(state)
     local root_node = state.tstree:root()
     local root_start_row, _, root_end_row, _ = root_node:range()
@@ -335,7 +354,7 @@ local function on_line_impl(self, buf, line, is_spell_nav)
 
         local url = get_url(match, buf, capture, metadata)
 
-        if hl and end_row >= line and (not is_spell_nav or spell ~= nil) then
+        if hl and end_row >= line and not on_conceal and (not on_spell or spell ~= nil) then
           api.nvim_buf_set_extmark(buf, ns, start_row, start_col, {
             end_line = end_row,
             end_col = end_col,
@@ -345,6 +364,16 @@ local function on_line_impl(self, buf, line, is_spell_nav)
             conceal = conceal,
             spell = spell,
             url = url,
+          })
+        end
+
+        if
+          (metadata.conceal_lines or metadata[capture] and metadata[capture].conceal_lines)
+          and #api.nvim_buf_get_extmarks(buf, ns, { start_row, 0 }, { start_row, 0 }, {}) == 0
+        then
+          api.nvim_buf_set_extmark(buf, ns, start_row, 0, {
+            end_line = end_row,
+            conceal_lines = '',
           })
         end
       end
@@ -366,7 +395,7 @@ function TSHighlighter._on_line(_, _win, buf, line, _)
     return
   end
 
-  on_line_impl(self, buf, line, false)
+  on_line_impl(self, buf, line, false, false)
 end
 
 ---@private
@@ -385,9 +414,40 @@ function TSHighlighter._on_spell_nav(_, _, buf, srow, _, erow, _)
   self:prepare_highlight_states(srow, erow)
 
   for row = srow, erow do
-    on_line_impl(self, buf, row, true)
+    on_line_impl(self, buf, row, true, false)
   end
   self._highlight_states = highlight_states
+end
+
+---@private
+---@param buf integer
+---@param row integer
+function TSHighlighter._on_conceal_line(_, _, buf, row)
+  local self = TSHighlighter.active[buf]
+  if not self or not self._conceal_line or self._conceal_checked[row] then
+    return self and self._conceal_line
+  end
+
+  -- Do not affect potentially populated highlight state.
+  local highlight_states = self._highlight_states
+  self.tree:parse({ row, row })
+  self:prepare_highlight_states(row, row)
+  on_line_impl(self, buf, row, false, true)
+  self._highlight_states = highlight_states
+  return true
+end
+
+---@private
+--- Clear conceal_lines marks whenever we redraw for a buffer change. Marks are
+--- added back as either the _conceal_line or on_win callback comes across them.
+function TSHighlighter._on_buf(_, buf)
+  local self = TSHighlighter.active[buf]
+  if not self or not self._conceal_line then
+    return
+  end
+
+  api.nvim_buf_clear_namespace(buf, ns, 0, -1)
+  self._conceal_checked = {}
 end
 
 ---@private
@@ -413,7 +473,9 @@ end
 api.nvim_set_decoration_provider(ns, {
   on_win = TSHighlighter._on_win,
   on_line = TSHighlighter._on_line,
+  on_buf = TSHighlighter._on_buf,
   _on_spell_nav = TSHighlighter._on_spell_nav,
+  _on_conceal_line = TSHighlighter._on_conceal_line,
 })
 
 return TSHighlighter
--- a/runtime/queries/markdown/highlights.scm
+++ b/runtime/queries/markdown/highlights.scm
@@ -49,12 +49,14 @@
 
 (fenced_code_block
   (fenced_code_block_delimiter) @markup.raw.block
-  (#set! conceal ""))
+  (#set! conceal "")
+  (#set! conceal_lines ""))
 
 (fenced_code_block
   (info_string
     (language) @label
-    (#set! conceal "")))
+    (#set! conceal "")
+    (#set! conceal_lines "")))
 
 (link_destination) @markup.link.url
 
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -22,6 +22,7 @@
 #include "nvim/marktree.h"
 #include "nvim/memory.h"
 #include "nvim/memory_defs.h"
+#include "nvim/move.h"
 #include "nvim/option_vars.h"
 #include "nvim/pos_defs.h"
 #include "nvim/sign.h"
@@ -122,6 +123,15 @@ void decor_redraw_sh(buf_T *buf, int row1, int row2, DecorSignHighlight sh)
       || (sh.flags & (kSHIsSign | kSHSpellOn | kSHSpellOff | kSHConceal))) {
     if (row2 >= row1) {
       redraw_buf_range_later(buf, row1 + 1, row2 + 1);
+    }
+  }
+  if (sh.flags & kSHConcealLines) {
+    FOR_ALL_WINDOWS_IN_TAB(wp, curtab) {
+      // TODO(luukvbaal): redraw only unconcealed lines, and scroll lines below
+      // it up or down. Also when opening/closing a fold.
+      if (wp->w_buffer == buf) {
+        changed_window_setting(wp);
+      }
     }
   }
   if (sh.flags & kSHUIWatched) {
@@ -830,6 +840,56 @@ next_mark:
   return attr;
 }
 
+static const uint32_t conceal_filter[kMTMetaCount] = {[kMTMetaConcealLines] = kMTFilterSelect };
+
+/// Called by draw, move and plines code to determine whether a line is concealed.
+/// Scans the marktree for conceal_line marks on "row", if necessary.
+///
+/// @param check_cursor If true, avoid an early return for an unconcealed cursorline.
+///                     Depending on the callsite, we still want to know whether the
+///                     cursor line would be concealed if it was not the cursorline.
+///
+/// @return whether "row" is concealed
+bool decor_conceal_line(win_T *wp, int row, bool check_cursor)
+{
+  if (wp->w_p_cole < 2 || !buf_meta_total(wp->w_buffer, kMTMetaConcealLines)
+      || (!check_cursor && wp == curwin && row + 1 == wp->w_cursor.lnum
+          && !conceal_cursor_line(wp))) {
+    return false;
+  }
+
+  // Scan the marktree for any conceal_line marks on this row.
+  MTPair pair;
+  MarkTreeIter itr[1];
+  marktree_itr_get_overlap(wp->w_buffer->b_marktree, row, 0, itr);
+  while (marktree_itr_step_overlap(wp->w_buffer->b_marktree, itr, &pair)) {
+    if (mt_conceal_lines(pair.start) && ns_in_win(pair.start.ns, wp)) {
+      return true;
+    }
+  }
+
+  marktree_itr_step_out_filter(wp->w_buffer->b_marktree, itr, conceal_filter);
+
+  while (itr->x) {
+    MTKey mark = marktree_itr_current(itr);
+    if (mark.pos.row > row) {
+      break;
+    }
+    if (mt_conceal_lines(mark) && ns_in_win(pair.start.ns, wp)) {
+      return true;
+    }
+    marktree_itr_next_filter(wp->w_buffer->b_marktree, itr, row + 1, 0, conceal_filter);
+  }
+  return false;
+}
+
+/// @return whether a window may have folded or concealed lines
+bool win_lines_concealed(win_T *wp)
+{
+  return hasAnyFolding(wp)
+         || (wp->w_p_cole >= 2 && buf_meta_total(wp->w_buffer, kMTMetaConcealLines));
+}
+
 int sign_item_cmp(const void *p1, const void *p2)
 {
   const SignItem *s1 = (SignItem *)p1;
@@ -850,8 +910,8 @@ int sign_item_cmp(const void *p1, const void *p2)
   return 0;
 }
 
-static const uint32_t sign_filter[4] = {[kMTMetaSignText] = kMTFilterSelect,
-                                        [kMTMetaSignHL] = kMTFilterSelect };
+static const uint32_t sign_filter[kMTMetaCount] = {[kMTMetaSignText] = kMTFilterSelect,
+                                                   [kMTMetaSignHL] = kMTFilterSelect };
 
 /// Return the signs and highest priority sign attributes on a row.
 ///
@@ -873,7 +933,7 @@ void decor_redraw_signs(win_T *wp, buf_T *buf, int row, SignTextAttrs sattrs[], 
   // TODO(bfredl): integrate with main decor loop.
   marktree_itr_get_overlap(buf->b_marktree, row, 0, itr);
   while (marktree_itr_step_overlap(buf->b_marktree, itr, &pair)) {
-    if (!mt_invalid(pair.start) && mt_decor_sign(pair.start)) {
+    if (!mt_invalid(pair.start) && mt_decor_sign(pair.start) && ns_in_win(pair.start.ns, wp)) {
       DecorSignHighlight *sh = decor_find_sign(mt_decor(pair.start));
       num_text += (sh->text[0] != NUL);
       kv_push(signs, ((SignItem){ sh, pair.start.id }));
@@ -940,7 +1000,7 @@ DecorSignHighlight *decor_find_sign(DecorInline decor)
   }
 }
 
-static const uint32_t signtext_filter[4] = {[kMTMetaSignText] = kMTFilterSelect };
+static const uint32_t signtext_filter[kMTMetaCount] = {[kMTMetaSignText] = kMTFilterSelect };
 
 /// Count the number of signs in a range after adding/removing a sign, or to
 /// (re-)initialize a range in "b_signcols.count".
@@ -1035,7 +1095,7 @@ bool decor_redraw_eol(win_T *wp, DecorState *state, int *eol_attr, int eol_col)
   return has_virt_pos;
 }
 
-static const uint32_t lines_filter[4] = {[kMTMetaLines] = kMTFilterSelect };
+static const uint32_t lines_filter[kMTMetaCount] = {[kMTMetaLines] = kMTFilterSelect };
 
 /// @param apply_folds Only count virtual lines that are not in folds.
 int decor_virt_lines(win_T *wp, int start_row, int end_row, int *num_below, VirtLines *lines,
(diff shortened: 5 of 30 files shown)


luukvbaal:
Have yet to figure out how to adapt `highlighter.lua` and markdown queries. It seems to work alright with
```lua
-- build/bin/nvim --clean -S test.lua
local ns = vim.api.nvim_create_namespace('')
vim.cmd.call('setline(1, range(1, 100))')
vim.api.nvim_buf_set_extmark(0, ns, 5, 0, { conceal_line = true })
vim.api.nvim_buf_set_extmark(0, ns, 6, 0, { conceal_line = true })
vim.o.conceallevel = 4
```

zeertzjq:
I think a single new `'conceallevel'` value isn't enough to control this. One may want `conceal_lines` to work in combination of any of the existing `'conceallevel'` values.

luukvbaal:
Ye wasn't sure about that yet either, was just following #25718 for now.

clason:
On the other hand, I don't think this (combining different levels) is common enough to warrant a new option. The main (and right now only) use for this is completely hiding code block markers in Markdown.

If you feel strongly about being able to hide only those lines, we could do the "bitflag" approach and add, say, 4 to each level to include vertical conceal (so 4 only hides lines, 7 is level 3 with hidden lines etc.)

zeertzjq:
> use for this is completely hiding code block markers in Markdown.

This requires a combination of `&conceallevel = 2`, which is already a combination that this PR currently doesn't support.

So why add a new level at all? Why not just add the concealing to existing levels?

luukvbaal:
Yeah going by the current help text for 'conceallevel', hiding concealed lines seems to fit fine with the current levels. Might be a bit surprising to have this enabled without user intervention but perhaps some kind of indicator on the next line similar to the 'smoothscroll' indicator can help with that?

clason:
> So why add a new level at all? Why not just add conceal_lines to existing levels?

That also works, and is probably the easiest for the MVP here.

src/nvim/decoration.c
```diff
@@ -702,8 +705,41 @@ int sign_item_cmp(const void *p1, const void *p2)
   return 0;
 }
 
-static const uint32_t sign_filter[4] = {[kMTMetaSignText] = kMTFilterSelect,
-                                        [kMTMetaSignHL] = kMTFilterSelect };
+static const uint32_t conceal_filter[kMTMetaCount] = {[kMTMetaConcealLine] = kMTFilterSelect };
+
+bool decor_conceal_line(win_T *wp, buf_T *buf, int row, bool moved)
```
bfredl, line 710: is `moved` for the cursor? perhaps `cursor_moved` is more explicit

src/nvim/plines.c
```diff
@@ -881,6 +885,10 @@ int plines_win_col(win_T *wp, linenr_T lnum, long column)
 int plines_win_full(win_T *wp, linenr_T lnum, linenr_T *const nextp, bool *const foldedp,
                     const bool cache, const bool limit_winheight)
 {
+  if (decor_conceal_line(wp, wp->w_buffer, lnum - 1, false)) {
+    return 0;
```
bfredl, line 889: callsites still expect `foldedp` to be set (I get UBSAN errors otherwise), could  use `if (foldedp) { foldedp = *false }`

src/nvim/decoration.c
luukvbaal: Yes but the argument only made sense for cursor movement to the next/previous line, so I removed it (and replaced it with separate calls).

zeertzjq (commented):
.

src/nvim/move.c
```diff
@@ -506,6 +508,14 @@ void check_cursor_moved(win_T *wp)
...
     wp->w_valid &= ~(VALID_WROW|VALID_WCOL|VALID_VIRTCOL
                      |VALID_CHEIGHT|VALID_CROW|VALID_TOPLINE);
+
+    // Concealed line visibility toggled.
+    if (wp->w_p_cole >= 2 && !conceal_cursor_line(wp)
+        && (decor_conceal_line(wp, wp->w_buffer, wp->w_cursor.lnum - 1)
+            || decor_conceal_line(wp, wp->w_buffer, wp->w_valid_cursor.lnum - 1))) {
+      // TODO(luukvbaal): shouldn't have to redraw entire window but scroll instead.
+      wp->w_lines_valid = 0;
+    }
```
zeertzjq, line 518: Does this work if you move it to `redraw_for_cursorline()`?

luukvbaal: Unfortunately it doesn't. I checked that before and debugged why it doesn't but forgot x)

runtime/doc/api.txt
```diff
@@ -2763,6 +2763,10 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
                     When a character is supplied it is used as |:syn-cchar|.
                     "hl_group" is used as highlight for the cchar if provided,
                     otherwise it defaults to |hl-Conceal|.
+                  • conceal_line: string which should be either empty or a
+                    single character. When empty, lines in the range are not
+                    drawn at all; the next unconcealed line is drawn instead.
+                    Supplying a character is currently unimplemented.
```
clason, line 2769:
```suggestion
                  • conceal_line: string (empty or a
                    single character). When provided, lines in the range are not
                    drawn at all; the next unconcealed line is drawn instead.
```
(Assuming that a non-empty string is simply ignored at the moment, meaning the behavior is the same.)

luukvbaal:
> Not sure if been discussed already but `conceal_lines` makes more sense to me than `conceal_line`.

Hasn't been discussed. Why? Because it will hide multiple lines if the range spans multiple lines? That's true for all flags, most of which aren't plural either. Keeping it singular seems more natural me, but I don't mind.

runtime/doc/api.txt
clason: > No objections, but I copied the first sentence from conceal verbatim.

That's fine then.

> If we keep that, your suggestion boils down to removing the mention of it being unimplemented?
Currently I also error when a single character is provided, could make it work like you say.

Yes, that was my suggestion.

clason:
`virt_lines`? (So +1 from me to Lewis' suggestion.)

luukvbaal:
`virt_lines` itself can take an array of lines so there the plural form makes sense. Setting a `conceal_lines` property for a singular `start_row` seems wrong to me. Setting `end_row` implies that all "singular-form" properties are applied to all lines in the range. Making the property itself plural seems off.

lewis6991:
> That's true for all flags, most of which aren't plural either. 

Which? From what I can tell, all flags are named appropriately for what they do.

clason:
> `virt_lines` itself can take an array of lines so there the plural form makes sense. Setting a `conceal_lines` property for a singular `start_row` seems wrong to me. Setting `end_row` implies that all "singular-form" properties are applied to all lines in the range. Making the property itself plural seems off.

Sorry, but `conceal_line` is not a singular property like `priority` (which would have to be called `char_priority` for that argument to hold). If a single `conceal_line` mark can skip multiple lines, it should be plural. (And the symmetry with `virt_lines` helps avoid needless mental typos.)

Not a deal-breaker, but to me the arguments are noticeably in favor of plural here.

lewis6991:
Likewise `nvim_buf_set/get_lines` can set/get a single line, but it is named with the plural because it can set/get multiple lines.

clason:
Could add a sentence or two to `:h treesitter-highlight-conceal` (doesn't need a separate paragraph or tag).

runtime/doc/treesitter.txt
```diff
@@ -490,6 +490,12 @@ still highlighted the same as other operators: >query
 
     "!=" @operator (#set! conceal "≠")
 <
+To conceal an entire line (next unconcealed line is drawn in it's stead), a
```
clason, line 493:
```suggestion
To conceal an entire line, a
```
I think that's clear enough in this context.

luukvbaal: IDK; in this context (without the knowledge of "vertical conceal"), "conceal an entire line" still reads as "conceal the entire line horizontally" to me. Should we refer to `nvim_buf_set_extmark()`'s properties above then?

clason (approved):

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -160,12 +160,39 @@ function TSHighlighter:destroy()
...
+    end
+
+    local range = vim.treesitter.get_range(node, buf, metadata and metadata[capture])
+    local start_row, _, end_row, _ = Range.unpack4(range)
+    -- The "conceal" attribute can be set at the pattern level or on a particular capture
+    if
+      capture and (metadata.conceal_lines or metadata[capture] and metadata[capture].conceal_lines)
+    then
+      api.nvim_buf_set_extmark(buf, ns, start_row, 0, { end_line = end_row, conceal_lines = '' })
+      marks_placed = true
```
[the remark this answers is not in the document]
luukvbaal: Yeah you're right. This is just to skip the clear_namespace API call though. Could remove it or make it buffer specific.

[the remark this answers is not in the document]
lewis6991: You mean unconditionally clear the namespace? That might be simpler. Shouldn't be expensive as the marktree will return early if there are no marks.

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -391,6 +420,7 @@ function TSHighlighter._on_win(_, _win, buf, topline, botline)
   if not self then
     return false
   end
+  vim.api.nvim_buf_clear_namespace(self.bufnr, ns, 0, -1)
```
lewis6991, line 423:
```suggestion
  api.nvim_buf_clear_namespace(self.bufnr, ns, 0, -1)
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -194,12 +218,17 @@ function TSHighlighter:prepare_highlight_states(srow, erow)
...
     -- for_each_tree traversal. This ensures that parents' highlight don't override children's.
-    table.insert(self._highlight_states, {
+    local state = {
       tstree = tstree,
       next_row = 0,
       iter = nil,
       highlighter_query = highlighter_query,
-    })
+    }
+    table.insert(self._highlight_states, state)
```
lewis6991, line 227: Is this change necessary?

luukvbaal: Not anymore.

src/nvim/decoration.c
```diff
@@ -983,6 +1023,10 @@ void decor_to_dict_legacy(Dict *dict, DecorInline decor, bool hl_name, Arena *ar
     PUT_C(*dict, "conceal", CSTR_TO_ARENA_OBJ(arena, buf));
   }
 
+  if (sh_hl.flags &kSHConcealLine) {
```
lewis6991, line 1026:
```suggestion
  if (sh_hl.flags & kSHConcealLine) {
```

src/nvim/plines.c
luukvbaal: We still need to update `nextp` as well so just moved it below the `hasFoldingWin()`.

luukvbaal:
We could instead have the on_win callback (that now already iterates from topline > botline), store captures/metadata per line, and have the on_line callback iterate that. Would that be a welcome refactor?

lewis6991:
I'm not sure what benefit that has over using the treesitter iterator. If anything wouldn't it be less performant due to the intermediate structure?

luukvbaal:
Benchmark confirms that it takes twice as long. So indeed most of the work seems to be done in the deduplicated function calls. Refactor also seems ~10% faster than on current HEAD, I would guess because it deduplicates some `root()` and `root():range()` calls as well.

luukvbaal:
So yeah it's about twice as fast, but it can't be stored per line, needs to be a flat list because otherwise multiline highlights are only applied to the start line. This doesn't seem to be tested (at least not in `treesitter/highlighter_spec.lua`). Even then, it's still around twice as fast as the current PR, and also ~10% faster than HEAD. 

So yeah it seems significant work is done in the de-duplicated functions. I'll add it in a separate commit (initially) to review.

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -324,29 +259,87 @@ local function on_line_impl(self, buf, line, is_spell_nav)
...
+
+---@param self vim.treesitter.highlighter
+---@param buf integer
+---@param line integer
+local function on_line_impl(self, buf, line)
+  if not self._highlight_states[line] then
+    return
+  end
+
+  for _, mark in ipairs(self._highlight_states[line]) do
```
lewis6991, line 329:
```suggestion
  for _, mark in ipairs(self._highlight_states[line] or {}) do
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -160,82 +156,13 @@ function TSHighlighter:destroy()
   if api.nvim_buf_is_loaded(self.bufnr) then
     vim.bo[self.bufnr].spelloptions = self.orig_spelloptions
     vim.b[self.bufnr].ts_highlight = nil
+    vim.api.nvim_buf_clear_namespace(self.bufnr, ns, 0, -1)
```
lewis6991, line 159:
```suggestion
    api.nvim_buf_clear_namespace(self.bufnr, ns, 0, -1)
```

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -276,42 +203,50 @@ local function get_spell(capture_name)
...
 
-    -- Only consider trees that contain this line
-    if root_start_row > line or root_end_row < line then
+    -- Only consider trees within the visible range
+    if root_start_row > erow or root_end_row < srow then
       return
     end
 
-    if state.iter == nil or state.next_row < line then
-      -- Mainly used to skip over folds
```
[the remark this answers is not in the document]
luukvbaal: It's not, don't think we can without querying folds ranges, or passing them along to on_win :/ If that perf gain outweighs it I guess it's a no-go.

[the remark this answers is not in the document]
lewis6991: It was pretty drastic from memory. You can probably git blame to find out the issue for more details. I think querying the folds wouldn't be too bad, I recently added that for semantic tokens.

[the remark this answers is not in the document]
luukvbaal: Restored the logic to skip over closed folds through `vim.fn.foldclosed(end)()`. Performance diff is negligible with HEAD on a 2k lines closed fold (without this logic it is indeed very slow). Still faster overall without folds.

Also restored your TODO note. If I understand what you wanted to do, that should be easier now that we use `foldclosedend()` anyways, if anything. Could take a look at it as well.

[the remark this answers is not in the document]
lewis6991: Well...if you're volunteering...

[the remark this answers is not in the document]
luukvbaal: > If anything wouldn't it be less performant due to the intermediate structure?

For transparency: this seems to be true in the case of many captures, I tested in a markdown file before.
With the C parser, in a large window, this refactor loses out to HEAD by ~5%.
So it depends on the situation, but I think it's the best we can do. Without the perf commit, this feature would introduce a significance performance hit.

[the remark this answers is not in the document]
lewis6991: Hmm ok. Generally we prefer the method that scales the best for larger files, even if it means making smaller files a bit slower.

vanaigr:
There's currently an issue with lines not being concealed if they weren't visible before other lines were concealed. I.e. for this [file.md](https://github.com/user-attachments/files/17966293/file.md), and this config:

```lua
vim.treesitter.start(0, 'markdown')
vim.cmd('set conceallevel=3')
vim.cmd('set number')
```

runnung `nvim --clean -u config.lua file.md`. 

On the first few redraws, lines at the bottom would not be concealed. Moving with `<C-d>` and `<C-u>` produces the same issues:

image


This happens because `srow` and `erow` are determined before some lines are concealed, and concealed lines below the window are not added even though they eventually become visible.

luukvbaal:
Thanks for noticing that, I'll look into it.

vanaigr:
Would it be possible to run the query cursor for the entire window only if the query has `conceal_lines` (similar to how #26827 [handles combined injections](https://github.com/neovim/neovim/blob/a092c61fd6b6f1b8cbd13bad7ada337786819a13/runtime/lua/vim/treesitter/query.lua#L35)) and `conceallevel` is 3, and keep the old highlighting logic?

Query cursor is the main bottleneck for long lines, and keeping the old logic would make it possible to replace full line highlighting with range highlighting.

If there's a slowdown from running the highlighting query twice, we can try moving everything that needs to run for the entire window into a separate query file to reduce the number of matches.

luukvbaal:
> If there's a slowdown from running the highlighting query twice, we can try moving everything that needs to run for the entire window into a separate query file to reduce the number of matches.

That seemed the most obvious solution to me as well and I proposed it in chat but @clason didn't like that.

lewis6991:
Nah, we should avoid that. If we need to partition them, then this should be done internally (somehow).

luukvbaal:
In addition to the new `_on_decor_line` decoration provider callback, I painstakingly went through all instances of `has(Any)Folding()` to make sure the logic handles concealed lines properly. Unless deemed necessary I don't think it will worth it to add tests for all the affected code paths...

clason:
What is the status here? (This is one of the main blockers for the 0.11 release.)

luukvbaal:
I've just been rebasing this until someone finds the time/courage to review this. I think this works well enough, though I believe scrolling doesn't always work as expected (with regards to the amount scrolled).

I did pose some questions in https://github.com/neovim/neovim/pull/31324#issuecomment-2498533444; I think the first and second behave in the least surprising way. 

> Do we need an indicator for (if the cursor is on) a concealed line (with set no(r)nu).

This may still be necessary but I held off on trying to implement it until I heard other opinions.

lewis6991:
If it doesn't conflict or regress with anything then I don't think this needs to be perfect.

lewis6991:
> this is necessary to fix a significant regression in 0.11.

Oh right. What regression is that? I can't see anything linked.

clason:
The concealed codeblock markers (which were stripped in a preprocessing step before we transitioned to relying on treesitter highlighting) lead to unacceptable whitespace. (I'm not sure there's a dedicated issue; I've been closing most complaints.)

clason:
> * What column should the cursor be at when it is on a concealed line (0 or actual `wcol` it would be at if it wasn't hidden).

I would say "actual" is least surprising. (It should make no difference on the concealed line, and moving to the next line would keep the cursor in the column.)

> * What should happen when the first line of a closed fold is concealed. For now I assume that would hide the line with foldtext and thus the entire fold.

I think that is acceptable. (Scenario: Folding codeblocks with concealed markers.) Can revisit as a "fix" after release if someone comes up with a better behavior.

> * Do we need an indicator for concealed lines (with `set no(r)nu`).

Not for the primary scenario (LSP hover/signature help), so let's punt on that.

src/nvim/decoration.c
```diff
@@ -850,8 +859,56 @@ int sign_item_cmp(const void *p1, const void *p2)
...
+  if (wp->w_p_cole < 2
+      || (!check_cursor && ((row + 1 == wp->w_cursor.lnum) && !conceal_cursor_line(wp)))) {
+    return false;
+  }
+
+  size_t keys = wp->w_buffer->b_marktree->n_keys;
+  if (!buf_meta_total(wp->w_buffer, kMTMetaConcealLines)) {
+    decor_providers_invoke_conceal_line(wp, row);
+    return wp->w_buffer->b_marktree->n_keys > keys;
+  }
```
bfredl, line 875: This assumes that there is only one source of concealed lines to be checked.

Which is fine for now as it is a private API, but maybe worth of a TODO comment.

src/nvim/decoration.c
```diff
@@ -124,6 +126,13 @@ void decor_redraw_sh(buf_T *buf, int row1, int row2, DecorSignHighlight sh)
       redraw_buf_range_later(buf, row1 + 1, row2 + 1);
     }
   }
+  if (sh.flags & kSHConcealLines) {
+    FOR_ALL_WINDOWS_IN_TAB(wp, curtab) {
+      if (wp->w_buffer == buf) {
+        changed_window_setting(wp);
```
bfredl, line 132: TODO comment: this forces a redraw off the entire window. we want something more fine grained later.

luukvbaal: Yep, also see https://github.com/neovim/neovim/pull/31324#discussion_r1857475348.

src/nvim/decoration.c
luukvbaal: How? I can't fully check/remember the code right now but doesn't the invoke call do the callback for all sources? The only assumption I see here is that those sources exclusively add `conceal_lines` marks.

bfredl: All sources might not add conceals at the same time. so if source A adds a conceal at line 1 and source B wants to add one at the call for line 10, that later call would be skipped. not a problem in the MVP but worthy of a TODO comment.

luukvbaal: Still don't see the problem; there is another decor provider conceal line invoke at line 899 below. This first one is only an extra early return if there are no conceal line marks in the buffer at all yet.

From memory it goes like this for every line that is checked:
1. Are there no conceal line marks? Then invoke callback and take increase in marks to mean the line is concealed.
2. If there are conceal line marks Check the mark tree for any marks on the checked line.
3. If no conceal line marks found invoke the callback and again take increase in marks to mean the line is concealed.

I'll check again when I get back home if I can understand what kind of TODO note is warranted, unless you can specify contrary to the above.

ribru17:
> If we want to be extra conservative, we could inject the query only when showing hover and signature help windows.

+1 for this, concealed lines for all markdown buffers in general seems a bit invasive

clason:
Our markdown queries are already tuned to this specific use case, so _if_ we do this, we should do it properly. But I don't think this is possible yet (buffer-local `vim.treesitter.query.set()`).

src/nvim/decoration.c
bfredl: So this entire block is rather an optimization as an early exit? then _that_ could be a good comment to write.

clason:
Question: What happens if both `conceal` and `conceal_line` are set on a line? (Relevant for backward-compatible markdown queries in nvim-treesitter.)

luukvbaal:
Well since we didn't introduce a new 'conceallevel', I think `conceal` will never actually take effect on a line that also has `conceal_line`. I'm not sure how that affects BWC, I guess it is advantageous?

clason:
Yes, if `conceal_line` takes priority over `conceal` and doesn't, say, error, that's exactly what I was hoping for.

luukvbaal:
Rebased and made some [changes]() to ensure the `_on_conceal_line` callback cost is only incurred for a window into a buffer where the highlighter may conceal lines.

src/nvim/decoration.c
luukvbaal: Added some comments in https://github.com/neovim/neovim/compare/23597cda3a1d230c8f1a6e0582968d0272778691..9b504a74e2127c3a528fb843e2f685d30d7d09bb.

seblyng:
I have been following this for a while and just decided to compile and test this since it seems to be in the review stage. I found an issue with this:

Minimal init.lua
```lua
	pattern = { "*" },
	callback = function(args)
		if not pcall(vim.treesitter.start, args.buf) then
			return
		end
	end,
})

	pattern = { "markdown" },
	callback = function()
		vim.opt.conceallevel = 2
	end,
})

vim.lsp.config("clangd", {
	cmd = { "clangd", "--clang-tidy" },
	root_markers = { ".clangd", "compile_commands.json" },
	filetypes = { "c" },
})

vim.lsp.enable("clangd")
```

foo.md:

````md
# This is a test

```c
int main() {
  int foo = 4;
}
```
````

main.c
```c
#include <stdio.h>

int main(void)
{
  printf("foo\n");
  return 1;
}
```

https://github.com/user-attachments/assets/1ba2e506-d051-4178-a669-42451dd6bad2

Notice that it doesn't seem to conceal before scrolling past the concealed lines

luukvbaal:
Thanks should be fixed. (The highlighter caches whether a line is concealed and relies on `on_changedtree` to tell it when that cache is invalidated. That logic didn't handle the first `on_changedtree` properly.)

seblyng:
Thanks! Seems to have fixed that specific issue. However, there still seems to be some things not working, and maybe especially for if it is the first line of the buffer (?)

Update init.lua

```lua
	pattern = { "*" },
	callback = function(args)
		if not pcall(vim.treesitter.start, args.buf) then
			return
		end
	end,
})

	pattern = { "markdown" },
	callback = function()
		vim.opt.conceallevel = 2
	end,
})

vim.lsp.config("clangd", {
	cmd = { "clangd", "--clang-tidy" },
	root_markers = { ".clangd", "compile_commands.json" },
	filetypes = { "c" },
})

vim.lsp.config("luals", {
	cmd = { "lua-language-server" },
	filetypes = { "lua" },
	root_markers = {
		".luarc.json",
		".luarc.jsonc",
		".luacheckrc",
		".stylua.toml",
		"stylua.toml",
		"selene.toml",
		"selene.yml",
	},
	settings = {
		Lua = {
			runtime = {
				version = "LuaJIT",
			},
			workspace = {
				checkThirdParty = false,
				library = {
					vim.env.VIMRUNTIME,
					"${3rd}/busted/library",
					"${3rd}/luv/library",
				},
			},
		},
	},
})

vim.lsp.enable("clangd")
vim.lsp.enable("luals")
```

https://github.com/user-attachments/assets/051520f3-d49d-412e-975d-d0531d42a984

NOTE: I wrote this before the two newest commits, but same behaviour with newest commit

luukvbaal:
> However, there still seems to be some things not working, and maybe especially for if it is the first line of the buffer

What are "some things"? I think the fact that the first line isn't hidden in the hover window is because the cursor is on that that line. Presumably it would be concealed if the LSP hover window added `vim.o.concealcursor = 'n'`. Though I think it _should_ be concealed if the hover is not yet the current window.

seblyng:
> What are "some things"? I think the fact that the first line isn't hidden in the hover window is because the cursor is on that that line. Presumably it would be concealed if the LSP hover window added vim.o.concealcursor = 'n'. Though I think it should be concealed if the hover is not yet the current window, that seems to be a bug.

Yes I totally agree. Maybe I was unclear. It is that it isn't concealed when the cursor is _not_ inside the popup that I consider a bug.

vanaigr:
> The highlighter caches whether a line is concealed and relies on `on_changedtree` to tell it when that cache is invalidated.

I think it should also check for text changes. The reason is the same as in https://github.com/neovim/neovim/pull/31975: if a pattern checks the text inside the node, whether it matches or not can change without changing the type of the node. And `on_changedtree` only gives ranges where types of nodes changed or nodes were moved.

Also does that mean that conceal patterns can only check nodes on the same line? If it includes a node from the line above, would it be removed when that line changes and it no longer matches?

luukvbaal:
The cache is also invalidated in `on_buf`.

> Also does that mean that conceal patterns can only check nodes on the same line? If it includes a node from the line above, would it be removed when that line changes and it no longer matches? (If this makes sense, I'm still reading the code)

I'm not well versed enough in TS lingo to even be sure what you're asking but I think that isn't handled by the highlighter? I just place conceal_lines marks whenever the query tells me to. If the text changes `on_buf` clears the namespace so it would be removed, and not placed again if it no longer matches I would assume.

clason:
Again, I remind that the primary use case for 0.11 is LSP hover and signature help, so there is no need for the MVP to be fully robust against all possible buffer changes; these can be deal with in (backported) fixes.

On the other hand, this:

> I think the fact that the first line isn't hidden in the hover window is because the cursor is on that that line. Presumably it would be concealed if the LSP hover window added vim.o.concealcursor = 'n'. Though I think it should be concealed if the hover is not yet the current window, that seems to be a bug.

should be handled (in a separate issue/PR).

seblyng:
Can confirm that the first line is now concealed if the cursor is not inside the popup after the latest commits👍 only tested this minimal but seems to work good and as expected now

clason:
With this PR? What happens with "normal" conceal (on `master`)?

luukvbaal:
Yeah need to adapt `util.lua:_make_floating_popup_size()`. Might be able to just replace it with with something akin to: open float and set its buffer text -> resize it to `api.nvim_win_text_height()`. I do the same for the ext_ui message box.

bfredl (approved):
I have mostly from the lower-level (ie marktree.c/screen.c ) perspective, but looks fine to me to merge soon.

src/nvim/marktree.c
```diff
@@ -1619,8 +1620,13 @@ bool marktree_itr_next_filter(MarkTree *b, MarkTreeIter *itr, int stop_row, int
   return marktree_itr_check_filter(b, itr, stop_row, stop_col, meta_filter);
 }
 
-const uint32_t meta_map[4] = { MT_FLAG_DECOR_VIRT_TEXT_INLINE, MT_FLAG_DECOR_VIRT_LINES,
-                               MT_FLAG_DECOR_SIGNHL, MT_FLAG_DECOR_SIGNTEXT };
+const uint32_t meta_map[kMTMetaCount] = {
```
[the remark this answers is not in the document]
luukvbaal: Fair. Being able to check `buf_meta_total()` for a specific kind is nice. Would need to be replaced by a full tree scan if we start merging kinds. Whether that weighs up to simd optimizations I couldn't say.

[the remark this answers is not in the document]
bfredl: > . Would need to be replaced by a full tree scan if we start merging metadata kinds.

No they won't. we would still use the metadata to limit the scan, it would just be slightly less precise (more false positives). These are all empirically chosen, false-positive-allowed filters to narrow down searches to start with.

runtime/queries/markdown/highlights.scm
```diff
@@ -49,12 +49,12 @@
 
 (fenced_code_block
   (fenced_code_block_delimiter) @markup.raw.block
-  (#set! conceal ""))
+  (#set! conceal_lines ""))
```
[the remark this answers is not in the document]
luukvbaal: I followed #25718 in this regard. Do you mean keep `conceal` _and_ add `conceal_lines` metadata? Yes I think we could do that. _Just_ keeping `conceal` might be problematic.

src/nvim/marktree.c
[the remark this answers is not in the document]
luukvbaal: Yes that's what I meant, _full_ scan being the wrong choice of words...

runtime/queries/markdown/highlights.scm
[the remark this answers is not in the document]
clason: > Do you mean keep conceal and add conceal_lines metadata? 

Yes, exactly (hence my earlier question about what happens when both metadata are set). 

Obviously we want to have `conceal_lines` in there, that's the whole point of this exercise ;)

runtime/doc/api.txt
```diff
@@ -2698,6 +2698,10 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
                     When a character is supplied it is used as |:syn-cchar|.
                     "hl_group" is used as highlight for the cchar if provided,
                     otherwise it defaults to |hl-Conceal|.
+                  • conceal_lines: string which should be either empty or a
+                    single character. When provided, lines in the range are
+                    not drawn at all; the next unconcealed line is drawn
+                    instead.
```
zeertzjq, line 2704: Mention `'conceallevel'` here?

src/nvim/drawscreen.c
```diff
@@ -1621,7 +1626,7 @@ static void win_update(win_T *wp)
       }
     }
 
-    if (mod_top != 0 && hasAnyFolding(wp)) {
+    if (mod_top != 0 && (win_lines_concealed(wp))) {
```
zeertzjq, line 1629:
```suggestion
    if (mod_top != 0 && win_lines_concealed(wp)) {
```

src/nvim/move.c
```diff
@@ -1504,10 +1508,8 @@ bool scrollup(win_T *wp, linenr_T line_count, bool byfold)
 
   check_topfill(wp, false);
 
-  if (hasAnyFolding(wp)) {
-    // Make sure w_topline is at the first of a sequence of folded lines.
-    hasFolding(wp, wp->w_topline, &wp->w_topline, NULL);
-  }
+  // Make sure w_topline is at the first of a sequence of folded lines.
+  hasFolding(wp, wp->w_topline, &wp->w_topline, NULL);
```
zeertzjq, line 1512: This change is a bit strange. Concealed lines don't affect `hasFolding()`, so why doesn't `hasFolding()` now have to be called even if there are no folds?

runtime/doc/api.txt
```diff
@@ -2698,6 +2698,10 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
                     When a character is supplied it is used as |:syn-cchar|.
                     "hl_group" is used as highlight for the cchar if provided,
                     otherwise it defaults to |hl-Conceal|.
+                  • conceal_lines: string which should be either empty or a
+                    single character. When provided, lines in the range are
```
zeertzjq, line 2702: Here the documentation mentions "string which should be either empty or a single character", but unlike `conceal` the API actually accepts any string without check for its content, which is a bit strange.

src/nvim/move.c
luukvbaal: It's been a while but I think I just removed it to match other callsites in this file. It's usually called unconditionally (it has this condition as an early return).

runtime/doc/api.txt
luukvbaal: Well the char part is as of yet unimplemented. The flag is a string to be future-proof towards that, while it is unimplemented it doesn't need to be checked in the API. But I guess there is no reason to already document the unimplemented behavior (although I do vaguely remember discussing that here or on matrix).

src/nvim/plines.c
```diff
@@ -893,6 +897,11 @@ int plines_win_full(win_T *wp, linenr_T lnum, linenr_T *const nextp, bool *const
   if (foldedp != NULL) {
     *foldedp = folded;
   }
+
+  if (decor_conceal_line(wp, lnum - 1, false)) {
+    return 0;
+  }
+
```
zeertzjq, line 904: I think this early return is wrong. When a line is concealed, there may still be virtual lines attached to the line above (with `virt_lines_above = false`).

luukvbaal: Hmm the early return maybe should be wrong but I don't think it currently is. I think because the line is concealed `virt_lines_above = false` for the above line will not be drawn currently?

zeertzjq: Oh, right. Add a TODO?

runtime/doc/api.txt
clason: I don't remember the outcome; there may have been concern that people already rely on the missing check and set multiple characters, which will make implementing the missing feature a breaking change.

Maybe we can add the check already even if it's not strictly necessary yet?

luukvbaal: I think that makes sense, done. (Though the current API check for `conceal` only checks that the first char is printable anyways, not that the string is a single character. No reason to stray from that for `conceal_lines` I guess.)

clason (approved):

src/nvim/plines.c
luukvbaal: Fixed in https://github.com/neovim/neovim/compare/e67e399ea01b3198070a821423e73eab6f89133a..d1b63fa0ed482479bd7add4cf2a94dd2d88d0145.

src/nvim/drawline.c
```diff
@@ -3036,8 +3037,8 @@ int win_line(win_T *wp, linenr_T lnum, int startrow, int endrow, int col_rows, s
       virt_line_index = -1;
       virt_line_flags = 0;
       // When the filler lines are actually below the last line of the
-      // file, don't draw the line itself, break here.
-      if (wlv.filler_todo == 0 && (wp->w_botfill || end_fill)) {
+      // file or line line is concealed, don't draw the line itself, break here.
```
zeertzjq, line 3040:
```suggestion
      // file or the line is concealed, don't draw the line itself, break here.
```

src/nvim/drawline.c
```diff
@@ -3036,8 +3037,8 @@ int win_line(win_T *wp, linenr_T lnum, int startrow, int endrow, int col_rows, s
...
+      if (wlv.filler_todo == 0 && (wp->w_botfill || end_fill || concealed)) {
```
zeertzjq, line 3041: It seems that `concealed` actually needs to be checked at almost every other place where `end_fill` is checked, except for the one in the `if (statuscol.draw)` branch.

zeertzjq (approved):

mathjiajia:
my latex conceal does not work after this

zeertzjq:
Please do not comment on the PR to report bugs. To report a bug, create an issue.

PR #31328, opened 2024-11-24
docs: misc

dundargoc:


Makefile
```diff
@@ -133,6 +133,9 @@ generated-sources benchmark $(FORMAT) $(LINT) $(TEST) doc: | build/.ran-cmake
 
 test: $(TEST)
 
+# iwyu-fix-includes can be downloaded from
+# https://github.com/include-what-you-use/include-what-you-use/blob/master/fix_includes.py.
+# Create a iwyu-fix-includes shell script in your $PATH that invokves the python script.
```
justinmk, line 138: 😆 
```suggestion
# Create a iwyu-fix-includes shell script in your $PATH that invokes the python script.
```

616b2f:
https://github.com/neovim/neovim/blob/d460928263d0ff53283f301dfcb85f5b6e17d2ac/runtime/lua/vim/lsp/diagnostic.lua#L211

a minor thing here, the parameter type should be `lsp.Diagnostics[]` at this line. Do you think this can be also handled in this PR?

dundargoc:
TODO:
- remove `unzip` from build requirements.
- spam getftype everywhere

runtime/lua/vim/lsp/diagnostic.lua
```diff
@@ -208,7 +208,7 @@ end
 
 --- @param uri string
 --- @param client_id? integer
---- @param diagnostics vim.Diagnostic[]
+--- @param diagnostics lsp.Diagnostic[]
```
clason, line 211: I don't think that's correct?

dundargoc: @616b2f

clason: (It's not correct)

clason: Sorry, didn't see this was under the lsp namespace. @goanders?

616b2f: This is used in `on_publish_diagnostics` function below which passes `diagnostics` property from `lsp.PublishDiagnosticsParams` class and there it has `lsp.Diagnostic` type. So it has to be here also `lsp.Diagnostics`.

PR #31337, opened 2024-11-24
fix(tui): only reset cursor color if it was changed

gpanders:
We already track this information so we might as well use it. This eliminates a bunch of unnecessary OSC sequences.


zeertzjq (commented):
It seems that the `cursor_color_changed` variable name is a bit confusing. Maybe it should be renamed to `cursor_has_color`?

PR #31338, opened 2024-11-24
docs(build): fix msys2 install example

axelhj:
I messed up with my previous PR #31312 the updated example correctly installs gcc for the UCRT environment but the build tools cannot be installed with that environment prefix (but pacman doesn't seem to error out). Sorry about the confusion.


zeertzjq:
Why is this in draft?

axelhj:
No reason. I think I was skimming the contribution guidelines a bit too quickly, the PR should be ready now.

dundargoc:

PR #5485, opened 2016-10-15
jobset(), jobget(). Deprecate termopen(), jobpid(), jobresize().

justinmk:
Nvim has `jobpid()` and `jobresize()`. Vim8 has `job_info()` and `job_status()` for no particular reason.  What's really wanted here is a way to get and set job properties. So let's do that.

Also deprecate `termopen()`, it is confusing and unnecessary: most people don't realize how similar it is to `jobstart()`. Nudging script/plugin authors towards using `jobstart()` to spawn a terminal makes it obvious that all of the relevant mechanisms are shared and available between the two.

To satisfy the purpose of `termopen()`, `jobstart()` needs a way to connect to a buffer. This is useful for all types of jobs (and Vim8 has a cloud of related options). 

- [ ] new job option: "buf" (or something like that, maybe "out_buf", "in_buf" like Vim8)
- [x] new job option: "type" = { rpc, pty, term } 
- [ ] ~~`jobset()`~~
- [x] 'jobget()`
    - [ ] `jobget(..., 'started')` start time
    - [ ] `jobget(..., 'started_by')` function/script that started the job
    - [ ] `jobget(..., 'stopped_by')` ? (Alternative: send this to `on_exit` handler ?)
- [ ] get a list of jobs (`jobget(0)` ?) https://github.com/neovim/neovim/issues/557
- [x] documentation
- [ ] tests
- [ ] mention "cwd" in error message https://github.com/neovim/neovim/commit/2c44d9257287c3e804dcc5a2bfe073c20554c0b9#commitcomment-19445677

PR #31343, opened 2024-11-25
feat(jobs): jobstart(…,{term=true}), deprecate termopen()

justinmk:
# Problem
`termopen` has long been a superficial wrapper around `jobstart`, and has no real purpose. Also, `vim.system` and `nvim_open_term` presumably will replace all features of `jobstart` and `termopen`, so centralizing the logic will help with that. cc @lewis6991 

From https://github.com/neovim/neovim/pull/5485 : 

> `termopen()` is misleading and unnecessary: most people don't realize how similar it is to `jobstart()`. Nudging script/plugin authors towards using `jobstart()` to spawn a terminal makes it obvious that all of the relevant mechanisms are shared and available between the two.

# Solution

- Introduce `eval/deprecated.c`, where all deprecated eval funcs will live.
- Introduce "term" flag of `jobstart`.
- Deprecate `termopen`.


runtime/doc/channel.txt
```diff
@@ -20,8 +20,8 @@ There are several ways to open a channel:
 
   2. Through stdin, stdout and stderr of a process spawned by |jobstart()|.
 
-  3. Through the PTY master end of a PTY opened with
-     `jobstart(..., {'pty': v:true})` or |termopen()|.
+  3. Through the PTY master end of a PTY opened with |nvim_open_term()| or
```
zeertzjq, line 23: `nvim_open_term()` does not open a PTY.

justinmk: Thanks, fixed in https://github.com/neovim/neovim/pull/31479

PR #31344, opened 2024-11-25
fix(lua): remove vim.loader.disable()

justinmk:
Problem:
`vim.loader.disable` does not conform to `:help dev-name-common` and `:help dev-patterns`.

Solution:
- Add `enable` parameter to `vim.loader.enable`
- Remove `vim.loader.disable`
- Note the change in `:help news-breaking-dev` (HEAD changes).
  - This is not a breaking change (except to "HEAD") because `vim.loader` is marked "experimental".

previous: 26765e8461c1ba1e9a351632212cf89900221781

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2496,7 +2496,7 @@ vim.loader.enable({enable})                              *vim.loader.enable()*
     • adds the libs loader
     • removes the default Nvim loader
 
-    Disable (`enable=true`):
+    Disable (`enable=false`):
     • removes the loaders
     • adds the default Nvim loader
 
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -2485,21 +2485,23 @@ vim.validate({name}, {value}, {validator}, {optional}, {message})
 ==============================================================================
 Lua module: vim.loader                                            *vim.loader*
 
-vim.loader.disable()                                    *vim.loader.disable()*
+vim.loader.enable({enable})                              *vim.loader.enable()*
     WARNING: This feature is experimental/unstable.
 
-    Disables the experimental Lua module loader:
-    • removes the loaders
-    • adds the default Nvim loader
+    Enables or disables the experimental Lua module loader:
 
-vim.loader.enable()                                      *vim.loader.enable()*
-    WARNING: This feature is experimental/unstable.
-
-    Enables the experimental Lua module loader:
-    • overrides loadfile
+    Enable (`enable=true`):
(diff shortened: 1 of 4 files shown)


gpanders:
Per your new proposed deprecation strategy, should we not keep a `vim.loader.disable` stub in `_deprecated.lua`? Otherwise this is a breaking change (with no `vim.deprecation` warning).

justinmk:
vim.loader is marked as experimental, so none of that is relevant.

gpanders:
Even so, it is a publicly exported symbol (no `_` prefix) and should go through the deprecation process, even if it is expedited (we do the same for tree-sitter, which is also marked experimental). Keeping a stub with `vim.deprecate` for a single release cycle is a small price to pay to minimize the impact of a breaking change.

justinmk:
Then we can't have experiments. Declaring something as an experiment implies that it will undergo breaking changes without a deprecation cycle, and this is a mechanism to reduce the overhead with introducing new modules. If experiments require a deprecation cycle then they are no different than non-experiments, and thus declaring them experimental is just a way to avoid the higher bar for non-experiments without actually committing to what an experiment implies (no deprecation cycle). I.e. "misleading", to put it generously.

clason:
> That prefix is for internal stuff.

No, we decided to use `__` for that so we can mark experimental stuff with `_` to make it crystal-clear. But `vim.loader` predates that decision.

But I agree with @gpanders that in this case, keeping a single-line stub (without documentation) is low-cost (and prevents high-cost community management (and keeps folke happy)).

justinmk:
Marking stuff "experimental" implies that it should not be blocked in code review to avoid one-way decisions. If experimental stuff requires deprecation cycles, then what is the practical difference vs non-experimental stuff? It essentially means nothing except a vague "this is messy" disclaimer.

gpanders:
>If experimental stuff requires deprecation cycles, then what is the practical difference vs non-experimental stuff?

We've since decided that experimental stuff will use the `_` prefix, which doesn't need a deprecation cycle since it's not a "public" symbol (by convention). Use of `_` prefixed symbols is a clear "use at your own risk" warning. `vim.loader` doesn't have the prefix because it predates that decision, so this is a special case.

clason:
It is, but as a courtesy for users since it wasn't marked as clearly as we now do, we could _still_ do a short deprecation cycle. (Especially against the backdrop of sentiment that we are currently breaking quite a few things -- not saying that we should stop development, but softening the blow where we can will send a signal of at least considering users.)

PR #31345, opened 2024-11-25
fix(lsp): retrigger diagnostics request on server cancellation

### LINKED ISSUE #30985: LSP: rust_analyzer: -32802: server cancelled the request

gpanders:
Resolves: https://github.com/neovim/neovim/issues/30985

Continuation of https://github.com/neovim/neovim/pull/30999 (credit to @Jesse-Bakker for most of the implementation).

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1532,12 +1532,13 @@ get_namespace({client_id}, {is_pull})
                      client. Defaults to push
 
                                           *vim.lsp.diagnostic.on_diagnostic()*
-on_diagnostic({_}, {result}, {ctx})
+on_diagnostic({error}, {result}, {ctx})
     |lsp-handler| for the method "textDocument/diagnostic"
 
     See |vim.diagnostic.config()| for configuration options.
 
     Parameters: ~
+      • {error}   (`lsp.ResponseError?`)
       • {result}  (`lsp.DocumentDiagnosticReport`)
       • {ctx}     (`lsp.HandlerContext`)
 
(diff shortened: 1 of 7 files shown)


mfussenegger:
See https://github.com/neovim/neovim/pull/30999#issuecomment-2448300472 I think the re-triggering aspect should be handled too - which might imply that the error needs to be surfaced to the request handler(?)

And having a test case would be great

mfussenegger (approved):
I think the test case might be easier to read if using the `create_server_definition` approach, but either way is fine.

emmanueltouzery:
I can't find evidence that this was backported to 0.10.x. I believe it should be, as rust is a popular language and its lsp is more or less unusable without this fix. I don't know how risky or not this change is, though.

gpanders:
Backport: https://github.com/neovim/neovim/pull/31427

PR #31347, opened 2024-11-25
fix(defaults): omit empty line from unimpaired mapping messages

marcuscaisey:
Problem:
The default unimpaired mappings display an empty line after the
command's output. This results (with default configuration) in the
`Press ENTER or type command to continue` prompt to be displayed, like
so:
image

Solution:
The cause is that we're checking the second return value from
`pcall(vim.api.nvim_cmd, opts, {})` to determine whether the call was
successful. `nvim_cmd` returns an empty string on success, so this value
is an empty string in the successful path which we then display.

The fix is simple: check the first return value instead which is the
"status code" of the call. 

After my change:
image


gpanders (approved):
Nice catch, thanks!

PR #31350, opened 2024-11-26
feat(tui): update 'background' on theme change events

### LINKED ISSUE #31294: Support theme change events
Spec: https://github.com/contour-terminal/contour/blob/f3c3334aa5c861348c5bbe8ffe572c872eef2e08/docs/vt-extensions/color-palette-update-notifications.md

Supporting terminals will send a CSI sequence when the operating system theme changes from light mode to dark mode or when the terminal emulator's background color changes. Nvim can detect this and update `'background'` accordingly. If the user's colorscheme automatically adapts based on the value of `'bg'` then this will update the user's colorscheme automatically.

gpanders:
Enabling private DEC mode 2031 tells the terminal to notify Nvim whenever the OS theme changes (i.e. light mode to dark mode or vice versa) or the terminal emulator's palette changes. When we receive one of these notifications we query the terminal color's background color again to see if it has changed and update the value of 'background' if it has.

We only do this though if the user has not explicitly set the value of 'bg' themselves. The help text is updated slightly to hint to users that they probably shouldn't set this value: on modern terminal emulators Nvim is able to completely determine this automatically.

Resolves: https://github.com/neovim/neovim/issues/31294

---

Demo (recorded on Ghostty):

https://github.com/user-attachments/assets/f18de29c-ed97-4bbc-8d7d-38aa12a2ec5f

### DIFF
--- a/src/nvim/tui/input.c
+++ b/src/nvim/tui/input.c
@@ -160,12 +160,16 @@ void tinput_init(TermInput *input, Loop *loop)
   // initialize a timer handle for handling ESC with libtermkey
   uv_timer_init(&loop->uv, &input->timer_handle);
   input->timer_handle.data = input;
+
+  uv_timer_init(&loop->uv, &input->bg_query_timer);
+  input->bg_query_timer.data = input;
 }
 
 void tinput_destroy(TermInput *input)
 {
   map_destroy(int, &kitty_key_map);
   uv_close((uv_handle_t *)&input->timer_handle, NULL);
+  uv_close((uv_handle_t *)&input->bg_query_timer, NULL);
   rstream_may_close(&input->read_stream);
   termkey_destroy(input->tk);
 }
@@ -179,6 +183,7 @@ void tinput_stop(TermInput *input)
 {
   rstream_stop(&input->read_stream);
   uv_timer_stop(&input->timer_handle);
+  uv_timer_stop(&input->bg_query_timer);
 }
 
 static void tinput_done_event(void **argv)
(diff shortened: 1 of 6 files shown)


will:
Built this branch, and the feature works great for me and my full config on ghostty 518f8b1048 on aarch64-darwin 15.1.1 for what it's worth.

src/nvim/tui/input.c
```diff
@@ -160,12 +160,16 @@ void tinput_init(TermInput *input, Loop *loop)
   // initialize a timer handle for handling ESC with libtermkey
   uv_timer_init(&loop->uv, &input->timer_handle);
   input->timer_handle.data = input;
+
+  uv_timer_init(&loop->uv, &input->bg_query_timer);
+  input->bg_query_timer.data = input;
```
justinmk, line 165: Guess we can't use the existing `timer_handle` because it has a different lifetime.

justinmk (approved):

gpanders:
We may want to eventually expose a `ThemeChanged` (or similar) autocommand event to allow plugins to hook into this, but since this PR doesn't expose any new APIs or user-facing options we're not committed to anything yet.

Terminal support for this feature is also still pretty rare (the only terminals I know of that support it are Ghostty and Contour).

justinmk:
> We may want to eventually expose a `ThemeChanged` (or similar) autocommand event to allow plugins to hook into this

`autocmd OptionSet background` ?

gpanders:
> > We may want to eventually expose a `ThemeChanged` (or similar) autocommand event to allow plugins to hook into this
> 
> `autocmd OptionSet background` ?

The (potential) `ThemeChanged` event would notify if the OS appearance changed, which is separate from the `'background'` option.

For instance, another way to implement this feature (if the `ThemeChanged` event existed) would be to add this autocommand:

```vim
au ThemeChanged * let &background=v:event.appearance
```

PR #31351, opened 2024-11-26
fix(lsp): vim.lsp.ClientConfig.commands type

snelling-a:
This PR fixes thet types of vim.lsp.ClientConfig commands

The current definition:

```lua
--- @field commands? table<string,fun(command: lsp.Command, ctx: table)>
```

fails validation when passed to `nvim-lspconfig` [here](https://github.com/neovim/nvim-lspconfig/blob/dafd61d6533bd90ecf6e2a3a972298fdddc74d82/lua/lspconfig/configs.lua?plain=1#L86-L93)

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1047,7 +1047,7 @@ Lua module: vim.lsp.client                                        *lsp-client*
       • {settings}?           (`table`) Map with language server specific
                               settings. See the {settings} in
                               |vim.lsp.Client|.
-      • {commands}?           (`table<string,fun(command: lsp.Command, ctx: table)>`)
+      • {commands}?           (`table<string, {[1]:function}|vim.api.keyset.user_command>`)
                               Table that maps string of clientside commands to
                               user-defined functions. Commands passed to
                               start_client take precedence over the global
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
This doesn't look right to me. lsp commands have nothing to do with user commands.

snelling-a:
> This doesn't look right to me. lsp commands have nothing to do with user commands.

@mfussenegger, i could be totally wrong, but given the docs:

```lua
--- Table that maps string of clientside commands to user-defined functions.
```

and the result of creating one of these commands:

```lua
commands = {
  TestLspCommand = {
    function()
      print("TestLspCommand")
    end,
    desc = "Test LSP Command",
  },
},
```

Screenshot 2024-11-26 at 12 25 54

it looks to me like this creates a user command.

i'm happy to hear otherwise and change it accordingly, but either way, the current type is incorrect

zeertzjq:
The user commands are created by nvim-lspconfig

mfussenegger:
Looks like the issue is that lspconfig defines `lspconfig.Config : vim.lsp.ClientConfig`

In the `vim.lsp.ClientConfig` the commands are for LSP client-commands. In lspconfig `commands` are for user commands and that clashes.
That needs fixing in lspconfig.

PR #31353, opened 2024-11-26
fix(lsp): improve LSP floating preview window cleanup

glepnir:
Problem: The current implementation creates a unique autocommand group for each floating preview window, which is inefficient and can lead to numerous autocommand groups.

Solution: Use a single shared autocommand group with improved window validation to properly clean up LSP floating preview windows.

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -1622,23 +1622,24 @@ function M.open_floating_preview(contents, syntax, opts)
       api.nvim_win_set_var(floating_winnr, opts.focus_id, bufnr)
     end
     api.nvim_buf_set_var(bufnr, 'lsp_floating_preview', floating_winnr)
+    api.nvim_win_set_var(floating_winnr, 'lsp_floating_bufnr', bufnr)
   end
 
-  local augroup_name = ('nvim.closing_floating_preview_%d'):format(floating_winnr)
-  local ok =
-    pcall(api.nvim_get_autocmds, { group = augroup_name, pattern = tostring(floating_winnr) })
-  if not ok then
-    api.nvim_create_autocmd('WinClosed', {
-      group = api.nvim_create_augroup(augroup_name, {}),
-      pattern = tostring(floating_winnr),
-      callback = function()
-        if api.nvim_buf_is_valid(bufnr) then
-          vim.b[bufnr].lsp_floating_preview = nil
-        end
-        api.nvim_del_augroup_by_name(augroup_name)
-      end,
-    })
-  end
+  api.nvim_create_autocmd('WinClosed', {
+    group = api.nvim_create_augroup('nvim.closing_floating_preview', { clear = true }),
+    callback = function(args)
+      local winid = tonumber(args.match)
+      local ok, preview_bufnr = pcall(api.nvim_win_get_var, winid, 'lsp_floating_bufnr')
+      if
+        ok
+        and api.nvim_buf_is_valid(preview_bufnr)
+        and winid == vim.b[preview_bufnr].lsp_floating_preview
+      then
+        vim.b[bufnr].lsp_floating_preview = nil
+        return true
+      end
+    end,
+  })
 
   vim.wo[floating_winnr].foldenable = false -- Disable folding.
   vim.wo[floating_winnr].wrap = opts.wrap -- Soft wrapping.
--- a/test/functional/plugin/lsp/utils_spec.lua
+++ b/test/functional/plugin/lsp/utils_spec.lua
@@ -267,38 +267,38 @@ describe('vim.lsp.util', function()
 
         eq(56, opts.height)
       end)
+    end)
+  end)
 
-      describe('vim.lsp.util.open_floating_preview', function()
-        local var_name = 'lsp_floating_preview'
-        local curbuf = api.nvim_get_current_buf()
+  describe('open_floating_preview', function()
+    before_each(function()
+      n.clear()
+      Screen.new(10, 10)
+      feed('9i<CR><Esc>G4k')
+    end)
 
-        it('clean bufvar after fclose', function()
-          exec_lua(function()
-            vim.lsp.util.open_floating_preview({ 'test' }, '', { height = 5, width = 2 })
-          end)
-          eq(true, api.nvim_win_is_valid(api.nvim_buf_get_var(curbuf, var_name)))
-          command('fclose')
-          eq(
-            'Key not found: lsp_floating_preview',
-            pcall_err(api.nvim_buf_get_var, curbuf, var_name)
-          )
-        end)
+    local var_name = 'lsp_floating_preview'
+    local curbuf = api.nvim_get_current_buf()
 
-        it('clean bufvar after CursorMoved', function()
-          local result = exec_lua(function()
-            vim.lsp.util.open_floating_preview({ 'test' }, '', { height = 5, width = 2 })
-            local winnr = vim.b[vim.api.nvim_get_current_buf()].lsp_floating_preview
-            local result = vim.api.nvim_win_is_valid(winnr)
-            vim.api.nvim_feedkeys(vim.keycode('G'), 'txn', false)
-            return result
-          end)
-          eq(true, result)
-          eq(
-            'Key not found: lsp_floating_preview',
-            pcall_err(api.nvim_buf_get_var, curbuf, var_name)
-          )
-        end)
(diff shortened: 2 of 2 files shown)


test/functional/plugin/lsp/utils_spec.lua
```diff
@@ -301,4 +301,36 @@ describe('vim.lsp.util', function()
       end)
     end)
   end)
+
+  describe('open_floating_preview', function()
```
zeertzjq, line 305: Why are the tests not removed from the old position?

glepnir: because before inside the `make_floating_popup_options` function describe block .

zeertzjq: So why aren't they removed?

runtime/lua/vim/lsp/util.lua
```diff
@@ -1632,6 +1632,7 @@ function M.open_floating_preview(contents, syntax, opts)
           vim.b[bufnr].lsp_floating_preview = nil
         end
         api.nvim_del_augroup_by_name(augroup_name)
+        return true
```
justinmk, line 1635: Why is the augroup needed if we are returning `true` here? I guess to avoid "duplicate" WinClosed events? But at least, `nvim_del_augroup_by_name` shouldn't be necessary because of the `if not ok` check on line 1626.

Should we instead create a single, global `WinClosed` handler that checks the win-id against `vim.b[bufnr].lsp_floating_preview` ?

glepnir: >  I guess to avoid "duplicate" WinClosed events?

yes because there can reuse a floating-window `  local floating_winnr = opts._update_win` in line 1531. 

Does the global WinClosed mean that winid is not needed as pattern? Any window closing will trigger it. Another thing we are missing is a function to check the existence of augroup or autocmd? The current method of checking the existence of augroup always uses pcall wrapper. Checking whether it has been registered an event requires `#nvim_get_autocmd({filter}) > 0`. It would be better if the filter supports `count` then return count number or something else 🤔

glepnir: Yes. exists can check everything on vim side. It would be better if we have a simple lua api to match it on autocmd level .  not a big problem.  nvm

runtime/lua/vim/lsp/util.lua
```diff
@@ -1620,18 +1620,14 @@ function M.open_floating_preview(contents, syntax, opts)
...
   end
 
-  local augroup_name = ('closing_floating_preview_%d'):format(floating_winnr)
-  local ok =
-    pcall(api.nvim_get_autocmds, { group = augroup_name, pattern = tostring(floating_winnr) })
-  if not ok then
+  local augroup_name = 'nvim_closing_floating_preview'
+  if vim.fn.exists(('#%s'):format(augroup_name)) == 0 then
     api.nvim_create_autocmd('WinClosed', {
       group = api.nvim_create_augroup(augroup_name, {}),
```
justinmk, line 1626: We could use `clear=true` here, which avoids all the `if exists(...)` code. But doesn't need to block this PR.

runtime/lua/vim/lsp/util.lua
```diff
@@ -1620,18 +1620,14 @@ function M.open_floating_preview(contents, syntax, opts)
...
-      pattern = tostring(floating_winnr),
-      callback = function()
-        if api.nvim_buf_is_valid(bufnr) then
+      callback = function(args)
+        if tonumber(args.match) == floating_winnr and api.nvim_buf_is_valid(bufnr) then
```
justinmk, line 1628: `floating_winnr` is bound to the closure when the autocmd was created. But the autocmd is long-lived. Can we check `vim.b[bufnr].lsp_floating_preview` instead?
```suggestion
        if tonumber(args.match) == vim.b[bufnr].lsp_floating_preview and api.nvim_buf_is_valid(bufnr) then
```

runtime/lua/vim/lsp/util.lua
glepnir: if i remember correctly default is clear =true

runtime/lua/vim/lsp/util.lua
glepnir: yes i also noticed it will make the variable move to heap..

justinmk: > and I think bufnr should not be also captured by closure either ? Can it be stored in w[floating_winnr].buf_hold_win?

oh, right. I thought that was an event arg. We should get it from event args, or `api.nvim_get_current_buf()` or whatever.

glepnir: > We should get it from event args, or api.nvim_get_current_buf() 

1. `args.buf` is current buffer in floating_winnr  
2.  if there trigger BufEnter which will close floating_winnr and get_current_buf will return a wrong buffer which maybe not hold the floating_winnr.

runtime/lua/vim/lsp/util.lua
```diff
@@ -1620,18 +1620,16 @@ function M.open_floating_preview(contents, syntax, opts)
...
-        if api.nvim_buf_is_valid(bufnr) then
+      group = api.nvim_create_augroup(augroup_name, { clear = true }),
+      callback = function(args)
+        if
+          tonumber(args.match) == vim.b[bufnr].lsp_floating_preview and api.nvim_buf_is_valid(bufnr)
+        then
           vim.b[bufnr].lsp_floating_preview = nil
         end
-        api.nvim_del_augroup_by_name(augroup_name)
       end,
```
justinmk, line 1633: what happened to the `return true` ?

runtime/lua/vim/lsp/util.lua
justinmk: either `vim.b[bufnr].lsp_floating_preview` is valid or not. why does any of that matter?

glepnir: > why does any of that matter?

we don't capture bufnr in closure ,then need to find bufnr which holds the floatinng_winnr. The event args or api.nvim_get_current_buf you proposed is not the previous bufnr in some cases, so this is invalid. Ineed to store this bufnr somewhere like win-var

justinmk: this is getting too complicated

runtime/lua/vim/lsp/util.lua
```diff
@@ -1620,18 +1620,16 @@ function M.open_floating_preview(contents, syntax, opts)
     api.nvim_buf_set_var(bufnr, 'lsp_floating_preview', floating_winnr)
   end
 
-  local augroup_name = ('closing_floating_preview_%d'):format(floating_winnr)
-  local ok =
-    pcall(api.nvim_get_autocmds, { group = augroup_name, pattern = tostring(floating_winnr) })
-  if not ok then
+  local augroup_name = 'nvim_closing_floating_preview'
```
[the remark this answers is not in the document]
glepnir: Okay, this will be clearer.

justinmk:
What's the status of this? Note that we use "dot" instead of underscore for augroup/namespace names.

glepnir:
just updated a way which not use closure

justinmk:
what is the updated problem/solution? the current pr title + description doesn't seem like the main problem that is being solved

runtime/lua/vim/lsp/util.lua
```diff
@@ -1622,20 +1622,24 @@ function M.open_floating_preview(contents, syntax, opts)
...
     api.nvim_buf_set_var(bufnr, 'lsp_floating_preview', floating_winnr)
+    api.nvim_win_set_var(floating_winnr, 'lsp_floating_bufnr', bufnr)
   end
 
-  local augroup_name = ('nvim.closing_floating_preview_%d'):format(floating_winnr)
-  local ok =
-    pcall(api.nvim_get_autocmds, { group = augroup_name, pattern = tostring(floating_winnr) })
-  if not ok then
+  local augroup_name = 'nvim.closing_floating_preview'
+  if vim.fn.exists(('#%s'):format(augroup_name)) == 0 then
```
justinmk, line 1629: is this check still needed , or can we just always recreate with `group = nvim_create_augroup(augroup_name, { clear = true })` ?

justinmk (approved):

runtime/lua/vim/lsp/util.lua
justinmk: But `clear = true` will clear the existing one, so what's the problem? It's worth the code simplification.

PR #31358, opened 2024-11-26
Fix test failures for release / tarball builds

jamessan:
Now that I've finally got the new dependencies packaged for 0.10.x, I've started running into test failures due to expectations of running with a debug build or from a git clone.

I'll adjust the CI to catch these problems and add fixes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -159,9 +159,15 @@ jobs:
           cmake --preset ci -D CMAKE_BUILD_TYPE=Debug -D CMAKE_INSTALL_PREFIX:PATH=$INSTALL_PREFIX ${{ matrix.build.flags }}
           cmake --build build
 
-      - name: ${{ matrix.test }}
+      - if: ${{ matrix.test == 'oldtest' }}
+        name: ${{ matrix.test }}
         timeout-minutes: 20
-        run: make ${{ matrix.test }}
+        run: make -C test/old/testdir NVIM_PRG=$(realpath build)/bin/nvim
+
+      - if: ${{ matrix.test != 'oldtest' }}
+        name: ${{ matrix.test }}
+        timeout-minutes: 20
+        run: cmake --build build --target ${{ matrix.test }}
 
       - name: Install
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -106,7 +106,7 @@ jobs:
           [
             { runner: ubuntu-24.04, os: ubuntu, flavor: asan, cc: clang, flags: -D ENABLE_ASAN_UBSAN=ON },
             { runner: ubuntu-24.04, os: ubuntu, flavor: tsan, cc: clang, flags: -D ENABLE_TSAN=ON },
-            { runner: ubuntu-24.04, os: ubuntu, cc: gcc },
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -145,6 +145,9 @@ jobs:
           sudo cpanm -n Neovim::Ext || cat "$HOME/.cpanm/build.log"
           perl -W -e 'use Neovim::Ext; print $Neovim::Ext::VERSION'
 
+      - name: Remove .git directory
+        run: cmake -E rm -rf -- .git
```
dundargoc, line 149: maybe just fix the three tests that fail instead of encoding this for eternity?

dundargoc, line 149: This is an expensive fix for a problem that probably won't happen again unless we create another root traversal function. I think a better tradeoff would just be to fix the tests and call it a day.

jamessan: I'm going to fix the tests. This just helps ensure there aren't other things relying on git (like the build part of the version inf). I'm not sure what's expensive about it. It takes less than a second in CI.

dundargoc: I meant more the maintaining this is expensive compared to the problem, which probably won't occur again.

.github/workflows/test.yml
```diff
@@ -145,14 +146,22 @@ jobs:
           sudo cpanm -n Neovim::Ext || cat "$HOME/.cpanm/build.log"
           perl -W -e 'use Neovim::Ext; print $Neovim::Ext::VERSION'
 
+      - if: ${{ matrix.build.flavor == 'release' }}
+        name: Setup Release build type
+        run: echo CMAKE_BUILD_TYPE=Release >> $GITHUB_ENV
```
dundargoc, line 151: Wut? Does adding `CMAKE_BUILD_TYPE=Release` to flags not work?

dundargoc: I checked locally and I can't reproduce. Have you tested it out?

dundargoc: We need to fix that then, this is pretty hacky. We can use this workaround for now in any case.

jamessan: We can touch the respective files earlier in the workflow. The root of the problem is we're using cmake to build and make to test.

dundargoc: Yes, that was mostly to save a ci step since we can't run the oldtests with cmake. I guess it's better to just run tests directly with cmake instead anyway?

.github/workflows/test.yml
```diff
@@ -194,7 +204,7 @@ jobs:
     uses: ./.github/workflows/test_windows.yml
 
   # This job tests the following things:
-  # - Check if Release, MinSizeRel and RelWithDebInfo compiles correctly.
+  # - Check if  MinSizeRel and RelWithDebInfo compiles correctly.
```
zeertzjq, line 207:
```suggestion
  # - Check if MinSizeRel and RelWithDebInfo compiles correctly.
```

jamessan:
N.B., this changes the name of the "ubuntu gcc ..." jobs (which are "required"), so that will need to be adjusted as part of this merge.

justinmk:
> this changes the name of the "ubuntu gcc ..." jobs (which are "required"), so that will need to be adjusted as part of this merge.

good call out , but I don't see any place where we actually have enumerated required status checks.

image

PR #31361, opened 2024-11-27
feat: add vim.fs.exists

dundargoc:
This is essentially upstreaming a modified version of util.path.exists
with the goal of upstreaming as much of non-configuration parts of
lspconfig to core as possible.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -234,6 +234,7 @@ LSP
 LUA
 
 • |vim.fs.rm()| can delete files and directories.
+• |vim.fs.exists()| can check whether file or directory exists.
 • |vim.validate()| now has a new signature which uses less tables,
   is more performant and easier to read.
 • |vim.str_byteindex()| and |vim.str_utfindex()| gained overload signatures
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -667,4 +667,14 @@ function M.rm(path, opts)
   end
 end
 
+--- @param filename string
+--- @return boolean
+function M.exists(filename)
+  local stat = uv.fs_stat(filename)
+  if stat then
+    return true
+  end
+  return false
+end
+
 return M


clason:
So the only difference to `uv.fs_stat` is that it returns `true` or `false` rather than table or `nil` and `err`?

dundargoc:
Idk, I think vim.fs should be a full module without needing to mix and match with vimscript and luv, it makes finding stuff harder if everything is scattered.

Also need a replacement suggestion for util.path.exists. If not this then I guess we just deprecated it as is and point them to uv.fs_stat?

clason:
> Also need a replacement suggestion for util.path.exists. If not this then I guess we just deprecated it as is and point them to uv.fs_stat?

Yes, that would be the default; it's a drop-in replacement (and used as such in our codebase and nvim-treesitter without issues).

> I think vim.fs should be a full module without needing to mix and match [...] makes finding stuff harder if everything is scattered.

That is a valid point of view, but afair we did (explicitly) not commit to that. 

In any case, that is a separate discussion; lspconfig can just yeet that function and switch to `if vim.uv.fs_stat('foo')` like everybody else, no _need_ to upstream anything.

clason:
Yes. But I don't think these are exported functions? This looks like an "upstreamed" version of plenary.path to me, which might have made sense at the time but not today.

Honestly, I would not recommend "working up" when upstreaming (I made that mistake with nvim-treesitter...); it'd be better to look at the user-facing and documented (which these aren't!) features and reimplement them from scratch (or deprecate without replacement). Once a feature is gone one way or another, we can just yeet all utility functions that are no longer used. Lather, rinse, repeat until we have reached the eternal sunshine of the spotless configs.

dundargoc:
> Yes. But I don't think these are exported functions? This looks like an "upstreamed" version of plenary.path to me, which might have made sense at the time but not today.

Unsure what you mean by exported. They are not local functions if that's what you mean, the code is misleading. [People use it in the wild](https://github.com/search?q=util.path.exists&type=code) and they will need to update the code if we remove it.

> Honestly, I would not recommend "working up" when upstreaming (I made that mistake with nvim-treesitter...); it'd be better to look at the user-facing and documented (which these aren't!) features and reimplement them in Neovim from scratch (or deprecate without replacement). Once a feature is gone one way or another, we can just yeet all utility functions that are no longer used. Lather, rinse, repeat until we have reached the eternal sunshine of the spotless configs.

OK, I'll try it.

justinmk:
We can at least add `:help fs-exists` or even `:help vim.fs.exists` which mentions the current "answer". It is valuable that users can reach for a single place to understand the recommended idioms.

dundargoc:
> We can at least add `:help fs-exists` or even `:help vim.fs.exists()` which mentions the current "answer". It is valuable that users can reach for a single place to understand the recommended idioms.

I don't mind this, but I'm wondering at that point if it's not more intuitive/useful to just provide the function instead of documenting it and then have the "bait and switch" with `vim.uv.fs_stat`?

dundargoc:
Deprecated in lspconfig in https://github.com/neovim/nvim-lspconfig/pull/3471.

justinmk:
> but I'm wondering at that point if it's not more intuitive/useful to just provide the function instead of documenting it and then have the "bait and switch" with `vim.uv.fs_stat`?

It's a two-way door. Starting with the doc hints is a cheap, incremental step which also "stops the bleeding" and buys time.

dundargoc:
OK, let's give it a shot: https://github.com/neovim/neovim/pull/31365

PR #31364, opened 2024-11-27
feat(vim.fs): find() can filter directories during search

mike325:


runtime/doc/lua.txt
```diff
@@ -3023,6 +3023,11 @@ vim.fs.find({names}, {opts})                                   *vim.fs.find()*
                  • {limit}? (`number`, default: `1`) Stop the search after
                    finding this many matches. Use `math.huge` to place no
                    limit on the number of matches.
+                 • {skip}? (`(fun(dir_name: string): boolean)?`) Do not
+                   traverse matching directories. If omitted, all directories
+                   are searched recursively.
+                 • {follow}? (`boolean`, default: `true`) Follow symbolic
```
justinmk, line 3029: This could be worth adding, since checking for a symlink may be more than a simple matter of a single `fs_stat` ?

runtime/lua/vim/fs.lua
```diff
@@ -308,6 +318,8 @@ function M.find(names, opts)
 
       for other, type_ in M.dir(dir) do
         local f = M.joinpath(dir, other)
+        local stat = vim.uv.fs_stat(f)
```
justinmk, line 321: this is a perf cost, which should only be paid if it's actually necessary (i.e. depending on the `follow` param)

runtime/doc/lua.txt
```diff
@@ -3023,6 +3023,11 @@ vim.fs.find({names}, {opts})                                   *vim.fs.find()*
                  • {limit}? (`number`, default: `1`) Stop the search after
                    finding this many matches. Use `math.huge` to place no
                    limit on the number of matches.
+                 • {skip}? (`(fun(dir_name: string): boolean)?`) Do not
+                   traverse matching directories. If omitted, all directories
```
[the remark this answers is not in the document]
mike325: not as it is now since name param just controls what gets found not whether or not to recursively search.

runtime/lua/vim/fs.lua
mike325: Ok, I'll change it

runtime/doc/lua.txt
mike325: it does probably need extra checks in windows, since I see the test failed there

runtime/lua/vim/fs.lua
```diff
@@ -190,6 +190,14 @@ end
 --- Use `math.huge` to place no limit on the number of matches.
 --- (default: `1`)
 --- @field limit? number
+---
+--- Do not traverse matching directories.
+--- If omitted, all directories are searched recursively.
+--- @field skip? (fun(dir_name: string): boolean)|nil
```
lewis6991, line 196:
```suggestion
--- @field skip? fun(dir: string): boolean
```

mike325:
I think the issue with windows might be cause because VM might not have the permissions create the link in the first place, but at the moment I don't have a windows system to confirm this.

runtime/doc/lua.txt
dundargoc: TIL about `getftype()`, maybe that works correctly. Might be worth testing it.

mike325:
Apparently the problem was indeed permissions related, not on creation but on read, `vim.uv.fs_stat(f)` was getting `EPERM` when the symlink was created without `junction=true` on windows, causing the test to fail.

runtime/doc/lua.txt
```diff
@@ -3039,6 +3039,11 @@ vim.fs.find({names}, {opts})                                   *vim.fs.find()*
                  • {limit}? (`number`, default: `1`) Stop the search after
                    finding this many matches. Use `math.huge` to place no
                    limit on the number of matches.
+                 • {skip}? (`fun(dir: string): boolean`) Do not traverse
```
justinmk, line 3042: please search for "predicate", "filter", "skip", and any other related concept, in the :help docs to see which is the most common. I'm wondering if positive logic ("match") or negative logic ("skip") is more common and more intuitive.

mike325: ### Skip
    - vim.fs.dir() - mentions both predicate and skip 
    - vim.iter:skip()
    - vim.iter:rskip()
    - search()
    - searchparipos()
    - searchpos()
    - strcharpart()
    - strchars()

### Filter
    - filter()
    - vim.inspect_pos()
    - vim.show_pos()
    - vim.tbl_filter()
    - vim.iter:filter()
    - vim.snippet.active()
    - vim.lsp.buf.code_action() - mentions both filter and predicate
    - vim.lsp.buf.format() - mentions both filter and predicate
    - vim.lsp.buf.rename() - mentions both filter and predicate
    - `filter` is mentions as prefer way to control `enable` functionality in dev-name-common and dev-api-patterns

### Match
    - string:match()
    - string:gmatch()
    - vim.lpeg.Pattern:match()
    - vim.filetype.match()
    - nvim_get_autocmds()
    - vim.regex:match_line()
    - vim.regex:match_str()
    - vim.diagnostic.match()
    - match() and match*() family

### Predicate
    - vim.fs.dir() - mentions both predicate and skip
    - vim.tbl_contains()
    - vim.tbl_filter()
    - vim.iter:all()
    - vim.iter:any()
    - vim.iter:find()
    - vim.iter:rfind()
    - vim.lsp.start() - reuse_client
    - vim.lsp.buf.code_action() - mentions both filter and predicate
    - vim.lsp.buf.format() - mentions both filter and predicate
    - vim.lsp.buf.rename() - mentions both filter and predicate
    - used by TS API

I might had missed some cases but I think this is a general overview of idioms.

it looks like `match` is by far the prefer idiom in the code base since it's used in a lot places to filter/narrow down in various scenarios, although is not so commonly used in as filter mechanism for functions/callbacks, `skip` is actually not that commonly used but is what I picked for `vim.fs.find` since I borrowed the concept from `vim.fs.dir`

justinmk: > I borrowed the concept from `vim.fs.dir`

Oh, I missed that an existing `vim.fs` function already has this concept. In that case this seems fine.

runtime/doc/lua.txt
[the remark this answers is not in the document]
gpanders: >@gpanders is names intended to provide "skip" functionality or do we need a new skip param as in this PR?

Was not intended (but was not considered). We could have the function return two values: the first if the item is a match, the second if the item should not be searched recursively.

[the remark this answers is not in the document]
mike325: > the doc for the `names` parameter says:
> 
> > If {names} is a function, it is called for each traversed item with args:
> > • name: base name of the current item
> > • path: full path of the current item The function should return `true` if the given item is considered a match.
> 
> 
> why can't this be used to skip anything that could conceivably be skipped by the new `skip` parameter in this PR?
> 
> @gpanders is `names` intended to provide "skip" functionality or do we need a new `skip` param as in this PR?

`The function should  return 'true' if the given item is considered a match.` is not suppose to be part of the `path` description, but the vimdoc script did not separate it, also as is now, `names` lazy evaluated the item when a type is provided, in the same example

```lua
-- get all files ending with .cpp or .hpp inside lib/
local cpp_hpp = vim.fs.find(function(name, path)
  return name:match('.*%.[ch]pp$') and path:match('[/\\\\]lib$')
end, {limit = math.huge, type = 'file'})
```
`names` will just evaluate files and wont even see the directories it traverse. it can be changed but in that case the evaluation will be always done for all items all time.

justinmk:
Please create a separate Pr for the "follow" paramter, so we can unblock it.

runtime/doc/lua.txt
```diff
@@ -3053,6 +3053,9 @@ vim.fs.find({names}, {opts})                                   *vim.fs.find()*
...
+                   matching directories. If omitted, all directories are
```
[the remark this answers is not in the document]
mike325: TBH I also think the semantics of using `skip()=false` to stop traversal is confusing.
 
Would it be better to use `filter` name and diverge from what `vim.fs.dir` does ?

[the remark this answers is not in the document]
justinmk: > Would it be better to use `filter` name and diverge from what `vim.fs.dir` does ?

Maybe if we also deprecate `dir({skip})` and add `dir({filter})`. WDYT @lewis6991 @gpanders ? Am I missing something in my impression that "skip=false is confusing"?

justinmk:
want to resolve https://github.com/neovim/neovim/pull/31364#discussion_r1916337288 before merging this.

runtime/doc/lua.txt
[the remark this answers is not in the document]
gpanders: I like `filter` over `skip`, though in https://github.com/neovim/neovim/pull/31364#discussion_r1868336855 I suggested avoiding adding a new parameter altogether by changing the semantics of the `names` parameter.

runtime/doc/lua.txt
```diff
@@ -3060,6 +3060,9 @@ vim.fs.find({names}, {opts})                                   *vim.fs.find()*
                    limit on the number of matches.
                  • {follow}? (`boolean`, default: `false`) Follow symbolic
                    links.
+                 • {filter}? (`fun(dir: string): boolean`) Traverse Matching
```
justinmk, line 3063: does returning `false` omit a directory (that matches the `vim.fs.dir` case)?
```suggestion
                 • {filter}? (`fun(dir: string): boolean`) Predicate that
                   decides if a directory is traversed. Return true to traverse
                   a directory, or false to skip. ...                   
```

runtime/doc/news.txt
```diff
@@ -338,6 +338,7 @@ LUA
 • |vim.fs.relpath()| gets relative path compared to base path.
 • |vim.fs.dir()| and |vim.fs.find()| can now follow symbolic links,
   the behavior can be turn on using the new `follow` option.
+• |vim.fs.find()| has an option to filter the directory traversal.
```
justinmk, line 341:
```suggestion
• |vim.fs.find()| accepts a `filter` predicate.
```

runtime/lua/vim/fs.lua
```diff
@@ -148,8 +148,17 @@ function M.dir(path, opts)
   vim.validate('path', path, 'string')
   vim.validate('depth', opts.depth, 'number', true)
   vim.validate('skip', opts.skip, 'function', true)
+  vim.validate('filter', opts.filter, 'function', true)
   vim.validate('follow', opts.follow, 'boolean', true)
 
+  if opts.skip ~= nil then
+    vim.deprecate('opts.skip', 'opts.filter', '0.13')
```
justinmk, line 155: This is one of those things that might just stick around. Don't need to rush its removal.

```suggestion
    vim.deprecate('opts.skip', 'opts.filter', '1.0')
```

runtime/lua/vim/fs.lua
```diff
@@ -148,8 +148,17 @@ function M.dir(path, opts)
...
+    if opts.filter == nil then
+      opts.filter = opts.skip
+      opts.skip = nil
+    end
```
justinmk, line 159:
```suggestion
    opts.filter = opts.filter or opts.skip
```

runtime/lua/vim/fs.lua
```diff
@@ -356,8 +370,10 @@ function M.find(names, opts)
         end
 
         if
-          type_ == 'directory'
-          or (type_ == 'link' and opts.follow and (vim.uv.fs_stat(f) or {}).type == 'directory')
+          (
+            type_ == 'directory'
+            or (type_ == 'link' and opts.follow and (vim.uv.fs_stat(f) or {}).type == 'directory')
+          ) and (not opts.filter or opts.filter(f) ~= false)
```
justinmk, line 376: So if filter returns `nil` that's considered truthy? Is that how `vim.fs.dir` works?

justinmk (approved):

runtime/lua/vim/fs.lua
mike325: This is the case now to keep the the same behavior of `skip` just with the new parameter, I can change filter to use falsey values instead but that would mean having a slightly more complex check in `vim.fs.dir`, now both functions would have different behaviors. 

How is it usually done in that case ? should the function error out if both `filter` and `skip` are provided or should it just perform both checks giving preference to `filter` ?

justinmk: > now both functions would have different behaviors.

whatever matches the current behavior of vim.fs.dir is preferred

mike325: ok then, let's keep the same behavior as `skip` at least for now,  let me know if more changes are require to move the PR forward

mike325:
any more thoughts on this @justinmk ?

justinmk:
spent lots of time reviewing this...

mike325:
if it's ok for you I'll open a new PR to clean up there changes and rebase to master

mike325:
continue in #39200

PR #31367, opened 2024-11-27
feat(diagnostic): Quickfix as builtin diagnostic handler

### LINKED ISSUE #31362: defaults: quickfix as builtin diagnostic handler
### Problem

The quickfix list is a very useful builtin feature, especially for showing diagnostics in your workspace / buffer. It however lacks auto-updating of these diagnostics. This is something that new users coming from an IDE (like vscode) might expect, as these IDEs often also have a window with diagnostics below the buffers. It makes the overall experience with using diagnostics in the qflist much better and gives working with diagnostics a more modern feel.

Implementing the qflist as diagnostic handler is fairly trivial thanks to `vim.diagnostic.handler`. However, users might not think to implement this themself. It would be great for discoverability to have this as builtin handler, even when it's turned off as default.

### Expected behavior

I currently use this in my own config:

```lua
vim.diagnostic.handlers["quickfix"] = {
  show = function(_, _, _, _)
    vim.diagnostic.setqflist({ open = false })
  end,

  hide = function(_, _)
    vim.fn.setqflist({}, 'r')
  end
}
```

and that works already quite well.

This solves the problem by creating a diagnostic handler that just sets the qflist with available diagnostics. I think the show function is called on DiagnosticChanged, thus this would update the qflist when the list of diagnostics changes. The cost is minimal, and provides a modern experience for working with diagnostics.

yochem:
Will close https://github.com/neovim/neovim/issues/31362.

This is a draft. I'll add/change the necessary documentation when the feature request gets more traction.

Since diagnostic handlers allow for namespace-specific diagnostic handling, it would be nice to 'tag' the qflist that is created with the namespace. This would also allow the `hide()` method to delete only the qflist from that namespace. See https://github.com/neovim/neovim/pull/30868.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -63,6 +63,10 @@ local M = {}
 --- (default: `true`)
 --- @field signs? boolean|vim.diagnostic.Opts.Signs|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Signs
 ---
+--- Use the quickfix list for diagnostics |quickfix|.
+--- (default: `true`)
+--- @field qflist? boolean|vim.diagnostic.Opts.Qflist|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Qflist
+---
 --- Options for floating windows. See |vim.diagnostic.Opts.Float|.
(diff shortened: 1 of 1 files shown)


yochem:
Closing this as it is added as example to the documentation instead. See https://github.com/neovim/neovim/issues/31362 for the discussion and https://github.com/neovim/neovim/pull/31371 for the documentation update.

PR #31370, opened 2024-11-28
vim-patch:9.1.{0867,0891,0896} 

### LINKED ISSUE #16128: after new refactor the select is wrong after leader add/delete
### Steps to reproduce

buffer with content `Foo Find five film` 
1:  `<C-x><C-P>`
2: intput F

Find not select 

this is a simple test case still have other cases i have find .  All the tests passed which led me to mistakenly believe that the refactoring was effective. In fact the test cases were lacking in many cases. I am trying to fix.

### Expected behaviour

work as before

### Version of Vim

build from source

### Environment

macos 

### Logs and stack traces

```shell
NAN
```

glepnir:
vim-patch:9.1.0867: ins_compl_add() has too many 

Problem:  ins_compl_add() has too many args
Solution: refactor it and use an int array instead of 2 separate int
          args (glepnir)

closes: https://github.com/vim/vim/pull/16062

https://github.com/vim/vim/commit/5c66e23c624717216d380d938d0bba5d34a004fe

patch 9.1.0891: building the completion list array is inefficient

Problem:  building the completion list array is inefficient
Solution: refactor and improve ins_compl_build_pum() func
          (glepnir)

current time complexity is O(n^2). I guess garray is not used here to save memory and avoid efficiency
is caused by heap memory allocation. A simple way is to add an extra pointer as a single linked list
to store the matching compl_T, and traverse this single linked list to generate compl_match_array.
The time complexity is O(n x m). The worst case is m=n, but we can still get a little improvement.
Because the if condition does not need to be run at one time. This should be a good solution for now.
Later we may be able to complete it in O(lgn) time. But this requires more reconstruction. So this is
the first step.

closes: https://github.com/neovim/neovim/pull/16125

https://github.com/vim/vim/commit/80b662009c0fe8f1728a3f3a2c8013b7eebf6745

https://github.com/vim/vim/pull/16129

vim-patch 9.1.0896: completion list wrong after v9.1.0891
Problem:  completion list wrong after v9.1.0891
Solution: update compl_mach_array after leader change
          (glepnir)

compl_shown_match update not correct after refactoring in v9.1.0891
Unfortunately, this regressed what item is selected after leader change.

So generate compl_match_array before updating compl_shown_match range,
and split generate compl_match_array into range match_head

fixes: https://github.com/vim/vim/issues/16128
closes: https://github.com/vim/vim/pull/16129

https://github.com/vim/vim/commit/a49c077a883b2566882df9069385ed1e1277ca64

### DIFF
--- a/test/old/testdir/test_popup.vim
+++ b/test/old/testdir/test_popup.vim
@@ -1675,4 +1675,29 @@ func Test_pum_completeitemalign()
   call StopVimInTerminal(buf)
 endfunc
 
+func Test_pum_keep_select()
+  CheckScreendump
+  let lines =<< trim END
+    set completeopt=menu,menuone,noinsert
+  END
+  call writefile(lines, 'Xscript', 'D')
+  let buf = RunVimInTerminal('-S Xscript', {})
+  call TermWait(buf)
+
+  call term_sendkeys(buf, "ggSFab\<CR>Five\<CR>find\<CR>film\<CR>\<C-X>\<C-P>")
+  call TermWait(buf, 50)
+  call VerifyScreenDump(buf, 'Test_pum_keep_select_01', {})
+  call term_sendkeys(buf, "\<C-E>\<Esc>")
+  call TermWait(buf, 50)
+
+  call term_sendkeys(buf, "S\<C-X>\<C-P>")
+  call TermWait(buf, 50)
+  call term_sendkeys(buf, "F")
+  call VerifyScreenDump(buf, 'Test_pum_keep_select_02', {})
+  call term_sendkeys(buf, "\<C-E>\<Esc>")
+
+  call TermWait(buf, 50)
+  call StopVimInTerminal(buf)
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab
--- a/src/nvim/insexpand.c
+++ b/src/nvim/insexpand.c
@@ -1176,6 +1176,20 @@ static int ins_compl_build_pum(void)
   bool compl_no_select = (cur_cot_flags & kOptCotFlagNoselect) != 0;
   bool compl_fuzzy_match = (cur_cot_flags & kOptCotFlagFuzzy) != 0;
   compl_T *match_head = NULL, *match_tail = NULL;
+
+  // If the current match is the original text don't find the first
+  // match after it, don't highlight anything.
+  bool shown_match_ok = match_at_original_text(compl_shown_match);
+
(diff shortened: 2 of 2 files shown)


zeertzjq (commented):
Missing comment about what `user_hl` does

zeertzjq (commented):
Test file update missing from the last commit

glepnir:
do we need add test case Test_pum_keep_select in popupmen_spec ?  there already have a similar case.

test/old/testdir/test_popup.vim
```diff
@@ -1675,4 +1675,26 @@ func Test_pum_completeitemalign()
...
 endfunc
 
+func Test_pum_keep_select()
+  CheckScreendump
+  let lines =<< trim END
+    set completeopt=menu,menuone,noinsert
+  END
+  call writefile(lines, 'Xscript', 'D')
+  let buf = RunVimInTerminal('-S Xscript', {})
+  call TermWait(buf)
```
zeertzjq, line 1685:
```suggestion
  call TermWait(buf)

```

test/old/testdir/test_popup.vim
```diff
@@ -1675,4 +1675,26 @@ func Test_pum_completeitemalign()
...
+  call term_sendkeys(buf, "ggSFab\<CR>Five\<CR>find\<CR>film\<CR>\<C-X>\<C-P>")
+  call TermWait(buf, 50)
+  call VerifyScreenDump(buf, 'Test_pum_keep_select_01', {})
+  call term_sendkeys(buf, "\<C-E>\<Esc>")
+  call TermWait(buf, 50)
```
zeertzjq, line 1690:
```suggestion
  call TermWait(buf, 50)

```

test/old/testdir/test_popup.vim
```diff
@@ -1675,4 +1675,26 @@ func Test_pum_completeitemalign()
...
+  call term_sendkeys(buf, "S\<C-X>\<C-P>")
+  call TermWait(buf, 50)
+  call term_sendkeys(buf, "F")
+  call VerifyScreenDump(buf, 'Test_pum_keep_select_02', {})
+  call term_sendkeys(buf, "\<C-E>\<Esc>")
```
zeertzjq, line 1695:
```suggestion
  call term_sendkeys(buf, "\<C-E>\<Esc>")

```

zeertzjq (approved):

PR #31371, opened 2024-11-28
docs(diagnostics): add quickfix handler example

yochem:
See https://github.com/neovim/neovim/issues/31362.

### DIFF
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -163,9 +163,9 @@ show a sign for the highest severity diagnostic on a given line: >lua
     }
 <
 
-                                          *diagnostic-handlers-example-loclist*
-To automatically update the |location-list| list with new diagnostics, we can
-create a new handler: >lua
+                                          *diagnostic-loclist-example*
+Whenever the |location-list| is opened, the following `show` handler will show
+the most recent diagnostics: >lua
 
   vim.diagnostic.handlers.loclist = {
     show = function(_, _, _, opts)
@@ -176,12 +176,8 @@ create a new handler: >lua
   }
 <
 
-The "hide" function can be omitted for this handler.
-
-Now, everytime the |location-list| is opened, it will show the most recent
-diagnostics. The handler accepts the same options as
-|vim.diagnostic.setloclist()| and can be configured using
-|vim.diagnostic.config()|: >lua
+The handler accepts the same options as |vim.diagnostic.setloclist()| and can be
+configured using |vim.diagnostic.config()|: >lua
 
   vim.diagnostic.config({
     loclist = {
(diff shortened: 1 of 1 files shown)


gpanders:
Diagnostic handlers are fired per-buffer, while the quickfix list is “global” (shared with all windows). It would make more sense for this to use the location list, which is scoped to a single window (thus allowing multiple windows containing different buffers to have their own diagnostics).

yochem:
That seems like a better approach indeed. Will change it.

gpanders (approved):
Thanks, this is a nice addition.

While testing this I found that if `open` is set to true, then the cursor automatically jumps into the location list after it is opened, which is extremely disruptive. In a follow up PR we should change the `open` parameter to open the quickfix/location list without moving the cursor into it.

EDIT: https://github.com/neovim/neovim/pull/31374

runtime/doc/diagnostic.txt
```diff
@@ -163,6 +163,37 @@ show a sign for the highest severity diagnostic on a given line: >lua
...
+    end
+  }
+<
+
+The "hide" function can be omitted for this handler.
+
+Now, everytime the |location-list| is opened, it will show the most recent
+diagnostics. The handler accepts the same options as
+|vim.diagnostic.setloclist()| and can be configured using
+|vim.diagnostic.config()|: >lua
```
justinmk, line 184: This is thoughtfully worded, but still way too verbose. Examples should be as small as possible, and only essential things should be noted. Otherwise it just gets messy and hard to read and maintain.

```suggestion
                                          *diagnostic-loclist-example*
Whenever the |location-list| is opened, the following `show` handler will show
the most recent diagnostics: >lua

  vim.diagnostic.handlers.loclist = {
    show = function(_, _, _, opts)
      -- Generally don't want it to open on every update
      opts.loclist.open = opts.loclist.open or false
      vim.diagnostic.setloclist(opts.loclist)
    end
  }
<

The handler accepts the same options as |vim.diagnostic.setloclist()| and can be
configured using |vim.diagnostic.config()|: >lua
```

runtime/doc/diagnostic.txt
```diff
@@ -163,6 +163,33 @@ show a sign for the highest severity diagnostic on a given line: >lua
...
+  vim.diagnostic.config({
+    loclist = {
+      open = true,
+      severity = { min = vim.diagnostic.severity.WARN },
+    }
+  })
+<
+
+This will open the location list window on every diagnostic change, and only
+list warnings and errors.
```
justinmk, line 191:
```suggestion
The handler accepts the same options as |vim.diagnostic.setloclist()| and can be
configured using |vim.diagnostic.config()|: >lua

  -- Open the location list on every diagnostic change (warnings/errors only).
  vim.diagnostic.config({
    loclist = {
      open = true,
      severity = { min = vim.diagnostic.severity.WARN },
    }
  })
<
```

justinmk (approved):

PR #31372, opened 2024-11-28
feat(options): add support for multitype options

famiu:
Problem: We don't support multitype options in the codebase.

Solution: Add preliminary support for options with multiple types, this will allow converting the callback options (`'*func'`, `'*expr'` options) to use actual Lua functions under the hood, as well as allow converting the list options to Array / Dictionary options.

Depends on #31405.


famiu:
Closing since it was decided that true multitype options won't be necessary, and that options should have one canonical type

PR #31374, opened 2024-11-28
fix(diagnostic): preserve cursor position when opening quickfix list

gpanders:
When diagnostics are set in the quickfix or location list with vim.diagnostic.setqflist or vim.diagnostic.setloclist, the cursor position should be preserved instead of jumping into the list's window.

---

Ref: https://github.com/neovim/neovim/pull/31371#pullrequestreview-2468606640

Should this be configurable? Leaning towards making it not configurable for now and we can add a knob if/when someone asks for it. Open to other opinions though.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -853,7 +853,9 @@ local function set_list(loclist, opts)
     vim.fn.setqflist({}, ' ', { title = title, items = items })
   end
   if open then
+    local winid = api.nvim_get_current_win()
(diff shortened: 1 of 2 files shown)


yochem:
I'm not sure about not making it configurable. I think it's a sensible thing to expect when you just run `setqflist()` on its own. It's also normal vim behaviour to switch to a window when it's opened. 

I could also update the loclist handler example in the docs (https://github.com/neovim/neovim/pull/31371) to have it switch back to the current window, as that is probably the usecase where the cursor switching is not desired.

justinmk (approved):

PR #31381, opened 2024-11-29
fix(lsp): completion insertion position bug

ddad431:
Problem: Currently, the completion result insertion position depends on
either the `word_boundary` or the `server_start_boundary` from **the last
server response**. When last `server_start_boundary` is nil , the insert position
**only** relies on `word_boundary` rather than `word_boundary` or the
previously response obtained `server_start_boundary`. As a result, the insert
position may be incorrect.

Solution: Update `server_start_boundary` only when the server responses with
match list.

Related: #31382

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -468,8 +468,7 @@ local function trigger(bufnr, clients)
         Context.isIncomplete = Context.isIncomplete or result.isIncomplete
         local client = lsp.get_client_by_id(client_id)
         local encoding = client and client.offset_encoding or 'utf-16'
-        local client_matches
-        client_matches, server_start_boundary = M._convert_results(
+        local client_matches, boundary = M._convert_results(
           line,
           cursor_row - 1,
           cursor_col,
@@ -479,7 +478,10 @@ local function trigger(bufnr, clients)
           result,
           encoding
         )
-        vim.list_extend(matches, client_matches)
+        if next(client_matches) then
+          vim.list_extend(matches, client_matches)
+          server_start_boundary = boundary
+        end
       end
     end
     local start_col = (server_start_boundary or word_boundary) + 1


MariaSolOs:
@ddad431 Thanks for the PR <3!
For reviewing purposes, did you encounter a bug with a specific LSP server that motivated this contribution? If so, could you please add a test that minimally reproduces the scenario so that we prevent regressions?

ddad431:
@MariaSolOs The problem happend on the buffer need to attched multi-server.(for my case, it occurs on volar hybrid mode which need volar and tsssever two language.

I also want add test. But I don't know how to imitate mult-server env.

MariaSolOs:
> @MariaSolOs The problem happend on the buffer need to attched multi-server.(for my case, it occurs on volar hybrid mode which need volar and tsssever two language server.
> 
> I also want add test. But I don't know how to imitate mult-server env.

Hmm. In that case could you start with a minimal reproducible example and create an issue to link to this change?

I know that it's a pain, but I just want to understand the problem that this is trying to fix. If reproducing the bug requires a complex setup, I'm more inclined to believe that this is a server issue and not an editor problem.

ddad431:
> > @MariaSolOs The problem happend on the buffer need to attched multi-server.(for my case, it occurs on volar hybrid mode which need volar and tsssever two language server.问题发生在缓冲区需要附加多服务器上。（就我而言，它发生在需要 Volar 和 TSSSEver 两种语言服务器的 Volar 混合模式下。
> > I also want add test. But I don't know how to imitate mult-server env.我还想要添加测试。但是我不知道如何模仿 mult-server env。
> 
> Hmm. In that case could you start with a minimal reproducible example and create an issue to link to this change?嗯。在这种情况下，您能否从一个最小可重现的示例开始，并创建一个问题来链接到此更改？
> 
> I know that it's a pain, but I just want to understand the problem that this is trying to fix. If reproducing the bug requires a complex setup, I'm more inclined to believe that this is a server issue and not an editor problem.我知道这很痛苦，但我只是想了解它试图解决的问题。如果重现错误需要一个复杂的设置，我更倾向于相信这是一个服务器问题，而不是编辑器问题。

Yes, wait a minute.

ddad431:
@MariaSolOs I have link to the issue.

I am a tyro and my english is poor. I don't know if I expressed it clearly
enough. So I give one clearly version: 

For `[].<C-x><C-o>`, the source code do:

1. `word_boundary` is 3
2. Two server(A, B) contencted one buffer
3. For completion, send requests to (A, B)
4. Only one server has matched list
    - A has, update `server_start_boundary` to 2 (match item is `.contact`)
    - B is empty, update `server_start_boundary` to nil
5. `server_start_boundary` is nil, so use `word_boundary(3)`
7. start insert pos is 4 (`word_boundary` + 1)
8. `[].(4)` + `.contact` = `[]..contact`

The correct doing:
- Only one server has matched list.
    - A has, update `server_start_boundary` to 2  (match item is `.contact`)
    - B is empty, don't update `server_start_boundary`
- `server_start_boundary` is 2, so use `server_start_boundary(2)`
- start insert pos is 3 (`word_boundary` + 1)
- `[](3)` + `.contact` = `[].contact`

hrsh7th:
I'm not good at English, but I'll try to explain this area.

1. The default completion start position may be different for each source.

- Plugins like nvim-cmp have keyword_pattern for each source. (VSCode does the same.)

- In such cases, the "completion start position" may be different for each source.

- If neovim itself refers to `'iskeyword'` on a per-buffer basis, it is per-buffer, so you don't need to consider this.

2. The completion start position may be different for each completion item.

- It is mainly determined by textEdit.start.character.

3. Use the minimum completion start position for vim's native completion menu.

- If the completion position in the completion menu is smaller than the completion position of the item, a character needs to be completed.

ddad431:
> I'm not good at English, but I'll try to explain this area.我不擅长英语，但我会尝试解释这个领域。
> 
> 1. The default completion start position may be different for each source.每个源的默认完成开始位置可能不同。
> 
> * Plugins like nvim-cmp have keyword_pattern for each source. (VSCode does the same.)像 nvim-cmp 这样的插件对每个源都有keyword_pattern。（VSCode 执行相同的操作。
> * In such cases, the "completion start position" may be different for each source.在这种情况下，每个源的“完成开始位置”可能不同。
> * If neovim itself refers to `'iskeyword'` on a per-buffer basis, it is per-buffer, so you don't need to consider this.如果 neovim 本身在每个缓冲区的基础上引用 `'iskeyword'`，那么它是每个缓冲区的，所以你不需要考虑这一点。
> 
> 2. The completion start position may be different for each completion item.每个完成项目的完成开始位置可能不同。
> 
> * It is mainly determined by textEdit.start.character.它主要由 textEdit.start.character 决定。
> 
> 3. Use the minimum completion start position for vim's native completion menu.使用 vim 的本机完成菜单的最小完成开始位置。
> 
> * If the completion position in the completion menu is smaller than the completion position of the item, a character needs to be completed.如果完成菜单中的完成位置小于项目的完成位置，则需要完成角色。

Thanks. Your `nvim-cmp` is the only one plugin(not only completion plugin) which I use. (But, I just can't lazyload it with `packadd` to reduce startuptime)

To tell the truth, I'm a programing tyro(8 months of front-end self-study) not vim(neovim) . This is my first pull request on GitHub. I can't fully understand what you're saying now, but I will find out carefully.

For this pull request, I took the easy way out. I just assume that there is only one completion source to get. Like above I said: "Only one server has matched list." I know my pull request is some problem. (e.g. when mount two server(lua, tssserver) in vue file, this change will still fail.) But there has reason.

I don't know the concept of `souce` which said. Is it like `lsp` source`, `buffer` source`? I don’t know if neovim currently does relevant work.  What I have learned so far(perhaps wrong): 
- For completion, the completion result come from language server. Currently, neo vim will request completion for every server which attach on current buffer.
- For completion, It should be based on language context（not code, not buffer）. This means that for one language context, it should get only one server result. Currently, most language server can't recognize language context, like above example which is  mount two server(lua, tsserver) in vue file, I can get lua completion source. But it doesn't make sense. This is why I create the request. For volar server, it only do recognize language context, this fit with "one language context, on language server result"

About another concept, I don't understand. But thanks for the information you gave.

PR #31390, opened 2024-11-29
fix(api): make `nvim_set_hl()` respect all `cterm` attributes

### LINKED ISSUE #31385: `nvim_set_hl()` does not set some highlight attributes in `cterm`
### Problem

Some new-ish highlight attributes are not set in `nvim_set_hl()` if supplied as part of `cterm`.

### Steps to reproduce

1. `nvim --clean`
2. `:lua vim.api.nvim_set_hl(0, 'AAA', { underdouble = true, cterm = { underdouble = true } })`.
3. `:hi AAA` shows only `gui=underdouble`. Same information in `:=vim.api.nvim_get_hl(0, { name = 'AAA' })`.

This happens for "underdouble", "underdotted", and "underdashed" attributes.

Note: setting attributes with `:hi AAA gui=underdouble cterm=underdouble` works as expected.

------

Here is the script used to check which attributes are not recognized:

```lua
local hl_attrs = {
  'bold',
  'underline',
  'undercurl',
  'underdouble',
  'underdotted',
  'underdashed',
  'strikethrough',
  'reverse',
  'italic',
  'standout',
  'altfont',
  'nocombine',
}

_G.no_cterm_attr = {}
for _, attr in ipairs(hl_attrs) do
  local hl_group = 'lua_' .. attr
  vim.api.nvim_set_hl(0, hl_group, { [attr] = true, cterm = { [attr] = true } })
  if vim.api.nvim_get_hl(0, { name = hl_group }).cterm == nil then table.insert(no_cterm_attr, attr) end
end

-- Execute this script and check `:=no_cterm_attr`
```

### Expected behavior

All highlight attributes can be set in `cterm` via `nvim_set_hl()`.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1252+gb1c907f219

### Operating system/version

### Terminal name/version

Ghostty 0.1.0-main+d47df211

xterm-ghostty

### Installation

appimage

echasnovski:
Fix #31385. Also added a test, but as it seems like a typo/undercopypaste, maybe not worth it?

### DIFF
--- a/src/nvim/highlight.c
+++ b/src/nvim/highlight.c
@@ -1100,6 +1100,9 @@ HlAttrs dict2hlattrs(Dict(highlight) *dict, bool use_rgb, int *link_id, Error *e
     CHECK_FLAG(cterm, cterm_mask, italic, , HL_ITALIC);
     CHECK_FLAG(cterm, cterm_mask, underline, , HL_UNDERLINE);
     CHECK_FLAG(cterm, cterm_mask, undercurl, , HL_UNDERCURL);
+    CHECK_FLAG(cterm, cterm_mask, underdouble, , HL_UNDERDOUBLE);
+    CHECK_FLAG(cterm, cterm_mask, underdotted, , HL_UNDERDOTTED);
+    CHECK_FLAG(cterm, cterm_mask, underdashed, , HL_UNDERDASHED);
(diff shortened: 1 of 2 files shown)


clason:
Are these attributes in Nvim 0.10 already (i.e., should we backport)?

echasnovski:
> Are these attributes in Nvim 0.10 already (i.e., should we backport)?

Yes.

PR #31391, opened 2024-11-29
docs: keybind suggestion for Inspect command

craigmac:


clason:
I think vimscript is left as an exercise to the reader for new docs; the Lua snippet is enough.

justinmk:
related: https://github.com/neovim/neovim/pull/31323

PR #31405, opened 2024-11-30
refactor(options): use `vim.inspect()` for converting OptVal to string

famiu:
Problem: Currently we have handcrafted code for converting an OptVal to its string representation. This may be fine for now, but it will be significantly tougher to implement once we add Array or Dictionary options. It also adds room for inconsistency between different methods of converting any object to a string.

Solution: Use `vim.inspect()` under the hood to convert an `OptVal` to a string. This ensures that the function is future-proofed for Array, Dictionary and Function option values, and also guarantees consistency with our Lua object-to-string conversion.

### DIFF
--- a/src/nvim/option.c
+++ b/src/nvim/option.c
@@ -3212,63 +3212,67 @@ static void set_option_varp(OptIndex opt_idx, void *varp, OptVal value, bool fre
   UNREACHABLE;
 }
 
-/// Return C-string representation of OptVal. Caller must free the returned C-string.
-static char *optval_to_cstr(OptVal o)
+/// Get C-string representation of OptVal. Uses vim.inspect() under the hood.
+/// Always returns a valid C-string.
+///
+/// @param  value  Option value.
+///
+/// @return  C-string representation of value.
+static const char *optval_to_cstr(OptVal value)
 {
-  switch (o.type) {
-  case kOptValTypeNil:
-    return xstrdup("");
-  case kOptValTypeBoolean:
-    return xstrdup(o.data.boolean ? "true" : "false");
-  case kOptValTypeNumber: {
-    char *buf = xmalloc(NUMBUFLEN);
-    snprintf(buf, NUMBUFLEN, "%" PRId64, o.data.number);
-    return buf;
+  Error err = ERROR_INIT;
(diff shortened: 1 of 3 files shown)


src/nvim/option.c
```diff
@@ -3212,63 +3212,67 @@ static void set_option_varp(OptIndex opt_idx, void *varp, OptVal value, bool fre
...
-    char *buf = xmalloc(o.data.string.size + 3);
-    snprintf(buf, o.data.string.size + 3, "\"%s\"", o.data.string.data);
-    return buf;
-  }
+  Error err = ERROR_INIT;
+  const char *res = nlua_inspect(optval_as_object(value), &err);
+
+  if (ERROR_SET(&err)) {
+    // This conversion should never fail unless something is seriously wrong.
+    abort();
```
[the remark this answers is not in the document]
justinmk: No

src/nvim/option.c
```diff
@@ -3212,63 +3212,67 @@ static void set_option_varp(OptIndex opt_idx, void *varp, OptVal value, bool fre
...
-    snprintf(buf, NUMBUFLEN, "%" PRId64, o.data.number);
-    return buf;
-  }
-  case kOptValTypeString: {
-    char *buf = xmalloc(o.data.string.size + 3);
-    snprintf(buf, o.data.string.size + 3, "\"%s\"", o.data.string.data);
-    return buf;
-  }
+  Error err = ERROR_INIT;
+  const char *res = nlua_inspect(optval_as_object(value), &err);
```
justinmk, line 3224: is there a reason you can't use `NLUA_EXEC_STATIC` instead of adding `nlua_inspect` ?

src/nvim/option.c
```diff
@@ -3212,63 +3212,67 @@ static void set_option_varp(OptIndex opt_idx, void *varp, OptVal value, bool fre
...
 
-/// Convert an OptVal to an API Object.
-Object optval_as_object(OptVal o)
+/// Convert an OptVal to an API Object. Always returns a valid Object.
+///
+/// @param  value  OptVal to convert.
+Object optval_as_object(OptVal value)
 {
-  switch (o.type) {
+  switch (value.type) {
```
justinmk, line 3239: seems like an unnecessary rename

PR #31407, opened 2024-11-30
test(treesitter): global highlight definitions and fold test

luukvbaal:
Add test for foldtext= highlighting. Change file to global highlight definitions while at it.


luukvbaal:
I caught a regression for the added test while working on I caused #31324. Extracted this to a separate PR because it's not necessarily related.

justinmk (approved):

PR #31408, opened 2024-12-01
test(vterm): move test functions into vterm_test fixture

jamessan:
In order to run unittests with a release build, we need the test
functions to be accessible when NDEBUG is defined. Moving the functions
into the test fixture ensures they are available and only available for
use by the unit tests.

---
Pulling this out of #31358 because it won't backport to release-0.10.


src/nvim/CMakeLists.txt
```diff
@@ -707,6 +707,13 @@ else()
   file(GLOB UNIT_TEST_FIXTURES CONFIGURE_DEPENDS ${PROJECT_SOURCE_DIR}/test/unit/fixtures/*.c)
   target_sources(nvim_bin PRIVATE ${UNIT_TEST_FIXTURES})
   target_compile_definitions(nvim_bin PRIVATE UNIT_TESTING)
+  if(MSVC)
+    set_source_files_properties(
+      ${UNIT_TEST_FIXTURES} PROPERTIES COMPILE_OPTIONS "-wd4090;-wd4244;-wd4267")
+  else()
+    set_source_files_properties(
```
dundargoc, line 714: Just fix the warnings instead? There are in total 5 warnings, all of them are simple conversion warnings.

justinmk (approved):

src/nvim/CMakeLists.txt
jamessan: These were warning suppressions we already apply to vterm as a whole. I just copied them to the fixture when I extracted the code.

I can fix the warnings, but I wasn't sure what the context was behind the existing suppressions.

dundargoc: Context was that we shouldn't needlessly change code we bundle/don't own as to diverge as little as possible from upstream. In this case we pretty much own the test code (arguably at this point I'd say we even basically "own" vterm code, just that no one has done the work to integrate it properly yet). So we no longer need to ignore the warnings for vterm code. I can fix the windows warnings at a future date if you don't have a windows machine on you.

dundargoc (approved):

PR #31409, opened 2024-12-01
Fix translation errors in src/nvim/po/vi.po

brianhuster:
### Why?
The Vietnamese keymaps in Vim is quite different from corresponding input methods that Vietnamese people use, so I think there should be a help document that tell users these differences

**Edit : The file vietnamese.txt has been removed because I have sent it to Vim**

## 2. Update vi.po
Surprisingly, the file `vi.po` in Neovim has so many errors, though fortunately they are mostly related to Vimscript runtime error or some other unpopular (and even removed) features. I have yet found an error in Vim's `vi.po` btw.  I have fixed many errors (but I think there may still some left), and add translation for a few other messages.

## Other questions
Vim provides Vimtutor for many languages, but there doesn't seem to be a way to open any localized Vimtutor in Vim. The only way I know is using `vimtutor` command in terminal, but it doesn't open the tutor in Vim. So I would like to know if this is intentional, or should I open an issue for this?


clason:
> I have sent a similar PR to https://github.com/vim/vim/pull/16144

Thank you! Then we need to wait until that is merged so we can port it using our standard approach. This prevents duplicate effort and conflicts.

 If something Neovim-specific is needed on top of that, we can add another PR on top of that.

brianhuster:
I decided to spend another PR here because the file `src/nvim/vi.po` is not the same as `src/po/vi.po` in Vim, and unlike the corresponding file in Vim, this file in Neovim has many errors in translation

brianhuster:
Btw, I would like to ask if there is a specific reason that lines 6394 - end in `src/nvim/vi.po` were commented?

clason:
`git blame` and searching Github for the corresponding PR is the best answer. I don't think anybody ever updated these files after forking them from Vim ten years ago. (One reason why a hard fork is problematic.)

brianhuster:
So when I update vi.po, do I need to add something like 
```po
#: ../api/private/helpers.c:201
```
before each id of the message? In Vim they don't do that

clason:
You don't need to, but it's helpful. (Check other po files, best sorted by date.)

brianhuster:
And if msgstr is "", will Neovim use English message instead?

brianhuster:
Since fixing the `vi.po` file in Neovim takes quite lot of time, I'll close this. @clason Do you think I should just copy the file `vi.po` from Vim to Neovim. The only downside is that it doesn't have Nvim-specific messages

dundargoc:
@brianhuster you might need to use `cmake --build build --target update-po-vi` to update the translation file. Haven't touched translations much but it might help.

brianhuster:
I don't think that helps. I think that command only build `.mo` files from corresponding `.po` files, right? But the problem here is in the file `.po`

PR #31411, opened 2024-12-01
docs: provide example for configuring LSP foldexpr

gpanders:
Using the "supports_method" function with a client capability inside of an LspAttach autocommand is the preferred method to do this, so we should be showing users how to do it.

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -720,6 +720,19 @@ foldexpr({lnum})                                          *vim.lsp.foldexpr()*
     Provides an interface between the built-in client and a `foldexpr`
     function.
 
+    To use, check for the "textDocument/foldingRange" capability in an
+    |LspAttach| autocommand. Example: >lua
+        vim.api.nvim_create_autocommand('LspAttach', {
+          callback = function(args)
+            local client = vim.lsp.get_client_by_id(args.data.client_id)
+            if client:supports_method('textDocument/foldingRange') then
+              vim.wo.foldmethod = 'expr'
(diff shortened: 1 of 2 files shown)


runtime/doc/lsp.txt
```diff
@@ -720,6 +720,18 @@ foldexpr({lnum})                                          *vim.lsp.foldexpr()*
     Provides an interface between the built-in client and a `foldexpr`
     function.
 
+    To use, check for the "textDocument/foldingRange" capability in an
+    |LspAttach| autocommand. Example: >lua
+        vim.api.nvim_create_autocommand('LspAttach', {
+          callback = function(args)
+            local client = vim.lsp.get_client_by_id(args.data.client_id)
+            if client.supports_method('textDocument/foldingRange') then
```
MariaSolOs, line 728:
```suggestion
            if client:supports_method('textDocument/foldingRange') then
```

Since the non-method function was deprecated.

runtime/doc/lsp.txt
```diff
@@ -720,6 +720,18 @@ foldexpr({lnum})                                          *vim.lsp.foldexpr()*
...
+              vim.wo.foldexpr = 'v:lua.vim.lsp.foldexpr()'
```
MariaSolOs, line 729: We should also add `vim.wo.foldmethod = 'expr'`.

runtime/doc/lsp.txt
gpanders: Addressed both in c0898f8. Thanks!

MariaSolOs (approved):

PR #31418, opened 2024-12-02
fix(decorator): make errors less noisy when decoration provider has exception

rudiejd:
Problem: https://github.com/neovim/neovim/issues/27235

Solution: 
Don't show error messages when errors for a decoration provider exceed `DP_MAX_ERROR`. I tried disabling the errors based on provider state being disabled, but that didn't fix the issue


justinmk (approved):

rudiejd:
I'm not sure a functional test is worth the trouble here because I'd have to figure out how to add and remove a busted parser as part of the test setup / cleanup. Happy to attempt that if people feel strongly about it, but otherwise I was unable to reproduce this issue when I manually tested following [the reproduction steps on the original issue.](https://github.com/neovim/neovim/issues/27235#issue-2103853854)

PR #31420, opened 2024-12-02
feat: add API for getting file permissions in octal

alichtman:
Problem: There is currently no API to return the current file permissions in octal.

Solution: Add one. The `getfpermoctal` API function returns the current file permissions, in octal.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -3528,6 +3528,20 @@ getfperm({fname})                                                   *getfperm()*
                 Return: ~
                   (`string`)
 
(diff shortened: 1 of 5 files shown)


zeertzjq:
No, thanks.

alichtman:
Why not?

zeertzjq:
If you want the file permission in another format you can format the value of `vim.uv.fs_stat(file).mode`. There doesn't need to be another function that achieves the same thing, but in a more limited way.

alichtman:
Thanks for the help!

PR #31432, opened 2024-12-03
docs: Correct name of nvim_create_autocmd in example

andrewferrier:
Example documentation for LSP folding contains invalid API function `nvim_create_autocommand`, it should be `nvim_create_autocmd`.


justinmk:
https://github.com/neovim/neovim/pull/31476

PR #31439, opened 2024-12-03
fix(column): check if signcolumn changed in all windows

### LINKED ISSUE #31438: Placing sign via extmark inside decoration provider might not result into showing sign column in another windows
### Problem

With `signcolumn=auto` and sign placed via extmark inside decoration provider, sign column might be not visible when it should be.

### Steps to reproduce

1. Create the following 'init.lua':
    ```lua
    -- Prepare UI: six windows all displaying same buffer
    local many_lines = {}
    for i = 1, 500 do
      table.insert(many_lines, 'Line #' .. i)
    end
    vim.api.nvim_buf_set_lines(0, 0, -1, false, many_lines)

    vim.cmd('wincmd s | wincmd s | wincmd v | wincmd j | wincmd v | wincmd j | wincmd v')
    vim.schedule(function() vim.cmd('wincmd j') end)

    -- Define helper data
    _G.log, _G.needs_clear = {}, false
    local ns_id, mark_id = vim.api.nvim_create_namespace('test'), nil

    -- Add decoration which possibly clears all extmarks and adds one on line 499
    local on_win = function(_, winid, bufnr, toprow, botrow)
      table.insert(_G.log, { winid, bufnr, toprow, botrow })

      if _G.needs_clear then
        vim.api.nvim_buf_clear_namespace(bufnr, ns_id, 0, -1)
        _G.needs_clear = false
      end

      if toprow < 499 and 499 <= botrow then
        mark_id = vim.api.nvim_buf_set_extmark(bufnr, ns_id, 499, 0, { id = mark_id, sign_text = '!', invalidate = true })
      end
    end
    vim.api.nvim_set_decoration_provider(ns_id, { on_win = on_win })
    ```
2. `nvim --clean -u path/to/init.lua`
3. Press `G` to go to last line. Sign `!` appears in current window **but** sign column in other windows does not appear.
4. Press `<C-l>` to redraw the screen - all sign columns appear.
5. Execute `:lua log, needs_clear = {}, true`. Sign column disappears in top two windows but still present in other windows.
6. Execute `:=log` to see that windows which were missing sign columns were processed by decoration provider *before* the one which added extmark.

Here is the demo:

https://github.com/user-attachments/assets/314312eb-bc78-4eca-b3b3-7d1757b04685

### Expected behavior

Adding sign to a buffer via extmark should result into sign column appearing in all windows displaying that buffer.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1272+g3d3a99e69c

### Operating system/version

### Terminal name/version

Ghostty 0.1.0-main+3c637a27

xterm-ghostty

### Installation

appimage

luukvbaal:
Fix #31438

### DIFF
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -1516,10 +1516,12 @@ static void win_update(win_T *wp)
 
   decor_providers_invoke_win(wp);
 
-  if (win_redraw_signcols(wp)) {
-    wp->w_lines_valid = 0;
-    wp->w_redr_type = UPD_NOT_VALID;
-    changed_line_abv_curs_win(wp);
+  FOR_ALL_WINDOWS_IN_TAB(win, curtab) {
+    if (win->w_buffer == wp->w_buffer && win_redraw_signcols(win)) {
+      win->w_lines_valid = 0;
+      changed_line_abv_curs_win(win);
+      redraw_later(win, UPD_NOT_VALID);
+    }
   }
 
   init_search_hl(wp, &screen_search_hl);
--- a/test/functional/ui/sign_spec.lua
+++ b/test/functional/ui/sign_spec.lua
@@ -607,4 +607,77 @@ describe('Signs', function()
     eq(6, infos[1].textoff)
     eq(6, infos[2].textoff)
   end)
+
+  it('auto width updated in all windows after sign placed in on_win #31438', function()
+    exec_lua([[
+      vim.cmd.call('setline(1, range(1, 500))')
+      vim.cmd('wincmd s | wincmd v | wincmd j | wincmd v')
+
+      _G.log, _G.needs_clear = {}, false
+      local ns_id, mark_id = vim.api.nvim_create_namespace('test'), nil
+
+      -- Add decoration which possibly clears all extmarks and adds one on line 499
+      local on_win = function(_, winid, bufnr, toprow, botrow)
+        if _G.needs_clear then
+          vim.api.nvim_buf_clear_namespace(bufnr, ns_id, 0, -1)
+          _G.needs_clear = false
+        end
+
+        if toprow < 499 and 499 <= botrow then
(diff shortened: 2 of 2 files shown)


src/nvim/drawscreen.c
```diff
@@ -1516,10 +1516,12 @@ static void win_update(win_T *wp)
 
   decor_providers_invoke_win(wp);
 
-  if (win_redraw_signcols(wp)) {
-    wp->w_lines_valid = 0;
-    wp->w_redr_type = UPD_NOT_VALID;
-    changed_line_abv_curs_win(wp);
+  FOR_ALL_WINDOWS_IN_TAB(win, curtab) {
```
zeertzjq, line 1519: Should this be now in `update_screen()` instead?

luukvbaal: Yeah agree that seems to more natural place for it now but that wouldn't solve the linked issue. It needs to happen after `on_win` callbacks.

test/functional/ui/sign_spec.lua
```diff
@@ -607,4 +607,63 @@ describe('Signs', function()
...
+      vim.cmd.call('setline(1, range(1, 500))')
+      vim.cmd('wincmd s | wincmd v | wincmd j | wincmd v')
+
+      local ns_id, mark_id = vim.api.nvim_create_namespace('test'), nil
+
+      -- Add decoration which possibly clears all extmarks and adds one on line 499
+      local on_win = function(_, winid, bufnr, toprow, botrow)
+        if _G.needs_clear then
+          vim.api.nvim_buf_clear_namespace(bufnr, ns_id, 0, -1)
+          _G.needs_clear = false
```
echasnovski, line 622: Should this also test the `:lua needs_clear = true` case? As it seems to have different behavior before this PR (all windows without signcolumn vs windows before target are without signcolumn).

justinmk (approved):

PR #31441, opened 2024-12-03
test(screen): adjust screen state to more closely follow stylua

luukvbaal:
Before:
```lua
screen:expect({        | screen:expect({
  grid = [[            |   grid = [[
    {10:>!}a        |  |     line ^1                   |
    {7:  }b        |   |     {1:~                        }|*4
    {10:>>}c        |  |   ]], messages={ {
    {7:  }^         |  |     content = { { "\ntest\n[O]k: ", 6, 11 } },
    {1:~          }|*9 |     kind = "confirm"
               |       |   } }
  ]]                   | })
})
```

After:
```lua
screen:expect([[       | screen:expect({
  {10:>!}a          |  |   grid = [[
  {7:  }b          |   |     line ^1                   |
  {10:>>}c          |  |     {1:~                        }|*4
  {7:  }^           |  |   ]],
  {1:~            }|*9 |   messages = { {
               |       |     content = { { "\ntest\n[O]k: ", 6, 11 } },
]])                    |     kind = "confirm"
                       |   } },
                       | })
```

### DIFF
--- a/test/functional/ui/screen.lua
+++ b/test/functional/ui/screen.lua
@@ -37,10 +37,10 @@
 -- Tests will often share a group of extra attribute sets to expect(). Those can be
 -- defined at the beginning of a test:
 --
---    screen:add_extra_attr_ids {
+--    screen:add_extra_attr_ids({
 --      [100] = { background = Screen.colors.Plum1, underline = true },
 --      [101] = { background = Screen.colors.Red1, bold = true, underline = true },
---    }
+--    })
 --
 -- To help write screen tests, see Screen:snapshot_util().
 -- To debug screen tests, see Screen:redraw_debug().
@@ -454,7 +454,7 @@ end
 ---    screen:expect(grid, [attr_ids])
 ---    screen:expect(condition)
 --- or keyword args (supports more options):
----    screen:expect{grid=[[...]], cmdline={...}, condition=function() ... end}
+---    screen:expect({ grid=[[...]], cmdline={...}, condition=function() ... end })
 ---
 --- @param expected string|function|test.function.ui.screen.Expect
 --- @param attr_ids? table<integer,table<string,any>>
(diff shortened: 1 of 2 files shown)


luukvbaal:
Unlikely to get all the way there without rolling a custom `vim.inspect` that follows stylua, but this at least gets us closer. Currently the no ext_state case always bites me for no reason when I forget to run stylua.

clason:
https://github.com/neovim/neovim/pull/28330

test/functional/ui/screen.lua
```diff
@@ -37,10 +37,10 @@
 -- Tests will often share a group of extra attribute sets to expect(). Those can be
 -- defined at the beginning of a test:
 --
---    screen:add_extra_attr_ids {
+--    screen:add_extra_attr_ids({
 --      [100] = { background = Screen.colors.Plum1, underline = true },
 --      [101] = { background = Screen.colors.Red1, bold = true, underline = true },
---    }
+--    })
```
justinmk, line 43: stylua doesn't enforce this does it?

luukvbaal: It doesn't but I changed it for consistency with what `_print_snapshot()` spits out for the `expect()` call. Do we have a preference? (Mine is with `()`).

justinmk: `expect{` is 3x more common than `expect({` so I guess our preference is `expect{`.

luukvbaal: On the other hand `expect([[` is 1000x more common than `expect [[`.

justinmk: My sense from observing other core contributors is that `expect{` is preferred

zeertzjq: Because that's the current output of `snapshot_util()`...

PR #24914, opened 2023-08-28
feat(terminal)!: open :terminal in a split by default

gpanders:
With a bang `[!]`, `:terminal` opens a terminal buffer in the current window, discarding any changes to the current buffer unless `'hidden'` is set (which it is by default as of #15410).

This is a BREAKING CHANGE.

---

Note this reverts some of the changes from #15427 (notably the doc changes to |:horizontal| and the news.txt update from that PR).

PR #31443, opened 2024-12-03
defaults: disable 'number', 'relativenumber', and 'signcolumn' in terminal buffers

gpanders:
Users who wish to have line numbers enabled in the terminal can use an autocommand:

```vim
au TermOpen * setlocal number relativenumber
```

### DIFF
--- a/runtime/doc/vim_diff.txt
+++ b/runtime/doc/vim_diff.txt
@@ -186,6 +186,9 @@ nvim_terminal:
     - 'textwidth' set to 0
     - 'nowrap'
     - 'nolist'
+    - 'nonumber'
+    - 'norelativenumber'
+    - 'signcolumn' set to "no"
     - 'winhighlight' uses |hl-StatusLineTerm| and |hl-StatusLineTermNC| in
       place of |hl-StatusLine| and |hl-StatusLineNC|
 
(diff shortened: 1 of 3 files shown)


justinmk:
No objection. IIRC the reason for not doing this (inheriting this current window options) was to avoid jank when switching to/from a `:terminal` in the current window.

The desire for this would be less pronounced if we move forward with https://github.com/neovim/neovim/pull/24914 ?

gpanders:
>IIRC the reason for inheriting this current window options, was to avoid jank when switching to/from a :terminal in the current window.

Do you recall what the issue was? I think with the multi-indexed options (`vim.wo[0][0]`) this works smoothly: the options are only set for the terminal buffer in the window.

Trying this:

```
VIMRUNTIME=runtime build/bin/nvim --clean --cmd 'set number' README.md
```

Then open a new window and create a terminal: `:vsplit|term`. The terminal buffer has no line numbers. Now switch back to the alt buffer `:e #` and the line numbers are back.

justinmk:
> Do you recall what the issue was?

To avoid the window resizing.

We can just try this though, it doesn't matter too much imo.

igorlfs:
Another option that affects terminal buffers is the `'scrolloff'`. When leaving terminal mode, the buffer may be unintentionally scrolled (and it also scrolls back when returning to terminal mode). To reproduce, just run a command long enough to fill the window's height, and run `clear`. On terminal mode, the cursor is at the top:

image

But after returning to normal mode, `'scrolloff'` is applied:

image

Given that this PR already leans into better defaults, I wonder if it makes sense to also set `scrolloff=0`?

PR #31446, opened 2024-12-04
vim-patch:fdfcce5: runtime(lua): add optional lua function folding

zeertzjq:
#### vim-patch:fdfcce5: runtime(lua): add optional lua function folding

closes: vim/vim#16151

https://github.com/vim/vim/commit/fdfcce56a6b8fe64ae155c822e3b920db97cde0d

Co-authored-by: Enno <Konfekt@users.noreply.github.com>

### DIFF
--- a/runtime/doc/filetype.txt
+++ b/runtime/doc/filetype.txt
@@ -695,6 +695,11 @@ To enable the recognition of Markdown comments each time after removing
 re-source "javaformat.vim" for Vim versions greater than `8.2.1397`: >
 	runtime autoload/javaformat.vim
 <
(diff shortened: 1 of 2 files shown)


clason:
I'm not sure we want this, to be honest, since we use treesitter for Lua by default and so should focus on treesitter folding.

zeertzjq:
What do you mean by "promoting" and "intrusive"?

clason:
I mean we should not suggest that people use this custom folding method over treesitter folding. And the ftplugin defines a lot of stuff that is not conditional and useless (for us).

zeertzjq:
Yes

Issue #31384, opened 2024-11-29
Built-in completion can still affect nearby extmarks


echasnovski:
### Problem

Built-in completion is prone to setting text even if that is not visibly necessary. This can affect extmarks. Although #23657 fixed some of them, there are still some out there it seems.

cc @luukvbaal, @bfredl 

### Steps to reproduce

1. Create the  following 'init.lua':
    ```lua
    vim.api.nvim_buf_set_lines(0, 0, -1, false, { 'abc abcd' })
    vim.api.nvim_win_set_cursor(0, { 1, 3 })

    _G.ns_id = vim.api.nvim_create_namespace('extmark')
    local extmark_opts = { end_col = 3, hl_group = 'Search', right_gravity = false, end_right_gravity = false }
    _G.mark_id = vim.api.nvim_buf_set_extmark(0, _G.ns_id, 0, 0, extmark_opts)

    -- Same issue even with 'menuone,noinsert,noselect'
    vim.o.completeopt = 'menuone,noinsert'
    ```
2. `nvim --clean -u path/to/init.lua`
3. Press `i` and `<C-n>`. It should show popup with `abcd` as candidate while extmark is still intact.
4. Type `e`. Extmark is affected although no visible changes to the text in its range was made. Executing `:vim.api.nvim_buf_get_extmark_by_id(0, ns_id, mark_id, { details = true })` confirms that it was squashed to the right (as both right gravity were disabled).

This happens both on latest `master` and 0.10.2, so doesn't look like a Nightly regression.

### Expected behavior

Built-in completion might affect extmarks only if it is intended to change text in extmark range.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1252+gb1c907f219

### Vim (not Nvim) behaves the same?

No, doesn't have this feature

### Operating system/version

EndeavourOS Linux x86_64, 6.12.1-arch1-1

### Terminal name/version

Ghostty 0.1.0-main+d47df211

### $TERM environment variable

xterm-ghostty

### Installation

appimage

PR #31448, opened 2024-12-04
fix(completion): avoid deleting text when completion leader changes

luukvbaal:
Problem:  When completion leader changes, text that might be reinserted
          immediately after is deleted. This unnecessarily affects
          extmarks. #31387 restored the original extmarks but that
          prevents end_right_gravity marks from growing.
Solution: Avoid deleting leader text that will be reinserted.

Fix #31437

### DIFF
--- a/src/nvim/insexpand.c
+++ b/src/nvim/insexpand.c
@@ -951,14 +951,14 @@ static void ins_compl_longest_match(compl_T *match)
     compl_leader = xstrdup(match->cp_str);
 
     bool had_match = (curwin->w_cursor.col > compl_col);
-    ins_compl_delete();
+    ins_compl_delete(false);
     ins_bytes(compl_leader + get_compl_len());
     ins_redraw(false);
 
     // When the match isn't there (to avoid matching itself) remove it
     // again after redrawing.
     if (!had_match) {
-      ins_compl_delete();
+      ins_compl_delete(false);
     }
     compl_used_match = false;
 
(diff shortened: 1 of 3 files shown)


luukvbaal:
Cannot guarantee this will not break something else, trying my best...
I also tried completely removing the save/restore extmark code in place of always avoiding to delete text that will be restored but this messes with the rest of the completion logic (it seems `find_next_completion_match()` expects the leader to be removed).

echasnovski:
@luukvbaal, all regressions after #31387 that I can think of seem to be gone with this PR. Along with #31384. And no obvious extra new issues. Huge thanks!

I can see one weird behavior on the intersection of extmarks and built-in completion, but it is also present on 0.10.2 and pre-#31387 versions, so it deserves a separate issue (if turns out to be an issue in the first place).

src/nvim/insexpand.c
```diff
@@ -3599,11 +3598,21 @@ static void ins_compl_update_shown_match(void)
...
+void ins_compl_delete(bool new_leader)
 {
+  // Avoid deleting text that will be reinserted when changing leader. This
+  // allows marks present on the original text to shrink/grow appropriately.
+  int i = 0;
+  if (new_leader) {
+    char *leader = ins_compl_leader();
+    while (leader[i] == compl_orig_text[i] && leader[i] != NUL) {
+      i++;
+    }
```
zeertzjq, line 3610: I think this should compare by characters instead of bytes, because `ins_bytes()` inserts the string by characters.

PR #31464, opened 2024-12-05
vim-patch:189e24b: runtime(doc): include vietnamese.txt

brianhuster:
Since Vietnamese keymaps in Vim is quite differences from the corresponding input methods, let's document the Vietnamese specifics in vietnames.txt

related: vim/vim#16144

https://github.com/vim/vim/commit/189e24bb1441abe87387a4e21c4387bbf2eac718

vim-patch::9.1.0902

Problem:    
            1. Vim's built-in Vietnamese keymaps have some important
            differences from the correspond Vietnamese input methods, but they
            are undocumented
            2. For some reason, the file vi.po in Neovim is so different
               from that of Vim, and it also contains many mistakes in
               translation, compared to that of Vim

Solution:   
            Add vietnamese.txt
            Copy the file vi.po from Vim to Neovim

https://github.com/vim/vim/commit/8a52587ee05a360cad4d42322200775fc74a4430

### DIFF
--- a/runtime/doc/help.txt
+++ b/runtime/doc/help.txt
@@ -150,6 +150,7 @@ LANGUAGE SUPPORT
 |arabic.txt|		Arabic language support and editing
 |hebrew.txt|		Hebrew language support and editing
 |russian.txt|		Russian language support and editing
+|vietnamese.txt|	Vietnamese language support and editing
 
 ------------------------------------------------------------------------------
 INTEROP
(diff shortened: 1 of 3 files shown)


justinmk:
All these different locale-related help files should have a common prefix like `l10n_`. There are 100s of human languages.

brianhuster:
So should I change file names of hebrew.txt, russian.txt and arabic.txt?

justinmk:
> So should I change file names of hebrew.txt, russian.txt and arabic.txt?

I'm in favor of that. Any objection @zeertzjq ?

clason:
Would need changes to the `vim-patch` script. (Which needs updating anyway, back to the filetype.lua change.)

PR #31470, opened 2024-12-06
fix(filetype): make filetype detection work with :doautocmd

### LINKED ISSUE #31306: Autodetecting filetype of compressed file
### Discussed in https://github.com/neovim/neovim/discussions/31300

<div type='discussions-op-text'>

<sup>Originally posted by **avidseeker** November 22, 2024</sup>
In Vim, `:set ft` returns the expected `markdown`
```
$ echo test | xz -c > test.md.xz && vim --clean test.md.xz
```
In Neovim, `nvim --clean test.md.xz`, the `:set ft` is not set.

This seems to be a regression from #27931.

It seems that autocmd pattern matching actually matches with both the unexpanded filename and the expanded file name:
```c
  if (regmatch.regprog != NULL
      && ((allow_dirs
           && (vim_regexec(&regmatch, fname, 0)
               || (sfname != NULL
                   && vim_regexec(&regmatch, sfname, 0))))
          || (!allow_dirs && vim_regexec(&regmatch, tail, 0)))) {
    result = true;
  }
```
That'll be a bit complicated to replicate with `filetype.lua`...

zeertzjq:
Fix #31306

### DIFF
--- a/runtime/filetype.lua
+++ b/runtime/filetype.lua
@@ -13,8 +13,8 @@ vim.api.nvim_create_autocmd({ 'BufRead', 'BufNewFile', 'StdinReadPost' }, {
     end
     local ft, on_detect = vim.filetype.match({
       -- The unexpanded file name is needed here. #27914
-      -- Neither args.file nor args.match are guaranteed to be unexpanded.
-      filename = vim.fn.bufname(args.buf),
+      -- However, bufname() can't be used, as it doesn't work with :doautocmd. #31306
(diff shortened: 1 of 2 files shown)


zeertzjq:
`args.file` being affected by `expand('<afile>')` is quite confusing. It should probably be fixed first...

zeertzjq:
With #31473 this should be better.

justinmk (approved):

PR #31472, opened 2024-12-06
docs(annotation): added missing `link` field

ColinKennedy:
`vim.api.nvim_get_hl` returns `vim.api.keyset.get_hl_info`, which can contain `link` but the definition for `vim.api.keyset.get_hl_info` is missing that key, which causes llscheck + lua-language-server to think the code is an error.

```lua
vim.api.nvim_set_hl(0, "FooBarFizzBuzz", {link="Normal"})
print(vim.inspect(vim.api.nvim_get_hl(0, {name="FooBarFizzBuzz"})))
```
prints
```
{
  link = "Normal"
}
```

### DIFF
--- a/runtime/lua/vim/_meta/api_keysets_extra.lua
+++ b/runtime/lua/vim/_meta/api_keysets_extra.lua
@@ -159,6 +159,7 @@ error('Cannot require a meta file')
 --- @field bg? integer
 --- @field sp? integer
 --- @field default? true
+--- @field link? string
 --- @field blend? integer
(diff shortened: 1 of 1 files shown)


dundargoc:

ColinKennedy:
@dundargoc how should I be suggesting new documentation fixes in the future? I'd imagine you wouldn't want me to directly edit the other PR. Should I still make PR suggestions but set the base to something non-master? Or just continue as I already did?

dundargoc:
Just make a regular PR like you're doing

PR #31473, opened 2024-12-06
fix(events): don't expand `args.file` for Lua callback

zeertzjq:
Problem:  In an autocommand Lua callback whether `args.file` is expanded
          depends on whether `expand('<afile>')` has been called.
Solution: Always use the unexpanded file name for `args.file`.

Related to #31306 and vim/vim#16106. This doesn't provide `sfname`, but
at least makes `args.file` have a consistent value.


justinmk (approved):
Nice

clason:
Backport?

zeertzjq:
Wait for some time to see if anything breaks

clason:
As long as we don't forget.

PR #31475, opened 2024-12-06
fix(marks): check gravity at range bounds when deleting text

### LINKED ISSUE #31465: Extmarks might not track text changes in combination with built-in completion and `TextChangedP` event
### Problem

Updating text and extmarks inside `TextChangedP` event can lead to incorrect tracking by extmarks of their ranges. In particular, when user navigates from selected entry in completion popup to no selected entry in completion popup.

I do understand that this is a rather niche use case with complicated reproduction. Yet it does seem like an indication of some issue in core.
Also, this situation can happen during snippet session with a `$1$1` snippet. At the moment, `vim.snippet` doesn't work with that snippet (no snippet session is started) and doesn't track changes in 'TextChangedP', but both seem reasonable to become reality.

**Note**: this happens on 0.10.2, current Nightly (`NVIM v0.11.0-dev-1287+g8f84167c30`), and after #31448 (at the state of `NVIM v0.11.0-dev-1291+gd817375d3e`).

### Steps to reproduce

1. Create the following (minimized as much as I was able to) 'init.lua':

    <details><summary>Code of 'init.lua'</summary>

    ```lua
    -- Define helper functions to work with extmarks
    local ns_id = vim.api.nvim_create_namespace('extmark')

    local get_extmark_data = function(extmark_id)
      local data = vim.api.nvim_buf_get_extmark_by_id(0, ns_id, extmark_id, { details = true })
      -- Make `data[3]` settable as `opts` in `nvim_buf_set_extmark()`
      data[3].id, data[3].ns_id = extmark_id, nil
      return data[1], data[2], data[3]
    end

    local set_gravity = function(extmark_id, right_gravity, end_right_gravity)
      local row, col, opts = get_extmark_data(extmark_id)
      opts.right_gravity, opts.end_right_gravity = right_gravity, end_right_gravity
      vim.api.nvim_buf_set_extmark(0, ns_id, row, col, opts)
    end

    -- Define logger for tracking extmark evolution
    _G.log = {}
    local add_log = function(state)
      local extmarks = vim.api.nvim_buf_get_extmarks(0, ns_id, 0, -1, { details = true })
      table.insert(_G.log, { state = state, line = vim.api.nvim_get_current_line(), extmarks = extmarks })
    end

    -- Set repro state: text for built-in completion and place to use it.
    vim.api.nvim_buf_set_lines(0, 0, -1, false, { 'aa bb', '' })
    vim.api.nvim_win_set_cursor(0, { 2, 0 })
    vim.o.completeopt = 'menuone,noselect'

    -- Create empty extmarks: `mark_1` as reference and `mark_2` to be synced
    -- IMPORTANT: marks are at the same column. No issue if they are apart.
    local mark_1 = vim.api.nvim_buf_set_extmark(0, ns_id, 1, 0, { end_row = 1, end_col = 0, hl_group = 'Search' })
    local mark_2 = vim.api.nvim_buf_set_extmark(0, ns_id, 1, 0, { end_row = 1, end_col = 0, hl_group = 'String' })
    set_gravity(mark_1, false, true)
    set_gravity(mark_2, true, true)

    -- Define synchronization of mark_2
    local n = 0
    local sync_mark_2 = function()
      -- Log before
      n = n + 1
      add_log('Before sync #' .. n)

      -- Change gravity for text change to properly affect extmarks:
      -- "left" for `mark_1` (untouched) and "expand" for `mark_2` (tracks text)
      set_gravity(mark_1, false, false)
      set_gravity(mark_2, false, true)

      -- Set text of mark_1 to be the same for mark_2
      local row_1, col_1, opts_1 = get_extmark_data(mark_1)
      local mark_1_text = vim.api.nvim_buf_get_text(0, row_1, col_1, opts_1.end_row, opts_1.end_col, {})
      local row_2, col_2, opts_2 = get_extmark_data(mark_2)
      vim.api.nvim_buf_set_text(0, row_2, col_2, opts_2.end_row, opts_2.end_col, mark_1_text)

      -- Change gravity back: "expand" for `mark_1` and "right" for mark_2
      set_gravity(mark_1, false, true)
      set_gravity(mark_2, true, true)

      -- Log after
      add_log('After sync #' .. n)
    end
    vim.api.nvim_create_autocmd({ 'TextChangedP' }, { callback = sync_mark_2 })
    ```
    </details>
2. `nvim --clean -u path/to/init.lua`
3. Press `<C-n>`. Popup should appear without anything selected.
4. Press `<C-n>` to select `aa` entry. It should be inserted in the text **twice** with different highlighting: first `Search` and then `String`. This is a result of text synchronization.
5. Press `<C-p>` to go to initial state. There is still *unhighlighted* `aa` text present, although it should have been deleted during synchronization. This is a result of (what looks like) incorrect tracking by extmarks of their ranges. Here is a pretty formatted log:

    <details><summary>Formatted log with exact place indicating the tracking issue</summary>

    ```
    {
      -- First `<C-n>` press
      {
        state = "Before sync #1",
        line = "",
        extmarks = {
          { 1, 1, 0, { end_row = 1, end_col = 0, right_gravity = false, end_right_gravity = true, hl_eol = false, hl_group = "Search", ns_id = 1, priority = 4096 } },
          { 2, 1, 0, { end_row = 1, end_col = 0, right_gravity = true,  end_right_gravity = true, hl_eol = false, hl_group = "String", ns_id = 1, priority = 4096 } },
        },
      },
      {
        state = "After sync #1",
        line = "",
        extmarks = {
          { 1, 1, 0, { end_row = 1, end_col = 0, right_gravity = false, end_right_gravity = true, hl_eol = false, hl_group = "Search", ns_id = 1, priority = 4096 } },
          { 2, 1, 0, { end_row = 1, end_col = 0, right_gravity = true,  end_right_gravity = true, hl_eol = false, hl_group = "String", ns_id = 1, priority = 4096 } },
        },
      },
      -- Second `<C-n>` press
      {
        state = "Before sync #2",
        line = "aa",
        extmarks = {
          { 1, 1, 0, { end_row = 1, end_col = 2, right_gravity = false, end_right_gravity = true, hl_eol = false, hl_group = "Search", ns_id = 1, priority = 4096 } },
          { 2, 1, 2, { end_row = 1, end_col = 2, right_gravity = true,  end_right_gravity = true, hl_eol = false, hl_group = "String", ns_id = 1, priority = 4096 } },
        },
      },
      {
        state = "After sync #2",
        line = "aaaa",
        extmarks = {
          { 1, 1, 0, { end_row = 1, end_col = 2, right_gravity = false, end_right_gravity = true, hl_eol = false, hl_group = "Search", ns_id = 1, priority = 4096 } },
          { 2, 1, 2, { end_row = 1, end_col = 4, right_gravity = true,  end_right_gravity = true, hl_eol = false, hl_group = "String", ns_id = 1, priority = 4096 } },
        },
      },
      -- `<C-p>` press
      {
        state = "Before sync #3",
        line = "aa",
        extmarks = {
          { 1, 1, 0, { end_row = 1, end_col = 0, right_gravity = false, end_right_gravity = true, hl_eol = false, hl_group = "Search", ns_id = 1, priority = 4096 } },
          -- Main issue indicator:
          -- - Before pressing `<C-p>` this extmark covered (1;2)-(1;4) range with both right gravity.
          -- - After pressing `<C-p>` this extmark says that it covers (1;0)-(1;0) range.
          --   This should not be the case, as only first two characters should have changed.
          { 2, 1, 0, { end_row = 1, end_col = 0, right_gravity = true,  end_right_gravity = true, hl_eol = false, hl_group = "String", ns_id = 1, priority = 4096 } },
        },
      },
      {
        state = "After sync #3",
        line = "aa",
        extmarks = {
          { 1, 1, 0, { end_row = 1, end_col = 0, right_gravity = false, end_right_gravity = true, hl_eol = false, hl_group = "Search", ns_id = 1, priority = 4096 } },
          { 2, 1, 0, { end_row = 1, end_col = 0, right_gravity = true,  end_right_gravity = true, hl_eol = false, hl_group = "String", ns_id = 1, priority = 4096 } },
        },
      },
    }
    ```

    <details>

Here the demo with showcased log:

https://github.com/user-attachments/assets/71f63ca1-a412-4c15-b6d5-fbc504845346

### Expected behavior

Extmarks always track text changes in the same way.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1287+g8f84167c30

### Operating system/version

### Terminal name/version

Ghostty 0.1.0-main+3c637a27

xterm-ghostty

### Installation

Appimage

luukvbaal:
Problem:  Marks that are properly restored by the splice associated with
          an undo edit, are unnecessarily pushed to the undo header. This
          results in incorrect mark tracking in the "copy_only"
          save/restore completion path.
Solution: Avoid pushing left gravity marks at the beginning of the range,
          and right gravity marks at the end of the range, to the undo
          header.

Fix #31465

### DIFF
--- a/src/nvim/extmark.c
+++ b/src/nvim/extmark.c
@@ -367,10 +367,21 @@ void extmark_splice_delete(buf_T *buf, int l_row, colnr_T l_col, int u_row, coln
   marktree_itr_get(buf->b_marktree, (int32_t)l_row, l_col, itr);
   while (true) {
     MTKey mark = marktree_itr_current(itr);
-    if (mark.pos.row < 0
-        || mark.pos.row > u_row
-        || (mark.pos.row == u_row && mark.pos.col > u_col)) {
+    if (mark.pos.row < 0 || mark.pos.row > u_row) {
       break;
+    }
+
+    bool copy = true;
+    // No need to copy left gravity marks at the beginning of the range,
+    // and right gravity marks at the end of the range, unless invalidated.
+    if (mark.pos.row == l_row && mark.pos.col - !mt_right(mark) < l_col) {
+      copy = false;
+    } else if (mark.pos.row == u_row) {
+      if (mark.pos.col > u_col + 1) {
+        break;
+      } else if (mark.pos.col + mt_right(mark) > u_col) {
+        copy = false;
+      }
     }
 
     bool invalidated = false;
@@ -388,6 +399,7 @@ void extmark_splice_delete(buf_T *buf, int l_row, colnr_T l_col, int u_row, coln
           extmark_del(buf, itr, mark, true);
           continue;
         } else {
+          copy = true;
           invalidated = true;
           mt_itr_rawkey(itr).flags |= MT_FLAG_INVALID;
           marktree_revise_meta(buf->b_marktree, itr, mark);
@@ -397,7 +409,7 @@ void extmark_splice_delete(buf_T *buf, int l_row, colnr_T l_col, int u_row, coln
     }
 
     // Push mark to undo header
-    if (only_copy || (uvp != NULL && op == kExtmarkUndo && !mt_no_undo(mark))) {
+    if (copy && (only_copy || (uvp != NULL && op == kExtmarkUndo && !mt_no_undo(mark)))) {
       ExtmarkSavePos pos = {
         .mark = mt_lookup_key(mark),
         .invalidated = invalidated,
@@ -541,10 +553,8 @@ void extmark_splice_impl(buf_T *buf, int start_row, colnr_T start_col, bcount_t 
 
   if (old_row > 0 || old_col > 0) {
     // Copy and invalidate marks that would be effected by delete
-    // TODO(bfredl): Be "smart" about gravity here, left-gravity at the
-    // beginning and right-gravity at the end need not be preserved.
-    // Also be smart about marks that already have been saved (important for
-    // merge!)
+    // TODO(bfredl): Be smart about marks that already have been
+    // saved (important for merge!)
     int end_row = start_row + old_row;
     int end_col = (old_row ? 0 : start_col) + old_col;
     u_header_T *uhp = u_force_get_undo_header(buf);
--- a/test/functional/api/extmark_spec.lua
+++ b/test/functional/api/extmark_spec.lua
@@ -249,7 +249,7 @@ describe('API/extmarks', function()
     set_extmark(ns, 2, 1, 0, { right_gravity = false })
     eq({ { 1, 0, 0 }, { 2, 1, 0 } }, get_extmarks(ns, { 0, 0 }, { -1, -1 }))
     feed('u')
-    eq({ { 1, 0, 0 }, { 2, 1, 0 } }, get_extmarks(ns, { 0, 0 }, { -1, -1 }))
+    eq({ { 1, 0, 0 }, { 2, 0, 0 } }, get_extmarks(ns, { 0, 0 }, { -1, -1 }))
     api.nvim_buf_clear_namespace(0, ns, 0, -1)
   end)
 


echasnovski:
As far as I can tell this indeed fixes #31465! Huge thanks again!

Moreover, if I am not mistaken in my verification, it also fixes all cases in #31437. The #31448 might still be a good change in on itself, though.

clason:
Is this backportable?

luukvbaal:
Given that @bfredl confirms this implements his TODO note, I believe so. This change should have no bearing on the final position of the affected marks after undo. The affected test is a bit suspicious in that sense. Maybe we need to check `mt_end()` as well:
```
diff --git a/src/nvim/extmark.c b/src/nvim/extmark.c
index 6119d838f9..5da68c6747 100644
--- a/src/nvim/extmark.c
+++ b/src/nvim/extmark.c
@@ -374,12 +374,12 @@ void extmark_splice_delete(buf_T *buf, int l_row, colnr_T l_col, int u_row, coln
     bool copy = true;
     // No need to copy left gravity marks at the beginning of the range,
     // and right gravity marks at the end of the range, unless invalidated.
-    if (mark.pos.row == l_row && mark.pos.col - !mt_right(mark) < l_col) {
+    if (mark.pos.row == l_row && mark.pos.col - (!mt_end(mark) && !mt_right(mark)) <
l_col) {
       copy = false;
     } else if (mark.pos.row == u_row) {
       if (mark.pos.col > u_col + 1) {
         break;
-      } else if (mark.pos.col + mt_right(mark) > u_col) {
+      } else if (mark.pos.col + (mt_end(mark) && mt_right(mark)) > u_col) {
         copy = false;
       }
     }
```

That leaves the test unaffected while still fixing #31465, but I'm not sure if it makes sense to single out end marks in that way. Are pair end marks treated differently in the marktree splice @bfredl?

justinmk (approved):
Waiting for @bfredl to take a look

bfredl:
> That leaves the test unaffected while still fixing https://github.com/neovim/neovim/issues/31465, but I'm not sure if it makes sense to single out end marks in that way. Are pair end marks treated differently in the marktree splice @bfredl?

No they are not (except the different _default_ gravity) . The behavior here is due to how deleting the only line of the buffer is removed (and thus how that is being undone), which is unrelated to the change in this PR and thus not a blocker.

echasnovski:
@bfredl, can/should this be backported?

PR #31480, opened 2024-12-06
defaults: disable 'foldcolumn' in terminal buffers

mehalter:
Follow up to #31443. The new terminal defaults are very sane and in the same vein I think it is a good idea to also by default disable foldcolumn since none of the default fold methods make sense for terminals and there is a minimal change that the user would want the fold column there by default when using an in editor terminal. I also want to make it clear I am not looking for this to lead to tons of default terminal buffer default settings, just that the `foldcolumn` seems to achieve the same goals as disabling sign and number columns.

Let me know what you think!

EDIT: looks like the commit linter doesn't like the commit name. I can happily change it if necessary, but it seems like it is the format used previously so I am thinking this is a false negative,


gpanders:
Thanks!

PR #31479, opened 2024-12-06
docs: misc

justinmk:
todo

- [x] https://github.com/neovim/neovim/pull/30085#discussion_r1741869280

Issue #31454, opened 2024-12-05
LSP: Improve textDocument/diagnostic performance by conforming to LSP spec


molostovvs:
### Problem

Users of the Roslyn LSP (used for C# in VS Code) have encountered significant delays when retrieving pull diagnostics in large documents while using Neovim. For instance, diagnostics in a 2000-line .cs file can take over 20 seconds to display after edits in Neovim, whereas in VS Code, diagnostics for the same file are displayed almost instantly.

As [@mparq noted](https://github.com/seblj/roslyn.nvim/issues/93#issuecomment-2508940330) in https://github.com/seblj/roslyn.nvim/issues/93, VS Code leverages additional parameters specified in the [LSP documentation for textDocument/diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentDiagnosticParams), specifically:

- previousResultId
- identifier

### Expected behavior

When requesting diagnostics, Neovim should include the `previousResultId` and `identifier` parameters as part of the request.
These parameters enable the server to utilize caching and return incremental results.

Support for maintaining state is already present in the [textDocument/semanticTokens implementation](https://github.com/neovim/neovim/blob/8f84167c30692555d3332565605e8a625aebc43c/runtime/lua/vim/lsp/semantic_tokens.lua#L289).
A similar mechanism could probably be implemented in `textDocument/diagnostic` handler.

tris203:
I am willing to work on this, but I think that it would benefit from having a central state management to hold the previous results identifier to avoid building another siloed LSP data store
- #31453

PR #31483, opened 2024-12-06
fix(lsp): cancel pending requests before refreshing

tris203:
Problem:
Diagnostics and inlay hints can be expensive to calculate, and we shouldn't stack them as this can cause noticeable lag.

Solution:
Check for duplicate inflight requests and cancel them before issuing a new one. This ensures that only the latest request is processed, improving performance and preventing potential conflicts.

This is part of an issue uncovered here:
https://github.com/seblj/roslyn.nvim/issues/93

Open as draft whilst some more people test it

### DIFF
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -2162,6 +2162,11 @@ function M._refresh(method, opts)
         local first = vim.fn.line('w0', window)
         local last = vim.fn.line('w$', window)
         for _, client in ipairs(clients) do
+          for rid, req in pairs(client.requests) do
+            if req.method == method and req.type == 'pending' and req.bufnr == bufnr then
+              client:cancel_request(rid)
+            end
+          end
           client:request(method, {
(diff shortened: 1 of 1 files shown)


seblj:
Without this PR:

https://github.com/user-attachments/assets/adc4f394-0fdf-4eaf-bd0b-aa8fb86b2faa

With this PR:

https://github.com/user-attachments/assets/9c1a6b17-3cac-4cd1-8b7b-8db817a1213c

This immediately seems _much_ better

tris203:
This is probably a temporary workaround until something like #31453 is implemented which means we can implement more of the diagnostic spec as per #31454.

Once we have a state store, then we can make use of incremental publishes which slow servers probably implement

lewis6991 (approved):

mfussenegger (approved):

justinmk:
Added to news in https://github.com/neovim/neovim/pull/31479

Perf improvements are definitely worth mentioning in news!

PR #31485, opened 2024-12-06
fix(inspect): show priority for treesitter highlights

clason:
Problem: `:Inspect` does not show priority for treesitter highlights,
leading to confusion why sometimes earlier highlights override later
highlights.

Solution: Also print priority metadata if set.

Screenshot 2024-12-06 at 19 21 29

### DIFF
--- a/runtime/lua/vim/_inspector.lua
+++ b/runtime/lua/vim/_inspector.lua
@@ -191,7 +191,12 @@ function vim.show_pos(bufnr, row, col, filter)
     append('Treesitter', 'Title')
     nl()
     for _, capture in ipairs(items.treesitter) do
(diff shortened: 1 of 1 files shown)


gpanders (approved):

justinmk (approved):
Nice. Similar to documentation, it's roughly true that anything that is not visible basically doesn't exist.

clason:
My distillation of recent (unrelated) discussions: Neovim is the editor for people who know what they're doing. Which is a great, open, niche, but needlessly exclusive. So Neovim should be the editor for people who _want_ to know what they're doing, so we should provide the tools (and docs) for that.

PR #31488, opened 2024-12-06
fix(treesitter): proper #trim! range for nodes ending at col 0

ribru17:
**Problem:** #30085 introduced char-wise folding support for `#trim!`, but the ranges were improperly calculated for nodes that end at column 0, due to the way `get_node_text` works

**Solution:** Add the blank line that `get_node_text` removes for for nodes ending at column 0. Also properly set column positions when performing linewise trims.

I apologize for not catching this sooner

### DIFF
--- a/test/functional/treesitter/parser_spec.lua
+++ b/test/functional/treesitter/parser_spec.lua
@@ -696,14 +696,14 @@ print()
     end)
 
     it('trims only empty lines by default (backwards compatible)', function()
-      insert [[
+      insert(dedent [[
       ## Heading
 
       With some text
 
       ## And another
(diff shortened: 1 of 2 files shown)


test/functional/treesitter/parser_spec.lua
```diff
@@ -696,14 +696,13 @@ print()
...
+      insert [[## Heading
 
-      With some text
+With some text
 
-      ## And another
+## And another
 
-      With some more]]
+With some more here]]
```
justinmk, line 705: can use `t.dedent()` to make this look nicer

PR #31491, opened 2024-12-07
refactor(autocmd): allow specifying window for autocmd context

famiu:
Problem: Currently we can only specify a buffer to use for an autocmd context through `aucmd_prepbuf()`, which finds a window that uses that buffer in the current tabpage, or creates an autocmd window. This means it's not possible to actually specify a window to use for an autocmd.

Solution: Add an `aucmd_prepbuf_win()` function which also takes a window as a parameter and uses it for the autocmd. If the window is not provided, then it behaves similarly to `aucmd_prepbuf()`

### DIFF
--- a/src/nvim/autocmd.c
+++ b/src/nvim/autocmd.c
@@ -1252,56 +1252,82 @@ bool check_nomodeline(char **argp)
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


famiu:
Moved back inside #31112

PR #31497, opened 2024-12-07
fix(inspect): always show priority

clason:
Problem: It is not obvious if a treesitter highlight priority shown in
`:Inspect` is higher or lower than the default.

Solution: Also print default priority (`vim.hl.priorities.treesitter`).
Add padding for better readability.

Follow-up to #31496 

**Question:** Should we still print the `language`? Strictly speaking it's redundant with the highlight group suffix and adds visual clutter; we also don't show it for semantic tokens.

Screenshot 2024-12-07 at 11 01 04

### DIFF
--- a/runtime/lua/vim/_inspector.lua
+++ b/runtime/lua/vim/_inspector.lua
@@ -178,7 +178,7 @@ function vim.show_pos(bufnr, row, col, filter)
     if data.hl_group ~= data.hl_group_link then
       append('links to ', 'MoreMsg')
       append(data.hl_group_link, data.hl_group_link)
-      append(' ')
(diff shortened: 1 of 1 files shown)


justinmk:
> Should we still print the `language`? Strictly speaking it's redundant with the highlight group suffix and adds visual clutter; we also don't show it for semantic tokens.

if it's 100% redundant then +1 for omitting it

clason:
I think so, but on the other hand this makes it more obvious where the highlight is coming from even if you haven't learned how the treesitter highlight groups are constructed?

justinmk:
🤷

PR #31508, opened 2024-12-08
fix(ui): update title in more cases

### LINKED ISSUE #31507: Mode doesn't update in title
### Problem

I have been encountering problems when having the mode in my `titlestring`, it only updates if i am on the normal or visual mode.

In neovim it doesn't work, but in vim it does, as you can see in this video:

https://github.com/user-attachments/assets/466ea954-5a4b-4284-8d82-5ebfa4d6d4c0

### Steps to reproduce

- nvim
- :lua vim.opt.titlestring = "%m %f (%{mode()}) | nvim"
- Go to insert mode (mode doesn't change in the title)
- Go to visual mode (it changes)
- Go back to normal mode (it changes)
- Go to visual line mode (it changes)
- Go to replace mode (it doesn't change)
- Go to terminal mode (it doesn't change)

### Expected behavior

The mode should change in the title

### Nvim version (nvim -v)

0.11.0-dev

no, vim 9.1.787

### Operating system/version

NixOS unstable commit fa42b5a

### Terminal name/version

Alacritty 0.14.0

alacritty

### Installation

Nix

zeertzjq:
Fix #31507


luukvbaal:
> sets need_maketitle instead of calling maketitle()

Instead of adding it in various places, perhaps an `if (need_maketitle) { maketitle() }` should be added to `update_screen()` and `redraw_statuslines()`? Presumably #22590 made the assumption that it would actually be called in the redraw_later code path.

Otherwise it might need to be added to `show_cursor_info_later()` callsites as well. The following does not update the title either:
```vim
set showmatch title titlestring=%{col('.')}
```

justinmk (approved):
@luukvbaal  ?

luukvbaal (approved):

PR #31509, opened 2024-12-08
fix(lsp): cancel pending requests on detach

ofseed:
This should resolve the bug mentioned in [this comment](https://github.com/neovim/neovim/pull/31311#issuecomment-2511024970).

The reproduction method is to open a relatively large file (so the server needs some time to process the request), and then immediately execute `:bd`. Once the request is completed, the handler will obtain the bufstate of a buffer already unloaded.

Should this fix be applied at a lower level, that is, should all pending requests of the client attached to the buffer be immediately canceled after the buffer is unloaded?

### DIFF
--- a/runtime/lua/vim/lsp/_folding_range.lua
+++ b/runtime/lua/vim/lsp/_folding_range.lua
@@ -197,6 +197,11 @@ local function setup(bufnr)
     -- `on_detach` also runs on buffer reload (`:e`).
     -- Ensure `bufstate` and hooks are cleared to avoid duplication or leftover states.
     on_detach = function()
+      util._cancel_requests({
+        bufnr = bufnr,
+        method = ms.textDocument_foldingRange,
+        type = 'pending',
+      })
       bufstates[bufnr] = nil
       api.nvim_clear_autocmds({ buffer = bufnr, group = augroup_setup })
     end,
--- a/runtime/lua/vim/lsp/util.lua
+++ b/runtime/lua/vim/lsp/util.lua
@@ -2135,6 +2135,41 @@ local function make_line_range_params(bufnr, start_line, end_line, position_enco
   }
 end
 
+---@class (private) vim.lsp.util._cancel_requests.Filter
+---@field bufnr? integer
+---@field clients? vim.lsp.Client[]
+---@field method? string
+---@field type? string
+
+---@private
+--- Cancel all {filter}ed requests.
+---
+---@param filter? vim.lsp.util._cancel_requests.Filter
+function M._cancel_requests(filter)
+  filter = filter or {}
+  local bufnr = filter.bufnr and vim._resolve_bufnr(filter.bufnr) or nil
+  local clients = filter.clients
+  local method = filter.method
+  local type = filter.type
+
+  for _, client in
+    ipairs(clients or vim.lsp.get_clients({
+      bufnr = bufnr,
+      method = method,
+    }))
+  do
+    for id, request in pairs(client.requests) do
+      if
+        (bufnr == nil or bufnr == request.bufnr)
+        and (method == nil or method == request.method)
+        and (type == nil or type == request.type)
+      then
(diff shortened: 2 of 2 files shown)


MariaSolOs:
> Should this fix be applied at a lower level, that is, should all pending requests of the client attached to the buffer be immediately canceled after the buffer is unloaded?

Yeah, I personally think that this cancellation should happen for all pending requests and not only for folding ranges.

ofseed:
I just thought of an edge case where jumping through `buf.reference` and similar methods may still require the request to not be canceled. At this point, the buffer in the request is the source rather than the target.

runtime/lua/vim/lsp/_folding_range.lua
```diff
@@ -197,6 +197,22 @@ local function setup(bufnr)
     -- `on_detach` also runs on buffer reload (`:e`).
     -- Ensure `bufstate` and hooks are cleared to avoid duplication or leftover states.
     on_detach = function()
+      for _, client in
+        ipairs(vim.lsp.get_clients({
+          bufnr = bufnr,
+          method = ms.textDocument_foldingRange,
+        }))
+      do
+        for id, req in pairs(client.requests) do
```
justinmk, line 206: Even if we decide to only do this for folding, it seems worth lifting this into a `lsp.util._cancel_all_requests(method)` util function?

here is another case: https://github.com/neovim/neovim/blob/92e61072ac7493d85c15e3403f17619446782b51/runtime/lua/vim/lsp/util.lua#L2164-L2169

justinmk:
> I just thought of an edge case where jumping through `buf.reference` and similar methods may still require the request to not be canceled. At this point, the buffer in the request is the source rather than the target.

Maybe we could enumerate the explicit methods that we don't want to cancel, in that low-level handler.

ofseed:
> Maybe we could enumerate the explicit methods that we don't want to cancel, in that low-level handler.

I am concerned about the methods in protocol extensions, as their behavior is completely uncertain. However, I might be overthinking it a bit; the cancellation of requests performed on a buffer after it has been deleted should be an expected behavior. I will handle it at a lower level later if this is the consensus.

MariaSolOs:
> > Maybe we could enumerate the explicit methods that we don't want to cancel, in that low-level handler.
> 
> I am concerned about the methods in protocol extensions, as their behavior is completely uncertain. However, I might be overthinking it a bit; the cancellation of requests performed on a buffer after it has been deleted should be an expected behavior. I will handle it at a lower level later if this is the consensus.

If by protocol extensions you mean custom messages that aren't specified in the LSP, then those would need special server/client handlers outside of Neovim, so I wouldn't worry about that.

ofseed:
Sorry I complicated this issue. Based on @justinmk's suggestion, I extracted a new `util._cancel_requests` to help us fix this potential bug that cannot be stably reproduced first.

runtime/lua/vim/lsp/util.lua
```diff
@@ -2135,6 +2135,41 @@ local function make_line_range_params(bufnr, start_line, end_line, position_enco
...
+---@field method? string
+---@field type? string
+
+---@private
+--- Cancel all {filter}ed requests.
+---
+---@param filter? vim.lsp.util._cancel_requests.Filter
+function M._cancel_requests(filter)
+  filter = filter or {}
+  local bufnr = filter.bufnr
```
justinmk, line 2150: I could see a caller accidentally sending buf=0 expecting it to mean current buffer. So we should probably handle that.

```suggestion
  local bufnr = filter.bufnr and vim._resolve_bufnr(filter.bufnr) or nil
```

justinmk (approved):

ofseed:
@hexium310 Oops, that's not intended. Thank you for mentioning that. I'm currently checking https://github.com/neovim/neovim/issues/32277, I will create a PR for both of them later.

PR #31519, opened 2024-12-08
fix(messages): no message kind for bufwrite

przepompownia:
- Problem: no good way to replace the initial bufwrite message (from `filemess`) by the final one (`"test.lua" [New] 0L, 0B written`), when using `vim.ui_attach`.
- Solution: add kind to both messages.

### DIFF
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -789,6 +789,7 @@ must handle.
 	kind
 	    Name indicating the message kind:
 		"" (empty)	Unknown (consider a |feature-request|)
+		"bufwrite"	|:write| message
 		"confirm"	|confirm()| or |:confirm| dialog
 		"confirm_sub"	|:substitute| confirm dialog |:s_c|
 		"emsg"		Error (|errors|, internal error, |:throw|, …)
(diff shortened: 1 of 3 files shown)


luukvbaal:
The `replace_last` argument to `msg_show` is `true` for the second message, so replacing is already possible (although I'm still not sure whether taking `replace_last` suggestions actually makes sense for a ui_attach implementation that doesnt write to the cmdline area).

But I think this kind can still be useful despite that.

luukvbaal (approved):

przepompownia:
> The `replace_last` argument to `msg_show` is `true` for the second message,

Yes, I didn't miss it but I'm not sure if there is no message between the message from `filemess` and the final one (and as a result of this, what is the previous message to replace).

runtime/doc/ui.txt
```diff
@@ -789,6 +789,7 @@ must handle.
 	kind
 	    Name indicating the message kind:
 		"" (empty)	Unknown (consider a |feature-request|)
+		"bufwrite"	|:write| message
```
justinmk, line 792: we have "list_cmd" and "search_cmd' already, should this be "write_cmd" ? Or is this triggered by other ways of writing a buffer?

luukvbaal: IDK; "list_cmd" contains output from one of various _commands_ that list something, "search_cmd" contains the actual search _command_ itself, not a message, so they're all a bit different. But yeah this message comes from the :write _command_ so the _cmd suffix could make sense.

PR #31521, opened 2024-12-08
docs: add guidance on if feature should be a plugin or not

dundargoc:
Can close if not wanted. Mostly checking if there is any interest in this or if the information is self-explanatory.

### DIFF
--- a/runtime/doc/develop.txt
+++ b/runtime/doc/develop.txt
@@ -654,5 +654,21 @@ External UIs are expected to implement these common features:
 - Support the text decorations/attributes given by |ui-event-hl_attr_define|.
   The "url" attr should be presented as a clickable hyperlink.
 
+Adding Features to core vs plugins ~
+
+Many features are better suited as plugins rather than core features. There
+will never be a surefire way to determine whether a feature should be added to
+the editor or if it should be a plugin, but here are some signs. A feature is
+more likely to be suited as a plugin if...
+
+- it sucks
+- can easily be achieved without changes to neovim editor
+- requires lots of maintenance
+- requires lots of code
+- the neovim maintainers aren't interested in maintaining it
+
+The last point is the least technical yet arguably the most important point.
+If the feature does not "spark joy" it's better to not do it (adding features
+is sexy and easy; maintaining features is boring and flaccid).
 
  vim:tw=78:ts=8:sw=4:et:ft=help:norl:


dundargoc:
@justinmk I dissed your `develop.txt` suggestions last time, feel free to return the favor 😉.

My idea was mostly to have a checklist of questions we want people to think about and have answers to before suggesting a feature to core. It's a somewhat recurring discussion, so thought it might be a good idea to start saving the rough conclusions from these discussions that we can then build upon. Not 100% sure if it happens enough to to warrant this, so inout welcome.

Will obviously clean up the language and elaborate if we want to go ahead with this.

runtime/doc/develop.txt
```diff
@@ -654,5 +654,21 @@ External UIs are expected to implement these common features:
...
 
+Adding Features to core vs plugins ~
+
+Many features are better suited as plugins rather than core features. There
+will never be a surefire way to determine whether a feature should be added to
+the editor or if it should be a plugin, but here are some signs. A feature is
+more likely to be suited as a plugin if...
+
+- it sucks
+- can easily be achieved without changes to neovim editor
```
dundargoc, line 665: I guess the "easily" is a bit of a weasel word, people have different gauges on what they constitute as easy. Also, "easy" probably should relate to how easy it is compared to the value of the feature, rather than how easy it is in a vacuum.

justinmk:
Does "plugin" here mean "third-party, not maintained by Nvim repo"? Hopefully https://neovim.io/charter/ helps answer this already. 

The general answer is we want to "lift" stuff into Nvim if it is important to our OOTB experience. And having a maintenance strategy for it is also required, as you mentioned.

dundargoc:
> Does "plugin" here mean "third-party, not maintained by Nvim repo"?

Yes, exactly.

> Hopefully https://neovim.io/charter/ helps answer this already.

Huh, didn't know that existed. I think it's pretty good as its own thing, but not sure how much guidance it provides in regards to this specifically. The most relevant is probably "Optimize 'out of the box', for new users but especially regular users"? Although not entirely sure what out of the box experience would mean for regular users, that seems like an oxymoron since it wouldn't be out of the box as they already should have existing configuration?
 
> The general answer is we want to "lift" stuff into Nvim if it is important to our OOTB experience. And having a maintenance strategy for it is also required, as you mentioned.

Makes sense. Unsure how we'd even start providing guidelines for what constitutes a good OOTB experience, that seems like pure guesswork/artform.

The existing texts might be good enough. I think I'll close this for now since it seems there is not an immediate need, thanks for the input.

PR #31525, opened 2024-12-09
feat(ui)!: emit prompt "messages" as cmdline events

luukvbaal:
Problem:  Prompts are emitted as message events, where cmdline events
          are more appropriate. The user input is also emitted as
          message events in fast context, so cannot be displayed with
          vim.ui_attach().
Solution: Ask for user input through cmdline prompts.

### DIFF
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -120,7 +120,7 @@ typedef struct {
   int indent;
   int c;
   bool gotesc;                          // true when <ESC> just typed
-  int do_abbr;                          // when true check for abbr.
+  bool do_abbr;                         // when true check for abbr.
   char *lookfor;                        // string to match
   int lookforlen;
   int hiscnt;                           // current history line in use
@@ -128,17 +128,17 @@ typedef struct {
                                         // to jump to next match
   int histype;                          // history type to be used
   incsearch_state_T is_state;
-  int did_wild_list;                    // did wild_list() recently
+  bool did_wild_list;                   // did wild_list() recently
   int wim_index;                        // index in wim_flags[]
   int save_msg_scroll;
   int save_State;                       // remember State when called
   int prev_cmdpos;
   char *save_p_icm;
-  int some_key_typed;                   // one of the keys was typed
+  bool some_key_typed;                  // one of the keys was typed
   // mouse drag and release events are ignored, unless they are
   // preceded with a mouse down event
-  int ignore_drag_release;
-  int break_ctrl_c;
+  bool ignore_drag_release;
+  bool break_ctrl_c;
   expand_T xpc;
   OptInt *b_im_ptr;
   buf_T *b_im_ptr_buf;  ///< buffer where b_im_ptr is valid
@@ -1848,6 +1848,12 @@ static int command_line_browse_history(CommandLineState *s)
 
 static int command_line_handle_key(CommandLineState *s)
 {
+  // For one key prompt, avoid putting ESC and Ctrl_C onto cmdline.
+  // For all other keys, just put onto cmdline and exit.
+  if (ccline.one_key && s->c != ESC && s->c != Ctrl_C) {
+    goto end;
+  }
+
   // Big switch for a typed command line character.
   switch (s->c) {
   case K_BS:
@@ -1998,6 +2004,12 @@ static int command_line_handle_key(CommandLineState *s)
     }
     FALLTHROUGH;
   case K_LEFTMOUSE:
+    // Return on left click above number prompt
+    if (ccline.mouse_used && mouse_row < cmdline_row) {
+      *ccline.mouse_used = true;
+      return 0;
+    }
+    FALLTHROUGH;
   case K_RIGHTMOUSE:
     command_line_left_right_mouse(s);
     return command_line_not_changed(s);
@@ -2155,6 +2167,14 @@ static int command_line_handle_key(CommandLineState *s)
     }
     return command_line_not_changed(s);
 
+  case 'q':
+    // Number prompts use the mouse and return on 'q' press
+    if (ccline.mouse_used) {
+      *ccline.cmdbuff = NUL;
+      return 0;
+    }
+    FALLTHROUGH;
+
   default:
     // Normal character with no special meaning.  Just set mod_mask
     // to 0x0 so that typing Shift-Space in the GUI doesn't enter
@@ -2175,6 +2195,7 @@ static int command_line_handle_key(CommandLineState *s)
     return command_line_changed(s);
   }
 
+end:
   // put the character in the command line
   if (IS_SPECIAL(s->c) || mod_mask != 0) {
     put_on_cmdline(get_special_key_name(s->c, mod_mask), -1, true);
@@ -2183,7 +2204,7 @@ static int command_line_handle_key(CommandLineState *s)
     IObuff[j] = NUL;                // exclude composing chars
     put_on_cmdline(IObuff, j, true);
   }
-  return command_line_changed(s);
+  return ccline.one_key ? 0 : command_line_changed(s);
 }
 
 static int command_line_not_changed(CommandLineState *s)
@@ -2721,8 +2742,11 @@ static void abandon_cmdline(void)
   if (msg_scrolled == 0) {
     compute_cmdrow();
   }
-  msg("", 0);
-  redraw_cmdline = true;
+  // Avoid overwriting key prompt
+  if (!ccline.one_key) {
+    msg("", 0);
+    redraw_cmdline = true;
+  }
 }
 
 /// getcmdline() - accept a command line starting with firstc.
@@ -2761,11 +2785,13 @@ char *getcmdline(int firstc, int count, int indent, bool do_concat FUNC_ATTR_UNU
 /// @param[in]  xp_context  Type of expansion.
 /// @param[in]  xp_arg  User-defined expansion argument.
 /// @param[in]  highlight_callback  Callback used for highlighting user input.
+/// @param[in]  one_key  Return after one key press for button prompt.
+/// @param[in]  mouse_used  Set to true when returning after right mouse click.
 ///
 /// @return [allocated] Command line or NULL.
 char *getcmdline_prompt(const int firstc, const char *const prompt, const int hl_id,
                         const int xp_context, const char *const xp_arg,
-                        const Callback highlight_callback)
+                        const Callback highlight_callback, bool one_key, bool *mouse_used)
   FUNC_ATTR_WARN_UNUSED_RESULT FUNC_ATTR_MALLOC
 {
   const int msg_col_save = msg_col;
@@ -2786,11 +2812,14 @@ char *getcmdline_prompt(const int firstc, const char *const prompt, const int hl
   ccline.xp_arg = (char *)xp_arg;
   ccline.input_fn = (firstc == '@');
   ccline.highlight_callback = highlight_callback;
+  ccline.one_key = one_key;
+  ccline.mouse_used = mouse_used;
 
   int msg_silent_saved = msg_silent;
   msg_silent = 0;
 
   char *const ret = (char *)command_line_enter(firstc, 1, 0, false);
+  ccline.redraw_state = kCmdRedrawNone;
 
   if (did_save_ccline) {
     restore_cmdline(&save_ccline);
@@ -4787,7 +4816,7 @@ void get_user_input(const typval_T *const argvars, typval_T *const rettv, const 
   const int save_ex_normal_busy = ex_normal_busy;
   ex_normal_busy = 0;
   rettv->vval.v_string = getcmdline_prompt(secret ? NUL : '@', p, get_echo_hl_id(),
-                                           xp_type, xp_arg, input_callback);
+                                           xp_type, xp_arg, input_callback, false, NULL);
   ex_normal_busy = save_ex_normal_busy;
   callback_free(&input_callback);
 
--- a/test/functional/ui/cmdline_spec.lua
+++ b/test/functional/ui/cmdline_spec.lua
@@ -1485,31 +1485,28 @@ describe('cmdheight=0', function()
   it('when substitute text', function()
     command('set cmdheight=0 noruler laststatus=3')
     feed('ifoo<ESC>')
-    screen:expect {
-      grid = [[
+    screen:try_resize(screen._width, 6)
+    screen:expect([[
       fo^o                      |
-      {1:~                        }|*3
+      {1:~                        }|*4
       {3:[No Name] [+]            }|
-    ]],
-    }
+    ]])
 
     feed(':%s/foo/bar/gc<CR>')
-    screen:expect {
-      grid = [[
+    screen:expect([[
       {2:foo}                      |
-      {1:~                        }|*3
-      {6:replace wi...q/l/^E/^Y)?}^ |
-    ]],
-    }
+      {3:                         }|
+                               |*2
+      {6:replace with bar (y/n/a/q}|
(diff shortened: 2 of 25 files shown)


przepompownia:
Please run the case

```lua
local api = vim.api

local function log(msg)
  if 0 == #vim.trim(msg) then
    return
  end
  vim.cmd.normal 'G'
end

local function redraw()
  vim.schedule(function ()
    vim.cmd.redraw()
  end)
end

api.nvim_create_autocmd('CmdlineEnter', {
  callback = function ()
    vim.ui_detach(ns)
    redraw()
  end
})

local qMsgs = {}

local function consumeQMsgs()
    log(vim.inspect(qmsg))
  end
  qMsgs = {}
end

local function attach()
  vim.ui_attach(ns, {ext_messages = true, ext_cmdline = false}, function (event, kind, content, replace_last)
    if event ~= 'msg_show' then
      return
    end
      api.nvim_input('\r')
    end

    local msg = {
      time = vim.uv.hrtime() / 1000000,
      fast = vim.in_fast_event(),
      e = event or 'none',
      k = kind or 'none',
      c = content or 'none',
      r = replace_last or 'none',
    }
      consumeQMsgs()
      log(vim.inspect(msg))
      redraw()
      return
    end

    qMsgs[#qMsgs + 1] = msg
    vim.schedule(function ()
      consumeQMsgs()
    end)
  end)
end

  callback = function ()
    attach()
    vim.fn.inputlist({'Select: ', 1, 2})
  end
})
```
(similar to https://github.com/neovim/neovim/discussions/31522). 


For some reason I don't see the message from `inputlist` called right after `ui_attach`. It probably doesn't matter in practice but worth checking. 

If I `:call inputlist([])` manually, I see `number_prompt`, which is a step forward.

Why not to display `list_cmd` above too?

luukvbaal:
Hoping to make this incremental change while leaving what is displayed on the current message grid unaffected. Invoking a cmdline prompt affects the message grid globals however, which makes that tricky. Spending a lot of time trying to make it work but might have to leave this until after the message grid is hopefully removed...

test/functional/ui/input_spec.lua
```diff
@@ -361,25 +361,16 @@ describe('input non-printable chars', function()
...
     feed('n')
     screen:expect([[
       foobar                                                      |
+      {1:~                                                           }|
       {3:                                                            }|
       "Xtest-overwrite"                                           |
       {9:WARNING: The file has been changed since reading it!!!}      |
       {6:Do you really want to write to it (y/n)?}u                   |
-      {6:Do you really want to write to it (y/n)?}                    |
       {6:Do you really want to write to it (y/n)?}n                   |
```
justinmk, line 374: For my understanding: why was only one of these "prompts"(?) removed?

justinmk (approved):
This looks great to me, it adds structure to flows that were spread around and implicit in various places.

It might be a bit risky but I think we should try it out. 

@fredizzimo @glacambre @theol0403 do you see any obvious problems?

test/functional/ui/input_spec.lua
luukvbaal: Old [implementation](https://github.com/neovim/neovim/pull/31525/files#diff-5134a5ea161c11fa8571f2cc0f9677d108ebda3e5741438147979ae504b7ef4aL58) emitted the message and pressed key on each invalid keypress. Now we stay inside the active cmdline state if the `command_line_handle_key()` switch returns early. Could probably make it behave the same if necessary.

glacambre:
I haven't looked at the code, just did a quick check and this doesn't work too well with firenvim - `:%s/x/y/gc` results in an empty command line popping up without additional information for the user to know what state they're in (i.e., the message saying `replace with b? (y/n)` is gone).
This might just be an issue with Firenvim itself though, I'll let you know once I've figure it out.

luukvbaal:
I just rebased my default ext_ui branch on top of this (which I admittedly hadn't done yet, just looked at the affected ext_ui tests which seemed to show correct state). Seems to work as expected if  you display the prompt message in the cmdline.

zeertzjq:
I think this is dangerous. Any code can be executed in cmdline mode, so this can lead to potential segfaults.

luukvbaal:
> With a non-fast message the executed code is limited to the callback. With a cmdline prompt there can be autocommands from unrelated plugins.

Yeah I realized that while typing my reply; if that's your concern we avoid firing those autocommands those when inside a cmdline prompt.

luukvbaal:
> Also, prompt_for_key() cannot return Ctrl-E, which is an possible key in the substitution prompt. It cannot return some other control characters like Ctrl-C and Esc either, despite them being checked for in the return value.

Right, I can go over these. So if timers callbacks can be executed in Vim during prompts, are we okay with processing cmdline autocmds/scheduled events as well, until we run into actual problems? Or should we preemptively avoid doing that?

test/functional/ui/input_spec.lua
luukvbaal: We now also return after keys that did not change the cmdline, so that the callers can still properly handle control characters. This test now displays `^E` rater than nothing.

luukvbaal:
One key prompts now always return early even if the cmdline did not change, so that callers can still handle control characters. Also further simplified this a bit by merging the number and one key prompts.

test/old/testdir/test_spell.vim
```diff
@@ -470,8 +470,10 @@ func Test_spellsuggest_option_number()
   \    "\n"
   \ .. "Change \"baord\" to:\n"
   \ .. " 1 \"board\"\n"
-  \ .. " 2 \"bard\"\n"
-  \ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
+  "\ Nvim: Prompt message is sent to cmdline prompt.
+  "\ .. " 2 \"bard\"\n"
+  \ .. " 2 \"bard\"\n", a)
+  "\ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
```
zeertzjq, line 476:
```suggestion
  \ .. " 2 \"bard\"\n"
  "\ Nvim: Prompt message is sent to cmdline prompt.
  "\ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
  \ , a)
```

test/old/testdir/test_spell.vim
```diff
@@ -508,8 +510,10 @@ func Test_spellsuggest_option_expr()
   \ .. "Change \"baord\" to:\n"
   \ .. " 1 \"BARD\"\n"
   \ .. " 2 \"BOARD\"\n"
-  \ .. " 3 \"BROAD\"\n"
-  \ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
+  "\ Nvim: Prompt message is sent to cmdline prompt.
+  "\ .. " 3 \"BROAD\"\n"
+  \ .. " 3 \"BROAD\"\n", a)
+  "\ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
```
zeertzjq, line 516:
```suggestion
  \ .. " 3 \"BROAD\"\n"
  "\ Nvim: Prompt message is sent to cmdline prompt.
  "\ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
  \ , a)
```

test/old/testdir/test_spell.vim
```diff
@@ -520,8 +524,10 @@ func Test_spellsuggest_option_expr()
   \ .. "Change \"baord\" to:\n"
   \ .. " 1 \"BARD\"                      (4 - 0)\n"
   \ .. " 2 \"BOARD\"                     (5 - 0)\n"
-  \ .. " 3 \"BROAD\"                     (5 - 0)\n"
-  \ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
+  "\ Nvim: Prompt message is sent to cmdline prompt.
+  "\ .. " 3 \"BROAD\"                     (5 - 0)\n"
+  \ .. " 3 \"BROAD\"                     (5 - 0)\n", a)
+  "\ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
```
zeertzjq, line 530:
```suggestion
  \ .. " 3 \"BROAD\"                     (5 - 0)\n"
  "\ Nvim: Prompt message is sent to cmdline prompt.
  "\ .. "Type number and <Enter> or click with the mouse (q or empty cancels): ", a)
  \ , a)
```

src/nvim/ex_getln.c
```diff
@@ -1852,6 +1852,18 @@ static int command_line_browse_history(CommandLineState *s)
   return CMDLINE_NOT_CHANGED;
 }
 
+static void put_int_on_cmdline(CommandLineState *s)
```
zeertzjq, line 1855: This function name is confusing. What does this have to do with `int`?

src/nvim/ex_getln.c
```diff
@@ -2181,19 +2207,18 @@ static int command_line_handle_key(CommandLineState *s)
...
 }
 
 static int command_line_not_changed(CommandLineState *s)
 {
+  // For one key prompt, put onto cmdline and exit.
+  if (ccline.one_key) {
+    put_int_on_cmdline(s);
+    return 0;
+  }
+
```
zeertzjq, line 2221: Doing this in `command_line_not_changed()` is a bit strange, as the effect of a control character in cmdline mode still takes effect before this is reached. For Ctrl-E it's not very obvious as it just moves the cursor, but the behavior may very strange if one presses some other keys (e.g Ctrl-R).

luukvbaal: Agreed, I thought some of the logic in `command_line_handle_key()` could still be useful/necessary, but we shouldn't do any of it for one key prompts. (Except `Esc` and `Ctrl_C`, to avoid putting their control codes onto the cmdline.)

zeertzjq:
Should mappings still be display at these prompts? Applying mappings is technically backward-incompatible, as if the user has some keys mapped in cmdline mode, they may have to press different keys.

src/nvim/ex_getln.c
```diff
@@ -2767,11 +2791,13 @@ char *getcmdline(int firstc, int count, int indent, bool do_concat FUNC_ATTR_UNU
 /// @param[in]  xp_context  Type of expansion.
 /// @param[in]  xp_arg  User-defined expansion argument.
 /// @param[in]  highlight_callback  Callback used for highlighting user input.
+/// @param[in]  one_key  Return after one key press for button prompt.
+/// @param[in]  mouse_used  Set to true when returning after right mouse click.
 ///
 /// @return [allocated] Command line or NULL.
 char *getcmdline_prompt(const int firstc, const char *const prompt, const int hl_id,
```
luukvbaal, line 2798: @bfredl `cmdline_show` events do not receive this prompt hl_id, so ext_cmdline cannot highlight the prompt accordingly. Is adding this hl_id as an additional parameter warranted?

folke:
fyi, I'll be on vacation till January 1, so would be great not to merge this till I'm back, since this will break some things in noice, which are easy to fix, but will need to be fixed. ty

luukvbaal:
Fine by me @folke. FYI (if we consider https://github.com/neovim/neovim/pull/31525#issuecomment-2541606319 to be resolved); after this PR you would no longer have to keep track of an internal message queue for prompt messages, scheduled messages will be displayed when Nvim enters the cmdline state (@przepompownia).

przepompownia:
After several hours of working without a queue on this PR, I did not encounter any of the previously observed issues, especially with `inputlist`.

src/nvim/ex_getln.c
luukvbaal: Added in #31661.

fredizzimo:
I tested this with Neovide, and didn't see any obvious problems, neither in practice nor when reviewing the code.  So, this LGTM.

test/functional/ui/cmdline_spec.lua
```diff
@@ -1485,31 +1485,28 @@ describe('cmdheight=0', function()
...
-    screen:expect {
-      grid = [[
+    screen:expect([[
       {2:foo}                      |
-      {1:~                        }|*3
-      {6:replace wi...q/l/^E/^Y)?}^ |
-    ]],
-    }
+      {3:                         }|
+                               |*2
```
luukvbaal, line 1499: This is a `:set cmdheight=0` issue that happens somewhere in the incsearch path. Whenever cmdheight == 0 and we type `/foo`, `?foo` or `:%s/foo` where foo exceeds the number of `Columns` we get a `cmdline/msg_row` conflict, resulting in the cleared row and extra scrolled line here. I think it's related to:
https://github.com/neovim/neovim/blob/adcd9360dfefc7b1e1edb0e86df460e074991c8d/src/nvim/message.c#L237-L240

I could attempt to fix this but that work will be in vain assuming we remove the message grid after #27855.

przepompownia: I probably had to reproduce and report the issue with multi-line commands and broken refresh (when switching between command line history e.g. by `<C-Up>`) a few days/weeks ago, but it wasn't necessarily related to this request. I am still not sure if you're talking about the same thing.

justinmk (approved):
Will wait for @folke / Jan 2.

Needs a news item, perhaps in the "UI" section of `news-features`. Or  add a "UI" section to news-breaking.

luukvbaal:
I think it fits the "EVENTS" section in `news-features`? Those items need updating anyhow for this change, and what is shown to the user is largely unchanged as a result of this change.

folke:
@luukvbaal was just looking into fixing noice, but I'm seeing some weird behavior.

I'm testing with:

```lua
vim.fn.confirm("Save changes?", "&Yes\n&No\n&Cancel")
```

When executing this:
- I receive a `msg_show` `confirm` event immediately
- No `cmdline_show` events yet
- Only after I type an invalid option, like `l`, I then get `cmdline_show` events with the prompt.

I assume the `cmdline_show` event should also come in immediately and right after the `msg_show`?

Additionally, it also seems no longer possible to know that the `cmdline_show` event is part of a `confirm`.
Previously I used that in noice, to show a different ui, with buttons based on the prompt text.

luukvbaal:
> I assume the cmdline_show event should also come in immediately and right after the msg_show?

It should, and that's precisely the behavior I'm seeing. I can try to reproduce with noice later.

> Additionally, it also seems no longer possible to know that the cmdline_show event is part of a confirm.
Previously I used that in noice, to show a different ui, with buttons based on the prompt text.

I think you can use the `cmdline_show->prompt` arg for that now.

folke:
Ok, thanks for confirming the behavior. Will see if I can create a repro without Noice.

folke:
And yes, noice shows the prompt, which is fine, but knowing it's a prompt triggered by a confirm would be a nice to have, but not that important.

folke:
Ok, figured it out. The `cmdline_show` is now shown when triggered from a keymap with `silent=true`.
Which makes sense I guess, but probably not wanted for `confirm` messages?

```lua
vim.keymap.set("n", "T", function()
	vim.fn.confirm("Save changes?", "&Yes\n&No\n&Cancel")
end, { silent = true })
```

luukvbaal:
>  Which makes sense I guess, but probably not wanted for confirm messages?

Yeah maybe, but that mapping yields the same behavior without ext_cmdline right?

folke:
After further testing, the behavior without `ui_attach`, so regular TUI also seems a bit off.

With my example, and without using a keymap and with `cmdheight=5` only the `save changes?` is printed, not the `yes/no` message.

> Yeah maybe, but that mapping yields the same behavior without ext_cmdline right?
Nope, without `ext_cmdline`, it shows just the `yes/no` with `silent=true`.

luukvbaal:
Yeah I meant the same broken behavior. That is a regression that should be fixed.

PR #31534, opened 2024-12-10
vim-patch: line shifting fixes

zeertzjq:
#### vim-patch:9.0.2112: [security]: overflow in shift_line

Problem:  [security]: overflow in shift_line
Solution: allow a max indent of INT_MAX

[security]: overflow in shift_line

When shifting lines in operator pending mode and using a very large
value, we may overflow the size of integer. Fix this by using a long
variable, testing if the result would be larger than INT_MAX and if so,
indent by INT_MAX value.

Special case: We cannot use long here, since on 32bit architectures (or
on Windows?), it typically cannot take larger values than a plain int,
so we have to use long long count, decide whether the resulting
multiplication of the shiftwidth value * amount is larger than INT_MAX
and if so, we will store INT_MAX as possible larges value in the long
long count variable.

Then we can safely cast it back to int when calling the functions to set
the indent (set_indent() or change_indent()). So this should be safe.

Add a test that when using a huge value in operator pending mode for
shifting, we will shift by INT_MAX

closes: vim/vim#13535

https://github.com/vim/vim/commit/6bf131888a3d1de62bbfa8a7ea03c0ddccfd496e

Skip the test for now, as it takes too long and require other changes.

#### vim-patch:9.0.2113: Coverity warns for another overflow in shift_line()

Problem:  Coverity warns for another overflow in shift_line()
Solution: Test for INT_MAX after the if condition, cast integer values
          to (long long) before multiplying.

https://github.com/vim/vim/commit/22a97fc241361aa91bda84e5344d5b7c0cda3e81

#### vim-patch:9.0.2122: [security]: prevent overflow in indenting

Problem:  [security]: prevent overflow in indenting
Solution: use long long and remove cast to (int)

The shiftwidth option values are defined as being long. However, when
calculating the actual amount of indent, we cast down to (int), which
may cause the shiftwidth value to become negative and later it may even
cause Vim to try to allocate a huge amount of memory.

We already use long and long long variable types to calculate the indent
(and detect possible overflows), so the cast to (int) seems superfluous
and can be safely removed. So let's just remove the (int) cast and
calculate the indent using longs.

Additionally, the 'shiftwidth' option value is also used when determining
the actual 'cino' options. There it can again cause another overflow, so
make sure it is safe in parse_cino() as well.

closes: vim/vim#13555

https://github.com/vim/vim/commit/3770574e4a70e810add9929973c51f9070c8c851

#### vim-patch:9.0.2124: INT overflow detection logic can be simplified

Problem:  INT overflow logic can be simplified
Solution: introduce trim_to_int() function

closes: vim/vim#13556

https://github.com/vim/vim/commit/2b0882fa6555b4d0197afbdfc32a4533cf6aacf4

vim-patch:9.0.2138: Overflow logic requires long long

Problem:  Overflow logic requires long long
Solution: Define vimlong_T data type to make life easier
          for porters

closes: vim/vim#13598

https://github.com/vim/vim/commit/fda700cb04612fe2f9301a9ba820309175decabf

Cherry-pick ops.c change from patch 9.1.0608.

Co-authored-by: Ernie Rael <errael@raelity.com>

#### vim-patch:9.1.0917: various vartabstop and shiftround bugs when shifting lines

Problem:  various vartabstop and shiftround bugs when shifting lines
Solution: Fix the bugs, add new tests for shifting lines in various ways
          (Gary Johnson)

closes: vim/vim#16193

https://github.com/vim/vim/commit/eed63f96d26723ff31a9728647eed526d06a553d

Co-authored-by: Gary Johnson <garyjohn@spocom.com>

### DIFF
--- a/src/nvim/math.c
+++ b/src/nvim/math.c
@@ -106,3 +106,9 @@ int vim_append_digit_int(int *value, int digit)
   *value = x * 10 + digit;
   return OK;
 }
+
+/// Return something that fits into an int.
+int trim_to_int(int64_t x)
+{
+  return x > INT_MAX ? INT_MAX : x < INT_MIN ? INT_MIN : (int)x;
+}
--- a/src/nvim/indent_c.c
+++ b/src/nvim/indent_c.c
@@ -1689,7 +1689,7 @@ void parse_cino(buf_T *buf)
       p++;
     }
     char *digits_start = p;   // remember where the digits start
-    int n = getdigits_int(&p, true, 0);
+    int64_t n = getdigits_int(&p, true, 0);
     divider = 0;
     if (*p == '.') {        // ".5s" means a fraction.
       fraction = atoi(++p);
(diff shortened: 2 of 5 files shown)


src/nvim/math.c
```diff
@@ -106,3 +106,9 @@ int vim_append_digit_int(int *value, int digit)
   *value = x * 10 + digit;
   return OK;
 }
+
+/// Return something that fits into an int.
+int trim_to_int(int64_t x)
```
justinmk, line 111: Is "truncate" a more conventional verb for this?

PR #31539, opened 2024-12-10
refactor(wininfo): change wininfo from a linked list to an array

bfredl:
"wininfo" is going to be my next victim. The main problem with wininfo is that it is "all or nothing", i e either all state about a buffer in a window is considered valid or none of it is. This needs to be fixed to address some long running grievances.

For now this is just a warmup: refactor it from a linked list to a vector.

### DIFF
--- a/src/klib/kvec.h
+++ b/src/klib/kvec.h
@@ -135,6 +135,10 @@
          : 0UL)), \
      &(v).items[(i)]))
 
+#define kv_shift(v, i, n) ((v).size -= (n), (i) < (v).size \
+                           && memmove(&kv_A(v, (i)), &kv_A(v, (i)+(n)), \
+                                      ((v).size-(i))*sizeof(kv_A(v, i))))
+
 #define kv_printf(v, ...) kv_do_printf(&(v), __VA_ARGS__)
 
 /// Type of a vector with a few first members allocated on stack
--- a/src/nvim/window.c
+++ b/src/nvim/window.c
@@ -5175,8 +5175,8 @@ win_T *win_alloc(win_T *after, bool hidden)
   return new_wp;
 }
 
-// Free one wininfo_T.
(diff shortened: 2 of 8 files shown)


justinmk (approved):

src/klib/kvec.h
```diff
@@ -135,6 +135,9 @@
          : 0UL)), \
      &(v).items[(i)]))
 
+#define kv_shift(v, i) ((v).size -= 1, (i) < (v).size && memmove(&kv_A(v, (i)), &kv_A(v, (i)+1), \
+                                                                 ((v).size-(i))*sizeof(kv_A(v, i))))
```
zeertzjq, line 139: Maybe this should be named `kv_shift1()`, or it should take an argument specifying how far to shift?

zeertzjq (commented):
The `FOR_ALL_BUF_WININFO()` defined in `globals.h` is now no longer valid.

src/nvim/window.c
```diff
@@ -5235,30 +5235,22 @@ void win_free(win_T *wp, tabpage_T *tp)
...
-    for (wininfo_T *wip = buf->b_wininfo; wip != NULL; wip = wip->wi_next) {
+    WinInfo *wip_wp = NULL;
+    size_t pos_null = kv_size(buf->b_wininfo);
+    for (size_t i = 0; i < kv_size(buf->b_wininfo); i++) {
+      WinInfo *wip = kv_A(buf->b_wininfo, i);
       if (wip->wi_win == wp) {
-        wininfo_T *wip2;
-
-        // If there already is an entry with "wi_win" set to NULL it
-        // must be removed, it would never be used.
```
zeertzjq, line 5243: Why is this comment removed?

zeertzjq: I think it should be moved above the `pos_null < kv_size(buf->b_wininfo)` check.

PR #31541, opened 2024-12-10
feat(runtime): add support for switching between if-else and switch statements in C files using `%`

ingenarel:
Problem:
even though you can switch between if else statements using matchit in a lot of files, you currently can't do that in c, even though it should be a simple regex

Solution:
extend the vim.b.match_words option in the c filetype plugin so that it includes the regexes for switching if else and switch statements.


clason:
Thank you, but we're not interested in special-casing this in every `ftplugin` (which, in principle, remain maintained by Vim). Instead, the plan is to replace `matchit` with a treesitter-compatible equivalent.

clason:
You can try contributing this to the Vim `ftplugin`, if it makes sense there -- in this case, we'll port it as a matter of course.

PR #31546, opened 2024-12-11
Added missing return types for the auto-generated Lua documentation

ColinKennedy:
A number of Lua annotations are missing `---@return` annotations, this PR is a pass at adding them.

I didn't add everything because of some reason or another. If curious, please see the sections below

<details>
<summary>Some type annotations should use `---@generic` - but out of scope for this PR, IMO</summary>

LuaLS supports "generic" types. See
https://github.com/LuaLS/lua-language-server/wiki/Annotations#generic

For example `:h deepcopy` takes argument `deepcopy({expr} [, {noref}])` and
returns a copy of it. So the function should ideally not return `any` like it
currently does but instead return the type of whatever `{expr}` is.

`scripts/gen_eval_files.lua` would probably need to be updated to handle that
and I considered it out of scope for this PR.

Here some functions that could benefit from `---@generic`.

- `:h deepcopy`
- `:h get`
- `:h items`
- `:h keys`
- `:h max`
- `:h min`
- `:h reduce`
- `:h sort`
- `:h values`

A future PR for these would be good.
</details>

<details>
<summary>Unsure how to handle primitive literals</summary>

A number of functions have explicit return types like "returns 0 on success or -1 on
failure". (Many) others are "return 0 as FALSE and 1 as TRUE". Some do the opposite. And
rare functions even have non-specific returns like `:h win_splitmove` that returns 0 for
success but non-zero for failure (there's no specificly defined failure value)

Anyway I was inclined to change the return values to be literal like
`---@returns 0 | 1` where-appropriate but I didn't want to assume anything. So
I opted out on this PR

For reference, these are the functions I found so far:

- `did_filetype` - 0 == FALSE, 1 == TRUE
- `inputrestore` - "Returns TRUE when there is nothing to restore, FALSE otherwise."
    - TRUE means "did nothing" and FALSE means "did something"
- `inputsave` - "Returns TRUE when out of memory, FALSE otherwise."
    - This means that unlike the others, TRUE actually is a fail condition
- `jobstart` - 0 / -1 == failure, 1-or-more if success
    - I guess in this one case `integer` would actually be correct
- `mkdir` - 0 == FALSE, 1 == TRUE
- `serverstop` - 0 == FALSE, 1 == TRUE
- `setbufline` - 1 == fail, 0 == success
- `setqflist` - 0 == success -1 == failure
- `wildmenumode` - 0 == FALSE, 1 == TRUE
- `win_move_separator` - 0 == FALSE, 1 == TRUE
- `win_move_statusline` - 0 == FALSE, 1 == TRUE
- `win_splitmove` - "Returns zero for success, non-zero for failure."
- `writefile` - 0 == success -1 == failure
</details>

<details>
<summary>Recommended To Use Custom Types</summary>

Some functions could use a proper `---@class` definition. For example `:h
vim.api.nvim_get_hl()` returns `vim.api.keyset.get_hl_info`.

I assumed I would define those in `vim/_meta/api_keysets_extra.lua` but didn't
know if there were preferred naming conventions. So I figured a follow-up PR for
that would be best.

These are the functions that could use custom class types, assuming an
equivalent doesn't already exist:

- `:h getcellwidths()`
- `:h getcursorcharpos()`
- `:h getmatches()`
- `:h gettagstack()`
- `:h menu_get()`
- `:h setcellwidths()`
- `:h taglist()`
- `:h timer_info()`
- `:h win_id2tabwin()`
- `:h win_screenpos()`
- `:h wordcount()`
</details>

<details>
<summary>Not Sure</summary>

All situations where I wasn't sure of the type (and an explanation of why)

- `digraph_get` - technically the type is char but lua has only string. So I guess just
  string?
- `funcref` is another weird one
- `sqrt()` can return a number or NaN. Would we define a fake NaN class and include that as a return type. e.g. `number | NaN`?
</details>

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1407,7 +1407,7 @@ ctxset({context} [, {index}])                                         *ctxset()*
                   • {index} (`integer?`)
 
                 Return: ~
-                  (`any`)
+                  (`integer`)
 
 ctxsize()                                                            *ctxsize()*
 		Returns the size of the |context-stack|.
@@ -1618,7 +1618,7 @@ did_filetype()                                                  *did_filetype()*
 		file.
 
                 Return: ~
-                  (`any`)
+                  (`integer`)
 
 diff_filler({lnum})                                              *diff_filler()*
 		Returns the number of filler lines above line {lnum}.
@@ -1633,7 +1633,7 @@ diff_filler({lnum})                                              *diff_filler()*
                   • {lnum} (`integer`)
 
                 Return: ~
-                  (`any`)
+                  (`integer`)
 
 diff_hlID({lnum}, {col})                                           *diff_hlID()*
 		Returns the highlight ID for diff mode at line {lnum} column
@@ -1674,7 +1674,7 @@ digraph_get({chars})                                       *digraph_get()* *E121
                   • {chars} (`string`)
 
                 Return: ~
-                  (`any`)
+                  (`string`)
 
 digraph_getlist([{listall}])                                 *digraph_getlist()*
 		Return a list of digraphs.  If the {listall} argument is given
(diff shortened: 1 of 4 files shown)


justinmk (approved):
Thanks!

PR #31549, opened 2024-12-11
fix(api): support get autocmds by id in api

### LINKED ISSUE #31531: nvim_get_autocmds: allow get by id
### Problem

Currently `nvim_get_autocmds` does not allow filter by id returned by `nvim_create_autocmd`. 

### Expected behavior

`id` option allows to do it without workarounds.

glepnir:
Close #31531

Problem: Can not use id as filter condition to get autocmds.

Solution: add id which support integer or array to filter get autocmds

PS: test and doc later

### DIFF
--- a/src/nvim/api/autocmd.c
+++ b/src/nvim/api/autocmd.c
@@ -68,32 +68,31 @@ static int64_t next_autocmd_id = 1;
 /// match any combination of them.
 ///
 /// @param opts Dict with at least one of the following:
-///             - group (string|integer): the autocommand group name or id to match against.
-///             - event (string|array): event or events to match against |autocmd-events|.
-///             - pattern (string|array): pattern or patterns to match against |autocmd-pattern|.
-///             Cannot be used with {buffer}
-///             - buffer: Buffer number or list of buffer numbers for buffer local autocommands
+///             - buffer: (integer) Buffer number or list of buffer numbers for buffer local autocommands
 ///             |autocmd-buflocal|. Cannot be used with {pattern}
+///             - event: (string|table) event or events to match against |autocmd-events|.
+///             - id: (integer) Autocommand ID to match.
+///             - group: (string|table) the autocommand group name or id to match against.
+///             - pattern: (string|table) pattern or patterns to match against |autocmd-pattern|.
+///             Cannot be used with {buffer}
 /// @return Array of autocommands matching the criteria, with each item
 ///             containing the following fields:
-///             - id (number): the autocommand id (only when defined with the API).
-///             - group (integer): the autocommand group id.
-///             - group_name (string): the autocommand group name.
-///             - desc (string): the autocommand description.
-///             - event (string): the autocommand event.
-///             - command (string): the autocommand command. Note: this will be empty if a callback is set.
-///             - callback (function|string|nil): Lua function or name of a Vim script function
+///             - buffer: (integer) the buffer number.
+///             - buflocal: (boolean) true if the autocommand is buffer local.
+///             - command: (string) the autocommand command. Note: this will be empty if a callback is set.
+///             - callback: (function|string|nil): Lua function or name of a Vim script function
 ///               which is executed when this autocommand is triggered.
-///             - once (boolean): whether the autocommand is only run once.
-///             - pattern (string): the autocommand pattern.
+///             - desc: (string) the autocommand description.
+///             - event: (string) the autocommand event.
+///             - id: (integer) the autocommand id (only when defined with the API).
+///             - group: (integer) the autocommand group id.
+///             - group_name: (string) the autocommand group name.
+///             - once: (boolean) whether the autocommand is only run once.
+///             - pattern: (string) the autocommand pattern.
 ///               If the autocommand is buffer local |autocmd-buffer-local|:
-///             - buflocal (boolean): true if the autocommand is buffer local.
-///             - buffer (number): the buffer number.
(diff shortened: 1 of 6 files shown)


src/nvim/api/autocmd.c
```diff
@@ -74,6 +74,7 @@ static int64_t next_autocmd_id = 1;
 ///             Cannot be used with {buffer}
 ///             - buffer: Buffer number or list of buffer numbers for buffer local autocommands
 ///             |autocmd-buflocal|. Cannot be used with {pattern}
+///             - id: (integer|array) autocmd ids
```
justinmk, line 77: Is there a strong need to support array? If someone has an id they usually just want 1 result.

justinmk: keeping the code simple (no array support) is probably worth it for now

glepnir: okay 👍

src/nvim/api/autocmd.c
```diff
@@ -72,8 +72,9 @@ static int64_t next_autocmd_id = 1;
 ///             - event (string|array): event or events to match against |autocmd-events|.
 ///             - pattern (string|array): pattern or patterns to match against |autocmd-pattern|.
 ///             Cannot be used with {buffer}
-///             - buffer: Buffer number or list of buffer numbers for buffer local autocommands
+///             - buffer number: Buffer number or list of buffer numbers for buffer local autocommands
 ///             |autocmd-buflocal|. Cannot be used with {pattern}
+///             - id number: Autocommand ID to match.
```
justinmk, line 77: while we're here, sort these key names? they are in random order.

src/nvim/api/autocmd.c
```diff
@@ -127,6 +128,11 @@ Array nvim_get_autocmds(Dict(get_autocmds) *opts, Arena *arena, Error *err)
     });
   }
 
+  int filter_id = -1;
```
justinmk, line 131:
```suggestion
  int id = -1;
```

src/nvim/api/autocmd.c
```diff
@@ -127,6 +128,11 @@ Array nvim_get_autocmds(Dict(get_autocmds) *opts, Arena *arena, Error *err)
...
+  if (HAS_KEY(opts, get_autocmds, id)) {
+    filter_id = (int)opts->id;
```
justinmk, line 133: ternary?

test/functional/api/autocmd_spec.lua
```diff
@@ -881,6 +881,93 @@ describe('autocmd api', function()
...
+      local _id, _id1
+      after_each(function()
+        api.nvim_del_autocmd(_id)
+        if _id1 then
+          api.nvim_del_autocmd(_id1)
+          _id1 = nil
+        end
+      end)
+
+      it('retrieves events using an ID', function()
```
justinmk, line 895: "retrieves" is an overly verbose word
```suggestion
      it('gets events by ID', function()
```

src/nvim/api/autocmd.c
```diff
@@ -127,6 +128,11 @@ Array nvim_get_autocmds(Dict(get_autocmds) *opts, Arena *arena, Error *err)
     });
```
justinmk, line 128: there is a TODO on line 96 that can be deleted now

test/functional/api/autocmd_spec.lua
```diff
@@ -881,6 +881,93 @@ describe('autocmd api', function()
...
       end)
     end)
+
+    describe('id', function()
+      local _id, _id1
+      after_each(function()
+        api.nvim_del_autocmd(_id)
+        if _id1 then
+          api.nvim_del_autocmd(_id1)
+          _id1 = nil
```
justinmk, line 891: why is this needed? is clear() not called for each test?

glepnir: Aha thanks.

src/nvim/api/autocmd.c
zeertzjq: Having the type immediately after the field name without any separation is strange. It's better to put the type in parentheses. For example, the docstring from `nvim_get_hl` `opts` looks like this:
```c
/// @param opts  Options dict:
///                 - name: (string) Get a highlight definition by name.
///                 - id: (integer) Get a highlight definition by id.
///                 - link: (boolean, default true) Show linked group name instead of effective definition |:hi-link|.
///                 - create: (boolean, default true) When highlight group doesn't exist create it.
```

glepnir: make sense.

glepnir: also update the type  use integer and table instead of number array.

przepompownia:
Works for me :)

test/functional/api/autocmd_spec.lua
```diff
@@ -899,6 +899,89 @@ describe('autocmd api', function()
...
+            command = 'echo "hello"',
+            event = 'BufEnter',
+            id = id,
+            once = false,
+            pattern = '*',
+          },
+        }, api.nvim_get_autocmds({ id = id }))
+      end)
+
+      it('gets events by ID and group', function()
```
justinmk, line 920: let's just have 1 test for the "with other filters" case. we don't need to test every permutation. you can merge some of the filters from the below tests into this test, if possible, then drop the tests below.

```suggestion
      it('gets events by ID and other filters', function()
```

justinmk:
LGTM after condensing the tests. Why in draft?

justinmk (approved):

PR #31551, opened 2024-12-11
feat(vim.fs): find(), dir() can "follow" symlinks

mike325:
This PR introduce a new option for `vim.fs.dir` and `vim.fs.find` to follow symbolic links.
The default behaviour now turns on the `follow` feature, users can opt out of it with `follow=false`

### DIFF
--- a/test/functional/lua/fs_spec.lua
+++ b/test/functional/lua/fs_spec.lua
@@ -217,12 +217,44 @@ describe('vim.fs', function()
   end)
 
   describe('find()', function()
+    before_each(function()
+      vim.uv.fs_symlink(
+        test_source_path .. '/build',
+        test_source_path .. '/build_link',
+        { junction = true, dir = true }
+      )
+    end)
+
+    after_each(function()
+      vim.uv.fs_unlink(test_source_path .. '/build_link')
+    end)
+
     it('works', function()
       eq(
         { test_build_dir .. '/build' },
         vim.fs.find('build', { path = nvim_dir, upward = true, type = 'directory' })
       )
       eq({ nvim_prog }, vim.fs.find(nvim_prog_basename, { path = test_build_dir, type = 'file' }))
+
+      eq(
+        { nvim_prog, test_source_path .. '/build_link/bin/' .. nvim_prog_basename },
+        vim.fs.find(nvim_prog_basename, {
+          path = test_source_path,
+          type = 'file',
+          limit = 2,
+          follow = true,
+        })
+      )
+
+      eq(
+        { nvim_prog },
+        vim.fs.find(nvim_prog_basename, {
+          path = test_source_path,
+          type = 'file',
+          limit = 2,
+          follow = false,
+        })
+      )
 
       local parent, name = nvim_dir:match('^(.*/)([^/]+)$')
       eq({ nvim_dir }, vim.fs.find(name, { path = parent, upward = true, type = 'directory' }))
--- a/test/functional/lua/fs_spec.lua
+++ b/test/functional/lua/fs_spec.lua
@@ -152,7 +152,7 @@ describe('vim.fs', function()
       )
     end)
 
-    it('works with opts.depth and opts.skip', function()
+    it('works with opts.depth, opts.skip and opts.follow', function()
       io.open('testd/a1', 'w'):close()
       io.open('testd/b1', 'w'):close()
       io.open('testd/c1', 'w'):close()
@@ -166,7 +166,7 @@ describe('vim.fs', function()
       io.open('testd/a/b/c/b4', 'w'):close()
       io.open('testd/a/b/c/c4', 'w'):close()
 
-      local function run(dir, depth, skip)
+      local function run(dir, depth, skip, follow)
         return exec_lua(function()
           local r = {} --- @type table<string, string>
           local skip_f --- @type function
@@ -177,7 +177,7 @@ describe('vim.fs', function()
               end
             end
           end
-          for name, type_ in vim.fs.dir(dir, { depth = depth, skip = skip_f }) do
+          for name, type_ in vim.fs.dir(dir, { depth = depth, skip = skip_f, follow = follow }) do
             r[name] = type_
           end
           return r
@@ -197,6 +197,7 @@ describe('vim.fs', function()
       exp['a/b2'] = 'file'
       exp['a/c2'] = 'file'
       exp['a/b'] = 'directory'
+      local lexp = vim.deepcopy(exp)
 
       eq(exp, run('testd', 2))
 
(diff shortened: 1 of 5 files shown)


justinmk (approved):

dundargoc:
nitpick: I think there's a risk of infinite recursion if the symbolic link is circular. To handle this, vimscript's `resolve()` just aborts after 100 times

> To cope with link cycles, resolving of symbolic links is stopped after 100 iterations.

I think it should also be possible to detect if it's a circular loop and abort in that case (I have vague memories of some path libraries going this route, but not swearing on it).


For what it's worth I think the PR is fine as is currently, mentioning it mostly for awareness or if it's an edge case we want to handle in the future.

mike325:
> nitpick: I think there's a risk of infinite recursion if the symbolic link is circular. Vimscript's `resolve()` handles this with the following behavior:
> 
> > To cope with link cycles, resolving of symbolic links is stopped after 100 iterations.
> 
> I think it should also be possible to detect if it's a circular loop and abort in that case (I have vague memories of some path libraries going this route, but not swearing on it).
> 
> For what it's worth I think the PR is fine as is currently, mentioning it mostly for awareness or if it's an edge case we want to handle in the future.

I also thought about the infinite recursion problem since I myself encountered this issue in a repo I was working on, although I found it using `fd|rg`.

dundargoc:
I tested out this PR, there is some already some internal mechanism that prevents infinite recursion but I don't know what. I can "only" recurse 33 times before it aborts: `/Users/dundargoc/programs/neovim/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link/link`

test/functional/lua/fs_spec.lua
```diff
@@ -213,17 +214,59 @@ describe('vim.fs', function()
...
+        vim.fs.find(nvim_prog_basename, {
+          path = test_source_path,
+          type = 'file',
+          limit = 2,
+          follow = true,
+        })
+      )
+
+      eq(
+        { nvim_prog },
```
justinmk, line 261: there's already a `describe('find()'` block, and no setup in this `it()`, so it makes sense to move these 2 `eq()` assertions into a separate `it('follow=true'` block. That makes it clear what these are testing, at zero code/verbosity-cost.

mike325:
Should this change be included in the `new.txt` file ? 
Also, is it ok to have it `follow=true`? or should it keep the old behavior by using  `follow=false` by default instead?

justinmk:
is there a test for the infinite loop case?

> Should this change be included in the `new.txt` file ?

definitely. probably the "Lua" section of news-features.

justinmk:
> Also, is it ok to have it `follow=true`? or should it keep the old behavior by using `follow=false` by default instead?

if we have a test for the infinite-loop case then defaulting to true seems fine.

justinmk:
> a test for the infinite-loop case then defaulting to true seems fine.
> 
> should be a separate test case like `it('follow=true handles symlink loop')`

the test can be skipped on windows for now, if that makes it easier

test/testutil.lua
```diff
@@ -394,15 +394,18 @@ end
...
-  if not (s == 'win' or s == 'mac' or s == 'freebsd' or s == 'openbsd' or s == 'bsd') then
+  if
+    not (s == 'win' or s == 'mac' or s == 'linux' or s == 'freebsd' or s == 'openbsd' or s == 'bsd')
+  then
     error('unknown platform: ' .. tostring(s))
   end
   return not not (
     (s == 'win' and (sysname:find('windows') or sysname:find('mingw')))
     or (s == 'mac' and sysname == 'darwin')
+    or (s == 'linux' and sysname == 'linux')
```
justinmk, line 408: can name it `unix`, and just assume that if nothing else matches then we treat it as `unix`

test/functional/lua/fs_spec.lua
```diff
@@ -166,8 +179,8 @@ describe('vim.fs', function()
       io.open('testd/a/b/c/b4', 'w'):close()
       io.open('testd/a/b/c/c4', 'w'):close()
 
-      local function run(dir, depth, skip)
-        return exec_lua(function()
+      local function run(dir, depth, skip, follow)
+        return exec_lua(function(follow_sym)
```
justinmk, line 183: nit: usual convention is something like:
```suggestion
        return exec_lua(function(follow_)
```

test/functional/lua/fs_spec.lua
```diff
@@ -17,6 +17,19 @@ local mkdir = t.mkdir
...
+  mac = 33,
+  win = 64,
+}
+
+local link_limit = platfom_link_limit.linux
+if is_os('win') then
+  link_limit = platfom_link_limit.win
+elseif is_os('mac') or is_os('bsd') then
+  link_limit = platfom_link_limit.mac
+end
```
justinmk, line 31:
```suggestion
local link_limit = is_os('win') and 64
    or (is_os('mac') or is_os('bsd')) and 33 or 41
```

justinmk (approved):

test/testutil.lua
```diff
@@ -405,10 +408,10 @@ end
 
 local architecture = uv.os_uname().machine
 
---- @param s 'x86_64'|'arm64'
+--- @param s 'x86_64'|'arm64'|'aarch64'
 --- @return boolean
 function M.is_arch(s)
-  if not (s == 'x86_64' or s == 'arm64') then
+  if not (s == 'x86_64' or s == 'arm64' or s == 'aarch64') then
```
justinmk, line 414: why is "aarch64" needed? not used in this PR.

and generally, we should have one token that means "arm", and then `is_arch` (or whatever calls it) should do the right thing for each platform.

mike325: oooo sorry about that, I was testing the linux side using a raspberry pi and a desktop, I wanted to know if the limit was different between x86 and ARM.
linux reports `aarch64` not `arm64` (or at least reaspbian does), I can remove it since it turned out not to be needed.

justinmk: ok let's remove it?

justinmk:
LGTM, waiting for `is_arch` change to be dropped

justinmk:
> > Also, is it ok to have it `follow=true`? or should it keep the old behavior by using `follow=false` by default instead?
> 
> if we have a test for the infinite-loop case then defaulting to true seems fine.

We probably need the default to be `follow=false`. There are some reports of "performance regressions" from `follow=true`.

mike325:
> We probably need the default to be `follow=false`. There are some reports of "performance regressions" from `follow=true`.
Do you want me to create an issue to track this @justinmk  ?

justinmk:
@mike325 i would prefer a PR :) we are release 0.11 soon, so it's urgent

PR #31553, opened 2024-12-11
fix: ensure we've selected the diagnostics quickfix list

### LINKED ISSUE #31540: LSP: vim.diagnostic.setqflist does not work after usage of vim.lsp.buf.document_symbol
### Problem

`vim.diagnostic.setqflist` stop make any changes to `quickfix list` after usage of `vim.lsp.buf.document_symbol`

1. run `vim.diagnostic.setqflist`: it shows diagnostic in quickfix list
2. run `vim.lsp.buf.document_symbol`: it shows symbols
3. run `vim.diagnostic.setqflist`: still shows symbols

But  e.g. `:grep` works fine.

https://github.com/user-attachments/assets/8684f56a-d093-44c0-89a0-1cb085034276

1. run `vim.diagnostic.setqflist`: it shows diagnostic in quickfix list
2. run `vim.lsp.buf.document_symbol`: it shows symbols
3. run `vim.diagnostic.setqflist`: still shows symbols

**minimal_init.lua**
```lua
local pattern = 'lua'
local root_markers = { '.' }

  pattern = pattern,
  callback = function(args)
    vim.lsp.start({
      name = 'bugged-ls',
      cmd = cmd,
      root_dir = root_dir,
      settings = settings
    })
  end
})
```

As far as I understand the problem is in **share/nvim/runtime/lua/vim/lsp/diagnostic.lua** `function set_list`. It has next code:
```lua
    -- If we already have a diagnostics quickfix, update it rather than creating a new one.
    -- This avoids polluting the finite set of quickfix lists, and preserves the currently selected
    -- entry.
    vim.fn.setqflist({}, _qf_id and 'u' or ' ', {
      title = title,
      items = items,
      id = _qf_id,
    })
```
Changing `vim.fn.setqflist({}, _qf_id and 'u' or ' ',` to just `vim.fn.setqflist({}, ' ',` fix this issue.

### Expected behavior

`vim.diagnostic.setqflist` should put diagnostick in quickfix list

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1334+gca760e645b

lua-language-server but probably any

### Operating system/version

6.12.3-arch1-1 Hyprland 0.45.2 built from branch  at commit 12f9a0d0b93f691d4d9923716557154d74777b0a 

### Log file

_No response_

jfly:
Previously, when updating the quickfix diagnostics list, we'd update it, and then open the quickfix buffer, but there was no guarantee that the quickfix buffer would be displaying the quickfix diagnostics list (it could very possibly be displaying some other quickfix list!).

This fixes things so we first select the quickfix list before opening the quickfix buffer. If `open` is not specified, the behavior is the same as before: we update the diagnostics quickfix list, but do not navigate to it.

This fixes https://github.com/neovim/neovim/issues/31540.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -865,7 +865,16 @@ local function set_list(loclist, opts)
     end
   end
   if open then
-    api.nvim_command(loclist and 'lwindow' or 'botright cwindow')
+    if loclist then
+      api.nvim_command('lwindow')
+    else
+      -- First navigate to the diagnostics quickfix list.
+      local nr = vim.fn.getqflist({ id = _qf_id, nr = 0 }).nr
+      api.nvim_command(nr .. 'chistory')
+
+      -- Now open the quickfix list.
+      api.nvim_command('botright cwindow')
+    end
   end
 end
 


runtime/lua/vim/diagnostic.lua
```diff
@@ -865,7 +865,16 @@ local function set_list(loclist, opts)
     end
   end
   if open then
-    api.nvim_command(loclist and 'lwindow' or 'botright cwindow')
+    if loclist then
+      api.nvim_command('lwindow')
+    else
+      -- First navigate to the diagnostics quickfix list.
+      local nr = vim.fn.getqflist({ id = _qf_id, nr = 0 }).nr
+      api.nvim_command(nr .. 'chi')
```
justinmk, line 873: best practice to use full names in code (helps with grep, etc)

```suggestion
      api.nvim_command(nr .. 'chistory')
```

jfly: Fixed! I actually did look for the full name, but am not very good at reading vim docs. `:help chi` lead me to:

```
:[count]chi[story]      Show the list of error lists.  The current list is
```

Which I glossed over as "optional count prefix" and "optional story suffix". Oops!

zeertzjq:
link failure

runtime/lua/vim/diagnostic.lua
```diff
@@ -865,7 +865,16 @@ local function set_list(loclist, opts)
     end
   end
   if open then
-    api.nvim_command(loclist and 'lwindow' or 'botright cwindow')
+    if loclist then
+      api.nvim_command('lwindow')
+    else
+      -- First navigate to the diagnostics quickfix list.
+      local nr = vim.fn.getqflist({ id = _qf_id, nr = 0 }).nr
```
zeertzjq, line 872: It should be `qf_id` after #31486

jfly: Oops, see https://github.com/neovim/neovim/pull/31557

PR #31557, opened 2024-12-12
fix: broken variable reference

jfly:
Unfortunately https://github.com/neovim/neovim/pull/31553 rebased cleanly on top of https://github.com/neovim/neovim/pull/31486, but still introduced a bug. Thanks to @zeertzjq for noticing this.

If someone can point me in the right direction, I'd be happy to add a test for this as penance.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -854,10 +854,11 @@ local function set_list(loclist, opts)
   -- numbers beyond the end of the buffer
   local diagnostics = get_diagnostics(bufnr, opts --[[@as vim.diagnostic.GetOpts]], false)
   local items = M.toqflist(diagnostics)
+  local qf_id = nil
   if loclist then
     vim.fn.setloclist(winnr, {}, 'u', { title = title, items = items })
   else
-    local qf_id = get_qf_id_for_title(title)
+    qf_id = get_qf_id_for_title(title)
 
     -- If we already have a diagnostics quickfix, update it rather than creating a new one.
     -- This avoids polluting the finite set of quickfix lists, and preserves the currently selected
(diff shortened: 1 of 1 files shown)


justinmk:
Luacheck caught this but I merged too quickly. Sorry everyone...

justinmk:
> I'd be happy to add a test for this as penance.

This looks like the right place to add a test: https://github.com/neovim/neovim/blob/ca760e645ba4d1fdb0b6fff3ac98231c3d683306/test/functional/lua/diagnostic_spec.lua#L3296

 You can [run the test](https://github.com/neovim/neovim/blob/master/test/README.md#filter-by-file) like this:

    TEST_FILE=test/functional/lua/diagnostic_spec.lua make functionaltest

delatorrejuanchi:
Hello! :wave:

It seems to me this change may have introduced a different bug.

If `open` is `true` and there is no existing diagnostics quickfix list (i.e.: `qf_id` is `nil`), then we'll attempt to open the location list instead ([ref](https://github.com/neovim/neovim/blob/433b342baa04b35fb1ab24d4ef38f126d12ea714/runtime/lua/vim/diagnostic.lua#L874-L883)). After calling this method again `qf_id` will not be `nil` again given that a diagnostics quickfix list would have been created by the previous call, so it'll open as expected.

Perhaps I'm missing something, but it looks like checking for `not loclist` instead of `qf_id` should fix it?

```lua
  if open then
    if not loclist then
      -- First navigate to the diagnostics quickfix list.
      local nr = vim.fn.getqflist({ id = qf_id, nr = 0 }).nr
      api.nvim_command(nr .. 'chistory')

      -- Now open the quickfix list.
      api.nvim_command('botright cwindow')
    else
      api.nvim_command('lwindow')
    end
  end
```

justinmk:
@delatorrejuanchi PR welcome. If you have bandwidth to add a test in the location mentioned above, that will also help a lot here :)

delatorrejuanchi:
I've opened #31585 to address this, including some tests for `vim.diagnostic.setqflist`. Some (unrelated?) checks seem to be failing though, and I'm uncertain about how to re-run them in case they are just flaky.

PR #31558, opened 2024-12-12
docs(annotations): added `---@generic` support

ColinKennedy:
Refined the auto-generated functions that need `---@generic`. For example `deepcopy()`, which currently returns `any`.

There are other functions that could benefit from this but I couldn't prove that the changes worked. I think lua-language-server itself might have some bugs related to generic. For now, I only included the ones that I could verify are working as intended.

This PR edits:

- `:h copy()`
- `:h deepcopy()`
- `:h reduce()`
- `:h reverse()`
- `:h sort()`

<details>
<summary>Testing `copy()` / `deepcopy()`</summary>

```lua
---@class FooDeep
---@field bar string
---@field another integer

---@type FooDeep
local foo = {bar="asdf", another=123}
local copy = vim.fn.deepcopy(foo)
print(copy.another) -- NOTE: Is found, as expected
print(copy.does_not_exist) -- NOTE: Is missing, as expected

---@type FooDeep
local foo2 = {bar="asdf", another=123}
local copy2 = vim.fn.copy(foo2)
print(copy2.another) -- NOTE: Is found, as expected
print(copy2.does_not_exist) -- NOTE: Is missing, as expected
```
</details>

<details>
<summary>Testing `reduce()`</summary>

```lua
---@param accumulator string
---@param value string
local function reducer(accumulator, value)
  return (accumulator or "") .. value
end

local result = vim.fn.reduce({"a", "b", "c"}, reducer)
print(result:match("asdf")) -- NOTE: Works as expected
print(result:does_not_exist()) -- NOTE: Errors as expected
```
</details>

<details>
<summary>Testing `reverse()`</summary>

```lua
local data = {"a", "z", "b"}

local reversed = vim.fn.reverse(data)
print(reversed[1]:match("asdf")) -- NOTE: is found, as expected
reversed[1]:does_not_exist() -- NOTE: is missing, as expected
```
</details>

<details>
<summary>Testing `sort()`</summary>

```lua
---@class Foo2
---@field bar string
---@field another integer

---@type Foo2[]
local foo = {{bar="asdf", another=123}}
local copy = vim.fn.sort(foo)
print(copy[1].another)
print(copy[1].does_not_exist)
```
</details>

Unrelated to `@generic` but because I was in the neighborhood -

- `max` - currently returning any
- `min` - currently returning any
- `items` - currently accepts any argument (but really it needs a `table`)

As far as I know `min` `max` functions return a number, always. Currently they're typed to return any. If that's wrong and they actually can return non-numbers then I'll revert it.

## Question For Reviewers
`runtime/doc/builtin.txt` uses the raw parameter type / return type. Previously `any` and is now `T`. `T` is pretty commonly understood as just a name for "Type" but I wondered if you wanted to do some kind of replacement. Something like "For builtin.txt, when a generic is used in a function and that name is found in the param / return, replace it back from `T` to the type name" or something.

### DIFF
--- a/src/nvim/eval.lua
+++ b/src/nvim/eval.lua
@@ -17,6 +17,7 @@
 --- @field deprecated? true
 --- @field returns? string|false
 --- @field returns_desc? string
+--- @field generics? string[] Used to write `---@generic` annotations over a function.
 --- @field signature? string
 --- @field desc? string
 --- @field params [string, string, string][]
@@ -1521,9 +1522,10 @@ M.funcs = {
       A |Dictionary| is copied in a similar way as a |List|.
       Also see |deepcopy()|.
     ]=],
+    generics = { 'T' },
     name = 'copy',
-    params = { { 'expr', 'any' } },
-    returns = 'any',
+    params = { { 'expr', 'T' } },
+    returns = 'T',
     signature = 'copy({expr})',
   },
   cos = {
@@ -1739,8 +1741,10 @@ M.funcs = {
       Also see |copy()|.
 
     ]=],
+    generics = { 'T' },
     name = 'deepcopy',
-    params = { { 'expr', 'any' }, { 'noref', 'boolean' } },
+    params = { { 'expr', 'T' }, { 'noref', 'boolean' } },
+    returns = 'T',
     signature = 'deepcopy({expr} [, {noref}])',
   },
   delete = {
@@ -5894,7 +5898,7 @@ M.funcs = {
       the index.
     ]=],
     name = 'items',
-    params = { { 'dict', 'any' } },
+    params = { { 'dict', 'table' } },
     signature = 'items({dict})',
   },
   jobclose = {
@@ -7309,6 +7313,7 @@ M.funcs = {
     ]=],
     name = 'max',
     params = { { 'expr', 'any' } },
+    returns = 'number',
     signature = 'max({expr})',
   },
   menu_get = {
@@ -8520,7 +8525,13 @@ M.funcs = {
       <
     ]=],
     name = 'reduce',
-    params = { { 'object', 'any' }, { 'func', 'function' }, { 'initial', 'any' } },
+    generics = { 'T' },
+    params = {
+      { 'object', 'any' },
+      { 'func', 'fun(accumulator: T, current: any): any' },
+      { 'initial', 'any' },
+    },
+    returns = 'T',
     signature = 'reduce({object}, {func} [, {initial}])',
   },
   reg_executing = {
@@ -8785,7 +8796,9 @@ M.funcs = {
       <
     ]=],
     name = 'reverse',
-    params = { { 'object', 'any' } },
+    generics = { 'T' },
+    params = { { 'object', 'T[]' } },
+    returns = 'T[]',
     signature = 'reverse({object})',
   },
   round = {
@@ -10924,7 +10937,9 @@ M.funcs = {
       <
     ]=],
     name = 'sort',
-    params = { { 'list', 'any' }, { 'how', 'string|function' }, { 'dict', 'any' } },
+    generics = { 'T' },
+    params = { { 'list', 'T[]' }, { 'how', 'string|function' }, { 'dict', 'any' } },
+    returns = 'T[]',
     signature = 'sort({list} [, {how} [, {dict}]])',
   },
   soundfold = {
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1305,10 +1305,10 @@ copy({expr})                                                            *copy()*
 		Also see |deepcopy()|.
 
                 Parameters: ~
-                  • {expr} (`any`)
+                  • {expr} (`T`)
 
                 Return: ~
-                  (`any`)
+                  (`T`)
 
 cos({expr})                                                              *cos()*
 		Return the cosine of {expr}, measured in radians, as a |Float|.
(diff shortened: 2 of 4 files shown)


src/nvim/eval.lua
```diff
@@ -17,9 +17,10 @@
 --- @field deprecated? true
 --- @field returns? string|false
 --- @field returns_desc? string
+--- @field generics string[] Used to write `---@generics` annotations over a function.
 --- @field signature? string
 --- @field desc? string
---- @field params [string, string, string][]
+--- @field params {[1]: string, [2]: string}[] The input of any `signature`.
```
lewis6991, line 23: This looks wrong. The old format was a 3 string tuple of `name, type, desc`.

lewis6991 (approved):

justinmk:
> `runtime/doc/builtin.txt` uses the raw parameter type / return type. Previously `any` and is now `T`.

This is fine.

src/nvim/eval.lua
```diff
@@ -17,9 +17,10 @@
 --- @field deprecated? true
 --- @field returns? string|false
 --- @field returns_desc? string
+--- @field generics string[] Used to write `---@generics` annotations over a function.
```
justinmk, line 20: This is perhaps another hint that suggests we need to migrate to using the meta files as the "source of truth" for the type info instead of re-inventing a type language in `eval.lua`.

justinmk (approved):

src/nvim/eval.lua
lewis6991: Keeping the source of truth as a pure structured format makes the generators much simpler overall and better accommodates all the other metadata eval functions have.

I wouldn't say we are re-inventing the type language, it is just being presented in a structured/unwound format.

justinmk: > Keeping the source of truth as a pure structured format makes the generators much simpler overall

? We have the luacats parser, and the meta files are structured.

lewis6991: The parser is far from perfect. It's just good enough to build documentation from it. And not requiring a parser is much simpler. It means people can add additional metadata without having to also update a parser.

src/nvim/eval.lua
ColinKennedy: I forgot to leave a note about this, thank you for reminding me - the original `[string, string, string][]` is incorrect.

This can be verified with `lua-language-server`

`lua-language-sever --check repro.lua`
`repro.lua`
```lua
---@class Foo
---@field [string, string string][]
```

`Diagnosis complete, 1 problems found, see ~/path/to/lua-language-server/libexec/log/check.json`

```json
{
    "file:///home/selecaoone/temp/annotation_key_test.lua": [
        {
            "code": "luadoc-miss-symbol",
            "data": "syntax",
            "message": "`]` expected.",
            "range": {
                "end": {
                    "character": 17,
                    "line": 1
                },
                "start": {
                    "character": 17,
                    "line": 1
                }
            },
            "severity": 2,
            "source": "Lua Syntax Check."
        }
    ]
}
```

`[string, string, string?][]` has the same problem.

Also if you run `:InspectTree`, the parsed tree contains ERROR on that line.

That said though my suggestion doesn't appear to work 100% either. I think it's a bug with LuaLS

```lua
---@class Foo
---@field [1] string

---@type Foo
local foo = {[1]="asdf"}

foo[2]:match("asdfasfd") -- This should error, but doesn't


---@class Bar
---@field thing string

---@type Bar
local foo = {thing="asdf"}

foo[2]:match("asdfasfd") -- This should error, but doesn't
foo.does_not_exist:match("asdfasfd") -- This should error, and it does
foo["does_not_exist"]:match("asdfasfd") -- This should error, but doesn't
```

It seems that accessing undefined numbered fields does not error. So either my suggestion is wrong and there's a better way to define it or it's a bug with LuaLS. Not sure. But the previous syntax does cause an error

lewis6991: > Also if you run :InspectTree, the parsed tree contains ERROR on that line.

The output of InspectTree doesn't matter for Lua annotations. What matters is how LuaLS processes them.

ColinKennedy: image

It errors for me both in-editor and using the `--check` terminal flag. With either `---@field params [string, string, string][]`  or `---@field params [string, string, string?][]`. What do you see when you run lua-language-server?

lewis6991: Tuples were added in 3.8.0 https://github.com/LuaLS/lua-language-server/pull/2505.

Newest version is 3.13

ColinKennedy: You're right, I'll revert this change.

On an unrelated note, it looks like 3.8.0+ has broken `lua-language-server --check`. But since that issue is unrelated to this PR, I'll make a separate issue over in their repo

Thank you for explaining it to me!

lewis6991: --check works for me on 3.13. There might be some changes in how it works though.

ColinKennedy: Interesting, I can't get it working on two machines so I'm not sure. Hopefully it's something simple.

About this PR, I reverted the change fully. If you want me to add `[string, string, string?][]` instead or something I can add that in.

src/nvim/eval.lua
ColinKennedy: Unrelated but realized to your chat but realized I forgot to add `?` so I changed it to `--- @field generics? string[] ...`

src/nvim/eval.lua
lewis6991: That or what it was before.

justinmk:
ready to merge?

ColinKennedy:
I'd say so yes but it's up to you all

lewis6991:
LGTM but I would not recommend anyone use any of the affected functions in Lua. There are much better alternatives: `vim.deepcopy`, `table.sort` etc.

A follow-up would be to document this.

PR #31559, opened 2024-12-12
feat(api): add/del folds, get fold information

ribru17:
Work on #19226


ofseed:
I have another perspective: implementing nvim's own fold API on the current foundation might be difficult. I glanced at the planned API, and there is a problem: if we allow a window to have foldings from different sources (for example, multiple LSP servers providing through `vim.lsp.foldexpr`), then if different sources create folds at the same start and end, then the target of these new APIs may be ambiguous. Therefore, I think if such an API exists, it should at least support namespaces. If it is implemented based on extmark, then this problem would not exist.

PR #31562, opened 2024-12-12
feat(terminal)!: cursor shape and blink

### LINKED ISSUE #3681: :terminal cursor doesn't match $TERM cursor
It looks like NeoVim is manually drawing the cursor in `:terminal` windows, rather than letting the external terminal (in my case, urxvt) draw the cursor. This results in some irksome inconsistency in how my cursor looks in `:terminal` windows vs. how it looks everywhere else, including other buffers in NeoVim. Specifically, I have my terminal cursor set to blink. Everywhere else in NeoVim, my cursor blinks, but in `:terminal` windows, it doesn't.

NeoVim really ought to be letting the external terminal draw the cursor.

### LINKED ISSUE #20325: Event `mode_change` does not report terminal mode
### Neovim version (nvim -v)

NVIM v0.7.0-dev+1299-g773516365-dirty

N/A

### Operating system/version

Gentoo

### Terminal name/version

Konsole 21.08.3

xterm-256color

### Installation

Portage (9999 Nightly)

[Running modified nvim-qt frontend to demonstrate issue]

The following patch has been applied to instrument `mode_change` events:
```patch
diff --git a/src/gui/shell.cpp b/src/gui/shell.cpp
index c91097b..1a469d1 100644
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -566,6 +566,7 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
        } else if (name == "mouse_off"){
                handleMouse(false);
        } else if (name == "mode_change"){
+               qDebug() << opargs;
                handleModeChange(opargs);
        } else if (name == "set_title"){
                handleSetTitle(opargs);

```

1. Open GUI frontend: `nvim-qt -- -u NORC`
2. Enter terminal mode with `:terminal`
3. Notice no mode_change is reported for terminal mode.

Here is a trace of all mode_change events:
```
(QVariant(QByteArray, "normal"), QVariant(qulonglong, 0))
(QVariant(QByteArray, "normal"), QVariant(qulonglong, 0))
(QVariant(QByteArray, "normal"), QVariant(qulonglong, 0)) // Occurs during startup
(QVariant(QByteArray, "cmdline_normal"), QVariant(qulonglong, 4)) // : - Command mode entered
(QVariant(QByteArray, "normal"), QVariant(qulonglong, 0)) // :terminal command completes
```

### Expected behavior

There should be a `mode_change` when the user enters and exits terminal mode.

### Actual behavior

There is no `mode_change` event, and the frontend cannot detect when the user is in terminal mode.

Related issues:
https://github.com/equalsraf/neovim-qt/issues/999
https://github.com/neovim/neovim/issues/19265

gpanders:
When a terminal application running inside the terminal emulator sets the cursor shape or blink status of the cursor, update the cursor in the parent terminal to match.

This removes the "virtual cursor" that has been in use by the terminal emulator since the beginning. The original rationale for using the virtual cursor was to avoid having to support additional UI methods to change the cursor color for other (non-TUI) UIs, instead relying on the TermCursor and TermCursorNC highlight groups.

The TermCursor highlight group is now used in the default `'guicursor'` value, which has a new entry for Terminal mode. However, the TermCursorNC highlight group is no longer supported: since terminal windows now use the real cursor, when the window is not focused there is no cursor displayed in the window at all, so there is nothing to highlight. Users can still use the `StatusLineTermNC` highlight group to differentiate non-focused terminal windows.

BREAKING CHANGE: The TermCursorNC highlight group is no longer supported.

Fixes #3681.
Fixes https://github.com/neovim/neovim/issues/20325.

### DIFF
--- a/runtime/doc/deprecated.txt
+++ b/runtime/doc/deprecated.txt
@@ -38,6 +38,10 @@ DIAGNOSTICS
 - The "cursor_position" parameter of |vim.diagnostic.JumpOpts| is renamed to
   "pos"
 
+HIGHLIGHTS
+• *TermCursorNC*	As of Nvim 0.11, unfocused |terminal| windows no
+			longer have any cursor.
+
 TREESITTER
 • *TSNode:child_containing_descendant()*	Use
   |TSNode:child_with_descendant()| instead; it is identical except that it can
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -87,6 +87,11 @@ EVENTS
 • |vim.ui_attach()| callbacks for |ui-messages| `msg_show` events are executed in
   |api-fast| context.
 
+HIGHLIGHTS
+
+• |TermCursorNC| is removed and no longer supported. Unfocused terminals no
+  longer have a cursor.
+
 LSP
 
 • Improved rendering of LSP hover docs. |K-lsp-default|
@@ -281,6 +286,12 @@ TERMINAL
   'scrollback' are not reflown.
 • The |terminal| now supports OSC 8 escape sequences and will display
   hyperlinks in supporting host terminals.
+• The |terminal| now uses the actual cursor, rather than a "virtual" cursor.
+  This means that escape codes sent by applications running in a terminal
+  buffer can change the cursor shape and visibility. However, it also
+  means that the |TermCursorNC| highlight group is no longer supported: an
+  unfocused terminal window will have no cursor at all (so there is nothing to
+  highlight).
 
 TREESITTER
 
--- a/src/nvim/cursor_shape.c
+++ b/src/nvim/cursor_shape.c
@@ -45,6 +45,7 @@ cursorentry_T shape_table[SHAPE_IDX_COUNT] = {
   { "more", 0, 0, 0,   0,   0,   0, 0, 0, "m", SHAPE_MOUSE },
   { "more_lastline", 0, 0, 0,   0,   0,   0, 0, 0, "ml", SHAPE_MOUSE },
   { "showmatch", 0, 0, 0, 100, 100, 100, 0, 0, "sm", SHAPE_CURSOR },
+  { "terminal", 0, 0, 0, 0, 0, 0, 0, 0, "t", SHAPE_CURSOR },
 };
 
 /// Converts cursor_shapes into an Array of Dictionaries
@@ -321,6 +322,8 @@ int cursor_get_mode_idx(void)
 {
   if (State == MODE_SHOWMATCH) {
     return SHAPE_IDX_SM;
+  } else if (State == MODE_TERMINAL) {
+    return SHAPE_IDX_TERM;
   } else if (State & VREPLACE_FLAG) {
     return SHAPE_IDX_R;
   } else if (State & REPLACE_FLAG) {
--- a/src/nvim/terminal.c
+++ b/src/nvim/terminal.c
@@ -52,6 +52,7 @@
 #include "nvim/channel.h"
 #include "nvim/channel_defs.h"
 #include "nvim/cursor.h"
+#include "nvim/cursor_shape.h"
 #include "nvim/drawline.h"
 #include "nvim/drawscreen.h"
 #include "nvim/eval.h"
@@ -160,14 +161,18 @@ struct terminal {
   int invalid_start, invalid_end;   // invalid rows in libvterm screen
   struct {
     int row, col;
+    int shape;
     bool visible;
+    bool blink;
   } cursor;
-  bool pending_resize;              // pending width/height
+
+  struct {
+    bool resize;          ///< pending width/height
+    bool cursor;          ///< pending cursor shape or blink change
+    StringBuilder *send;  ///< When there is a pending TermRequest autocommand, block and store input.
+  } pending;
 
   bool color_set[16];
-
-  // When there is a pending TermRequest autocommand, block and store input.
-  StringBuilder *pending_send;
 
   char *selection_buffer;  /// libvterm selection buffer
   StringBuilder selection;  /// Growable array containing full selection data
@@ -207,24 +212,24 @@ static void emit_termrequest(void **argv)
   apply_autocmds_group(EVENT_TERMREQUEST, NULL, NULL, false, AUGROUP_ALL, buf, NULL, &data);
   xfree(payload);
 
-  StringBuilder *term_pending_send = term->pending_send;
-  term->pending_send = NULL;
+  StringBuilder *term_pending_send = term->pending.send;
+  term->pending.send = NULL;
   if (kv_size(*pending_send)) {
     terminal_send(term, pending_send->items, pending_send->size);
     kv_destroy(*pending_send);
   }
   if (term_pending_send != pending_send) {
-    term->pending_send = term_pending_send;
+    term->pending.send = term_pending_send;
   }
   xfree(pending_send);
 }
 
 static void schedule_termrequest(Terminal *term, char *payload, size_t payload_length)
 {
-  term->pending_send = xmalloc(sizeof(StringBuilder));
-  kv_init(*term->pending_send);
+  term->pending.send = xmalloc(sizeof(StringBuilder));
+  kv_init(*term->pending.send);
   multiqueue_put(main_loop.events, emit_termrequest, term, payload, (void *)payload_length,
-                 term->pending_send);
+                 term->pending.send);
 }
 
 static int parse_osc8(VTermStringFragment frag, int *attr)
@@ -363,7 +368,7 @@ void terminal_open(Terminal **termpp, buf_T *buf, TerminalOptions opts)
   // Create a new terminal instance and configure it
   Terminal *term = *termpp = xcalloc(1, sizeof(Terminal));
   term->opts = opts;
-  term->cursor.visible = true;
+
   // Associate the terminal instance with the new buffer
   term->buf_handle = buf->handle;
   buf->terminal = term;
@@ -386,6 +391,28 @@ void terminal_open(Terminal **termpp, buf_T *buf, TerminalOptions opts)
   term->selection_buffer = xcalloc(SELECTIONBUF_SIZE, 1);
   vterm_state_set_selection_callbacks(state, &vterm_selection_callbacks, term,
                                       term->selection_buffer, SELECTIONBUF_SIZE);
+
+  VTermValue cursor_shape;
+  switch (shape_table[SHAPE_IDX_TERM].shape) {
+  case SHAPE_BLOCK:
+    cursor_shape.number = VTERM_PROP_CURSORSHAPE_BLOCK;
+    break;
+  case SHAPE_HOR:
+    cursor_shape.number = VTERM_PROP_CURSORSHAPE_UNDERLINE;
+    break;
+  case SHAPE_VER:
+    cursor_shape.number = VTERM_PROP_CURSORSHAPE_BAR_LEFT;
+    break;
+  }
+  vterm_state_set_termprop(state, VTERM_PROP_CURSORSHAPE, &cursor_shape);
+
+  VTermValue cursor_blink;
+  if (shape_table[SHAPE_IDX_TERM].blinkon != 0 && shape_table[SHAPE_IDX_TERM].blinkoff != 0) {
+    cursor_blink.boolean = true;
+  } else {
+    cursor_blink.boolean = false;
+  }
+  vterm_state_set_termprop(state, VTERM_PROP_CURSORBLINK, &cursor_blink);
 
   // force a initial refresh of the screen to ensure the buffer will always
   // have as many lines as screen rows when refresh_scrollback is called
@@ -565,7 +592,7 @@ void terminal_check_size(Terminal *term)
 
   vterm_set_size(term->vt, height, width);
   vterm_screen_flush_damage(term->vts);
-  term->pending_resize = true;
+  term->pending.resize = true;
   invalidate_terminal(term, -1, -1);
 }
 
@@ -614,15 +641,27 @@ bool terminal_enter(void)
   curwin->w_p_so = 0;
   curwin->w_p_siso = 0;
 
+  // Save the existing cursor entry since it may be modified by the application
+  cursorentry_T save_cursorentry = shape_table[SHAPE_IDX_TERM];
+
+  // Update the cursor shape table and flush changes to the UI
+  s->term->pending.cursor = true;
+  refresh_cursor(s->term);
+
   adjust_topline(s->term, buf, 0);  // scroll to end
-  // erase the unfocused cursor
-  invalidate_terminal(s->term, s->term->cursor.row, s->term->cursor.row + 1);
   showmode();
   curwin->w_redr_status = true;  // For mode() in statusline. #8323
   redraw_custom_title_later();
-  ui_busy_start();
+  if (!s->term->cursor.visible) {
+    // Hide cursor if it should be hidden
+    ui_busy_start();
+  }
+  ui_cursor_shape();
   apply_autocmds(EVENT_TERMENTER, NULL, NULL, false, curbuf);
   may_trigger_modechanged();
+
+  // Tell the terminal it has focus
+  terminal_focus(s->term, true);
 
   s->state.execute = terminal_execute;
   s->state.check = terminal_check;
@@ -634,6 +673,9 @@ bool terminal_enter(void)
   State = save_state;
   RedrawingDisabled = s->save_rd;
   apply_autocmds(EVENT_TERMLEAVE, NULL, NULL, false, curbuf);
+
+  shape_table[SHAPE_IDX_TERM] = save_cursorentry;
+  ui_mode_info_set();
 
   if (save_curwin == curwin->handle) {  // Else: window was closed.
     curwin->w_p_cul = save_w_p_cul;
@@ -649,8 +691,9 @@ bool terminal_enter(void)
     free_string_option(save_w_p_culopt);
   }
 
-  // draw the unfocused cursor
-  invalidate_terminal(s->term, s->term->cursor.row, s->term->cursor.row + 1);
+  // Tell the terminal it lost focus
+  terminal_focus(s->term, false);
+
   if (curbuf->terminal == s->term && !s->close) {
     terminal_check_cursor();
   }
@@ -659,7 +702,11 @@ bool terminal_enter(void)
   } else {
     unshowmode(true);
   }
-  ui_busy_stop();
+  if (!s->term->cursor.visible) {
+    // If cursor was hidden, show it again
+    ui_busy_stop();
+  }
+  ui_cursor_shape();
   if (s->close) {
     bool wipe = s->term->buf_handle != 0;
     s->term->destroy = true;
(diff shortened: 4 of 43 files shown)


runtime/doc/news.txt
```diff
@@ -281,6 +281,8 @@ TERMINAL
   'scrollback' are not reflown.
 • The |terminal| now supports OSC 8 escape sequences and will display
   hyperlinks in supporting host terminals.
+• The |terminal| now updates the cursor shape and blink when set by an
+  application.
```
justinmk, line 285: should we also mention something like "defers to the actual UI cursor instead of drawing a fake cursor"

src/nvim/cursor_shape.c
```diff
@@ -45,6 +45,7 @@ cursorentry_T shape_table[SHAPE_IDX_COUNT] = {
   { "more", 0, 0, 0,   0,   0,   0, 0, 0, "m", SHAPE_MOUSE },
   { "more_lastline", 0, 0, 0,   0,   0,   0, 0, 0, "ml", SHAPE_MOUSE },
   { "showmatch", 0, 0, 0, 100, 100, 100, 0, 0, "sm", SHAPE_CURSOR },
+  { "terminal", 0, 0, 0, 0, 0, 0, 0, 0, "te", SHAPE_CURSOR },
```
justinmk, line 48: Just for reference: interesting that vim doesn't have this https://github.com/vim/vim/blob/c942f84aadffd0c8969ecf81e3e9103722b2714f/src/misc2.c#L2042

justinmk, line 48: the `mode()` representation of terminal-mode is "t", does it make sense to match that here? 

```suggestion
  { "terminal", 0, 0, 0, 0, 0, 0, 0, 0, "t", SHAPE_CURSOR },
```

src/nvim/terminal.c
```diff
@@ -615,15 +620,16 @@ bool terminal_enter(void)
   curwin->w_p_siso = 0;
 
   adjust_topline(s->term, buf, 0);  // scroll to end
-  // erase the unfocused cursor
-  invalidate_terminal(s->term, s->term->cursor.row, s->term->cursor.row + 1);
   showmode();
   curwin->w_redr_status = true;  // For mode() in statusline. #8323
   redraw_custom_title_later();
-  ui_busy_start();
+  ui_cursor_shape();
```
justinmk, line 626: for GUIs should news.txt mention something like "terminal-mode no longer signals 'busy' event" ? cc @fredizzimo

justinmk (approved):
Very excited for this!

src/nvim/terminal.c
gpanders: AFAICT this was only being used to hide the cursor in the TUI. I'm not sure how GUIs might have been using it though.

src/nvim/cursor_shape.c
gpanders: That's not surprising, we are adding this here only so that we can communicate cursor shape information to the UI in the `mode_info_set` event. Since Vim doesn't have the UI architecture it makes sense they would have another way of implementing this.

cosminadrianpopescu:
I am trying to test it and the cursor seems to be always one character to the left while editing. Having, for example:

$> ls |-

I would expect in this situation, when i click "a", the a to appear before the "-" character, but it appears after. See the attached movie.
[capture.webm](https://github.com/user-attachments/assets/416a876f-a034-4963-b4cb-807fcf4c4915)

gpanders:
> I am trying to test it and the cursor seems to be always one character to the left while editing. Having, for example:
> 
> $> ls |-
> 
> I would expect in this situation, when i click "a", the a to appear before the "-" character, but it appears after. See the attached movie. [capture.webm](https://github.com/user-attachments/assets/416a876f-a034-4963-b4cb-807fcf4c4915)

This is mentioned in the PR description:

>There is one bug I'm chasing down where exiting and then re-entering Terminal mode causes a disagreement in cursor position between Nvim and the terminal. But otherwise it seems to be working ok.

gpanders:
I was able to solve the cursor issue so this is fully working now. I also implemented cursor hiding, so if an application in the terminal emulator hides the cursor it will be hidden in Nvim too.

Because we are using the real cursor now and not the virtual cursor, I had to update any test that uses the terminal cursor in the expected screen output... which was a lot of them. There are some tests that fail now because of TermCursor. I've marked those as `pending` for now.

runtime/doc/news.txt
```diff
@@ -281,6 +281,8 @@ TERMINAL
   'scrollback' are not reflown.
 • The |terminal| now supports OSC 8 escape sequences and will display
   hyperlinks in supporting host terminals.
+• The |terminal| now updates the cursor shape and blink when set by an
```
justinmk, line 284: we should also mention `TermCursor` and `TermCursorNC` highlights currently don't work. maybe in the `:help news-breaking-dev` if we expect to fix this before 0.11 release.

test/functional/ui/cursor_spec.lua
```diff
@@ -190,6 +190,19 @@ describe('ui/cursor', function()
...
+        blinkon = 0,
+        blinkwait = 0,
+        cell_percentage = 0,
+        cursor_shape = 'block',
+        name = 'terminal',
+        hl_id = 0,
+        id_lm = 0,
+        attr = {},
+        attr_lm = {},
+        short_name = 't',
```
justinmk, line 204: does `:help 'guicursor'` need an update?

test/functional/terminal/highlight_spec.lua
```diff
@@ -248,8 +248,21 @@ it('CursorLine and CursorColumn work in :terminal buffer in Normal mode', functi
     {1:bar fooba^r                                        }|
                                                       |
   ]])
-  -- CursorLine and CursorColumn are combined with TermCursorNC.
+end)
+
+-- Pending until TermCursor highlights are working again #31562
+pending('CursorLine and CursorColumn are combined with TermCursorNC', function()
```
zeertzjq, line 254: This is a test for the drawline.c terminal highlight combining code. If the terminal cursor is no longer highlighted by drawline.c, some other highlights should be added to the terminal.

runtime/doc/news.txt
gpanders: I fixed `TermCursor`, so we won't regress that in this PR (it was easy, just set it in `'guicursor'` and everything magically worked).

`TermCursorNC` will no longer be supported at all though. This is not a temporary regression, but a permanent change. When the terminal window doesn't have focus, there is simply no cursor, so there's no highlight group to apply. We still have `StatusLineTermNC` though.

I'll add this to `news.txt` and `deprecated.txt`.

test/functional/ui/cursor_spec.lua
gpanders: Yes, done.

test/functional/terminal/highlight_spec.lua
```diff
@@ -248,8 +248,9 @@ it('CursorLine and CursorColumn work in :terminal buffer in Normal mode', functi
     {1:bar fooba^r                                        }|
                                                       |
   ]])
-  -- CursorLine and CursorColumn are combined with TermCursorNC.
-  command('highlight TermCursorNC gui=reverse')
+  -- CursorLine and CursorColumn are combined with other highlight groups
+  command('highlight Search NONE gui=reverse')
+  command('match Search /\\%6l\\%11c/')
```
zeertzjq, line 253: Match highlight does not come from the terminal and does not cover the drawline.c terminal highlight code. The highlights need to come from the program running in the terminal.

test/functional/terminal/highlight_spec.lua
```diff
@@ -269,8 +270,8 @@ it('CursorLine and CursorColumn work in :terminal buffer in Normal mode', functi
     bar foobar{4: }                                       |
                                                       |
   ]])
-  -- TermCursorNC has higher precedence.
-  command('highlight TermCursorNC gui=NONE guibg=Red')
+  -- Search has higher precedence.
+  command('highlight Search gui=NONE guibg=Red')
```
gpanders, line 274: @zeertzjq I updated this test to use `Search` instead of `TermCursorNC`. All of the tests pass except for the one on line 264: the highlight group on line 270 is `2` instead of `4`. I am not sure why this is the case because the test on line 254 passes correctly.

Will look at it more later, but if you have any ideas why this test is failing let me know.

test/functional/terminal/cursor_spec.lua
```diff
@@ -93,39 +93,44 @@ describe(':terminal cursor', function()
...
+      ]],
+        unchanged = true,
+      })
     end)
   end)
 end)
 
-describe('cursor with customized highlighting', function()
+-- Pending until screen.lua supports coloring the cursor
+pending('cursor with customized highlighting', function()
```
gpanders, line 125: `TermCursor` highlights work but AFAICT screen.lua doesn't handle coloring the cursor based on hl groups from `'guicursor'`. I tried implementing this but ran out of time tonight. Will try again later.

test/functional/terminal/highlight_spec.lua
gpanders: See https://github.com/neovim/neovim/pull/31562#discussion_r1884787479

zeertzjq: ?

gpanders: TermCursorNC doesn’t come from the program running in the terminal either? What is this supposed to be testing? Your PR adding this test had zero context or other information.

By the way, editing your comment after the fact and then asking if I didn’t understand it is poor form.

zeertzjq: > By the way, editing your comment after the fact and then asking if I didn’t understand it is poor form.

After what fact? I only made a one-character edit to my comment before I saw your reply, and it didn't change the overall meaning.

> By “comes from the terminal” do you mean a highlight that has the `kHlTerminal` kind?

No, I mean it needs to come from `terminal_get_line_attributes()`.

gpanders: >No, I mean it needs to come from terminal_get_line_attributes(). But since TermCursorNC is removed, it can now only come from the program running in the terminal.

Ok the updated test ought to be sufficient now I think. Instead of the old cursor I printed text to the terminal using different SGRs.

runtime/doc/deprecated.txt
```diff
@@ -66,6 +66,9 @@ LSP
 • `client.on_attach()`			Use |Client:on_attach()| instead.
 • `vim.lsp.start_client()`		Use |vim.lsp.start()| instead.
 
+HIGHLIGHTS
```
justinmk, line 69: nit: sections are ordered alphabetically

runtime/doc/news.txt
```diff
@@ -219,6 +219,11 @@ EVENTS
   for completion being done.
 • |vim.on_key()| callbacks can consume the key by returning an empty string.
 
+HIGHLIGHTS
+
+• |TermCursorNC| is removed and no longer supported. Unfocused terminals no
+  longer have a cursor.
```
justinmk, line 225: is this intended to be in `news-breaking` ?

gpanders: Yes, thanks

test/functional/terminal/highlight_spec.lua
```diff
@@ -195,7 +195,7 @@ it('CursorLine and CursorColumn work in :terminal buffer in Normal mode', functi
   local screen = Screen.new(50, 7)
   screen:set_default_attr_ids({
     [1] = { background = Screen.colors.Grey90 }, -- CursorLine, CursorColumn
-    [2] = { reverse = true }, -- TermCursor
+    [2] = { reverse = true }, -- Search
```
zeertzjq, line 198:
```suggestion
    [2] = { reverse = true },
```

test/functional/terminal/highlight_spec.lua
```diff
@@ -248,46 +248,54 @@ it('CursorLine and CursorColumn work in :terminal buffer in Normal mode', functi
     {1:bar fooba^r                                        }|
                                                       |
   ]])
-  -- CursorLine and CursorColumn are combined with TermCursorNC.
-  command('highlight TermCursorNC gui=reverse')
+  -- CursorLine and CursorColumn are combined with other highlight groups
```
zeertzjq, line 251:
```suggestion
  -- CursorLine and CursorColumn are combined with terminal highlights.
```

test/functional/terminal/cursor_spec.lua
gpanders: I'm wondering if we should just remove this test. The terminal cursor color is set in `'guicursor'` now, so is colored by the UI. So this is a UI test now, not a Nvim core test, and we already have a test for `'guicolor'` in `ui/cursor_spec.lua`.

gpanders:
@cosminadrianpopescu Should be fixed now, can you please try again?

cosminadrianpopescu:
@gpanders it seems to be working now. I will test more later. But the initial tests look good. The issue dissapeared. 

Thank you for the fix and for this PR.

gpanders:
Windows only test failure:

```
FAILED   1 test, listed below:
FAILED   test/functional\terminal\highlight_spec.lua @ 193: CursorLine and CursorColumn work in :terminal buffer in Normal mode
test/functional\terminal\highlight_spec.lua:255: Row 6 did not match.
Expected:
  |tty ready{1: }                                        |
  | foobar f{1:o}obar foobar foobar foobar foobar foobar |
  |foobar fo{1:o}bar foobar foobar foobar foobar foobar f|
  |oobar foo{1:b}ar foobar foobar foobar foobar foobar fo|
  |obar foob{1:a}r foobar foobar foobar foobar foobar foo|
  |*{1:bar fooba^r}{4: foobar}{1:                                 }|
  |                                                  |
Actual:
  |tty ready{1: }                                        |
  | foobar f{1:o}obar foobar foobar foobar foobar foobar |
  |foobar fo{1:o}bar foobar foobar foobar foobar foobar f|
  |oobar foo{1:b}ar foobar foobar foobar foobar foobar fo|
  |obar foob{1:a}r foobar foobar foobar foobar foobar foo|
  |*{1:bar fooba^r foobar                                 }|
  |                                                  |
```

But I cannot repro locally in a Windows VM:

image

gpanders:
Windows only test failure:

```
FAILED   test/functional\terminal\highlight_spec.lua @ 193: CursorLine and CursorColumn work in :terminal buffer in Normal mode
test/functional\terminal\highlight_spec.lua:255: Row 6 did not match.
Expected:
  |tty ready{1: }                                        |
  | foobar f{1:o}obar foobar foobar foobar foobar foobar |
  |foobar fo{1:o}bar foobar foobar foobar foobar foobar f|
  |oobar foo{1:b}ar foobar foobar foobar foobar foobar fo|
  |obar foob{1:a}r foobar foobar foobar foobar foobar foo|
  |*{1:bar fooba^r}{4: foobar}{1:                                 }|
  |                                                  |
Actual:
  |tty ready{1: }                                        |
  | foobar f{1:o}obar foobar foobar foobar foobar foobar |
  |foobar fo{1:o}bar foobar foobar foobar foobar foobar f|
  |oobar foo{1:b}ar foobar foobar foobar foobar foobar fo|
  |obar foob{1:a}r foobar foobar foobar foobar foobar foo|
  |*{1:bar fooba^r foobar                                 }|
  |                                                  |
```

I can reproduce this locally in a Windows VM as well. Curiously, I can't get the terminal to support any SGR sequences on Windows. Even when I run Nvim normally and try to emit an SGR (e.g. `vim.api.nvim_chan_send(vim.b.terminal_job_id, '\027[7m foobar\027[0m')`), it doesn't work.

Are SGRs known not to work on Windows for some reason? This seems unrelated to this PR so I might slap a conditional skip on this test for now.

gpanders:
>At this moment in the first terminal, we have a block cursor, although we should have the line cursor. The cursor will change back to line if I pass through normal mode, by doing a <C-\><C-n> and then i again. But if I don't pass through normal mode, the cursor will stay a block cursor.

Another good catch. Should be fixed now, can you please try again?

cosminadrianpopescu:
> Another good catch. Should be fixed now, can you please try again?

No, unfortunately it's still not working. Same behavior as previous. Did I missed something? I did a `git pull`, compile again, make install and restarted nvim,

gpanders:
Oh I see, your mappings change the window without exiting and Terminal mode. We need to also update the cursor if the window changes.

cosminadrianpopescu:
> EDIT: Should be fixed now in 9360236. Thanks for finding all these corner cases, let me know if you find any others!

Yes, it works now. Thank you very much. At the moment everything seems to be in order.

Tomorrow I will stress it some more, as I will use it a real environment, at work and I'll let you know if I find anything else. 

Again, great PR.

cosminadrianpopescu:
Related with the previous bug, I've noticed that if the actions are in the other order, then the cursor is lost until a new terminal is created. 

Given the mapping file (/tmp/mappings.lua):

```lua
vim.api.nvim_set_keymap('t', '<f1>', '', {
    callback = function()
        vim.api.nvim_command('tabn 1')
        vim.api.nvim_command('startinsert')
    end
})


vim.api.nvim_set_keymap('t', '<f2>', '', {
    callback = function()
        vim.api.nvim_command('tabn 2')
        vim.api.nvim_command('startinsert')
    end
})
```

then, in a new terminal: 

```
nvim --clean<cr>
:luafile /tmp/mappings.lua<cr>
:terminal<cr>i
<C-\><C-n>
:tabnew<cr>
:terminal<cr> # the line cursor is visible now
vifm<cr> # there is no cursor now
<f1> # still no cursor in the first tab
<C-\><C-n>i # still no cursor
```

It appears that there is no way to get back the cursor, after this. Even by doing `echo -e '\e[5 q'` or `reset` will not bring back the cursor. Maybe this helps.

gpanders:
I believe I've fixed the issues with cursor visibility (and I've added tests for those as well).

I can repro your issue with the nested `nvim` instance as well. I'm not sure exactly what's happening there so I'll have to dig into that a bit more.

cosminadrianpopescu:
Thank you for the latest fixes. I can confirm that indeed, the cursor visibility issue is fixed. Let me know when you want me to test nvim/vim nested issue.

gpanders:
@cosminadrianpopescu Please try again, I've pushed a fix that at least addresses the reproduction you gave me.

cosminadrianpopescu:
> @cosminadrianpopescu Please try again, I've pushed a fix that at least addresses the reproduction you gave me.

The nvim nested issue seems to be fixed. 

However, the previous issue (when the cursor is hidden by apps like `vifm`) seems I could reproduce it again. 

If I follow the steps I gave you, then everything is fine. But I manage to still encounter the issue using some scripts to send keys to the terminal. I need to isolate it and find a reproducible scenario to pass you with a `nvim --clean` and I'll be back with that issue.

gpanders:
@cosminadrianpopescu Since we’re now just looking at some rarer corner cases, I’m going to move forward merging this. Please open a new issue with your reproduction steps when you’re able.

cosminadrianpopescu:
> @cosminadrianpopescu Since we’re now just looking at some rarer corner cases, I’m going to move forward merging this. Please open a new issue with your reproduction steps when you’re able.

Sure. 

However, at the moment I'm not able to reproduce with a clean nvim. I tried all this morning and I didn't manage to do it. I still have to identify the issue on my config which is still generating the issue. But normal usage, without sending keys programmatically to the terminal and without other plugins (I also have 4 external plugins in my config) does not seem to produce the issue.

lifei:
Is this feature enable in v0.10.3?

lifei:
> > Is this feature enable in v0.10.3?
> 
> No, we do not usually backport breaking changes.

Thank you for quick replying. Which version will this feature enabled in?

zeertzjq:
0.11

Issue #15462, opened 2021-08-23
"if" in a commented line parsed as an actual "if statement" in regards to matchit working with nvim-treesitter


comiluv:
### Neovim version (nvim -v)

NVIM v0.6.0-dev+142-g5a111c1b0

### Vim (not Nvim) behaves the same?

no, vim doesn't support nvim-treesitter

### Operating system/version

Microsoft Windows [Version 10.0.19043.1165]

### Terminal name/version

Windows Terminal/1.10.1933.0

### $TERM environment variable

(nothing is in $TERM env)

### Installation

choco install neovim --pre

### How to reproduce the issue

1. Install [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter)
2. Make a viml file: test-vim.vim
3. Type following
```viml
if
	" A comment containing a word if
else
	echo ELSE
endif
```
4. Press % while the cursor is on endif

### Expected behavior

Commented line is ignored for matching if-else-endif with % key.
![GIF 2021-08-22 오후 7-43-05](https://user-images.githubusercontent.com/19642705/130394191-feeb0654-fb68-4ea9-b8a1-bc13ce5272d6.gif)

### Actual behavior

The cursor jumps to the word if in the commented line
![GIF 2021-08-22 오후 7-42-23](https://user-images.githubusercontent.com/19642705/130394073-6c0a3dd8-83c8-4e02-89d0-5ff38ded2131.gif)

comiluv:
This issue was posted on nvim-treesitter/nvim-treesitter#1732 but was discussed to be a problem with matchit

vigoux:
Is a matchit issue when `syntax off`, thus unrelated to tree-sitter i'd say ?

justinmk:
hopefully a (very small and simple) matchit replacement will materialize, which uses treesitter queries. Will keep this open to track that.

timtyrrell:
@justinmk I know this exists: https://github.com/andymass/vim-matchup#tree-sitter-integration

kaddkaka:
This behavior for `%` is surprising and manifests as a regression when a user activates treesitter.

Would it be possible to incorporate something like https://github.com/andymass/vim-matchup into neovim core to be more robust when someone activates treesitter?

clason:
That very much depends on the "something". A minimal(!) Lua alternative to matchit that is treesitter-aware is welcome.

justinmk:
[...]

PR #31563, opened 2024-12-12
fix(runtime): let match(it+paren) skips use treesitter

emiasims:
Problem: When treesitter is enabled, by default syntax groups are not defined, but these
groups are used to identify where to skip matches in matchit and matchparen.

Solution: This patch does three things:
1. If syntax is enabled regardless of treesitter (`vim.bo.syntax='on'`):
   Use original implementation.
2. If treesitter is enabled and syntax is not:
   Match the syntax groups (i.e. `comment\|string`) against treesitter captures
   to check for skipped groups.
3. Add an explicit treesitter syntax for marking captures to skip:
   matchit uses `b:match_skip` to determine what counts as skippable
   Where 's:comment\|string' uses a match of the named syntax groups against
   a regex match of comment\|string, 't:comment\|string' now uses vim regex
   to match against the names of the treesitter capture groups.

ref #15462

### DIFF
--- a/runtime/pack/dist/opt/matchit/autoload/matchit.vim
+++ b/runtime/pack/dist/opt/matchit/autoload/matchit.vim
@@ -222,6 +222,7 @@ function matchit#Match_wrapper(word, forward, mode) range
   let view = winsaveview()
   call cursor(0, curcol + 1)
   if skip =~ 'synID' && !(has("syntax") && exists("g:syntax_on"))
+        \ || skip =~ 'v:lua.vim.treesitter' && !exists('b:ts_highlight')
     let skip = "0"
   else
     execute "if " .. skip .. "| let skip = '0' | endif"
@@ -678,6 +679,7 @@ fun! matchit#MultiMatch(spflag, mode)
   let middlepat = substitute(middlepat, ',', '\\|', 'g')
 
   if skip =~ 'synID' && !(has("syntax") && exists("g:syntax_on"))
+        \ || skip =~ 'v:lua.vim.treesitter' && !exists('b:ts_highlight')
     let skip = '0'
   else
     try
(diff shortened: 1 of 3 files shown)


justinmk (approved):
No objection, though https://github.com/neovim/neovim/issues/15462 can remain open to track a full Lua rewrite of matchit.

clason:
Can you explain (also in the commit message) what this patch does and how it improves the situation with treesitter (without regressing on legacy syntax)?

clason:
(And, yes, this would be a stopgap until the planned total rewrite, which cannot come too soon for me.)

milisims:
When treesitter is enabled, by default syntax groups are not defined, but these
groups are used to identify where to skip matches in matchit and matchparen.

This patch does three things:
1. If syntax is enabled regardless of treesitter (`vim.bo.syntax='on'`): Use original implementation.
2. If treesitter is enabled and syntax is not: Match the syntax groups (i.e. `comment\|string`) against treesitter captures to check for skipped groups.
3. Add an explicit treesitter syntax for marking captures to skip: matchit uses `b:match_skip` to determine what counts as skippable Where 's:comment\|string' uses a match of the named syntax groups against a regex match of comment\|string, 't:comment\|string' now uses vim regex to match against the names of the treesitter capture groups.

clason:
Hmm, I'm not sure this is going to work in its current form: the expected captures don't match well to the standard captures we use: https://neovim.io/doc/user/treesitter.html#treesitter-highlight-groups

I'd feel much better if this PR came with tests, but the bundled plugins are unfortunately not covered.

milisims:
Comment and strings work very well, which is a huge improvement, and is the default for ftplugins that don’t define `b:match_skip`. Other languages use ‘special’ and ‘character’ like c, which works fine.

I went through every bundled ftplugin for `match_skip`, and I found:
- basic: custom match_skip, requires syntax. Probably could be edited to work pretty well with a test if ts is loaded and just defining a `t:` match skip.
- clojure: regex syn group is highlighted as string.regex which already works (the match isn’t anchored). string/comment work
- erlang: erlangmodifier and erlangquotedatom would possibly be ignored, I don’t know how they’re parsed or highlighted in the ts grammar. otherwise comment/string work
- julia: needs a `t:` definition
- perl: comment and string work, but others wouldn’t without a `t:`
- php: custom, uses syntax but could fallback at least to comment and strings. possibly just a `t:` def
- ruby: custom, but most could be covered with `t:` I think
- rust: would need a `t:` but easy, comment & string still work
- vim: needs a `t:`
Most of these could be improved from by simply falling back to comment/strings if ts is enabled and syntax is not.

These langs define match_skip but have no ts grammar: eiffel, mediawiki, mf, modula2, zimbu, zsh
At least, not in nvim-treesitter.

clason:
Yes, I am specifically concerned about the "subcaptures". For example, `@string.special` or `@comment.documentation` -- shouldn't these be covered, too?

milisims:
As is, the spec is just matched against the group. So `‘comment\|string` matches against `comment.documentation`, which is how the matchit implementation works, so subgroups are all matched. Should we anchor it?

clason:
No, then that should be fine.

justinmk:
https://github.com/neovim/neovim/pull/31861 is underway, should we merge this meanwhile? If not then let's close it.

clason:
Low risk, so let's just merge (and backport).

PR #31567, opened 2024-12-13
vim-patch:9.1.0921: popupmenu logic is a bit convoluted

glepnir:
Problem:  popupmenu logic is a bit convoluted
Solution: slightly refactor logic and use MIN/MAX() macros to simplify
          (glepnir)

Define the MAX/MIN macros. Since we support some older platforms, C compilers may not be as smart. This helps reduce unnecessary if statements and redundant ternary expressions. Pre-calculate some expressions by defining variables. Remove unnecessary parentheses. Adjust certain lines to avoid exceeding 80 columns.

closes: vim/vim#16205

https://github.com/vim/vim/commit/c942f84aadffd0c8969ecf81e3e9103722b2714f

### DIFF
--- a/src/nvim/popupmenu.c
+++ b/src/nvim/popupmenu.c
@@ -234,12 +234,7 @@ void pum_display(pumitem_T *array, int size, int selected, bool array_changed, i
     }
 
     // Figure out the size and position of the pum.
-    if (size < PUM_DEF_HEIGHT) {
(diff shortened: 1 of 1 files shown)


zeertzjq:
You are missing two changes. Closing in favor of #31566

glepnir:
aha i missed 31566 ❤️

PR #31576, opened 2024-12-14
fix(api): nvim_win_set_buf(0, 0) fails if 'winfixbuf' is set

phanen:
## Problem
With `winfixbuf` enabled, `nvim_win_set_buf` and `nvim_set_current_buf` cannot switch to current buffer.

Reproduce:
```lua
vim.wo.winfixbuf = true
vim.api.nvim_win_set_buf(0, vim.api.nvim_get_current_buf())
```

```lua
vim.wo.winfixbuf = true
vim.api.nvim_set_current_buf(0)
```

### DIFF
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -888,7 +888,7 @@ void nvim_set_current_buf(Buffer buffer, Error *err)
 {
   buf_T *buf = find_buffer_by_handle(buffer, err);
 
-  if (!buf) {
+  if (!buf || curwin->w_buffer == buf) {
     return;
   }
 
--- a/src/nvim/api/window.c
+++ b/src/nvim/api/window.c
@@ -59,7 +59,7 @@ void nvim_win_set_buf(Window window, Buffer buffer, Error *err)
 {
   win_T *win = find_window_by_handle(window, err);
   buf_T *buf = find_buffer_by_handle(buffer, err);
-  if (!win || !buf) {
+  if (!win || !buf || win->w_buffer == buf) {
     return;
   }
 
--- a/test/functional/options/winfixbuf_spec.lua
+++ b/test/functional/options/winfixbuf_spec.lua
@@ -8,8 +8,8 @@ describe("Nvim API calls with 'winfixbuf'", function()
     clear()
   end)
 
-  it("Calling vim.api.nvim_win_set_buf with 'winfixbuf'", function()
-    local results = exec_lua([[
+  it('vim.api.nvim_win_set_buf on non-current buffer', function()
+    local ok = exec_lua([[
       local function _setup_two_buffers()
         local buffer = vim.api.nvim_create_buf(true, true)
 
@@ -23,16 +23,16 @@ describe("Nvim API calls with 'winfixbuf'", function()
 
       local other_buffer = _setup_two_buffers()
       local current_window = 0
-      local results, _ = pcall(vim.api.nvim_win_set_buf, current_window, other_buffer)
+      local ok, _ = pcall(vim.api.nvim_win_set_buf, current_window, other_buffer)
 
-      return results
+      return ok
     ]])
 
-    assert(results == false)
+    assert(not ok)
   end)
 
-  it("Calling vim.api.nvim_set_current_buf with 'winfixbuf'", function()
-    local results = exec_lua([[
+  it('vim.api.nvim_set_current_buf on non-current buffer', function()
+    local ok = exec_lua([[
       local function _setup_two_buffers()
         local buffer = vim.api.nvim_create_buf(true, true)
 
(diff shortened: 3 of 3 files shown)


src/nvim/api/window.c
```diff
@@ -63,6 +63,10 @@ void nvim_win_set_buf(Window window, Buffer buffer, Error *err)
     return;
   }
 
+  if (win->w_buffer == buf) {
+    return;
```
justinmk, line 67: probably makes sense to combine with the condition on line 62?

same for the other case.

test/functional/options/winfixbuf_spec.lua
```diff
@@ -8,7 +8,7 @@ describe("Nvim API calls with 'winfixbuf'", function()
     clear()
   end)
 
-  it("Calling vim.api.nvim_win_set_buf with 'winfixbuf'", function()
+  it("Calling vim.api.nvim_win_set_buf on other buffer with 'winfixbuf'", function()
```
justinmk, line 11: let's unverbose these test names. there is no need to mention 'winfixbuf' in the tests since they are in a `describe()` scope that already mentions 'winfixbuf'

```suggestion
  it("vim.api.nvim_win_set_buf on non-current buffer", function()
```

test/functional/options/winfixbuf_spec.lua
```diff
@@ -52,4 +52,24 @@ describe("Nvim API calls with 'winfixbuf'", function()
 
     assert(results == false)
   end)
+
+  it('Calling vim.api.nvim_win_set_buf on current buffer with `winfixbuf`', function()
+    local results = exec_lua([[
+      vim.wo.winfixbuf = true
+      local results, _ = pcall(vim.api.nvim_win_set_buf, 0, vim.api.nvim_get_current_buf())
+      return results
```
justinmk, line 60: this is mislelading because `results` actually contains the status, not the result. The usual convention is to name this `ok` or `status`. Let's do that for all tests in this file.

```suggestion
      local ok, _ = pcall(vim.api.nvim_win_set_buf, 0, vim.api.nvim_get_current_buf())
      return ok
```

test/functional/options/winfixbuf_spec.lua
```diff
@@ -45,11 +45,31 @@ describe("Nvim API calls with 'winfixbuf'", function()
...
+  end)
+
+  it('vim.api.nvim_win_set_buf on current buffer', function()
+    local ok = exec_lua([[
+      vim.wo.winfixbuf = true
+      local ok, _ = pcall(vim.api.nvim_win_set_buf, 0, vim.api.nvim_get_current_buf())
+      return ok
+    ]])
+
+    assert(ok)
```
justinmk, line 63: sorry, just noticed that `pcall` is not needed here because we don't expect failure. same for the other case. in order to actually assert something useful, maybe we should check that the buffer didn't change.

```suggestion
      local curbuf = vim.api.nvim_get_current_buf()
      vim.api.nvim_win_set_buf(0, curbuf)
      assert(vim.api.nvim_get_current_buf() == curbuf)
    ]])
```

justinmk (approved):

src/nvim/api/vim.c
```diff
@@ -888,7 +888,7 @@ void nvim_set_current_buf(Buffer buffer, Error *err)
 {
   buf_T *buf = find_buffer_by_handle(buffer, err);
 
-  if (!buf) {
+  if (!buf || curwin->w_buffer == buf) {
     return;
```
justinmk, line 892: what happens if we just delete the `w_p_wfb ` check below? `do_buffer` already checks it. then this problem wouldn't have happened. @ColinKennedy

src/nvim/api/window.c
```diff
@@ -59,7 +59,7 @@ void nvim_win_set_buf(Window window, Buffer buffer, Error *err)
 {
   win_T *win = find_window_by_handle(window, err);
   buf_T *buf = find_buffer_by_handle(buffer, err);
-  if (!win || !buf) {
+  if (!win || !buf || win->w_buffer == buf) {
```
justinmk, line 62: what happens if we just delete the `w_p_wfb ` check below? `win_set_buf` should already check it. then this problem wouldn't have happened. @ColinKennedy

src/nvim/api/vim.c
ColinKennedy: The unittests are fairly thorough and this PR covers the other cases. As long as the tests all pass we should be good.

ColinKennedy: I personally prefer the "add the check at the caller instead of in the core" because if gives callers more control but we shouldn't double check if it can be avoided.

src/nvim/api/window.c
justinmk: Followup: https://github.com/neovim/neovim/pull/31595

src/nvim/api/vim.c
justinmk: https://github.com/neovim/neovim/pull/31595

PR #31577, opened 2024-12-14
fix(diagnostic): support function type of float in open_float()

### LINKED ISSUE #31573: `vim.diagnostic.config` does not apply options returned from functions
### Problem

`:h vim.diagnostic.Opts` says

```
    Each of the configuration options below accepts one of the following:
    • `false`: Disable this feature
    • `true`: Enable this feature, use default settings.
    • `table`: Enable this feature with overrides. Use an empty table to use
      default values.
    • `function`: Function with signature (namespace, bufnr) that returns any
      of the above.
...
      • {float}?             (`boolean|vim.diagnostic.Opts.Float|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Float`)
                             Options for floating windows. See
                             |vim.diagnostic.Opts.Float|.
```

but

```lua
vim.diagnostic.config({
  float = {
    border = 'rounded',
    header = '',
    scope = 'cursor',
  },
})
```

and

```lua
vim.diagnostic.config({
  float = function (namespace, bufnr)
    return {
      border = 'rounded',
      header = '',
      scope = 'cursor',
    }
  end,
})
```

have totally different behaviors; the function version doesn't seem to do anything at all.

### Steps to reproduce

Add

```lua
vim.diagnostic.config({
  float = function (namespace, bufnr)
    return {
      border = 'rounded',
      header = '',
      scope = 'cursor',
    }
  end,
})
```

to your config and open a diagnostic floating window `vim.diagnostic.show_float()`

### Expected behavior

Should look the same as 

```lua
vim.diagnostic.config({
  float = {
    border = 'rounded',
    header = '',
    scope = 'cursor',
  },
})
```

### Nvim version (nvim -v)

 v0.10.2 

n/a

### Operating system/version

macOS 15.2

### Terminal name/version

Terminal.app

xterm-256color

### Installation

brew

glepnir:
Problem: the float field of vim.diagnostic.config can be a function type but function was ignored in open_float function.

Solution: check global_diagnostic_options.float type before use for funciton type

Fix #31573

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -92,7 +92,7 @@ end
 --- @field signs? boolean|vim.diagnostic.Opts.Signs|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Signs
 ---
 --- Options for floating windows. See |vim.diagnostic.Opts.Float|.
---- @field float? boolean|vim.diagnostic.Opts.Float|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Float
+--- @field float? boolean|vim.diagnostic.Opts.Float|fun(namespace?: integer|integer[], bufnr:integer): vim.diagnostic.Opts.Float
 ---
 --- Options for the statusline component.
 --- @field status? vim.diagnostic.Opts.Status
@@ -2408,10 +2408,13 @@ function M.open_float(opts, ...)
     -- Resolve options with user settings from vim.diagnostic.config
     -- Unlike the other decoration functions (e.g. set_virtual_text, set_signs, etc.) `open_float`
     -- does not have a dedicated table for configuration options; instead, the options are mixed in
-    -- with its `opts` table which also includes "keyword" parameters. So we create a dedicated
-    -- options table that inherits missing keys from the global configuration before resolving.
-    local t = global_diagnostic_options.float
-    local float_opts = vim.tbl_extend('keep', opts, type(t) == 'table' and t or {})
+    -- with its `opts` table. We create a dedicated options table (`float_opts`) that inherits
+    -- missing keys from the global configuration (`global_diagnostic_options.float`), which can
+    -- be a table or a function.
+    local o = global_diagnostic_options
+    local t = type(o.float) == 'table' and o.float
+      or (type(o.float) == 'function' and o.float(opts.namespace, bufnr) or {})
+    local float_opts = vim.tbl_extend('keep', opts, t)
     opts = get_resolved_options({ float = float_opts }, nil, bufnr).float
   end
 
--- a/test/functional/lua/diagnostic_spec.lua
+++ b/test/functional/lua/diagnostic_spec.lua
@@ -3215,28 +3215,40 @@ describe('vim.diagnostic', function()
       )
     end)
 
-    it(
-      'creates floating window and returns float bufnr and winnr without header, if requested',
-      function()
-        -- One line (since no header):
-        --    1. <msg>
-        eq(
-          1,
-          exec_lua(function()
-            local diagnostics = {
-              _G.make_error('Syntax error', 0, 1, 0, 3),
-            }
-            vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
-            vim.diagnostic.set(_G.diagnostic_ns, _G.diagnostic_bufnr, diagnostics)
-            local float_bufnr, winnr =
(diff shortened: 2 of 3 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -1824,8 +1824,13 @@ function M.open_float(opts, ...)
     -- does not have a dedicated table for configuration options; instead, the options are mixed in
     -- with its `opts` table which also includes "keyword" parameters. So we create a dedicated
     -- options table that inherits missing keys from the global configuration before resolving.
-    local t = global_diagnostic_options.float
-    local float_opts = vim.tbl_extend('keep', opts, type(t) == 'table' and t or {})
```
justinmk, line 1828: can't we just change this to:

```lua
local float_opts = vim.tbl_extend('keep', opts, type(t) == 'table' and t or (
    type(t) == 'function' and global_diagnostic_options.float(0, bufnr) or {}
))

also, does the comment above need an small update?

```

runtime/lua/vim/diagnostic.lua
```diff
@@ -1824,8 +1824,13 @@ function M.open_float(opts, ...)
     -- does not have a dedicated table for configuration options; instead, the options are mixed in
     -- with its `opts` table which also includes "keyword" parameters. So we create a dedicated
     -- options table that inherits missing keys from the global configuration before resolving.
-    local t = global_diagnostic_options.float
```
justinmk, line 1827: can't we just change this to:

    local t = global_diagnostic_options.float or global_diagnostic_options.float(0, bufnr)

glepnir: yes i forget it use `if_nil` to set a default value always.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1822,10 +1822,11 @@ function M.open_float(opts, ...)
...
     -- Unlike the other decoration functions (e.g. set_virtual_text, set_signs, etc.) `open_float`
     -- does not have a dedicated table for configuration options; instead, the options are mixed in
-    -- with its `opts` table which also includes "keyword" parameters. So we create a dedicated
-    -- options table that inherits missing keys from the global configuration before resolving.
-    local t = global_diagnostic_options.float
-    local float_opts = vim.tbl_extend('keep', opts, type(t) == 'table' and t or {})
+    -- with its `opts` table. We create a dedicated options table (`float_opts`) that inherits
+    -- missing keys from the global configuration (`global_diagnostic_options.float`), which can
+    -- be a table or a function.
+    local t = global_diagnostic_options.float or global_diagnostic_options.float(0, bufnr)
```
justinmk, line 1828: oh I may have made a mistake, should it be more like this?:

```suggestion
    local o = global_diagnostic_options
    local t = type(o.float) == 'table' and o.float or (type(o.float) == 'function' and o.float(0, bufnr) or {})
```

test/functional/lua/diagnostic_spec.lua
```diff
@@ -1749,6 +1749,18 @@ describe('vim.diagnostic', function()
           return _G.count_extmarks(_G.diagnostic_bufnr, _G.diagnostic_ns)
         end)
       )
+
+      eq(
```
justinmk, line 1753: -- `float` option defined as a function.

test/functional/lua/diagnostic_spec.lua
```diff
@@ -1749,6 +1749,18 @@ describe('vim.diagnostic', function()
...
+        true,
+        exec_lua(function()
+          vim.diagnostic.config({
+            float = function(_, _)
+              return { border = 'rounded' }
+            end,
+          })
+          return type(vim.diagnostic.config().float) == 'function'
```
justinmk, line 1761: instead of asserting this (which doesn't really test much), can we assert `eq('rounded', ...)` ?

And before this assertion, check that border is NOT 'rounded', to make sure we are actually testing something.

glepnir: I have added in open_float

test/functional/lua/diagnostic_spec.lua
```diff
@@ -2500,6 +2513,37 @@ describe('vim.diagnostic', function()
       )
     end)
 
+    it('', function()
```
justinmk: 👀

justinmk:
rebase, and please double-check if an existing test can be updated instead (if it saves a lot of code)

PR #31585, opened 2024-12-15
fix(diagnostic): vim.diagnostic.setqflist() opens location list on first call

delatorrejuanchi:
Problem: #31557 introduced a bug where calling `vim.diagnostic.setqflist` with `{ open = true }` attempts to open the location list instead of the diagnostics quickfix list if it didn't exist before. This is because we were using `qf_id` to decide if we should open the location list or the quickfix list ([ref](https://github.com/delatorrejuanchi/neovim/blob/6c2c77b128e74c69be0d4fd79637c742f7eaba34/runtime/lua/vim/diagnostic.lua#L873-L884)), but `qf_id` is `nil` when there is no existing diagnostics quickfix list with a given title (`'Diagnostics'` by default).

Solution: revert to using `loclist` to decide whether to open the location list or the quickfix list.

This PR also adds some tests to assert the existing functionality of `vim.diagnostic.setqflist`.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -871,7 +871,7 @@ local function set_list(loclist, opts)
   end
 
   if open then
-    if qf_id then
+    if not loclist then
       -- First navigate to the diagnostics quickfix list.
       local nr = vim.fn.getqflist({ id = qf_id, nr = 0 }).nr
       api.nvim_command(nr .. 'chistory')
--- a/test/functional/lua/diagnostic_spec.lua
+++ b/test/functional/lua/diagnostic_spec.lua
@@ -5,6 +5,7 @@ local command = n.command
 local clear = n.clear
 local exec_lua = n.exec_lua
 local eq = t.eq
+local neq = t.neq
 local matches = t.matches
 local api = n.api
 local pcall_err = t.pcall_err
@@ -3261,6 +3262,22 @@ describe('vim.diagnostic', function()
       end)
 
       neq(result[1], result[2])
+    end)
+
+    it('opens quickfix window when open=true', function()
+      local qf_winid = exec_lua(function()
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.diagnostic.set(_G.diagnostic_ns, _G.diagnostic_bufnr, {
+          _G.make_error('Error', 1, 1, 1, 1),
+        })
+
+        vim.diagnostic.setqflist({ open = true })
+
+        return vim.fn.getqflist({ winid = 0 }).winid
+      end)
+
+      neq(0, qf_winid)
     end)
   end)
 
--- a/test/functional/lua/diagnostic_spec.lua
+++ b/test/functional/lua/diagnostic_spec.lua
@@ -3212,6 +3212,58 @@ describe('vim.diagnostic', function()
     end)
   end)
 
+  describe('setqflist()', function()
+    it('updates existing diagnostics quickfix if one already exists', function()
+      local result = exec_lua(function()
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.fn.setqflist({}, ' ', { title = 'Diagnostics' })
+        local diagnostics_qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        vim.diagnostic.setqflist({ title = 'Diagnostics' })
+        local qf_id = vim.fn.getqflist({ id = 0, nr = '$' }).id
+
+        return { diagnostics_qf_id, qf_id }
+      end)
+
+      eq(result[1], result[2])
+    end)
+
+    it('navigates to existing diagnostics quickfix if one already exists and open=true', function()
+      local result = exec_lua(function()
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.fn.setqflist({}, ' ', { title = 'Diagnostics' })
+        local diagnostics_qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        vim.fn.setqflist({}, ' ', { title = 'Other' })
+
+        vim.diagnostic.setqflist({ title = 'Diagnostics', open = true })
+        local qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        return { diagnostics_qf_id, qf_id }
+      end)
+
+      eq(result[1], result[2])
+    end)
+
+    it('sets new diagnostics quickfix as active when open=true', function()
+      local result = exec_lua(function()
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.fn.setqflist({}, ' ', { title = 'Other' })
+        local other_qf_id = vim.fn.getqflist({ id = 0 }).id
+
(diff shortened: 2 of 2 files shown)


delatorrejuanchi:
It seems to me some checks are failing due to flakiness, but I'm uncertain about how to proceed. If someone could point me in the right direction, that would be appreciated! 🙂

runtime/lua/vim/diagnostic.lua
```diff
@@ -871,7 +871,7 @@ local function set_list(loclist, opts)
   end
 
   if open then
-    if qf_id then
+    if not loclist then
```
jfly, line 874: I think this it the wrong fix? The next line of code uses `qf_id` to figure out the index of the diagnostics quickfix list, and I don't know what passing `id = nil` would do (probably select the current quickfix list, which isn't necessarily the right thing).

Instead, we should compute `qf_id` right after creating a new quickfix. That is, above right after `setqflist`, we should have something like:

```lua
    -- Get the id of the newly created quickfix list.
    if qf_id == nil then
      qf_id = vim.fn.getqflist({ id = 0 }).id
    end
```

That will ensure that `qf_id` is always set if we're dealing with the quickfix list.

If you wanted to add a test for this, I think you'd do something like:

1. Create a non-diagnostics quickfix list, open it.
2. Call `vim.diagnostic.setqflist({ open = true })`
3. Assert that the currently opened/selected quickfix list is the qf list from step 2.

I believe that with the code you have here, it would still have the qf list from step 1 open.

test/functional/lua/diagnostic_spec.lua
```diff
@@ -3212,6 +3213,67 @@ describe('vim.diagnostic', function()
...
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.fn.setqflist({}, ' ', { title = 'Diagnostics' })
+        local diagnostics_qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        vim.diagnostic.set(_G.diagnostic_ns, _G.diagnostic_bufnr, {
+          _G.make_error('Error', 1, 1, 1, 1),
+        })
+
+        vim.diagnostic.setqflist()
```
jfly, line 3228: nit (feel free to ignore): it might be nice to be explicit here and specify the title, rather than relying upon the default value (which happens to be "Diagnostics")

test/functional/lua/diagnostic_spec.lua
```diff
@@ -3212,6 +3213,67 @@ describe('vim.diagnostic', function()
...
+        vim.fn.setqflist({}, ' ', { title = 'Diagnostics' })
+        local diagnostics_qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        vim.fn.setqflist({}, ' ', { title = 'Other' })
+
+        vim.diagnostic.set(_G.diagnostic_ns, _G.diagnostic_bufnr, {
+          _G.make_error('Error', 1, 1, 1, 1),
+        })
+
+        vim.diagnostic.setqflist()
```
jfly, line 3250: nit (feel free to ignore): I'd explicitly set `open = true` here. I would also be explicit about the title.

test/functional/lua/diagnostic_spec.lua
```diff
@@ -3212,6 +3213,67 @@ describe('vim.diagnostic', function()
...
+    it('updates existing diagnostics quickfix if one already exists', function()
+      local result = exec_lua(function()
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.fn.setqflist({}, ' ', { title = 'Diagnostics' })
+        local diagnostics_qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        vim.diagnostic.set(_G.diagnostic_ns, _G.diagnostic_bufnr, {
+          _G.make_error('Error', 1, 1, 1, 1),
+        })
```
jfly, line 3226: nit: I don't understand how this affects the test. We're not looking at the actual contents of the qf list. Consider removing? (same comment for the other tests you added).

jfly (commented):
Thanks for fixing the bug I introduced, and adding tests!

test/functional/lua/diagnostic_spec.lua
delatorrejuanchi: Addressed in a9fde686434a

test/functional/lua/diagnostic_spec.lua
delatorrejuanchi: Addressed in a9fde686434a

runtime/lua/vim/diagnostic.lua
delatorrejuanchi: Unless I'm missing something, I believe this is fine. Let me explain my reasoning here:

- if `qf_id` is `nil`, then that means we've not found an existing diagnostics quickfix
- for that reason, the call to `vim.fn.setqflist` would have created a new quickfix list and made it active
- `vim.fn.getqflist({ id = qf_id, nr = 0 }).nr` will return the `nr` of the current quickfix list, which is the one we just created

- if `qf_id` is not `nil`, then that means we've found an existing diagnostics quickfix
- the call to `vim.fn.setqflist` with `'u'` would have updated it, instead of creating a new one
  - note that this does not make the diagnostics quickfix active, which is why we need to call `api.nvim_command(nr .. 'chistory')`
- `vim.fn.getqflist({ id = qf_id, nr = 0 }).nr` will return the `nr` of the existing diagnostics quickfix

> If you wanted to add a test for this, I think you'd do something like:
> Create a non-diagnostics quickfix list, open it.
> Call vim.diagnostic.setqflist({ open = true })
> Assert that the currently opened/selected quickfix list is the qf list from step 2.

When we call `vim.fn.setqflist` with the new diagnostics quickfix, it's also set as active as a new one is created. That means that even when `open=false`, if a quickfix list is already open, the new diagnostics quickfix list will be displayed as it's also active.
When `open=true`, `vim.fn.getqflist({ id = qf_id, nr = 0 }).nr` will return the `nr` of the newly created diagnostics quickfix, as it's been activated when it was created.

I've added a new test to describe this behavior: `sets new diagnostics quickfix as active`. Together with the other tests, it seems to me it covers the behavior we want to assert.

test/functional/lua/diagnostic_spec.lua
```diff
@@ -3212,6 +3213,74 @@ describe('vim.diagnostic', function()
...
+    end)
+
+    it('sets new diagnostics quickfix as active when open=true', function()
+      local result = exec_lua(function()
+        vim.api.nvim_win_set_buf(0, _G.diagnostic_bufnr)
+
+        vim.fn.setqflist({}, ' ', { title = 'Other' })
+        local other_qf_id = vim.fn.getqflist({ id = 0 }).id
+
+        vim.diagnostic.setqflist({ title = 'Diagnostics', open = true })
```
delatorrejuanchi, line 3258: With `open = false`, the new diagnostics quickfix would also be set as active.

I've added `open = true`, to assert that [these lines](https://github.com/delatorrejuanchi/neovim/blob/1eaf4637f49092a38d026ddc84840686a427b954/runtime/lua/vim/diagnostic.lua#L875-L877) change that behavior.

I'd be happy to add another test with `open=false`, if we want to be more thorough.

justinmk (approved):
Thanks! And thanks @jfly for reviewing.

PR #31590, opened 2024-12-16
fix(messages): no message kind for :undo messages

przepompownia:
Problem: cannot handle `:undo` and `:redo` messages in a special way, e.g. replace one by another.

Solution: add `undo` kind.

### DIFF
--- a/src/nvim/undo.c
+++ b/src/nvim/undo.c
@@ -1862,6 +1862,7 @@ static void u_doit(int startcount, bool quiet, bool do_buf_event)
     u_oldcount = -1;
   }
 
+  msg_ext_set_kind("undo");
   int count = startcount;
   while (count--) {
     // Do the change warning now, so that it triggers FileChangedRO when
--- a/test/functional/ui/messages_spec.lua
+++ b/test/functional/ui/messages_spec.lua
@@ -240,6 +240,36 @@ describe('ui/ext_messages', function()
         },
       },
     })
+
+    -- undo
+    feed('uu')
+    screen:expect({
+      grid = [[
+        ^                         |
+        {1:~                        }|*4
+      ]],
+      messages = {
+        {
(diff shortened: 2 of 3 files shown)


src/nvim/undo.c
```diff
@@ -1862,6 +1862,7 @@ static void u_doit(int startcount, bool quiet, bool do_buf_event)
     u_oldcount = -1;
   }
 
+  msg_ext_set_kind("undo");
```
przepompownia, line 1865: Please verify if there is no better place to this call.

justinmk (approved):

test/functional/ui/messages_spec.lua
```diff
@@ -826,6 +826,27 @@ describe('ui/ext_messages', function()
...
+        },
+      },
+    }
+
+    feed('u')
+    screen:expect {
+      messages = {
+        {
+          content = { { 'Already at oldest change' } },
+          kind = 'undo',
```
justinmk, line 845: before this PR, `kind = ''` here?

src/nvim/undo.c
luukvbaal: I think so yes, which makes the one you added in `u_undo_end()` redundant.
I pushed a change in which I removed it, moved the test to the umbrella kind test, and exchanged the messages for ones without the time of the change, which tends to result in flakiness (1 vs 0 seconds).

test/functional/ui/messages_spec.lua
luukvbaal: Yep

luukvbaal (approved):

src/nvim/undo.c
przepompownia: Indeed, I see that after removal the test case passes too. Thanks!

PR #31597, opened 2024-12-16
fix(wininfo): when freeing windows, free the lowest priority wininfo

bfredl:
On master (also before #31539) deleting a window could cause the used wininfo for a buffer to change. This is due to always removing the previous NULL wininfo when deleting a window, even if that wininfo had higher priority than the the deleted window's own wininfo.

Instead delete the wininfo with lowest priority. This retains the memory saving efect while not affecting the effective value of window options and so on.

### DIFF
--- a/test/functional/api/window_spec.lua
+++ b/test/functional/api/window_spec.lua
@@ -488,6 +488,48 @@ describe('API/win', function()
       assert_alive()
     end)
 
+    describe('after closing', function()
+      local buf, win0, win1, win2
+
+      before_each(function()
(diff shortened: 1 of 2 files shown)


test/functional/api/window_spec.lua
```diff
@@ -488,6 +488,48 @@ describe('API/win', function()
       assert_alive()
     end)
 
+    describe('after deleteting', function()
```
justinmk, line 491:
```suggestion
    describe('after deleting', function()
```

PR #31596, opened 2024-12-16
fix(api): generic error messages, not using TRY_WRAP

justinmk:
Problem:
- API functions using `try_start` directly, do not surface the underlying error message, and instead show generic messages.
- Error-handling code is duplicated in the API impl.
- Failure modes are not tested.

Solution:
- Use `TRY_WRAP`.
- Add tests.

continues https://github.com/neovim/neovim/pull/31595
related: https://github.com/neovim/neovim/pull/22734

PR #31600, opened 2024-12-16
refactor(api): always use TRY_WRAP

luukvbaal:
Problem:  Two separate try/end wrappers, that only marginally differ by
          restoring a few variables. Wrappers that don't restore
          previous state are dangerous to use in "api-fast" functions.
Solution: Remove wrappers that don't restore the previous state.
          Always use TRY_WRAP.

Continues #31595, #31596

### DIFF
--- a/src/nvim/api/private/helpers.c
+++ b/src/nvim/api/private/helpers.c
@@ -42,8 +42,10 @@
 
 /// Start block that may cause Vimscript exceptions while evaluating another code
 ///
-/// Used when caller is supposed to be operating when other Vimscript code is being
-/// processed and that “other Vimscript code” must not be affected.
+/// Used just in case caller is supposed to be operating when other Vimscript code
+/// is being processed and that “other Vimscript code” must not be affected.
+///
+/// @warning Avoid calling directly; use TRY_WRAP instead.
 ///
 /// @param[out]  tstate  Location where try state should be saved.
 void try_enter(TryState *const tstate)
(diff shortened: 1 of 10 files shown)


src/nvim/api/private/helpers.c
```diff
@@ -166,7 +126,19 @@ bool try_end(Error *err)
...
 
-  return ERROR_SET(err);
+  assert(!need_rethrow);
+  assert(!got_int);
+  assert(!did_throw);
+  assert(!did_emsg);
+  assert(msg_list == &tstate->private_msg_list);
+  assert(*msg_list == NULL);
+  assert(current_exception == NULL);
+  msg_list = (msglist_T **)tstate->msg_list;
```
justinmk, line 136:
```suggestion
  msg_list = (msglist_T **)tstate->msg_list;  // Restore the exception context.
```

src/nvim/api/tabpage.c
```diff
@@ -146,9 +146,10 @@ void nvim_tabpage_set_win(Tabpage tabpage, Window win, Error *err)
...
 
   if (tp == curtab) {
-    try_start();
-    win_goto(wp);
-    if (!try_end(err) && curwin != wp) {
+    TRY_WRAP(err, {
+      win_goto(wp);
+    });
+    if (!ERROR_SET(err) && curwin != wp) {
       api_set_error(err, kErrorTypeException, "Failed to switch to window %d", win);
```
justinmk, line 153: This probably can be dropped as in 022449b52236 , but harmless I guess.

justinmk (approved):

justinmk:
@Julian if you can find a test-case that reproduces the issue, that will decide whether the error is intentional or not. This change makes error handling more consistent and may raise errors that previously were hidden.

PR #31609, opened 2024-12-17
feat(terminal)!: include cursor position in TermRequest event data

gpanders:
When a plugin registers a TermRequest handler there is currently no way for the handler to know where the terminal's cursor position was when the sequence was received. This is often useful information, e.g. for OSC 133 sequences which are used to annotate shell prompts.

Modify the event data for the TermRequest autocommand to be a table instead of just a string. The "sequence" field of the table contains the sequence string and the "cursor" field contains the cursor position when the sequence was received.

To maintain consistency between TermRequest and TermResponse (and to future proof the latter), TermResponse's event data is also updated to be a table with a "sequence" field.

BREAKING CHANGE: event data for TermRequest and TermResponse is now a table

---

This enables plugins to do things like add signs to terminal buffers to indicate shell prompts. Demo:

[Screencast from 2024-12-17 10-44-18.webm](https://github.com/user-attachments/assets/120059fc-6d47-4349-8b93-2d019de95e14)

### DIFF
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -144,8 +144,8 @@ directory indicated in the request. >lua
   vim.api.nvim_create_autocmd({ 'TermRequest' }, {
     desc = 'Handles OSC 7 dir change requests',
     callback = function(ev)
-      if string.sub(vim.v.termrequest, 1, 4) == '\x1b]7;' then
-        local dir = string.gsub(vim.v.termrequest, '\x1b]7;file://[^/]*', '')
+      if string.sub(ev.data.sequence, 1, 4) == '\x1b]7;' then
+        local dir = string.gsub(ev.data.sequence, '\x1b]7;file://[^/]*', '')
         if vim.fn.isdirectory(dir) == 0 then
           vim.notify('invalid dir: '..dir)
           return
--- a/src/nvim/terminal.c
+++ b/src/nvim/terminal.c
@@ -186,7 +186,7 @@ struct terminal {
   char *selection_buffer;  ///< libvterm selection buffer
   StringBuilder selection;  ///< Growable array containing full selection data
 
-  StringBuilder termrequest_buffer;  ///< Growable array containing unfinished request payload
+  StringBuilder termrequest_buffer;  ///< Growable array containing unfinished request sequence
 
   size_t refcount;                  // reference count
 };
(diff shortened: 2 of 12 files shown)


src/nvim/terminal.c
```diff
@@ -229,7 +237,8 @@ static void schedule_termrequest(Terminal *term, char *payload, size_t payload_l
   term->pending.send = xmalloc(sizeof(StringBuilder));
   kv_init(*term->pending.send);
   multiqueue_put(main_loop.events, emit_termrequest, term, payload, (void *)payload_length,
-                 term->pending.send);
+                 term->pending.send, (void *)(intptr_t)term->cursor.row,
+                 (void *)(intptr_t)term->cursor.col);
```
gpanders, line 241: This doesn't completely work. In some cases, vterm seems to call the `on_osc` callback before it's updated its cursor position, so it reports the wrong position. This seems to happen more frequently when there is a lot of text (e.g. `ls -al` vs just `ls`).

runtime/lua/vim/_meta/vvars_extra.lua
```diff
@@ -63,8 +63,10 @@ error('Cannot require a meta file')
 --- complete item selected.
 --- @field height? integer
 --- @field width? integer Height of popup menu on |CompleteChanged|
---- @field row? integer Width of popup menu on |CompleteChanged|
---- Col count of popup menu on |CompleteChanged|, relative to screen.
+--- @field row? integer For |CompleteChanged|: width of popup menu. For |TermRequest|: current
+--- cursor row when escape sequence was received.
+--- For |CompleteChanged|: col count of popup menu, relative to screen. For |TermRequest|: current
+--- cursor column when escape sequence was received.
```
justinmk, line 69: Typically `:help v:event` should not need to enumerate every event for a field; rather the field should describe is general purpose, then each help doc for each event mentions the `v:event` fields that it sets.

IOW: `:help TermRequest` should mention these v:event fields. 

Also, `:help TermRequest` already mentions `:help event-data` but I see that it sets `args.data` to a string. I think in general we shouldn't be shy about adding more fields to `args` (and those fields should ideally match `v:event` where possible).

justinmk (approved):

runtime/lua/vim/_meta/vvars_extra.lua
gpanders: `args.data` has been the place where we put arbitrary event data, reserving the top level table for well-known (and documented) fields like "buf", "file", etc. If we start putting arbitrary data in the top-level then it (1) risks conflicts with the well-known fields (users can add arbitrary data with `nvim_exec_autocmds`), and (2) makes the `data` sub-field meaningless (or, at best, confusing).

I don't think we should add new event data to the top-level field, but we could consider making a breaking change to the TermRequest autocommand to make `args.data` a table instead. Then we would have `args.data.payload` for the actual escape sequence, and `args.data.row` and `args.data.col`.

justinmk: >  but we could consider making a breaking change to the TermRequest autocommand to make `args.data` a table instead. Then we would have `args.data.payload` for the actual escape sequence, and `args.data.row` and `args.data.col`.

SGTM

runtime/doc/terminal.txt
```diff
@@ -144,8 +144,8 @@ directory indicated in the request. >lua
   vim.api.nvim_create_autocmd({ 'TermRequest' }, {
     desc = 'Handles OSC 7 dir change requests',
     callback = function(ev)
-      if string.sub(vim.v.termrequest, 1, 4) == '\x1b]7;' then
-        local dir = string.gsub(vim.v.termrequest, '\x1b]7;file://[^/]*', '')
+      if string.sub(ev.data.payload, 1, 4) == '\x1b]7;' then
+        local dir = string.gsub(ev.data.payload, '\x1b]7;file://[^/]*', '')
```
justinmk, line 148: `ev.data.data` is probably the right name, though it's a bit redundant I guess. Not a blocker.

justinmk (approved):

clason:
@gpanders What's the status here? Do we want to get this into 0.11?

gpanders:
> @gpanders What's the status here? Do we want to get this into 0.11?

I'll take another look at it. We might be able to get it into 0.11 but I'm fine deferring it if it gets complicated.

clason:
Yup, not a blocker as far as I can tell, just triaging.

runtime/doc/terminal.txt
gpanders: I'm going to call it `sequence` instead, since that is almost universally the term used for these things.

src/nvim/terminal.c
gpanders: I think I got this working. The issue was that `term->cursor.row` is the cursor position in the window. vterm has no concept of the Nvim buffer, as far as it's concerned the window is the whole terminal. Anything not visible in the window is part of the scrollback buffer. So we need to convert the terminal/window cursor row into the buffer row using `row_to_linenr`.

However this does not get us all the way there, because an OSC sequence can be processed before Nvim/vterm have updated the terminal buffer contents (moving old lines into the scrollback and appending new lines to the end of the buffer). If we emit the event too soon, then the line number in the event payload will be invalid. So we also have to check to make sure `sb_pending` is zero (indicating that there are no pending scrollback lines) before emitting the event. If it is not, then we re-schedule the event.

This all seems to be working for me locally, but I'll try and write up a test case as well.

clason:
https://github.com/neovim/neovim/pull/32736

PR #31613, opened 2024-12-17
test(cursor_spec): global highlight definitions

luukvbaal:


luukvbaal:
IDK yet if the second commit is effective (or if it even makes sense), don't think it's possible to run cirrus CI on a fork?

PR #31615, opened 2024-12-17
vim-patch:9.1.{0936,0941,0942}: ComplMatchIns highlight

zeertzjq:
#### vim-patch:9.1.0936: cannot highlight completed text

Problem:  cannot highlight completed text
Solution: (optionally) highlight auto-completed text using the
          ComplMatchIns highlight group (glepnir)

closes: vim/vim#16173

https://github.com/vim/vim/commit/6a38aff218f5b99a1aed7edaa357df24b9092734

Co-authored-by: glepnir <glephunter@gmail.com>

#### vim-patch:9.1.0941: ComplMatchIns doesn't work after multibyte chars

Problem:  ComplMatchIns doesn't work after multibyte chars
          (after v9.1.0936)
Solution: Use (ptr - line) instead of wlv.col (zeertzjq).

closes: vim/vim#16233

https://github.com/vim/vim/commit/f4ccada5c372b2c14cc32490860c6995cd00268c

#### vim-patch:9.1.0942: a few typos were found

Problem:  a few typos were found
Solution: fix them (zeertzjq)

closes: vim/vim#16232

https://github.com/vim/vim/commit/d32bf0a06762f9ad08334d67b4d7f235f87f9063


echasnovski:
Also, the last character is not highlighted. Is it intentional?

Here is the demo:

https://github.com/user-attachments/assets/497e5a8b-c0eb-4f13-8dba-1d4d62f5485a

glepnir:
https://github.com/vim/vim/pull/16244

echasnovski:
> [vim/vim#16244](https://github.com/vim/vim/pull/16244)

Yes, it seems to solve the off-by-one issue, but not the coloring.
I think `ComplMatchIns` should be completely invisible by default and strictly opt-in for color schemes. Linking to `Normal` does not achieve that.

zeertzjq:
Perhaps it should default to having `NONE` for both foreground and background.

echasnovski:
> Perhaps it should default to having `NONE` for both foreground and background.

Yes, I'd say so. Do you think it is a reasonable default for Vim also, or should there be a PR for Neovim separately?

zeertzjq:
This seems to be equivalent to linking to `Normal` in Vim, but I'll add it to vim/vim#16244 anyway.

zeertzjq:
No, I mean it's equivalent to linking to `Normal` in Vim even if `Normal` has background and foreground color.

justinmk:
> `ComplMatchIns` should be completely invisible by default and strictly opt-in for color schemes.

Discussed in https://github.com/neovim/neovim/pull/32251#issuecomment-2624012534

PR #31618, opened 2024-12-18
fix(coverity): error handling (CHECKED_RETURN)

justinmk:
CID 516406:  Error handling issues  (CHECKED_RETURN)
    /src/nvim/api/vimscript.c: 284 in nvim_call_dict_function()
    278       Object rv = OBJECT_INIT;
    279
    280       typval_T rettv;
    281       bool mustfree = false;
    282       switch (dict.type) {
    283       case kObjectTypeString:
    >>>     CID 516406:  Error handling issues  (CHECKED_RETURN)
    >>>     Calling "eval0" without checking return value (as is done elsewhere 10 out of 12 times).
    284         TRY_WRAP(err, {
    285           eval0(dict.data.string.data, &rettv, NULL, &EVALARG_EVALUATE);
    286           clear_evalarg(&EVALARG_EVALUATE, NULL);
    287         });
    288         if (ERROR_SET(err)) {
    289           return rv;

### DIFF
--- a/src/nvim/api/vimscript.c
+++ b/src/nvim/api/vimscript.c
@@ -229,10 +229,9 @@ static Object _call_function(String fn, Array args, dict_T *self, Arena *arena, 
   funcexe.fe_selfdict = self;
 
   TRY_WRAP(err, {
-    // call_func() retval is deceptive, ignore it.  Instead we set `msg_list`
-    // (see above) to capture abort-causing non-exception errors.
-    call_func(fn.data, (int)fn.size, &rettv, (int)args.size,
-              vim_args, &funcexe);
+    // call_func() retval is deceptive, ignore it.  Instead TRY_WRAP sets `msg_list` to capture
+    // abort-causing non-exception errors.
+    (void)call_func(fn.data, (int)fn.size, &rettv, (int)args.size, vim_args, &funcexe);
   });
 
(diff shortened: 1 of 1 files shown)


src/nvim/api/vimscript.c
```diff
@@ -282,7 +282,10 @@ Object nvim_call_dict_function(Object dict, String fn, Array args, Arena *arena,
   switch (dict.type) {
   case kObjectTypeString:
     TRY_WRAP(err, {
-      eval0(dict.data.string.data, &rettv, NULL, &EVALARG_EVALUATE);
+      int rv = eval0(dict.data.string.data, &rettv, NULL, &EVALARG_EVALUATE);
+      if (rv != OK && !ERROR_SET(err)) {
+        abort();
```
justinmk, line 287: Assume this never happens. If it does, then coverity was right and I need to restore the explicit handling removed in 022449b52236

zeertzjq: `:call nvim_call_dict_function('{', 'foo', [])`

zeertzjq: `err` has never been used up to this point (as `try_leave()` hasn't been executed yet), so `!ERROR_SET(err)` is always true here.

PR #12826, opened 2020-09-01
treesitter: use lua-match? instead of match?

vigoux:
Following [this comment](https://github.com/neovim/neovim/pull/12739#issuecomment-681139014), this PR reverts the switch to using lua for the default `match?` predicate, to use the vim implementation.

The lua-based `match?` predicate is accessible through `lua-match?`.
We still expose `vim-match?` to avoid confusion, and to be clear.

We also fire an error when a predicate has no handlers.

And also do a drive-by simplification of `Query:match_preds`.

PR #31625, opened 2024-12-18
perf(decor): improve `iter_captures()` cache

vanaigr:
Currently `iter_captures()` caches predicate and directives results separately. For each of the 2 lookups, it has to call into C to get the match id. Additionally, both `match_preds()` and `apply_directives()` call to `match:captures()`, which creates a new table each time.

I updated `iter_captures()` to only have 1 cache with the final metadata. This is possible because if predicates match, the directives are always applied; and if predicates don't match, the query match is removed and no future captures from it will be returned.

`match:captures()` and `match:info()` (which call to C) are now called only once.

Also split predicates and directives into 2 lists.

Theoretically, this should also make `iter_matches()` faster.

### Results

Median, in ms:

|before|after|
|-|-|
|42.5|25.6|

### DIFF
--- a/runtime/lua/vim/treesitter/query.lua
+++ b/runtime/lua/vim/treesitter/query.lua
@@ -7,6 +7,59 @@ local memoize = vim.func._memoize
 
 local M = {}
 
+local function is_directive(name)
+  return string.sub(name, -1) == '!'
+end
+
+---@nodoc
+---@class vim.treesitter.query.ProcessedPredicate
+---@field [1] string predicate name
+---@field [2] boolean should match
+---@field [3] (integer|string)[] the original predicate
+
+---@alias vim.treesitter.query.ProcessedDirective (integer|string)[]
+
+---@nodoc
+---@class vim.treesitter.query.ProcessedPattern {
+---@field predicates vim.treesitter.query.ProcessedPredicate[]
+---@field directives vim.treesitter.query.ProcessedDirective[]
+
+--- Splits the query patterns into predicates and directives.
+---@param patterns table<integer, (integer|string)[][]>
+---@return table<integer, vim.treesitter.query.ProcessedPattern>
+local function process_patterns(patterns)
+  ---@type table<integer, vim.treesitter.query.ProcessedPattern>
+  local processed_patterns = {}
+
+  for k, pattern_list in pairs(patterns) do
+    ---@type vim.treesitter.query.ProcessedPredicate[]
+    local predicates = {}
+    ---@type vim.treesitter.query.ProcessedDirective[]
+    local directives = {}
+
+    for _, pattern in ipairs(pattern_list) do
+      -- Note: tree-sitter strips the leading # from predicates for us.
+      local pred_name = pattern[1]
+      ---@cast pred_name string
+
+      if is_directive(pred_name) then
+        table.insert(directives, pattern)
+      else
+        local should_match = true
+        if pred_name:match('^not%-') then
+          pred_name = pred_name:sub(5)
+          should_match = false
+        end
+        table.insert(predicates, { pred_name, should_match, pattern })
+      end
+    end
+
+    processed_patterns[k] = { predicates = predicates, directives = directives }
+  end
+
+  return processed_patterns
+end
+
 ---@nodoc
 ---Parsed query, see |vim.treesitter.query.parse()|
 ---
@@ -15,6 +68,7 @@ local M = {}
 ---@field captures string[] list of (unique) capture names defined in query
 ---@field info vim.treesitter.QueryInfo query context (e.g. captures, predicates, directives)
 ---@field query TSQuery userdata query object
+---@field private _processed_patterns table<integer, vim.treesitter.query.ProcessedPattern>
 local Query = {}
 Query.__index = Query
 
@@ -33,6 +87,7 @@ function Query.new(lang, ts_query)
     patterns = query_info.patterns,
   }
   self.captures = self.info.captures
+  self._processed_patterns = process_patterns(self.info.patterns)
   return self
 end
 
@@ -751,67 +806,50 @@ function M.list_predicates()
   return vim.tbl_keys(predicate_handlers)
 end
 
-local function xor(x, y)
-  return (x or y) and not (x and y)
-end
-
-local function is_directive(name)
-  return string.sub(name, -1) == '!'
-end
-
 ---@private
----@param match TSQueryMatch
+---@param pattern_i integer
+---@param predicates vim.treesitter.query.ProcessedPredicate[]
+---@param captures table<integer, TSNode[]>
 ---@param source integer|string
-function Query:match_preds(preds, pattern, captures, source)
-  for _, pred in pairs(preds) do
-    -- Here we only want to return if a predicate DOES NOT match, and
-    -- continue on the other case. This way unknown predicates will not be considered,
-    -- which allows some testing and easier user extensibility (#12173).
-    -- Also, tree-sitter strips the leading # from predicates for us.
-    local is_not = false
+---@return boolean whether the predicates match
+function Query:_match_predicates(predicates, pattern_i, captures, source)
+  for _, predicate in ipairs(predicates) do
+    local processed_name = predicate[1]
+    local should_match = predicate[2]
+    local orig_predicate = predicate[3]
 
-    -- Skip over directives... they will get processed after all the predicates.
-    if not is_directive(pred[1]) then
-      local pred_name = pred[1]
-      if pred_name:match('^not%-') then
-        pred_name = pred_name:sub(5)
-        is_not = true
-      end
+    local handler = predicate_handlers[processed_name]
+    if not handler then
+      error(string.format('No handler for %s', orig_predicate[1]))
+      return false
+    end
 
-      local handler = predicate_handlers[pred_name]
-
-      if not handler then
-        error(string.format('No handler for %s', pred[1]))
-        return false
-      end
-
-      local pred_matches = handler(captures, pattern, source, pred)
-
-      if not xor(is_not, pred_matches) then
-        return false
-      end
+    local does_match = handler(captures, pattern_i, source, orig_predicate)
+    if does_match ~= should_match then
+      return false
     end
   end
   return true
 end
 
 ---@private
----@param match TSQueryMatch
+---@param pattern_i integer
+---@param directives vim.treesitter.query.ProcessedDirective[]
+---@param source integer|string
+---@param captures table<integer, TSNode[]>
 ---@return vim.treesitter.query.TSMetadata metadata
-function Query:apply_directives(preds, pattern, captures, source)
+function Query:_apply_directives(directives, pattern_i, captures, source)
   ---@type vim.treesitter.query.TSMetadata
   local metadata = {}
 
-  for _, pred in pairs(preds) do
-    if is_directive(pred[1]) then
-      local handler = directive_handlers[pred[1]]
+  for _, directive in pairs(directives) do
+    local handler = directive_handlers[directive[1]]
 
-      if not handler then
-        error(string.format('No handler for %s', pred[1]))
-      end
-
-      handler(captures, pattern, source, pred, metadata)
+    if not handler then
+      error(string.format('No handler for %s', directive[1]))
     end
+
+    handler(captures, pattern_i, source, directive, metadata)
   end
 
   return metadata
@@ -833,12 +871,6 @@ local function value_or_node_range(start, stop, node)
   end
 
   return start, stop
-end
-
---- @param match TSQueryMatch
---- @return integer
-local function match_id_hash(_, match)
-  return (match:info())
 end
 
 --- Iterates over all captures from all matches in {node}.
@@ -897,7 +929,7 @@ function Query:iter_captures(node, source, start, stop)
       return
     end
 
-    local match_id, pattern = match:info()
+    local match_id, pattern_i = match:info()
 
     --- @type vim.treesitter.query.TSMetadata
     local metadata
@@ -906,11 +938,14 @@ function Query:iter_captures(node, source, start, stop)
     end
 
     if not metadata then
-      local preds = self.info.patterns[pattern]
-      if preds then
+      metadata = {}
+
+      local processed_pattern = self._processed_patterns[pattern_i]
+      if processed_pattern then
         local captures = match:captures()
 
-        if not self:match_preds(preds, pattern, captures, source) then
+        local predicates = processed_pattern.predicates
+        if not self:_match_predicates(predicates, pattern_i, captures, source) then
           cursor:remove_match(match_id)
           if end_line and captured_node:range() > end_line then
             return nil, captured_node, nil, nil
@@ -918,9 +953,8 @@ function Query:iter_captures(node, source, start, stop)
           return iter(end_line) -- tail call: try next match
         end
 
-        metadata = self:apply_directives(preds, pattern, captures, source)
-      else
-        metadata = {}
+        local directives = processed_pattern.directives
+        metadata = self:_apply_directives(directives, pattern_i, captures, source)
       end
 
       highest_cached_match_id = math.max(highest_cached_match_id, match_id)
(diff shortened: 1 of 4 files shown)


runtime/lua/vim/treesitter/query.lua
```diff
@@ -4,6 +4,56 @@ local memoize = vim.func._memoize
...
+---@field [1] string predicate name
+---@field [2] boolean should match
+---@field [3] Pattern
+
+---@alias ProcessedDirective Pattern
+
+---@private
+---@param patterns table<integer, Pattern[]>
+---@return table<integer, { preds: ProcessedPredicate[], directives: ProcessedDirective[] }>
+local function process_patterns(patterns)
```
justinmk, line 24: is there a brief docstring that can give a hint about what this does

runtime/lua/vim/treesitter/query.lua
```diff
@@ -899,16 +915,35 @@ function Query:iter_captures(node, source, start, stop)
...
+          end
+          return iter(end_line) -- tail call: try next match
+        end
+
+        metadata = self:apply_directives(patterns.directives, pattern, captures, source)
+      else
+        metadata = {}
+      end
+
+      highest_cached_match_id = math.max(highest_cached_match_id, match_id)
```
ribru17, line 944: Is there a way to make this work without keeping track of the highest seen match id? Maybe by just checking if `match_cache[match_id]` exists? Just wondering

ribru17, line 944: Is there a way to make this work without keeping track of the highest seen match id? Maybe by just checking if `match_cache[match_id]` exists? Just wondering

runtime/lua/vim/treesitter/query.lua
```diff
@@ -972,16 +1007,21 @@ function Query:iter_matches(node, source, start, stop, opts)
...
+    --- @type vim.treesitter.query.TSMetadata
+    local metadata
+    if patterns then
+      if not self:match_preds(patterns.preds, pattern, captures, source) then
+        cursor:remove_match(match_id)
+        return iter() -- tail call: try next match
+      end
+      metadata = self:apply_directives(patterns.directives, pattern, captures, source)
+    else
+      metadata = {}
```
ribru17, line 1022: Maybe
```suggestion
    local metadata = {}
    if patterns then
      if not self:match_preds(patterns.preds, pattern, captures, source) then
        cursor:remove_match(match_id)
        return iter() -- tail call: try next match
      end
      metadata = self:apply_directives(patterns.directives, pattern, captures, source)
```

runtime/lua/vim/treesitter/query.lua
```diff
@@ -899,16 +915,35 @@ function Query:iter_captures(node, source, start, stop)
...
+    --- @type vim.treesitter.query.TSMetadata
+    local metadata
+    if match_id <= highest_cached_match_id then
+      metadata = match_cache[match_id]
     end
 
-    local metadata = apply_directives(self, match, source)
+    if not metadata then
+      local patterns = self.processed_patterns[pattern]
+      if patterns then
```
ribru17, line 928: Nit: maybe it would be clearer to call this `processed_pattern` instead of `patterns`? Same with below

runtime/lua/vim/treesitter/query.lua
```diff
@@ -740,84 +793,52 @@ function M.list_predicates()
...
-      local pred_name = pred[1]
-      if pred_name:match('^not%-') then
-        pred_name = pred_name:sub(5)
-        is_not = true
-      end
 
-      local handler = predicate_handlers[pred_name]
+    local processed_name = pred[1]
+    local should_match = pred[2]
+    local pattern = pred[3]
```
ribru17, line 808: Maybe can just change this to
```suggestion
    local pred_name = pred[3][1]
```
and then use no handler for `pred_name` below? Since this is the only place `pattern` is used currently. Then maybe we can also change the parameter name from `pattern_i` to `pattern` which seems clearer to me

ribru17 (approved):

runtime/lua/vim/treesitter/query.lua
vanaigr: Yes, it can be removed. Work correctly either way. 

I added this check since the cache might become a hash map at some point, and indexing it would become more expensive.

 Most of the time, captures returned by `next_capture()` are new and have match id higher than anything in the cache, so this check is cheaper.

It's about 1ms IIRC

runtime/lua/vim/treesitter/query.lua
vanaigr: also passed to the handler as the last argument.

Regarding the rename, `pattern_i` is clearer since it's not the actual pattern, but its index.

ribru17: Ah thanks, that makes sense. nevermind then

vanaigr:
Currently, the term "predicate" is used to refer to both tree-sitter predicates and a specific type of predicate that tests a condition.

Updated query.lua to refer to the second meaning as "guard". Let me know if you think the name should be different. I can also remove this change if it's unnecessary.

vanaigr:
What happened to macos clang unittest?

<details>

<summary>part of the log</summary>

```console
CMake Error at /Users/runner/work/neovim/neovim/cmake/RunTests.cmake:103 (message):
  unit tests failed with error: 1


-- Tests exited non-zero: 1
-- Output to stderr:
In file included from tmp_pseudoheader.h:6859:
In file included from ././src/nvim/eval/typval.h:10:
/Users/runner/work/neovim/neovim/src/nvim/gettext_defs.h:4:11: fatal error: 'libintl.h' file not found
    4 | # include <libintl.h>  // IWYU pragma: export
      |           ^~~~~~~~~~~
1 error generated.
In file included from tmp_pseudoheader.h:6859:
In file included from ././src/nvim/eval/typval.h:10:
/Users/runner/work/neovim/neovim/src/nvim/gettext_defs.h:4:11: fatal error: 'libintl.h' file not found
    4 | # include <libintl.h>  // IWYU pragma: export
      |           ^~~~~~~~~~~
1 error generated.
tmp_pseudoheader.h:3880:10: fatal error: './src/nvim/api/private/t.h' file not found
 3880 | #include "./src/nvim/api/private/t.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
tmp_pseudoheader.h:3880:10: fatal error: './src/nvim/api/private/t.h' file not found
 3880 | #include "./src/nvim/api/private/t.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
In file included from tmp_pseudoheader.h:4:
In file included from ././src/nvim/buffer.h:8:
/Users/runner/work/neovim/neovim/src/nvim/gettext_defs.h:4:11: fatal error: 'libintl.h' file not found
    4 | # include <libintl.h>  // IWYU pragma: export
      |           ^~~~~~~~~~~
1 error generated.
In file included from tmp_pseudoheader.h:4:
In file included from ././src/nvim/buffer.h:8:
/Users/runner/work/neovim/neovim/src/nvim/gettext_defs.h:4:11: fatal error: 'libintl.h' file not found
    4 | # include <libintl.h>  // IWYU pragma: export
      |           ^~~~~~~~~~~
1 error generated.
E94: No matching buffer for _test_
E94: No matching buffer for _test_
E94: No matching buffer for _test_E94: No matching buffer for testIn file included from tmp_pseudoheader.h:1950:
In file included from ././src/nvim/eval/typval.h:10:
/Users/runner/work/neovim/neovim/src/nvim/gettext_defs.h:4:11: fatal error: 'libintl.h' file not found
    4 | # include <libintl.h>  // IWYU pragma: export
      |           ^~~~~~~~~~~
1 error generated.
In file included from tmp_pseudoheader.h:1950:
In file included from ././src/nvim/eval/typval.h:10:
/Users/runner/work/neovim/neovim/src/nvim/gettext_defs.h:4:11: fatal error: 'libintl.h' file not found
    4 | # include <libintl.h>  // IWYU pragma: export
      |           ^~~~~~~~~~~
1 error generated.
E474: No container to close: ]
E474: Closing list with curly bracket: }
E474: Closing dictionary with square bracket: ]
E474: Trailing comma: ]
E474: Expected value after colon: }
E474: Expected value: }
E474: Comma not inside container: ,
E474: Duplicate comma: ,1]
E474: Comma after colon: ,}
E474: Using comma in place of colon: ,}
E474: Leading comma: ,}
E474: Leading comma: ,]
E474: Colon not inside container: :
E474: Using colon not in dictionary: :]
E474: Unexpected colon: :}
E474: Duplicate colon: :1}
E474: Expected null: n
E474: Expected true: t
E474: Expected false: f
E474: Unfinished escape sequence: "\
E474: Unfinished unicode escape sequence: "\u"
E474: Expected four hex digits after \u: \uXXXX"
E474: Unknown escape sequence: \?"
E474: ASCII control characters cannot be present inside string: ^I"
E474: Only UTF-8 strings allowed: <c2>"
E474: Only UTF-8 code points up to U+10FFFF are allowed to appear unescaped: ������"
E474: Expected string end: "
E474: Missing number after minus sign: -
E474: Missing number after decimal dot: -1.
E474: Missing exponent: -1.0e
E474: Unidentified byte: ?
E474: Trailing characters: ?
```

</details>

runtime/lua/vim/treesitter/query.lua
```diff
@@ -737,87 +790,50 @@ end
...
-  local preds = self.info.patterns[pattern]
-
-  if not preds then
-    return true
-  end
-
-  local captures = match:captures()
-
-  for _, pred in pairs(preds) do
-    -- Here we only want to return if a predicate DOES NOT match, and
```
vanaigr, line 765: Also removed this comment, as unknown predicates throw an error since https://github.com/neovim/neovim/pull/12826

clason:
> What happened to macos clang unittest?

Nothing; it's always been flaky.

runtime/lua/vim/treesitter/query.lua
vanaigr: Range selection seems to be broken. This is the second time I selected one range and it displays as a different range.

runtime/lua/vim/treesitter/query.lua
```diff
@@ -4,14 +4,69 @@ local memoize = vim.func._memoize
...
+  return string.sub(name, -1) == '!'
+end
+
+---A predicate is a list of (integer|string); integer represents `capture_id`, and
+---string represents (literal) arguments to predicate. A predicate is either a guard
+---or a directive. See |treesitter-predicates| and |treesitter-directives| for more details.
+---@alias Predicate (integer|string)[]
+
+---@nodoc
+---@class ProcessedGuard
```
lewis6991, line 17:
```suggestion
---@class vim.treesitter.query.ProcessedGuard
```

runtime/lua/vim/treesitter/query.lua
```diff
@@ -4,14 +4,69 @@ local memoize = vim.func._memoize
...
+  for k, pattern in pairs(patterns) do
+    ---@type ProcessedGuard[]
+    local guards = {}
+    ---@type ProcessedDirective[]
+    local directives = {}
+
+    for _, pred in ipairs(pattern) do
+      -- Note: tree-sitter strips the leading # from predicates for us.
+      local pred_name = pred[1]
+      ---@cast pred_name string
```
lewis6991, line 41: How do we know this is safe? Is the type definition correct?

justinmk:
> Currently, the term "predicate" is used to refer to both tree-sitter predicates and a specific type of predicate that tests a condition. Updated query.lua to refer to the second meaning as "guard".

We use `cond` (condition) in various places.

runtime/lua/vim/treesitter/query.lua
```diff
@@ -737,87 +790,50 @@ end
...
-
-      local handler = predicate_handlers[pred_name]
-
-      if not handler then
-        error(string.format('No handler for %s', pred[1]))
-        return false
-      end
-
-      local pred_matches = handler(captures, pattern, source, pred)
+function Query:match_guards(pattern_i, guards, captures, source)
```
justinmk, line 801: Even if `@private` , still need `_` prefix, especially if it's on a public interface.
```suggestion
function Query:_match_guards(pattern_i, guards, captures, source)
```

runtime/doc/treesitter.txt
```diff
@@ -1285,7 +1285,7 @@ lint({buf}, {opts})                              *vim.treesitter.query.lint()*
     Use |treesitter-parsers| in runtimepath to check the query file in {buf}
     for errors:
     • verify that used nodes are valid identifiers in the grammar.
-    • verify that predicates and directives are valid.
+    • verify that guards and directives are valid.
```
clason, line 1288: No, I think that refers (or should refer) specifically to `:h treesitter-predicates`, in parallel to `treesitter-directives`. I don't think "guards" is helpful here.

clason:
> Currently, the term "predicate" is used to refer to both tree-sitter predicates and a specific type of predicate that tests a condition.

Please stick to the industry-standard "predicate"  here; using our own terminology without need just introduces confusion.

runtime/lua/vim/treesitter/query.lua
vanaigr: This is what the original code did. And it also expected string.

https://github.com/neovim/neovim/blob/e1c2179dd93ed2cd787b1cd016606b1901a1acfe/runtime/lua/vim/treesitter/query.lua#L772-L777

vanaigr:
> _All_ predicates test a condition.

Tree-sitter refers to `#set!` directive as a predicate on its website: 
https://tree-sitter.github.io/tree-sitter/syntax-highlighting
image

And the term "directive" doesn't appear in tree-sitter repo.

clason:
Yeah, that's unfortunate, but the docs are... in need of improvement. Still, our usage has been consistent for years, so we shouldn't change it now (without _very_ good reason).

runtime/lua/vim/treesitter/query.lua
vanaigr: renamed `apply_directives()` as well.

runtime/lua/vim/treesitter/query.lua
```diff
@@ -4,6 +4,59 @@ local memoize = vim.func._memoize
 
 local M = {}
 
+local function is_directive(name)
+  return string.sub(name, -1) == '!'
+end
+
+---@nodoc
+---@class vim.treesitter.query.ProcessedPred
```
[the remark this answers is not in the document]
vanaigr: This is a problem because we need to refer to a neovim "predicate" and tree-sitter "predicate" in the same place, and I can't rename either. What should the names be then? Here's an example of this:

https://github.com/neovim/neovim/blob/6c69b0ba7976e8767c0f2d07ac4fb8a15eea4625/runtime/lua/vim/treesitter/query.lua#L805-L808

805 refers to this processed neovim predicate, while 808 refers to a treesitter predicate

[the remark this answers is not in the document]
clason: We are indeed not 100% consistent (esp. in metadata and optional fields), but this is out of scope here. (We _do_ have `add_directive()` though!)

So here's my suggested convention for now:
1. Keep the distinction between directives (`#foo!`) and predicates (`#bar?`) wherever it exists.
2. If there is a need for a term covering both, use "predicate". (I know that's not ideal, but it's not a regression, and that is all I care about in the context of this PR.)
3. In code, `pred` is fine as a shorthand for both -- only exposed API and docs are critical.

(Treesitter docs are still a bit of a mess and require a general cleanup once we're happier with the general shape of things and have a clearer view of what we _want_ to document, and how.)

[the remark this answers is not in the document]
vanaigr: > (We do have add_directive() though!).

I did a poor job explaining what I was referring to. I meant that docs for `add_predicate()` refer to neovim predicates and docs for `directive_handlers` refer to treesitter predicates. Updated my comment.

As for the convention, i think the PR adheres to all three points.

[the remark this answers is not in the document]
clason: Yep, just noting down for reference; no objections to your PR.

runtime/lua/vim/treesitter/query.lua
```diff
@@ -7,6 +7,59 @@ local memoize = vim.func._memoize
...
+---@field [1] string predicate name
+---@field [2] boolean should match
+---@field [3] (integer|string)[] the original predicate
+
+---@alias vim.treesitter.query.ProcessedDirective (integer|string)[]
+
+---@nodoc
+---@class vim.treesitter.query.ProcessedPattern {
+---@field predicates vim.treesitter.query.ProcessedPredicate[]
+---@field directives vim.treesitter.query.ProcessedDirective[]
```
justinmk, line 25: is "resolved" the right term here?
```suggestion
---@field predicates vim.treesitter.query.ResolvedPredicate[]
---@field directives vim.treesitter.query.ResolvedDirective[]
```

runtime/lua/vim/treesitter/query.lua
```diff
@@ -33,6 +87,7 @@ function Query.new(lang, ts_query)
     patterns = query_info.patterns,
   }
   self.captures = self.info.captures
+  self._processed_patterns = process_patterns(self.info.patterns)
```
justinmk, line 90: Here it looks like a new cache is added. The PR description mentions:

> Currently iter_captures() caches predicate and directives results separately

can you point out where the old caches were removed?

justinmk (approved):
Looks mostly like a lateral move in terms of complexity 👍

runtime/lua/vim/treesitter/query.lua
vanaigr: https://github.com/neovim/neovim/blob/86770108e2c6e08c2b8b95f1611923ba99b854dd/runtime/lua/vim/treesitter/query.lua#L905-L906

I removed this memoization and replaced it with 1 table.

The new field is part of:
> Also split predicates and directives into 2 lists.

PR #31627, opened 2024-12-18
fix(messages): better formatting for :highlight with ext_messages

### LINKED ISSUE #31624: messages: ui_attach: sometimes whitespaces are not preserved in chunks
### Problem

Sometimes message chunks passed to `vim.ui_attach` callback have lost some whitespace characters compared to the built-in UI appearance. 

Image

### Steps to reproduce

`init.lua`:
```lua
local api = vim.api

local function log(msg)
  vim.cmd.normal 'G'
end

local function detach()
  vim.ui_detach(ns)
end

local redirect = true

local function attach()
  if not redirect then
    return
  end
  vim.ui_attach(ns, {ext_messages = true, ext_cmdline = false}, function (event, _, content, _)
    if event ~= 'msg_show' then
      return
    end

    vim.schedule(function ()
      log(vim.inspect(content))
    end)
  end)
end

  callback = attach
})

  redirect = not redirect
  if not redirect then
    detach()
  end
end, {nargs = 0})
```
- `nvim --clean -u init.lua`
- enter `:filter Search hi` command
- enter `:MsgRedirToggle`
- enter `:filter Search hi` command again and compare results, especially look at the chunk contents before `xxx`. I get

Image

Only the first line has any space before `xxx`.

### Expected behavior

Whitespace characters are preserved in ui_attach callback. 

### Nvim version (nvim -v)

v0.11.0-dev-1380+g7121983c45

no

### Operating system/version

Debian Sid

### Terminal name/version

alacritty

alacritty

### Installation

from repo

luukvbaal:
Also avoid going down message callstack with empty message, and remove expected grid for some tests where it did not change, and we are just testing for expected messages.

Fix #31624

### DIFF
--- a/src/nvim/highlight_group.c
+++ b/src/nvim/highlight_group.c
@@ -1887,8 +1887,7 @@ bool syn_list_header(const bool did_header, const int outlen, const int id, bool
     if (got_int) {
       return true;
     }
-    msg_outtrans(hl_table[id - 1].sg_name, 0, false);
-    name_col = msg_col;
+    msg_col = name_col = msg_outtrans(hl_table[id - 1].sg_name, 0, false);
     endcol = 15;
   } else if ((ui_has(kUIMessages) || msg_silent) && !force_newline) {
     msg_putchar(' ');
--- a/src/nvim/message.c
+++ b/src/nvim/message.c
@@ -2129,8 +2129,8 @@ void msg_puts_len(const char *const str, const ptrdiff_t len, int hl_id, bool hi
   // If redirection is on, also write to the redirection file.
   redir_write(str, len);
 
-  // Don't print anything when using ":silent cmd".
-  if (msg_silent != 0) {
+  // Don't print anything when using ":silent cmd" or empty message.
+  if (msg_silent != 0 || *str == NUL) {
     return;
   }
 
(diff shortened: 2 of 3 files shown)


src/nvim/message.c
```diff
@@ -2129,8 +2129,8 @@ void msg_puts_len(const char *const str, const ptrdiff_t len, int hl_id, bool hi
   // If redirection is on, also write to the redirection file.
   redir_write(str, len);
 
-  // Don't print anything when using ":silent cmd".
-  if (msg_silent != 0) {
+  // Don't print anything when using ":silent cmd" or empty message.
+  if (msg_silent != 0 || *str == NUL) {
```
luukvbaal, line 2133: Without this the change in `msg_puts_display()` resulted in strlen - 1 overflow; there is no need to try to put an empty message onto the screen, or emit it.

src/nvim/highlight_group.c
```diff
@@ -1887,8 +1887,7 @@ bool syn_list_header(const bool did_header, const int outlen, const int id, bool
     if (got_int) {
       return true;
     }
-    msg_outtrans(hl_table[id - 1].sg_name, 0, false);
-    name_col = msg_col;
+    msg_col = name_col = msg_outtrans(hl_table[id - 1].sg_name, 0, false);
```
justinmk, line 1890: nit: I think we discourage this

justinmk (approved):

PR #31637, opened 2024-12-19
refactor: iwyu

justinmk:
Result of `make iwyu` (after some "fixups").


src/nvim/tui/termkey/termkey.c
```diff
@@ -13,9 +13,7 @@
 #include "nvim/tui/termkey/termkey_defs.h"
 
 #ifndef _WIN32
-# include <poll.h>
-# include <strings.h>
-# include <unistd.h>
+# include <termios.h>
```
justinmk, line 16: 👀

src/nvim/context.h
```diff
@@ -1,7 +1,5 @@
 #pragma once
 
-#include <stddef.h>
```
zeertzjq, line 3: This is needed by the generated header

src/nvim/api/private/helpers.h
```diff
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <stdbool.h>
-#include <stddef.h>
```
zeertzjq, line 4: ditto

src/nvim/cmdhist.h
```diff
@@ -1,7 +1,8 @@
 #pragma once
 
+#include <stddef.h>
+
 #include "nvim/cmdexpand_defs.h"  // IWYU pragma: keep
-#include "nvim/eval/typval_defs.h"
```
zeertzjq, line 4: ditto

src/nvim/option.h
```diff
@@ -1,13 +1,10 @@
 #pragma once
 
-#include <stdbool.h>
-#include <stdint.h>
 #include <stdio.h>  // IWYU pragma: keep
 
 #include "nvim/api/private/defs.h"  // IWYU pragma: keep
 #include "nvim/api/private/helpers.h"
 #include "nvim/cmdexpand_defs.h"  // IWYU pragma: keep
-#include "nvim/eval/typval_defs.h"
```
zeertzjq, line 10: ditto

src/nvim/context.h
justinmk: Thanks :) If IWYU doesn't see that, and the build passes... hmm...

PR #31638, opened 2024-12-19
Fixed MAX_TIMEOUT value for 32-bit systems

### LINKED ISSUE #31633: Invalid default timeout value for 32-bit systems
### Problem

`MAX_TIMEOUT` is defined in `share/nvim/runtime/lua/vim/_system.lua` on line 82 as `local MAX_TIMEOUT = 2 ^ 31`.

On 32-bit systems this is bigger than the max signed value which is `2 ^ 31 - 1` and as a result when calling wait without an argument (like `vim.system(cmd):wait()`) this default is used, but because it overflows it is interpreted as a negative value (`-1`) and I get a traceback:

```
E5108: Error executing lua /home/dan/.neovim/share/nvim/runtime/lua/vim/_system.lua:89: timeout must be >= 0
stack traceback:
        [C]: in function 'wait'
        /home/dan/.neovim/share/nvim/runtime/lua/vim/_system.lua:89: in function 'wait'
```

### Steps to reproduce

Type 

```:lua vim.system({"ls"}):wait()```

on a 32-bit system.

### Expected behavior

I would expect no traceback. The way to fix this is to define `MAX_TIMEOUT` as `2 ^ 31 - 1`.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1383+g02bc40c194

probably not. they don't have lua

### Operating system/version

Linux Debian unstable

### Terminal name/version

alacritty 0.14.0-dev (8dfd2e56)

alacritty

### Installation

built from repo

### PR DESCRIPTION
**Pull request by danpascu**
The maximum signed value on 32-bit systems is 2 ^ 31 - 1. When using 2 ^ 31 for the default timeout, the value would overflow on such systems resulting in a negative value, which caused a stack trace when calling wait() without a timeout. Fixes #31633

### DIFF
--- a/runtime/lua/vim/_system.lua
+++ b/runtime/lua/vim/_system.lua
@@ -79,6 +79,7 @@ function SystemObj:_timeout(signal)
   self:kill(signal or SIG.TERM)
 end
 
+-- Use max 32-bit signed int value to avoid overflow on 32-bit systems. #31633
 local MAX_TIMEOUT = 2 ^ 31 - 1
 
 --- @param timeout? integer
--- a/runtime/lua/vim/_system.lua
+++ b/runtime/lua/vim/_system.lua
@@ -79,7 +79,7 @@ function SystemObj:_timeout(signal)
   self:kill(signal or SIG.TERM)
 end
 
-local MAX_TIMEOUT = 2 ^ 31
+local MAX_TIMEOUT = 2 ^ 31 - 1
 
 --- @param timeout? integer
 --- @return vim.SystemCompleted

### CODE REVIEW

**Review by lewis6991 — APPROVED** [review]
A comment would be valuable.

**Comment by danpascu** [comment]
You mean in the code where MAX_TIMEOUT is defined?

Something like this:

```lua

-- #31633: Use max 32-bit signed int value to avoid overflow on 32-bit systems
local MAX_TIMEOUT = 2 ^ 31 - 1
```

danpascu:
Also will you add it when you merge, or do you want me to make a new commit adding the comment?

runtime/lua/vim/_system.lua
```diff
@@ -79,7 +79,7 @@ function SystemObj:_timeout(signal)
   self:kill(signal or SIG.TERM)
 end
 
-local MAX_TIMEOUT = 2 ^ 31
+local MAX_TIMEOUT = 2 ^ 31 - 1
```
justinmk, line 82:
```suggestion
-- Use max 32-bit signed int value to avoid overflow on 32-bit systems. #31633
local MAX_TIMEOUT = 2 ^ 31 - 1
```

danpascu: Should I use a particular message for the commit? I see you have conventions, but while I could figure out that the type of this commit is `fix`, I have no idea what the scope is.

Or is it possible that you update the commit message during the merge process to conform to the convention?

justinmk: will fix it during merge

PR #31646, opened 2024-12-20
fix(messages): no message kind for completion menu messages

przepompownia:
Problem: no way to replace older insert mode completion messages by newer.

Solution: add kind for them.

### DIFF
--- a/runtime/doc/ui.txt
+++ b/runtime/doc/ui.txt
@@ -796,6 +796,7 @@ must handle.
 		"echo"		|:echo| message
 		"echomsg"	|:echomsg| message
 		"echoerr"	|:echoerr| message
+		"completion"    |ins-completion-menu| message
 		"list_cmd"	List output for various commands (|:ls|, |:set|, …)
 		"lua_error"	Error in |:lua| code
 		"lua_print"	|print()| from |:lua| code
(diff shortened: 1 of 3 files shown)


przepompownia:
- Please verify the place of setting the new kind.

- For some reason the test output on my local machine looks like not working `feed('ccli<C-n>')`. Feel free to fix it.

runtime/doc/ui.txt
```diff
@@ -796,6 +796,7 @@ must handle.
 		"echo"		|:echo| message
 		"echomsg"	|:echomsg| message
 		"echoerr"	|:echoerr| message
+		"insert_expand" |insert_expand| message
```
justinmk, line 799: `:help insert_expand` seems to be about CTRL-X. Should we name this "completion" and use it for any other potential completion-related messages (e.g. for ctrl-n)?
```suggestion
		"completion" |ins-completion| message
```

luukvbaal: Ye seems more appropriate to me, if not `"completion_menu"`. The `'shortmess'` flag that guards these messages links to `|ins-completion-menu|`, and we only emit these as actual messages with `:set noshowmode`, both of which seem good to mention in this help text.

luukvbaal:
> For some reason the test output on my local machine looks like not working `feed('ccli<C-n>')`. Feel free to fix it.

These messages are only emitted with `:set noshowmode`. Also, when adding a step to an existing test, try to add it to the end of the `it()` block so as to not affect pre-existing steps.

justinmk (approved):

przepompownia:
> > For some reason the test output on my local machine looks like not working `feed('ccli<C-n>')`. Feel free to fix it.
> 
> These messages are only emitted with `:set noshowmode`. Also, when adding a step to an existing test, try to add it to the end of the `it()` block so as to not affect pre-existing steps.

OK, that's why the text appeared in `showmode` in the test and `messages` still contained the previous result. 

I started from appending the new subcase to the end of `it`, but then, on failure, moved it up and observed that the message is still "old"… Thanks for the fix before I looked here again.

PR #31650, opened 2024-12-20
refactor(eval): move funcs to deprecated.c

justinmk:


src/nvim/eval/deprecated.c
```diff
@@ -15,6 +18,117 @@
 # include "eval/deprecated.c.generated.h"
 #endif
 
+/// "rpcstart()" function (DEPRECATED)
+void f_rpcstart(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
```
justinmk, line 22: Disadvantage: can't declare these as `static`. 

I tried reworking this file as `deprecated.c.h` and including it in `src/nvim/eval/funcs.c`, but that failed because `build/src/nvim/auto/funcs.generated.h` failed to find the declarations.

PR #31660, opened 2024-12-20
fix stable branch mention in BUILD.md

mwcz:
BUILD.md still mentions the `stable` branch which does not exist.  This changes the recommended stable branch to `release-0.10`, though that means it needs to be bumped with each new release.


clason:
`stable` is a tag, not a branch, and that is intentional.

PR #31661, opened 2024-12-20
feat(ui): additional arguments for cmdline_show/hide events

### LINKED ISSUE #31643: UI messages: specify whether `msg_show` event is saved in :messages history
### Problem

If I'm not wrong, in `vim.ui_attach` (with `ext_messages = true`) callback there is no way to determine if the message is added to the history or not. 

So far I felt like I had to decide this on my own in each case but I found `msg_hist_off` global that could be passed the callback. 

https://github.com/neovim/neovim/blob/8ef41f590224dfeea2e51d9fec150e363fd72ee0/src/nvim/insexpand.c#L4601-L4611

### Expected behavior

`msg_hist_off` (or maybe its negation) is available from `vim.ui_attach` callback.

luukvbaal:
**feat(ui): additional arguments for cmdline_show/hide events**

Problem:  Unable to tell what highlight the prompt part of a
          cmdline_show event should have, and whether cmdline_hide was
          emitted after aborting.
Solution: Add additional arguments hl_id to cmdline_show, and abort to
          cmdline_hide.

**feat(ui): specify whether msg_show event is added to history**

Pass along whether message in msg_show event is added to the internal
:messages history.

Fix #31643.

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -86,6 +86,11 @@ EVENTS
 
 • |vim.ui_attach()| callbacks for |ui-messages| `msg_show` events are executed in
   |api-fast| context.
+• Various additions for the following UI events:
+  • `cmdline_show`: `hl_id` to highlight the prompt text.
+  • `cmdline_hide`: `abort` indicating if the cmdline was aborted.
+  • `msg_show`: new message kinds: "bufwrite", "completion", "list_cmd",
+    "lua_print", "number_prompt", "search_cmd", "undo", "wildlist".
 
 HIGHLIGHTS
 
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -86,11 +86,13 @@ EVENTS
 
 • |vim.ui_attach()| callbacks for |ui-messages| `msg_show` events are executed in
   |api-fast| context.
-• Various additions for the following UI events:
-  • `cmdline_show`: `hl_id` to highlight the prompt text.
-  • `cmdline_hide`: `abort` indicating if the cmdline was aborted.
-  • `msg_show`: new message kinds: "bufwrite", "completion", "list_cmd",
-    "lua_print", "number_prompt", "search_cmd", "undo", "wildlist".
+• New/enhanced arguments in these existing UI events:
+  • `cmdline_show`: `hl_id` argument to highlight the prompt text.
+  • `cmdline_hide`: `abort` argument indicating if the cmdline was aborted.
+  • `msg_show`:
+    • `history` argument indicating if the message was added to the history.
+    • new message kinds: "bufwrite", "completion", "list_cmd",
+      "lua_print", "number_prompt", "search_cmd", "undo", "wildlist".
 
 HIGHLIGHTS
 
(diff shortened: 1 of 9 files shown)


test/functional/ui/screen.lua
```diff
@@ -79,6 +79,7 @@ end
 --- @field win_position table<integer,table<string,integer>>
 --- @field float_pos table<integer,table>
 --- @field cmdline table<integer,table>
+--- @field cmdline_abort? boolean
```
luukvbaal, line 82: Didn't think it necessary to add this as an ext_keys, to avoid having to add this to a bunch of existing tests.

luukvbaal: Not sure I like this actually, considering doing this anyways; adding the `abort` state as a field in `self.cmdline[level]` instead.

runtime/doc/news.txt
```diff
@@ -86,6 +86,13 @@ EVENTS
 
 • |vim.ui_attach()| callbacks for |ui-messages| `msg_show` events are executed in
   |api-fast| context.
+• Various additions for the following UI events:
```
justinmk, line 89:
```suggestion
• New fields in these existing UI events:
```

runtime/doc/news.txt
```diff
@@ -86,6 +86,13 @@ EVENTS
...
+  • `cmdline_show`: `hl_id` argument to highlight the prompt text.
+  • `cmdline_hide`: `abort` argument indicating if the cmdline was aborted.
```
justinmk, line 91: nit: for consistency, this should probably be `cancel`. In `:help api`, `:help lsp`, and `:help CompleteDone`, we use "cancel" in a few places and "abort" nowhere.

"abort" can be useful when we want to indicate a "ungraceful exit" where state may be invalid, but I guess that's not the case here.
```suggestion
  • `cmdline_hide`: `cancel` indicates the cmdline was canceled.
```

luukvbaal: Fair, but I think that is trumped by the fact that this is analogous to the `v:event.abort` field for `CmdlineLeave`.

runtime/doc/news.txt
luukvbaal: What about:
"New/enhanced arguments in these existing UI events:`

src/nvim/message.c
```diff
@@ -3159,13 +3162,14 @@ void msg_ext_ui_flush(void)
   msg_ext_emit_chunk();
   if (msg_ext_chunks->size > 0) {
     Array *tofree = msg_ext_init_chunks();
-    ui_call_msg_show(cstr_as_string(msg_ext_kind), *tofree, msg_ext_overwrite);
+    ui_call_msg_show(cstr_as_string(msg_ext_kind), *tofree, msg_ext_overwrite, msg_ext_history);
```
justinmk, line 3165: is it useful to pass the index? 
```suggestion
    ui_call_msg_show(cstr_as_string(msg_ext_kind), *tofree, msg_ext_overwrite,
                     msg_ext_history ? msg_hist_len : -1);
```

justinmk (approved):

src/nvim/message.c
luukvbaal: Inclined to say no, even though I'm still not sure how useful the argument itself is. Just added it because of #31643, and because @bfredl said: "I just don't see any reason to not expose it. Like it is observable state anyway, this just gives the information at an earlier time."

But since the message history is a FIFO, there is no guarantee that the message will still be at the same index when the history is inspected, unless done immediately, at which point it will simply be the last element.

justinmk (approved):

PR #31669, opened 2024-12-21
fix(messages): bug and unwanted truncation in msg_outtrans_long

luukvbaal:
- Small typo/bug in msg_outtrans_long passing string length as "hist" argument.
- Avoid truncating message in msg_outtrans_long with ext_messages as a
  followup to #21709.
- Remove _hl from msg_keep and smsg_keep as there is no variant without hl_id.
- msg_printf_hl is removed (identical to smsg except that it sets
  `msg_scroll = true`, seemingly as a caveat to force a more prompt in
  cmdline mode). Move this logic to the only the only place this was
  used in ex_getln.c.

### DIFF
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -3142,8 +3142,9 @@ static bool color_cmdline(CmdlineInfo *colored_ccline)
 
 #define PRINT_ERRMSG(...) \
   do { \
+    msg_scroll = true; \
     msg_putchar('\n'); \
-    msg_printf_hl(HLF_E, __VA_ARGS__); \
+    smsg(HLF_E, __VA_ARGS__); \
     printed_errmsg = true; \
   } while (0)
   bool ret = true;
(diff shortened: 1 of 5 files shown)


justinmk:
>  "_id" missing from msg_putchar_hl, msg_puts_hl and msg_home_replace_hl.

"Missing" in what way? I don't see other cases in this module, and typically `_id` functions *return* an id, whereas these `_hl` functions are performing a highlight. Thus `_hl_id` actually seems confusing here.

> Remove _hl from msg_keep and smsg_keep as there is no variant without hl_id.

👍 

> msg_printf_hl is removed

👍

luukvbaal:
> "Missing" in what way? I don't see other `_id` functions in this module, and typically `_id` functions _return_ an id, whereas these `_hl` functions are performing a highlight. Thus `_hl_id` actually seems confusing here.

Oh oops, I agree that for function names the `_hl` suffix is more natural. This was prompted by coming across some comments mentioning functions ending in `_hl_id`. I thought we now had some functions that ended in `_hl_id`, and updated these to be consistent, thinking that was the direction chosen in the previous PR. I'll fix the comments instead (likely made it in as an oversight after a substitute command).

PR #31674, opened 2024-12-21
docs: split compound step into two steps in BUILD.md

jj-du-plessis:
Problem:
The instruction for installing the stable release comes after a compound instruction (cd ... && make ...), which is too late.

Solution:
Split the compound instruction into two, giving users the opportunity to first checkout the stable branch before proceeding with the make command.


dundargoc:

PR #31697, opened 2024-12-23
feat(test): support and document lua test case debugging

mfussenegger:
Similar to how there is a `GDB` environment variable to let the nvim
test instances to be run under `gdbserver` this adds a `OSV_PORT`
variable to start nvim test instances with `osv` in blocking mode to let
a debug client attach to it for debugging of `exec_lua` code blocks.

recording

(I haven't really used this yet, so not sure what other limitations or caveats exist)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -138,6 +138,163 @@ Debugging tests
   Then put `screen:snapshot_util()` anywhere in your test. See the comments in
   `test/functional/ui/screen.lua` for more info.
 
+Debugging Lua test code
+-----------------------
+
+Debugging Lua test code is a bit involved. Get your shopping list ready, you'll
+need to install and configure:
+
+1. [nvim-dap](https://github.com/mfussenegger/nvim-dap)
+2. [local-lua-debugger-vscode](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#local-lua-debugger-vscode)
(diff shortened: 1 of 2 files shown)


justinmk (approved):
Beautiful!

test/README.md
```diff
@@ -138,6 +138,163 @@ Debugging tests
...
+if os.getenv("LOCAL_LUA_DEBUGGER_VSCODE") == "1" then
+  require("lldebugger").start()
+end
+```
+
+This is a [local-lua-debugger
+limitation](https://github.com/tomblind/local-lua-debugger-vscode?tab=readme-ov-file#busted)
+
+- You cannot step into code of files which get baked into the nvim binary like
+  the `shared.lua`.
```
luukvbaal, line 295: Unless `--luamod-dev` is used? Should that be mentioned here?

mfussenegger: I wasn't aware of `--luamod-dev`. It should work but probably needs to be set for testnvim instance - could probably do that also automatically under the `OSV_PORT` env block?

luukvbaal: I have not yet tried this out, but yes that seems appropriate to me. Making sure we can step into all of our runtime when `OSV_PORT` is set sounds reasonable.

And yeah I have at times forgot about `--luamod-dev` and then found it to be missing from our manpage. Should it be added there, or was it omitted on purpose?

PR #31700, opened 2024-12-23
fix: fix broken wasmtime build

dundargoc:
Regression from 2a7d0ed6145bf3f8b139c2694563f460f829813a, which removed
header that is only needed if wasmtime support is enabled. Prevent this
from happening again by wrapping the include in a `HAVE_WASMTIME` check.


dundargoc:
> optional dependencies are the root of all evil

me, right now

PR #31701, opened 2024-12-23
fix(diagnostic): silence chistory command

yorickpeterse:
vim.diagnostic.set_list() uses chistory to restore the actively selected entry whenever necessary. This however also results in it displaying some output in the message bar, but this output isn't useful (and can even be distracting) when opening the quickfix window. This fixes this by silencing the chistory command.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -874,7 +874,7 @@ local function set_list(loclist, opts)
     if not loclist then
       -- First navigate to the diagnostics quickfix list.
       local nr = vim.fn.getqflist({ id = qf_id, nr = 0 }).nr
-      api.nvim_command(nr .. 'chistory')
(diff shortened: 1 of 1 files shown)


yorickpeterse:
The commit refers to the output that can be seen in the bottom of this screenshot:
Screenshot From 2024-12-23 04-10-27

runtime/lua/vim/diagnostic.lua
```diff
@@ -874,7 +874,7 @@ local function set_list(loclist, opts)
     if not loclist then
       -- First navigate to the diagnostics quickfix list.
       local nr = vim.fn.getqflist({ id = qf_id, nr = 0 }).nr
-      api.nvim_command(nr .. 'chistory')
+      api.nvim_command('silent ' .. nr .. 'chistory')
```
justinmk, line 877:
```suggestion
      api.nvim_command('silent %dchistory')
```

PR #31702, opened 2024-12-23
feat: add vim.fs.is_relative_to

dundargoc:
This is needed to replace the nvim-lspconfig function is_descendant that
some lspconfg configurations still use.


dundargoc:
Chose name because it's what python calls it and you can't go wrong with doing what python does.

Will add tests once there's rough consensus we want this.

runtime/lua/vim/fs.lua
```diff
@@ -679,4 +679,22 @@ function M.rm(path, opts)
   end
 end
 
+--- Returns `true` if `path` is relative to `root`
+--- @param path string?
+--- @param root string?
+--- @return boolean
+function M.is_relative_to(path, root)
+  if root == nil or path == nil then
```
justinmk, line 687: if they are both optional then we should return `true` if they are both nil.

not sure if they should be optional though. callers can opt-in to "relaxed" handling via `path or ''`.

justinmk:
> once there's rough consensus we want this.

It's a common need, and not obvious how to do it in a reliable way.

Just for reference, what is the vimscript way to do this? chatgpt suggests:

```vim
function! IsPathRelativeTo(pathA, pathB) abort
    let l:absoluteA = fnamemodify(a:pathA, ':p')
    let l:absoluteB = fnamemodify(a:pathB, ':p')

    " Ensure pathB ends with a slash for proper comparison
    if l:absoluteB[-1] !=# '/'
        let l:absoluteB .= '/'
    endif

    return l:absoluteA[:len(l:absoluteB) - 1] ==# l:absoluteB
endfunction
```

runtime/lua/vim/fs.lua
```diff
@@ -679,4 +679,22 @@ function M.rm(path, opts)
...
+    return false
+  end
+  path = vim.fs.normalize(path)
+  root = vim.fs.normalize(root)
+  for dir in vim.fs.parents(path) do
```
justinmk, line 692: did we decide not to use the  absolute path or `realpath`  approach? avoids an iterator.

runtime/lua/vim/fs.lua
dundargoc: No strong opinions on this, string-only sounds good.

runtime/lua/vim/fs.lua
dundargoc: No decision has been made as far as I'm aware. I went with this approach mostly because it seemed safer, as to avoid false positives for edge cases where `root` is `/home/user` and `path` is `/home/username`. If that's not really a concern, then the following code snippet should work I think:

```lua
function M.is_relative_to(path, root)
  vim.validate('path', path, 'string')
  vim.validate('root', root, 'string')
  return root == path or vim.startswith(vim.fs.normalize(path), vim.fs.normalize(root) .. '/')
end
```

Shall we go with this?

dundargoc: Answer: we shall not, because it fails the tests. So we could go with absolute/realpath but we'd need to ensure it passes the tests. I welcome any challenger :)

dundargoc: No decision has been made, but I'm not sure it can be done in a robust way. I tried but it didn't pass the tests.

dundargoc:
This is good to go. I'll leave it open for a while for comments. I am alright with bikeshedding the name. On lspconfig this is called `is_descendant`, but this type of "ancestor/descendent" glossary isn't common in other language path so don't think it's particularly common.

justinmk:
Naming suggetion: `isrelto`, or `relto`

test/functional/lua/fs_spec.lua
```diff
@@ -454,4 +454,74 @@ describe('vim.fs', function()
...
+  describe('is_relative_to()', function()
+    local function is_rel(path, root)
+      return exec_lua([[return vim.fs.is_relative_to(...)]], path, root)
+    end
+    it('works', function()
+      local p = 'a/b'
+      eq(true, is_rel(p, 'a/'))
+      eq(true, is_rel(p, 'a/'))
+      eq(true, is_rel(p, 'a/b'))
+      eq(false, is_rel(p, 'c'))
```
justinmk, line 467: nit: `ok` and `ok(not ...)` look more readable. Or `assert`.
```suggestion
      ok(is_rel(p, 'a/b'))
      eq(not is_rel(p, 'c'))
```

dundargoc:
Potentiall superseded by https://github.com/neovim/neovim/pull/31790

Issue #31687, opened 2024-12-22
Cursor blink behaviour is unclear


fredizzimo:
### Problem

The [guicursor](https://neovim.io/doc/user/options.html#cursor-blinking) documentation says this:
> blinkwait{N}[cursor-blinking](https://neovim.io/doc/user/options.html#cursor-blinking)
blinkon{N} blinkoff{N} blink times for cursor: blinkwait is the delay before the cursor starts blinking, blinkon is the time that the cursor is shown and blinkoff is the time that the cursor is not shown. Times are in msec. **When one of the numbers is zero, there is no blinking.**

In Neovide we interpret this as the blinking is disabled when only `blinkon` is set, but not the other two values. The TUI Neovim seems to interpret this differently, and enables blinking in that case.

This leads to confusions like this:
* https://github.com/neovide/neovide/issues/2898
* https://github.com/neovide/neovide/issues/2164

So, it would be good to get a clarification on how that should be interpreted. Preferably, UIs should not need to care about these special cases at all, and Neovim automatically sends all zeroes, or suitable values, depending on the configuration.

### Steps to reproduce

1. Start nvim or neovide
2. set guicursor=n:block-blinkon100,i:ver25-blinkon100

Observe that the option works differently depending on if TUI nvim is started or neovide is started

### Expected behavior

Neovim, always sets all three values in `mode_info_set`, if there should be blinking, otherwise all of them are set to 0.

### Nvim version (nvim -v)

NVIM v0.10.3

### Vim (not Nvim) behaves the same?

N/A

### Operating system/version

Arch Linux

### Terminal name/version

alacritty

### $TERM environment variable

alacritty

### Installation

pacman

justinmk:
> Neovim, always sets all three values in `mode_info_set`, if there should be blinking, otherwise all of them are set to 0.

👍

gpanders:
Blinking should not be disabled if blinkwait is 0. How else would a user configure a 0 wait time?

I interpreted the docs to mean “if either blinkon or blinkoff is 0, blinking is disabled”. I think this is what the text *means* to say, but it should be clarified to be unambiguous.

fredizzimo:
We can change that in Neovide, but something else still needs to be done, since setting only `blinkon` (which enables blinking in the TUI) sends a `blinkoff` with a value of 0 in the `mode_info_set` structure.

gpanders:
>setting only `blinkon` (which enables blinking in the TUI)

[...]

PR #31703, opened 2024-12-23
fix(terminal): set cursor cell percentage

### LINKED ISSUE #31685: UI protocol: mode_info_set.cell_percentage is always 0 in terminal mode
### Problem

https://github.com/neovim/neovim/pull/31562 added proper cursor shape and blink support, but unfortunately `cell_percentage` (in the `mode_info_set` event) is always set to 0, which makes the cursor invisible in GUIs like Neovide.

This was first reported here https://github.com/neovide/neovide/issues/2897

### Steps to reproduce

1. Start neovide
2. Open `:terminal` with a bash shell using the default vertical bar cursor shape
3. Observe that the cursor is invisible

### Expected behavior

Neovim should set `cell_percentage` to something reasonable, for example `25` for the vertical bar and `20` for the vertical bar, like the `guicursor` defaults.

Alternatively, it should be documented that 0 is GUI specific, and that the GUI can choose a suitable value. But in that case, it would probably make sense to set the `guicursor` default to 0 as well.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1399+gd1e00a5f6d

N/A

### Operating system/version

Arch Linux

### Terminal name/version

Neovide 0.13.3

alacritty

### Installation

build from repo

gpanders:
Fixes: https://github.com/neovim/neovim/issues/31685

cc @fredizzimo. I took your suggestion and set the percentage to 25 for vertical cursors and 20 for horizontal. This isn't configurable for now (I'll wait until enough people ask for it, because it's not clear how we even would make this configurable: the escape sequence to modify the cursor shape can't specify parameters such as the width).

Btw I noticed while testing in Neovide that the cursor doesn't blink, even when it is configured to. Is this a known issue in Neovide or is this something else broken in core?

### DIFF
--- a/src/nvim/terminal.c
+++ b/src/nvim/terminal.c
@@ -1961,9 +1961,11 @@ static void refresh_cursor(Terminal *term)
     break;
   case VTERM_PROP_CURSORSHAPE_UNDERLINE:
     shape_table[SHAPE_IDX_TERM].shape = SHAPE_HOR;
+    shape_table[SHAPE_IDX_TERM].percentage = 20;
     break;
   case VTERM_PROP_CURSORSHAPE_BAR_LEFT:
     shape_table[SHAPE_IDX_TERM].shape = SHAPE_VER;
+    shape_table[SHAPE_IDX_TERM].percentage = 25;
     break;
   }
 
--- a/test/functional/terminal/cursor_spec.lua
+++ b/test/functional/terminal/cursor_spec.lua
@@ -152,7 +152,7 @@ describe(':terminal cursor', function()
     end)
   end)
 
-  it('can be modified by application #3681', function()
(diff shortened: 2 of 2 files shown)


justinmk (approved):

fredizzimo:
Thank you, I can confirm that it works in Neovide. We probably have to tweak the cursor movement animation at some point though, since it's moving around a lot. 

> Btw I noticed while testing in Neovide that the cursor doesn't blink, even when it is configured to. Is this a known issue in Neovide or is this something else broken in core?

Yes, that's another issue, it has been known for long, but I only reported it here yesterday https://github.com/neovim/neovim/issues/31687, basically you need to configure all three times `wait`, `on` and `off`.

PR #31708, opened 2024-12-24
docs(api): fix return type of `nvim[_buf]_get_keymap()`

### LINKED ISSUE #31691: Wrong lua type annotation for the return value of `nvim_get_keymap` and `nvim_buf_get_keymap`
### Problem

`nvim_get_keymap` and `nvim_buf_get_keymap` return a list of mapping definitions which has the following fields:

- abbr
- buffer
- desc
- expr
- lhs
- lhsraw
- lhsrawalt
- lnum
- mode
- mode_bits
- noremap
- nowait
- rhs
- script
- scriptversion
- sid
- silent

However in `runtime/lua/vim/_meta/api.lua` they are annotated to return `vim.api.keyset.keymap[]` with extra fields `callback`, `replace_keycodes`, `unique` and is missing fields `abbr`, `buffer`, `lhs`, `lhsraw`, `lhsrawalt`, `lnum`, `mode`, `mode_bits`, `rhs`, `scriptversion`, and `sid`

### Steps to reproduce

N/A

### Expected behavior

`nvim_get_keymap` and `nvim_buf_get_keymap` should have correct type annotation in `_meta/api.lua`.

### Nvim version (nvim -v)

0.10.3

N/A

### Operating system/version

Linux

### Terminal name/version

Alacritty

tmux-256color

### Installation

AUR

Bekaboo:
Shoud fix #31691

### DIFF
--- a/runtime/lua/vim/_meta/api_keysets_extra.lua
+++ b/runtime/lua/vim/_meta/api_keysets_extra.lua
@@ -173,6 +173,26 @@ error('Cannot require a meta file')
 --- @field force? true
 --- @field cterm? vim.api.keyset.hl_info.cterm
 
+--- @class vim.api.keyset.get_keymap
+--- @field abbr? 0|1
+--- @field buffer? 0|1
+--- @field callback? function
+--- @field desc? string
+--- @field expr? 0|1
+--- @field lhs? string
+--- @field lhsraw? string
+--- @field lhsrawalt? string
+--- @field lnum? integer
+--- @field mode? string
+--- @field mode_bits? integer
+--- @field noremap? 0|1
+--- @field nowait? 0|1
+--- @field rhs? string
+--- @field script? 0|1
+--- @field scriptversion? integer
+--- @field sid? integer
+--- @field silent? 0|1
+
 --- @class vim.api.keyset.get_mode
 --- @field blocking boolean
 --- @field mode string
--- a/runtime/lua/vim/_meta/api.lua
+++ b/runtime/lua/vim/_meta/api.lua
@@ -452,7 +452,7 @@ function vim.api.nvim_buf_get_extmarks(buffer, ns_id, start, end_, opts) end
 ---
 --- @param buffer integer Buffer handle, or 0 for current buffer
 --- @param mode string Mode short-name ("n", "i", "v", ...)
---- @return vim.api.keyset.keymap[] # Array of |maparg()|-like dictionaries describing mappings.
+--- @return vim.api.keyset.get_keymap[] # Array of |maparg()|-like dictionaries describing mappings.
 --- The "buffer" key holds the associated buffer handle.
 function vim.api.nvim_buf_get_keymap(buffer, mode) end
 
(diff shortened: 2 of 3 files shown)


runtime/lua/vim/_meta/api_keysets_extra.lua
```diff
@@ -173,6 +173,26 @@ error('Cannot require a meta file')
 --- @field force? true
 --- @field cterm? vim.api.keyset.hl_info.cterm
 
+--- @class vim.api.keyset.get_keymap
+--- @field abbr? 0|1
```
glepnir, line 177: does there shouldn't boolean? and in doc says is return maparg() like dict maybe we can also link this object for maparg?

zeertzjq: No, they are integers

Bekaboo: I checked the result of `:lua =vim.api.nvim_get_keymap('n')` and found out that fields that should be boolean are represented by 0/1

Bekaboo: > maybe we can also link this object for maparg?

`maparg()` can returns the rhs of a mapping if `{dict}` if falsy or not specified, e.g. `maparg('<C-l>')`. Currently it is annotated to return a string.

We can do this for `maplist()` though which gives the same result as `nvim_get_keymap`

Bekaboo:
@glepnir Is it a good idea to set `vim.api.keyset.get_keymap[]` as the return value of vimscript `maparg()` or `maplist()`? Different from `nvim_get_keymap()`, they are not API functions. The type annotations are generally more vague for the vimscript function compared to the detailed annotations for the `nvim_` API functions, and according to `vimfn.lua` there's no other vimcript function has type annotations of `vim.api.keyset.*`

glepnir:
Not sure 😮‍💨. Just saw the documentation described it as maparg() like .  Maybe that's good enough for now.

justinmk:
Obvious question is why not enhance `vim.api.keyset.keymap` instead of defining a new type? But I see that the new type has integer values whereas `vim.api.keyset.keymap` has booleans. 

Then the next question is WHY `nvim_buf_get_keymap` doesn't coerce stuff to booleans?

justinmk (approved):
Meh.

Bekaboo:
> Then the next question is WHY `nvim_buf_get_keymap` doesn't coerce stuff to booleans?

@justinmk Won't that break the API?

justinmk:
> I'd like that too, but won't that break the API?

Maybe. That's a good reminder of why "roundtripping" is important: it helps discover this kind of thing from the start. Roundtrips are also discussed in https://github.com/neovim/neovim/pull/29464 @max397574 though I need to review if it's relevant here (the "coerce to boolean" topic).

Bekaboo:
Adding `vim.keymap.get` indeed ease the urgent to change the return type of `nvim_get_keymap()` and avoid breaking the API.

PR #31631, opened 2024-12-19
feat(treesitter): async parsing

ribru17:
**Problem:** Parsing can be slow for large files, and it is a blocking operation which can be disruptive and annoying.

**Solution:** Provide a function for asynchronous parsing, which accepts a callback to be run after parsing completes.

This borrows from #22420 and from the pulsar implementation a bit. It currently works with the big linux file in my local testing, and I think the code is a bit simpler than before.

That said, there are still some stutters at times (with the gigantic file), which is disappointing. Start up time is massively improved but cursor movement stalls towards the end of parsing completion. ~If anyone can provide some insights on how to smooth over the asynchronous chunks it would be much appreciated.~

Thanks to some digging by @vanaigr, it is clear that the stutters are due to injection processing, specifically in the `iter_matches()` call in `LanguageTree:_get_injections()`. This is essentially the only highlighting bottleneck now. Deleting injection queries for the language gives no stutter whatsoever, and completely asynchronous highlighting.

PR #31710, opened 2024-12-24
fix: clamp range lines in `nvim__redraw()`

vanaigr:
### Problem

`nvim__redraw()` doesn't clamp the lines in the `range` parameter before truncating to int. The resulting range may be empty when the original range contained buffer lines and vice versa.
Originally found in: https://github.com/neovim/neovim/pull/31631#discussion_r1896414116

E.g. for a buffer with 4 lines, these are the redrawn lines:

```lua
{ 2, 2 ^ 31 } -> none (should be { 2, 3 })
{ 2, 2 ^ 32 } -> none (should be { 2, 3 })
{ 2 ^ 32 - 1, 2 } -> { 0, 1 } (should be none)
```

### Solution

Clamp `range` values before truncating to int.

### DIFF
--- a/test/functional/api/vim_spec.lua
+++ b/test/functional/api/vim_spec.lua
@@ -5385,16 +5385,11 @@ describe('API', function()
         13                                                          |
       ]],
     })
-    -- takes buffer line count from correct buffer with "win" and {0, -1} "range"
   end)
 
   it('nvim__redraw range parameter', function()
     Screen.new(10, 5)
-
-    fn.setline(1, '1')
-    fn.setline(2, '2')
-    fn.setline(3, '3')
-    fn.setline(4, '4')
+    fn.setline(1, fn.range(4))
 
     exec_lua([[
       _G.lines_list = {}
--- a/test/functional/api/vim_spec.lua
+++ b/test/functional/api/vim_spec.lua
@@ -5386,7 +5386,57 @@ describe('API', function()
       ]],
     })
     -- takes buffer line count from correct buffer with "win" and {0, -1} "range"
-    api.nvim__redraw({ win = 0, range = { 0, -1 } })
+  end)
+
+  it('nvim__redraw range parameter', function()
+    Screen.new(10, 5)
+
+    fn.setline(1, '1')
+    fn.setline(2, '2')
+    fn.setline(3, '3')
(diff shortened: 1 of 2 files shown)


test/functional/api/vim_spec.lua
```diff
@@ -5386,7 +5386,57 @@ describe('API', function()
       ]],
     })
     -- takes buffer line count from correct buffer with "win" and {0, -1} "range"
```
luukvbaal, line 5388: Should (re)move this comment.

test/functional/api/vim_spec.lua
```diff
@@ -5386,7 +5386,57 @@ describe('API', function()
...
-    api.nvim__redraw({ win = 0, range = { 0, -1 } })
+  end)
+
+  it('nvim__redraw range parameter', function()
+    Screen.new(10, 5)
+
+    fn.setline(1, '1')
+    fn.setline(2, '2')
+    fn.setline(3, '3')
+    fn.setline(4, '4')
```
luukvbaal, line 5397:
```suggestion
    fn.setline(1, fn.range(4))    
```

luukvbaal (approved):

vanaigr:
I also triggered this assert while writing a test with a row being `2 ^ 63 - 1`: 
```console
/neovim/src/mpack/conv.c:113: mpack_pack_number: Assertion `v <= 9007199254740991. && v >= -9007199254740991.' failed.
```

luukvbaal:
> I also triggered this assert while writing a test with a row being 2 ^ 63 - 1:

Interesting, libmpack treats integer types as floating point so the maximum value it can handle is `2^53 - 1`. I suppose we need to adapt the api dispatch to error for numbers that overflow `double`. (Or add `mpack_pack_integer()`, IDK, @bfredl?)

bfredl:
Maximum possible line number is MAX_LNUM currently 2^31 - 1.  We could just clamp a larger end value to that at arg parsing time

luukvbaal:
But we are not talking specifically about the end value for `range` here (not sure if you are). Any integer argument that reaches libmpack (over the RPC layer?) can hit this assert. Erroring rather than clamping seems more reasonable to me.

PR #31711, opened 2024-12-24
vim-patch:9.1.0958: filetype: supertux2 config files detected as lisp

clason:
Problem:  filetype: supertux2 config files detected as lisp
Solution: detect supertux2 config files as scheme instead
          (Wu, Zhenyu)

References:
https://github.com/SuperTux/supertux/wiki/S-Expression

supertux uses #t and #f as bool type, which is same as scheme, not
common lisp

closes: vim/vim#16287

https://github.com/vim/vim/commit/e62d93ead10b4c5818e3c0b7551f1784d24bfe33

Co-authored-by: Wu, Zhenyu <wuzhenyu@ustc.edu>


runtime/lua/vim/filetype.lua
```diff
@@ -2159,7 +2159,7 @@ local pattern = {
     ['/gitolite%-admin/conf/'] = starsetf('gitolite'),
     ['/%.i3/config$'] = 'i3config',
     ['/i3/config$'] = 'i3config',
-    ['/supertux2/config$'] = 'lisp',
+    ['/supertux2/config$'] = 'scheme',
     ['/%.mplayer/config$'] = 'mplayerconf',
```
zeertzjq, line 2163:
```suggestion
    ['/%.mplayer/config$'] = 'mplayerconf',
    ['/supertux2/config$'] = 'scheme',
```

zeertzjq (approved):

PR #31716, opened 2024-12-24
fix: alignment with hjkl directional keys

xames3:
- not much but this commit fixes the alignment with hjkl directional keys where the j & l keys were offset a little to left.


zeertzjq:
No. Search existing PRs.

PR #31719, opened 2024-12-25
fix: enhance builtin autocompletion and fix LSP snippet bug

latteyt:
- Replaced InsertCharPre with TextChangedI and TextChangedP for improved efficiency in builtin autocompletion.
- Fixed a bug where multi-line snippets from LSP caused the "Using a Blob as a String" error.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -31,9 +31,9 @@ local buf_handles = {}
 --- @nodoc
 --- @class vim.lsp.completion.Context
 local Context = {
-  cursor = nil, --- @type [integer, integer]?
+  cursor = nil,            --- @type [integer, integer]?
   last_request_time = nil, --- @type integer?
-  pending_requests = {}, --- @type function[]
+  pending_requests = {},   --- @type function[]
   isIncomplete = false,
 }
 
@@ -277,10 +277,11 @@ function M._lsp_to_complete_items(result, prefix, client_id)
   for _, item in ipairs(items) do
     if matches(item) then
       local word = get_completion_word(item)
+      word = string.gsub(word, "[\r\n\t]", " ") -- fix complete() bug on lsp-snippet
       local hl_group = ''
       if
-        item.deprecated
-        or vim.list_contains((item.tags or {}), protocol.CompletionTag.Deprecated)
+          item.deprecated
+          or vim.list_contains((item.tags or {}), protocol.CompletionTag.Deprecated)
       then
         hl_group = 'DiagnosticDeprecated'
       end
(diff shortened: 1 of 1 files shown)


mfussenegger:
> Replaced InsertCharPre with TextChangedI and TextChangedP for improved efficiency in builtin autocompletion.

Why'd they be more efficient? `TextChangedI` also doesn't trigger if the popup menu is visible, which breaks the logic to fetch more candidates if the previous result was incomplete.

> Fixed a bug where multi-line snippets from LSP caused the "Using a Blob as a String" error.

I thought the nil byte issue was fixed in the C side of the completion system. Do you have a reproduction for the error? cc: @glepnir

6Seeks:
> > Replaced InsertCharPre with TextChangedI and TextChangedP for improved efficiency in builtin autocompletion.
> 
> Why'd they be more efficient? `TextChangedI` also doesn't trigger if the popup menu is visible, which breaks the logic to fetch more candidates if the previous result was incomplete.
> 
> > Fixed a bug where multi-line snippets from LSP caused the "Using a Blob as a String" error.
> 
> I thought the nil byte issue was fixed in the C side of the completion system. Do you have a reproduction for the error? cc: @glepnir

sorry, i donot why use the `InsertCharPre` instead of  `{ "TextChangedI", "TextChangedP" }` for triggerring autocompletion. So i replace it. Anyway,  `InsertCharPre` can only work when inserting the trigger charactors, but  `{ "TextChangedI", "TextChangedP" }`  can work on more, e.g., deleting charactor or inserting an alphabet.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -31,9 +31,9 @@ local buf_handles = {}
 --- @nodoc
 --- @class vim.lsp.completion.Context
 local Context = {
-  cursor = nil, --- @type [integer, integer]?
+  cursor = nil,            --- @type [integer, integer]?
```
MariaSolOs, line 34: Could we please remove the formatting changes in this PR?

6Seeks: i will withdraw this pr. because i found too-frequent trigger autocompletion will bring more bug.

`clangd`  fails to set `isIncomplete` to `true` in time when calling lsp-snippets:
[录屏 12-26-2024 10:22:06 AM.webm](https://github.com/user-attachments/assets/5d2c96b8-2a3d-4fe8-9ca3-2304dd85fd06)

`lua_ls` will still have the bug of `Using a Blob as a String`  when inserting the lsp-snippets:

[录屏 12-26-2024 10:24:44 AM.webm](https://github.com/user-attachments/assets/61c1f7a4-316b-4f4d-9521-9af6f8e1ad8d)


it needs more considerations. Anyway, both `"TextChangedI", "TextChangedP", "InsertCharPre"` are  not very suitable for this feature (in my opinion).

6Seeks: I now use a very dirty method, directly blocking lsp-snippet to make it work normally.

mfussenegger: It would still be nice to have reproduction steps for your issue. If there's a bug we ought to fix it.
But a PR that fixes a bug shouldn't mix other features into it (re-trigger on backspace)

6Seeks: sorry for it. Above bugs do not exist in the master; maybe i need more coding and debuging for re-trigger on backspace/alphabet.

6Seeks: can you close this pr?

Issue #27119, opened 2024-01-21
architecture: eliminate "provider" pattern in favor of vim.ui


justinmk:
### Problem

The old `{runtime}/provider/*` pattern is basically unnecessary since Lua monkey-patching is actually more flexible (and allows stacking, unlike the old pattern). Thus the old "provider" pattern is an unnecessary concept that adds overhead to Nvim development and usage.

### Expected behavior

- Move all the old providers to `vim.ui` (or possibly some other module, but `vim.ui` is probably a fine place for all providers to live).
- Remove support for the old `{runtime}/provider/*` concept, and related documentation.

wookayin:
I agree with the main idea and point made here, i.e. to deprecate all the `*/provider/*` pattern. I think Justin was mainly thinking of clipboard. Existing providers at the moment are rplugins (python, node, perl, ruby) and clipboard:

- Clipboard is a member that belongs to `vim.ui` (as osc52 already does), see `vim.ui.clipboard`
- But rplugins are more like for RPC support and not about UI, so `vim.ui` wouldn't be a good choice. I propose `vim.remote`, `vim.rpc`, etc.

justinmk:
> But rplugins are more like for RPC support and not about UI, so `vim.ui` wouldn't be a good choice. I can suggest: `vim.remote`

I suggested `vim.ui` because introducing something like `vim.providers` just seems pointless because most "providers" really _are_ user-facing interfaces ("UI"). The python/ruby/node providers exist to provide `:pydo`, `:rubydo`, etc, which are directly user-facing features.

wookayin:
> OTOH, they also provide API, so it would be a stretch...

Right. we have `pyeval()` a la `luaeval()`, etc. Even `:python`, `:pydo`, etc. can be seen as for scripting. The concern of "rplugin providers" is too general and wide. If `:pydo`, etc. were about user interface then all other core features (buffers, windows, treesitter, etc.) would also concerning UIs too, after all. Some mid-level utilities and APIs such as notifications, prompts (e.g. `vim.ui.select()`), pop-up (floating windows) as a core util #25514, etc. are directly concerning user interfaces though.

I think a better categorization can focus on their remote/RPC nature (their implementation and protocol) or external Language support for scripting, rather than on their function, i.e., what they are for (remote "plugins", ultimately for better user interface).

justinmk:
Updated to mention `vim.os` instead of `vim.ui`.

PR #31730, opened 2024-12-26
feat(clipboard)!: use OSC 52 as fallback clipboard provider

gpanders:
We currently enable the OSC 52 clipboard provider by setting g:clipboard when a list of conditions are met, one of which is that $SSH_TTY must be set. We include this condition because often OSC 52 is not the best clipboard provider, so if there are "local" providers available Nvim should prefer those over OSC 52.

However, if no other providers are available, Nvim should use OSC 52 even when $SSH_TTY is not set. When a user is in an SSH session then the checks for the other clipboard providers will still (typically) fail, so OSC 52 continues to be enabled by default in SSH sessions.

This is marked as a breaking change because there are some cases where OSC 52 wasn't enabled before and is now (or vice versa). However, I think this provides a better default user experience.

:christmas_tree:

### DIFF
--- a/runtime/plugin/osc52.lua
+++ b/runtime/plugin/osc52.lua
@@ -6,7 +6,15 @@ for _, ui in ipairs(vim.api.nvim_list_uis()) do
   end
 end
 
-if not tty or vim.g.clipboard ~= nil or vim.o.clipboard ~= '' or not os.getenv('SSH_TTY') then
+-- Do not query when any of the following is true:
+--   * TUI is not attached
+--   * OSC 52 support is explicitly disabled via g:termfeatures
+--   * Using a badly behaved terminal
+if
+  not tty
+  or (vim.g.termfeatures ~= nil and vim.g.termfeatures.osc52 == false)
+  or vim.env.TERM_PROGRAM == 'Apple_Terminal'
+then
   return
 end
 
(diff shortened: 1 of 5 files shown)


runtime/plugin/osc52.lua
```diff
@@ -17,28 +17,13 @@ require('vim.termcap').query('Ms', function(cap, found, seq)
...
-      ['*'] = osc52.copy('*'),
-    },
-    paste = {
-      ['+'] = osc52.paste('+'),
-      ['*'] = osc52.paste('*'),
-    },
-  }
+  local termfeatures = vim.split(vim.g.termfeatures or '', ',', { trimempty = true })
+  termfeatures[#termfeatures + 1] = 'osc52'
+  vim.g.termfeatures = table.concat(termfeatures, ',')
```
gpanders, line 28: Using `g:termfeatures` here is a bit of a kludge so that we can communicate OSC 52 support to the clipboard.vim script.

I think a more elegant solution would be to add a new UI event for copying and pasting text. The TUI would implement this event using OSC 52, but other UIs can implement it using their own GUI toolkits. How do Neovim GUIs handle copy-paste today? Do they still shell out to external tools like the TUI does? If so then having a UI event seems like it would be a nice addition since GUIs should trivially be able to interface with the system clipboard.

cc @justinmk

justinmk:
> Doing something with `nvim_set_client_info()` (e.g. to set TUI capabilities) would be better.

Definitely sounds a bit better than `g:termfeatures`.

> How do Neovim GUIs handle copy-paste today?

For pasting they are advised to call `nvim_paste` (`:help dev-ui`). They presumably link to OS libraries so don't need to shell out.

> Another, perhaps even better, solution would be to add a new UI event for copying and pasting. 

Does it need to be a UI event or could it be a regular event (autocmd)?

gpanders:
>For pasting they are advised to call nvim_paste

What about copying?

justinmk:
> What about copying?

We don't have an answer for that yet and I agree we should. In theory, UIs should ship a "clipboard provider", but our "provider" model is clumsy. We should replace all "providers" with either Lua functions which can be overridden, or global events (which we also need to enhance so they can return a "result"). https://github.com/neovim/neovim/issues/27119

PR #31746, opened 2024-12-27
docs(eval): search viml function return lnum or 0

glepnir:
Problem: search() return line number when found otherwise 0

Soliton: add returns with integer


dundargoc:

PR #31761, opened 2024-12-28
feat(vim.fs): support ~user paths in vim.fs.abspath and vim.fs.normalize

famiu:
Problem: `vim.fs.abspath()` and `vim.fs.normalize()` expand tilde to
home directory in file paths, but don't have support for `~user` paths.

Solution: Add `~user` path support for the aforementioned functions.

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -505,22 +505,18 @@ local function path_resolve_dot(path)
   return (is_path_absolute and '/' or '') .. table.concat(new_path_components, '/')
 end
 
---- Expand tilde (~) character at the beginning of the path to the user's home directory.
+--- Expand ~ and ~user (POSIX) in paths. Requires path to use / as path separator.
 ---
 --- @param path string Path to expand.
 --- @param sep string|nil Path separator to use. Uses os_sep by default.
 --- @return string Expanded path.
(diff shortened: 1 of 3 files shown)


famiu:
@dundargoc @justinmk I don't know how we could possibly add tests for this.

Another question remains as to whether `~user` should work in Windows? (I don't think it should)

dundargoc:
[Related comment](https://github.com/neovim/neovim/pull/28187#discussion_r1899039945): this behavior is technically incorrect since it's possible to create a local directory named `~user`, and expanding that to the users homedir would be technically incorrect. It's possible to make the argument that it's more beneficial/practical to expand than to allow asinine (but technically allowed) paths with tilde on them.

I'm starting to wonder if we should just rely on a separate expansion function ala `expand()` for these types of expansions (haven't fully thought it through, perhaps it would be too verbose/clumsy).

justinmk:
> starting to wonder if we should just rely on a separate expansion function ala `expand()` for these types of expansions

That is what I was hinting. In https://github.com/neovim/neovim/pull/28187#discussion_r1898952208 I didn't say "we should do this", I said it is similar, and one seems to imply the other.

Later comments from various people indicate that expanding "~" is expected and avoids some awkwardness, so we can leave it at that.

PR #31782, opened 2024-12-30
feat: windows normalize path after vim.fs.joinpath

GustavEikaas:
## Description
Hello!

I have been working with neovim a while when developing this plugin [easy-dotnet.nvim](https://github.com/GustavEikaas/easy-dotnet.nvim). Given the nature of this plugin there is quite a lot of work in resolving paths. Pretty often I read a file, get a partial path and then have to concatenate two parts of a path together. For this I use `vim.fs.joinpath`. The luadoc for the function reads as follows 
```lua
--- Concatenate directories and/or file paths into a single path with normalization
--- (e.g., `"foo/"` and `"bar"` get joined to `"foo/bar"`)
```
One thing I noticed after a while is that the "normalization" it refers to only applies between the parts being concatenated. It does not impact the parts being joined together. This resulted in me having to always `vim.fs.normalize` around the paths after joining. This would ensure clean and fully compatible paths. 

I wouldnt consider `vim.fs.joinpath` to be buggy, but the word quirky comes to mind. There might be good reasons for not applying normalization to the entire path (not that I can think of any). But if there is can we update the luadoc with more examples to fully capture the behaviour? Whats also interesting to note is that `:gsub('//+', '/')` strips out duplicate `/` but pays no attention to `\`

## Examples of this behaviour in the wild 
- [#160 easy-dotnet.nvim](https://github.com/GustavEikaas/easy-dotnet.nvim/pull/160/files#diff-387a51eeb83581adeff4281e9238e7e91cdd1615c6c19d2d84a87cec5d27b151R158)
- [#149 easy-dotnet.nvim](https://github.com/GustavEikaas/easy-dotnet.nvim/pull/149/files#diff-3a973d76d2eb1ec0b2252e59200543785e714d8cb33b34e2fc3f5cf5864e6800R6)
- [#110 easy-dotnet.nvim](https://github.com/GustavEikaas/easy-dotnet.nvim/pull/110/files#diff-17054b1b101af4c8f772007883695f3583bb3c409bcfc77b162306c7a8020656R7)
- [roslyn.nvim](https://github.com/seblj/roslyn.nvim/blob/47d97e0a3f81b778409e742877b8b03fdf6c762d/lua/roslyn/sln/utils.lua#L15-L16)

## Suggestion
My suggested fix is for `vim.fs.joinpath` to normalize the entire path after the operation. I cant see any consequence in doing this.

### Before
```lua
vim.fs.joinpath("C:\\Users\\Vim\\",  "btw")
-- Output
-- "C:\Users\Vim\/btw"
```
### After
```lua
vim.fs.joinpath("C:\\Users\\Vim\\",  "btw")
-- Output
-- "C:/Users/Vim/btw"
```

### Considerations
- Why would anyone want mixed separators in their paths?
- How can anyone rely on mixed separators in their paths? (breaking change argument)
- Is there a performance cost worth considering? (no)

I couldnt find any existing issue on this
I will fix the failing tests if this functionality is wanted
Any input would be greatly appreciated!

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -105,14 +105,19 @@ function M.basename(file)
   return file:match('/$') and '' or (file:match('[^/]*$'))
 end
 
---- Concatenate directories and/or file paths into a single path with normalization
---- (e.g., `"foo/"` and `"bar"` get joined to `"foo/bar"`)
+--- Concatenates partial paths into one path. Slashes are normalized (redundant slashes are removed, and on Windows backslashes are replaced with forward-slashes)
+--- (e.g., `"foo/"` and `"/bar"` get joined to `"foo/bar"`)
+--- (windows: e.g `"a\foo\"` and `"\bar"` are joined to `"a/foo/bar"`)
 ---
 ---@since 12
 ---@param ... string
 ---@return string
 function M.joinpath(...)
-  return (table.concat({ ... }, '/'):gsub('//+', '/'))
+  local path = table.concat({ ... }, '/')
+  if iswin then
+    path = path:gsub('\\', '/')
+  end
+  return (path:gsub('//+', '/'))
 end
 
 ---@alias Iterator fun(): string?, string?
--- a/test/functional/lua/fs_spec.lua
+++ b/test/functional/lua/fs_spec.lua
@@ -323,6 +323,20 @@ describe('vim.fs', function()
       eq('foo/bar/baz', vim.fs.joinpath('foo', 'bar', 'baz'))
       eq('foo/bar/baz', vim.fs.joinpath('foo', '/bar/', '/baz'))
     end)
+    it('rewrites backslashes on Windows', function()
+      if is_os('win') then
+        eq('foo/bar/baz/zub/', vim.fs.joinpath([[foo]], [[\\bar\\\\baz]], [[zub\]]))
+      else
+        eq([[foo/\\bar\\\\baz/zub\]], vim.fs.joinpath([[foo]], [[\\bar\\\\baz]], [[zub\]]))
+      end
+    end)
+    it('strips redundant slashes', function()
+      if is_os('win') then
+        eq('foo/bar/baz/zub/', vim.fs.joinpath([[foo//]], [[\\bar\\\\baz]], [[zub\]]))
+      else
+        eq('foo/bar/baz/zub/', vim.fs.joinpath([[foo]], [[//bar////baz]], [[zub/]]))
+      end
+    end)
   end)
 
   describe('normalize()', function()
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -3062,8 +3062,11 @@ vim.fs.find({names}, {opts})                                   *vim.fs.find()*
         items
 
 vim.fs.joinpath({...})                                     *vim.fs.joinpath()*
-    Concatenate directories and/or file paths into a single path with
-    normalization (e.g., `"foo/"` and `"bar"` get joined to `"foo/bar"`)
+    Concatenates partial paths into one path. Slashes are normalized
+    (redundant slashes are removed, and on Windows backslashes are replaced
+    with forward-slashes) (e.g., `"foo/"` and `"/bar"` get joined to
+    `"foo/bar"`) (windows: e.g `"a\foo\"` and `"\bar"` are joined to
+    `"a/foo/bar"`)
 
     Attributes: ~
         Since: 0.10.0


justinmk:
`normalize` is a rather expensive routine, so if this is only about slashes, then we should perhaps use a more narrow approach.

OTOH, `joinpath()` would be more useful if it could support both kinds of slashes. It could support an overload like:

    joinpath(parts: list, opts: table)

Then one can opt-in to `\` support like this:

    joinpath({'a', 'b'}, {backslash=true})

I suppose that overload can wait until we get a request for it. 

But the main question still is whether callers of `joinpath()` should pay the cost of a full string-replace for an edge-case that the caller can opt-in to via `joinpath(normalize(...))`.

# Proposal

To avoid the performance cost on non-Windows, replace backslashes only on Windows:

```lua
if iswin then
  return x:gsub(...)
end

**Comment by GustavEikaas** [comment]
Im not sure I see how much more useful joinpath would be if it supported the overload. It requires the developer to check which platform its on. Very often the code is joining a path coming from cmd output/file parsing with a hardcoded known path like `/bin/abc` which contains separators. If vim.fs.joinpath doesnt fix all slashes then argument b would also have to contain correct separators.

e.g
```lua
local some_starter_path = os.execute("return a path")
if isWin then
  joinpath({some_starter_path, '\bin\\program'}, {backslash=true})
else
 joinpath({some_starter_path, '/bin/program'}, {backslash=false})
end
```


if performance cost is a concern then I think the below proposal is the best

```lua
if iswin then
  return x:gsub(...)
end
```


>But the main question still is whether callers of joinpath() should pay the cost of a full string-replace for an edge-case that the caller can opt-in to via joinpath(normalize(...)).

Clarity was my main issue with this. Extending the luadoc to highlight this would be a perfectly acceptable solution. Its just not clear that `Concatenate directories and/or file paths into a single path with normalization` means that the only `normalized` section of the path is between the parts that the function joined



Speaking of overloads, one possible solution is the overload you suggested but instead of backslash, we go with normalize?

joinpath({'a', 'b'}, {normalize=true})

justinmk:
the proposal I gave is what I suggest

GustavEikaas:
Great! Ill adjust the PR and the appropriate tests.

GustavEikaas:
@justinmk Did I break a test in `test / macos arm clang functionaltest (pull_request) ` or is it flaky. My change shouldn't affect non-windows in any way. I can't see why this test would break?

I inspected the test that fails but not really sure how its related to my code at all

justinmk:
`watch_spec.lua` is in bad shape, you can ignore it

runtime/lua/vim/fs.lua
```diff
@@ -112,7 +112,11 @@ end
 ---@param ... string
 ---@return string
 function M.joinpath(...)
-  return (table.concat({ ... }, '/'):gsub('//+', '/'))
+  local path = table.concat({ ... }, '/')
+  if iswin then
+    return (path:gsub('\\', '/'):gsub('//+', '/'))
```
justinmk, line 117: nit: this makes the idea more obvious, and avoids some duplicate code

```suggestion
    path = path:gsub('\\', '/')
```

runtime/lua/vim/fs.lua
```diff
@@ -112,7 +112,11 @@ end
 ---@param ... string
```
justinmk, line 112: the docstring (line 108) mentions "with normalization"; let's edit that to explicitly mention: `slashes are normalized (redundant slashes are removed, and on Windows backslashes are replaced with forward-slashes)`

justinmk:
the tests live here: https://github.com/neovim/neovim/blob/fe7be7ae95d7627fc7e0f0cb7d778803ba2424d7/test/functional/lua/fs_spec.lua#L321

please add a test for both windows and unix. e.g. (untested):

```lua
      if is_os('win') then
        eq('foo/bar/baz/zub', vim.fs.joinpath([[foo]], [[\\bar\\\\baz]], [[zub\]]))
      else
        eq([[foo/\\bar\\\\baz/zub\]], vim.fs.joinpath([[foo]], [[\\bar\\\\baz]], [[zub\]]))
      end
```

runtime/lua/vim/fs.lua
GustavEikaas: **Suggestion**
```
--- Concatenate directories and/or file paths into a single path with normalization (redundant slashes are removed)
--- Windows: backward slashes are replaced with forward slashes
--- (e.g., `"foo/"` and `"/bar"` get joined to `"foo/bar"`)
--- (windows: e.g `"a\foo\"` and `"\bar"` are joined to `"a/foo/bar"`)
```

justinmk: would prefer the wording i gave

GustavEikaas: Done

runtime/lua/vim/fs.lua
```diff
@@ -105,14 +105,19 @@ function M.basename(file)
   return file:match('/$') and '' or (file:match('[^/]*$'))
 end
 
---- Concatenate directories and/or file paths into a single path with normalization
---- (e.g., `"foo/"` and `"bar"` get joined to `"foo/bar"`)
+--- Concatenate directories and/or file paths into a single path slashes are normalized (redundant slashes are removed, and on Windows backslashes are replaced with forward-slashes)
```
justinmk, line 108:
```suggestion
--- Concatenates partial paths into one path. Slashes are normalized (redundant slashes are removed, and on Windows backslashes are replaced with forward-slashes)
```

test/functional/lua/fs_spec.lua
```diff
@@ -323,6 +323,20 @@ describe('vim.fs', function()
       eq('foo/bar/baz', vim.fs.joinpath('foo', 'bar', 'baz'))
       eq('foo/bar/baz', vim.fs.joinpath('foo', '/bar/', '/baz'))
     end)
+    it('it only rewrites backslashes on windows', function()
```
justinmk, line 326:
```suggestion
    it('rewrites backslashes on Windows', function()
```

test/functional/lua/fs_spec.lua
```diff
@@ -323,6 +323,20 @@ describe('vim.fs', function()
...
+      if is_os('win') then
+        eq('foo/bar/baz/zub/', vim.fs.joinpath([[foo]], [[\\bar\\\\baz]], [[zub\]]))
+      else
+        eq([[foo/\\bar\\\\baz/zub\]], vim.fs.joinpath([[foo]], [[\\bar\\\\baz]], [[zub\]]))
+      end
+    end)
+    it('it strips redundant slashes', function()
```
justinmk, line 333:
```suggestion
    it('strips redundant slashes', function()
```

GustavEikaas:
LGTM

PR #31787, opened 2024-12-30
feat(ui): more user friendly substitute message

### LINKED ISSUE #31667: Improve usability for search/replace "with confirm"
### Problem

when doing search/replace the user is presented with a "cryptic" help in the form of 
` (y/n/a/q/l/^E/^Y)`

### Expected behavior

I propose we elaborate a bit on the help text to cater for the inexperienced by explaining each option, for example 
` (y)es/(n)o/(a)ll/(q) stop/(l) replace and stop/(^E) scroll down/(^Y) scroll up` 

or if this is too verbose we could skip the readily understandable ones (y/n/a/q)
` (y/n/a/q/l (replace+stop)/^E (down)/^Y (up)) `

luukvbaal:
Problem:  Unknown key mappings listed in substitute confirm message.
Solution: Include hints as to what the key mappings do.

Resolve #31667

### DIFF
--- a/src/nvim/po/pt_BR.po
+++ b/src/nvim/po/pt_BR.po
@@ -4216,8 +4216,8 @@ msgstr "E146: Expressões regulares não podem ser delimitadas por letras"
 
 #: ../ex_cmds.c:3958
 #, c-format
-msgid "replace with %s (y/n/a/q/l/^E/^Y)?"
-msgstr "substituir por %s (y/n/a/q/l/^E/^Y)?"
+msgid "replace with %s? (y)es/(n)o/(a)ll/(q)uit/(l)ast/scroll up(^E)/down(^Y)"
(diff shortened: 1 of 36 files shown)


justinmk (approved):

src/nvim/po/pt_BR.po
```diff
@@ -4216,8 +4216,8 @@ msgstr "E146: Express
 
 #: ../ex_cmds.c:3958
 #, c-format
-msgid "replace with %s (y/n/a/q/l/^E/^Y)?"
-msgstr "substituir por %s (y/n/a/q/l/^E/^Y)?"
+msgid "replace with %s? (y)es/(n)o/(a)ll/(q)uit/(l)ast/scroll up(^E)/down(^Y)"
+msgstr "substituir por %s? (y)es/(n)o/(a)ll/(q)uit/(l)ast/scroll up(^E)/down(^Y)"
```
justinmk, line 4220: For non-english maybe the old way should remain? We can see if we get feedback either way though.

justinmk (approved):

PR #31789, opened 2024-12-30
docs: Drop `--beta` since this channel isn't available anymore on snap store

locnnil:
Since now the `neovim` snap `nvim` is only available on `stable` and `edge` and for `stable` there is no need to specify the channel since it's the default one that snap command line utility will try to install from.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -302,7 +302,7 @@ Neovim nightly and stable are available on the [snap store](https://snapcraft.io
 **Stable Builds**
 
 ```sh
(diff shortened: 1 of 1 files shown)


justinmk:
Should we change it to `--edge` (or whatever the flag is for that)?

locnnil:
@justinmk the `--edge` is for Nightly Builds releases, which is already listed below of my modification. 

I'm aiming to target the instruction for installing the stable releases.

dundargoc:

PR #31790, opened 2024-12-30
feat: add vim.fs.relpath

dundargoc:
This is needed to replace the nvim-lspconfig function is_descendant that
some lspconfg configurations still use.

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -741,4 +741,37 @@ function M.abspath(path)
   return M.joinpath(cwd, path)
 end
 
+--- Gets `target` path relative to `base`, or `nil` if `base` is not an ancestor.
+---
+--- Example:
+---
+--- ```lua
+--- vim.fs.relpath('/var', '/var/lib') -- 'lib'
+--- vim.fs.relpath('/var', '/usr/bin') -- nil
+--- ```
+---
+--- @param base string
+--- @param target string
+--- @param opts table? Reserved for future use
+--- @return string|nil
+function M.relpath(base, target, opts)
+  vim.validate('base', base, 'string')
+  vim.validate('target', target, 'string')
+  vim.validate('opts', opts, 'table', true)
+
+  base = vim.fs.normalize(vim.fs.abspath(base))
+  target = vim.fs.normalize(vim.fs.abspath(target))
+  if base == target then
+    return '.'
+  end
+
+  local prefix = ''
+  if iswin then
+    prefix, base = split_windows_path(base)
+  end
+  base = prefix .. base .. (base ~= '/' and '/' or '')
+
+  return vim.startswith(target, base) and target:sub(#base + 1) or nil
+end
+
 return M
--- a/test/functional/lua/fs_spec.lua
+++ b/test/functional/lua/fs_spec.lua
@@ -168,8 +168,8 @@ describe('vim.fs', function()
 
       local function run(dir, depth, skip)
         return exec_lua(function()
-          local r = {}
-          local skip_f
+          local r = {} --- @type table<string, string>
+          local skip_f --- @type function
           if skip then
             skip_f = function(n0)
               if vim.tbl_contains(skip or {}, n0) then
@@ -493,8 +493,8 @@ describe('vim.fs', function()
   end)
 
   describe('abspath()', function()
-    local cwd = is_os('win') and vim.uv.cwd():gsub('\\', '/') or vim.uv.cwd()
-    local home = is_os('win') and vim.uv.os_homedir():gsub('\\', '/') or vim.uv.os_homedir()
+    local cwd = assert(t.fix_slashes(assert(vim.uv.cwd())))
+    local home = t.fix_slashes(assert(vim.uv.os_homedir()))
 
     it('works', function()
       eq(cwd .. '/foo', vim.fs.abspath('foo'))
@@ -525,5 +525,58 @@ describe('vim.fs', function()
         eq(cwd .. '/foo', vim.fs.abspath(cwd_drive .. 'foo'))
       end)
     end
+  end)
+
+  describe('relpath()', function()
+    it('works', function()
+      local cwd = assert(t.fix_slashes(assert(vim.uv.cwd())))
+      local my_dir = vim.fs.joinpath(cwd, 'foo')
+
+      eq(nil, vim.fs.relpath('/var/lib', '/var'))
+      eq(nil, vim.fs.relpath('/var/lib', '/bin'))
+      eq(nil, vim.fs.relpath(my_dir, 'bin'))
+      eq(nil, vim.fs.relpath(my_dir, './bin'))
+      eq(nil, vim.fs.relpath(my_dir, '././'))
+      eq(nil, vim.fs.relpath(my_dir, '../'))
+      eq(nil, vim.fs.relpath('/var/lib', '/'))
+      eq(nil, vim.fs.relpath('/var/lib', '//'))
+      eq(nil, vim.fs.relpath(' ', '/var'))
+      eq(nil, vim.fs.relpath(' ', '/var'))
+      eq('.', vim.fs.relpath('/var/lib', '/var/lib'))
+      eq('lib', vim.fs.relpath('/var/', '/var/lib'))
+      eq('var/lib', vim.fs.relpath('/', '/var/lib'))
+      eq('bar/package.json', vim.fs.relpath('/foo/test', '/foo/test/bar/package.json'))
+      eq('foo/bar', vim.fs.relpath(cwd, 'foo/bar'))
+      eq('foo/bar', vim.fs.relpath('.', vim.fs.joinpath(cwd, 'foo/bar')))
+      eq('bar', vim.fs.relpath('foo', 'foo/bar'))
+      eq(nil, vim.fs.relpath('/var/lib', '/var/library/foo'))
+
+      if is_os('win') then
+        eq(nil, vim.fs.relpath('/', ' '))
+        eq(nil, vim.fs.relpath('/', 'var'))
+      else
+        local cwd_rel_root = cwd:sub(2)
+        eq(cwd_rel_root .. '/ ', vim.fs.relpath('/', ' '))
+        eq(cwd_rel_root .. '/var', vim.fs.relpath('/', 'var'))
+      end
+
+      if is_os('win') then
+        eq(nil, vim.fs.relpath('c:/aaaa/', '/aaaa/cccc'))
+        eq(nil, vim.fs.relpath('c:/aaaa/', './aaaa/cccc'))
+        eq(nil, vim.fs.relpath('c:/aaaa/', 'aaaa/cccc'))
+        eq(nil, vim.fs.relpath('c:/blah\\blah', 'd:/games'))
+        eq(nil, vim.fs.relpath('c:/games', 'd:/games'))
+        eq(nil, vim.fs.relpath('c:/games', 'd:/games/foo'))
+        eq(nil, vim.fs.relpath('c:/aaaa/bbbb', 'c:/aaaa'))
+        eq('cccc', vim.fs.relpath('c:/aaaa/', 'c:/aaaa/cccc'))
+        eq('aaaa/bbbb', vim.fs.relpath('C:/', 'c:\\aaaa\\bbbb'))
+        eq('bar/package.json', vim.fs.relpath('C:\\foo\\test', 'C:\\foo\\test\\bar\\package.json'))
+        eq('baz', vim.fs.relpath('\\\\foo\\bar', '\\\\foo\\bar\\baz'))
+        eq(nil, vim.fs.relpath('a/b/c', 'a\\b'))
+        eq('d', vim.fs.relpath('a/b/c', 'a\\b\\c\\d'))
+        eq('.', vim.fs.relpath('\\\\foo\\bar\\baz', '\\\\foo\\bar\\baz'))
+        eq(nil, vim.fs.relpath('C:\\foo\\test', 'C:\\foo\\Test\\bar\\package.json'))
+      end
+    end)
   end)
 end)
--- a/test/testutil.lua
+++ b/test/testutil.lua
@@ -22,13 +22,6 @@ local M = {
   paths = Paths,
 }
 
---- @param p string
---- @return string
-local function relpath(p)
-  p = vim.fs.normalize(p)
-  return (p:gsub('^' .. uv.cwd, ''))
-end
-
 --- @param path string
 --- @return boolean
 function M.isdir(path)
@@ -45,14 +38,15 @@ end
 --- (Only on Windows) Replaces yucky "\\" slashes with delicious "/" slashes in a string, or all
 --- string values in a table (recursively).
 ---
---- @param obj string|table
---- @return any
+--- @generic T: string|table
+--- @param obj T
+--- @return T|nil
 function M.fix_slashes(obj)
   if not M.is_os('win') then
     return obj
   end
   if type(obj) == 'string' then
-    local ret = obj:gsub('\\', '/')
+    local ret = string.gsub(obj, '\\', '/')
     return ret
   elseif type(obj) == 'table' then
     --- @cast obj table<any,any>
@@ -482,7 +476,8 @@ function M.check_cores(app, force) -- luacheck: ignore
   -- "./Xtest-tmpdir/" => "Xtest%-tmpdir"
   local local_tmpdir = nil
   if tmpdir_is_local and tmpdir then
-    local_tmpdir = vim.pesc(relpath(tmpdir):gsub('^[ ./]+', ''):gsub('%/+$', ''))
+    local_tmpdir =
+      vim.pesc(vim.fs.relpath(assert(vim.uv.cwd()), tmpdir):gsub('^[ ./]+', ''):gsub('%/+$', ''))
   end
 
   local db_cmd --- @type string
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -3148,6 +3148,23 @@ vim.fs.parents({start})                                     *vim.fs.parents()*
         (`nil`)
         (`string?`)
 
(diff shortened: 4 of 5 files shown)


dundargoc:
Alternative to https://github.com/neovim/neovim/pull/31702

echasnovski:
> I think `../apache` is technically more correct but that returning `nil` would make the implementation easier.

Returning '../apache' definitely makes `relpath()` more versatile. My personal opinion is that it comes at somewhat high cost of (significantly) improved complexity of the function itself and the inability to use it to decide whether a path is a descendant of the base (thus requiring a separate function for that). Plus I don't think I can get a good enough practical use case for supporting returning '../apache' like output.

If the requirement is dropped, then implementation is fairly straightforward and allows `vim.fs.relpath(path, dir) == nil` be used to check whether `path` is descendant of `dir`:

```lua
vim.fs.relpath = function(from, to)
  from, to = vim.fs.normalize(vim.fs.abspath(from)), vim.fs.normalize(vim.fs.abspath(to))
  to = to .. (to ~= '/' and '/' or '')
  return vim.startswith(from, to) and from:sub(#to + 1) or nil
end
```

> For reference, the relative path function for golang, zig, java, D and dart would return `../apache`.

Resorting for what is commonly used in other languages might indeed be the way to go here. You can also add [one of R's popular file system library](https://fs.r-lib.org/reference/path_math.html) to the list of "... would return `../apache`".

dundargoc:
I was somewhat hesitant to mention how other languages implements this as there's risk of cargo culting involved in these cases. I can remove it if it's not useful/detrimental to the discussion. If I'm being honest I'm don't see a super strong reason for allowing outside paths at the moment other than "it is more correct". I will lurk around in python code/issues to check why they decided to add this functionality in 3.12 when they didn't before, presumably there was a need for this.

test/testutil.lua
```diff
@@ -22,13 +22,6 @@ local M = {
   paths = Paths,
 }
 
---- @param p string
---- @return string
-local function relpath(p)
-  p = vim.fs.normalize(p)
-  return (p:gsub('^' .. uv.cwd, ''))
```
justinmk, line 29: This is an interesting impl, maybe we should try it? But make sure that `cmd` ends with a slash :P

dundargoc: I'm pretty sure I wrote and that it doesn't work. I half-assed it when I wrote it because we weren't using it anyway (it's currently gated behind a check that is set to false).

justinmk: The tests will tell us :)

May also need `normalize(uv.cwd)`. And technically un-normalize the result, but that's a simple matter of counting slashes.

dundargoc: > Would be very nice if we can delegate all the heavy lifting to normalize, for as many purposes as possible.

Sure, if possible we should do that. Need to figure out what the "correct"/our intended behavior we wanna be, let me know if you have any ideas.

echasnovski: That's basically [this comment](https://github.com/neovim/neovim/pull/31790#issuecomment-2565799250) and has the drawbacks of not working for path outside of target directory.

gpanders:
The `realpath` tool also returns `../apache`:

```
$ realpath --relative-to /var/lib /var/apache
../apache
```

justinmk:
> If the requirement is dropped, then implementation is fairly straightforward and allows `vim.fs.relpath(path, dir) == nil` be used to check whether `path` is descendant of `dir`:

Let's do that way then. We can add support for the "outside of target directory" case as a optional flag, later, if anyone actually requests it. But the default behavior should allow us to avoid `is_relative_to`.

test/testutil.lua
lewis6991: Maybe needs wrapping with `vim.pesc()` for good measure since cwd may contain `.`.

runtime/lua/vim/fs.lua
```diff
@@ -732,4 +732,28 @@ function M.abspath(path)
   return M.joinpath(cwd, path)
 end
 
+--- Return relative path from directory `base` to `target`. Note that `base` must be an ancestor to
+--- `target`, meaning relative paths starting with `../` are not supported.
+--- Example:
+---
+--- ```lua
+--- vim.fs.relpath('/var', '/var/lib') -- lib
```
justinmk, line 740:
```suggestion
--- Gets `target` path relative to `base`, or `nil` if `base` is not an ancestor.
---
--- Example:
---
--- ```lua
--- vim.fs.relpath('/var', '/var/lib') -- 'lib'
```

runtime/lua/vim/fs.lua
```diff
@@ -732,4 +732,28 @@ function M.abspath(path)
   return M.joinpath(cwd, path)
 end
 
+--- Return relative path from directory `base` to `target`. Note that `base` must be an ancestor to
+--- `target`, meaning relative paths starting with `../` are not supported.
```
justinmk, line 736: Not clear what `paths starting with ../ are not supported` means. `abspath()` would handle such inputs.

runtime/lua/vim/fs.lua
```diff
@@ -732,4 +732,28 @@ function M.abspath(path)
...
+--- @param target string
+--- @return string|nil
+function M.relpath(base, target)
+  vim.validate('base', base, 'string')
+  vim.validate('target', target, 'string')
+
+  base = vim.fs.normalize(vim.fs.abspath(base))
+  target = vim.fs.normalize(vim.fs.abspath(target))
+  if base == target then
+    return ''
```
justinmk, line 753: Should this be `'.'`  ?

dundargoc: I guess we have to since `vim.fs.joinpath` doesn't support empty strings to denote current directory.

test/functional/lua/fs_spec.lua
```diff
@@ -503,4 +503,27 @@ describe('vim.fs', function()
       end)
     end
   end)
+
+  describe('relpath()', function()
+    it('works', function()
+      eq(nil, vim.fs.relpath('/var/lib', '/var'))
+      eq(nil, vim.fs.relpath('/var/lib', '/bin'))
```
justinmk, line 510: test for relative case, plus some weird inputs?

maybe some whitespace and redundant slashes too...

```suggestion
      eq(nil, vim.fs.relpath('/var/lib', '/bin'))
      eq(nil, vim.fs.relpath('/var/lib', 'bin'))
      eq(nil, vim.fs.relpath('/var/lib', './bin'))
      -- ?
      eq(nil, vim.fs.relpath('/var/lib', '././'))
      -- ?
      eq(nil, vim.fs.relpath('/var/lib', '../'))
      -- ?
      eq(nil, vim.fs.relpath('/var/lib', '/'))
      -- ?
      eq(nil, vim.fs.relpath('/var/lib', '//'))
```

test/functional/lua/fs_spec.lua
```diff
@@ -503,4 +503,27 @@ describe('vim.fs', function()
...
+      eq('.', vim.fs.relpath('/var/lib', '/var/lib'))
+      eq('lib', vim.fs.relpath('/var/', '/var/lib'))
+      eq('var/lib', vim.fs.relpath('/', '/var/lib'))
+      eq('bar/package.json', vim.fs.relpath('/foo/test', '/foo/test/bar/package.json'))
+
+      if is_os('win') then
+        eq(nil, vim.fs.relpath('c:/blah\\blah', 'd:/games'))
+        eq(nil, vim.fs.relpath('c:/aaaa/bbbb', 'c:/aaaa'))
+        eq('cccc', vim.fs.relpath('c:/aaaa/', 'c:/aaaa/cccc'))
```
justinmk, line 519: more cases

```
eq(?, vim.fs.relpath('c:/aaaa/', '/aaaa/cccc'))
eq(?, vim.fs.relpath('c:/aaaa/', './aaaa/cccc'))
eq(?, vim.fs.relpath('c:/aaaa/', 'aaaa/cccc'))
```

test/functional/lua/fs_spec.lua
```diff
@@ -503,4 +503,27 @@ describe('vim.fs', function()
...
+        eq('aaaa/bbbb', vim.fs.relpath('c:/', 'c:\\aaaa\\bbbb'))
+        eq('bar/package.json', vim.fs.relpath('C:\\foo\\test', 'C:\\foo\\test\\bar\\package.json'))
```
justinmk, line 521: test case insensitive, redundant slashes?

```suggestion
        eq('bar/package.json', vim.fs.relpath([[C:\foo\test]], [[C:\fOo\\tEst\bar\\\\package.json]]))
```

dundargoc: Nothing in `vim.fs` checks for case insensitiveness though. If anything I'd assume `normalize()` to take care of that.

Also heads up: it's possible to set case-insensitiveness for a specific directory only on windows, which can break this assumptions at times: https://github.com/dart-lang/core/issues/529

justinmk: 1 line of coverage for that gives us some sort of signal in CI, why not?

dundargoc: I don't understand what you're saying. I'll make a PR to fix normalize()

dundargoc: https://github.com/neovim/neovim/pull/31833

echasnovski (changes requested):

test/functional/lua/fs_spec.lua
```diff
@@ -529,4 +526,49 @@ describe('vim.fs', function()
...
+  describe('relpath()', function()
+    it('works', function()
+      local cwd = assert(t.fix_slashes(assert(vim.uv.cwd())))
+
+      eq(nil, vim.fs.relpath('/var/lib', '/var'))
+      eq(nil, vim.fs.relpath('/var/lib', '/bin'))
+      eq(nil, vim.fs.relpath('/var/lib', 'bin'))
+      eq(nil, vim.fs.relpath('/var/lib', './bin'))
+      eq(nil, vim.fs.relpath('/var/lib', '././'))
+      eq(nil, vim.fs.relpath('/var/lib', '../'))
```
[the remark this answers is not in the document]
dundargoc: Wait, you're right. Should we just remove them for the time being?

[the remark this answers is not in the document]
echasnovski: Of use construct some directory that is guaranteed to not be ancestor for any of the four cases. Something like `vim.fs.joinpath(vim.uv.cwd(), 'foo')` should work.

test/functional/lua/fs_spec.lua
```diff
@@ -529,4 +526,53 @@ describe('vim.fs', function()
...
+        local cwd_rel_root = cwd:sub(2)
+        eq(cwd_rel_root .. '/ ', vim.fs.relpath('/', ' '))
+        eq(cwd_rel_root .. '/var', vim.fs.relpath('/', 'var'))
+      end
+
+      if is_os('win') then
+        eq(nil, vim.fs.relpath('c:/aaaa/', '/aaaa/cccc'))
+        eq(nil, vim.fs.relpath('c:/aaaa/', './aaaa/cccc'))
+        eq(nil, vim.fs.relpath('c:/aaaa/', 'aaaa/cccc'))
+        eq(nil, vim.fs.relpath('c:/blah\\blah', 'd:/games'))
```
justinmk, line 566: would be good to have a case where everything looks relative except the drive letter

```lua
        eq(nil, vim.fs.relpath('c:/games', 'd:/games'))
        eq(nil, vim.fs.relpath('c:/games', 'd:/games/foo'))
``

File: test/functional/lua/fs_spec.lua [hunk 10]
```diff
@@ -529,4 +526,53 @@ describe('vim.fs', function()
...
+        eq(nil, vim.fs.relpath('c:/aaaa/bbbb', 'c:/aaaa'))
+        eq('cccc', vim.fs.relpath('c:/aaaa/', 'c:/aaaa/cccc'))
+        eq('aaaa/bbbb', vim.fs.relpath('c:/', 'c:\\aaaa\\bbbb'))
```
justinmk, line 569: use mixed casing to drive home the point that drive-letter casing is ignored
```suggestion
        eq('aaaa/bbbb', vim.fs.relpath('C:/', 'c:\\aaaa\\bbbb'))
```

test/functional/lua/fs_spec.lua
```diff
@@ -529,4 +526,53 @@ describe('vim.fs', function()
...
+        eq('bar/package.json', vim.fs.relpath('C:\\foo\\test', 'C:\\foo\\test\\bar\\package.json'))
```
justinmk, line 570: add  mixed-case coverage

        eq(nil, vim.fs.relpath('C:\\foo\\test', 'C:\\foo\\Test\\bar\\package.json'))

justinmk (approved):

test/functional/lua/fs_spec.lua
```diff
@@ -529,4 +526,53 @@ describe('vim.fs', function()
...
+      eq('var/lib', vim.fs.relpath('/', '/var/lib'))
+      eq('bar/package.json', vim.fs.relpath('/foo/test', '/foo/test/bar/package.json'))
+      eq('foo/bar', vim.fs.relpath(cwd, 'foo/bar'))
+      eq('foo/bar', vim.fs.relpath('.', vim.fs.joinpath(cwd, 'foo/bar')))
+      eq('bar', vim.fs.relpath('foo', 'foo/bar'))
+      eq(nil, vim.fs.relpath('/var/lib', '/var/library/foo'))
+
+      if is_os('win') then
+        eq(nil, vim.fs.relpath('/', ' '))
+        eq(nil, vim.fs.relpath('/', 'var'))
```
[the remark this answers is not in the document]
dundargoc: > And besides that, Vim usually treats `/` on Windows as "root of the current drive".

We could start doing that but to my knowledge there is no precedence for that currently in `vim.fs`. Shouldn't `normalize` or `abspath` be taking care of this in that case?

justinmk (approved):

runtime/lua/vim/fs.lua
```diff
@@ -741,4 +741,35 @@ function M.abspath(path)
   return M.joinpath(cwd, path)
 end
 
+--- Gets `target` path relative to `base`, or `nil` if `base` is not an ancestor.
```
wookayin, line 744: > or `nil` if `base` is not an ancestor.

What's the main motivation behind this design? Wouldn't `base="/var/lib", target="/var/foo"` => `relpath = ../foo"` also make sense (`relpath` in other programming languages implements such a behavior)?

wookayin: Sorry for a duplicate comment, I missed the previous discussion. But I'm still worried about potential breaking changes in the future and the design of the API.

justinmk: See https://github.com/neovim/neovim/pull/31790#issuecomment-2565837243

The current behavior allows `relpath()` to also provide "is relative to" functionality.

> Wouldn't `base="/var/lib", target="/var/foo"` => `relpath = "../foo"` also make sense

We can get that in the future if it's wanted, by adding a `opts` flag. I think it's uncommon.

Or if I'm wrong, then we can make the functionality from this PR activated by `opts`. Can be a followup.

wookayin: I agree with we can avoid a complication for uncommon scenario, but as a low-level stdlib I believe the correct and reasonable default behavior is to make it behave the same and consistentyl as the common `relpath` and other programming languages (including python's `os.path.relpath`).

I believe eventually we will implement the remaining case ("*walk_up*") sometime in the future, or at least some people will find the behavior of API surprising (arguably). If the function would work correctly only under a certain `opt`, that would sound weird in my humble opinion.

If the main purpose of adding this low-level vim.fs API is to implement `is_descendant` (with being performant), why don't we add a dedicated API instead?

So please consider requiring an explicit option such as `opt = {walk_up = False}` to achieve the (current) behavior, where the default option for the API (`{ walk_up = true }`) could remain yet-to-be implemented for the time being.

dundargoc: I guess we could do that, although adding an option for an unimplemented feature sounds a bit odd. I'm fine with either option, I guess you'd need to convince @justinmk and @echasnovski though.

justinmk: > I believe the correct and reasonable default behavior is to make it behave the same and consistentyl as the common `relpath` and other programming languages (including python's `os.path.relpath`).

I didn't think about that. Though I can see that it would be useful for *sibling* files. But I have never needed that... 

The common use-case for wanting a relative path is for a target that you *know* is descendant, and you want to avoid an absolute path (e.g. "workspace roots" for LSP or other project-like resolution). 

Do you have some real-world examples where a "sibling" relative path is wanted?

dundargoc: @wookayin I'd prefer not to delay with this for much longer. If you have any examples in mind let us know, otherwise I'll go ahead and merge this soon.

echasnovski: > If you have any examples in mind let us know, otherwise I'll go ahead and merge this soon.

Might be worth adding a third `opts` argument for future use and explicitly say that it is "reserved for future use". Not mandatory, since adding it later seems to follow the API contract for Lua stdlib.

dundargoc: OK, add an `opts` argument

echasnovski (approved):

runtime/lua/vim/fs.lua
```diff
@@ -741,4 +741,38 @@ function M.abspath(path)
...
+--- ```
+---
+--- @param base string
+--- @param target string
+--- @param opts table? Reserved for future use
+--- @return string|nil
+function M.relpath(base, target, opts)
+  vim.validate('base', base, 'string')
+  vim.validate('target', target, 'string')
+  vim.validate('opts', target, 'table', true)
```
echasnovski, line 760:
```suggestion
  vim.validate('opts', opts, 'table', true)
```

runtime/lua/vim/fs.lua
dundargoc: I'll merge this, we can change the signature before release if it turns out doing the other way is more useful.

PR #31799, opened 2024-12-31
ci(news): trigger job for `perf` commit type

dundargoc:
There is a "performance" section in news.txt so it makes sense we should
also give a reminder to update news for performance improvements.


justinmk (approved):
Nice

PR #31812, opened 2025-01-01
fix(treesitter): injections highlighting order

### LINKED ISSUE #31777: Combined injections not entirely highlighted
### Problem

The treesitter highlighter fails to consider all text of a combined injection, even when it is visible on screen. Sometimes it will draw it correctly but the highlights go back to being incorrect on the next redraw(?)

### Steps to reproduce

minimal.lua:

```lua
vim.treesitter.query.set(
  'lua',
  'injections',
  [[((comment_content) @injection.content (#set! injection.self) (#set! injection.combined))]]
)

vim.api.nvim_buf_set_lines(0, 0, -1, false, {
  '-- print([[some',
  '-- text',
  '-- here]])',
})
vim.bo.ft = 'lua'
```

```sh
nvim --clean -u minimal.lua
```

Observe:
Image

Weird redrawing:
https://github.com/user-attachments/assets/8e845f91-29fe-46ae-a3bd-617e2d8387cd

### Expected behavior

In the above example, `here]]` should also have a string highlight, and the final closing parenthesis should have a bracket highlight. Instead they are not parsed as injections.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1422+g35247b00a4-dirty Build type: Debug LuaJIT 2.1.1734355927

NA

### Operating system/version

Linux 6.12.5 @ NixOS

### Terminal name/version

kitty 0.37.0

xterm-kitty

### Installation

Build from source

vanaigr:
### Problem

1. Highlights from injections might get obscured by parent highlights with the same priority. See https://github.com/neovim/neovim/issues/31777
2. Ephemeral and regular extmarks with the same priorities might change the order they appear depending on how they are redrawn. See https://github.com/neovim/neovim/pull/30869#issuecomment-2430570060

### Solution

Add `subpriority` field to `nvim_buf_set_extmarks()` `opts` that allows further control over the order of highlights. 

1 is fixed by drawing each highlight state with a higher subpriority than the previous. Regular `priority` values still work as expected. Subpriority only breaks the ties in favor of child injections.
2 is fixed by making regular extmarks use subpriority of 0 when drawn. The highlighter starts subpriorities from 1, so ephemeral extmarks are always drawn on top of regular extmarks with the same priority. 

Should `subpriority` be documented?

fix https://github.com/neovim/neovim/issues/31777

### DIFF
--- a/runtime/lua/vim/treesitter/highlighter.lua
+++ b/runtime/lua/vim/treesitter/highlighter.lua
@@ -203,11 +203,11 @@ function TSHighlighter:prepare_highlight_states(srow, erow)
   end)
 end
 
----@param fn fun(state: vim.treesitter.highlighter.State)
+---@param fn fun(state: vim.treesitter.highlighter.State, index: number)
 ---@package
 function TSHighlighter:for_each_highlight_state(fn)
-  for _, state in ipairs(self._highlight_states) do
-    fn(state)
+  for i, state in ipairs(self._highlight_states) do
+    fn(state, i)
   end
 end
 
@@ -281,7 +281,7 @@ end
 ---@param line integer
 ---@param is_spell_nav boolean
 local function on_line_impl(self, buf, line, is_spell_nav)
-  self:for_each_highlight_state(function(state)
+  self:for_each_highlight_state(function(state, state_i)
     local root_node = state.tstree:root()
     local root_start_row, _, root_end_row, _ = root_node:range()
 
@@ -333,6 +333,7 @@ local function on_line_impl(self, buf, line, is_spell_nav)
             hl_group = hl,
             ephemeral = true,
             priority = priority,
+            subpriority = state_i,
             conceal = conceal,
             spell = spell,
             url = url,
--- a/runtime/lua/vim/treesitter/highlighter.lua
+++ b/runtime/lua/vim/treesitter/highlighter.lua
@@ -203,7 +203,7 @@ function TSHighlighter:prepare_highlight_states(srow, erow)
   end)
 end
 
----@param fn fun(state: vim.treesitter.highlighter.State, index: number)
+---@param fn fun(state: vim.treesitter.highlighter.State, index: integer)
 ---@package
 function TSHighlighter:for_each_highlight_state(fn)
   for i, state in ipairs(self._highlight_states) do
--- a/src/nvim/api/keysets_defs.h
+++ b/src/nvim/api/keysets_defs.h
@@ -39,6 +39,7 @@ typedef struct {
   Boolean invalidate;
   Boolean ephemeral;
   Integer priority;
+  Integer subpriority;
   Boolean right_gravity;
   Boolean end_right_gravity;
   Array virt_lines;
--- a/src/nvim/decoration.c
+++ b/src/nvim/decoration.c
@@ -509,24 +509,28 @@ static void decor_range_add_from_inline(DecorState *state, int start_row, int st
   if (decor.ext) {
     DecorVirtText *vt = decor.data.ext.vt;
     while (vt) {
-      decor_range_add_virt(state, start_row, start_col, end_row, end_col, vt, owned);
+      decor_range_add_virt(state, start_row, start_col, end_row, end_col,
+                           vt, 0, owned);
       vt = vt->next;
     }
     uint32_t idx = decor.data.ext.sh_idx;
     while (idx != DECOR_ID_INVALID) {
       DecorSignHighlight *sh = &kv_A(decor_items, idx);
-      decor_range_add_sh(state, start_row, start_col, end_row, end_col, sh, owned, ns, mark_id);
+      decor_range_add_sh(state, start_row, start_col, end_row, end_col,
+                         sh, owned, 0, ns, mark_id);
       idx = sh->next;
     }
   } else {
     DecorSignHighlight sh = decor_sh_from_inline(decor.data.hl);
-    decor_range_add_sh(state, start_row, start_col, end_row, end_col, &sh, owned, ns, mark_id);
+    decor_range_add_sh(state, start_row, start_col, end_row, end_col,
+                       &sh, owned, 0, ns, mark_id);
   }
 }
 
 static void decor_range_insert(DecorState *state, DecorRange *range)
 {
-  range->ordering = state->new_range_ordering++;
+  range->priority |= (uint64_t)state->new_range_ordering;
+  state->new_range_ordering++;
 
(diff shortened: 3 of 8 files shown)


src/nvim/decoration.c
```diff
@@ -692,16 +697,15 @@ int decor_redraw_col_impl(win_T *wp, int col, int win_col, bool hidden, DecorSta
...
+    uint64_t const priority = r->priority;
 
     int begin = 0;
     int end = cur_end;
     while (begin < end) {
       int mid = begin + ((end - begin) >> 1);
       int mi = indices[mid];
       DecorRange *mr = &slots[mi].range;
-      if (mr->priority < priority || (mr->priority == priority && mr->ordering < ordering)) {
+      if (mr->priority <= priority) {
```
vanaigr, line 708: Comparison changed from `<` to `<=`. Should work identically, since `ordering` (now a part of priority) is unique.

runtime/lua/vim/treesitter/highlighter.lua
```diff
@@ -203,11 +203,11 @@ function TSHighlighter:prepare_highlight_states(srow, erow)
   end)
 end
 
----@param fn fun(state: vim.treesitter.highlighter.State)
+---@param fn fun(state: vim.treesitter.highlighter.State, index: number)
```
lewis6991, line 206:
```suggestion
---@param fn fun(state: vim.treesitter.highlighter.State, index: integer)
```

src/nvim/api/keysets_defs.h
```diff
@@ -39,6 +39,7 @@ typedef struct {
   Boolean invalidate;
   Boolean ephemeral;
   Integer priority;
+  Integer subpriority;
```
lewis6991, line 42: I think we should try to implement this without subpriority. We can just shift all the current base priorities by some magnitude to give us enough space for priorities in between.

clason:
How does this relate to https://github.com/neovim/neovim/pull/27131?

lewis6991:
We wanted to introduce subpriorities then because the highlighter was changed to use `iter_matches` from `iter_captures` in order to maintain highlight behaviour.

In the end I think I (or someone) figured out how to implement what we wanted with `iter_captures` so we reverted the subpriorities change since it was no longer needed.

clason:
Yes, but there was some (critical) discussion around subpriorities, and I think those points apply here as well and should be considered.

lewis6991:
I don't think anyone really wanted it, but it was seen as a simple means to an ends, but @bfredl did somewhat object that it wasn't necessary, and my opinion has tipped over time into that direction. Priority is already a 64bit integer, which should allow plenty of room for everything we need.

vanaigr:
> Priority is already a 64bit integer, which should allow plenty of room for everything we need.

Priority is limited to 65'535 at least in the extmarks.

src/nvim/api/keysets_defs.h
vanaigr: I don't know what priorities other than 100 should do. I know there's at leas 1 case in which priority 99 is used to make a highlight get obscured by parent highlight: https://github.com/nvim-treesitter/nvim-treesitter/pull/5649.

There's also a question about whether highlights > 100 should appear above all other highlights from other trees, or only child highlights, or only this tree's highlights.

lewis6991:
Ok my mistake. That's still a large number though.

justinmk:
> 2 is fixed by making regular extmarks use subpriority of 0 when drawn. The highlighter starts subpriorities from 1, so ephemeral extmarks are always drawn on top of regular extmarks with the same priority.

Does this suggest that we need a `kind:string` field on extmarks? It looks like subpriorities are used to encode semantics with numbers instead of an explicit "kind".

Subpriorities only resolve 1 level of conflicts, right?

vanaigr:
> Does this suggest that we need a `kind:string` field on extmarks? It looks like subpriorities are used to encode semantics with numbers instead of an explicit "kind". (Correct me if wrong :)
> 
> Subpriorities only resolve 1 level of conflicts, right?

Not sure what that means.

Usually, regular extmarks are added at the last possible moment, and thus their highlight is applied on top of other decorations with the same priority.

But that is not the case for when the process is first started. The first extmark (and possibly others) are added before everything else, and get obscured by other decprations with the same priority.

The same happens with multiline regular extmarks. When they are first added, they appear above. But on the next line, new decorations with the same priprity are added and now the extmark is behind them.

justinmk:
> regular extmarks are added at the last possible moment

So does it make sense to annotate "regular" and other (non-regular) extmarks with a field that indicates this with a name, instead of a secondary (numeric) priority?

lewis6991:
How would we use that to control ordering/priority of extmarks? How are the names compared?

justinmk:
Would need to be enum-like so the names have ordering. However, reviewing how `treesitter/highlighter.lua` uses `subpriority`, I see now that names aren't a good fit.

ribru17:
Can #31777 be solved without messing with priorities? I wonder if this could be done completely within the highlighter, seeing as it is not the case that all injections render weirdly, just combined ones

vanaigr:
Not sure if it matters, but there are situations where the same behavior will happen for regular injections: 

<details>

```lua
for name, url in pairs {
  'https://github.com/rose-pine/neovim.git'
} do
  end
end

require'nvim-treesitter.configs'.setup {
  sync_install = true,
  ensure_installed = { "javascript", "styled" },
}

vim.api.nvim_buf_set_lines(0, 0, -1, true, {
  'css`',
  '.test {',
  '    prop: "5',
  '    ${a} ${b}  3";',
  '}',
  '`',
})

vim.treesitter.start(0, 'javascript')

vim.treesitter.query.set(
  'javascript',
  'highlights',
  [[(string_fragment) @fragment]]
)

vim.api.nvim_set_hl(0, "@fragment.javascript", { bg = 'green', italic = true })
vim.api.nvim_set_hl(0, "@string.styled", { bg = 'red' })
```

</details>

image
image

src/nvim/api/keysets_defs.h
ribru17: @vanaigr Do you mind if I open a PR trying this solution? I will credit you in the commit :) (See https://github.com/neovim/neovim/commit/26c350ab84e6c6fab29001be474f0c4bccebec14)

> I know priority 99 was used at least once to make a highlight get obscured by parent highlight

This is used in a few places (usually using priority 90) where we know injections will be added, so we do this to prevent the parent highlights from overriding them. I think these changes will be unnecessary if we just shift the base priorities over

PR #31815, opened 2025-01-01
fix(vim.fs.abspath): correctly handle UNC paths

dundargoc:


runtime/lua/vim/fs.lua
```diff
@@ -587,15 +587,14 @@ function M.normalize(path, opts)
   end
 
   local win = opts.win == nil and iswin or not not opts.win
-  local os_sep_local = win and '\\' or '/'
```
justinmk, line 590: `opts.win` was intentionally supposed to allow the caller to opt-in to Windows behavior regardless of the current platform. Mostly for tests but also for control/flexibility

justinmk (approved):

PR #31819, opened 2025-01-01
Visual block insert - preview results

### LINKED ISSUE #20329: Visual block insert – preview results
### Feature already in Vim?

No

### Feature description

Since you can preview the results of `:s/` for example, I think you should also be able to preview the results of visual block inserting. Let's say we have this block of text:

```
text1
text2
text3
```

At the top left, I press <kbd>Control-V</kbd> to enter visual block mode. I go down twice and then press <kbd>Shift-I</kbd>. At this point I should be able to see three cursors in front of each line. As I type some text, I should be seeing the text on all of the lines, not just the top one. This should definitely be the default behavior, just like `incsearch`.

TheLeoP:
Should fix #20329


src/nvim/block_defs.h
justinmk, line 1: curious about this file, is it to avoid circular dependency? a name like `ops_block_defs.h` is probably more discoverable.

src/nvim/undo_defs.h
```diff
@@ -2,13 +2,42 @@
...
+  int save_b_u_save_nr_last;
+  int save_b_u_seq_cur;
+  time_t save_b_u_time_cur;
+  int save_b_u_save_nr_cur;
+  char *save_b_u_line_ptr;
+  linenr_T save_b_u_line_lnum;
+  colnr_T save_b_u_line_colnr;
+} UndoInfo;
+
+typedef struct {
```
justinmk, line 31: a one-line docstring that describes this would be useful. I assume it's `preview buffer info`

src/nvim/block_defs.h
TheLeoP: I added it because trying to include `ops.h` on another header file (`undo_defs.h` in this case) gave me a build error because of  `EXTERN LuaRef repeat_luaref INIT( = LUA_NOREF);  ///< LuaRef for "."`

src/nvim/undo_defs.h
TheLeoP: I'll add it

PR #31820, opened 2025-01-01
[WIP] refactor(options): use reference counted strings for options

famiu:


justinmk:
What actual problem is this solving? AFAICT, there are some small performance benefits, and a big complexity cost.

pro:

- avoids deep-copying option strings when they're shared.
- copying buffer-local options for a new buffer (~30 options copied)
- Inheritance / `:setlocal opt<`, copying a global value to local.

con:

1. premature optimization: option values are small, and buffer/window creation is not a hot path.
2. more complexity. big, high-risk change.

PR #31829, opened 2025-01-02
fix(lsp): ensure watcher cancel

tomtomjhj:
* Cancel watcher in the "force" case.
* Cancel watcher outside the async callback. It seems nvim doesn't wait async jobs on quitting, leaving detached inotifywait processes.
* Clean up cancelling callbacks.


justinmk:
> It seems nvim doesn't wait async jobs on quitting, leaving detached inotifywait processes.

Yeah, because we got complaints about Nvim taking time to exit. Though ideally we would at least cancel and deref all the libuv/luv handles. Though I think there was a difficulty with that, I don't remember.

PR #30483, opened 2024-09-23
use "nvim -l" as test runner

justinmk:
## Problem:
61e99217e684 replaced usages of `vim.fn`. This duplicates non-trivial logic and may have introduced bugs like 38e38d1b401e.

Later on, b02eeb6a7281 graduated `fnamemodify` to `fast`, so avoiding it in `vim.fs` is no longer necessary.

## Solution:
- Use `vim.fn` to deduplicate logic and increase quality.
- Use `nvim -l` instead of `nvim -ll` in the test runner.
    - This is a test balloon to see what CI does.

PR #31833, opened 2025-01-02
fix(vim.fs.normalize): normalize case for windows drive letter

dundargoc:
Also add tests for the current path casing behavior so it doesn't get
accidentally changed.


justinmk:
> Relevant: [dart-lang/core#529](https://github.com/dart-lang/core/issues/529)

hmmm...

> Note that this means that, in general, paths cannot be _truly_ canonicalized (in the sense of "guaranteeing that two paths that refer to the same location on disk will have the same string") without actually interacting with the filesystem on Windows

an example of why "pure" path operations are a leaky abstraction.

though "normalize", by definition, yearns for a "canonical", universal representation :(

echasnovski:
I think this might be a serious enough change (if merged) to deserve an explicit documentation and mention in 'news.txt'.

test/functional/lua/fs_spec.lua
```diff
@@ -340,22 +340,25 @@ describe('vim.fs', function()
   end)
 
   describe('normalize()', function()
+    local function normalize(path, opts)
+      return exec_lua([[return vim.fs.normalize(...)]], path, opts)
```
justinmk, line 344: can this be avoided by using Lua functions instead of of `vim.fn.tolower`? In https://github.com/neovim/neovim/pull/30483 I am endeavoring to revert this.

dundargoc: I don't think lua is unicode-aware. Perhaps we can get away with not considering unicode for the time being?

justinmk: > Perhaps we can get away with not considering unicode for the time being?

Yeah I wouldn't worry about that. Case-insensitivity is not well-defined for non-ASCII chars. I can't find a reference, but I wonder if Windows "case insensitivity" applies to non-ASCII chars anyways.

justinmk:
Reading https://stackoverflow.com/a/34000339/152142 makes me wonder if we should treat Windows as case-sensitive.

It's trivial for callers to use `string.lower()` , so maybe it's better that we just mention this in the docstring for normalize()? Any tests you added here would still be useful.

echasnovski:
> It's trivial for callers to use `string.lower()` , so maybe it's better that we just mention this in the docstring for normalize()? Any tests you added here would still be useful.

I am pretty sure it would have to be `vim.fn.tolower()`. There is pretty well defined case sensitivity for many reasonable cases, like non-Latin alphabets (Cyrillic, for example).

justinmk:
> There is pretty well defined case sensitivity for many reasonable cases, like non-Latin alphabets (Cyrillic, for example).

But how does the Windows API define it? It's an extremely old API, and based on that SO link, this case-insensitive behavior is only for backwards-compatibility, and may be limited to ASCII.

test/functional/lua/fs_spec.lua
dundargoc: It does. `кириллица` is the same file as `КИРИЛЛИЦА`.

dundargoc:
That "may depend" tho. I tested it, changing to turkish locale doesn't affect how windows treats the file regards to turkish i vs. İ, likely because this specific case transformation isn't well defined by unicode.

dundargoc:
> Reading [stackoverflow.com/a/34000339/152142](https://stackoverflow.com/a/34000339/152142) makes me wonder if we should treat Windows as case-sensitive.

I don't have strong feelings on this. I only created this because you sent me on this sidequest [here](https://github.com/neovim/neovim/pull/31790#discussion_r1900883389). If we're not we can close this and skip testing for case for the time being?

justinmk:
> skip testing for case for the time being?

we should still have a couple test cases that assert case-sensitivity. i.e. add some uppercase chars to an existing test case, since I think most/all of them are lowercase

dundargoc:
As in, check that we *don't* consider case insensitivity at all in vim.fs? So, an example would be

```
neq('C:/', vim.fs.normalize('c:/'))
```
?

echasnovski:
I've just realized that there is a talk here only about case insensitivity on Windows, right? The MacOS is also case insensitive or at least it is like that by default.

So my guess would be to follow what Vim does and just delegate that decision to `vim.fn.has('fname_case') == 0`. If it returns incorrect result for some OS-settings pair, then adjust that in Vim.

dundargoc:
> I've just realized that there is a talk here only about case insensitivity on Windows, right? The MacOS is also case insensitive or at least it is like that by default.

I know, but people didn't like assuming case insensitive on macos because it is possible to use a case sensitive file system.

> So my guess would be to follow what Vim does and just delegate that decision to vim.fn.has('fname_case') == 0. If it returns incorrect result for some OS-settings pair, then adjust that in Vim.

This is actually decided via a cmake variable/build system, so following vim won't help us here. But perhaps relying on `fname_case` is still a better idea. Also, it's unfortunately not a get out of jail-free card (see linked dart issue) but it's a much better 98% solution IMO. (Honestly, directory-specific case sensitive windows directories is super cursed and I'm pretty fine not supporting it lol)

mfussenegger:
>> I've just realized that there is a talk here only about case insensitivity on Windows, right? The MacOS is also case insensitive or at least it is like that by default.

> I know, but people didn't like assuming case insensitive on macos because it is possible to use a case sensitive file system.

Same on linux, common examples: usb-drives, smb mounts. Less common: cd-roms (Hello iso9660 and Joliet)


A bit Haskell specific, but [this article](https://hasufell.github.io/posts/2022-06-29-fixing-haskell-filepaths.html) does a decent job of outlining some problems with the path handling. (Encoding/locale being mount specific, but almost all languages and frameworks assume utf-8 encoding or use system-locale, some with a way to override. Java has a `sun.jnu.encoding` system property, glib has a `G_FILENAME_ENCODING`, etc.)


> so, it's unfortunately not a get out of jail-free card (see linked dart issue) but it's a much better 98% solution IMO. (Honestly, directory-specific case sensitive windows directories is super cursed and I'm pretty fine not supporting it lol)

I'm not sure if a 98% solution is better in this case because the 2% where it doesn't work can manifest in obscure bugs that are difficult to track down. I'd either keep treating it case-sensitive because there it is easier to notice what's wrong, or aim for a 100% solution with some path abstractions.

fredizzimo:
Just linking the actual documentation here, rather than random stackoverflow and chatgpt answers
https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity

The usage might be more widespread than assumed due to wsl, but on the other hand in those cases you probably run nvim from within wsl, so I don't know. I still think, there should be at least an option to disable case normalization globally, if it's not done correctly per directory.

justinmk:
> there should be at least an option to disable case normalization globally, if it's not done correctly per directory.

That would defeat the entire purpose of normalization. Anyway, we currently don't normalize casing, so no option is needed.

It seems clear now that in order to normalize casing, we need "full" support, otherwise we are making things worse. Though there are some low-risk things we can do before then, such as normalizing drive-letter casing.

fredizzimo:
Normalizing might be a bad idea in general, Windows does actually support case sensitive files in other dirctories as well https://www.tiraniddo.dev/2019/02/ntfs-case-sensitivity-on-windows.html. 

The solution is probably to open the file with the `FILE_FLAG_POSIX_SEMANTICS` instead.

justinmk:
> I'm not sure what the actual problem is that this PR tries to fix and the linked dart issue tells the opposite, to _never_ normalize it.

You mean don't normalize the casing. That is different than "don't normalize". "Normalize" is a synonym for "canonicalize".

> It can be cannonicialized though https://learn.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchcanonicalizeex

`vim.fs.normalize` already has that functionality.

fredizzimo:
> You mean don't normalize the casing

Yes, I edited the comment to say that. And please din't hide this excellent link https://www.tiraniddo.dev/2019/02/ntfs-case-sensitivity-on-windows.html.

dundargoc:
> NOTE: I'm not sure what the actual problem is that this PR tries to fix 

Addressing review comments in https://github.com/neovim/neovim/pull/31790#discussion_r1900883389

fredizzimo:
> > NOTE: I'm not sure what the actual problem is that this PR tries to fix
> 
> Addressing review comments in [#31790 (comment)](https://github.com/neovim/neovim/pull/31790#discussion_r1900883389)

Ah, then it aound like the case normalization should be added to realpath, just like in python for example https://docs.python.org/3/library/os.path.html#os.path.realpath. The difference is that realpath is allowed to read the file system, and actually can do it correctly, while normalize just operates on strings.

dundargoc:
Pivoting PR according to https://github.com/neovim/neovim/pull/31833#issuecomment-2571320869

dundargoc:
Discussion continued in https://github.com/neovim/neovim/issues/31863

runtime/lua/vim/fs.lua
```diff
@@ -629,6 +629,10 @@ function M.normalize(path, opts)
       return prefix .. path
     end
 
+    if prefix:match('%w:') then
+      prefix = string.upper(prefix)
+    end
+
     -- Remove extraneous slashes from the prefix
     prefix = prefix:gsub('/+', '/')
```
echasnovski, line 637:
```suggestion
    -- Ensure capital drive and remove extraneous slashes from the prefix
    prefix = prefix:gsub('^%a:', string.upper):gsub('/+', '/')
```

A neat little trick is that `gsub` allows function as replacement, which is appropriate here.

echasnovski (approved):

PR #31836, opened 2025-01-02
doc: Fix missing newline in vim.system() example output

mhellstr:
Problem:
Docs were missing a trailing newline character for stdout output.

Solution:
Newline added.

This then also matches the test in https://github.com/neovim/neovim/blob/master/test/functional/lua/system_spec.lua#L56-L58


dundargoc:

Issue #30940, opened 2024-10-24
WinSeparator highlight not applied


dstein64:
### Problem

`WinSeparator` highlight is not applied when specified in a config file after calling `colorscheme`. This appears to only happen when the terminal has a light background. Perhaps the issue relates to the light/dark background detection and handling.

### Steps to reproduce

Set the terminal background to a light color.

Create an init file, `~/.config/nvim/init.vim`, with the following contents:

```vim
colorscheme vim                                                        
highlight WinSeparator term=reverse cterm=reverse gui=reverse
```

Restart nvim and create a vertical split:

```vim
:vsplit
```

### Actual and Expected behavior

The specified highlighting is not applied.

<img src="https://github.com/user-attachments/assets/12a34cac-5561-465e-8e8a-fbb60b44bc1e" width="640" />

Executing `:highlight WinSeparator term=reverse cterm=reverse gui=reverse` results in the expected highlighting.

<img src="https://github.com/user-attachments/assets/b858405f-5885-43b9-8021-0c6b1920347b" width="640" />

### Nvim version (nvim -v)

v0.10.0, v0.10.2, NVIM v0.11.0-dev-1044+g01739d467

### Vim (not Nvim) behaves the same?

Vim v9.1does not have the issue when using VertSplit instead of WinSeparator

### Operating system/version

Kubuntu 24.01

### Terminal name/version

Konsole 23.08.5

### $TERM environment variable

xterm-256color

### Installation

appimage

justinmk:
Please test the latest development version. The [Releases](https://github.com/neovim/neovim/releases/tag/nightly) page has pre-built archives for Linux/Windows/macOS. Check `:version` to confirm the version.

dstein64:
> "Latest stable release is 0.10.2, why use 0.10.0?"

I should have checked what the latest version is and also try the development version to see if the issue was resolved.

> "Meanwhile, please test the latest development version. The [Releases](https://github.com/neovim/neovim/releases/tag/nightly) page has pre-built archives for Linux/Windows/macOS. Check :version to confirm the version."

I updated the report above to include v0.10.2, and the latest development version. I tested on both of those, and encountered the issue.

Including text from your comment as a caption under _Nvim version_ on the bug report page might help to prevent similar occurrences in the future (e.g., "Please test the latest development version. The [Releases](https://github.com/neovim/neovim/releases/tag/nightly) page has pre-built archives for Linux/Windows/macOS.").

[...]

PR #31842, opened 2025-01-03
feat(health): add bug report section

### LINKED ISSUE #31847: :checkhealth can create a bug report
Should be somehow integrated with `:checkhealth`. As @clason hinted, the checkhealth report should have all info that we want. Then in the checkhealth buffer, there should be a way the user can start a bug report, and it opens a URL like:

https://github.com/neovim/neovim/issues/new?labels=bug&title=user+docs+HTML%3A+gui.txt+&body=%23%23%20Problem%3A%0D%0D%23%23%20Steps%20to%20reproduce:%0D%0D%60%60%60%0DTODO%0D%60%60%60%0D%0D%23%23%20Expected%20behavior:%0D%0D%23%23%20System%20info:%0D%0D

which contains all the info that we want.

Instead of a `:bugreport` command, a buffer-local command like `gX` ? Or `:checkhealth bug`.

_Originally posted by @justinmk in https://github.com/neovim/neovim/issues/31842#issuecomment-2569460601_

glepnir:
Problem: there is no simple way to retrieve various pieces of information
for a bug report.

Solution: add a bug report section to checkhealth in vim.health

close https://github.com/neovim/neovim/issues/31847

### DIFF
--- a/runtime/lua/vim/health/health.lua
+++ b/runtime/lua/vim/health/health.lua
@@ -493,7 +493,153 @@ local function check_external_tools()
   end
 end
 
+local function detect_terminal()
+  local e = vim.env
+  if e.TERM_PROGRAM then
+    local v = e.TERM_PROGRAM_VERSION --- @type string?
+    return e.TERM_PROGRAM_VERSION and (e.TERM_PROGRAM .. ' ' .. v) or e.TERM_PROGRAM
+  end
+
+  local map = {
+    KITTY_WINDOW_ID = 'kitty',
+    ALACRITTY_SOCKET = 'alacritty',
+    ALACRITTY_LOG = 'alacritty',
+    WEZTERM_EXECUTABLE = 'wezterm',
+    KONSOLE_VERSION = function()
+      return 'konsole ' .. e.KONSOLE_VERSION
+    end,
+    VTE_VERSION = function()
+      return 'vte ' .. e.VTE_VERSION
+    end,
+  }
+
+  for key, val in pairs(map) do
+    local env = e[key] --- @type string?
+    if env then
+      return type(val) == 'function' and val() or val
+    end
+  end
+
+  return 'unknown'
+end
+
+local function check_sysinfo()
+  vim.health.start('System Info')
+
+  -- Use :version because vim.version().build returns "Homebrew" for brew installs.
+  local version_output = vim.api.nvim_exec2('version', { output = true }).output
+  local nvim_version = version_output:match('NVIM v[^\n]+') or 'unknown'
+  local commit_hash = nvim_version:match('%+g?(%x+)')
+  local version_for_report = nvim_version
+  if commit_hash then
+    version_for_report = nvim_version:gsub('%+g' .. commit_hash, ' neovim/neovim@' .. commit_hash)
+    local has_git = vim.fn.executable('git') == 0
+    local has_curl = vim.fn.executable('curl') == 1
+    local cmd = has_git and { 'git', 'ls-remote', 'https://github.com/neovim/neovim', 'nightly' }
+      or has_curl and {
+        'curl',
+        '-s',
+        'https://api.github.com/repos/neovim/neovim/commits/nightly',
+        '-H',
+        'Accept: application/vnd.github.sha',
+      }
+      or nil
+
+    if cmd then
+      local result = vim.system(cmd, { text = true }):wait()
+      if result.code == 0 then
+        local sha = has_git and result.stdout:match('^(%x+)') or result.stdout
+        local latest_commit = sha and sha:sub(1, 10)
+        if latest_commit and commit_hash ~= latest_commit then
+          vim.health.warn(
+            string.format('Build is outdated. Local: %s, Latest: %s', commit_hash, latest_commit)
+          )
+        end
+      end
+    else
+      vim.health.warn('Cannot check for updates: git or curl not found')
+    end
+  end
+
+  local os_info = vim.uv.os_uname()
+  local os_string = os_info.sysname .. ' ' .. os_info.release
+  local terminal = detect_terminal()
+  local term_env = vim.env.TERM or 'unknown'
+
+  vim.health.info('Nvim version: ' .. nvim_version)
+  vim.health.info('Operating system: ' .. os_string)
+  vim.health.info('Terminal: ' .. terminal)
+  vim.health.info('$TERM: ' .. term_env)
+
+  local body = vim.text.indent(
+    0,
+    string.format(
+      [[
+    ## Problem:
+
+
+
+    ## Steps to reproduce:
+    ```
+    nvim --clean
+    ```
+
+    ## Expected behavior:
+
+
+
+    ## Nvim version (nvim -v):
+
+    %s
+
+    ## Vim (not Nvim) behaves the same?
+
+
+
+    ## Operating system/version:
+
+    %s
+
+    ## Terminal name/version:
+
+    %s
+
+    ## $TERM environment variable:
+
+    %s
+
+    ## Installation:
+
+]],
+      version_for_report,
+      os_string,
+      terminal,
+      term_env
+    )
+  )
+
+  local encoded_body = vim.uri_encode(body) --- @type string
+  local issue_url = 'https://github.com/neovim/neovim/issues/new?labels=bug&body=' .. encoded_body
+  vim.schedule(function()
+    local win = vim.api.nvim_get_current_win()
+    local buf = vim.api.nvim_win_get_buf(win)
+    if vim.bo[buf].filetype ~= 'checkhealth' then
+      return
+    end
+    _G.nvim_health_bugreport_open = function()
+      vim.ui.open(issue_url)
+    end
+    vim.wo[win].winbar =
+      '%#WarningMsg#%@v:lua.nvim_health_bugreport_open@Click to Create Bug Report on GitHub%X%*'
+    vim.api.nvim_create_autocmd('BufDelete', {
+      buffer = buf,
+      command = 'lua _G.nvim_health_bugreport_open = nil',
+    })
+  end)
+end
+
 function M.check()
+  check_sysinfo()
   check_config()
   check_runtime()
   check_performance()


clason:
What does this offer over `:checkhealth`?

glepnir:
ref https://github.com/neovim/neovim/issues/30940#issuecomment-2438052648

clason:
Yes, if `:checkhealth` and `:version` are missing relevant information, they should be added there.

(Most important fields of the issue template are situational and/or not available to Nvim, for example the installation method.)

glepnir:
Yes. As a unified external interface. It should be there. I will change it later.

src/nvim/ex_cmds.c
```diff
@@ -4802,3 +4804,37 @@ void ex_oldfiles(exarg_T *eap)
...
+  list_version();
+
+  msg("$TERM environment variable:", 0);
+  const char *term = os_getenv("TERM");
+  msg(term != NULL ? term : "UNKNOWN", 0);
+
+  msg("Operating system/version:", 0);
+  Error err = ERROR_INIT;
+  Object ret = NLUA_EXEC_STATIC(
+    "if jit then "
```
justinmk, line 4820: just call a runtime file, don't inline it

justinmk:
Should be somehow integrated with `:checkhealth`. As @clason hinted, the checkhealth report should have all info that we want. Then in the checkhealth buffer, there should be a way the user can start a bug report, and it opens a URL like:

https://github.com/neovim/neovim/issues/new?labels=bug&title=user+docs+HTML%3A+gui.txt+&body=%23%23%20Problem%3A%0D%0D%23%23%20Steps%20to%20reproduce:%0D%0D%60%60%60%0DTODO%0D%60%60%60%0D%0D%23%23%20Expected%20behavior:%0D%0D%23%23%20System%20info:%0D%0D

which contains all the info that we want.

Instead of a `:bugreport` command, a buffer-local command like `gX` ? Or `:checkhealth bug`.

glepnir:
Last night before I went to sleep, I suddenly thought of this idea. I feel it would make creating issues very convenient. I’m not sure whether others would accept it, though.  there create an acwrite buffer where the user fills in the essential information. After that collects the required system details, merges everything together, and uses gh to create the issue — all without opening a browser.

test4

justinmk:
> collects the required system details, merges everything together, and uses gh to create the issue — all without opening a browser.

Opening a browser is a feature. It's less surprising, and will give users a way to continue, if the bug report tool doesn't work perfectly for some reason.

glepnir:
alright… then maybe the approach is to run `check bug`, collect the necessary information, and then use gx to open the “Create Issue” link in browser.  just suddenly realized there might be a way to create issues without leaving Neovim, so I tried it out.

glepnir:
I’ve updated the implementation. does this look good🤔

test4

runtime/lua/vim/bugreport/health.lua
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  return 'unknown'
+end
+
+function M.check()
+  vim.health.start('Bug Report Information')
+
+  local version_output = vim.api.nvim_exec2('version', { output = true }).output
+  local nvim_version = version_output:match('NVIM v[^\n]+') or 'unknown'
```
justinmk: ? `vim.version`

runtime/lua/vim/bugreport/health.lua
```diff
@@ -0,0 +1,126 @@
...
+  vim.schedule(function()
+    local bufnr = vim.api.nvim_get_current_buf()
+    local ns = vim.api.nvim_create_namespace('bugreport')
+    local line_count = vim.api.nvim_buf_line_count(bufnr)
+
+    vim.api.nvim_buf_set_extmark(bufnr, ns, line_count - 2, 2, {
+      end_col = #text + 2,
+      url = issue_url,
+      hl_group = 'Underlined',
+    })
```
justinmk: if we want a "button" maybe we should use 'winbar'

not a blocker.

runtime/lua/vim/bugreport/health.lua
justinmk, line 1: Rather than add a new module , just add to the existing `vim.health`  section, in `runtime/lua/vim/health/health.lua`

this "system info" can just always display in the `vim.health` section, with a "do [something] to create a bug report" info message in that section.

runtime/lua/vim/bugreport/health.lua
```diff
@@ -0,0 +1,126 @@
...
+    return vim.env.TERM_PROGRAM .. (version ~= '' and ' ' .. version or '')
+  end
+
+  if vim.env.KITTY_WINDOW_ID then
+    local result = vim.system({ 'kitty', '--version' }, { text = true }):wait()
+    if result.code == 0 then
+      local v = vim.version.parse(result.stdout, { strict = false })
+      return v and ('kitty ' .. tostring(v)) or 'kitty'
+    end
+    return 'kitty'
```
justinmk: dedupe this logic

justinmk:
Looks like a fine start.

runtime/lua/vim/bugreport/health.lua
glepnir: because I can't get commit hash from vim.version 🤔

justinmk: It's reported as `build` field:

```
:lua =vim.version()
{
  build = "ge0308dc14730",
  major = 0,
  ...
}
```

btw while we're doing this, would be very nice for the bug report to replace the `g` at start of the version with a space char, so that it auto-hyperlinks in github.

glepnir: Oh if build from package manager there just show name of it.. so can regex can as a fallback ?

```
{
  api_compatible = 0,
  api_level = 14,
  api_prerelease = true,
  build = "Homebrew",
  major = 0,
  minor = 12,
  patch = 0,
  prerelease = "dev",
  <metatable> = {
    __eq = <function 1>,
    __index = <function 2>,
    __le = <function 3>,
    __lt = <function 4>,
    __newindex = <function 5>,
    __tostring = <function 6>
  }
}
```

justinmk: oh. hmm... ok, let's use `:version` as you did then.

but add a 1-line comment explaining it:

    -- Use :version because vim.version().build returns "Homebrew" for brew installs.

glepnir: I’ll take a look at version.lua later and see if I can resolve it there ..

glepnir: I've tried use `neovim/neovim@commithas` it works like this 

image

runtime/lua/vim/health/health.lua
```diff
@@ -493,6 +493,120 @@ local function check_external_tools()
...
+  local os_string = os_info.sysname .. ' ' .. os_info.release
+  local terminal = detect_terminal()
+  local term_env = vim.env.TERM or 'unknown'
+
+  vim.health.info('Nvim version: ' .. nvim_version)
+  vim.health.info('Operating system: ' .. os_string)
+  vim.health.info('Terminal: ' .. terminal)
+  vim.health.info('$TERM: ' .. term_env)
+
+  local body = string.format(
```
justinmk: use `vim.text.indent(0, ...)` so this string can be indented

runtime/lua/vim/bugreport/health.lua
glepnir: image

glepnir:
add a remote and local commit hash compare

image

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+
+  local encoded_body = vim.uri_encode(body) --- @type string
+  local issue_url = 'https://github.com/neovim/neovim/issues/new?labels=bug&body=' .. encoded_body
+  vim.schedule(function()
+    local win = vim.api.nvim_get_current_win()
+    local buf = vim.api.nvim_win_get_buf(win)
+    if vim.bo[buf].filetype ~= 'checkhealth' then
+      return
+    end
+    _G._bugreport_open = function()
```
justinmk: As always, we want to "namespace" anything in a global scope.
```suggestion
    _G._nvim_health_bugreport_open = function()
```

Should probably do something like `_G._... = nil` on BufDelete.

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+  vim.health.info('Nvim version: ' .. nvim_version)
+  vim.health.info('Operating system: ' .. os_string)
+  vim.health.info('Terminal: ' .. terminal)
+  vim.health.info('$TERM: ' .. term_env)
+
+  local body = vim.text.indent(
+    0,
+    string.format(
+      '## Problem:\n\n\n\n'
+        .. '## Steps to reproduce:\n\n```\nnvim --clean\n\n```\n\n'
```
justinmk: we have `vim.text.indent(0, [[..]])` which dedents text. Never need to concat a bunch of strings like this.

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+local function check_sysinfo()
+  vim.health.start('Bug Report Information')
+
+  -- Use :version because vim.version().build returns "Homebrew" for brew installs.
+  local version_output = vim.api.nvim_exec2('version', { output = true }).output
+  local nvim_version = version_output:match('NVIM v[^\n]+') or 'unknown'
+  local commit_hash = nvim_version:match('%+g?(%x+)')
+  local version_for_report = nvim_version
+  if commit_hash then
+    version_for_report = nvim_version:gsub('%+g' .. commit_hash, ' neovim/neovim@' .. commit_hash)
```
justinmk, line 535: nice 👍

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+    if env then
+      return type(val) == 'function' and val() or val
+    end
+  end
+
+  return 'unknown'
+end
+
+local function check_sysinfo()
+  vim.health.start('Bug Report Information')
```
justinmk: let's call it System Info. the bug report button tells its own story, don't need to repeat the story in this heading.

justinmk:
```suggestion
  vim.health.start('System Info')
```

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+  -- Use :version because vim.version().build returns "Homebrew" for brew installs.
+  local version_output = vim.api.nvim_exec2('version', { output = true }).output
+  local nvim_version = version_output:match('NVIM v[^\n]+') or 'unknown'
+  local commit_hash = nvim_version:match('%+g?(%x+)')
+  local version_for_report = nvim_version
+  if commit_hash then
+    version_for_report = nvim_version:gsub('%+g' .. commit_hash, ' neovim/neovim@' .. commit_hash)
+    local result = vim
+      .system({ 'git', 'ls-remote', 'https://github.com/neovim/neovim', 'HEAD' }, { text = true })
+      :wait()
```
justinmk: what happens if `git` is not on the system? checkhealth shouldn't just fail. this will raise an error.

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+
+  -- Use :version because vim.version().build returns "Homebrew" for brew installs.
+  local version_output = vim.api.nvim_exec2('version', { output = true }).output
+  local nvim_version = version_output:match('NVIM v[^\n]+') or 'unknown'
+  local commit_hash = nvim_version:match('%+g?(%x+)')
+  local version_for_report = nvim_version
+  if commit_hash then
+    version_for_report = nvim_version:gsub('%+g' .. commit_hash, ' neovim/neovim@' .. commit_hash)
+    local result = vim
+      .system({ 'git', 'ls-remote', 'https://github.com/neovim/neovim', 'HEAD' }, { text = true })
```
justinmk: it should check `nightly`, otherwise it will always show "outdated" because HEAD almost always has new commits

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+      :wait()
+
+    if result.code == 0 then
+      local sha = result.stdout:match('^(%x+)')
+      local latest_commit = sha:sub(1, 10) --- @type string?
+      if latest_commit and commit_hash ~= latest_commit then
+        vim.health.warn(
+          string.format('Build is outdated. Local: %s, Latest: %s', commit_hash, latest_commit)
```
justinmk: this is a great idea for HEAD users. Since this is in `if commit_hash`, it won't show for stable users, which is good 👍

runtime/lua/vim/health/health.lua
```diff
@@ -493,11 +493,112 @@ local function check_external_tools()
...
+      '%#WarningMsg#%@v:lua._bugreport_open@Click to Create Bug Report on GitHub%X%*'
+  end)
+end
+
 function M.check()
   check_config()
   check_runtime()
   check_performance()
   check_rplugin_manifest()
+  check_sysinfo()
```
justinmk: should be the first one i guess. not randomly in the middle.

justinmk:
We will probably improve this but it's a good start. Some important ideas from https://github.com/neovim/neovim/issues/31847 which we should do in the future:

- let other plugins use this
- let healthchecks say which info should be included in the bugreport
    - especially don't want private info in the bugreport

runtime/lua/vim/health/health.lua
```diff
@@ -493,7 +493,153 @@ local function check_external_tools()
...
+      or nil
+
+    if cmd then
+      local result = vim.system(cmd, { text = true }):wait()
+      if result.code == 0 then
+        local sha = has_git and result.stdout:match('^(%x+)') or result.stdout
+        local latest_commit = sha and sha:sub(1, 10)
+        if latest_commit and commit_hash ~= latest_commit then
+          vim.health.warn(
+            string.format('Build is outdated. Local: %s, Latest: %s', commit_hash, latest_commit)
```
justinmk, line 555: Er... why does this assume that `commit_hash ~= latest_commit` means "outdated" ? :)

glepnir: build hash does not match the nightly hash, does that indicate the build is outdated?

justinmk: hashes can be anything. Note that `:version` shows `0.12.0-dev-1688-...`, where the `-1688` is a hint that I have 1688 commits since the last tag (`v0.11.x`). So if that is available, that could be compared. But then the logic would also need to count the commits in github.

I think this check should be skipped for now.

glepnir: Here it is matching the `g+commit hash` not commits

justinmk: Actually, since it's getting the latest nightly, you are right that it can be assumed the upstream commit is probably newer.

Maybe this is fine for now, although it will be wrong for core developers.

But also not sure if this is actually useful because anyone that didn't update in the last 24 hours will see an "outdated" message.

glepnir: Initially I compared against HEAD to ensure it was always checked against the latest commits, since some users' builds are outdated. This helps remind them to try the latest build before reporting an issue, as the problem may already have been fixed in a recent commit.

justinmk: Yes, but the message will almost always show "outdated", so people will stop noticing it. It will get ignored.

Counting commits would give a sense of how outdated they are.

glepnir: tried it. commits are based on `git describe --first-parent`. package management may be based on a different branch like  homebrew outputs `-dev-3779`.

runtime/lua/vim/health/health.lua
```diff
@@ -493,7 +493,153 @@ local function check_external_tools()
...
+  -- Use :version because vim.version().build returns "Homebrew" for brew installs.
+  local version_output = vim.api.nvim_exec2('version', { output = true }).output
+  local nvim_version = version_output:match('NVIM v[^\n]+') or 'unknown'
+  local commit_hash = nvim_version:match('%+g?(%x+)')
+  local version_for_report = nvim_version
+  if commit_hash then
+    version_for_report = nvim_version:gsub('%+g' .. commit_hash, ' neovim/neovim@' .. commit_hash)
+    local has_git = vim.fn.executable('git') == 0
+    local has_curl = vim.fn.executable('curl') == 1
+    local cmd = has_git and { 'git', 'ls-remote', 'https://github.com/neovim/neovim', 'nightly' }
```
clason, line 538: Not nice: For people building from HEAD, this prints a warning that the build is outdated (even though it's not).

glepnir: originally, comparing HEAD..but nowuse nightly with the source build compare  would trigger an “outdated” warning. I'm still experimenting with the commit count.

runtime/lua/vim/health/health.lua
glepnir: I’m confused. If a commit hash is detected, meaning it’s a source build or a nightly build, shouldn’t it be compared against HEAD to determine whether it’s outdated?

runtime/lua/vim/health/health.lua
clason: Yeah, you can't have it both ways. (And this will be even worse for stable users once 0.12 is released, who will constantly be taunted with being outdated -- pushing more people to use nightly even when they shouldn't.)

Also, remember that people use Neovim in airgapped systems -- will this trigger an error for them?

Personally, I think this is cute but not worth the complexity (and it'd need a lot more complexity to make it working robustly for all users).

clason: Sorry, I see now that this was already discussed earlier. Ignore me and concentrate on the other discussion.

runtime/lua/vim/health/health.lua
justinmk: Again, it will almost always show "outdated" so it's not useful. Counting commits probably also won't work for various reasons.

glepnir: emm not exactly .. run neovim which i build yesterday it show me outdated 
image

build with latest commit  there is no warning..

image

justinmk: > build with latest commit there is no warning

Yes. But every day there are new commits. So the warning isn't useful.

glepnir: Some package managers update  nightly builds  slowly, and some people build Neovim from older commits, so their installation can be outdated for a long time. In those cases, this warning is useful. Of course, if you think it’s unnecessary, we can remove this check later since there still need a refactor for plugin :)

justinmk: In those cases the warning will say "outdated" even more often than people building from source.

I build from source every day and this will still show "outdated" more often than not. People using slow nightly builds will just always see "outdated". That is not useful.

glepnir: I’ve thought about this before. I noticed that most PRs are merged during the night for people in certain time zones, so they rarely encounter merges during their daytime, which means the “outdated” warning doesn’t show up very often. There may not be a good solution for this. I just wanted a way to remind them to try building from the latest commit first and then verify whether the issue still exists, so  can avoid wasting the maintainers’ time 😮‍💨

justinmk: It definitely would be useful if we have a way to compare the age. And also if they are using an older stable release. We *should* show a warning if they are using 0.10 or 0.11.3, instead of the latest stable release.

glepnir: Check the release might be easier.

image

glepnir: https://github.com/neovim/neovim/pull/36715

PR #31846, opened 2025-01-03
refactor: adopt vterm

dundargoc:
We have changed too much to consider it a mere bundled dependency (such
as unicode handling in e3bfcf2fd4a4ebf00b104b082cfe83c8144a842d), and
can consider it our own at this point.


justinmk:
why not keep the dir name `vterm` ? `src/nvim/vterm/`

MAINTAIN.md
```diff
@@ -148,8 +148,6 @@ These dependencies are "vendored" (inlined), we must update the sources manually
 * `src/xdiff/`: [xdiff](https://github.com/git/git/tree/master/xdiff)
 * `src/cjson/`: [lua-cjson](https://github.com/openresty/lua-cjson)
 * `src/klib/`: [Klib](https://github.com/attractivechaos/klib)
-* `src/vterm/`: [libvterm](https://www.leonerd.org.uk/code/libvterm/),
```
justinmk, line 151: are we not planning to look at upstream ever?

dundargoc: Already discussed here: https://github.com/neovim/neovim/pull/30047#discussion_r1740455806

I can re-add it if you disagree with the thought process.

PR #31854, opened 2025-01-03
docs: cleanup news.txt

gpanders:
Move non-breaking changes out of the breaking changes section

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -40,7 +40,6 @@ These changes may require adaptations in your config or plugins.
 
 API
 
-• Improved API "meta" docstrings and :help documentation.
 • `vim.rpcnotify(0)` and `rpcnotify(0)` broadcast to ALL channels. Previously
   they would "multicast" only to subscribed channels (controlled by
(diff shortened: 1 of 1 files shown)


runtime/doc/news.txt
```diff
@@ -91,16 +85,6 @@ EVENTS
 
 • |vim.ui_attach()| callbacks for |ui-messages| `msg_show` events are executed in
   |api-fast| context.
-• New/enhanced arguments in these existing UI events:
-  • `cmdline_hide`: `abort` argument indicating if the cmdline was aborted.
-  • `cmdline_show`:
-    • Prompts that were previously emitted as `msg_show` events, are now routed
-      through `cmdline_show`.
```
luukvbaal, line 98: At least this pullet should stay under |news-breaking|. But entire section seems fine to stay for the extra notice; I think I also encounter GUIs that broke on added arguments to UI events (even though they should handle that gracefully).

justinmk (approved):

runtime/doc/news.txt
gpanders: I've moved these back.

PR #31865, opened 2025-01-04
fix(health): set modifiable=false in checkhealth buffers

MariaSolOs:
^^

### DIFF
--- a/runtime/lua/vim/health.lua
+++ b/runtime/lua/vim/health.lua
@@ -407,6 +407,9 @@ function M._check(mods, plugin_names)
   -- Clear the 'Running healthchecks...' message.
   vim.cmd.redraw()
   vim.print('')
+
+  -- Once we're done writing checks, set nomodifiable.
+  vim.bo[bufnr].modifiable = false
 end
 
 return M


justinmk:
SGTM. Tests need a tweak though.

I think we should also set `q` to quit in a :checkhealth buffer...

MariaSolOs:
> SGTM. Tests need a tweak though.

Yep, working on it.

> I think we should also set `q` to quit in a :checkhealth buffer...

Agreed, but I will add this in a separate PR since it's an unrelated feature to this change :)

dundargoc:
> I think we should also set q to quit in a :checkhealth buffer...

Didn't this turn out to be a massive headache with `:Man`? Also, I don't understand what is special about checkhealth buffers that a `q` should exit it (can't say I'm personally a fan of this behavior as things like `q/` won't work)

dundargoc:
Can't get this to work, I get

```
E5009: Invalid $VIMRUNTIME: /Users/dundargoc/programs/neovim/runtime
Error executing lua: Vim:E21: Cannot make changes, 'modifiable' is off
stack traceback:
        [C]: in function 'append'
        /Users/dundargoc/programs/neovim/runtime/lua/vim/health.lua:403: in function '_check'
        [string "<nvim>"]:1: in main chunk
```

The change itself seems good though.

dundargoc (approved):

MariaSolOs:
> Can't get this to work, I get
> 
> ```
> E5009: Invalid $VIMRUNTIME: /Users/dundargoc/programs/neovim/runtime
> Error executing lua: Vim:E21: Cannot make changes, 'modifiable' is off
> stack traceback:
>         [C]: in function 'append'
>         /Users/dundargoc/programs/neovim/runtime/lua/vim/health.lua:403: in function '_check'
>         [string "<nvim>"]:1: in main chunk
> ```
> 
> The change itself seems good though.

Ohhh, I guess that this is because we log to the buffer while running the checks. And so we should probably set `nomodifiable` after logging all checks.

justinmk (approved):

PR #31866, opened 2025-01-05
fix(cmdline): always show cmdline when it is a prompt

luukvbaal:
Cmdline prompts should ignore `cmd_silent`.


luukvbaal:
Ref https://github.com/neovim/neovim/pull/31525#issuecomment-2571342673

justinmk (approved):

Issue #31706, opened 2024-12-23
defaults: `q` should close floating windows


igorlfs:
### Problem

Within the plugin ecosystem, `q` (in normal mode) is a common mapping to close floating windows. It's also used within core, to close the LSP Hover window.

Usually, floating windows aren't modifiable, so having `q` mapped to recording a macro isn't all that useful.

### Expected behavior

`q` should be mapped to `:quit` for floats.

justinmk:
Why would it only be for floats? We could consider a default `q` mapping to do this. But that (or even the "floats only" case) is blocked until we think of a good way to opt-in to the normal purpose of `q` (start a macro).

igorlfs:
> Why would it only be for floats?

Mostly to not interfere with the behavior of starting a macro for regular windows. But I do agree it might sound odd to have different actions for the same key, depending on context.

wsdjeg:
> pt-in to the normal purpose

Please don't, I use `q` to start macro, and `q:` and `q?` to open cmdline windows.

justinmk:
What about `<M-q>` ?

telemachus:
> What about `<M-q>` ?

No particular preference other than leave `q` itself alone for macros since so many people are used to that and no point in changing from vim about something so basic.

justinmk:
# Proposal

We can define a `<Plug>(nvim-closewin)` , which will be globally available with the following behavior:

- Closes the current window.
    - Note: this is different than `ZB` [#2434](https://github.com/neovim/neovim/issues/2434) which closes current buffer without closing the window.
- Closes special-purpose buffers (in `:Man`, `:checkhealth`, vim.pack, etc.) even if it's the last window/tabpage.
- By default we map it to `<M-q>` and users can override it with a single line.

barrettruth:
@echasnovski as the developer of the great and superior mini.bufremove, are you interested in tackling this or have any insights?

EDIT: meant #2434...

echasnovski:
> @echasnovski as the developer of the great and superior mini.bufremove, are you interested in tackling this or have any insights?

I appreciate kind words, but 'mini.bufremove' is the opposite of what is described here: it is meant to delete/wipeout/hide a buffer *without* closing a window. I.e. similar to #2434 and other connected issues/solutions.

If anything, I think the separate `<Plug>(nvim-closewin)` is redundant because if users want custom map, `<Cmd>close<CR>` is pretty much the same.

PR #31870, opened 2025-01-05
feat(health): close checkhealth buffers with q

MariaSolOs:
As suggested in https://github.com/neovim/neovim/pull/31865#issuecomment-2571441200.

CC @dundargoc since it sounds like you're not a fan of this.


dundargoc:
I think this is *fine*, didn't mean to sound too negative. I won't make too much noise if this is merged. I just don't fully understand why pressing `q` for readonly buffers is expected/desired, and also why we're not doing this consistently. For example, shouldn't pressing `q` when using `:help` also quit the window by this logic?

runtime/lua/vim/health.lua
```diff
@@ -407,6 +407,15 @@ function M._check(mods, plugin_names)
   -- Clear the 'Running healthchecks...' message.
   vim.cmd.redraw()
   vim.print('')
+
+  -- Quit with 'q' inside healthcheck buffers.
+  vim.api.nvim_buf_set_keymap(
+    bufnr,
+    'n',
+    'q',
+    '<cmd>bdelete<cr>',
```
justinmk, line 416: I think `<c-w>q` is a good convention to continue. No need to close all other windows that are potentially showing this buffer. Plus it's better if `q` behaves consistently where possible.
```suggestion
    '<c-w>q',
```

justinmk:
> I just don't fully understand why pressing `q` for readonly buffers is expected/desired, and also why we're not doing this consistently.

This is partly to ease the transition for `:LspInfo` users. 

> shouldn't pressing `q` when using `:help` also quit the window by this logic?

Maybe, but there is 10x more friction there than for checkhealth. Long term, we probably want a default global mapping like `<m-q>` instead of making buffer-local decisions.

mfussenegger:
See also https://github.com/neovim/neovim/issues/31706
I think the same concerns regarding mapping `q` to `<c-w>q` that apply to floats apply here too.

So -1 from my side.
I also don't like the `<m-q>` idea. nvim so far doesn't use any `<m-` mappings and I reverse those for window manager keymaps and it would be annoying if nvim started to use that namespace too.

Not sure why `<c-w>q` is such a problem. It's easy enough to define a autocmd to setup `q` for floats or read-only buffers for those who really want to have that.

justinmk:
> nvim so far doesn't use any `<m-` mappings and I reverse those for window manager keymaps and it would be annoying if nvim started to use that namespace too.

Why not use super for window manager? I think using alt/meta for window-related default Nvim mappings is a good fit, and solves a lot of problems...

mfussenegger:
> Why not use super for window manager? I think using alt/meta for window-related default Nvim mappings is a good fit,

I actually thought `<m-...>` refers to super as I would have expected alt to be `<alt-q>`.  I'd be more open towards `<alt>` mappings but there's still potential for conflicts with window managers or the terminals and still doesn't explain why quit justified a _second_ keymap.

Aside from that, `<c-w>` is already established for window related mappings.

> and solves a lot of problems...

Which ones?

justinmk:
> > and solves a lot of problems...
> 
> Which ones?

There are 10+ useful navigation-related mappings that vim/nvim is missing OOTB. Most obvious is `<m-h>`/j/k/l, also mappings for "new window", "new tab", etc.

Off topic here though

dundargoc:
> This is partly to ease the transition for `:LspInfo` users.

I didn't know we held `:LspInfo` users in such high regard to dictate core design decisions ;). Would a simple map for the checkhealth buffer not suffice for the time being?

Tbh I think I'm more confused by this rather than against it, so I don't mind if it gets merged. But it does strike me as odd as I don't really understand the reasoning or what problem is being solved. Is this just about saving time by using `q` instead of `:q`? Or is this about this being a better default or more intuitive somehow?

justinmk:
> Would a simple map for the checkhealth buffer not suffice

that is what this pr is

really doesn't need so much debate. we own checkhealth and making it fancy is part of the plan.

if we decide a more global/universal approach for "quit" later, we can revert this.

mfussenegger:
> > Would a simple map for the checkhealth buffer not suffice
> 
> that is what this pr is
> 
> really doesn't need so much debate. we own checkhealth and making it fancy is part of the plan.
> 
> if we decide a more global/universal approach for "quit" later, we can revert this.

Having inconsistent `q` behaviour with no discernible pattern for users other than remembering what special buffer maps it doesn't strike me as improvement or as fancy but rather looks like a usability issue to me.

Anyway, go ahead if you must. I won't comment on this further

dundargoc:
> that is what this pr is

Breh, I meant as a code snippet for those users who wanted this.

Anyway, I don't see the "vision" yet but it's true that this a two-way door so I guess we don't have to be overly cautious. For reference, what would be the best way for users to disable this behavior?

MariaSolOs:
Despite being the author, I don't have strong feelings about merging this or not.

However I do agree that I would dislike using a `<m-...>` mapping here. As others have said I already use alt mappings for outside-nvim-navigation stuff.

justinmk:
> Having inconsistent `q` behaviour with no discernible pattern for users other than remembering what special buffer maps it doesn't strike me as improvement or as fancy but rather looks like a usability issue to me.

That's true, but we don't have a better answer currently.

The rationale is that `:che` is positioned as a replacement for `:LspInfo` (and potentially similar use-cases), where a brief "current status" is wanted. So a way to quickly dismiss that "current status" is needed.

Although `q` has unfortunate inconsistencies, it is very common in the plugin ecosystem, including tpope's plugins (until pivoting to `gq`, but that's a tangent).

justinmk:
I will document this in my docs pr

clason:
Pressing `q` on a `:check` window closes Nvim for me -- that can't be intended?

justinmk:
will fix

PR #31874, opened 2025-01-05
fix(filetype): use .slang extension for slang instead of .sl

mTvare6:
The extension used by slang is .slang.
[Examples] and [docs] for reference.

[Examples]: https://github.com/shader-slang/slang/tree/master/examples
[docs]: https://shader-slang.com/slang/user-guide/get-started.html


clason:
Thank you, but all filetype detection changes must go through vim/vim.

PR #31879, opened 2025-01-06
fix(snippet): remove stale snippet keymap examples

MariaSolOs:
After https://github.com/neovim/neovim/pull/27339/commits/e6cfcaed184d4ecdc8a8638429e1bd9e1b3251dc, these examples are misleading since these keymaps are set by default today.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -4527,16 +4527,6 @@ vim.snippet.active({filter})                            *vim.snippet.active()*
     Returns `true` if there's an active snippet in the current buffer,
     applying the given filter if provided.
 
-    You can use this function to navigate a snippet as follows: >lua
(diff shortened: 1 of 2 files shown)


justinmk:
can we reword the example like "to change the default" or maybe an example that shows how to remove the default (or is that not possible yet?)

MariaSolOs:
> can we reword the example like "to change the default" or maybe an example that shows how to remove the default (or is that not possible yet?)

As of today changing/removing this default isn’t possible, but #31884 would change that.

clason:
@MariaSolOs can you rebase and check against the changed default keymap settings? (Or close if it's no longer applicable?)

MariaSolOs:
@clason closing as the example was removed from `vim.snippet.active` and the one in `vim.snippet.jump` has been updated to reflect the new behavior.

PR #31880, opened 2025-01-06
feat(startup): distinct exitcodes for config/-c vs -l error

justinmk:
followup to #21723

# Problem

- Exit code is always 1 for all kinds of failures.
- `nvim -l` exits with success even if there is a config/-c error:
  ```
  echo 'print(vim.v.servername)' | nvim -c "throw 'foo'" -l -
  ```

# Solution

For `nvim -l` set a different exit code for config errors (`-u`, `-c`) vs `-l` (Lua code) errors.

### DIFF
--- a/runtime/doc/starting.txt
+++ b/runtime/doc/starting.txt
@@ -193,10 +193,15 @@ argument.
                 disables most prompts and messages. Unrelated to |-s|.
                 See also |-S| to run script files.
 
+                Sets exit code as follows:
+                • 0 on success
+                • 1 on config error (including -u/-c/--cmd)
+                • 2 on -l script error
+
                 -es reads/executes stdin as Ex commands. >
                         printf "put ='foo'\n%%print\n" | nvim -es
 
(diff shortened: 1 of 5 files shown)


runtime/doc/starting.txt
```diff
@@ -193,10 +193,15 @@ argument.
...
+                Sets exit code as follows:
+                • 0 on success
+                • 1 on config error (including -u/-c/--cmd)
+                • 2 on -l script error
+
                 -es reads/executes stdin as Ex commands. >
                         printf "put ='foo'\n%%print\n" | nvim -es
 
-<                -Es reads stdin as text (into buffer 1).  Use |-c| or "+" to
+<               -Es reads stdin as text (into buffer 1).  Use |-c| or "+" to
```
zeertzjq, line 204: I think the `<` should be moved to the previous line.

runtime/doc/starting.txt
zeertzjq, line 208: This one too.

PR #31884, opened 2025-01-06
feat(snippet): add navigation_mapping completion option

MariaSolOs:
Possible solution to https://github.com/neovim/neovim/issues/30198.

### DIFF
--- a/runtime/lua/vim/snippet.lua
+++ b/runtime/lua/vim/snippet.lua
@@ -2,8 +2,6 @@ local G = vim.lsp._snippet_grammar
 local snippet_group = vim.api.nvim_create_augroup('vim/snippet', {})
 local snippet_ns = vim.api.nvim_create_namespace('vim/snippet')
 local hl_group = 'SnippetTabstop'
-local jump_forward_key = '<tab>'
-local jump_backward_key = '<s-tab>'
 
 --- Returns the 0-based cursor position.
 ---
@@ -179,13 +177,16 @@ function Tabstop:set_right_gravity(right_gravity)
   })
 end
 
+--- @alias vim.snippet.NavigationMappings { jump_forward: string, jump_backward: string }
+
 --- @class (private) vim.snippet.Session
 --- @field bufnr integer
 --- @field extmark_id integer
 --- @field tabstops table<integer, vim.snippet.Tabstop[]>
 --- @field current_tabstop vim.snippet.Tabstop
---- @field tab_keymaps { i: table<string, any>?, s: table<string, any>? }
---- @field shift_tab_keymaps { i: table<string, any>?, s: table<string, any>? }
+--- @field navigation_mappings vim.snippet.NavigationMappings?
+--- @field jump_forward_keymaps { i: table<string, any>?, s: table<string, any>? }
+--- @field jump_backward_keymaps { i: table<string, any>?, s: table<string, any>? }
 local Session = {}
 
 --- Creates a new snippet session in the current buffer.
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/snippet.lua
```diff
@@ -194,13 +195,15 @@ local Session = {}
...
+--- @param navigation_mappings vim.snippet.NavigationMappings?
 --- @return vim.snippet.Session
-function Session.new(bufnr, snippet_extmark, tabstop_data)
+function Session.new(bufnr, snippet_extmark, tabstop_data, navigation_mappings)
   local self = setmetatable({
     bufnr = bufnr,
     extmark_id = snippet_extmark,
     tabstops = {},
     current_tabstop = Tabstop.new(0, bufnr, { 0, 0, 0, 0 }),
+    navigation_mappings = navigation_mappings,
```
justinmk, line 206:
```suggestion
    mappings = mappings,
```

runtime/lua/vim/snippet.lua
```diff
@@ -179,13 +177,16 @@ function Tabstop:set_right_gravity(right_gravity)
   })
 end
 
+--- @alias vim.snippet.NavigationMappings { jump_forward: string, jump_backward: string }
```
justinmk, line 180: could leave the door open for other kinds of mappings in the future?

also given `vim.keymap` maybe we should stick to "keymap" 

```suggestion
--- @alias vim.snippet.Keymap { jump_forward: string, jump_backward: string }
```

echasnovski:
A note: allowing configurable mappings on `vim.lsp.completion` level and not on `vim.snippet` level places plugins/users who directly use `vim.snippet` at a disadvantage. That means that any plugin wanting to use `vim.snippet` directly (which is an eventual `vim.snippet` design and is already a reality for several popular plugins) will have no good way of preventing `<Tab>` / `<S-Tab>` being forced after snippet expansion.

clason:
>  will have no good way of preventing <Tab> / <S-Tab> being forced after snippet expansion.

Wat. The whole point of this is to _set_ these mappings ONLY in a specific context (and opt-in). Plugins can do whatever they want, and instruct users _not_ to opt-in if they handle this themselves.

mfussenegger:
> A note: allowing configurable mappings on `vim.lsp.completion` level and not on `vim.snippet` level places plugins/users who directly use `vim.snippet` at a disadvantage. That means that any plugin wanting to use `vim.snippet` directly (which is an eventual `vim.snippet` design and is already a reality for several popular plugins) will have no good way of preventing `<Tab>` / `<S-Tab>` being forced after snippet expansion.

To extend on this, we'd also need a configuration option in core once we add support for LSP SnippetTextEdit - which can be triggered via code actions, code lens or inlay hints, etc. 

To me this seems like one of the worst compromises to tie this to lsp.completion.

echasnovski:
> > will have no good way of preventing  /  being forced after snippet expansion.
> 
> Wat. The whole point of this is to _set_ these mappings ONLY in a specific context (and opt-in). Plugins can do whatever they want, and instruct users _not_ to opt-in if they handle this themselves.

I did miss the `if not self.navigation_mappings then return end` part in `vim.snippet`, which indeed makes for a better scoped usage of "forced" mappings.

Yet it still looks like having `vim.lsp.completion.enable()` will affect possibly completely unrelated usage of `vim.snippet.expand()` (with potentially different use cases). Due to [this monkey patching](https://github.com/neovim/neovim/blob/aa9c38e04227fa52bda1efed532b0e5e3b91fe73/runtime/lua/vim/lsp/completion.lua#L663).

> We heard your mini.snippet desires, but frankly that's not enough to base a core design on.

@clason, please get your facts straight before making this kind of personal comments, okay? #30198 doesn't mention anything about 'mini.snippets' but only common `<Tab>` and `<S-Tab>` Insert mode mappings. 'mini.snippets' is already a thing with its way of allowing users to customize their mapping as they want to, which should be the case from the start.

runtime/lua/vim/snippet.lua
MariaSolOs: Sure, but we should keep the plural.

MariaSolOs:
> > A note: allowing configurable mappings on `vim.lsp.completion` level and not on `vim.snippet` level places plugins/users who directly use `vim.snippet` at a disadvantage. That means that any plugin wanting to use `vim.snippet` directly (which is an eventual `vim.snippet` design and is already a reality for several popular plugins) will have no good way of preventing `<Tab>` / `<S-Tab>` being forced after snippet expansion.
> 
> To extend on this, we'd also need a configuration option in core once we add support for LSP SnippetTextEdit - which can be triggered via code actions, code lens or inlay hints, etc.
> 
> To me this seems like one of the worst compromises to tie this to lsp.completion.
> 
> Alternative proposal: #31887

FWIW I do like your proposal better :)

MariaSolOs:
Closing this in favour of #31887

PR #31887, opened 2025-01-06
feat(snippet): set snippet keymaps permanent instead of dynamic

mfussenegger:
Problem:

Given that `vim.snippet.expand()` sets temporary `<tab>`/`<s-tab>`
keymaps there is no way to build "smart-tab" functionality where `<tab>`
chooses the next completion candidate if the popup menu is visible.

Solution:

Set the keymap permanent in `_defaults`.

The downside of this approach is that users of multiple snippet engine's
need to adapt their keymaps to handle all their engines that are in use.
For example:

    vim.keymap.set({ 'i', 's' }, "<Tab>", function()
      if foreign_snippet.active() then
        return "<Cmd>lua require('foreign_snippet').jump()<CR>"
      elseif vim.snippet.active({ direction = 1 }) then
        return "<Cmd>lua vim.snippet.jump(1)<CR>"
      else
        return key
      end
    end, { expr = true })

Upside is that using `vim.keymap.set` to override keymaps is a well
established pattern - no need for an extra snippet.config keymap DSL - and `vim.snippet.expand` calls made by nvim itself
or plugins have working keymaps out of the box.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -4551,16 +4551,6 @@ vim.snippet.active({filter})                            *vim.snippet.active()*
     Returns `true` if there's an active snippet in the current buffer,
     applying the given filter if provided.
 
-    You can use this function to navigate a snippet as follows: >lua
-        vim.keymap.set({ 'i', 's' }, '<Tab>', function()
-           if vim.snippet.active({ direction = 1 }) then
-             return '<Cmd>lua vim.snippet.jump(1)<CR>'
-           else
-             return '<Tab>'
-           end
-         end, { expr = true })
-<
-
     Parameters: ~
       • {filter}  (`vim.snippet.ActiveFilter?`) Filter to constrain the search
                   with:
@@ -4585,14 +4575,15 @@ vim.snippet.jump({direction})                             *vim.snippet.jump()*
     Jumps to the next (or previous) placeholder in the current snippet, if
     possible.
 
-    For example, map `<Tab>` to jump while a snippet is active: >lua
+    By default `<Tab>` is setup to jump if a snippet is active. The default
+    mapping looks like: >lua
         vim.keymap.set({ 'i', 's' }, '<Tab>', function()
            if vim.snippet.active({ direction = 1 }) then
              return '<Cmd>lua vim.snippet.jump(1)<CR>'
            else
              return '<Tab>'
            end
-         end, { expr = true })
+         end, { descr = '...', expr = true, silent = true })
 <
 
     Parameters: ~
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -218,6 +218,27 @@ do
     end, { desc = 'vim.lsp.buf.signature_help()' })
   end
 
+  do
+    ---@param direction vim.snippet.Direction
+    ---@param key string
+    local function set_snippet_jump(direction, key)
+      vim.keymap.set({ 'i', 's' }, key, function()
+        if vim.snippet.active({ direction = direction }) then
+          return string.format('<Cmd>lua vim.snippet.jump(%d)<CR>', direction)
+        else
+          return key
+        end
+      end, {
+        desc = 'vim.snippet.jump if active, otherwise ' .. key,
+        expr = true,
+        silent = true,
+      })
+    end
+
+    set_snippet_jump(1, '<Tab>')
+    set_snippet_jump(-1, '<S-Tab>')
+  end
+
   --- Map [d and ]d to move to the previous/next diagnostic. Map <C-W>d to open a floating window
   --- for the diagnostic under the cursor.
   ---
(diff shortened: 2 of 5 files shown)


clason:
That won't work, since `<tab>` is used to expand a trigger -- i.e., _before_ a "foreign snippet" session is active.

mfussenegger:
> That won't work, since `<tab>` is used to expand a trigger -- i.e., _before_ a "foreign snippet" session is active.

I don't follow. `<tab>` here would only jump if a vim.snippet session is already active. It won't expand snippets. 

If you setup `<tab>` to expand or jump with either vim.snippet or a foreign snippet engine then you already customised it and can account for both engines, no? 

I thought I understood your use case about using two engines in the same buffer at different times but looks like I'm missing something.

clason:
The point is that this configuration is not as trivial as your code example indicates. My best effort so far is a monstrosity, and I was hoping we wouldn't have to force this boilerplate on other users.

mfussenegger:
> The point is that this configuration is not as trivial as your code example indicates.

Can you share your configuration or elaborate on how exactly you currently use vim.snippet + luasnip?

justinmk (approved):

runtime/lua/vim/_defaults.lua
```diff
@@ -184,6 +184,23 @@ do
...
+    ---@param direction vim.snippet.Direction
+    ---@param key string
+    local function set_snippet_jump(direction, key)
+      vim.keymap.set({ 'i', 's' }, key, function()
+        if vim.snippet.active({ direction = direction }) then
+          return string.format('<Cmd>lua vim.snippet.jump(%d)<CR>', direction)
+        else
+          return key
+        end
+      end, { desc = 'vim.snippet.jump if active, otherwise ' .. key, expr = true })
```
MariaSolOs, line 197:
```suggestion
      end, { desc = 'vim.snippet.jump if active, otherwise ' .. key, expr = true, silent = true })
```

MariaSolOs (approved):

MariaSolOs:
@mfussenegger Could we also piggy-back the changes from https://github.com/neovim/neovim/pull/31879 in this PR and document on how to remove/override these default mappings?

runtime/doc/lua.txt
```diff
@@ -4551,16 +4551,6 @@ vim.snippet.active({filter})                            *vim.snippet.active()*
     Returns `true` if there's an active snippet in the current buffer,
     applying the given filter if provided.
 
-    You can use this function to navigate a snippet as follows: >lua
-        vim.keymap.set({ 'i', 's' }, '<Tab>', function()
```
justinmk, line 4555: instead of dropping this , worth updating it with the example from your PR description?

justinmk: oh, there's already an example below at `:help vim.snippet.jump()` 👍

justinmk (approved):
🚀 🚀 🚀

test/functional/lua/snippet_spec.lua
```diff
@@ -18,6 +18,20 @@ local retry = t.retry
...
     clear()
+    exec_lua(function()
+      local function set_snippet_jump(direction, key)
+        vim.keymap.set({ 'i', 's' }, key, function()
+          if vim.snippet.active({ direction = direction }) then
+            return string.format('<Cmd>lua vim.snippet.jump(%d)<CR>', direction)
+          else
+            return key
+          end
+        end, { desc = 'vim.snippet.jump if active, otherwise ' .. key, expr = true })
```
MariaSolOs, line 29:
```suggestion
        end, { silent = true, expr = true })
```

For test purposes the description is irrelevant, but I think we should set `silent` and `expr` like the default mapping does.

MariaSolOs (approved):

tokisuno:
just a question - does this mean once this is merged, i can disable the super-tab functionality?

michaelfortunato:
Hi all, I am on nightly and tried to do this, but it seems that on certain start ups  neovim will sometimes respect this configuration and then on other startup sessions will apply the original default keymap. Has anyone else encountered this? I can elaborate. 
```lua
del({"i","s"} "<Tab>")

map({ "i", "s" }, "<Tab>", function()
  local pos = vim.fn.getpos(".")
  local line = vim.fn.getline(".")

  -- If at the beginning of the line, you usually just want to insert
  -- a tab
  if pos[3] == 0 or string.sub(line, 0, pos[3]):match("^%s+$") then
       return "<Tab>"
  end

  local luasnip = require("luasnip")
  if luasnip.locally_jumpable(1) then
    require("blink.cmp").snippet_forward()
    return
  else
    return "<Tab>"
  end
end, { noremap = true, expr = true, silent = true })
```

For instance if I nvim <file> a terminal the first time, then neovim seems to respect the above changes. But then if I quit neovim, and run nvim <file> again it no longer does. It's almost exactly working 50% of the time.

Issue #31871, opened 2025-01-05
deprecate nvim_out_write/nvim_err_write/nvim_err_writeln


przepompownia:
### Problem

For example, for 
```lua
vim.api.nvim_err_writeln('1\n2\n3')
```
- `\n` in the input message are used to split the output into separate messages
- each such `\n` is passed as an additional message (respecting the order of appearing)
- the logic of assigning `kind` value (empty or `emsg`) is unclear for me

```lua
{
  c = { { 24, "1", 6 } },
  k = "emsg"
}
{
  c = { { 0, "\n", 0 } },
  k = "emsg"
}
{
  c = { { 24, "2", 6 } },
  k = ""
}
{
  c = { { 0, "\n", 0 } },
  k = "emsg"
}
{
  c = { { 24, "3", 6 }, { 0, "\n", 0 } },
  k = ""
}
```

This makes it difficult to display such ~messages~ message correctly.

### Steps to reproduce

`init.lua`:
```lua
local api = vim.api
local ns = api.nvim_create_namespace('messageRedirection')

local function log(msg)
  api.nvim_buf_set_lines(1, -1, -1, true, vim.split(msg, '\n'))
  vim.cmd.normal 'G'
end

local function detach()
  api.nvim__redraw({flush = true})
  vim.ui_detach(ns)
  api.nvim__redraw({flush = true})
end

api.nvim_create_autocmd('CmdlineEnter', {callback = detach})

local qMsgs = {}

local function consumeQMsgs()
  for _, qmsg in ipairs(qMsgs) do
    log(vim.inspect(qmsg))
  end
  qMsgs = {}
end

local redirect = true

local function attach()
  if not redirect then
    return
  end
  api.nvim__redraw({flush = true})
  vim.ui_attach(ns, {ext_messages = true, ext_cmdline = false}, function (event, kind, content, replace)
    if event ~= 'msg_show' then
      return
    end
    if kind == 'return_prompt' then
      api.nvim_input('\r')
      return
    end

    local msg = {
      k = kind or 'none',
      c = content or 'none',
    }
    if not vim.in_fast_event() then
      consumeQMsgs()
      log(vim.inspect(msg))
      return
    end

    qMsgs[#qMsgs + 1] = msg
    vim.schedule(function ()
      consumeQMsgs()
      api.nvim__redraw({flush = true})
    end)
  end)
end

api.nvim_create_autocmd({'CmdlineLeave'}, {callback = attach})

[...]

PR #31895, opened 2025-01-07
fix(api): replace `nvim_out_write`, `nvim_err_writeln` usages by `nvim_echo`

przepompownia:
It's a step for #31871

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -224,7 +224,7 @@ do
     local function cmd(opts)
       local ok, err = pcall(vim.api.nvim_cmd, opts, {})
       if not ok then
-        vim.api.nvim_err_writeln(err:sub(#'Vim:' + 1))
+        vim.api.nvim_echo({ { err:sub(#'Vim:' + 1), 'ErrorMsg' } }, true, {})
       end
     end
 
--- a/runtime/lua/vim/_editor.lua
+++ b/runtime/lua/vim/_editor.lua
@@ -621,7 +621,7 @@ end
 ---@diagnostic disable-next-line: unused-local
 function vim.notify(msg, level, opts) -- luacheck: no unused args
   if level == vim.log.levels.ERROR then
-    vim.api.nvim_err_writeln(msg)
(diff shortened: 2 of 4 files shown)


justinmk (approved):

przepompownia:
Please change the commit title to something more appropriate (no idea what to put into parentheses after `fix`).

MariaSolOs:
For my own learning: What's the reasoning behind this change?

glepnir:
> Please change the commit title to something more appropriate (no idea what to put into parentheses after `fix`, `api` for example).

git commit --amend then edit the commist msg and git push --force

luukvbaal:
Might need to add `opts->error` to `nvim_echo()` that calls `emsg_multiline()` to make this an exact replacement.

przepompownia:
> Might need to add `opts->error` to `nvim_echo()` that calls `emsg_multiline()` to make this an exact replacement.

The only allowed opts key is `verbose`. Do you mean the fourth, separate parameter? It's probably unavailable from Lua API.

https://github.com/neovim/neovim/blob/d8bc08db7fd8d0efbf2e9ebf39fecb6f732f84e8/src/nvim/api/vim.c#L771-L781

In `build/src/nvim/auto/lua_api_c_bindings.generated.h` `err` I see it is set internally.

przepompownia:
> > Please change the commit title to something more appropriate (no idea what to put into parentheses after `fix`, `api` for example).
> 
> git commit --amend then edit the commist msg and git push --force

Thanks, I know - I meant the correctness of the content, not the way of change.

luukvbaal:
> The only allowed opts key is verbose.

Yes, that is the problem. I'm saying that before we deprecate `nvim_out_write(ln)`, we should consider extending `nvim_echo` to support an `opts->error` argument if that is to be it's replacement.

luukvbaal:
> we are simulating :emsg

Actually, we are simulating `:echoerr` and so perhaps `opts->err` is more appropriate? `emsg()` is an internal function.

justinmk:
Thanks for getting this started. Looks like https://github.com/neovim/neovim/pull/31900 will address this.

PR #31900, opened 2025-01-07
feat(api): add err field to nvim_echo() opts

### LINKED ISSUE #31871: deprecate nvim_out_write/nvim_err_write/nvim_err_writeln

luukvbaal:
**feat(api): add err field to nvim_echo() opts**

Problem:  We want to deprecate `nvim_err_write(ln)()` but there is no
          obvious replacement. Meanwhile we already have `nvim_echo()`
          with an `opts` argument.
Solution: Add `err` argument to `nvim_echo()` that directly maps to
          `:echoerr`.

**feat(api): deprecate nvim_out/err_write(ln)**

Resolve #31871

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -662,6 +662,8 @@ nvim_echo({chunks}, {history}, {opts})                           *nvim_echo()*
                    `hl_group` element can be omitted for no highlight.
       • {history}  if true, add to |message-history|.
       • {opts}     Optional parameters.
+                   • err: Treat the message like |:echoerr|. Omitted `hlgroup`
+                     uses |hl-ErrorMsg| instead.
                    • verbose: Message is printed as a result of 'verbose'
                      option. If Nvim was invoked with -V3log_file, the message
                      will be redirected to the log_file and suppressed from
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -669,23 +669,6 @@ nvim_echo({chunks}, {history}, {opts})                           *nvim_echo()*
                      will be redirected to the log_file and suppressed from
                      direct output.
 
-nvim_err_write({str})                                       *nvim_err_write()*
-    Writes a message to the Vim error buffer. Does not append "\n", the
-    message is buffered (won't display) until a linefeed is written.
-
-    Parameters: ~
-      • {str}  Message
-
-nvim_err_writeln({str})                                   *nvim_err_writeln()*
-    Writes a message to the Vim error buffer. Appends "\n", so the buffer is
-    flushed (and displayed).
-
-    Parameters: ~
-      • {str}  Message
-
-    See also: ~
-      • nvim_err_write()
-
 nvim_eval_statusline({str}, {opts})                   *nvim_eval_statusline()*
     Evaluates statusline string.
 
@@ -1155,13 +1138,6 @@ nvim_open_term({buffer}, {opts})                            *nvim_open_term()*
 
     Return: ~
         Channel id, or 0 on error
-
-nvim_out_write({str})                                       *nvim_out_write()*
-    Writes a message to the Vim output buffer. Does not append "\n", the
-    message is buffered (won't display) until a linefeed is written.
-
-    Parameters: ~
-      • {str}  Message
 
 nvim_paste({data}, {crlf}, {phase})                             *nvim_paste()*
     Pastes at cursor (in any mode), and sets "redo" so dot (|.|) will repeat
--- a/src/nvim/api/vim.c
+++ b/src/nvim/api/vim.c
@@ -86,8 +86,6 @@
 #include "nvim/vim_defs.h"
 #include "nvim/window.h"
 
-#define LINE_BUFFER_MIN_SIZE 4096
-
 #ifdef INCLUDE_GENERATED_DECLARATIONS
 # include "api/vim.c.generated.h"
 #endif
@@ -806,37 +804,6 @@ void nvim_echo(Array chunks, Boolean history, Dict(echo_opts) *opts, Error *err)
 
 error:
   hl_msg_free(hl_msg);
-}
-
-/// Writes a message to the Vim output buffer. Does not append "\n", the
-/// message is buffered (won't display) until a linefeed is written.
-///
-/// @param str Message
-void nvim_out_write(String str)
-  FUNC_API_SINCE(1)
-{
-  write_msg(str, false, false);
-}
-
-/// Writes a message to the Vim error buffer. Does not append "\n", the
-/// message is buffered (won't display) until a linefeed is written.
-///
-/// @param str Message
-void nvim_err_write(String str)
-  FUNC_API_SINCE(1)
-{
-  write_msg(str, true, false);
-}
-
-/// Writes a message to the Vim error buffer. Appends "\n", so the buffer is
-/// flushed (and displayed).
-///
-/// @param str Message
-/// @see nvim_err_write()
-void nvim_err_writeln(String str)
-  FUNC_API_SINCE(1)
-{
-  write_msg(str, true, true);
 }
 
 /// Gets the current list of buffer handles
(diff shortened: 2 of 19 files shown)


src/nvim/api/vim.c
```diff
@@ -790,7 +793,11 @@ void nvim_echo(Array chunks, Boolean history, Dict(echo_opts) *opts, Error *err)
     verbose_enter();
   }
 
-  msg_multihl(hl_msg, history ? "echomsg" : "echo", history);
+  if (opts->err) {
+    emsg_multiline(chunks.items[0].data.array.items[0].data.string.data, true);
```
luukvbaal, line 797: Should we error when `chunks.size > 1` or `history == false` or just silently execute the documented behavior?

justinmk: Why not print all chunks?

src/nvim/api/vim.c
```diff
@@ -790,14 +791,18 @@ void nvim_echo(Array chunks, Boolean history, Dict(echo_opts) *opts, Error *err)
...
+  }
 
   if (opts->verbose) {
     verbose_leave();
     verbose_stop();  // flush now
   }
 
-  if (history) {
+  if (history && !opts->err) {
     // history takes ownership
```
justinmk, line 806: is that the case even when "echo" is used on line 797?

justinmk (approved):
Fantastic

src/nvim/api/vim.c
luukvbaal: `nvim_echo` supports multiple chunks to give a different highlight to different chunks. For `opts->err` I think it makes sense to implicitly highlight with `ErrorMsg`. I suppose we could still support multiple chunks though, even if that doesn't serve any purpose for the caller to pass in multiple chunks?

src/nvim/api/vim.c
luukvbaal: "echo" is used when `history == false`, so then hint in this comment doesn't apply either? Not sure what you mean, but yes when `history == false` `msg_multihl()` doesn't take ownership, and instead we free at the end of `nvim_echo()`.

src/nvim/api/vim.c
luukvbaal: > If people want to send chunks with random highlights, we might as well let them

With the connotation that those highlight's aren't actually applied? Wouldn't it be better to error in that case?

Or are you saying that `opts->err` with different highlight chunks _should_ apply those highlights? In that case I suppose we would need to add a `hl_id` argument to `emsg_multiline`.

justinmk: > In that case I suppose we would need to add a `hl_id` argument to `emsg_multiline`.

Oh, let's not bother with that. For now we'll just document that chunk highlights are ignored. Or we could throw an error.

But at least supporting multiple chunks seems harmless?

luukvbaal: Harmless yes, but also pointless. Indicating to me that an error might be more appropriate. Though supporting different highlight chunks even for `opts->err` does sound the least surprising/easiest to document to me. I don't think the change would be too involved.

justinmk: > I don't think the change required would be too involved.

SGTM

runtime/doc/api.txt
```diff
@@ -662,28 +662,13 @@ nvim_echo({chunks}, {history}, {opts})                           *nvim_echo()*
                    `hl_group` element can be omitted for no highlight.
       • {history}  if true, add to |message-history|.
       • {opts}     Optional parameters.
+                   • err: Print an error message for which `history` is always
+                     true and omitting `hl_group` uses |hl-ErrorMsg|.
```
justinmk, line 666: `history` is a required arg so it's strange to ignore it. Is the `history is always true` behavior a limitation of the internals? If not, it seems better to just let the `history` arg decide this.

luukvbaal: >  Is the history is always true behavior a limitation of the internals

Yes, that limitation could be lifted by also passing along `history`; so I guess we should do that?

src/nvim/api/vim.c
```diff
@@ -790,14 +790,14 @@ void nvim_echo(Array chunks, Boolean history, Dict(echo_opts) *opts, Error *err)
     verbose_enter();
   }
 
-  msg_multihl(hl_msg, history ? "echomsg" : "echo", history);
+  msg_multihl(hl_msg, opts->err ? "echoerr" : history ? "echomsg" : "echo", history, opts->err);
```
justinmk, line 793: I guess this is "echoerr" instead of "emsg" because this is a "user generated" error, not an exception, etc.

src/nvim/message.c
```diff
@@ -302,9 +302,13 @@ void msg_multihl(HlMessage hl_msg, const char *kind, bool history)
   msg_ext_set_kind(kind);
   for (uint32_t i = 0; i < kv_size(hl_msg); i++) {
     HlMessageChunk chunk = kv_A(hl_msg, i);
-    msg_multiline(chunk.text, chunk.hl_id, true, false, &need_clear);
+    if (err) {
+      emsg_multiline(chunk.text.data, chunk.hl_id, true);
```
justinmk, line 306: This makes me think the `kind` should be "emsg" ?

runtime/doc/api.txt
justinmk: I don't see why not. Maybe it raises the question of "why are we bothering to add `err` flag"? I assume it still has a purpose :) Maybe this doc isn't fully communicating what its purpose is.

src/nvim/message.c
luukvbaal: IDK; see my comment above and `eval.c:7972` for the `:echoerr` behavior. We are trying to mirror that here.

src/nvim/api/vim.c
justinmk: `echoerr` to me implies that it was produced by `:echoerr` (and perhaps also by this new flag)

runtime/doc/api.txt
luukvbaal: > Maybe it raises the question of "why are we bothering to add err flag"? I assume it still has a purpose

Yes, the thought crossed my mind. There is still some logic in `emsg_multiline` that applies to "emitting a message as an error". If nothing else, the message kind is appropriately conveyed to ext_messages. Not sure if all the other logic only applies to Vimscript.

src/nvim/message.c
justinmk: ok I see that :echoerr uses the same call: https://github.com/neovim/neovim/blob/17b46d01e29443452ae8b607017f8f5c585d3f0a/src/nvim/eval.c#L7968-L7972

runtime/doc/api.txt
justinmk: Sounds like it has a purpose then.

src/nvim/api/vim.c
luukvbaal: > assuming that this new err flag actually plumbs through the same codepaths as :echoerr

It does. And to me the fact that non-`opts->err` invocations we're already emitted with the `echo(msg)` kind, using `echoerr` for `opts->err` seems natural.

test/functional/api/vim_spec.lua
```diff
@@ -3680,6 +3680,18 @@ describe('API', function()
...
+    it('can print error message', function()
+      async_meths.nvim_echo({ { 'Error' }, { 'Message', 'Special' } }, false, { err = true })
+      screen:expect([[
+                                                |
+        {1:~                                       }|*3
+        {3:                                        }|
+        {9:Error}                                   |
+        {16:Message}                                 |
+        {6:Press ENTER or type command to continue}^ |
+      ]])
```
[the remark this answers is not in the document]
luukvbaal: Yeah it would, easier to add it to `messages_spec.lua`. Which I did, and that revealed some more work was needed to avoid emitting multiple messages (with unset `kind`/`history`) for a single multihl message (because `msg_keep()` logic interferes with that).

src/nvim/message.c
```diff
@@ -618,6 +633,9 @@ void msg_source(int hl_id)
     msg_scroll = true;  // this will take more than one line
     msg(p, hl_id);
     xfree(p);
+    if (is_multihl) {
+      msg_start();  // avoided in msg_keep() but necessary here
```
luukvbaal, line 637: This is a bit questionable since this goes against the point of the `is_multihl` guards, but it currently resolves a failing test related to newline formatting. `msg_start()` emits ext_messages (which we don't want for a single multihl message), while also taking care of formatting the message grid. Once the message grid is removed I want to re-evaluate whether `msg_start/end()` are necessary at all. Currently it results unnecessary message chunk splitting in various places (also seen in https://github.com/neovim/neovim/issues/31901).

justinmk (approved):

PR #31905, opened 2025-01-07
fix(highlight): make `TablineSel` more noticeable with `notermguicolors`

echasnovski:
Problem: Linking `TablineSel` to `Normal` makes it more noticeable with `notermguicolors` but less so with `termguicolors` (compared to using bold text in both cases).

Solution: use bold text with `termguicolors` and regular with `notermguicolors`.

------

@justinmk: *this* is a better way to make `TabLineSel` more visible. Although it was pretty visible even before #31896. Here is comparison:

`notermguicolors` before #31896:

screenshot_2025-01-07_20:02:29

`notermguicolors` after #31896 and this PR (absence of underline is quite visible as is; but removing bold text is even better, I agree):

screenshot_2025-01-07_20:01:59

------

`termguicolors` before #31896 and after this PR

screenshot_2025-01-07_20:02:16

`termguicolors` after #31896 (objectively harder to distinguish):

screenshot_2025-01-07_20:01:28


justinmk:
This is what I see with latest master + 'termguicolors' : 

image


"bold" is not at all noticeable enough. The current tab should look like it's connected-to the buffer.

justinmk:
It's possible that I got mixed when testing the previous PR, because of my custom tabline: https://github.com/justinmk/config/blob/b03530933f5d307c18fef92899ec65739d4316a6/.config/nvim/plugin/plugins.lua#L443-L453

jmbaur:
Thanks @echasnovski !

justinmk:
> > "bold" is not at all noticeable enough.
> 
> That depends on the font, I guess.

Main point is that the tab should visually connect-to the buffer. As long as that effect is in place, and the non-selected tabs have a clearly visually-distinct background color, I have no objections.

Issue #31652, opened 2024-12-20
:terminal doesn't propagate DEC mode 2031 (for reporting dark/light mode)


jonathanslenders:
### Problem

Recently, Neovim implemented DEC mode 2031 so that the terminal theme (dark/light) would remain synchronized with the outer terminal.

This works, but doesn't propagate to any application running within Neovim's embedded terminal, including nested Neovim instances.

cc: @gpanders

### Expected behavior

Applications within the embedded terminal should be able to enable mode 2031 and receive notifications when the background color of Neovim (or the outer terminal) change, and these applications should be able to query the foreground/background color.

justinmk:
You can define a `OptionSet bg` event handler which uses `nvim_chan_send()` to send a sequence to the `:terminal` buffer.

jonathanslenders:
@justinmk, I'm not sure it can be done with `nvim_chan_send()`. The sequence should only be sent to the terminal if the application running within the terminal enabled mode 2031. And it should probably be done as a response to a change of the 'bg' setting, regardless of whether or not 'bg' was updated because the host terminal notified nvim for a theme update.

gpanders:
> > The sequence should only be sent to the terminal if the application running within the terminal enabled mode 2031.
> 
> `:help TermRequest`
> 

This won't work today since TermRequest only fires for OSC and DSC commands, while enabling a DEC mode is a CSI command.

vterm has an option to call a callback for "unknown" CSIs, which we could quite easily hook up to TermRequest. The only difficulty there is documenting what exactly is an "unknown" CSI (the short answer is basically, "whatever vterm doesn't handle itself" but that's not very useful).

PR #31916, opened 2025-01-08
feat(terminal): propagate DECSET/DECRST to TermRequest

gpanders:
Modify vterm to add a new callback for when a DEC mode is set or reset. This callback emits a TermRequest event with the DECSET/DECRST sequence used.

This shouldn't be too noisy as DEC modes don't change that often (the noisiest mode will likely be bracketed paste since many shells enable that mode at the prompt and then disable it before running a command).

Ref: https://github.com/neovim/neovim/issues/31652#issuecomment-2578127646

### DIFF
--- a/runtime/doc/autocmd.txt
+++ b/runtime/doc/autocmd.txt
@@ -1004,9 +1004,10 @@ TermClose			When a |terminal| job ends.
 				Sets these |v:event| keys:
 				    status
 							*TermRequest*
-TermRequest			When a |:terminal| child process emits an OSC
-				or DCS sequence. Sets |v:termrequest|. The
-				|event-data| is the request string.
+TermRequest			When a |:terminal| child process emits an OSC,
+				DCS, DECSET, or DECRST sequence. Sets
+				|v:termrequest|. The |event-data| is the
+				request string.
 							*TermResponse*
 TermResponse			When Nvim receives an OSC or DCS response from
 				the host terminal. Sets |v:termresponse|. The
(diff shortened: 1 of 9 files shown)


justinmk (approved):

gpanders:
I haven't merged this yet because I'm not 100% sure it's the best approach.

For one thing, the phrasing is confusing: setting a DEC mode is not a "request" from the terminal which requires a response, so shoving this into `TermRequest` already bends the mental model a bit. I'm also not sure what other DEC modes aside from 2031 plugins would ever be interested in or could do anything about.

Another approach would be to simply teach vterm about DEC mode 2031 and handle it like other modes (i.e. set a terminal property when the mode is enabled). We could hook into `did_set_background` to loop through all terminal buffers and send the DSR sequence if the mode is enabled.

justinmk:
> setting a DEC mode is not a "request" from the terminal which requires a response, so shoving this into `TermRequest` already bends the mental model a bit.

As long as the `TermRequest` handler has a way to detect what kind of event it is, I think using `TermRequest` is fine. Though this is an example of why I wanted the event "pattern" to represent a name of some sort :) ("osc", "dcs", ...)

gpanders:
> Though this is an example of why I wanted the event "pattern" to represent a name of some sort :) ("osc", "dcs", ...)

This is unambiguous from the first 2 bytes of the event data, and using the buffer/filename as the pattern allows buffer-local events.

But even aside from the naming question, adding DECSET/DECRST to TermRequest doesn't seem useful aside from this _one_ use case, and may actually hurt due to the increase in (unused) TermRequest events. I think it'd be better to just handle this in vterm like we do for the other DEC modes.

gpanders:
Superseded by https://github.com/neovim/neovim/pull/31999

PR #31919, opened 2025-01-08
fix(api): validate option name for `nvim_get_option()`

### LINKED ISSUE #31894: Assertion failed: (opt_idx != kOptInvalid) when accessing invalid options
### Problem

Nvim 0.11-dev crashes with the following error

```
Assertion failed: (opt_idx != kOptInvalid), function get_option, file option.c, line 3399.
```

when accessing an option that does not exist. This is a regression since 29ded889579a9d590e8ea885a9a402ff4bae87be (#31066). /cc @famiu

### Steps to reproduce

```
:lua vim.api.nvim_get_option('invalid')
```

To my knowledge, this bug was first reported in https://github.com/neovim/pynvim/pull/584:
```bash
nvim --clean --headless -c "py3 vim.current.buffer.options['invalid-option']"
```

### Expected behavior

Should raise an error: `Invalid option name: 'invalid-option'`

### Nvim version (nvim -v)

0.11 commit 29ded8895

n/a

### Operating system/version

macOS

### Terminal name/version

N/A

N/A

### Installation

build from source

famiu:
Problem: `nvim_get_option()` doesn't validate the option name, which leads to an assertion failure.

Solution: Validate option name in `nvim_get_option()`.

Fix #31894

### DIFF
--- a/src/nvim/api/deprecated.c
+++ b/src/nvim/api/deprecated.c
@@ -646,6 +646,10 @@ static Object get_option_from(void *from, OptScope scope, String name, Error *er
   });
 
   OptIndex opt_idx = find_option(name.data);
+  VALIDATE_S(opt_idx != kOptInvalid, "option name", name.data, {
+    return (Object)OBJECT_INIT;
+  });
+
   OptVal value = NIL_OPTVAL;
 
   if (option_has_scope(opt_idx, scope)) {


famiu:
Should I add regression tests? This is a deprecated API so it seems like tests for it it were intentionally removed

justinmk:
> This is a deprecated API so it seems like tests for it it were intentionally removed

That shouldn't normally be the case. Instead we should move such tests to `deprecated_spec.lua`

PR #31921, opened 2025-01-08
feat(extmarks): virtual text can be right-aligned, truncated

georgev93:
When setting virtual text to be right aligned, the virtual text will print on top of buffer text if the virtual text is too long.

I would think a more desirable behavior for virtual text applications (such as LSP diagnostics) would be to keep the virtual text right aligned if it is short enough to fit without being truncated, but to truncate it and insert it at EOL if it is too long (thus avoiding the "virtual text is covering up buffer text" problem).

Right Aligned Example (note that the function name "getMemoryBlock" gets partially covered up by virtual text):

Right Aligned Truncated Example (since the virtual text in the right window is too long, it is truncated):


MariaSolOs:
@georgev93 if your main concern is LSP diagnostics, can't you achieve the truncation using `vim.diagnostic.config.virtual_text.format`?

georgev93:
> @georgev93 if your main concern is LSP diagnostics, can't you achieve the truncation using `vim.diagnostic.config.virtual_text.format`?

I'm uncertain how to accomplish this with `vim.diagnostic.config.virtual_text.format`. If there were a way to evaluate where the column position of the EOL character as well as the right-most column on the screen, I could certainly do the math in that function. I'm not exactly sure how to get those column positions from the context of that function though.

Additionally, this was the type of feature that I thought others may be interested in, so I figured extending neovim might be a nice thing to do. Considering the default way nvim shows virtual text is to truncate the virtual text rather than cover up buffer text, I interpreted that to indicate that most people prefer truncated virtual text over covering up buffer text. Therefore, this enables that "popular" feature, but just with right alignment as well.

georgev93:
I think the FreeBSD functionaltest failure is unrelated?

zeertzjq:
This can still overlap with buffer text if the line is wrapped and the extmark appears in non-last screen line. Since you don't want the virtual text to overlap with buffer text, you may want to skip the virtual text when `do_eol` is false, and name it `eol_right_align` instead.

src/nvim/drawline.c
```diff
@@ -275,6 +275,15 @@ static void draw_virt_text(win_T *wp, buf_T *buf, int col_off, int *end_col, int
     if (decor_virt_pos(item) && item->draw_col == -1) {
       bool updated = true;
       VirtTextPos pos = decor_virt_pos_kind(item);
+
+      if (pos == kVPosRightAlignTrunc) {
+        if (do_eol && (vt->width <= (right_pos - state->eol_col))) {
+          pos = kVPosRightAlign;
+        } else {
+          pos = kVPosEndOfLine;
+        }
```
zeertzjq, line 284: This will lead to inconsistent sorting order. `kVPosRightAlign` places virtual text from right to left, while `kVPosEndOfLine` places virtual text from left to right.

georgev93: You're absolutely right! I'll take another swipe at this

georgev93: Now I'm checking for the combined length of the eol_right_aligned virtual text, and setting the eol_col to a position that would result in combined right alignment. If the combined virtual text is too long to fit, it doesn't change the eol_col.

In either case, once it has conditionally changed the eol_col, it just treats the "eol_right_align" virtual text as "eol" virtual text.

src/nvim/drawline.c
```diff
@@ -275,6 +275,52 @@ static void draw_virt_text(win_T *wp, buf_T *buf, int col_off, int *end_col, int
     if (decor_virt_pos(item) && item->draw_col == -1) {
       bool updated = true;
       VirtTextPos pos = decor_virt_pos_kind(item);
+
+      if (do_eol && pos == kVPosEndOfLineRightAlign) {
```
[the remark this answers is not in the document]
georgev93: Hmm... I don't think there's a way to do this with less code since this "eol_right_align" option requires a "lookahead" to correctly determine its spacing. I could add a simple lookahead to all alignment cases, then there may be less code and more symmetry between the five options, but that would be at the expense of speed to the other four options. I figured speed was the priority over code size.

Is the problem the asymmetry rather than the code size of my PR? Looking at the other PRs, I don't think mine is particularly large, but it does cause an asymmetric look to the code for the five alignment options (again, because they have asymmetric requirements).

Also, I apologize if I failed to understand "the scope" of neovim in submitting this PR if this is truely "out of scope." I figured since I'm simply extending a feature that neovim already has (that vim does not). This extension doesn't change the behavior of existing code in neovim, so all things considered I would think this is a pretty benign change.

Lastly, if you're worried about the aesthetics of the code (such as line count), I can certainly accomplish what I'm doing in less lines by combining conditional logic, removing comments, and compressing variable names. I'd argue that would be a sacrifice of readability for a non-performative goal like line count. However, I couldn't argue with the sentiment of "keeping the coding style consistent."

georgev93:
Sorry for all the "bait and switching" in requesting for a review then flipping back to a draft PR! Caught one last issue right before the wife went in to labor... but looking through all of this one last time from the hospital I'm saying this is good to go! _final_final_v2... thanks for your patience.

test/functional/ui/decorations_spec.lua
```diff
@@ -509,6 +509,98 @@ describe('decorations providers', function()
...
+      // on Mulholland Drive            +{2:     }|
+      try_start();                      +{2:     }|
+      bufref_T save_buf;                +{2:     }|
+      switch_buffer(&save_buf, buf);    +{2:     }|
+      posp = getmark(mark, false);      +{2:     }|
+      restore_buffer(&save_buf);^        +{2:     }|
+                                              |
+    ]]}
+  end)
+
```
zeertzjq, line 603: This test is not needed. It is nearly identical to the test `can have virtual text of the style: right_align`.

runtime/doc/api.txt
```diff
@@ -2685,6 +2685,10 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
                     • "overlay": display over the specified column, without
                       shifting the underlying text.
                     • "right_align": display right aligned in the window.
+                    • "eol_right_align": display right aligned in the window
+                      unless the virtual text is longer than the space
+                      available. If the virtual text is too long, it is
+                      truncated to fit in the window after the EOL character.
```
zeertzjq, line 2691: This doesn't make it clear that when the line is wrapped, the virtual text is shown after the end of the line, not on a previous screen line.

zeertzjq:
Not related to this PR, but I noticed that the truncation behavior may be suboptimal for `virt_text_pos = 'eol'` as well. With `virt_text_pos = 'eol'`, if there isn't enough space, the virtual text with the highest priority is truncated.

test/functional/ui/decorations_spec.lua
georgev93: You are correct; I've removed the test. I was attempting to more explicitly document the expected "write over buffer text" behavior through the test, but this benefit hardly outweighs the extra time it takes to run the test every time.

runtime/doc/api.txt
georgev93: I added on a sentence for clarification on this subject.

georgev93:
> Not related to this PR, but I noticed that the truncation behavior may be suboptimal for `virt_text_pos = 'eol'` as well. With `virt_text_pos = 'eol'`, if there isn't enough space, the virtual text with the highest priority is truncated.

Good catch! Certainly not ideal. I also agree that this eol behavior is outside the scope of this PR.

src/nvim/drawline.c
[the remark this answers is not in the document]
justinmk: > less lines by combining conditional logic, removing comments, and compressing variable names.

Not looking for code golf, but want to avoid extra code in general. 

For example, maybe the position-finding loop is common to other code that operates on DecorRange.

I retract my "out of scope" comment, since the screentests do indicate that this looks valuable.

runtime/doc/api.txt
```diff
@@ -2685,6 +2685,13 @@ nvim_buf_set_extmark({buffer}, {ns_id}, {line}, {col}, {opts})
                     • "overlay": display over the specified column, without
                       shifting the underlying text.
                     • "right_align": display right aligned in the window.
+                    • "eol_right_align": display right aligned in the window
```
justinmk, line 2688: keys should be alphabetical. and for this PR at least, "eol_right_align" should go after "eol".

src/nvim/drawline.c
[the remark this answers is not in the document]
georgev93: Understood! Within the context of the `draw_virt_text` function, I was thinking I could change the overall structure to loop through the DecorRange slots, build a new dynamically allocated array containing the identified virtual text instances this function needs to display, then incrementing through that dynamically allocated array to display all the virtual text in the second half of that function (in the same manner that it does now). That would eliminate the extra pass for the "lookahead" in this `eol_right_align` case, making the `eol_right_align` case much more efficient, barely decreasing the efficiency of the other four cases, slight increase in memory use for all cases, and behavior between all cases would be much more similar.

^ I can push what I'm talking about here soon. Might be a day or so with the newborn!

Would this be in line with what you're looking for? Or are you looking for some common function to be pulled out of `draw_virt_text` that can also be used elsewhere in the same source file? If you're looking for the latter, that might take me an extra minute as I haven't studied the source as much at that scope.

wincent:
> I'm uncertain how to accomplish this with `vim.diagnostic.config.virtual_text.format`. If there were a way to evaluate where the column position of the EOL character as well as the right-most column on the screen, I could certainly do the math in that function. I'm not exactly sure how to get those column positions from the context of that function though.

Here's how I've been doing it:

```lua
local virtual_text = {
  virt_text_pos = 'right_align',
  format = function(diagnostic)
    local bufnr = diagnostic.bufnr
    local lnum = diagnostic.lnum
    local window = get_window(bufnr)
    if window == nil then
      return diagnostic.message
    end
    local win_width = vim.api.nvim_win_get_width(window)
    local line = vim.api.nvim_buf_get_lines(bufnr, lnum, lnum + 1, false)[1]
    local line_length = #line
    local padding = 20
    local excess = (line_length + padding + #diagnostic.message) - win_width
    if excess > 0 then
      local trimmed = string.sub(diagnostic.message, 1, -excess) .. '…'
      return trimmed
    else
      return diagnostic.message
    end
  end,
}
```

It only "kind of" works. In particular, it doesn't seem to cope well if I resize the window. There may be edge cases that it fails to handle. I'd be glad to replace it with a built-in option that does the right thing.

runtime/doc/diagnostic.txt
```diff
@@ -619,8 +619,8 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
       • {hl_mode}?            (`'replace'|'combine'|'blend'`) See
                               |nvim_buf_set_extmark()|.
       • {virt_text}?          (`[string,any][]`) See |nvim_buf_set_extmark()|.
-      • {virt_text_pos}?      (`'eol'|'overlay'|'right_align'|'inline'`) See
-                              |nvim_buf_set_extmark()|.
+      • {virt_text_pos}?      (`'eol'|'overlay'|'right_align'|'eol_right_align'|'inline'`)
```
justinmk, line 622: these could be alphabetical too. that helps discover/intuition.

runtime/doc/news.txt
```diff
@@ -180,6 +180,8 @@ API
 • |nvim__ns_set()| can set properties for a namespace
 • |nvim_echo()| `err` field to print error messages and `chunks` accepts
   highlight group IDs.
+• |nvim_buf_set_extmark()| virt_text_pos now accpets `eol_right_align` to
```
justinmk, line 183: typo
```suggestion
• |nvim_buf_set_extmark()| virt_text_pos accepts `eol_right_align` to
```

src/nvim/decoration.h
```diff
@@ -15,8 +15,8 @@
 // actual Decor* data is in decoration_defs.h
 
 /// Keep in sync with VirtTextPos in decoration_defs.h
-EXTERN const char *const virt_text_pos_str[]
-INIT( = { "eol", "overlay", "win_col", "right_align", "inline" });
+EXTERN const char *const virt_text_pos_str[] INIT( = { "eol", "overlay", "win_col", "right_align",
+                                                       "eol_right_align", "inline" });
```
justinmk, line 19: alphabetical

src/nvim/drawline.c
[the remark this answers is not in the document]
justinmk: > Or are you looking for some common function to be pulled out of `draw_virt_text` that can also be used elsewhere in the same source file

That. Not really worried about performance here, since this logic isn't working with much data.

src/nvim/drawline.c
```diff
@@ -275,7 +278,50 @@ static void draw_virt_text(win_T *wp, buf_T *buf, int col_off, int *end_col, int
...
+              continue;
+            }
+
+            if (!decor_virt_pos(lookaheadItem)) {
+              continue;
+            }
+
+            if (lookaheadItem->draw_col != -1) {
+              continue;
+            }
```
justinmk, line 300: this seems a bit verbose, and isn't really typical for this codebase.
```suggestion
            if (lookaheadItem->start_row != state->row
                || !decor_virt_pos(lookaheadItem)
                || lookaheadItem->draw_col != -1) {
              continue;
            }
```

src/nvim/decoration.h
georgev93: 401d944881941b87a51ad62b7ce4379be17117e5

src/nvim/drawline.c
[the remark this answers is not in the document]
georgev93: @justinmk, I've been shuffling the code around a bit and I'm not sure if there are any gains to be made from additional refactoring. Do you see anything specific that I'm missing? I'm not really familiar with the broader codebase, so my view may be a bit myopic for this task.

test/functional/ui/decorations_spec.lua
```diff
@@ -509,6 +509,69 @@ describe('decorations providers', function()
...
+      try_start();                 +{2:1234567890}|
+      bufref_T save_buf;           +{2:1234567890}|
+      switch_buffer(&save_buf, buf); +{2:12345678}|
+      posp = getmark(mark, false); +{2:1234567890}|
+      restore_buffer(&save_buf);^   +{2:1234567890}|
+                                              |
+    ]]}
+  end)
+
+  it('multiple instances of eol right align', function()
```
justinmk, line 541:
```suggestion
  it('multiple eol_right_align', function()
```

justinmk (approved):
LGTM, the +1 from @wincent adds useful feedback :)

Needs conflict resolution

georgev93:
Excellent! Thanks @justinmk . Conflict (news.txt) resolved. Do you want me to squash and fix the lintcommit or would you rather do the honors?

wincent:
Thanks for this @georgev93, @justinmk! 🙇🏻 

[Trying it out](https://github.com/wincent/wincent/commit/5f1c70c6107362b5356bcbf770b2500269cfc5a7) with the latest nightly and will report back if I notice anything noteworthy.

PR #31926, opened 2025-01-09
fix(terminal): disable CSIu encoding

### LINKED ISSUE #24093: Terminal mode should probably not send CSI-u sequences for shift+enter and shift+space
### Problem

When used with a terminal app which supports CSI-u sequences, neovim will use those to get more fine-grained key combo information for key combos, to be able to detect sequences such as shift+enter or shift+space. In `:terminal` mode, nvim will reproduce those CSI-u sequences. That means, when you hit shift+space in `:terminal` mode, neovim will send `\e[32;2u` to the terminal instead of a space character, and when you hit shift+enter, neovim will send `\e[13;2u` instead of a line feed character.

This is problematic. On most systems I've encountered, `\e[32;2u` and `\e13;2u` aren't interpreted as space or enter by default by any shell, which means the user ends up with garbage in their terminal if they accidentally hit space before letting go of enter. While some programs can be manually configured to understand the CSI-u sequences properly, it seems unreasonable to expect the user to manually go through every system they ever interact with (directly, through SSH, through serial, whatever) and reconfigure readline and whatever other line reading libraries are in use by terminal software they use.

I believe neovim should, by default, do what iTerm2 does by default: send a space character when shift+space is pressed and a line feed character when shift+enter is pressed, unless CSI-u mode is enabled by a program printing `\e[>1u` to the terminal, at which point it starts reporting those sequences using CSI-u. If that's too much work, I think disabling CSI-u for shift+enter and shift+space would be better than the current situation.

Note: As a workaround, shift+space and shift+enter can be remapped to space and enter with `tmap <S-Space> <Space>` and `tmap <S-Enter> <Enter>`.

### Steps to reproduce

1. `nvim --clean`
2. `:terminal`
3. Press shift+enter or shift+space

### Expected behavior

The shift key should be ignored, a space or enter should be written to the terminal. This is what iTerm2 and, I believe, most other terminals with CSI-u support do.

### Neovim version (nvim -v)

0.9.1

No, vim's :terminal mode sends a space and enter character for shift+space/shift+enter

### Operating system/version

macOS 13.4

### Terminal name/version

3.4.19

xterm-256color

### Installation

Homebrew

gpanders:
Resolves: https://github.com/neovim/neovim/issues/24093

vterm unconditionally uses the CSIu encoding proposed in [1]. This is roughly equivalent to the "Disambiguate escape codes" mode of the kitty keyboard protocol. However, because it's enabled unconditionally (and not enabled explicitly by an application), applications will receive escape sequences they don't understand (e.g. CSI 13 ; 2 u for Shift+Enter).

We _could_ teach vterm to understand the escape sequence for applications to opt-in to the "Disambiguate escape codes" mode and only use CSIu encoding when that mode is enabled, but then another problem arises: Nvim itself does not disambiguate those keys so by the time the user's input arrives at vterm, it is impossible to distinguish between e.g. `<C-I>` and `<Tab>` (even though these keys are encoded separately when received by the parent terminal!). If we supported the kitty sequence to enable the CSIu encoding then we wouldn't actually be disambiguating some escape codes, so the mode doesn't actually work properly.

Instead, we just disable CSIu encoding completely for now. If Nvim's input processing is ever changed so that `<C-I>` and `<Tab>` remain disambiguated, allowing us to send separate keys to vterm, then we could enable support for kitty's sequence to enable the disambiguate mode.

[1]: http://www.leonerd.org.uk/hacks/fixterms/

### DIFF
--- a/src/nvim/vterm/keyboard.c
+++ b/src/nvim/vterm/keyboard.c
@@ -25,29 +25,32 @@ void vterm_keyboard_unichar(VTerm *vt, uint32_t c, VTermModifier mod)
     return;
   }
 
-  int needs_CSIu;
+  bool needs_CSIu = false;
   switch (c) {
   // Special Ctrl- letters that can't be represented elsewise
   case 'i':
   case 'j':
   case 'm':
   case '[':
-    needs_CSIu = 1;
+    // CSIu encoding is disabled #31926
+    //needs_CSIu = true;
     break;
   // Ctrl-\ ] ^ _ don't need CSUu
   case '\\':
   case ']':
   case '^':
   case '_':
-    needs_CSIu = 0;
     break;
   // Shift-space needs CSIu
   case ' ':
-    needs_CSIu = !!(mod & VTERM_MOD_SHIFT);
+    // CSIu encoding is disabled #31926
+    //needs_CSIu = !!(mod & VTERM_MOD_SHIFT);
     break;
   // All other characters needs CSIu except for letters a-z
   default:
-    needs_CSIu = (c < 'a' || c > 'z');
+    // CSIu encoding is disabled #31926
+    //needs_CSIu = (c < 'a' || c > 'z');
+    break;
   }
 
   // ALT we can just prefix with ESC; anything else requires CSI u
@@ -187,7 +190,8 @@ void vterm_keyboard_key(VTerm *vt, VTermKey key, VTermModifier mod)
 
   case KEYCODE_LITERAL:
                         case_LITERAL:
-    if (mod & (VTERM_MOD_SHIFT|VTERM_MOD_CTRL)) {
+    // CSIu encoding is disabled #31926
+    if (mod & /*(VTERM_MOD_SHIFT|VTERM_MOD_CTRL)*/ 0) {
       vterm_push_output_sprintf_ctrl(vt, C1_CSI, "%d;%du", k.literal, mod + 1);
     } else {
       vterm_push_output_sprintf(vt, mod & VTERM_MOD_ALT ? ESC_S "%c" : "%c", k.literal);


bfredl:
> Nvim itself does not disambiguate those keys.

Not sure I follow. The only reason this could be problem in the first place is that neovim does disambiguate these keys, so that there are `<C-Enter>` etc sequences to forward.

_Unconditionally_ disabling these sequences in libvterm seems like a regression and loss of functionality which has existed and worked as intended for a long time. Working around "too much" disambiguation is a a lot easier than the other way around. Like if you frequently `<C-Enter>` by mistake and don't care about that particular key you can use

    tnoremap <c-Enter> <cr>

The opposite is of course possible but a lot more involved, you would need `nvim_chan_send()` and lookup the exact sequence to send yourself, i e reimplementing the keys layer of libvterm yourself.

I know people hate options but if using `tnoremap` is seen as too cumbersome, this c-layer disablement should at least be configurable.

bfredl:
Github ate part of your message so I missed it, but yes the default as-if `noremap <c-i> <tab>` mapping should apply to normal mode but not to terminal mode (although I still think this is an orthogonal issue to whether we use detection or not) . I have worked with this code before and will come up with something.

gpanders:
> Github ate part of your message 

My fault, forgot backticks around be angle brackets. Fixed now.

zeertzjq:
> Superseded by #31932

How? #31932 doesn't solve #24093. It makes it worse: #32019

gpanders:
> #31932 doesn't solve #24093. It makes it worse: #32019

With #31932 we can adapt vterm to use progressive enhancement. I’ll do that in a new PR.

gpanders:
https://github.com/neovim/neovim/pull/32038

PR #31930, opened 2025-01-09
feat(eval): allow writefile() to accept strings

AbhijithKumble:
Fix #30763

This is my first time contributing to this project, so how do i add test case coverage to the pr ?

### DIFF
--- a/src/nvim/eval/fs.c
+++ b/src/nvim/eval/fs.c
@@ -1393,6 +1393,35 @@ write_blob_error:
   return false;
 }
 
+/// Write a String to file with descriptor `fp`.
+///
+/// @param[in]  fp  File to write to.
+/// @param[in]  data String to write.
+///
+/// @return true on success, or false on failure.
+
+static bool write_string(FileDescriptor *const fp, const char *const data)
+  FUNC_ATTR_NONNULL_ARG(1)
+{
+  int error = 0;
+  const int len = strlen(data);
+  if (len > 0) {
+    const ptrdiff_t written = file_write(fp, data, (size_t)len);
+    if (written < (ptrdiff_t)len) {
+      error = (int)written;
+      goto write_string_error;
+    }
+  }
+  error = file_flush(fp);
+  if (error != 0) {
+    goto write_string_error;
+  }
+  return true;
+write_string_error:
+  semsg(_(e_error_while_writing_str), os_strerror(error));
+  return false;
+}
+
 /// "writefile()" function
 void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
 {
@@ -1408,9 +1437,9 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
         return;
       }
     });
-  } else if (argvars[0].v_type != VAR_BLOB) {
+  } else if (argvars[0].v_type != VAR_BLOB && argvars[0].v_type != VAR_STRING) {
     semsg(_(e_invarg2),
-          _("writefile() first argument must be a List or a Blob"));
+          _("writefile() first argument must be a List or a Blob or String"));
     return;
   }
 
@@ -1478,6 +1507,8 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
     bool write_ok;
     if (argvars[0].v_type == VAR_BLOB) {
       write_ok = write_blob(&fp, argvars[0].vval.v_blob);
+    } else if (argvars[0].v_type == VAR_STRING) {
+      write_ok = write_string(&fp, argvars[0].vval.v_string);
     } else {
       write_ok = write_list(&fp, argvars[0].vval.v_list, binary);
     }
--- a/src/nvim/eval/fs.c
+++ b/src/nvim/eval/fs.c
@@ -1436,7 +1436,7 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
       }
     });
   } else if (argvars[0].v_type != VAR_BLOB
-             && !(argvars[0].v_type == VAR_STRING && current_sctx.sc_sid != SID_LUA)) {
+             && !(argvars[0].v_type == VAR_STRING && current_sctx.sc_sid == SID_LUA)) {
     semsg(_(e_invarg2),
           _("writefile() first argument must be a List or a Blob"));
     return;
--- a/src/nvim/eval/fs.c
+++ b/src/nvim/eval/fs.c
@@ -1435,14 +1435,10 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
         return;
       }
     });
-  } else if (argvars[0].v_type != VAR_BLOB) {
-    if (argvars[0].v_type == VAR_STRING && current_sctx.sc_sid != SID_LUA) {
-      semsg(_(e_invarg2),
-            _("writefile() does not support Vim calls if the first argument is a string"));
-    } else {
-      semsg(_(e_invarg2),
-            _("writefile() first argument must be a List or a Blob"));
-    }
+  } else if (argvars[0].v_type != VAR_BLOB
+             && !(argvars[0].v_type == VAR_STRING && current_sctx.sc_sid != SID_LUA)) {
+    semsg(_(e_invarg2),
+          _("writefile() first argument must be a List or a Blob"));
     return;
   }
 
--- a/src/nvim/eval/fs.c
+++ b/src/nvim/eval/fs.c
@@ -1365,32 +1365,46 @@ write_list_error:
   return false;
 }
 
+/// Write data to file with descriptor `fp`.
+///
+/// @param[in]  fp    File to write to.
+/// @param[in]  data  Pointer to the data to write.
+/// @param[in]  len   Length of the data to write.
+///
+/// @return true on success, or false on failure.
+static bool write_data(FileDescriptor *const fp, const char *const data, const size_t len)
+  FUNC_ATTR_NONNULL_ARG(1)
+{
+  int error = 0;
+  if (len > 0) {
+    const ptrdiff_t written = file_write(fp, data, len);
+    if (written < (ptrdiff_t)len) {
+      error = (int)written;
+      goto write_data_error;
+    }
+  }
+  error = file_flush(fp);
+  if (error != 0) {
+    goto write_data_error;
+  }
+  return true;
+
+write_data_error:
+  semsg(_(e_error_while_writing_str), os_strerror(error));
+  return false;
+}
+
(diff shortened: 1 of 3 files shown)


luukvbaal:
Thanks for working on this. Test coverage should probably go in `test/functional/vrimscript/writefile_spec.lua`. Adding a testcase there should not prove to be too difficult if you look at the pre-existing tests.

src/nvim/eval/fs.c
```diff
@@ -1393,6 +1393,35 @@ static bool write_blob(FileDescriptor *const fp, const blob_T *const blob)
...
 }
 
+/// Write a String to file with descriptor `fp`.
+///
+/// @param[in]  fp  File to write to.
+/// @param[in]  data String to write.
+///
+/// @return true on success, or false on failure.
+
+static bool write_string(FileDescriptor *const fp, const char *const data)
```
justinmk, line 1403: nearly identical to `write_blob`. can we parameterize `write_blob` instead of copy-pasting it

justinmk:
> how do i add test case coverage to the pr ?

https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md links to https://github.com/neovim/neovim/blob/master/test/README.md#running-tests

src/nvim/eval/fs.c
AbhijithKumble: I have create a `write_data` function which will accept `char*` from `write_string` or `write_blob` and then `write_data` will write to the file.

AbhijithKumble:
other than string , blob or list where should the write test fail, i have removed tests that were passing lists, dicts, blobs as string, so those will become redundant now.

zeertzjq:
> other than string , blob or list where should the write test fail, i have removed tests that were passing lists, dicts, blobs as string, so those will become redundant now.

Why? Checks for invalid types should be tested.

AbhijithKumble:
> > other than string , blob or list where should the write test fail, i have removed tests that were passing lists, dicts, blobs as string, so those will become redundant now.
> 
> Why? Checks for invalid types should be tested.

Those are basically strings right and the test cases where checking if lists, dicts were passed as string, but now strings are accepted

zeertzjq:
> > > other than string , blob or list where should the write test fail, i have removed tests that were passing lists, dicts, blobs as string, so those will become redundant now.
> > 
> > 
> > Why? Checks for invalid types should be tested.
> 
> Those are basically strings right and the test cases where checking if lists, dicts were passed as string, but now strings are accepted

You are confusing the two arguments. Only strings are accepted for the second argument, both before and after the second PR. And for the first argument, there are still types that are not accepted.

AbhijithKumble:
> You are confusing the two arguments. Only strings are accepted for the second argument, both before and after the second PR. And for the first argument, there are still types that are not accepted.

oh got it, i will add the test case back to it

test/functional/vimscript/writefile_spec.lua
```diff
@@ -148,12 +148,6 @@ describe('writefile()', function()
       'Vim(call):E118: Too many arguments for function: writefile',
       pcall_err(command, ('call writefile([], "%s", "b", 1)'):format(fname))
     )
-    for _, arg in ipairs({ '0', '0.0', 'function("tr")', '{}', '"test"' }) do
-      eq(
-        'Vim(call):E475: Invalid argument: writefile() first argument must be a List or a Blob',
-        pcall_err(command, ('call writefile(%s, "%s", "b")'):format(arg, fname))
-      )
```
zeertzjq, line 155: The other types are still disallowed. Please only remove `'"test"'`, keeping the others and adjusting the expected error message.

test/old/testdir/test_writefile.vim
```diff
@@ -18,7 +18,6 @@ func Test_writefile()
   call assert_equal("vimmers", l[4])
   call delete(f)
```
zeertzjq, line 20:
```suggestion
  " call assert_fails('call writefile("text", "Xfile")', 'E475: Invalid argument: writefile() first argument must be a List or a Blob')
```

test/old/testdir/test_writefile.vim
```diff
@@ -18,7 +18,6 @@ func Test_writefile()
...
 
-  call assert_fails('call writefile("text", "Xfile")', 'E475: Invalid argument: writefile() first argument must be a List or a Blob')
```
zeertzjq, line 21: Lines in this files are commented out by adding `" ` at the beginning of the line, not removed. Please see other commented out lines in this file.

AbhijithKumble: done

AbhijithKumble: thanks for the help

zeertzjq:
I though about this a bit more. Although this change makes sense for Lua, it leads to inconsistencies with other Vim script functions. For example, in `setline()`, `\n` is translated to `\0` both when passing a single string and when passing a list of strings. However, here in `writefile()` this is only done when passing a list of strings, not a single string. Of course, `setline()` doesn't accept a blob, but this makes me wonder whether it's better to add a more general way to allow explicitly distinguishing between Vimscript blob and strings in Lua (that also works for variables instead of only function arguments).

justinmk:
> but this makes me wonder whether it's better to add a more general way to allow explicitly distinguishing between Vimscript blobs and strings in Lua

Not really in favor of adding this distinction on the Lua side. Lua shouldn't have to care about blobs at all.

> For example, in `setline()`, `\n` is translated to `\0` both when passing a single string and when passing a list of strings.

When calling `setline(string)` from Lua, let's treat it as blob? Could make this a general rule.

zeertzjq:
> > For example, in `setline()`, `\n` is translated to `\0` both when passing a single string and when passing a list of strings.
> 
> When calling `setline(string)` from Lua, let's treat it as blob? Could make this a general rule.

That seems even more complicated, and not backward-compatible, and requires information about whether each string comes from Lua to be available in typval strings.

bfredl:
I think `setline()` is irrelevant as context mandates it must be a _single_ line, so only NUL:s make sense, not interpreting `\n` as `\n`

> However, here in writefile() this is only done when passing a list of strings, not when passing a single string.

This should be completely fine for `writefile()`. There is already countless examples of vim scripts functions which are inconsistent with themselves and with other vim script functions. Like `setreg()` with a string _already_ behaves like this PR does for `writefile()`. 

In the end, for `writefile()` we are changing an **error** into an actual behavior, so we can just pick whatever behavior makes sense for us. Worrying about vimL self-consistency is the mind-killer.

test/functional/vimscript/writefile_spec.lua
```diff
@@ -148,9 +148,9 @@ describe('writefile()', function()
       'Vim(call):E118: Too many arguments for function: writefile',
       pcall_err(command, ('call writefile([], "%s", "b", 1)'):format(fname))
     )
-    for _, arg in ipairs({ '0', '0.0', 'function("tr")', '{}', '"test"' }) do
+    for _, arg in ipairs({ '0', '0.0', 'function("tr")', '{}' }) do
       eq(
-        'Vim(call):E475: Invalid argument: writefile() first argument must be a List or a Blob',
+        'Vim(call):E475: Invalid argument: writefile() first argument must be a List, a Blob or String',
```
justinmk, line 153: This message should not be changed. In Lua a blob *is* a string. And in Vimscript, string is not accepted for this file.

src/nvim/eval/fs.c
```diff
@@ -1478,6 +1505,8 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
     bool write_ok;
     if (argvars[0].v_type == VAR_BLOB) {
       write_ok = write_blob(&fp, argvars[0].vval.v_blob);
+    } else if (argvars[0].v_type == VAR_STRING) {
```
justinmk, line 1508: Instead of changing writefile to accept `VAR_STRING`, it should continue to only accept `VAR_BLOB`, but it should also check `current_sctx.sc_sid == SID_LUA`.

If `current_sctx.sc_sid == SID_LUA`, then use `vval.v_string` (I think? Or pass `vval.v_blob` and treat it as a string).

If that doesn't work, then possibly the "bridge" somewhere may need to either...

- special-case writefile when called from Lua
- convert all Lua calls to "blob". or maybe a list of certain functions.

test/functional/vimscript/writefile_spec.lua
zeertzjq: In this PR's implementation, a string is accepted. Otherwise it'll require
> information about whether each string comes from Lua to be available in typval strings.

src/nvim/eval/fs.c
zeertzjq: I think adding an additional `&& current_sctx.sc_sid == SID_LUA` here is enough. `vval.v_blob` is not valid when `v_type` is `VAR_STRING`.

AbhijithKumble: neovim crashes when i add this && current_sctx.sc_sid == SID_LUA ?

zeertzjq: Ah, not here. Near the error message above.

src/nvim/eval/fs.c
```diff
@@ -1409,8 +1436,13 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
...
-    semsg(_(e_invarg2),
-          _("writefile() first argument must be a List or a Blob"));
+    if (argvars[0].v_type == VAR_STRING && current_sctx.sc_sid != SID_LUA) {
+      semsg(_(e_invarg2),
+            _("writefile() does not support Vim calls if the first argument is a string"));
+    } else {
+      semsg(_(e_invarg2),
+            _("writefile() first argument must be a List or a Blob"));
+    }
     return;
```
zeertzjq, line 1446: This doesn't work, as the error message is shown on both branches. Also according to @justinmk there is no need to use two different error messages. So keep the original error message and just change the `(argvars[0].v_type != VAR_BLOB)` condition to `(argvars[0].v_type != VAR_BLOB && !(argvars[0].v_type == VAR_STRING && current_sctx.sc_sid == SID_LUA))`.

src/nvim/eval/fs.c
```diff
@@ -1408,7 +1435,8 @@ void f_writefile(typval_T *argvars, typval_T *rettv, EvalFuncData fptr)
         return;
       }
     });
-  } else if (argvars[0].v_type != VAR_BLOB) {
+  } else if (argvars[0].v_type != VAR_BLOB
+             && !(argvars[0].v_type == VAR_STRING && current_sctx.sc_sid != SID_LUA)) {
```
zeertzjq, line 1439:
```suggestion
             && !(argvars[0].v_type == VAR_STRING && current_sctx.sc_sid == SID_LUA)) {
```

zeertzjq:
Superseded by #39098

PR #31931, opened 2025-01-09
vim-patch:9.1.0996: ComplMatchIns may highlight wrong text

glepnir:
Problem:  ComplMatchIns may highlight wrong text
Solution: don't highlight in case of fuzzy match,
          skip-highlight when not inserting anything
          (glepnir)

closes: vim/vim#16404

https://github.com/vim/vim/commit/e890887b8052561ac5f8dce218e578ed28599cc6

### DIFF
--- a/test/functional/ui/popupmenu_spec.lua
+++ b/test/functional/ui/popupmenu_spec.lua
@@ -5853,6 +5853,39 @@ describe('builtin popupmenu', function()
           {2:-- INSERT --}                    |
         ]])
         feed('<Esc>')
+
+        -- Does not highlight the compl leader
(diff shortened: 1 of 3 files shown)


test/functional/ui/popupmenu_spec.lua
```diff
@@ -5853,6 +5853,38 @@ describe('builtin popupmenu', function()
           {2:-- INSERT --}                    |
         ]])
         feed('<Esc>')
+
```
zeertzjq, line 5856:
```suggestion

-- Does not highlight the compl leader
```

PR #31932, opened 2025-01-09
fix(getchar): do not simplify keycodes in terminal mode

bfredl:
The code represents a useful pattern in normal mode where remapping `<tab>` will implicitly also remap `<c-i>` unless you remap that explicitly. This however relies on the _unmapped_ behavior being identical which is not true in terminal mode, as vterm can distinguish these keys.

Vim seems to entangle this with kitty keyboard mode detection which is irrelevant for us. Conditional fallbacks depending on keyboard mode could be done completely inside `vterm/` without `getchar.c` getting involved, I would think.

ref https://github.com/neovim/neovim/pull/31926#issuecomment-2580000207

### DIFF
--- a/src/nvim/getchar.c
+++ b/src/nvim/getchar.c
@@ -1518,9 +1518,11 @@ int merge_modifiers(int c_arg, int *modifiers)
 
   if (*modifiers & MOD_MASK_CTRL) {
     if ((c >= '`' && c <= 0x7f) || (c >= '@' && c <= '_')) {
-      c &= 0x1f;
-      if (c == NUL) {
-        c = K_ZERO;
+      if (!(State & MODE_TERMINAL) || !(c == 'I' || c == 'J' || c == 'M' || c == '[')) {
+        c &= 0x1f;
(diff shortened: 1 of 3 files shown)


zeertzjq:
Does that mean this part in Vim's terminal code is not relevant?
```c
    void
check_no_reduce_keys(void)
{
    if (no_reduce_key_state != NRKS_CHECK
	    || no_reduce_keys >= 1
	    || curbuf->b_term == NULL
	    || curbuf->b_term->tl_vterm == NULL)
	return;

    if (vterm_using_key_protocol())
    {
	// "modify_other_keys" or kitty keyboard protocol was enabled while
	// waiting.
	no_reduce_key_state = NRKS_SET;
	++no_reduce_keys;
    }
}
```

bfredl:
I would like to avoid code like that, we don't need more global entangled state just because that is how vim ended up solving it. (just like the vim code is also entangled with processing incoming host terminal keys which we have fully separated )

zeertzjq:
Related #24093

PR #31934, opened 2025-01-09
refactor: delete duplicate utf8-functionality

dundargoc:
Also remove British National Replacement Character Set.

We keep the DEC Special Graphics and ASCII despite it not being unicode
as some old software such as calcurse still rely on this functionality.

References:
- https://github.com/neovim/neovim/pull/31934#discussion_r1911046426
- https://en.wikipedia.org/wiki/DEC_Special_Graphics
- https://vt100.net/docs/vt220-rm/chapter2.html#S2.4.3

### DIFF
--- a/src/nvim/mbyte_defs.h
+++ b/src/nvim/mbyte_defs.h
@@ -74,3 +74,5 @@ typedef struct {
 } CharBoundsOff;
 
 typedef utf8proc_int32_t GraphemeState;
+
+enum { UNICODE_INVALID = 0xFFFD, };
--- a/src/nvim/tui/termkey/termkey.c
+++ b/src/nvim/tui/termkey/termkey.c
@@ -634,40 +634,13 @@ static void eat_bytes(TermKey *tk, size_t count)
   tk->buffcount -= count;
 }
 
-// TODO(dundargoc): we should be able to replace this with utf_char2bytes from mbyte.c
 int fill_utf8(int codepoint, char *str)
 {
-  int nbytes = utf_char2len(codepoint);
-
+  int nbytes = utf_char2bytes(codepoint, str);
   str[nbytes] = 0;
-
-  // This is easier done backwards
-  int b = nbytes;
-  while (b > 1) {
-    b--;
-    str[b] = (char)0x80 | (codepoint & 0x3f);
-    codepoint >>= 6;
-  }
-
-  switch (nbytes) {
-  case 1:
-    str[0] = (codepoint & 0x7f); break;
-  case 2:
-    str[0] = (char)0xc0 | (codepoint & 0x1f); break;
-  case 3:
-    str[0] = (char)0xe0 | (codepoint & 0x0f); break;
-  case 4:
-    str[0] = (char)0xf0 | (codepoint & 0x07); break;
-  case 5:
-    str[0] = (char)0xf8 | (codepoint & 0x03); break;
-  case 6:
-    str[0] = (char)0xfc | (codepoint & 0x01); break;
-  }
-
   return nbytes;
 }
 
-#define UTF8_INVALID 0xFFFD
 static TermKeyResult parse_utf8(const unsigned char *bytes, size_t len, int *cp, size_t *nbytep)
 {
   unsigned nbytes;
@@ -681,7 +654,7 @@ static TermKeyResult parse_utf8(const unsigned char *bytes, size_t len, int *cp,
     return TERMKEY_RES_KEY;
   } else if (b0 < 0xc0) {
     // Starts with a continuation byte - that's not right
-    *cp = UTF8_INVALID;
+    *cp = UNICODE_INVALID;
     *nbytep = 1;
     return TERMKEY_RES_KEY;
   } else if (b0 < 0xe0) {
@@ -700,7 +673,7 @@ static TermKeyResult parse_utf8(const unsigned char *bytes, size_t len, int *cp,
     nbytes = 6;
     *cp = b0 & 0x01;
   } else {
-    *cp = UTF8_INVALID;
+    *cp = UNICODE_INVALID;
     *nbytep = 1;
     return TERMKEY_RES_KEY;
   }
@@ -714,7 +687,7 @@ static TermKeyResult parse_utf8(const unsigned char *bytes, size_t len, int *cp,
 
     cb = bytes[b];
     if (cb < 0x80 || cb >= 0xc0) {
-      *cp = UTF8_INVALID;
+      *cp = UNICODE_INVALID;
       *nbytep = b;
       return TERMKEY_RES_KEY;
     }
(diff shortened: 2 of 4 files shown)


src/nvim/tui/termkey/termkey.c
```diff
@@ -637,34 +637,7 @@ static void eat_bytes(TermKey *tk, size_t count)
...
-
-  // This is easier done backwards
-  int b = nbytes;
-  while (b > 1) {
-    b--;
-    str[b] = (char)0x80 | (codepoint & 0x3f);
-    codepoint >>= 6;
-  }
-
-  switch (nbytes) {
```
dundargoc, line 652: I guess we could replace the implementation in `utf_char2bytes` with this as they do the same thing, but don't think it's a big readability win either way. Can replace if anyone likes this version more.

dundargoc:
```
test_edit.vim:                  application/octet-stream; charset=binary
```

???

src/nvim/mbyte_defs.h
```diff
@@ -74,3 +74,5 @@ typedef struct {
 } CharBoundsOff;
 
 typedef utf8proc_int32_t GraphemeState;
+
+#define UNICODE_INVALID 0xFFFD
```
zeertzjq, line 78:
```suggestion
enum { UNICODE_INVALID = 0xFFFD, };
```

zeertzjq:
You are removing features from `:terminal`. Isn't this a breaking change?

src/nvim/vterm/state.c
```diff
@@ -645,7 +645,7 @@ static int on_escape(const char *bytes, size_t len, void *user)
 
     {
       int setnum = bytes[0] - 0x28;
-      VTermEncoding *newenc = vterm_lookup_encoding(ENC_SINGLE_94, bytes[1]);
+      VTermEncoding *newenc = vterm_lookup_encoding();
```
gpanders, line 648: I think instead we should delete this entire `case` block with a comment indicating only UTF-8 is supported. Leaving the code here provides the (wrong) impression that we are actually handling other character sets.

src/nvim/vterm/encoding.c
```diff
@@ -164,115 +163,7 @@ static VTermEncoding encoding_utf8 = {
...
-    [0x77] = 0x252C,  // BOX DRAWINGS LIGHT DOWN AND HORIZONTAL
-    [0x78] = 0x2502,  // BOX DRAWINGS LIGHT VERTICAL
-    [0x79] = 0x2A7D,  // LESS-THAN OR SLANTED EQUAL-TO
-    [0x7a] = 0x2A7E,  // GREATER-THAN OR SLANTED EQUAL-TO
-    [0x7b] = 0x03C0,  // GREEK SMALL LETTER PI
-    [0x7c] = 0x2260,  // NOT EQUAL TO
-    [0x7d] = 0x00A3,  // POUND SIGN
-    [0x7e] = 0x00B7,  // MIDDLE DOT
-  }
-};
```
gpanders, line 248: I think we should keep the line drawing character set since there is at least one known application which uses it (and box drawing is not an extremely uncommon use case).

foot, Ghostty, and [libvaxis](https://github.com/rockorager/libvaxis) all support the box drawing character set (possibly others, these are just the ones I know of for sure).

src/nvim/vterm/state.c
gpanders: See my [other comment](https://github.com/neovim/neovim/pull/31934/files#r1910927726), I think we should keep the DEC drawing encoding (removing the others is probably fine), in which case this should stay as-is.

gpanders:
This will definitely need an item in news.txt (even though there are probably only 3 users who will know what it means).

src/nvim/vterm/encoding.c
clason: But are they using _these_ characters and not the unicode ones?

(Link from chat for the "one known application": https://github.com/pimutils/khal)

rockorager: Actually it's the [vaxis](https://git.sr.ht/~rockorager/vaxis) embedded terminal that supports that character set. It didn't initially but we received bug reports in `aerc` for `khal` which used that charset for the box drawing glyphs.

clason: OK, just wanted to make sure. And they don't prefer the unicode set if that is available?

clason: And for the record, we're talking about  https://en.wikipedia.org/wiki/DEC_Special_Graphics

rockorager: These work by mapping ascii codepoints to other codepoints. So switching the charset to `0`, you can print a `j` and get a `┘`

clason: Just like `#` to `£` 🫖

rockorager: Also sorry, I told Greg the wrong application. It was [calcurse](https://calcurse.org/) which uses the box drawing charset

dundargoc: > At least [Ghostty is](https://github.com/ghostty-org/ghostty/blob/799f5b823904a2076a49c494ae50b486ac519368/src/terminal/charsets.zig#L54-L84).

[ghostty implements the british and us encoding as well though](https://github.com/ghostty-org/ghostty/blob/799f5b823904a2076a49c494ae50b486ac519368/src/terminal/charsets.zig#L17-L20)

dundargoc: > It didn't initially but we received bug reports in aerc for khal which used that charset for the box drawing glyphs.

@rockorager do you have a link to it so I can get more context?

dundargoc: OK, the drawing character set is widely supported by most terminals and tmux, so it makes sense to keep it.

dundargoc:
TODO:

- link to this discussion and to https://en.wikipedia.org/wiki/DEC_Special_Graphics

src/nvim/vterm/state.c
```diff
@@ -645,7 +645,7 @@ static int on_escape(const char *bytes, size_t len, void *user)
 
     {
       int setnum = bytes[0] - 0x28;
-      VTermEncoding *newenc = vterm_lookup_encoding(ENC_SINGLE_94, bytes[1]);
+      VTermEncoding *newenc = vterm_lookup_encoding(ENC_SINGLE_94);
 
       if (newenc) {
         state->encoding[setnum].enc = newenc;
```
gpanders, line 651: You need to keep the designator, otherwise any `ESC (` sequence will always switch to the DEC graphics mode here (and it will be impossible to switch back).

Another issue here is that I think some applications will switch to DEC graphics char set and then switch back to the "normal" char set with `ESC ( B` (i.e. ASCII). That won't work here now since `ESC ( B` will return NULL from `vterm_lookup_encoding` and the char set will be stuck in the graphics char set.

We could either (1) keep the `usascii` encoding as-is, or (2) just use UTF-8 for the `B` charset.

clason: Vote for option (2).

dundargoc: I think we might need to keep the usaascii encoding as well as single shift 2 and 3 doesn't work without it, so it's not possible to temporarily switch to another encoding. Let's just keep it as is, I didn't except non-unicode to still be this entrenched in terminal space.

dundargoc:
Uh, maybe. It's the Single Shift commands that makes me nervous if we can remove it, since you're supposed to be able to temporarily switch to DEC and back and without the GR/GL circlejerking in `usaascii` it might not be possible? It might be, but the uncertainty makes me think this is not a good investment of time unless someone know beforehand for sure this isn't needed.

PR #31948, opened 2025-01-10
fix(docs): use syntactically correct Lua code example

echasnovski:
`function` is a reserved keyword in Lua and can not be used in the `mymod.function` form.


dundargoc:

PR #31950, opened 2025-01-10
fix(comment): always get innermost ts commentstring

ribru17:
~The `level > res_level` is always true since `level >= 1`, and `res_level` is never reassigned~

### DIFF
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -37,7 +37,7 @@ local function get_commentstring(ref_position)
     for _, ft in ipairs(filetypes) do
       local cur_cs = vim.filetype.get_option(ft, 'commentstring')
       if cur_cs ~= '' and level > res_level then
-        ts_cs = cur_cs
+        ts_cs, res_level = cur_cs, level
       end
     end
 


echasnovski:
Oh, indeed. I think it rather should be this patch:

```diff
diff --git a/runtime/lua/vim/_comment.lua b/runtime/lua/vim/_comment.lua
index de7f62632c..803129ad09 100644
--- a/runtime/lua/vim/_comment.lua
+++ b/runtime/lua/vim/_comment.lua
@@ -37,7 +37,7 @@ local function get_commentstring(ref_position)
     for _, ft in ipairs(filetypes) do
       local cur_cs = vim.filetype.get_option(ft, 'commentstring')
       if cur_cs ~= '' and level > res_level then
-        ts_cs = cur_cs
+        ts_cs, res_level = cur_cs, level
       end
     end
 
```

As it was intended to get a 'commentstring' from the "deepest" local tree-sitter language. Not sure how to easily construct a failing test for this, though, but it seems necessary.

ribru17:
In my experience, the comments do always get the deepest tree even without this :thinking:

echasnovski:
> and it is very unlikely (impossible?) that we iterate:

It is indeed very unlikely, but I am not confident in my tree-sitter knowledge to say that it is impossible. There doesn't seem to be a restriction for parsers to *not* cover intersecting regions. So the situation like two top-level (level 1) sibling parsers covering whole document and the first one contains single child covering cursor (level 2). As traversing is done depth first, the the sibling level 1 parser will be tested after the deepest one and might produce the result.

For example, I've had an issue about three weeks ago on Nightly with installed 'markdown', 'markdown_inline', and 'html' parsers when code block with tree-sitter supported language (like `lua`) was recognized as 'html' (which was still the parser covering cursor, I believe). It was both for my use case and built-in commenting. Sadly, I can not reproduce this on currently latest Nightly and updated parsers, so it might as well has been an issue in Neovim, parser, or my repro steps. There was no issue on 0.10.2 at the moment with the same setup.

All in all, I don't see big harm in following patch from [this comment](https://github.com/neovim/neovim/pull/31950#issuecomment-2584061995). Would you mind repurposing this PR or should I do a separate one?

ribru17:
Gotcha. Sure I can repurpose this PR, hopefully that issue you mentioned will not crop up again then haha

echasnovski (approved):

ribru17:
I found a repro for your original issue, and I think this PR doesn't handle it fully sadly. The issue is with combined injections:

```md
# Here is md

~~~lua
ffi.cdef([[
int (*fun)(int, char *);
]])
~~~

<!-- html -->

~~~lua
ffi.cdef([[ // This sometimes gets html commentstring
int (*fun)(int, char *);
]])

ffi.cdef([[
int (*fun)(int, char *);
]])
~~~

<!-- html -->

~~~c
int (*fun)(int, char *); // This sometimes gets html commentstring
~~~

<!-- html -->
```

echasnovski:
Yeah, indeed. Thanks for finding this! It does look like the example I had.

After investigation, I believe the problem here is the combination of:
1. `LanguageTree:contains()` returns `true` for both 'html' and 'lua' language trees here. I don't think this should be the case and created #31984 for this.
2. Both 'html' and 'lua' trees are siblings. Combined with no order guarantees of `pairs()`, this results into either of them.

If 1. (i.e. #31984) is deemed to be the intended behavior here, then instead of `lang_tree:contains()` there needs to be a separate `for` loop through language tree's included regions with manual comparison.

chernetskyi:
Maybe the fix does not solve fully the @echasnovski's issue, but don't we still need it?

ribru17:
The issue was solved by #33948 I believe

Issue #31000, opened 2024-10-30
virt_lines doesn't scroll horizontally


WhyNotHugo:
### Problem

[lsp_lines.nvim](https://git.sr.ht/~whynothugo/lsp_lines.nvim) uses `nvim_buf_set_extmark` to create extmarks in a buffer. These extmarks use `virtual_lines` with some spaces on the left so that the properly align with the position of a diagnostic on the line above.

When I scroll to the right on buffer, the extmarks remain anchored in position relative to the _window_, not to the buffer. As a result, the extmarks are not properly positioned relative to the buffer, and the overflowing portion of the extmarks are unreadable.

extmarks and their position are declared for buffers, so their position should be relative to buffers, not windows.

Given that multiple windows can render the same buffer, I cannot take the scroll offset into account when rendering the extmarks, because one the "correct" value varies per-window.

### Steps to reproduce

Add this to your `init.lua`:

```lua
vim.keymap.set("n", "<Leader>x", function()
  local ns = vim.api.nvim_create_namespace("test")
  local line = {
    "hello, this is just some test text. It should be anchored under first line",
  }
  local virt_lines = { { line } }

  vim.api.nvim_buf_set_extmark(0, ns, 0, 0, {
    virt_lines = virt_lines,
  })
end)
```

Open some file and press `250ix`, then press <kbd>Esc</kbd> and <kbd>Leader</kbd><kbd>x</kbd>.

Scroll to the right.

The virtual text will maintain its absolute position relative to the window, not relative to the buffer.

### Expected behavior

When scrolling to the right, the virtual line should scroll too, keeping its position relative to the buffer.

### Nvim version (nvim -v)

NVIM v0.10.2

### Vim (not Nvim) behaves the same?

n/a

### Operating system/version

Alpine Linux

### Terminal name/version

foot

### $TERM environment variable

foot

### Installation

Alpine packages (edge)

xzbdmw:
A related problem https://github.com/windwp/nvim-autopairs/pull/474

MariaSolOs:
@WhyNotHugo thanks for the ping. No idea of how to fix this either, but I'll keep it in my mental backlog.

zeertzjq:
Considering #18282, I think there are two options here:
1. Add a single new flag to control both whether virtual lines scroll horizontally and whether virtual lines wrap.
   - With `'wrap'`, the flag enables wrapping for virtual lines attached to an extmark.
[...]

Issue #16545, opened 2021-12-06
Allow to right align diagnostic virtual text and allow it to have all diagnostics of the current line instead of just one of them


vatosarmat:
### Feature description

There is `vim.diagnostic.config` `opts.virtual_text.format` but it allows to format and display only one diagnostic. How about to add something like `opts.virtual_text.format_line` which would allow to format and display in virtual text all the diagnostics of the current line?

Currently it is possible via re-implementing `vim.diagnostic.handlers.virtual_text.show`, but I had to hack-export `private local function save_extmarks(namespace, bufnr)`
![diagnostic_virtual_text](https://user-images.githubusercontent.com/35213999/144816888-64dd6706-bc14-48ae-8dfa-f8e6fe53a6dd.png)

gpanders:
I think finding out why (and if) `save_extmarks` and `restore_extmarks` are necessary and fixing the underlying issue there (and thus allowing us to remove those two functions) is the best solution here, as that then allows the underline and virtual text handlers to be fully overridable.

ColinKennedy:
@vatosarmat Could you share your implementation on how you got virtual text to right-align?

I'm trying to get one of the following behaviors in Neovim

Right align (@vatosarmat 's request)

``` 
foo text                     # virtual stuff
bar                              # more blah
more text here                # text here tt
```

Or even just "starts at a certain column"

``` 
foo text                         # virtual stuff
bar                              # more blah
more text here                   # text here tt
really long text that exceeds column  # more text
```

I've been looking around for any way to do this via `vim.diagnostic.config()` but haven't found anything. So far I'm just setting the virtual text spacing very high while I wait to find a cleaner solution.

MariaSolOs:
@gpanders (cc @justinmk @glepnir) what do you think of changing the API of `vim.diagnostic.config.virtual_text.format` to take an list of diagnostics instead? This would address this issue and it seems like it wouldn't require a lot from neovim's side [(given that `reformat_diagnostics` already receives all diagnostics)](https://github.com/neovim/neovim/blob/58a1ef8e6a93c615379f6fbe7234697bcdc42b3e/runtime/lua/vim/diagnostic.lua#L131). It does introduce a breaking change, but I find it reasonable for the virtual text handler to have all the diagnostic information available for the current line.

gpanders:
[...]

PR #31959, opened 2025-01-11
feat(diagnostic): virtual_lines

MariaSolOs:
Other editors (like [helix](https://helix-editor.com/news/release-25-01-highlights/#diagnostics)) offer built-in functionality for displaying multiline diagnostics as virtual lines. Plugins like [`lsp_lines`](https://git.sr.ht/~whynothugo/lsp_lines.nvim) already offer this in Neovim, so let's bring it to core :)

Note that this PR is basically a port of `lsp_lines`, and hence all fame and glory should be forwarded to @WhyNotHugo.

EDIT: Including some pictures of what this looks like (with the `current_line` option set)
image
image


WhyNotHugo:
The latest version of lsp_lines implements `only_current_line`, so you should be able to use the whole implementation as a starting point.

Personally, I map `<leader>l` to toggle virtual_lines. But I wouldn't provide a mapping out of the box; it's probabl best to provide usage and setup recommendations in the docs, and let users pick a mapping of their liking. Especially since some will want to do multiple actions on the same mapping (e.g.: toggle virtual_text for other lines and toggle virtual_lines for the current line).

@clason: The current implementation in lsp_lines looks extremely similar to your screenshot of helix (I wonder if they took inspiration from lsp_lines, or if it's simply a naturally convergent design):

image

To achieve the same result from your screenshot, you will also need a feature to "hide virtual_text for current line". This can be implemented independently of this virtual lines.

clason:
> @clason: The current implementation in lsp_lines looks extremely similar to your screenshot of helix (I wonder if they took inspiration from lsp_lines, or if it's simply a naturally convergent design):

🤔 

Yes, these are two different questions: 
1. Adding support for showing diagnostics in virtual lines as a diagnostic handler.
2. Wiring up and exposing as diagnostic UX.

We should focus on 1. for now and leave 2. as a separate discussion.

clason:
The question came up of lsp-lines' license: As far as we understand it, it's compatible with Apache 2.0 so we could include it (provided we keep the current license text, e.g., in a header of a new Lua module), but it would make things simpler if you explicitly gave permission to include the (adapted) code under Neovim's own Apache 2.0 license. Not a deal-breaker either way.

WhyNotHugo:
You have my explicit permission to redistribute my work on lsp_lines.nvim under the Apache 2.0 licence.

Please do note that other contributors have collaborated to the project.

clason:
Noted, but your general willingness is the most important question. We can follow up with the others if we care sufficiently once this is done.

MariaSolOs:
@WhyNotHugo perhaps it's obvious, but how is https://github.com/neovim/neovim/pull/30070 meant to ease the implementation of `lsp_lines`?

runtime/lua/vim/diagnostic.lua
```diff
@@ -228,6 +229,14 @@ end
 --- See |nvim_buf_set_extmark()|.
 --- @field virt_text_hide? boolean
 
+--- @class vim.diagnostic.Opts.VirtualLines
+---
+--- Only show diagnostics for the current line.
+--- @field only_current_line? boolean
+---
+--- The characters used to draw the lines between diagnostic messages.
+--- @field drawing_characters? {cross?:string, horizontal?:string, horizontal_up?:string, up_right?:string, vertical?:string, vertical_right?:string}
```
MariaSolOs, line 238: Wonder if instead of indicating the characters here we want to expose options like "rounded"/"rectangular" etc

runtime/lua/vim/diagnostic.lua
```diff
@@ -228,6 +229,14 @@ end
 --- See |nvim_buf_set_extmark()|.
 --- @field virt_text_hide? boolean
 
+--- @class vim.diagnostic.Opts.VirtualLines
```
MariaSolOs, line 232: `lsp_lines` also exposes an [option](https://git.sr.ht/~whynothugo/lsp_lines.nvim#usage) to highlight the entire diagnostic line. I didn't include that here, both as a matter of personal taste and because neither the pictures in the README nor Helix's usage have such highlighting. Open to discussion though.

MariaSolOs:
I'll publish this so that we can inaugurate the bikeshedding ceremony but note that I'm still missing tests.

runtime/lua/vim/diagnostic.lua
gpanders: I'm not sure we need to make this configurable at all, at least for now. I think we hard code it for now and if people ask for it, we can make it an option. But if possible, the fewer knobs the better.

funcref:
Against this. Please don’t do this. Most people switch from IDEs to Vim/Neovim because it is lightweight. Don’t let Neovim continue down the path of becoming bloated. This kind of UI is not needed by everyone, so please don’t force it into the core. Of course, there may be options to turn it on or off by default, but keep in mind that Neovim is becoming more bloated with each release. Helix lacks a robust plugin ecosystem, so they had to implement such features internally. Neovim, on the other hand, already has a mature and excellent plugin ecosystem.

runtime/lua/vim/diagnostic.lua
WhyNotHugo: Agreed that we can skip it for the initial version.

For future versions, we can consider using a format similar the `border` parameter for `nvim_open_win`.

WhyNotHugo:
> @WhyNotHugo perhaps it's obvious, but how is #30070 meant to ease the implementation of `lsp_lines`?

https://todo.sr.ht/~whynothugo/lsp_lines.nvim/3

MariaSolOs:
> > @WhyNotHugo perhaps it's obvious, but how is #30070 meant to ease the implementation of `lsp_lines`?
> 
> https://todo.sr.ht/~whynothugo/lsp_lines.nvim/3

Ohh so it looks like now that https://github.com/neovim/neovim/pull/30070 went it, this should be automagically supported.

runtime/doc/diagnostic.txt
```diff
@@ -170,6 +170,16 @@ show a sign for the highest severity diagnostic on a given line: >lua
     }
 <
 
+                                          *diagnostic-toggle-virtual-lines-example*
+Diagnostic handlers can also be toggled. For example, you might want to toggle
+the `virtual_lines` handler with the following keymap: >lua
+
+    vim.keymap.set('<leader>l', function()
```
justinmk, line 177:
```suggestion
    vim.keymap.set('gK', function()
```

runtime/doc/diagnostic.txt
```diff
@@ -590,6 +600,12 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
                      diagnostics matching the given severity
                      |diagnostic-severity|.
 
+*vim.diagnostic.Opts.VirtualLines*
+
+    Fields: ~
+      • {only_current_line}?  (`boolean`, default: `true`) Only show
```
justinmk, line 606:
```suggestion
      • {current_line}?  (`boolean`, default: `true`) Only show
```

runtime/lua/vim/diagnostic.lua
```diff
@@ -1603,6 +1611,267 @@ M.handlers.virtual_text = {
...
+    opts = opts or {}
+
+    if not vim.api.nvim_buf_is_loaded(bufnr) then
+      return
+    end
+
+    local ns = M.get_namespace(namespace)
+    if not ns.user_data.virt_lines_ns then
+      ns.user_data.virt_lines_ns =
+        api.nvim_create_namespace(string.format('%s/diagnostic/virtual_lines', ns.name))
```
justinmk, line 1845: nit, non blocker, just for reference: I think we should name our namespaces/augroups with `nvim.foo.bar` format

src/nvim/highlight_group.c
```diff
@@ -232,6 +232,11 @@ static const char *highlight_init_both[] = {
   "default link DiagnosticVirtualTextInfo  DiagnosticInfo",
   "default link DiagnosticVirtualTextHint  DiagnosticHint",
   "default link DiagnosticVirtualTextOk    DiagnosticOk",
+  "default link DiagnosticVirtualLinesError DiagnosticError",
+  "default link DiagnosticVirtualLinesWarn  DiagnosticWarn",
+  "default link DiagnosticVirtualLinesInfo  DiagnosticInfo",
+  "default link DiagnosticVirtualLinesHint  DiagnosticHint",
+  "default link DiagnosticVirtualLinesOk    DiagnosticOk",
```
justinmk, line 239: If we really need new hl groups they should probably link to `DiagnosticVirtualText*` ? I don't really see why we need new hl groups though.

justinmk (approved):

src/nvim/highlight_group.c
clason: Conceivably, you might wish to use, e.g., italics for virtual text (to more clearly distinguish it from actual buffer text), while this would be too much for virtual lines (where you potentially see more text, including all diagnostics).

Conversely, you might wish to use a different background to make clear these are boguslines.
(Although that would probably be better handled by allowing for a uniform background color different from normal.)

MariaSolOs: I do think that we need separate highlight groups. @clason gave some good examples in favor of it.

> If we really need new hl groups they should probably link to `DiagnosticVirtualText*` ? I don't really see why we need new hl groups though.

Sure, but I find it a bit weird that this would be the only diagnostic handler whose highlight groups don't link to `Diagnostic*`. I chose this for a matter of consistency.

runtime/doc/diagnostic.txt
justinmk: AFAICT there are no other `only_` fields. I doubt this is the first time we have needed such a boolean. And it's true that names can have minor ambiguities without knowing the type--but the type is available to resolve those minor ambiguities when they arise. That's a good tradeoff to avoid over-explaining names in general.

MariaSolOs: Fair enough. I'll rename it.

runtime/lua/vim/diagnostic.lua
MariaSolOs: Here you go: https://github.com/neovim/neovim/pull/32001

mfussenegger (commented):
Just to put it on the table: Could consider making this a `style = lines` option for the existing virt-text handler. 

Reasons: 
- Having both active at the same time would be ... noisy
- No new hl groups 
- The current line option is useful for both
- Could later consider a dynamic/best-fit style that uses either lines or text depending on the diagnostic length

MariaSolOs:
> * Having both active at the same time would be ... noisy

As @WhyNotHugo mentioned above, some people (including myself) might be interested in having `virtual_lines` only for the current line and `virtual_text` for the rest. I would also like to be able to toggle `virtual_lines` only, but not `virtual_text`.

> * No new hl groups

I think this is a con (previous discussion [here](https://github.com/neovim/neovim/pull/31959#discussion_r1913084967)).

> * Could later consider a dynamic/best-fit style that uses either lines or text depending on the diagnostic length

Given that `virtual_text` already allows formatting the diagnostic, the length is slightly irrelevant here IMO since it can always be truncated there.

gpanders:
Nit on the commit message: should be `feat(diagnostic)`, not `feat(lsp)`

runtime/lua/vim/diagnostic.lua
```diff
@@ -228,6 +229,12 @@ end
 --- See |nvim_buf_set_extmark()|.
 --- @field virt_text_hide? boolean
 
+--- @class vim.diagnostic.Opts.VirtualLines
+---
+--- Only show diagnostics for the current line.
+--- (default: `true`)
```
clason, line 235: Should be `false`

clason, line 235: I think the current interface is more akin to inlay hints, so default `false` makes more sense.

(If and how this is wired up to the "show diagnostic" and "next diagnostic" functions is a different -- follow-up -- question.)

MariaSolOs:
From Matrix: Make sure I disable the handler by default. cc @gpanders

runtime/doc/news.txt
```diff
@@ -70,6 +70,8 @@ DIAGNOSTICS
   the "severity_sort" option.
 • Diagnostics are filtered by severity before being passed to a diagnostic
   handler |diagnostic-handlers|.
+• A "virtual_lines" diagnostic handler was added to render diagnostics using
+  virtual lines below the respective code.
```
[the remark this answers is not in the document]
MariaSolOs: Oh my bad, I just looked for the diagnostics section when adding this bullet.

gpanders:
Some diagnostics can be very long, e.g. from rust-analyzer:

image

We should probably have a `max_lines` configuration that truncates the number of virtual lines that are added (with something like `and N more lines` as the last line). The full message can still be seen with `open_float`.

This should probably be set to something low by default, like 3.

WhyNotHugo:
> We should probably have a max_lines configuration that truncates the number of virtual lines that are added (with something like and N more lines as the last line). The full message can still be seen with open_float.
> 
> This should probably be set to something low by default, like 3.

-1 If a long diagnostic is an problem, the handler can be [temporarily] disabled to hide them. The same is true if a single line has too many diagnostics.

We shouldn't truncate diagnostics by default (especially considering that this handler is disabled by default anyway).

gpanders:
> We shouldn't truncate diagnostics by default (especially considering that this handler is disabled by default anyway).

Then perhaps a knob to not show virtual lines _at all_ (instead of truncating) if there are more than `N` lines.

I never want to see half my screen taken up by virt lines as in that example above. Frankly, I don't even really want to see more than one diagnostic as virtual lines. If there is that much information to display, I'd prefer to suppress the virtual lines and open the diagnostics in a floating window instead. People will have different opinions about this, but we should provide the flexibility to configure this to meet people's different needs.

Note that the `virtual_text` handler already truncates diagnostics. It only shows the first diagnostic on a line, and even then it only shows the first line of that diagnostic. So it's not like this is unprecedented.

WhyNotHugo:
> I never want to see half my screen taken up by virt lines as in that example above. Frankly, I don't even really want to see more than one diagnostic as virtual lines.

I find this situation more useful that having to opt between truncating surrounding code (floating_text) or truncating diagnostics and showing only one (virtual_text).

That's the whole reason that virtual_lines came to be: as a way to read the entire diagnostic while not obstructing the surrounding code.

> Then perhaps a knob to not show virtual lines at all (instead of truncating) if there are more than N lines.

Generally, I've found it more useful to manually toggle diagnostics. This is why the recommendation for lsp_lines was to set a convenient mapping to do this quickly. Sometimes I have 12 lines of diagnostics on screen and need to read them to understand what's up, but sometimes I have just 1 or 2 but don't care about them because I'm in the middle of some refactor.

With knobs such as "hide lines if they are more than N", the only way to read the whole diagnostic is to change this value of `N`.

WhyNotHugo:
Sorry for the long message; couldn't figure out how to summarise that better.

MariaSolOs:
I agree with @WhyNotHugo regarding showing the full diagnostic.

> Note that the `virtual_text` handler already truncates diagnostics. It only shows the first diagnostic on a line, and even then it only shows the first line of that diagnostic. So it's not like this is unprecedented.

Issues like https://github.com/neovim/neovim/issues/16545 show that often people want to see more of the diagnostics, not a truncated version.

It will be hard to please everyone of course, but one can easily override the handler and pass diagnostics that have truncated messages.

MariaSolOs:
> Sorry for the long message; couldn't figure out how to summarise that better.

Not long at all, this was very well-put.

MariaSolOs:
As agreed on Matrix, I've added a `format` option to let users figure out how to handle particularly verbose servers.

WhyNotHugo:
> As agreed on Matrix, I've added a format option to let users figure out how to handle particularly verbose servers.

Some folks have also requested the ability to add an icon to diagnostics. The `format` feature enables this functionality too, without the need to explicitly implement it inside the handler.

MariaSolOs:
@gpanders I've addressed all your feedback (thanks for that btw <3). Lmk if you still feel like something's missing/not great here.

runtime/lua/vim/diagnostic.lua
```diff
@@ -101,6 +101,7 @@ end
 --- @field update_in_insert boolean
 --- @field underline vim.diagnostic.Opts.Underline
 --- @field virtual_text vim.diagnostic.Opts.VirtualText
+--- @field virtual_lines vim.diagnostic.Opts.VirtualLines
```
gpanders, line 104: We also need to add a `virtual_lines` entry to `vim.diagnostic.Opts` (after line 74)

runtime/lua/vim/diagnostic.lua
```diff
@@ -228,6 +229,16 @@ end
...
 
+--- @class vim.diagnostic.Opts.VirtualLines
+---
+--- Only show diagnostics for the current line.
+--- (default: `false`)
+--- @field current_line? boolean
+---
+--- A function that takes a diagnostic as input and returns a string.
+--- The return value is the text used to display the diagnostic.
+--- @field format? fun(diagnostic:vim.Diagnostic): string
```
gpanders, line 240: Doesn't need to be done in this PR, but I think we should change the semantics of the `format` function to allow it to return a `nil` value, where `nil` means the diagnostic is not displayed by that handler.

This could be used to support the "best fit" approach that @mfussenegger mentioned (e.g. by using a `format` function for both virtual text and virtual lines that chooses when and how a given diagnostic should be displayed).

runtime/lua/vim/diagnostic.lua
```diff
@@ -1603,6 +1616,271 @@ M.handlers.virtual_text = {
...
+---@param opts vim.diagnostic.Opts.VirtualLines
+local function render_virtual_lines(namespace, bufnr, diagnostics, opts)
+  if not api.nvim_buf_is_loaded(bufnr) then
+    return
+  end
+
+  vim.validate('namespace', namespace, 'number')
+  vim.validate('bufnr', bufnr, 'number')
+  vim.validate('diagnostics', diagnostics, vim.islist, 'a list of diagnostics')
+  vim.validate('opts', opts, 'table', true)
```
gpanders, line 1651: We shouldn't need to validate in a local (internal) function. We have full control over the inputs to these functions so should assume all parameters are valid. Validation should only occur in public/user-facing functions.

We can also assume that the buffer is loaded at this point so can skip the `buf_is_loaded` check.

runtime/lua/vim/diagnostic.lua
```diff
@@ -228,6 +229,16 @@ end
 --- See |nvim_buf_set_extmark()|.
 --- @field virt_text_hide? boolean
 
+--- @class vim.diagnostic.Opts.VirtualLines
+---
+--- Only show diagnostics for the current line.
+--- (default: `false`)
+--- @field current_line? boolean
```
gpanders, line 236: It'd be nice to port this option to the `virtual_text` handler too (in a different PR)

runtime/lua/vim/diagnostic.lua
```diff
@@ -1603,6 +1616,271 @@ M.handlers.virtual_text = {
   end,
 }
 
+local virtual_lines_augroup =
+  api.nvim_create_augroup('vim_lsp_diagnostic_virt_lines', { clear = true })
```
gpanders, line 1620: Rather than defining a single augroup at the top-level scope, would it make sense to make the augroup ID a field of `ns.user_data` in the same way that we do for the extmark's namespace?

runtime/lua/vim/diagnostic.lua
MariaSolOs: Sounds good. I'll make sure to do this in a follow-up PR.

runtime/lua/vim/diagnostic.lua
MariaSolOs: Good idea. I'll do it after merging this :)

runtime/lua/vim/diagnostic.lua
MariaSolOs: I guess we could. This is what `lsp_lines` does. @WhyNotHugo is there a reason why [this `LspLines` augroup](https://git.sr.ht/~whynothugo/lsp_lines.nvim/tree/a92c755f182b89ea91bd8a6a2227208026f27b4d/item/lua/lsp_lines/init.lua#L32) is created at the top level?

runtime/lua/vim/diagnostic.lua
WhyNotHugo: > I think we should change the semantics of the format function to allow it to return a nil value, where nil means the diagnostic is not displayed by that handler.

Nice idea, this could also be used for situations like "I want to hide virtual_lines in specific buffers or filetypes".

runtime/lua/vim/diagnostic.lua
WhyNotHugo: No reason, this was simply the easiest way to do it

MariaSolOs:
@gpanders all of your latest feedback has been addressed in https://github.com/neovim/neovim/compare/5e7e46032ce95a9189987e0799d482bd584423d8..31c78507794ecdb136ad16bb987b2dbcc50ed2a6

runtime/lua/vim/diagnostic.lua
MariaSolOs: https://github.com/neovim/neovim/pull/32187

runtime/doc/diagnostic.txt
```diff
@@ -97,8 +97,8 @@ If a diagnostic handler is configured with a "severity" key then the list of
 diagnostics passed to that handler will be filtered using the value of that
 key (see example below).
 
-Nvim provides these handlers by default: "virtual_text", "signs", and
-"underline".
+Nvim provides these handlers by default: "virtual_text", "virtual_lines",
+"signs", and "underline".
```
justinmk, line 101: In `:help nvim_buf_set_extmark()` the related fields are named `virt_text`, `virt_lines`. Do we want to be consistent (though that would imply deprecating `virtual_text` here and aliasing it to `virt_text`)?

MariaSolOs: Yeah. I find consistency with the existing handler's name preferable, and I personally don't find the hassle of renaming (and deprecating) `virtual_text` worth it. Letting others weigh in though.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1603,6 +1620,261 @@ M.handlers.virtual_text = {
...
+    end
+
+    local ns = M.get_namespace(namespace)
+    if not ns.user_data.virt_lines_ns then
+      ns.user_data.virt_lines_ns =
+        api.nvim_create_namespace(string.format('nvim.%s.diagnostic.virtual_lines', ns.name))
+    end
+    if not ns.user_data.virt_lines_augroup then
+      ns.user_data.virt_lines_augroup =
+        api.nvim_create_augroup('nvim.lsp.diagnostic.virt_lines', { clear = true })
```
gpanders, line 1844: This is going to cause the augroup to be re-created (and cleared) multiple times (once for every namespace). The augroup name should be unique per namespace, similar to the extmark namespace name.

MariaSolOs: That's a good idea. But maybe I'll use `handlers.virtual_lines._augroup` just to indicate that it's part of our internal implementation. WDYT?

gpanders: That sounds fine.

gpanders (approved):

MariaSolOs:
@gpanders note that I just pushed a new commit where I use `save_extmarks` to follow what the `underline` and `virtual_text` handlers do.

MariaSolOs:
@justinmk @gpanders @mfussenegger I think this is ready to go :)

divaltor:
Hello, thanks for PR. It would be nice to have possibility to automatically hide virtual line if there is only 1 error on that line and display it as virtual text instead.

Examples in Helix:
image
image

WhyNotHugo:
I guess https://github.com/neovim/neovim/issues/31000 still applies, given that it's a big in how extmarks are rendered.

gegoune:
> I really like the `current_line` option. Is there a way to debounce it? So when doing `j` in a file with many warnings, it does not expand and shrink in every line?

On top of that I would like to not debounce jumping to diagnostic with `vim.diagnostic.jump()' should this (debouncing for `j/k`, etc) ever be added.

MariaSolOs:
> > I really like the `current_line` option. Is there a way to debounce it? So when doing `j` in a file with many warnings, it does not expand and shrink in every line?
> 
> On top of that I would like to not debounce jumping to diagnostic with `vim.diagnostic.jump()` should this (debouncing for `j/k`, etc) ever be added.

@gegoune @leiserfg Please create issues for these. Commenting on PRs is not the best way to ask for a feature request.

PR #31974, opened 2025-01-11
perf(treesitter): cache queries strongly

### LINKED ISSUE #31969: Treesitter query cache is cleared too soon
### Problem

Some Treesitter queries take a really long time to parse. For example, `vim.treesitter.query.get("haskell", "highlights")` can take ~350ms to complete. Although `query.get(...)` [uses memoize to cache the result](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/treesitter/query.lua#L274), the cache is extremely weak. The weak table is [implemented using "kv" mode](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/func/_memoize.lua#L42) which means that the key and the value have to be kept alive or else the entry is garbage collected. But because the key is a string created by the memoize function, there's no way to keep the key alive which means the cache is largely useless and only lives for one gc cycle (if I'm understanding the implementation correctly) .

Could Neovim instead pass `false` to the memoize function's `weak` parameter so that the table contains strong references? That would make the cache useful and remove tons of stuttering when plugins+Neovim constantly try to run Haskell Treesitter queries. I haven't looked into how much additional memory this would use, but for extremely slow queries like Haskell's highlighter, this would make using Neovim much more pleasant.

### Steps to reproduce

Run the following lua code in Neovim:
```lua
local start_time = vim.uv.hrtime()
-- Attempt to keep the cache alive by storing the result in a global. That doesn't work though because the key can still get garbage collected.
First_run = vim.treesitter.query.get('haskell', 'highlights')
print((vim.uv.hrtime() - start_time) / 1e6)
collectgarbage()
start_time = vim.uv.hrtime()
vim.treesitter.query.get('haskell', 'highlights')
print((vim.uv.hrtime() - start_time) / 1e6)
```

### Expected behavior

The first number printed should be large, but the second number should be much smaller. That's the case if you remove the `collectgarbage` call, but with the `collectgarbage` call they both print large numbers.

### Nvim version (nvim -v)

NVIM v0.10.3 Build type: Release LuaJIT 2.1.1713773202

N/A

### Operating system/version

Nixos 25.05beta732261.6df24922a140

### Terminal name/version

N/A, but I'm using Alacritty

alacritty

### Installation

System package manager

ribru17:
**Problem:** Query parsing uses a weak cache which is invalidated frequently

**Solution:** Make the cache strong, and invalidate it manually when necessary (that is, when `rtp` is changed or `query.set()` is called)

Closes #31969

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -292,6 +292,9 @@ PERFORMANCE
   inflight requests). This greatly improves performance with slow LSP servers.
 • 10x speedup for |vim.treesitter.foldexpr()| (when no parser exists for the
   buffer).
+• Strong |treesitter-query| caching makes repeat  |vim.treesitter.query.get()|
+  and |vim.treesitter.query.parse()| calls significantly faster for large
+  queries.
 
 PLUGINS
 
--- a/runtime/lua/vim/treesitter/query.lua
+++ b/runtime/lua/vim/treesitter/query.lua
@@ -262,6 +262,7 @@ local explicit_queries = setmetatable({}, {
 ---@param query_name string Name of the query (e.g., "highlights")
 ---@param text string Query text (unparsed).
 function M.set(lang, query_name, text)
+  M.get:clear(lang, query_name)
   explicit_queries[lang][query_name] = M.parse(lang, text)
 end
 
@@ -284,7 +285,15 @@ M.get = memoize('concat-2', function(lang, query_name)
   end
 
   return M.parse(lang, query_string)
-end)
+end, false)
+
+api.nvim_create_autocmd('OptionSet', {
+  pattern = { 'runtimepath' },
+  group = api.nvim_create_augroup('ts_query_cache_reset', { clear = true }),
+  callback = function()
+    M.get:clear()
+  end,
+})
 
 --- Parses a {query} string and returns a `Query` object (|lua-treesitter-query|), which can be used
 --- to search the tree for the query patterns (via |Query:iter_captures()|, |Query:iter_matches()|),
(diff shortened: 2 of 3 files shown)


lewis6991 (approved):

runtime/doc/news.txt
```diff
@@ -292,6 +292,9 @@ PERFORMANCE
   inflight requests). This greatly improves performance with slow LSP servers.
 • 10x speedup for |vim.treesitter.foldexpr()| (when no parser exists for the
   buffer).
+• The |treesitter-query| parsing cache is now strong, so
+  |vim.treesitter.query.get()| and |vim.treesitter.query.parse()| are now much
+  quicker, consistently.
```
clason, line 297:
```suggestion
• Stronger |treesitter-query| caching makes repeat  |vim.treesitter.query.get()| and |vim.treesitter.query.parse()| calls significantly faster for large queries.
```

clason:
@MariaSolOs This impacts the PR to allow extending queries via `query.set()`.

runtime/lua/vim/treesitter/query.lua
```diff
@@ -284,7 +285,15 @@ M.get = memoize('concat-2', function(lang, query_name)
   end
 
   return M.parse(lang, query_string)
-end)
+end, false)
+
+api.nvim_create_autocmd('OptionSet', {
+  pattern = { 'runtimepath' },
+  group = api.nvim_create_augroup('ts_query_cache_reset', { clear = true }),
```
justinmk, line 292: For namespaces let's choose a consistent convention, I suggest `nvim.foo.bar` instead of undescores/slashes/whatever else we're doing.

PR #31981, opened 2025-01-12
refactor: simplify bump_deps.lua

dundargoc:
Simplify usage and remove redundant flags and code.

### DIFF
--- a/scripts/bump_deps.lua
+++ b/scripts/bump_deps.lua
@@ -3,40 +3,45 @@
 -- Usage:
 --    ./scripts/bump_deps.lua -h
 
-local M = {}
+assert(vim.fn.executable('sed') == 1)
 
-local _trace = false
 local required_branch_prefix = 'bump-'
 local commit_prefix = 'build(deps): '
 
--- Print message
-local function p(s)
-  vim.cmd('set verbose=1')
-  vim.api.nvim_echo({ { s, '' } }, false, {})
-  vim.cmd('set verbose=0')
+local repos = {
+  'luajit/luajit',
+  'libuv/libuv',
+  'luvit/luv',
+  'neovim/unibilium',
+  'juliastrings/utf8proc',
+  'tree-sitter/tree-sitter',
+  'tree-sitter/tree-sitter-c',
+  'tree-sitter-grammars/tree-sitter-lua',
+  'tree-sitter-grammars/tree-sitter-vim',
+  'neovim/tree-sitter-vimdoc',
+  'tree-sitter-grammars/tree-sitter-query',
+  'tree-sitter-grammars/tree-sitter-markdown',
+  'bytecodealliance/wasmtime',
+  'uncrustify/uncrustify',
+}
+
+local dependency_table = {} --- @type table<string, string>
+for _, repo in pairs(repos) do
+  dependency_table[vim.fs.basename(repo)] = repo
 end
 
-local function die()
-  p('')
+local function die(msg)
+  print(msg)
   vim.cmd('cquit 1')
 end
 
 -- Executes and returns the output of `cmd`, or nil on failure.
 -- if die_on_fail is true, process dies with die_msg on failure
---
--- Prints `cmd` if `trace` is enabled.
 local function _run(cmd, die_on_fail, die_msg)
-  if _trace then
-    p('run: ' .. vim.inspect(cmd))
-  end
-  local rv = vim.trim(vim.fn.system(cmd)) or ''
+  local rv = vim.trim(vim.system(cmd, { text = true }):wait().stdout) or ''
   if vim.v.shell_error ~= 0 then
     if die_on_fail then
-      if _trace then
-        p(rv)
-      end
-      p(die_msg)
-      die()
+      die(die_msg)
     end
     return nil
   end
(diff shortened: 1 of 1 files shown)


clason:
* move out the dependency table to a top-level table
* remove all indirections
* align behavior with `vim-patch` (create new branch unless current branch starts with `bump-`, remove `--branch` switch, don't special-case `master`)

scripts/bump_deps.lua
```diff
@@ -136,23 +125,25 @@ local function get_dependency(dependency_name)
 end
 
 local function get_gh_commit_sha(repo, ref)
-  require_executable('gh')
+  assert(vim.fn.executable('gh') == 1)
```
clason, line 128: Just do these checks once at the top of the file.

scripts/bump_deps.lua
```diff
@@ -200,268 +145,86 @@ local function verify_branch(new_branch_suffix)
...
         --dep <dependency>    bump to a specific release or tag.
 
     Dependency Options:
-        --version <tag>       bump to a specific release or tag.
-        --commit <hash>       bump to a specific commit.
-        --HEAD                bump to a current head.
+        --ref <ref>           bump to a specific commit or tag.
+        --head                bump to a current head.
 
         <dependency> is one of:
```
clason, line 198: I would just add this as a `--list` switch, not bake this into the boilerplate.

clason: For bonus points (not in this PR), `--list` could show the latest and tracked versions...

scripts/bump_deps.lua
```diff
@@ -178,19 +116,25 @@ local function write_cmakelists_line(symbol, kind, value)
   }, 'Failed to write ' .. deps_file)
 end
 
-local function explicit_create_branch(dep)
-  require_executable('git')
+local function ref(dependency_name, _ref)
+  -- Allow all variations of tree-sitter, treesitter, tree_sitter_c, tree_sitter-c etc
```
clason, line 120: nah, that's not necessary. all deps are lower-case, and if you are unsure, just use `--list`. This iS n_ot NI_M.

dundargoc: I meant more variations between: `treesitter` vs `tree-sitter` as well as `treesitter-c` vs `treesitter_c`. If you still think it's overkill I'll remove it

clason: I still think it's overkill.

scripts/bump_deps.lua
```diff
@@ -178,290 +106,106 @@ local function write_cmakelists_line(symbol, kind, value)
...
     Usage:  nvim -l %s [options]
         Bump to HEAD, tagged version, commit, or branch:
-            nvim -l %s --dep Luv --head
-            nvim -l %s --dep Luv --version 1.43.0-0
-            nvim -l %s --dep Luv --commit abc123
-            nvim -l %s --dep Luv --branch
-        Create a PR:
-            nvim -l %s --pr
+            nvim -l %s luv --head
+            nvim -l %s luv --ref 1.43.0-0
```
clason, line 159:
```suggestion
            %s luv --ref 1.43.0-0
```
And so on (yay shebang!)

scripts/bump_deps.lua
```diff
@@ -3,21 +3,38 @@
 -- Usage:
 --    ./scripts/bump_deps.lua -h
 
-local M = {}
+assert(vim.fn.executable('curl') == 1)
+assert(vim.fn.executable('sed') == 1)
+assert(vim.fn.executable('git') == 1)
 
 local _trace = false
```
clason, line 10: Is that still needed?

dundargoc: Technically yes, practically no. We can absolutely remove `curl` since it's available everywhere. `git` should be given. I guess we can keep `sed` as it would be a surprise for windows users?

dundargoc: Should I remove sed too?

clason: Sorry, I was talking about the `_trace` line.

The sed check can stay; it's the only command where failure would leave stale files around.

dundargoc: Oh, the trace. Eh, I guess it could be useful? to see which run commands are used? Idk, Not super attached to it.

clason: As you can't set it from the outside, I would say "no". More code to yeet :)

scripts/bump_deps.lua
```diff
@@ -178,290 +94,107 @@ local function write_cmakelists_line(symbol, kind, value)
...
-  local this_script = _G.arg[0]:match('[^/]*.lua$')
-  print(([=[
+  local this_script = tostring(vim.fs.basename(_G.arg[0]))
+  local script_exe = './' .. this_script
+  local help = ([=[
     Bump Nvim dependencies
 
-    Usage:  nvim -l %s [options]
+    Usage:  %s [options]
         Bump to HEAD, tagged version, commit, or branch:
```
clason, line 146: The "branch" variant is not documented (and implemented?) It's not really needed, though.

* If it already works, just add a corresponding line to the example below;
* if not, just remove "or branch" from this line.

dundargoc: It's implemented (since it's just a ref). Let's remove it from docs if it's not needed.

clason (approved):
Nice cleanup!

PR #25612, opened 2023-10-12
fix(lsp): handle NUL bytes in popup text

zeertzjq:
Fix #25610

PR #31990, opened 2025-01-13
fix(lsp): ensure minimum height for floating popup

Xuyuanp:
For a long time, I’ve been quite annoyed because the floating window for hover and signature help always cuts off a few lines, forcing me to jump into the floating window to see the full documentation. Recently, I finally had the time to investigate the issue and found out that the `_make_floating_popup_size` function counts empty lines as having zero height.

before:
image
after:
image

Related: #25612


justinmk (approved):

MariaSolOs (approved):

zeertzjq:
backport?

PR #31993, opened 2025-01-13
fix(float): nvim_win_set_config cannot set title/footer independently

### LINKED ISSUE #31992: The `vim.api.nvim_win_set_config` function does not work as documented.
### Problem

The `title` and `footer` cannot be set independently. When only `title` or `footer` is set, `title_pos/footer_pos` are reset to their default value of `left`. 
This behavior does not align with the documentation, which states: 
> When reconfiguring a window, absent option keys will not be changed."

### Steps to reproduce

```lua

vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, { 'Hello, world!' })

---@type vim.api.keyset.win_config
local win_config = {
    relative = 'cursor',
    style = 'minimal',
    width = 20,
    height = 10,
    row = 0,
    col = 0,
    title = 'old',
    title_pos = 'center',
    border = 'rounded',
}

local winnr = vim.api.nvim_open_win(bufnr, true, win_config)

vim.api.nvim_win_set_config(winnr, { title = 'new' })
```

Run this code, and you'll find that the title has been aligned to the left.

### Expected behavior

The new title should remain centered.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1537+gf4cff3077b

no

### Operating system/version

macOS 14.6

### Terminal name/version

ghostty 1.0.1

xterm-256color

### Installation

homebrew

glepnir:
Problem: reset win config doesn't take before title footer pos value

Solution: reuse before titile footer pos value

Fix #31992

### DIFF
--- a/src/nvim/api/win_config.c
+++ b/src/nvim/api/win_config.c
@@ -895,7 +895,7 @@ static void parse_bordertext(Object bordertext, BorderTextType bordertext_type, 
   *is_present = true;
 }
 
-static bool parse_bordertext_pos(String bordertext_pos, BorderTextType bordertext_type,
+static bool parse_bordertext_pos(win_T *wp, String bordertext_pos, BorderTextType bordertext_type,
                                  WinConfig *fconfig, Error *err)
 {
   AlignTextPos *align;
@@ -909,7 +909,9 @@ static bool parse_bordertext_pos(String bordertext_pos, BorderTextType bordertex
   }
 
   if (bordertext_pos.size == 0) {
-    *align = kAlignLeft;
+    if (!wp) {
+      *align = kAlignLeft;
+    }
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


src/nvim/api/win_config.c
```diff
@@ -901,7 +901,7 @@ static bool parse_bordertext_pos(String bordertext_pos, BorderTextType bordertex
   }
 
   if (bordertext_pos.size == 0) {
-    *align = kAlignLeft;
+    *align = bordertext_type == kBorderTextTitle ? fconfig->title_pos : fconfig->footer_pos;
```
zeertzjq, line 904: This assignment doesn't do anything. The value is `*align` is already equal to RHS.

justinmk:
The problem description is hard to understand. The linked issue gives a clear problem. Please add that here. 

Also worth a rebase.

(Future: Is there any way to make these tests less verbose? Perhaps by a shared function?)

justinmk (approved):

PR #32001, opened 2025-01-14
refactor: use `nvim.foo.bar` format for namespace/augroup names

MariaSolOs:
Addresses https://github.com/neovim/neovim/pull/31959#discussion_r1912810198.


runtime/filetype.lua
```diff
@@ -3,7 +3,7 @@ if vim.g.did_load_filetypes then
 end
 vim.g.did_load_filetypes = 1
 
-vim.api.nvim_create_augroup('filetypedetect', { clear = false })
+vim.api.nvim_create_augroup('nvim.filetypedetect', { clear = false })
 
 vim.api.nvim_create_autocmd({ 'BufRead', 'BufNewFile', 'StdinReadPost' }, {
   group = 'filetypedetect',
```
lucario387, line 9: This should match the groupname above

MariaSolOs: I decided to not modify this group after all. Docs and tests rely on this name and it seems to come from vim.

lewis6991 (approved):

justinmk (approved):

yochem:
Maybe also: https://github.com/neovim/neovim/blob/e8ddb7a46938f8843abc1c321cfd83cee2ba0020/runtime/plugin/shada.vim#L6

MariaSolOs:
> Maybe also:
> 
> https://github.com/neovim/neovim/blob/e8ddb7a46938f8843abc1c321cfd83cee2ba0020/runtime/plugin/shada.vim#L6

@yochem what about it?

yochem:
Whoops, sorry if I was not clear. Shouldn't that also be included in this PR, and changed to "nvim.shadacommands" or "nvim.shada"? Or does that only work with with augroups defined in lua code?

MariaSolOs:
> Whoops, sorry if I was not clear. Shouldn't that also be included in this PR, and changed to "nvim.shadacommands" or "nvim.shada"? Or does that only work with with augroups defined in lua code?

I think that @justinmk's original suggestion was limited to the names we create via `nvim_create_augroup` and `nvim_create_namespace` (in Lua). I think it would be weird to rename this pascal-cased vim names, but I'll let others weigh in.

justinmk:
> What I think is more important is that all the groups created by core have a consistent naming scheme.

Agreed, but let's skip it to unblock this PR, since it at least sets a pattern for Lua, where most activity is happening anyway.

Some tests are failing, just need to update them then let's merge this.

MariaSolOs:
> Agreed, but let's ignore vimscript / legacy vim augroup names for now, to unblock this PR, since it sets a pattern for Lua, where most activity is happening anyway.
> 
> Some tests are failing, just need to update them then let's merge this.

Yep yep, pardon my skill issues. Will get to it after work today :)

MariaSolOs:
idk why but [this line](https://github.com/neovim/neovim/blob/611ef354919f1c6564efd2ff8074545941458ccc/runtime/lua/vim/treesitter/highlighter.lua#L163) fails (with an 'invalid group' error) if I change the autocommand name to `nvim.syntaxset` (and update both line 144 and 163). Leaving it as is for now, but

MariaSolOs:
Okay help wanted to figure out why the menu popup tests are still failing

justinmk:
> but [this line](https://github.com/neovim/neovim/blob/611ef354919f1c6564efd2ff8074545941458ccc/runtime/lua/vim/treesitter/highlighter.lua#L163) fails (with an 'invalid group' error) if I change the autocommand name to `nvim.syntaxset` (and update both line 144 and 163). Leaving it as is for now.

👍 that's a legacy augroup defined here (and elsewhere): https://github.com/neovim/neovim/blob/47866cd8d20c62afa8a3c3929d3aada2db9162f5/runtime/syntax/manual.vim#L22

PR #32002, opened 2025-01-14
fix(lsp): make passing false to vim.lsp.enable() disable given LSP(s)

andrewbraxton:
Problem:
Per the documentation, passing `false` as the `enable` parameter of
`vim.lsp.enable()` should disable the given LSP(s), but it does not work
due to a logic error.

Specifically, `enable == false and nil or {}` will always evaluate to
`{}` because `nil` is falsy.

Solution:
Correct the conditional statement.

### DIFF
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -546,7 +546,7 @@ function lsp.enable(name, enable)
     if nm == '*' then
       error('Invalid name')
     end
-    lsp._enabled_configs[nm] = enable == false and nil or {}
+    lsp._enabled_configs[nm] = enable ~= false and {} or nil
   end
 
   if not next(lsp._enabled_configs) then


MariaSolOs (approved):

runtime/lua/vim/lsp.lua
```diff
@@ -544,7 +544,11 @@ function lsp.enable(name, enable)
     if nm == '*' then
       error('Invalid name')
     end
-    lsp._enabled_configs[nm] = enable == false and nil or {}
+    if enable == false then
+      lsp._enabled_configs[nm] = nil
+    else
+      lsp._enabled_configs[nm] = {}
+    end
```
justinmk, line 551:
```suggestion
    lsp._enabled_configs[nm] = enable ~= false and {} or nil
```

justinmk:
Thanks for the fix. A test would be good too :)

Issue #31954, opened 2025-01-10
Expose `vim.hl.on_yank()` blink functionality


lsvmello:
### Problem

`vim.hl.on_yank()` has a very nice implementation of a visual hint that might be useful for plugins or even the newly introduced `yxx` keymap (https://github.com/neovim/neovim/pull/31904).

### Expected behavior

Add a new `timeout` option to `vim.hl.range` or maybe create a new `vim.hl.blink` function for it.

justinmk:
> Add a new `timeout` option to `vim.hl.range`

That sounds like a good direction

siddhantdev:
I would like to work on this. Where are these functions (`vim.hl.on_yank()`, `vim.hl.range`) implemented in the code?

craigmac:
runtime/lua/vim/hl.lua

PR #32012, opened 2025-01-14
feat: expose vim.hl.on_yank() blink functionality

siddhantdev:
Fix for https://github.com/neovim/neovim/issues/31954

The code works but I don't know how to write tests for it, so I have marked the PR as a draft for now.

### DIFF
--- a/runtime/doc/lua.txt
+++ b/runtime/doc/lua.txt
@@ -685,6 +685,8 @@ vim.hl.range({bufnr}, {ns}, {higroup}, {start}, {finish}, {opts})
                      whether the range is end-inclusive
                    • {priority}? (`integer`, default:
                      `vim.hl.priorities.user`) Highlight priority
+                   • {timeout}? (`integer`, default: -1 no timeout) Time in ms
+                     before highlight is cleared
 
 
 ==============================================================================
--- a/runtime/lua/vim/hl.lua
+++ b/runtime/lua/vim/hl.lua
@@ -17,6 +17,9 @@ M.priorities = {
   user = 200,
 }
 
+local range_timer --- @type uv.uv_timer_t?
+local range_hl_clear --- @type fun()?
+
 --- @class vim.hl.range.Opts
 --- @inlinedoc
 ---
@@ -31,6 +34,10 @@ M.priorities = {
 --- Highlight priority
 --- (default: `vim.hl.priorities.user`)
 --- @field priority? integer
+---
+--- Time in ms before highlight is cleared
+--- (default: -1 no timeout)
+--- @field timeout? integer
 
 --- Apply highlight group to range of text.
 ---
@@ -45,6 +52,7 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
   local regtype = opts.regtype or 'v'
   local inclusive = opts.inclusive or false
   local priority = opts.priority or M.priorities.user
+  local timeout = opts.timeout or -1
 
   local v_maxcol = vim.v.maxcol
 
@@ -100,6 +108,19 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
     end
   end
 
+  if range_timer and not range_timer:is_closing() then
+    range_timer:close()
+    assert(range_hl_clear)
+    range_hl_clear()
+  end
+
+  range_hl_clear = function()
+    range_timer = nil
+    range_hl_clear = nil
+    pcall(vim.api.nvim_buf_clear_namespace, bufnr, ns, 0, -1)
+    pcall(vim.api.nvim__ns_set, { wins = {} })
+  end
+
   for _, res in ipairs(region) do
     local start_row = res[1][2] - 1
     local start_col = res[1][3] - 1
@@ -113,11 +134,13 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
       strict = false,
     })
   end
+
+  if timeout ~= -1 then
+    range_timer = vim.defer_fn(range_hl_clear, timeout)
+  end
 end
 
 local yank_ns = api.nvim_create_namespace('nvim.hlyank')
-local yank_timer --- @type uv.uv_timer_t?
-local yank_cancel --- @type fun()?
 
 --- Highlight the yanked text during a |TextYankPost| event.
 ---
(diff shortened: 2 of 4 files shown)


clason:
You should refactor the `hl.on_yank()` function to use the new option rather than set its own timer.

siddhantdev:
Done

runtime/lua/vim/hl.lua
```diff
@@ -113,11 +121,26 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
       strict = false,
     })
   end
+
+  if timeout ~= -1 then
+    print(timeout)
```
jakubbortlik, line 126: Is this `print` intentional or just a forgotten debug print?

siddhantdev: Oh no! That was left by mistake. Fixed it.

runtime/lua/vim/hl.lua
```diff
@@ -100,6 +107,19 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
     end
   end
 
+  if range_timer then
```
glepnir, line 110:
```suggestion
  if range_timer and not range_timer:closing() then
```

runtime/lua/vim/hl.lua
```diff
@@ -156,27 +178,14 @@ function M.on_yank(opts)
...
-    assert(yank_cancel)
-    yank_cancel()
-  end
 
   vim.api.nvim__ns_set(yank_ns, { wins = { winid } })
   M.range(bufnr, yank_ns, higroup, "'[", "']", {
     regtype = event.regtype,
     inclusive = event.inclusive,
     priority = opts.priority or M.priorities.user,
+    timeout = timeout,
```
clason, line 187: just inline the local here

runtime/lua/vim/hl.lua
```diff
@@ -31,6 +33,10 @@ M.priorities = {
 --- Highlight priority
 --- (default: `vim.hl.priorities.user`)
 --- @field priority? integer
+---
+--- Time in ms before highlight is cleared
+--- (default: -1 (inactive))
```
clason, line 38:
```suggestion
--- (default: -1 (no timeout))
```

runtime/lua/vim/hl.lua
glepnir: aha typo it's `is_closing` safely check timer is not closing ..

justinmk:
Needs a test in `test/functional/lua/hl_spec.lua`. You can [run the test](https://github.com/neovim/neovim/blob/master/test/README.md#filter-by-file) like this:

    TEST_FILE=test/functional/lua/hl_spec.lua make functionaltest

runtime/lua/vim/hl.lua
```diff
@@ -17,6 +17,8 @@ M.priorities = {
   user = 200,
 }
 
+local range_timer --- @type uv.uv_timer_t?
+local range_hl_clear --- @type fun()?
 --- @class vim.hl.range.Opts
```
justinmk, line 22: add a blank line above this.

siddhantdev:
For the test, I have written this:

```lua
  it('correctly removes highlight after timeout time', function ()
    local timeout = 1000
    exec_lua(function ()
      local ns = vim.api.nvim_create_namespace('')
      vim.hl.range(0, ns, 'Search', { 0, 0 }, { 4, 0 }, { timeout = timeout })
    end)
    screen:expect([[
      {10:^asdfghjkl}{100:$}                                                  |
      {10:«口=口»}{100:$}                                                    |
      {10:qwertyuiop}{100:$}                                                 |
      {10:口口=口口}{1:$}                                                  |
      zxcvbnm{1:$}                                                    |
                                                                  |
    ]])
    vim.uv.sleep(timeout)
    screen:expect([[
      ^asdfghjkl{1:$}                                                  |
      «口=口»{1:$}                                                    |
      qwertyuiop{1:$}                                                 |
      口口=口口{1:$}                                                  |
      zxcvbnm{1:$}                                                    |
                                                                  |
    ]])
  end)
```

The `screen:expect` calls have been filled in using `screen:snapshot_util()`. I'm not sure if using `vim.uv.sleep` is the best way to check if the timeout works but I couldn't think of any other way. `screen:expect` has a timeout option, but setting `{ timeout = timeout }` after the string doesn't seem to work.

justinmk:
> I'm not sure if using `vim.uv.sleep` is the best way to check if the timeout works but I couldn't think of any other way.

`screen:expect` will wait for the expected screen state. sleep() shouldn't be needed

siddhantdev:
So do I put just one `screen:expect call for the final state, or two: one for the initial highlight and then one for after the highlight is cleared?

justinmk:
> two: one for the initial highlight and then one after that to check that the highlight is cleared?

that's the usual recommended way.

test/functional/lua/hl_spec.lua
```diff
@@ -104,6 +104,30 @@ describe('vim.hl.range', function()
                                                                   |
     ]])
   end)
+
+  it('correctly removes highlight after timeout time', function ()
```
justinmk, line 108: "correctly" is never needed in test names.
```suggestion
  it('removes highlight after given `timeout`', function ()
```

runtime/doc/lua.txt
```diff
@@ -685,6 +685,8 @@ vim.hl.range({bufnr}, {ns}, {higroup}, {start}, {finish}, {opts})
                      whether the range is end-inclusive
                    • {priority}? (`integer`, default:
                      `vim.hl.priorities.user`) Highlight priority
+                   • {timeout}? (`integer`, default: -1 (no timeout) Time in
+                     ms before highlight is cleared)
```
justinmk, line 689: docs generator did something ... interesting... here :)

runtime/lua/vim/hl.lua
```diff
@@ -100,6 +108,19 @@ function M.range(bufnr, ns, higroup, start, finish, opts)
     end
   end
 
+  if range_timer and not range_timer:is_closing() then
+    range_timer:close()
+    assert(range_hl_clear)
+    range_hl_clear()
+  end
+
+  range_hl_clear = function()
```
justinmk, line 117: Correct me if wrong, but it looks like this logic will only support *one* timed highlight at any given time. 

That's probably fine for now, I want to many more revisions to this PR. But may need a followup. The "one timed highlight only" concept is specific to the `.on_yank()` feature. Whereas for `.range()` should presumably allow the caller to control stuff, by returning a handle of some sort.

justinmk (approved):
LGTM after comments are addressed, and after the news.txt CI is resolved.

runtime/lua/vim/hl.lua
siddhantdev: Yeah, currently it only supports one timed highlight at a time.

PR #32013, opened 2025-01-14
feat(messages): add :!cmd shell message kinds

### LINKED ISSUE #31901: messages: add kind(s) for external command output
### Problem

For example: assuming the repro described below, `:!echo -e "1\n2"` gives

```lua
{
  c = { { 0, ':!echo -e "1\\n2"\r\n', 0 } },
  k = ""
}
{
  c = { { 0, "1\n2\n", 0 } },
  k = ""
}
{
  c = { { 0, "\n", 0 } },
  k = ""
}
```

### Steps to reproduce

`init.lua`:
```lua
local api = vim.api

local function log(msg)
  vim.cmd.normal 'G'
end

local function detach()
  vim.ui_detach(ns)
end

local qMsgs = {}

local function consumeQMsgs()
    log(vim.inspect(qmsg))
  end
  qMsgs = {}
end

local redirect = true

local function attach()
  if not redirect then
    return
  end
    if event ~= 'msg_show' then
      return
    end
      api.nvim_input('\r')
      return
    end

    local msg = {
      k = kind or 'none',
      c = content or 'none',
    }
      consumeQMsgs()
      log(vim.inspect(msg))
      return
    end

    qMsgs[#qMsgs + 1] = msg
    vim.schedule(function ()
      consumeQMsgs()
    end)
  end)
end

  callback = function ()
    attach()
  end
})

  redirect = not redirect
  if not redirect then
    detach()
  end
end, {nargs = 0})
```
- `nvim --clean -u init.lua`
- `:!echo -e "1\n2"`

### Expected behavior

Rather single message with not empty kind. See also https://github.com/neovim/neovim/issues/31871#issuecomment-2574695005

### Nvim version (nvim -v)

v0.11.0-dev-1495+gb67fcd0488

n/a

### Operating system/version

Debian Sid

### Terminal name/version

alacritty

alacritty

### Installation

from repo

luukvbaal:
**feat(messages): add :!cmd shell message kinds**

Also print stderr error messages with ErrorMsg highlight group.

Fix #31901

**fix(messages): verbose kind for nvim_echo()**

Problem:  No "verbose" kind for nvim_echo() opts->verbose.
Solution: Pass NULL "kind" to indicate no new kind.


luukvbaal:
Not sure if this granularity is warranted, considering we [rejected](https://github.com/neovim/neovim/pull/31279#discussion_r1850566968) more specific kinds for various "list_cmd" output. Though those messages can still be parsed for known formatting whereas shell stderr/out/return can contain anything so the distinction may be useful. Otherwise just use "shell" for all chunks.

justinmk:
> Not sure if this granularity is warranted

No strong feelings either way. Perhaps an argument against the granularity is that `:!` will become not so special once we reimplement it as a PTY wrapper. In theory it won't be any different than `:echo system(...,{pty=true})`. Though it will continue to be able to handle stdout/stderr/exitcode, so these granular kinds can continue to be supported.

runtime/doc/news.txt
```diff
@@ -96,7 +96,7 @@ EVENTS
   • `msg_show`:
     • `history` argument indicating if the message was added to the history.
     • new message kinds: "bufwrite", "completion", "list_cmd", "lua_print",
-      "search_cmd", "undo", "verbose", wildlist".
+      "search_cmd", "shell_out/err/return", "undo", "verbose", wildlist".
```
zeertzjq, line 99:
```suggestion
      "search_cmd", "shell_out/err/ret", "undo", "verbose", wildlist".
```

src/nvim/event/proc.h
```diff
@@ -21,8 +21,8 @@ static inline Proc proc_init(Loop *loop, ProcType type, void *data)
     .argv = NULL,
     .exepath = NULL,
     .in = { .closed = false },
-    .out = { .s.closed = false },
-    .err = { .s.closed = false },
+    .out = { .s.closed = false, .s.fd = 1 },
+    .err = { .s.closed = false, .s.fd = 2 },
```
justinmk, line 25: should we use `STDOUT_FILENO` / `STDERR_FILENO` ?

src/nvim/os/shell.c
```diff
@@ -1067,7 +1068,7 @@ static void out_data_ring(const char *output, size_t size)
   }
 
   if (output == NULL && size == SIZE_MAX) {   // Print mode
-    out_data_append_to_screen(last_skipped, &last_skipped_len, true);
+    out_data_append_to_screen(last_skipped, &last_skipped_len, 1, true);
```
justinmk, line 1071: STDOUT_FILENO ?

justinmk (approved):

PR #32017, opened 2025-01-14
fix(marks): revise metadata for start mark of revalidated pair

luukvbaal:
Problem:  Metadata may be revised for end mark of a revalidated pair.
Solution: Revise metadata for start mark of a revalidated pair.

### DIFF
--- a/src/nvim/extmark.c
+++ b/src/nvim/extmark.c
@@ -112,10 +112,10 @@ static void extmark_setraw(buf_T *buf, uint64_t mark, int row, colnr_T col, bool
 {
   MarkTreeIter itr[1] = { 0 };
   MTKey key = marktree_lookup(buf->b_marktree, mark, itr);
-  bool move = key.pos.row >= 0 && (key.pos.row != row || key.pos.col != col);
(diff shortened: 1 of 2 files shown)


src/nvim/extmark.c
```diff
@@ -112,10 +112,10 @@ static void extmark_setraw(buf_T *buf, uint64_t mark, int row, colnr_T col, bool
 {
   MarkTreeIter itr[1] = { 0 };
   MTKey key = marktree_lookup(buf->b_marktree, mark, itr);
-  bool move = key.pos.row >= 0 && (key.pos.row != row || key.pos.col != col);
-  // Already valid keys were being revalidated, presumably when encountering a
-  // SavePos from a modified mark. Avoid adding that to the decor again.
```
luukvbaal, line 117: I took another good look at what was actually going on here. The `mt_itr_rawkey(itr).flags &= (uint16_t) ~MT_FLAG_INVALID;` at line 144 was responsible for this, where `itr` is no longer valid after `marktree_move()`. Refactored this to `itr` is no longer used after `marktree_move()`.

PR #32023, opened 2025-01-15
fix: vim.diagnostic.config does not take func as float config

### LINKED ISSUE #32022: Diagnostic float opts does not accept function
### Problem

According to `vim.diagnostic.Opts` documentation, float should accept a function as well. However passing a function would use the default config instead.

```
      • {float}?             (`boolean|vim.diagnostic.Opts.Float|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Float`)
                             Options for floating windows. See
                             |vim.diagnostic.Opts.Float|.
```

From the code, it only considers the table type as float config.

https://github.com/neovim/neovim/blob/a78eddd54112033eea0212865efd2f75cc59fc93/runtime/lua/vim/diagnostic.lua?plain=1#L1021-L1031

### Steps to reproduce

1. Add following config to `init.lua`

```lua
vim.diagnostic.config({
	virtual_text = false,
	float = function()
		return {
			format = function()
				return 'hello world'
			end,
		}
	end,
})
```

2. open a lua file, and create a syntax error in the file
3. jump to the diagnostic

### Expected behavior

floating window should show 'hello world' instead of the actual error message

### Nvim version (nvim -v)

v0.11.0-nightly+a8ace2c

N/A

### Operating system/version

NixOs

### Terminal name/version

Alacritty

alacritty

### Installation

nix

s1n7ax:
Fixes: https://github.com/neovim/neovim/issues/32022

Unfortunately I don't know how to test this in NixOS. Though I can successfully build the project launching `nvim` in build directory will result in following error. Hope anyone else check it or is there a way to get around this error?

```
Error detected while processing /home/s1n7ax/.config/nvim/init.lua:
E5113: Error while calling lua chunk: vim/_init_packages.lua:0: module 'vim.uri' not found:
```

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -1018,9 +1018,18 @@ local function goto_diagnostic(diagnostic, opts)
     vim.cmd('normal! zv')
   end)
 
-  local float_opts = opts.float
+  --- @type vim.diagnostic.Opts.Float
+  local float_opts
+
+  if type(opts.float) == 'table' then
+    float_opts = opts.float --[[@as vim.diagnostic.Opts.Float]]
+  elseif type(opts.float) == 'function' then
+    float_opts = opts.float(diagnostic.namespace, diagnostic.bufnr)
+  elseif opts.float then
+    float_opts = {}
+  end
+
   if float_opts then
-    float_opts = type(float_opts) == 'table' and float_opts or {}
     vim.schedule(function()
       M.open_float(vim.tbl_extend('keep', float_opts, {
         bufnr = api.nvim_win_get_buf(winid),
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/diagnostic.lua
```diff
@@ -1018,9 +1018,18 @@ local function goto_diagnostic(diagnostic, opts)
     vim.cmd('normal! zv')
   end)
 
-  local float_opts = opts.float
+  --- @type vim.diagnostic.Opts.Float
+  local float_opts
```
justinmk, line 1022:
```suggestion
  local float_opts = type(opts.float) == 'table' and opts.float or (... and ... or ...)
```

justinmk:
> Though I can successfully build the project launching `nvim` in build directory will result in following error.

you didn't say how you are launching nvim. anyways, try this:

    VIMRUNTIME=/path/to/runtime ./build/bin/nvim --luamod-dev

in any case, a test is probably needed (or update an existing test)

s1n7ax:
Sorry about the inconvenience. I got neovim working in local machine and turn out this is not working at all. I'll be creating a new PR. I'll close this now

Issue #30756, opened 2024-10-11
make install does not work with cmake 3.13 / 3.14


ImNotaGit:
### Problem

I tried to build and install v0.10.2 (tags/v0.10.2) from source with gcc/10.3.0 and cmake/3.13.4 on CentOS.

There was no error. But in the end of `make install` there was a warning message `WARNING: Target "nvim_bin" has EXCLUDE_FROM_ALL set and will not be built by default but an install rule has been provided for it.  CMake does not define behavior for this case.`, and it appeared that the install did not complete. In the end `$HOME/neovim` is not present.

BTW I used exactly the same process and dependencies to build v0.9.2 without any issue.

### Steps to reproduce

```
git clone https://github.com/neovim/neovim.git 
cd neovim
git checkout tags/v0.10.2
make CMAKE_BUILD_TYPE=RelWithDebInfo CMAKE_EXTRA_FLAGS="-DCMAKE_INSTALL_PREFIX=$HOME/neovim"
make install
```

### Expected behavior

I expect the install to complete with the files installed to the specified destination.

### Nvim version (nvim -v)

tags/v0.10.2

### Vim (not Nvim) behaves the same?

NA

### Operating system/version

CentOS

### Terminal name/version

NA

### $TERM environment variable

xterm-256color

### Installation

build from repo

dundargoc:
Also, could you link to the full build/install logs in a gist or something? I'd like to see what actually happens.

ImNotaGit:
> I can't recreate this, neither on macos or on centos docker container. I wonder if this might be related to the cmake version somehow. Do you have the possibility to try out a newer cmake version? It would really help in pinpointing the problem I think.

cmake/3.13.4 is the newest version available on our system... I'm afraid it will be hard for me to get a newer version... sorry.

> Also, could you link to the full build/install logs in a gist or something? I'd like to see what actually happens.

Sure, please find it here: https://gist.github.com/ImNotaGit/4121920dccd6e6a8a3f094f63e708860
This gist contains output (i.e. the messages printed to stdout) from `make` and `make install`, as well as the `CMakeOutput.log` and `CMakeError.log`. It turned out that `CMakeError.log` does include some errors (which I cannot decipher), but they were not shown during `make` or `make install` (both of which did reach 100% w/o displaying any error).

justinmk:
> I'm afraid it will be hard for me to get a newer version... sorry.

[...]

PR #32024, opened 2025-01-15
build: fix `make install` on cmake 3.13 and 3.14

dundargoc:
Closes https://github.com/neovim/neovim/issues/30756.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ distclean:
 	$(MAKE) clean
 
 install: checkprefix nvim
-	$(CMAKE) --install build
+	$(CMAKE_GENERATOR) -C build install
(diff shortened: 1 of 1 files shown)


steev:
This doesn't seem right - Ninja's generator is Ninja, but the binary is ninja, so when you attempt to install you now get

```
[136/137] Generating zh_CN.UTF-8.mo
"Ninja" -C build install
/bin/sh: 1: Ninja: not found
make: *** [Makefile:160: install] Error 127
```

dundargoc:
Oh right, this works on chad windows and macos because "Ninja" is the same as "ninja" but not on linux. Will create a followup.

PR #32029, opened 2025-01-15
fix(treesitter): replace child_containing_descendant with child_with_…

hellobbn:
…descendant

Problem: As of tree-sitter/tree-sitter@344a88c, child_containing_descendant is removed
Solution: Replace it with `child_with_descendant`


clason:
Yes but we are still on 0.24. We will handle this as part of migrating to 0.25 _when that is released_.

PR #32034, opened 2025-01-15
fix: show previews for nomodifiable buffers

### LINKED ISSUE #32016: command-preview: does not show in nomodifiable buffers
### Problem

Hello,

https://github.com/neovim/neovim/blob/e8a6c1b02122852da83dc52184e78369598d8240/src/nvim/ex_getln.c#L2706-L2709

`&& cmdpreview_may_show(s)) { `

won't run if the buffer is not modifiable, this makes it that so by default I can't add incremental previews for quickfix in custom user commands

### Steps to reproduce

```lua
vim.opt.icm = "split"

vim.api.nvim_create_user_command("PreviewTest", function() end, {
	preview = function()
		vim.api.nvim_buf_set_lines(0, 0, -1, false, { "if you see this it works" })
		return 2
	end,
})
```

`:set modifiable` works
`:set nomodifiable` does not work

### Expected behavior

I would expect incremental preview to work with nomodifiable buffers.

But I understand that it kinda makes sense to skip nomodifiable, because incremental preview should in theory change the buffer. But the preview function would just fail at that point and the plugin author is responsible that it wouldn't do that

Anyway, while I don't like it, this workaround works

```lua
	vim.api.nvim_create_autocmd("CmdlineEnter", {
		callback = function()
			vim.bo.modifiable = true 
		end,
	})
	vim.api.nvim_create_autocmd("CmdlineLeave", {
		callback = function()
			vim.bo.modifiable = false 
		end,
	})
```

### Nvim version (nvim -v)

0.10.3

no, does not have this functionality

### Operating system/version

Linux

### Terminal name/version

tmux

screen256

### Installation

pacman / tested with it built with source also

### Extra notes

I think that this should be in the help page ("Please note that incremental preview is skipped on nomodifiable buffers")

tronikelis:
closes #32016

The test fails if I add the check back in

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -319,6 +319,8 @@ PLUGINS
 
 • EditorConfig
   • spelling_language property is now supported.
+• 'inccommand' incremental preview can run on 'nomodifiable' buffers and
+  restores their 'modifiable' state
 
 STARTUP
 
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -168,6 +168,7 @@ typedef struct {
   pos_T save_b_op_end;
   varnumber_T save_changedtick;
   CpUndoInfo undo_info;
+  int save_b_p_ma;
 } CpBufInfo;
 
 typedef struct {
@@ -2419,6 +2420,7 @@ static void cmdpreview_prepare(CpInfo *cpinfo)
     if (!set_has(ptr_t, &saved_bufs, buf)) {
       CpBufInfo cp_bufinfo;
       cp_bufinfo.buf = buf;
+      cp_bufinfo.save_b_p_ma = buf->b_p_ma;
       cp_bufinfo.save_b_p_ul = buf->b_p_ul;
       cp_bufinfo.save_b_changed = buf->b_changed;
       cp_bufinfo.save_b_op_start = buf->b_op_start;
@@ -2509,6 +2511,7 @@ static void cmdpreview_restore_state(CpInfo *cpinfo)
     }
 
     buf->b_p_ul = cp_bufinfo.save_b_p_ul;        // Restore 'undolevels'
+    buf->b_p_ma = cp_bufinfo.save_b_p_ma;        // Restore 'modifiable'
   }
 
   for (size_t i = 0; i < cpinfo->win_info.size; i++) {
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -2704,7 +2704,6 @@ static int command_line_changed(CommandLineState *s)
       && current_sctx.sc_sid == 0    // only if interactive
       && *p_icm != NUL       // 'inccommand' is set
       && !exmode_active      // not in ex mode
-      && curbuf->b_p_ma      // buffer is modifiable
       && cmdline_star == 0   // not typing a password
       && !vpeekc_any()
       && cmdpreview_may_show(s)) {
--- a/src/nvim/ex_getln.c
+++ b/src/nvim/ex_getln.c
@@ -163,12 +163,12 @@ typedef struct {
 typedef struct {
   buf_T *buf;
   OptInt save_b_p_ul;
+  int save_b_p_ma;
   int save_b_changed;
   pos_T save_b_op_start;
   pos_T save_b_op_end;
   varnumber_T save_changedtick;
   CpUndoInfo undo_info;
-  int save_b_p_ma;
 } CpBufInfo;
 
 typedef struct {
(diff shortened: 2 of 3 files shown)


test/functional/ui/inccommand_user_spec.lua
```diff
@@ -253,6 +262,29 @@ describe("'inccommand' for user commands", function()
     ]]
   end)
 
+  it('preview shown with nomodifiable', function()
```
justinmk, line 265:
```suggestion
  it("can preview 'nomodifiable' buffer", function()
```

test/functional/ui/inccommand_user_spec.lua
```diff
@@ -253,6 +262,29 @@ describe("'inccommand' for user commands", function()
     ]]
   end)
 
+  it("can preview 'nomodifiable' buffer", function()
+    command('set nomodifiable')
```
justinmk, line 266: You mentioned that you want to edit a quickfix buffer, so maybe it's worth setting up quickfix in this test. That gives extra coverage.

Also this test should make a small edit to the quickfix, then cancel the preview-command, then assert that the quickfix buffer was restored.

Also at the start and end of the test, assert that 'nomodifiable' is set on the buffer.

    eq(false, api.nvim_get_option_value('modifiable', {buf=<quickfix_buf_num>}))

test/functional/ui/inccommand_user_spec.lua
tronikelis: thanks

justinmk (approved):

test/functional/ui/inccommand_user_spec.lua
justinmk: What happens if the preview callback sets 'modifiable', is it able to edit the buffer?

tronikelis: > What happens if the preview callback sets 'modifiable', is it able to edit the buffer?

Don't know, will check

tronikelis: https://github.com/user-attachments/assets/f7542773-e37b-4dad-99fd-39b37e269aa4


nvim does not reset the modifiable, only the buffer lines it seems like

tronikelis: I should add restoring  `modifiable` then right?

tronikelis: Added in [`7b2e553` (#32034)](https://github.com/neovim/neovim/pull/32034/commits/7b2e553e47739081edc07e555246da5dc9ab54e5)

We can delete this commit if we decide that not restoring is the correct way to go

src/nvim/ex_getln.c
```diff
@@ -168,6 +168,7 @@ typedef struct {
   pos_T save_b_op_end;
   varnumber_T save_changedtick;
   CpUndoInfo undo_info;
+  int save_b_p_ma;
```
justinmk, line 171: looking at `save_b_p_ul` above, should this be a `Opt...` type instead of `int`? 

rebase on latest master to make sure you have recent changes related to the options code.

test/functional/ui/inccommand_user_spec.lua
```diff
@@ -253,6 +253,61 @@ describe("'inccommand' for user commands", function()
     ]]
   end)
 
+  it("restores 'modifiable'", function()
```
justinmk, line 256: Don't need separate tests. The "can preview nomodifiable" test can also edit the buffer and assert the stuff this test is doing.

justinmk (approved):
This looks good after comments are resolved, thanks for following up. Would you mind adding an item to `:help news-features` "PLUGINS" section

src/nvim/ex_getln.c
tronikelis: https://github.com/neovim/neovim/blob/8a236c242a76825a6a9266feda45794c7328c807/src/nvim/buffer_defs.h#L576-L576

It is `int` in the original spec, while ul is OptInt, so not sure, I guess we either change both to OptInt or leave as int?

zeertzjq: No

runtime/doc/news.txt
```diff
@@ -148,7 +148,8 @@ OPTIONS
 
 PLUGINS
 
-• TODO
+• 'inccomand' incremental preview now runs on 'nomodifiable' buffers and
+  restores their 'modifiable' state
```
justinmk, line 152: this is in news-breaking. please find the PLUGINS section in news-features, below :)

runtime/doc/news.txt
```diff
@@ -148,7 +148,8 @@ OPTIONS
 
 PLUGINS
 
-• TODO
+• 'inccomand' incremental preview now runs on 'nomodifiable' buffers and
```
justinmk, line 151:
```suggestion
• 'inccommand' incremental preview can run on 'nomodifiable' buffers and
```

PR #32038, opened 2025-01-15
feat(terminal): add support for kitty keyboard protocol

### LINKED ISSUE #18735: ctrl-/ doesn't work for fzf in terminal
### Neovim version (nvim -v)

NVIM v0.8.0-dev+1751-ga20892c4b

No

### Operating system/version

macOS 12.4

### Terminal name/version

xterm

xterm-256color

### Installation

homebrew (HEAD)

- Start from `nvim --clean`
- Open a terminal buffer: `:term`
- Run: `fzf --preview cat --bind='ctrl-/:toggle-preview'`
- Press `CTRL-/`

### Expected behavior

`CTRL-/` should work well in the terminal (FZF should toggle the preview window)

On neovim 0.7.1, this work correctly as expected.

### Actual behavior

`^[[47;5u` is directly inserted. FZF prompt will have a bunch of `;5u` letters inserted.

image

### LINKED ISSUE #24093: Terminal mode should probably not send CSI-u sequences for shift+enter and shift+space

### LINKED ISSUE #32019: not all terminal keycodes are being passed to built-in `:terminal` correctly
### Problem

Since #31932, I've noticed some terminal keycodes are not being properly passed to the underlying terminal. Specifically, `<c-j>` is appearing as `106;5u` instead of receiving the full escape code for the keypress.

I haven't had time to dig too far into this, but it's easily reproducible (see repro below). 

This does cause issues when using terminal applications inside a `:term` that rely on that keycode (`<c-j>`).

### Steps to reproduce

1. nvim --clean
2. `:term`
3. `<c-v><c-j>` (control-v and then control-j)

### Expected behavior

This should cause the cursor to move down a line, but instead it inserts `^[[106;5u`. If you do the same thing (`<c-v><c-j>`) in any other terminal emulator, the cursor should move down a line. This is also what happens in neovim's terminal emulator in the commit just prior to the bisected commit.

### Nvim version (nvim -v)

v0.11.0-dev-1511+gdcaf8bef08

no

### Operating system/version

macOS 15.2/RHEL 8

### Terminal name/version

ghostty, kitty, tmux

xterm-ghostty, xterm-kitty, tmux-256color

### Installation

built from git, bisected

gpanders:
This commit adds basic support for the kitty keyboard protocol to Neovim's builtin terminal. For now only the first mode ("Disambiguate escape codes") is supported.

Resolves: https://github.com/neovim/neovim/issues/18735
Resolves: https://github.com/neovim/neovim/issues/24093
Resolves: https://github.com/neovim/neovim/issues/32019

CSIu encoding is now disabled by default unless an application in the terminal opts into it.

### DIFF
--- a/src/nvim/vterm/keyboard.c
+++ b/src/nvim/vterm/keyboard.c
@@ -10,54 +10,77 @@
 # include "vterm/keyboard.c.generated.h"
 #endif
 
+static VTermKeyEncodingFlags vterm_state_get_key_encoding_flags(const VTermState *state)
+{
+  int screen = state->mode.alt_screen ? BUFIDX_ALTSCREEN : BUFIDX_PRIMARY;
+  const struct VTermKeyEncodingStack *stack = &state->key_encoding_stacks[screen];
+  assert(stack->size > 0);
+  return stack->items[stack->size - 1];
+}
+
 void vterm_keyboard_unichar(VTerm *vt, uint32_t c, VTermModifier mod)
 {
-  // The shift modifier is never important for Unicode characters apart from Space
-  if (c != ' ') {
-    mod &= (unsigned)~VTERM_MOD_SHIFT;
+  bool passthru = false;
+  if (c == ' ') {
+    // Space is passed through only when there are no modifiers (including shift)
+    passthru = mod == VTERM_MOD_NONE;
+  } else {
+    // Otherwise pass through when there are no modifiers (ignoring shift)
+    passthru = (mod & (unsigned)~VTERM_MOD_SHIFT) == 0;
   }
 
-  if (mod == 0) {
-    // Normal text - ignore just shift
+  if (passthru) {
     char str[6];
     int seqlen = fill_utf8((int)c, str);
     vterm_push_output_bytes(vt, str, (size_t)seqlen);
     return;
   }
 
-  int needs_CSIu;
-  switch (c) {
-  // Special Ctrl- letters that can't be represented elsewise
-  case 'i':
-  case 'j':
-  case 'm':
-  case '[':
-    needs_CSIu = 1;
-    break;
-  // Ctrl-\ ] ^ _ don't need CSUu
-  case '\\':
-  case ']':
-  case '^':
-  case '_':
-    needs_CSIu = 0;
-    break;
-  // Shift-space needs CSIu
-  case ' ':
-    needs_CSIu = !!(mod & VTERM_MOD_SHIFT);
-    break;
-  // All other characters needs CSIu except for letters a-z
-  default:
-    needs_CSIu = (c < 'a' || c > 'z');
-  }
+  VTermKeyEncodingFlags flags = vterm_state_get_key_encoding_flags(vt->state);
+  if (flags.disambiguate) {
+    // Always use unshifted codepoint
+    if (c >= 'A' && c <= 'Z') {
+      c += 'a' - 'A';
+      mod |= VTERM_MOD_SHIFT;
+    }
 
-  // ALT we can just prefix with ESC; anything else requires CSI u
-  if (needs_CSIu && (mod & (unsigned)~VTERM_MOD_ALT)) {
     vterm_push_output_sprintf_ctrl(vt, C1_CSI, "%d;%du", c, mod + 1);
     return;
   }
 
   if (mod & VTERM_MOD_CTRL) {
-    c &= 0x1f;
+    // Handle special cases. These are taken from kitty, but seem mostly
+    // consistent across terminals.
+    switch (c) {
+    case '2':
+    case ' ':
+      // Ctrl+2 is NUL to match Ctrl+@ (which is Shift+2 on US keyboards)
+      // Ctrl+Space is also NUL for some reason
+      c = 0x00;
+      break;
+    case '3':
+    case '4':
+    case '5':
+    case '6':
+    case '7':
+      // Ctrl+3 through Ctrl+7 are sequential starting from 0x1b. Importantly,
+      // this means that Ctrl+6 emits 0x1e (the same as Ctrl+^ on US keyboards)
+      c = 0x1b + c - '3';
+      break;
+    case '8':
+      // Ctrl+8 is DEL
+      c = 0x7f;
+      break;
+    case '/':
+      // Ctrl+/ is equivalent to Ctrl+_ for historic reasons
+      c = 0x1f;
+      break;
+    default:
+      if (c >= '@' && c <= 0x7f) {
+        c &= 0x1f;
+      }
+      break;
+    }
   }
 
   vterm_push_output_sprintf(vt, "%s%c", mod & VTERM_MOD_ALT ? ESC_S : "", c);
@@ -75,7 +98,7 @@ typedef struct {
     KEYCODE_CSINUM,
     KEYCODE_KEYPAD,
   } type;
-  char literal;
+  int literal;
   int csinum;
 } keycodes_s;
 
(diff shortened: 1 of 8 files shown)


test/functional/terminal/buffer_spec.lua
```diff
@@ -629,6 +629,9 @@ describe('terminal input', function()
   -- TODO(bfredl): getcharstr() erases the distinction between <C-I> and <Tab>.
   -- If it was enhanced or replaced this could get folded into the test above.
   it('can send TAB/C-I and ESC/C-[ separately', function()
+    if skip(is_os('win'), "The escape sequence to enable kitty keyboard mode doesn't work on Windows") then
+      return
+    end
```
gpanders, line 634: This test fails on Windows.

With help from @dundargoc, we determined that vterm never sees the query from the nested Nvim instance to detect if the kitty keyboard protocol is supported.

It gets stranger: requesting a primary device attributes report (DA1) in the nested terminal with `` Write-Host "`e[c" `` _should_ return `CSI [ 1 ; 2 c` (see [here]()), but @dundargoc instead received `CSI ?61;6;7;21;22;23;24;28;32;42c`. There is something fishy going on between the shell and vterm on Windows. Probably something with Windows' PTY emulation (which I don't claim to understand at all).

I feel comfortable skipping this since the entire `terminal/tui_spec.lua` file is already skipped on Windows. @dundargoc also confirmed that the terminal on Windows is still fully functional, so this doesn't regress anything other than an inability to distinguish `<C-I>` and `<Tab>` in nested Neovim instances, which was already the case anyway until https://github.com/neovim/neovim/pull/31932 was recently merged.

dundargoc: This was tested with `pwsh`, meaning powershell version 7 and not the default powershell as the default version can't interpret escape code.

gpanders: > It gets stranger: requesting a primary device attributes report (DA1) in the nested terminal with `` Write-Host "`e[c" `` _should_ return `CSI ? 1;2c` (see [here](https://github.com/neovim/neovim/blob/5801cbd1af65ff9de1602bbd4cb66e76b6bf4328/src/nvim/vterm/state.c#L1388)), but @dundargoc instead received `CSI ?61;6;7;21;22;23;24;28;32;42c`. There is something fishy going on between the shell and vterm on Windows. Probably something with Windows' PTY emulation (which I don't claim to understand at all).

This is where this reply comes from https://github.com/microsoft/terminal/blob/25392ea6042242eceeb95cca29ca18784cf0ee7b/src/terminal/adapter/adaptDispatch.cpp#L1494

It does indeed look like the ConPTY layer in Windows is intercepting some of these sequences and sending its own responses rather than delegating to libvterm. So either libvterm never receives the query, or it receives it but Nvim never gets the response (I'm not sure which right now).

zeertzjq:
To match the requirements of kitty keyboard protocol (any key with a Ctrl modifier is sent as CSI u), you need some other changes:
```diff
diff --git a/src/nvim/getchar.c b/src/nvim/getchar.c
index 6cf4556a9f..516de37444 100644
--- a/src/nvim/getchar.c
+++ b/src/nvim/getchar.c
@@ -1518,11 +1518,9 @@ int merge_modifiers(int c_arg, int *modifiers)
 
   if (*modifiers & MOD_MASK_CTRL) {
     if ((c >= '`' && c <= 0x7f) || (c >= '@' && c <= '_')) {
-      if (!(State & MODE_TERMINAL) || !(c == 'I' || c == 'J' || c == 'M' || c == '[')) {
-        c &= 0x1f;
-        if (c == NUL) {
-          c = K_ZERO;
-        }
+      c &= 0x1f;
+      if (c == NUL) {
+        c = K_ZERO;
       }
     } else if (c == '6') {
       // CTRL-6 is equivalent to CTRL-^
@@ -2058,6 +2056,10 @@ static bool at_ins_compl_key(void)
 /// @return  the length of the replaced bytes, 0 if nothing changed, -1 for error.
 static int check_simplify_modifier(int max_offset)
 {
+  if (State & MODE_TERMINAL) {
+    return 0;
+  }
+
   for (int offset = 0; offset < max_offset; offset++) {
     if (offset + 3 >= typebuf.tb_len) {
       break;
diff --git a/src/nvim/terminal.c b/src/nvim/terminal.c
index fb8565e534..09099ac5b2 100644
--- a/src/nvim/terminal.c
+++ b/src/nvim/terminal.c
@@ -783,7 +783,10 @@ static int terminal_execute(VimState *state, int key)
 {
   TerminalState *s = (TerminalState *)state;
 
-  switch (key) {
+  int tmp_mod_mask = mod_mask;
+  int mod_key = merge_modifiers(key, &tmp_mod_mask);
+
+  switch (mod_key) {
   case K_LEFTMOUSE:
   case K_LEFTDRAG:
   case K_LEFTRELEASE:
@@ -841,13 +844,13 @@ static int terminal_execute(VimState *state, int key)
     FALLTHROUGH;
 
   default:
-    if (key == Ctrl_C) {
+    if (mod_key == Ctrl_C) {
       // terminal_enter() always sets `mapped_ctrl_c` to avoid `got_int`. 8eeda7169aa4
       // But `got_int` may be set elsewhere, e.g. by interrupt() or an autocommand,
       // so ensure that it is cleared.
       got_int = false;
     }
-    if (key == Ctrl_BSL && !s->got_bsl) {
+    if (mod_key == Ctrl_BSL && !s->got_bsl) {
       s->got_bsl = true;
       break;
     }
```

zeertzjq:
Note: I meant Ctrl-digits behave strangely when kitty keyboard protocol is disabled.

gpanders:
> Note: I meant Ctrl-digits behave strangely when kitty keyboard protocol is disabled.

Yes I noticed the same thing. As far as I can tell there is no real rhyme or reason to how Ctrl+digits are encoded, it seems to be special case handling to match expected historic behavior. I've updated vterm to copy what kitty does, so this should be working now. If you can find a reference to a better way that these keys should be encoded please let me know...

test/unit/vterm_spec.lua
```diff
@@ -2478,6 +2555,77 @@ putglyph 1f3f4,200d,2620,fe0f 2 0,4]])
     expect_output('\x1b[I')
     vterm.vterm_state_focus_out(state)
     expect_output('\x1b[O')
+
+    -- Disambiguate escape codes enabled
```
zeertzjq, line 2559:
```suggestion
    -- Disambiguate escape codes disabled
```

src/nvim/vterm/keyboard.c
```diff
@@ -10,54 +10,77 @@
...
+      // this means that Ctrl+6 emits 0x1e (the same as Ctrl+^ on US keyboards)
+      c = 0x1b + c - '3';
+      break;
+    case '8':
+      // Ctrl+8 is DEL
+      c = 0x7f;
+      break;
+    case '/':
+      // Ctrl+/ is equivalent to Ctrl+_ for historic reasons
+      c = 0x5f;
```
zeertzjq, line 76: 0x5f is \_, Ctrl+\_ is 0x1f.
```suggestion
      c = 0x1f;
```

gpanders: Nice catch, thanks.

zeertzjq (approved):

PR #32044, opened 2025-01-15
build: fix install

dundargoc:


dundargoc:
@steev please test

steev:
Does not work, on Linux systems, the generator itself is still called Ninja, just that the binary is ninja

PR #32059, opened 2025-01-16
ci: add arm64 runner

clason:
Problem: No tests on Linux Arm.

Solution: Add tests on Linux Arm with `RelWithDebInfo`.   Remove duplicate and low-leverage `build-types` job.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -204,37 +204,6 @@ jobs:
   windows:
     uses: ./.github/workflows/test_windows.yml
 
-  # This job tests the following things:
-  # - Check if MinSizeRel and RelWithDebInfo compiles correctly.
-  # - Test the above build types with the GCC compiler specifically.
-  #   Empirically the difference in warning levels between GCC and other
-  #   compilers is particularly big.
-  # - Test if the build works with multi-config generators. We mostly use
-  #   single-config generators so it's nice to have a small sanity check for
-  #   multi-config.
-  build-types:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 10
-    env:
-      CC: gcc
-    steps:
-      - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup
-
-      - name: Build third-party deps
-        run: |
-          cmake -S cmake.deps -B .deps -G "Ninja Multi-Config"
-          cmake --build .deps
-
-      - name: Configure
-        run: cmake --preset ci -G "Ninja Multi-Config"
-
-      - name: RelWithDebInfo
-        run: cmake --build build --config RelWithDebInfo
-
-      - name: MinSizeRel
-        run: cmake --build build --config MinSizeRel
-
   with-external-deps:
     runs-on: ubuntu-24.04
     timeout-minutes: 10
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,6 +107,7 @@ jobs:
             { runner: ubuntu-24.04, os: ubuntu, flavor: asan, cc: clang, flags: -D ENABLE_ASAN_UBSAN=ON },
             { runner: ubuntu-24.04, os: ubuntu, flavor: tsan, cc: clang, flags: -D ENABLE_TSAN=ON },
             { runner: ubuntu-24.04, os: ubuntu, flavor: release, cc: gcc, flags: -D CMAKE_BUILD_TYPE=Release },
+            { runner: ubuntu-24.04-arm, os: ubuntu, flavor: arm, cc: gcc, flags: -D CMAKE_BUILD_TYPE=RelWithDebInfo },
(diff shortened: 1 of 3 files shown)


clason:
Looks good, only issue is permissions with the `vim.ui.open()` tests around nonexistent files (permission issues, I think we had similar on Cirrus). At least the first (validation) test sounds like an actually relevant issue and should likely be fixed before providing releases?

Follow-up question would be whether and how we want to test this platform (add to matrix? switch one of the other (`release`) tests to ARM?)

runtime/doc/support.txt
```diff
@@ -13,6 +13,7 @@ Supported platforms					 *supported-platforms*
 
 `System`          `Tier`      `Versions`                  `Tested versions`
 Linux            1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
```
dundargoc, line 15: I guess add x86-64 or arm64 or whatever the kids use these day?

.github/workflows/test.yml
```diff
@@ -107,6 +107,7 @@ jobs:
             { runner: ubuntu-24.04, os: ubuntu, flavor: asan, cc: clang, flags: -D ENABLE_ASAN_UBSAN=ON },
             { runner: ubuntu-24.04, os: ubuntu, flavor: tsan, cc: clang, flags: -D ENABLE_TSAN=ON },
             { runner: ubuntu-24.04, os: ubuntu, flavor: release, cc: gcc, flags: -D CMAKE_BUILD_TYPE=Release },
+            { runner: ubuntu-24.04-arm, os: ubuntu, flavor: arm, cc: gcc, flags: -D CMAKE_BUILD_TYPE=RelWithDebInfo },
```
dundargoc, line 110: You can remove the `RelWithDebInfo` build in the `build-types` job. In fact, with this I'd say we can remove the `build-types` job entirely as MinSizeRel or Ninja Multi-Config isn't really important enough to warrant a CI job.

dundargoc (approved):

.github/workflows/test.yml
dundargoc: Either do it in this PR or I can do it in a follow up.

clason: @bfredl You'd need to remove the `build-types` job from the list of required checks then.

clason:
Let's get the tests in now; release workflow will be a follow-up in the next days.

zeertzjq:
Why remove build-types jobs?

clason:
Problem: Some CI jobs are redundant: `ResWithDebInfo` is already tested
on Linux-Arm64; `MinSizeRel` and Ninja Multi Config are not sufficiently
relevant in practice to spend CI cycles on.

Solution: Remove `build-types` job.

runtime/doc/support.txt
```diff
@@ -11,15 +11,16 @@ Support                                                              *support*
...
-Linux            1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
-macOS (Intel)    1      >= 11                        macOS 13
-macOS (M1)       1      >= 11                        macOS 15
-Windows 64-bit   1      >= Windows 10 Version 1809   Windows Server 2022
-FreeBSD          1      >= 10                        FreeBSD 14
-OpenBSD          2      >= 7
-MinGW            2      MinGW-w64
-Windows 64-bit   3      < Windows 10 Version 1809
+`System`            `Tier`      `Versions`                  `Tested versions`
+Linux (x864_64)    1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
```
zeertzjq, line 15:
```suggestion
Linux (x86_64)     1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
```

runtime/doc/support.txt
```diff
@@ -11,15 +11,16 @@ Support                                                              *support*
...
+Linux (aarch64)    1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
```
zeertzjq, line 16:
```suggestion
Linux (x86_64)     1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
Linux (arm64)      1      >= 2.6.32, glibc >= 2.12     Ubuntu 24.04
```

clason: That arch was intentional. Can you tell my _why_ we should use one over the other?

zeertzjq: No

clason: wat

zeertzjq: Well I'm not sure either. But GitHub calls it arm64.

clason: So does Debian, so let's go with that. (It's also shorter.)

adamency:
# PSA

This PR adds ARM64 builds to all future releases (i.e. after this PR was merged).

zeertzjq:
No, it's #32194

PR #31575, opened 2024-12-14
feat(ui): allow 'sign/statuscolumn' to combine highlight attrs

luukvbaal:
For consistency; this was already done for rulerformat, statusline, tabline and winbar in #29976.

PR #32060, opened 2025-01-16
feat(api): combined highlights in nvim_eval_statusline()

luukvbaal:
Problem:  Combined highlighting was not applied to nvim_eval_statusline(),
          and 'statuscolumn' sign segment/numhl highlights.
Solution: Add an additional `groups` element to the return value of
          `nvim_eval_statusline()->highlights`. This is an array of stacked
          highlight groups (highest priority last). Also resolve combined
          highlights for the 'statuscolumn' sign segment/numhl highlights.
          Expose/synchronize some drawline.c logic that is now mimicked in
          three different places.

Followup to #29976, #31575


echasnovski:
Thanks, @luukvbaal! This indeed fixes combining highlight attributes for line numbers.

There also seems to be a similar issue for signs, though. Here is the repro:

```lua
vim.o.cursorline = true
vim.o.statuscolumn = '%C%s%=%l'

vim.api.nvim_set_hl(0, 'CursorLineSign', { underline = true })
vim.api.nvim_set_hl(0, 'Green', { fg = 'NvimLightGreen' })

vim.api.nvim_buf_set_lines(0, 0, -1, false, { 'aaa', 'bbb', 'ccc' })
local ns_id = vim.api.nvim_create_namespace('test')
vim.api.nvim_buf_set_extmark(0, ns_id, 0, 0, { sign_text = '!', sign_hl_group = 'Green' })
vim.api.nvim_buf_set_extmark(0, ns_id, 1, 0, { sign_text = '*', sign_hl_group = 'Green' })
```

With this PR it behaves like this (seems to combine with `CursorLineNr` instead of `CursorLineSign`):

https://github.com/user-attachments/assets/c04b937b-3415-4a0c-bcfa-dfc494bfe446

Maybe similar also happens with foldcolumn.

luukvbaal:
Yeah I thought as much. That will be more cumbersome to solve.

luukvbaal:
Also worth noting that since we now combine highlights, the highlights returned by `nvim_eval_statusline()` are no longer representative of what will actually be drawn. Not sure how that API is most often used, but I know at least `nvim-treesitter-context` uses it to build a statuscolumn in a floating window, whose highlights could now be incorrect. Would have to create some kind of dummy highlight groups if we want to support combined highlights there.

justinmk (approved):
Nice choices, as usual!

runtime/doc/api.txt
```diff
@@ -704,6 +704,8 @@ nvim_eval_statusline({str}, {opts})                   *nvim_eval_statusline()*
           • start: (number) Byte index (0-based) of first character that uses
             the highlight.
           • group: (string) Name of highlight group.
+          • groups: (array) Names of stacked highlight groups (highest
+            priority last).
```
luukvbaal, line 708: With this, would/can we deprecate `group` (now/ever)?
I also considered omitting `groups` from the return value when there is only 1 group, so no stacking, but always including it is probably worth it for simplicity sake.

justinmk: > With this, would/can we deprecate `group` (now/ever)?

Sure. We don't have a "pattern" for deprecating kwargs fields, but I suggest something like: if the deprecated key is used, call `vim.deprecate(..., '0.13')` using `NLUA_EXEC_STATIC`.

> I also considered omitting `groups` from the return value when there is only 1 group, so no stacking, but always including it is probably worth it for simplicity sake.

Predictability is strongly preferred. Polymorphic returns are painful.

luukvbaal: > but I suggest something like: if the deprecated key is used

I don't think detecting that is feasible though? At best we could maybe slap a `---@deprecated` on the return class, but that is currently a generic `table<string,any>` for all `Dict` return types, and I'm not even sure the deprecated annotation can be used for class fields.

justinmk: Oh sorry I thought "group" was a kwargs field, not a return value.

For deprecating return values we can just mention it in its docstring, for now.

luukvbaal: > For deprecating return values we can just mention it in its docstring, for now.

OK, so should we do that here?

justinmk: Sure

PR #32065, opened 2025-01-17
docs(gh): use new issue types

clason:
Advantage: Easier visibility, better filtering, can be extended to new custom issue types (have to be defined for the org).

Tested on https://github.com/nvim-treesitter/nvim-treesitter/issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
 name: Bug Report
 description: Report a problem in Nvim
-labels: [bug]
+type: 'bug'
 body:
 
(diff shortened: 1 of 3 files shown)


lewis6991 (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,6 +1,6 @@
 name: Bug Report
 description: Report a problem in Nvim
-labels: [bug]
+type: 'bug'
```
justinmk, line 3: why not also keep the label?

clason: Single source of truth.

PR #32070, opened 2025-01-17
fix(lsp): document_symbol should use loclist as default

### LINKED ISSUE #31832: `vim.lsp.buf.document_symbol` should use loclist
### Problem

`vim.lsp.buf.document_symbol` uses the global quickfix list. This means a user can't view the document symbols of two different files at the same time. 

Assuming you have a Lua lsp working correctly:
1. `echo "local function aa() end" > a.lua`
2. `echo "local function bb() end" > b.lua`
3. `nvim -o a.lua b.lua`
4. `gO`
5. Switch to b.lua
6. `gO`

Notice the document symbol list changes.

Current behavior: one possible qflist

### Expected behavior

The `loclist` argument of `vim.lsp.buf.document_symbol` is set to `true` by default, but can still be overridden.

Each window should be able to have its separate list of document symbols.

In the example above, this would mean two loclists: one list for a.lua with `aa`, and one for b.lua with `bb`.

Expected behavior: loclist for every window

---

I can create a PR for this if you want.

yochem:
Problem: Not able to open document symbols for different buffers

Solution: Use the location list as default.

To switch back to previous behavior (qflist):

  vim.lsp.buf.document_symbol({ loclist = false })

Fixes: #31832

### DIFF
--- a/runtime/doc/lsp.txt
+++ b/runtime/doc/lsp.txt
@@ -1445,12 +1445,11 @@ Lua module: vim.lsp.buf                                              *lsp-buf*
                         vim.lsp.buf.definition({ on_list = on_list })
                         vim.lsp.buf.references(nil, { on_list = on_list })
 <
-
-                    If you prefer loclist instead of qflist: >lua
+      • {loclist}?  (`boolean`) Whether to use the |location-list| or the
+                    |quickfix| list. >lua
                         vim.lsp.buf.definition({ loclist = true })
-                        vim.lsp.buf.references(nil, { loclist = true })
+                        vim.lsp.buf.references(nil, { loclist = false })
 <
-      • {loclist}?  (`boolean`)
 
 *vim.lsp.LocationOpts*
     Extends: |vim.lsp.ListOpts|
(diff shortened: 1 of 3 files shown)


MariaSolOs:
Not against the change, but the docs are a bit confusing because the the description of `vim.lsp.ListOpts` makes it sound as if `loclist` is always false by default:
https://github.com/neovim/neovim/blob/ba8ce7498cd530b211d4a02df5e71b9eef022f6d/runtime/doc/lsp.txt#L1449-L1451

yochem:
Thanks for the suggestion, I've changed it.

MariaSolOs:
> Thanks for the suggestion, I've changed it.

Hmm, I'm not seeing any changes.

yochem:
I'm an idiot, forgot to actually push it 🙈 sorry!

MariaSolOs (approved):

yochem:
Shoul be updated now.

yochem:
Btw, should this be marked as a breaking change?

justinmk (approved):

PR #32076, opened 2025-01-17
docs(float): Add tag `floating-windows`

brianhuster:
Problem:
- It was hard for me to find documents about "floating windows" in Nvim
- ~~Due to the difference in terms between Nvim (floating windows) and Vim (popup windows), people transitioning from Vim may also find it hard to find the relevant doc in Nvim~~
- ~~The tag `extmark` seems redundant for me (we already have `extmarks`)~~

Solution:
- Add tags `api-floating-windows` and `popupwin`
- ~~Remove tag `extmark`~~

### DIFF
--- a/runtime/doc/api.txt
+++ b/runtime/doc/api.txt
@@ -453,7 +453,7 @@ Example using the API from Vimscript: >vim
 
 
 ==============================================================================
-Floating windows                                                 *api-floatwin*
+Floating windows				*api-floatwin* *floating-windows*
 
 Floating windows ("floats") are displayed on top of normal windows.  This is
 useful to implement simple widgets, such as tooltips displayed next to the


runtime/doc/api.txt
```diff
@@ -502,7 +502,7 @@ Example: create a float with scratch buffer: >vim
 <
 
 ==============================================================================
-Extended marks                       *api-extended-marks* *extmarks* *extmark*
+Extended marks					*api-extended-marks* *extmarks*
```
clason, line 505: You can't do that since the tag is linked in (multiple) other places.

dundargoc:
Thank you for your contribution! We collect all small documentation fixes into a single pull request and merge it once it gets big enough: https://github.com/neovim/neovim/pull/31995.

PR #32077, opened 2025-01-17
feat(treesitter): allow is_valid() to accept a range

### LINKED ISSUE #32075: Too many redraws when reparsing becomes async
### Problem

The highlighter runs an async parse in `on_win`, which redraws in the callback to update once parsing finishes. This redraw is *not* run when the parsing was able to complete synchronously, because we can just call `on_line` after so it is not necessary. Reparses will *usually* run synchronously because they are much quicker, so we don't need an async parse all the time.

*However*, some files are so large that even reparsing will get broken up over multiple event loop iterations. In this case, a reparse triggers a redraw, which triggers `on_win`, which triggers a reparse, which triggers a redraw, etc. The result is that large files (like the big linux one) will have CPU usage at 100% when idling in a buffer, and high amounts of flickering when scrolling.

### Steps to reproduce

Download https://raw.githubusercontent.com/torvalds/linux/master/drivers/gpu/drm/amd/include/asic_reg/dcn/dcn_3_2_0_sh_mask.h

```lua
-- minimal.lua
for name, url in pairs {
} do
  end
end

-- Increase if needed
vim.o.rdt = 6000

require('nvim-treesitter.configs').setup {
  ensure_installed = { 'lua', 'cpp', 'c', 'comment' },
}
```

run

`nvim --clean -u minimal.lua dcn_3_2_0_sh_mask.h`

Note that CPU usage is consistently at 100% due to the above reasons

### Expected behavior

Reparsing should not be run in the first place; rather the `LanguageTree` should recognize that the region is valid and should return the trees immediately.

A proper solution (in my opinion) is to let `is_valid()` accept a range, which will allow it to tell if a languagetree is valid over that range. The downside is it will no longer be able to flatten `_valid` to `true` (since we won't know if it is valid for all possible ranges). IMO this is still overall an optimization

### Nvim version (nvim -v)

v0.11.0-nightly+e8a6c1b

NA

### Operating system/version

NixOS 25.05

### Terminal name/version

Ghostty 1.0.1

xterm-ghostty

### Installation

Nixpkgs nightly overlay

ribru17:
This commit allows `LanguageTree:is_valid()` to accept a range parameter which will narrow the scope of the validity check. This allows the tree to return earlier much more often in the highlighter, saving lots of CPU cycles (especially when re-parsing would've taken multiple event loop iterations).

Fixes #32075

### DIFF
--- a/runtime/lua/vim/treesitter/languagetree.lua
+++ b/runtime/lua/vim/treesitter/languagetree.lua
@@ -91,7 +91,7 @@ local TSCallbackNames = {
 ---@field private _source (integer|string) Buffer or string to parse
 ---@field private _trees table<integer, TSTree> Reference to parsed tree (one for each language).
 ---Each key is the index of region, which is synced with _regions and _valid.
----@field private _valid boolean|table<integer,boolean> If the parsed tree is valid
+---@field private _valid table<integer,boolean>? Map from tree IDs to whether they are valid.
 ---@field private _logger? fun(logtype: string, msg: string)
 ---@field private _logfile? file*
 local LanguageTree = {}
@@ -134,7 +134,7 @@ function LanguageTree.new(source, lang, opts)
       or query.get(lang, 'injections'),
     _has_regions = false,
     _injections_processed = false,
-    _valid = false,
+    _valid = nil,
     _parser = vim._create_ts_parser(lang),
     _ranges_being_parsed = {},
     _cb_queues = {},
@@ -239,7 +239,7 @@ end
 --- tree in treesitter. Doesn't clear filesystem cache. Called often, so needs to be fast.
 ---@param reload boolean|nil
 function LanguageTree:invalidate(reload)
-  self._valid = false
+  self._valid = nil
   self._parser:reset()
 
   -- buffer was reloaded, reparse all trees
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/treesitter/languagetree.lua
```diff
@@ -271,38 +271,74 @@ function LanguageTree:lang()
...
+  end
+
+  return false
+end
+
 --- Returns whether this LanguageTree is valid, i.e., |LanguageTree:trees()| reflects the latest
 --- state of the source. If invalid, user should call |LanguageTree:parse()|.
----@param exclude_children boolean|nil whether to ignore the validity of children (default `false`)
+---@param exclude_children boolean? whether to ignore the validity of children (default `false`)
+---@param range boolean|Range? range to check for validity (entire tree is checked if the value is `boolean|nil`)
```
clason, line 302: That doesn't sound right  -- surely `range = false` will _not_ check the entire tree?

runtime/lua/vim/treesitter/languagetree.lua
```diff
@@ -271,38 +271,74 @@ function LanguageTree:lang()
   return self._lang
 end
 
+--- @param region Range6[]
+--- @param range? boolean|Range
+--- @return boolean
+local function intercepts_region(region, range)
```
clason, line 277: Isn't that `vim._range.intercepts`? If not, why can't that be used (or changed)?

clason: Or put another way: we should think about formalizing Regions as `Range[]` and add a (private) API for that (either as part of `vim._range` or -- if necessary -- a `vim._region`).

runtime/lua/vim/treesitter/languagetree.lua
ribru17: It's tricky; for backwards compatibility, I had it so that `not range` meant the entire tree was checked (which is incidentally the same as `range == true`), so the only difference was when range is a true `Range`. But I think to reduce confusion `is_valid()` should only accept a `Range?` rather than a `boolean|Range?`

runtime/lua/vim/treesitter/languagetree.lua
ribru17: > Isn't that `vim._range.intercepts`? If not, why can't that be used (or changed)?

It calls that function, but adds a few small conveniences for the language tree

I think formalizing the regions is a good idea. Maybe can be done as a follow-up

PR #32079, opened 2025-01-17
fix(diagnostic)!: make virtual text handler opt-in

gpanders:
Making this opt-out (on by default) was the wrong choice from the beginning. It is too visually noisy to be enabled by default.

BREAKING CHANGE: Users must opt-in to the diagnostic virtual text handler by adding `vim.diagnostic.config({ virtual_text = true })` to their config.

### DIFF
--- a/runtime/doc/diagnostic.txt
+++ b/runtime/doc/diagnostic.txt
@@ -464,7 +464,7 @@ Lua module: vim.diagnostic                                    *diagnostic-api*
     Fields: ~
       • {underline}?         (`boolean|vim.diagnostic.Opts.Underline|fun(namespace: integer, bufnr:integer): vim.diagnostic.Opts.Underline`, default: `true`)
                              Use underline for diagnostics.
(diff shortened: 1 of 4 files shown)


MariaSolOs (approved):

mfussenegger (approved):

justinmk:
Why was this obviously the wrong choice, where was this discussed? This seems contrary to the other work we've done to enable common lsp features by default, if lsp is working.

If discovering how to opt-out of this is hard, disabling it by default means it will be discovered by much fewer people from the other direction.

justinmk:
As a compromise, how about a mode that shows something like `x warnings ("grq" for details)` on relevant lines?

clason:
Would be just as noisy by default (and enabling the actual virtual text is a single line).

justinmk:
> Would be just as noisy by default

"Noisy"? I have not seen any report from users about this. Is there one?

The "noise" comes from the LSP server, which the user has already opted-in to.

justinmk:
Followup: 

- we still enable signs and highlights/underlines.
- future idea for "discovery" more context hints in the right-click mouse menu.

So this is probably fine.

PR #32080, opened 2025-01-17
feat(options): writetimestamp

liamrlawrence:
New option `writetimestamp` `wt` that can be set to a format string. This will cause a timestamp to be displayed after writing or appending a file. Disabled if empty.

Created `os_ctimefmt()` which is like `os_ctime()` with an additional argument to accept a format string for the timestamp.

`os_ctime_r()` used a constant format string, moved the core logic into `os_ctimefmt_r()` for increased flexibility.

:set writetimestamp=%H:%M:%S

### DIFF
--- a/src/nvim/bufwrite.c
+++ b/src/nvim/bufwrite.c
@@ -42,6 +42,7 @@
 #include "nvim/os/fs_defs.h"
 #include "nvim/os/input.h"
 #include "nvim/os/os_defs.h"
+#include "nvim/os/time.h"
 #include "nvim/path.h"
 #include "nvim/pos_defs.h"
 #include "nvim/sha256.h"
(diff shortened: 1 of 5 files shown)


clason:
I don't think this needs an option. If it's useful, it's useful and should simply be always shown. (Otherwise people will want to configure the format, and...)

Also, what problem does this actually solve?  (Not having the option is not a problem in itself...)

liamrlawrence:
Screenshots:

image


image


image

justinmk:
How does this compare to adding a `BufWritePost` autocmd in the user's config?

> Sometimes when I'm working with multiple Neovim panes and I temporarily navigate away and come back, I'm not always sure if or when I've saved my buffers

Seems like something that might be useful as part of the `:help <c-g>` (or `g<c-g>`) output. (I've also wanted that to report the filesize in human-readable KiB/MiB/etc...)

When the use-case is "sometimes I need to check this", that's a signal that it should be opt-in via a "status" command such as `<c-g>`.

If an option is needed to control timestamp _formatting_, it should be a generic option (e.g. 'datetimeformat'), not one that is special for "writes". But I would guess that is already achieved by your OS locale settings, which Vim/Nvim try to respect.

asmodeus812:
Isnt that too easy to do vimscript, to merit changes in core ?

```vimscript
function! s:AutosaveWriteFile()
    if mode() == 'c'
        return
    endif

    let oldmod = &modified
    silent! update
    let newmod = &modified
    if newmod == 0 && newmod != oldmod
        let fn = expand('%:t') | let ln = line("$")
        let off = line2byte(ln) | let tm = strftime("%H:%M:%S")
        echo printf("Autosaved \"%s\" %dL, %dB [w] %s", fn, ln, off, tm)
    endif
endfunction

```

justinmk:
I'm in favor of enhancing CTRL-g, otherwise this is wontfix.

PR #32081, opened 2025-01-17
fix(coverity/530826): validate_opt_idx unchecked negative idx

devon-g:
Problem:
opt_idx possible negative value used as index

Solution:
check opt_idx not less than zero (kOptInvalid)

### DIFF
--- a/src/nvim/option.c
+++ b/src/nvim/option.c
@@ -980,12 +980,12 @@ static int validate_opt_idx(win_T *win, OptIndex opt_idx, int opt_flags, uint32_
 
   // Skip all options that are not window-local (used when showing
   // an already loaded buffer in a window).
-  if ((opt_flags & OPT_WINONLY) && (opt_idx == kOptInvalid || !option_is_window_local(opt_idx))) {
+  if ((opt_flags & OPT_WINONLY) && !option_is_window_local(opt_idx)) {
     return FAIL;
   }
 
   // Skip all options that are window-local (used for :vimgrep).
-  if ((opt_flags & OPT_NOWIN) && opt_idx != kOptInvalid && option_is_window_local(opt_idx)) {
+  if ((opt_flags & OPT_NOWIN) && option_is_window_local(opt_idx)) {
     return FAIL;
   }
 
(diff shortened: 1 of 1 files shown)


famiu:
for `option_has_type`, the check should be added directly as an `assert`. For `option_is_window_local` the check already exists. Also your check is incorrect since `opt_idx` should never equal `kOptInvalid` in those cases

devon-g:
I see the mistake I made, thanks! I moved the invalid opt check into option_has_type.

src/nvim/option.c
```diff
@@ -980,12 +980,12 @@ static int validate_opt_idx(win_T *win, OptIndex opt_idx, int opt_flags, uint32_
 
   // Skip all options that are not window-local (used when showing
   // an already loaded buffer in a window).
-  if ((opt_flags & OPT_WINONLY) && (opt_idx == kOptInvalid || !option_is_window_local(opt_idx))) {
+  if ((opt_flags & OPT_WINONLY) && !option_is_window_local(opt_idx)) {
```
famiu, line 983: This check is unnecessary. `option_is_window_local` already has that check.

src/nvim/option.c
```diff
@@ -980,12 +980,12 @@ static int validate_opt_idx(win_T *win, OptIndex opt_idx, int opt_flags, uint32_
...
     return FAIL;
   }
 
   // Skip all options that are window-local (used for :vimgrep).
-  if ((opt_flags & OPT_NOWIN) && opt_idx != kOptInvalid && option_is_window_local(opt_idx)) {
+  if ((opt_flags & OPT_NOWIN) && option_is_window_local(opt_idx)) {
```
famiu, line 988: ditto

src/nvim/option.c
devon-g: I'm sorry I'm a bit confused this check is removed in my commit

src/nvim/option.c
devon-g: ditto

src/nvim/option.c
famiu: I might've misread, sorry.

famiu (approved):

PR #32089, opened 2025-01-18
fix(messages): compute msg_col after last newline in ext_messages

luukvbaal:
Problem:  We want to keep track of the current message column, which is
          done very rudimentary for ext_messages; only checking if the
          message ends in a newline to reset the column, while computing
          the entire cellwidth of the message, which may contain
          (multiple) newlines not necessarily at the end (since 21718c6).
          This introduced a noticeable delay for large messages (e.g. :=vim).
Solution: Calculate the cellwidth of the message after the last newline.
          Use it to keep track of the current message column. This might
          not be a functional change currently, since it only affects
          messages with (multiple) newlines not at the end of a message,
          which I don't think we emit internally, and msg_col is reset for
          a new kind. It does fix the performance problem.

### DIFF
--- a/src/nvim/message.c
+++ b/src/nvim/message.c
@@ -153,7 +153,6 @@ static Array *msg_ext_chunks = NULL;
 static garray_T msg_ext_last_chunk = GA_INIT(sizeof(char), 40);
 static sattr_T msg_ext_last_attr = -1;
 static int msg_ext_last_hl_id;
-static size_t msg_ext_cur_len = 0;
 
 static bool msg_ext_history = false;  ///< message was added to history
 static bool msg_ext_overwrite = false;  ///< will overwrite last message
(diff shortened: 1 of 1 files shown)


src/nvim/message.c
```diff
@@ -3472,14 +3470,6 @@ void msg_advance(int col)
     msg_col = col;              // for redirection, may fill it up later
     return;
   }
-  if (ui_has(kUIMessages)) {
-    // TODO(bfredl): use byte count as a basic proxy.
-    // later on we might add proper support for formatted messages.
```
luukvbaal, line 3477: I'm not sure if this resolves the todo note, but it does make the logic for ext_messages in msg_advance() redundant. And msg_col is used in some places to determine the formatting for internal messages. Do you remember what you had in mind for "proper support for formatted mesages" @bfredl?

bfredl: Don't remember exactly, maybe stuff like markdown formatting (which implies a different column width). But it is fine to remove this `TODO` regardless, as there is no special case for ext_messages to consider anymore.

justinmk (approved):

Issue #16245, opened 2021-11-05
Unicode processing issue in terminal buffers


foonathan:
### Neovim version (nvim -v)

NVIM v0.5.1 (also present on HEAD)

### Vim (not Nvim) behaves the same?

n/a (vim doesn't have `termopen()`)

### Operating system/version

ArchLinux

### Terminal name/version

kitty 0.23.1-1 (also present with xterm and a TTY)

### $TERM environment variable

xterm-kitty

### Installation

pacman -S neovim

### How to reproduce the issue

Take the following C program that repeatedly prints a Unicode character with a pause in between:

```c
#include <stdio.h>

int main()
{
    // I haven't tested other Unicode characters.
    // The multilines and the initial whitespace seem to trigger it easier.
    const char str[] = "   │\n   │\n"; // not ASCII bar, but U+2502

    while (1)
    {
        // If I use fputs(), the error happens a lot less often.
        for (const char* ptr = str; *ptr; ++ptr)
            fputc(*ptr, stderr);

        // We need some sort of pause here to trigger it.
        int i;
        scanf("%d", &i);
    }
}
```

Compile it (e.g `gcc print.c`) and start nvim --clean. Enter `:call termopen('./a.out')` and go to insert/terminal mode. Repeatedly enter some number to execute ~10 loop iterations.

### Expected behavior

The terminal buffer contains output like this:

```
   │
   │
0
   │
   │
0
etc
```

### Actual behavior

The terminal buffer contains output like this:

```
   │
   │
0
   │
   │
0
   ��
�   │
0
   �
�   │
0
   │
   │
0
   │
   │
0
   │
   │
0
   │
   ��
�0
   │
   ��
�0
```

This output was created by copy-pasting the terminal buffer, so it is not a rendering issue - it actually contains invalid Unicode. Note that not every output is mangled, only some, and this can vary from execution to execution (sometimes it works completely fine, sometimes almost everything is mangled). It also persists when you use C++'s `u8` string literal, so it's not a Unicode encoding issue of the source file or something like that.

Given this behavior and the fact that using `fputs()` over `fputc()` fixes it, I suspect it is some sort of buffering issue where neovim receives partial UTF-8 input and concludes that it has an invalid code unit: when it then receives the rest of the code point, it thinks that this is another invalid code unit, instead of combining the two into a valid code point.

zeertzjq:
I guess this is the problem?
[...]

PR #32095, opened 2025-01-19
refactor(terminal): move split UTF-8 handling to libvterm

zeertzjq:
Apply the patch from https://github.com/neovim/neovim/issues/16245#issuecomment-964074087

Co-authored-by: erw7 <erw7.github@gmail.com>


zeertzjq:
This seems more hacky than the original solution.

PR #32099, opened 2025-01-19
feat(api): deprecate nvim_buf_add_highlight()

bfredl:
This was kept for a while as it was a useful short hand and initially (pre-0.10) matched what highlights what actually properly implemented. But now `vim.hl.range()` is a better high-level shorthand with full support for native multi-line ranges.

### DIFF
--- a/runtime/doc/deprecated.txt
+++ b/runtime/doc/deprecated.txt
@@ -22,6 +22,7 @@ API
 • nvim_out_write()	Use |nvim_echo()|.
 • nvim_err_write()	Use |nvim_echo()| with `{err=true}`.
 • nvim_err_writeln()	Use |nvim_echo()| with `{err=true}`.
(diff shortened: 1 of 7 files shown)


justinmk (approved):

runtime/doc/deprecated.txt
```diff
@@ -22,6 +22,7 @@ API
 • nvim_out_write()	Use |nvim_echo()|.
 • nvim_err_write()	Use |nvim_echo()| with `{err=true}`.
 • nvim_err_writeln()	Use |nvim_echo()| with `{err=true}`.
+• nvim_buf_add_highlight()	use |vim.hl.range()| or |nvim_buf_set_extmark()|
```
justinmk, line 25:
```suggestion
• nvim_buf_add_highlight()	Use |vim.hl.range()| or |nvim_buf_set_extmark()|
```

justinmk (approved):

PR #32122, opened 2025-01-20
feat: add Show Diagnostics to context menu

### LINKED ISSUE #32115: mouse: context menu (right click) helps discover lsp/diagnostics
## Problem

We show diagnostics (by default) in the signs column and highlight/underline, but getting more details about the warning/error is not obvious. 

Image

## Expected behavior

- When user right-clicks on a diagnostic, show...
    - a `Show diagnostics` menu item which opens the quickfix list
    - a `Configure diagnostics` menu item which goes to help that mentions
        - `vim.diagnostic.open_float()`
        - `vim.diagnostic.config({ virtual_text = true })`
- Also when right-clicking the sign column item.

siddhantdev:
Adds Show Diagnostics, Show All Diagnostics and Configure Diagnostics buttons to the context menu.

Fixes: https://github.com/neovim/neovim/issues/32115

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -442,7 +442,6 @@ do
       local urls = require('vim.ui')._get_urls()
       local url = vim.startswith(urls[1], 'http')
       local ctx = url and 'url' or (vim.lsp.get_clients({ bufnr = 0 })[1] and 'lsp' or nil)
-
       enable_ctx_menu(ctx)
     end,
   })
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -383,9 +383,12 @@ end
 do
   --- Right click popup menu
   vim.cmd([[
-    anoremenu PopUp.Go\ to\ definition      <Cmd>lua vim.lsp.buf.definition()<CR>
     amenu     PopUp.Open\ in\ web\ browser  gx
     anoremenu PopUp.Inspect                 <Cmd>Inspect<CR>
+    anoremenu PopUp.Go\ to\ definition      <Cmd>lua vim.lsp.buf.definition()<CR>
+    anoremenu PopUp.Show\ Diagnostics       <Cmd>lua vim.diagnostic.open_float()<CR>
+    anoremenu PopUp.Show\ All\ Diagnostics  <Cmd>lua vim.diagnostic.setqflist()<CR>
+    anoremenu PopUp.Configure\ Diagnostics  <Cmd>help vim.diagnostic.config()<CR>
     anoremenu PopUp.-1-                     <Nop>
     vnoremenu PopUp.Cut                     "+x
     vnoremenu PopUp.Copy                    "+y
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/_defaults.lua
```diff
@@ -383,10 +383,14 @@ end
 do
   --- Right click popup menu
   vim.cmd([[
+    anoremenu PopUp.Show\ Diagnostics       <Cmd>lua vim.diagnostic.open_float()<CR>
+    anoremenu PopUp.Show\ All\ Diagnostics  <Cmd>lua vim.diagnostic.setqflist()<CR>
+    anoremenu PopUp.Configure\ Diagnostics  <Cmd>help vim.diagnostic.config()<CR>
+    anoremenu PopUp.-1-                     <Nop>
     anoremenu PopUp.Go\ to\ definition      <Cmd>lua vim.lsp.buf.definition()<CR>
```
justinmk, line 390: `go to defintion` should be in the same "group" as diagnostics and other lsp-like menu items.

siddhantdev: Updated it accordingly.

mike325: shouldn't lsp actions have a fallback to other methods in case there's no server attached to the buffer ? maybe get a tag definition using `taglist()`

siddhantdev: The lsp actions only get added if an lsp is attached. If there's no lsp the actions aren't available to prevent any errors.

runtime/lua/vim/_defaults.lua
```diff
@@ -422,6 +442,7 @@ do
       local urls = require('vim.ui')._get_urls()
       local url = vim.startswith(urls[1], 'http')
       local ctx = url and 'url' or (vim.lsp.get_clients({ bufnr = 0 })[1] and 'lsp' or nil)
+
```
justinmk, line 445:
```suggestion
```

runtime/lua/vim/_defaults.lua
```diff
@@ -403,13 +406,30 @@ do
...
     ]])
 
     if ctx == 'url' then
       vim.cmd([[amenu enable PopUp.Open\ in\ web\ browser]])
     elseif ctx == 'lsp' then
       vim.cmd([[anoremenu enable PopUp.Go\ to\ definition]])
     end
+
+    local lnum = vim.fn.getcurpos()[2] - 1 ---@type integer
+    local diagnostic = false
```
justinmk, line 421: This doesn't fit the `ctx` pattern of this function. But will go ahead and merge to unblock.

siddhantdev: Perhaps instead of `ctx` being just a string, it could be a table instead? That way additional info such as the diagnostic stuff could be passed and checking for it won't be awkward. Should I make that change?

justinmk: Not sure. Maybe `ctx` isn't needed anymore. Its original purpose is not clear after some refactors.

siddhantdev: Even I don't see why the checking is done before the function is called. It could just be done in place.

justinmk: It was needed, but later refactors changed the logic so now all the logic can live in this function.

siddhantdev: I have opened a PR https://github.com/neovim/neovim/pull/32391

PR #32131, opened 2025-01-20
vim-patch:9.1.1042: filetype: just files are not recognized

clason:
Problem:  filetype: just files are not recognized
Solution: adjust filetype detection pattern, detect just shebang line,
          include just ftplugin, indent and syntax plugin
          (Peter Benjamin)

closes: vim/vim#16466

https://github.com/vim/vim/commit/72755b3c8e91ec90447969b736f080e0de36003d

Co-authored-by: Peter Benjamin <petermbenjamin@gmail.com>


zeertzjq:
Why is this in draft?

clason:
Because I haven't decided yet how to deal with the new case-insensitive flag.

clason:
I have decided that "case insensitive" means "reasonable variants", so only consider the initial letter. (This is consistent with how we handled other cases, pun intended.)

zeertzjq:
How about all-caps?

zeertzjq (approved):

PR #32138, opened 2025-01-21
feat(lsp): support color preview for completion items

glepnir:
Problem: Color completion items display as plain text without visual preview

Solution: Parse RGB/hex colors from documentation and render with colored symbol ■

### DIFF
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -309,6 +309,7 @@ LSP
 • Support for `workspace/codeLens/refresh`:
   https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens_refresh
 • `gx` will use `textDocument/documentLink` if available.
+• |vim.lsp.completion.enable()| adds colored symbol preview for color items.
 
 LUA
 
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -276,6 +276,51 @@ local function match_item_by_value(value, prefix)
   return vim.startswith(value, prefix), nil
 end
 
+--- Generate kind text for completion color items
+--- Parse color from doc and return colored symbol ■
+---
+---@param item table completion item with kind and documentation
+---@return string? kind text or "■" for colors
+---@return string? highlight group for colors
+local function generate_kind(item)
+  if not lsp.protocol.CompletionItemKind[item.kind] then
+    return 'Unknown'
+  end
+  if item.kind ~= lsp.protocol.CompletionItemKind.Color then
+    return lsp.protocol.CompletionItemKind[item.kind]
+  end
+  local doc = get_doc(item)
+  if #doc == 0 then
+    return
+  end
+
+  -- extract hex from RGB format
+  local r, g, b = doc:match('rgb%((%d+)%s*,?%s*(%d+)%s*,?%s*(%d+)%)')
+  local hex = r and string.format('%02x%02x%02x', tonumber(r), tonumber(g), tonumber(b))
+    or doc:match('#?([%da-fA-F]+)')
+
+  if not hex then
+    return
+  end
+
+  -- expand 3-digit hex to 6-digit
+  if #hex == 3 then
+    hex = hex:gsub('.', '%1%1')
+  end
+
+  if #hex ~= 6 then
+    return
+  end
+
+  hex = hex:lower()
+  local group = ('@lsp.color.%s'):format(hex)
+  if #api.nvim_get_hl(0, { name = group }) == 0 then
+    api.nvim_set_hl(0, group, { fg = '#' .. hex })
+  end
+
+  return '■', group
+end
+
 --- Turns the result of a `textDocument/completion` request into vim-compatible
 --- |complete-items|.
 ---
@@ -358,16 +403,18 @@ function M._lsp_to_complete_items(
       then
         hl_group = 'DiagnosticDeprecated'
       end
+      local kind, kind_hlgroup = generate_kind(item)
       local completion_item = {
         word = word,
         abbr = item.label,
-        kind = protocol.CompletionItemKind[item.kind] or 'Unknown',
+        kind = kind,
         menu = item.detail or '',
         info = get_doc(item),
         icase = 1,
         dup = 1,
         empty = 1,
         abbr_hlgroup = hl_group,
+        kind_hlgroup = kind_hlgroup,
         user_data = {
           nvim = {
             lsp = {
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -164,6 +164,21 @@ describe('vim.lsp.completion: item conversion', function()
     eq(expected, result)
   end)
 
+  it('generate "■" symbol with highlight group for CompletionItemKind.Color', function()
+    local completion_list = {
+      { label = 'text-red-300', kind = 16, documentation = 'color: rgb(252, 165, 165)' },
+    }
+    local result = complete('|', completion_list)
+    result = vim.tbl_map(function(x)
+      return {
+        word = x.word,
+        kind_hlgroup = x.kind_hlgroup,
+        kind = x.kind,
+      }
+    end, result.items)
+    eq({ { word = 'text-red-300', kind_hlgroup = '@lsp.color.fca5a5', kind = '■' } }, result)
+  end)
+
   ---@param prefix string
   ---@param items lsp.CompletionItem[]
   ---@param expected table[]


runtime/lua/vim/lsp/completion.lua
```diff
@@ -284,6 +304,7 @@ function M._lsp_to_complete_items(result, prefix, client_id)
   local candidates = {}
   local bufnr = api.nvim_get_current_buf()
   local user_convert = vim.tbl_get(buf_handles, bufnr, 'convert')
+  local kind_highlights = generate_kind_hlgroup()
```
MariaSolOs, line 307: Why generate this for every completion request? It only needs to be constructed once.

glepnir: highlight group can be added in runtime . i don't think this need generate once

MariaSolOs: You can still use an autocommand for tracking when the highlight groups have changed right?

glepnir: idk what you mean. there still have group does not defined in default .  you can add a highlight group from ex highlight command .  how you tracking this when you do `:hi  Reference guifg=#fffffff`

runtime/lua/vim/lsp/completion.lua
```diff
@@ -284,6 +305,7 @@ function M._lsp_to_complete_items(result, prefix, client_id)
...
```
justinmk, line 308:
```suggestion
  local kind_hlgroups = generate_kind_hlgroup()
```

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -168,6 +168,71 @@ describe('vim.lsp.completion: item conversion', function()
     eq(expected, result)
   end)
 
+  it('uses exists highlight group for kind', function()
```
justinmk, line 171:
```suggestion
  it('uses existing highlight group for kind_hlgroup', function()
```

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -168,6 +168,71 @@ describe('vim.lsp.completion: item conversion', function()
...
+    local completion_list = {
```
justinmk, line 172: is this list meant to be exhaustive (i.e. does it need to cover ALL highlight group examples), or could it be reduced to cover various kinds of edge cases

justinmk:
What is the current behavior, and how does this change it? For example, is the current behavior that there is "no highlight", or "wrong highlight because of ..." ?

mfussenegger:
> What is the current behavior, and how does this change it? For example, is the current behavior that there is "no highlight", or "wrong highlight because of ..." ?
> 
> ^ This is the general intention of the Problem description. Describe the symptom(s).

Currently there is no highlight for the kind:

image

With this change there would be:

image

I kinda prefer without, but not opinionated enough about colors to make the call on this one.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -244,6 +244,44 @@ local function match_item_by_value(value, prefix)
...
+      local r, g, b = doc:match('rgb%((%d+)%s*,?%s*(%d+)%s*,?%s*(%d+)%)')
+      if r then
+        local color = string.format('%02x%02x%02x', tonumber(r), tonumber(g), tonumber(b))
+        local group = ('@lsp.color.%s'):format(color)
+        if vim.fn.hlID(group) < 1 then
+          api.nvim_set_hl(0, group, { fg = '#' .. color })
+        end
+        return '■', group
+      end
+    end
```
mfussenegger, line 271: What kind of server is providing the color codes in the document property?

I just tested this with the css server because I was curios how this looks like, but there the color codes are in the text:

image

No strong opinion on whether we should do this at all.

glepnir: from tailwindcss..seems like also can do some work on css. I remember vscode do it

glepnir:
> I kinda prefer without, but not opinionated enough about colors to make the call on this one.

Actually, these can be implemented in `convert`. Is it just provided out of the box? Or will we provide some completion configuration interface here in the future, `vim.completion.config` then adding a field to enable highlight group or not

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -168,6 +168,92 @@ describe('vim.lsp.completion: item conversion', function()
     eq(expected, result)
   end)
 
+  it('#generate kind and kind_hlgroup', function()
```
justinmk, line 171: mention the LSP spec name in this test name?

src/nvim/highlight_group.c
```diff
@@ -339,6 +339,16 @@ static const char *highlight_init_both[] = {
   "default link @lsp.type.typeParameter @type.definition",
   "default link @lsp.type.variable      @variable",
 
+  "default link @lsp.type.text          Pmenu",
```
glepnir, line 342: @mfussenegger  Several new highlight groups have been added here and temporarily placed in the lsp.type scope.. Any ideas?

clason: Note  that these are LSP semantic tokens, which are specified in the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens). We should _not_ invent new ones.

clason: Use different names (and a different scheme) for those, e.g., `PmenuText` etc.

glepnir: I  don't think `Pmenu` prefix is available,  these are two different things.  kind also in [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind)

clason: Then `CompletionKindText` etc. The point is that you shouldn't (ab)use the semantic token highlight groups for this.

clason: Or `@lsp.kind.text` would work too (depending on whether you can and want to reuse them for `SymbolKind`).

clason: And in any case, these need to be documented _somewhere_.

(Reminder to self: I need to add a consolidated list or "table of contents" in an appropriate place since it's getting hard for colorscheme authors to know which "default groups" they need to set.)

glepnir: `@lsp.kind` soud a good scope

glepnir: maybe also need `cmp` as a secondary prefix ? `lsp.cmp.kind.xx`..because there also have symbolkind https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolKind

clason: Yes, that's what I was asking. Personally, since there is a large overlap between these and I don't see much valid reason for giving them different colors where they coincide, I would prefer to have a single `kind` table with everything that is _either_ a `CompletionItemKind` or a `SymbolKind` (or both).

glepnir: Oh, I ignored that sentence. .. becasue i was busy with other PRs before.. sry nvm

runtime/doc/news.txt
```diff
@@ -313,6 +313,7 @@ LSP
 • |vim.lsp.enable()| has been added to enable servers.
 • |vim.lsp.buf.code_action()| resolves the `command` property during the
   `codeAction/resolve` request.
+• |vim.lsp.compltion()| add kind_hlgroup and color preview.
```
max397574, line 316:
```suggestion
• |vim.lsp.completion()| add kind_hlgroup and color preview.
```

clason (changes requested):

src/nvim/highlight_group.c
```diff
@@ -339,6 +339,31 @@ static const char *highlight_init_both[] = {
   "default link @lsp.type.typeParameter @type.definition",
   "default link @lsp.type.variable      @variable",
 
+  "default link @lsp.kind.method        @lsp.type.method",
```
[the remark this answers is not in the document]
glepnir: Considering in completion, it defaults  link to Pmenu. Because kind_hlgroup will be combined with the original attr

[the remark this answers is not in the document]
clason: Yeah, I think this needs careful documentation.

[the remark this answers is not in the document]
clason: I've just merged https://github.com/neovim/neovim/pull/32789, so could you rebase on `master` and add the tag to the list at the top of `:highlight`?

clason (approved):

justinmk:
In principle this looks good, just needs cleanup

comfysage:
are the lsp kind highlights supposed to fall back to Pmenu as a default or PmenuKind?

glepnir:
If no kind hlgroup exists, the default PmenuKind will be used.

glepnir:
Removed highlight group generation for regular completion item kinds, now it only handles color completion items.
tries to extract a hex value from the doc info, then generates the preview symbol and the corresponding kind highlight group. For regular kinds, users can handle them via opts.convert.

PR #32140, opened 2025-01-21
fix(mouse): 'statuscolumn' fold and popopmenu handling

luukvbaal:
Problem:  A right-click on the 'statuscolumn' does not open the
          popupmenu, even if a cell without a clickdef is clicked.
          Clicking the %C fold item does not open/close the fold.
Solution: Open the popupmenu when there is no clickdef like right-clicking
          the sign/numbercolumn does. Fill "linebuf_vcol" when drawing the
          'statuscolumn' to handle foldcolumn item clicks.

Ref https://github.com/neovim/neovim/issues/32115#issuecomment-2602584032

### DIFF
--- a/src/nvim/drawline.c
+++ b/src/nvim/drawline.c
@@ -368,14 +368,14 @@ static int draw_virt_text_item(buf_T *buf, int col, VirtText vt, HlMode hl_mode,
 
 // TODO(bfredl): integrate with grid.c linebuf code? madness?
 static void draw_col_buf(win_T *wp, winlinevars_T *wlv, const char *text, size_t len, int attr,
-                         bool vcol)
+                         const colnr_T *fold_vcol, bool inc_vcol)
 {
   const char *ptr = text;
   while (ptr < text + len && wlv->off < wp->w_grid.cols) {
     int cells = line_putchar(wp->w_buffer, &ptr, &linebuf_char[wlv->off],
                              wp->w_grid.cols - wlv->off, wlv->off);
     int myattr = attr;
-    if (vcol) {
+    if (inc_vcol) {
       advance_color_col(wlv, wlv->vcol);
       if (wlv->color_cols && wlv->vcol == *wlv->color_cols) {
         myattr = hl_combine_attr(win_hl_attr(wp, HLF_MC), myattr);
@@ -383,7 +383,7 @@ static void draw_col_buf(win_T *wp, winlinevars_T *wlv, const char *text, size_t
     }
     for (int c = 0; c < cells; c++) {
       linebuf_attr[wlv->off] = myattr;
-      linebuf_vcol[wlv->off] = vcol ? wlv->vcol++ : -1;
+      linebuf_vcol[wlv->off] = inc_vcol ? wlv->vcol++ : fold_vcol ? *(fold_vcol++) : -1;
       wlv->off++;
     }
   }
@@ -412,7 +412,7 @@ static void draw_foldcolumn(win_T *wp, winlinevars_T *wlv)
   int fdc = compute_foldcolumn(wp, 0);
   if (fdc > 0) {
     int attr = win_hl_attr(wp, use_cursor_line_highlight(wp, wlv->lnum) ? HLF_CLF : HLF_FC);
-    fill_foldcolumn(wp, wlv->foldinfo, wlv->lnum, attr, fdc, &wlv->off, NULL);
+    fill_foldcolumn(wp, wlv->foldinfo, wlv->lnum, attr, fdc, &wlv->off, NULL, NULL);
   }
 }
 
@@ -421,8 +421,9 @@ static void draw_foldcolumn(win_T *wp, winlinevars_T *wlv)
 /// @param fdc  Current width of the foldcolumn
 /// @param[out] wlv_off  Pointer to linebuf offset, incremented for default column
 /// @param[out] out_buffer  Char array to fill, only used for 'statuscolumn'
+/// @param[out] out_vcol  vcol array to fill, only used for 'statuscolumn'
 void fill_foldcolumn(win_T *wp, foldinfo_T foldinfo, linenr_T lnum, int attr, int fdc, int *wlv_off,
-                     schar_T *out_buffer)
+                     colnr_T *out_vcol, schar_T *out_buffer)
 {
   bool closed = foldinfo.fi_level != 0 && foldinfo.fi_lines > 0;
   int level = foldinfo.fi_level;
@@ -448,10 +449,12 @@ void fill_foldcolumn(win_T *wp, foldinfo_T foldinfo, linenr_T lnum, int attr, in
       symbol = schar_from_ascii('>');
     }
 
+    int vcol = i >= level ? -1 : (i == closedcol - 1 && closed) ? -2 : -3;
     if (out_buffer) {
+      out_vcol[i] = vcol;
       out_buffer[i] = symbol;
     } else {
-      linebuf_vcol[*wlv_off] = i >= level ? -1 : (i == closedcol - 1 && closed) ? -2 : -3;
+      linebuf_vcol[*wlv_off] = vcol;
       linebuf_attr[*wlv_off] = attr;
       linebuf_char[(*wlv_off)++] = symbol;
     }
@@ -577,7 +580,7 @@ static void draw_lnum_col(win_T *wp, winlinevars_T *wlv)
           char *num = skipwhite(buf);
           rl_mirror_ascii(num, skiptowhite(num));
         }
-        draw_col_buf(wp, wlv, buf, (size_t)width, attr, false);
+        draw_col_buf(wp, wlv, buf, (size_t)width, attr, NULL, false);
       } else {
         draw_col_fill(wlv, schar_from_ascii(' '), width, attr);
       }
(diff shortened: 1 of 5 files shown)


src/nvim/drawline.c
```diff
@@ -448,6 +451,7 @@ void fill_foldcolumn(win_T *wp, foldinfo_T foldinfo, linenr_T lnum, int attr, in
 
     if (out_buffer) {
       out_buffer[i] = symbol;
+      out_vcol[i] = i >= level ? -1 : (i == closedcol - 1 && closed) ? -2 : -3;
```
luukvbaal, line 454: @bfredl me touchy your code; `fill_foldcolumn()` now fills an array to pass on to `draw_col_buf()` so that it can set the magic vcol values for open/closing a fold when the `%C` foldcolumn item is clicked in the statuscolumn.

src/nvim/drawline.c
```diff
@@ -368,7 +368,7 @@ static int draw_virt_text_item(buf_T *buf, int col, VirtText vt, HlMode hl_mode,
 
 // TODO(bfredl): integrate with grid.c linebuf code? madness?
 static void draw_col_buf(win_T *wp, winlinevars_T *wlv, const char *text, size_t len, int attr,
-                         bool vcol)
+                         colnr_T *fold_vcol, bool vcol)
```
zeertzjq, line 371: Can the `vcol` parameter be omitted by passing `&wlv->vcol` as `fold_vcol` here?

luukvbaal: I think so yeah.

luukvbaal: I checked that out when writing this but no I don't think so. `wlv->vcol` is not an array, its actual value needs to be incremented here. Whereas `fold_vcol` was filled beforehand and we increment its address.

Though I guess we could check if `fold_vcol` == `&wlv->vcol` and change the statement based on that but I'm not sure that would be an improvement.

zeertzjq: That's not what you are doing here. `(*fold_vcol)++` is incrementing the value.

luukvbaal: I.e.

```c
linebuf_vcol[wlv->off] = vcol == &wlv->vcol ? wlv->vcol++ : vcol ? (*vcol)++ : -1;
```

zeertzjq: > I.e.
> 
> ```c
> linebuf_vcol[wlv->off] = vcol == &wlv->vcol ? wlv->vcol++ : vcol ? (*vcol)++ : -1;
> ```

Look more carefully. Aren't `wlv->vcol++` and `(*vcol)++` the same thing?

luukvbaal: Oops you're right, it's working by accident then, because the fold close/open values are `-3` and `-2`...

zeertzjq: If you want to increment the address you should use `*(foldvcol++)`.

zeertzjq, line 371:
```suggestion
                         const colnr_T *fold_vcol, bool vcol)
```

luukvbaal: So would this be an improvement then?
```diff
-                         colnr_T *fold_vcol, bool vcol)
+                         colnr_T *vcol)
 {
   const char *ptr = text;
   while (ptr < text + len && wlv->off < wp->w_grid.cols) {
     int cells = line_putchar(wp->w_buffer, &ptr, &linebuf_char[wlv->off],
                              wp->w_grid.cols - wlv->off, wlv->off);
     int myattr = attr;
-    if (vcol) {
+    if (vcol == &wlv->vcol) {
       advance_color_col(wlv, wlv->vcol);
       if (wlv->color_cols && wlv->vcol == *wlv->color_cols) {
         myattr = hl_combine_attr(win_hl_attr(wp, HLF_MC), myattr);
@@ -383,7 +383,7 @@ static void draw_col_buf(win_T *wp, winlinevars_T *wlv, const char *text, size_t
     }
     for (int c = 0; c < cells; c++) {
       linebuf_attr[wlv->off] = myattr;
-      linebuf_vcol[wlv->off] = vcol ? wlv->vcol++ : fold_vcol ? *(fold_vcol++) : -1;
+      linebuf_vcol[wlv->off] = vcol == &wlv->vcol ? (*vcol)++ : vcol ? *(vcol++) : -1;
       wlv->off++;
     }
   }
```

zeertzjq: Also I wonder if the `vcol` parameter name should now be renamed to `inc_vcol` to avoid confusion.

src/nvim/drawline.c
```diff
@@ -450,6 +451,7 @@ void fill_foldcolumn(win_T *wp, foldinfo_T foldinfo, linenr_T lnum, int attr, in
 
     if (out_buffer) {
       out_buffer[i] = symbol;
+      out_vcol[i] = i >= level ? -1 : (i == closedcol - 1 && closed) ? -2 : -3;
     } else {
       linebuf_vcol[*wlv_off] = i >= level ? -1 : (i == closedcol - 1 && closed) ? -2 : -3;
```
zeertzjq, line 456: This expression is present in both branches of the `if (out_buffer)` condition. Assign it to a variable before the condition?

src/nvim/mouse.c
```diff
@@ -844,13 +836,10 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
...
     if (fixindent) {
-      c1 = (dir == BACKWARD) ? '[' : ']';
-      c2 = 'p';
+      prep_redo(regname, count, NUL, (dir == BACKWARD) ? '[' : ']', NUL, 'p', NUL);
     } else {
-      c1 = (dir == FORWARD) ? 'p' : 'P';
-      c2 = NUL;
+      prep_redo(regname, count, NUL, (dir == FORWARD) ? 'p' : 'P', NUL, NUL, NUL);
     }
-    prep_redo(regname, count, NUL, c1, NUL, c2, NUL);
```
zeertzjq, line 853: Why remove the `c1` and `c2` variables? That makes the code harder to understand.

src/nvim/mouse.c
```diff
@@ -376,6 +406,7 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
     return false;
   }
 
+  static bool in_tab_line = false;   // mouse clicked in tab line
```
zeertzjq, line 409: Having a static variable appear randomly in the middle of a function is strange. Move it back to the top.

src/nvim/mouse.c
```diff
@@ -829,9 +821,9 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
     stuffcharReadbuff(Ctrl_O);
   }
 
+  static pos_T orig_cursor;
```
zeertzjq, line 824: ditto

src/nvim/mouse.c
luukvbaal: I tend to agree, though conversely reading these at the top of such a large function tells nothing about where they're actually used.

test/functional/ui/statuscolumn_spec.lua
```diff
@@ -713,6 +744,40 @@ describe('statuscolumn', function()
...
+          {8:|}{7:  }{8:|}aaaaa                                            |*10
+                                                               |
+        ]])
+        -- Note: alternate between left and right mouse buttons to avoid triggering multiclicks
+        api.nvim_input_mouse('right', 'press', '', 0, 0, 2)
+        feed('<Esc>') -- Close popupmenu
+        api.nvim_input_mouse('left', 'press', '', 0, 0, 1)
+        screen:expect(s1)
+        api.nvim_input_mouse('right', 'press', '', 0, 0, 2)
+        feed('<Esc>') -- Close popupmenu
```
zeertzjq, line 773: Adding right clicks for the sole purpose of avoid multiclicks is a bit awkward. How about setting `'mousetime'` to zero instead?

src/nvim/mouse.c
zeertzjq: Yes, but these two variables are used in multiple places that are quite far apart :)

test/functional/ui/statuscolumn_spec.lua
luukvbaal: I copied this from a test in `ui/statusline_spec.lua` (which you originally added), but if that works sure.

src/nvim/mouse.c
```diff
@@ -705,6 +688,12 @@ bool do_mouse(oparg_T *oap, int c, int dir, int count, bool fixindent)
     if (click_defs != NULL) {
       switch (click_defs[click_col].type) {
       case kStlClickDisabled:
+        // If there is no click definition, still open the popupmenu for a
+        // statuscolumn click like a click in the sign/numbercolumn does.
```
zeertzjq, line 692:
```suggestion
        // statuscolumn click like a click in the sign/number column does.
```

src/nvim/mouse.c
```diff
@@ -1496,7 +1489,7 @@ int jump_to_mouse(int flags, bool *inclusive, int which_button)
       }
     }
   }
-
+foldclick:;
```
zeertzjq, line 1492: I think there should be a blank line above this label.
```suggestion

foldclick:;
```

src/nvim/mouse.c
```diff
@@ -1258,8 +1253,6 @@ int jump_to_mouse(int flags, bool *inclusive, int which_button)
   if (flags & MOUSE_SETPOS) {
     goto retnomove;                             // ugly goto...
   }
-  win_T *old_curwin = curwin;
-  pos_T old_cursor = curwin->w_cursor;
```
zeertzjq, line 1262: What is the point of this change?

luukvbaal: No idea, probably failed to move it back to it's original position after a different change. Thanks for the thorough review. 

The jump happens at line 1359, so it seems this could be moved down to before that if statement.

zeertzjq (approved):

Issue #30905, opened 2024-10-22
LSP completion: wrong startcol if label is not a keyword


przepompownia:
### Problem

For a snippet `/**  */` labeled with `/**` I get `/**/**  */` after completion at the end of line containing `/**` (see repros).

### Steps to reproduce

I have prepared both the test case and the repro to manual usage (extracted from existing tests).

Test case:
```lua
  it('starts in correct place if label is not a keyword', function()
    local completion_list = {
      isIncomplete = false,
      items = {
        {
          label = '/**',
          insertText = '/** ${1:hello} */',
          insertTextFormat = 2,
        },
      },
    }
    exec_lua(function()
      vim.o.completeopt = 'menuone,noselect'
    end)
    create_server(completion_list)
    feed('0i/<c-x><c-o>')
    retry(nil, nil, function()
      eq(
        1,
        exec_lua(function()
          return vim.fn.pumvisible()
        end)
      )
    end)
    feed('<C-n><C-y>')
    eq(
      { true, { '/** ${1:hello} */' } },
      exec_lua(function()
        return {
          vim.snippet.active({ direction = 1 }),
          vim.api.nvim_buf_get_lines(0, 0, -1, true),
        }
      end)
    ) -- this assertion fails
    feed('<tab>')
    eq(
      #'/** hello */',
      exec_lua(function()
        return vim.api.nvim_win_get_cursor(0)[2]
      end)
    )
  end)
```

init.lua:
```lua
function _G._create_server(opts)
  opts = opts or {}
  local server = {}
  server.messages = {}

  function server.cmd(dispatchers)
    local closing = false
    local handlers = opts.handlers or {}
    local srv = {}

    function srv.request(method, params, callback)
      table.insert(server.messages, {
        method = method,
        params = params,
      })
      local handler = handlers[method]
      if handler then
        handler(method, params, callback)
      elseif method == 'initialize' then
        callback(nil, {
          capabilities = opts.capabilities or {},
        })
      elseif method == 'shutdown' then
        callback(nil, nil)
      end
      local request_id = #server.messages
      return true, request_id
    end

    function srv.notify(method, params)
      table.insert(server.messages, {
        method = method,
        params = params,
      })
      if method == 'exit' then
        dispatchers.on_exit(0, 15)
      end
    end

    function srv.is_closing()
      return closing
    end

    function srv.terminate()
      closing = true
    end

    return srv
  end

  return server
end

local completion_result = {
  isIncomplete = false,
  items = {
    {
[...]

PR #32142, opened 2025-01-21
fix(lsp): support editsNearCursor in completion 

### LINKED ISSUE #39001: LSP: omnifunc (clangd) unexpectedly deletes "->" when completeopt contains "longest"
### Problem

Since version 0.12.0, LSP omni completion in C++ code with clangd unexpectedly deletes `->` in some contexts.

https://github.com/user-attachments/assets/bb862d24-6815-4a48-87f2-c16e0619c4e6

I have bisected this bug and found the first bad commit to be 36db6ff2c128864840e2820491a2172d6b1b7e62 (https://github.com/neovim/neovim/pull/37491)

### Steps to reproduce using "nvim --clean -u minimal_init.lua"

Minimal init.lua:
```lua
vim.opt.completeopt = { "menu", "popup", "longest" } -- "longest" required for repro

vim.lsp.config['clangd'] = {
  cmd = { 'clangd' },
  filetypes = { 'c', 'cpp', 'objc', 'objcpp', 'cuda' },
  root_markers = {
    '.clangd',
    '.clang-tidy',
    '.clang-format',
    'compile_commands.json',
    'compile_flags.txt',
    'configure.ac', -- AutoTools
    '.git',
  },
  capabilities = {
    textDocument = {
      completion = {
        editsNearCursor = true, -- this option is required for repro
      },
    },
    offsetEncoding = { 'utf-8', 'utf-16' },
  },
}

vim.lsp.enable('clangd')

  callback = function(ev)
    local client = vim.lsp.get_client_by_id(ev.data.client_id)
    if client:supports_method('textDocument/completion') then
      vim.lsp.completion.enable(true, client.id, ev.buf, { autotrigger = false })
    end
  end,
})
```

I found that `editsNearCursor = true` (as configured by [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/clangd.lua#L84)) is required to reproduce this bug.

This bug only seems to occur when `completeopt` contains `longest`

---

To trigger the issue, invoke omni completion for a function call to a C++ object wrapper in `std::unique_ptr`.
The issue does not occur when using omni completion for plain / bare pointers.
Using the following code example, omni completion works as expected for `foo` and `bar` but not `baz`.
```c++
#include <memory>
#include <vector>

int main(int argc, const char *argv[]) {
	std::vector<int> foo;
	std::vector<int> *bar {nullptr};
	std::unique_ptr<std::vector<int>> baz;

	return 0;
}
```

Typing `baz->i_CTRL-X_CTRL-O` leads to `->` being deleted when the completion menu is shown.

If using LSP `autotrigger = true` this bug makes it extremely difficult to write any code using unique_ptr or similar wrapper classes.

### Expected behavior

In C++ code, omni completion should never delete the `->` (or `.`) operators.

This bug did not exist in neovim 0.11

### Nvim version (nvim -v)

NVIM v0.12.1

Apple clangd version 17.0.0 (clang-1700.6.4.2)

### Operating system/version

macOS 15.7.5

### Log file

[lsp.log](https://github.com/user-attachments/files/26666567/lsp.log)

glepnir:
Problem: clangd with editsNearCursor returns items with different
textEdit ranges. Prefix matching and word construction
break for these items, and dot-to-arrow edits are not applied on confirm.

Solution: Use per-item prefix based on textEdit range.end. Apply
textEdit.newText on CompleteDone when it differs from the inserted word.

Fix #39001

ref: https://github.com/clangd/clangd/issues/2201

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -64,7 +64,6 @@ local buf_handles = {}
 --- @nodoc
 --- @class vim.lsp.completion.Context
 local Context = {
-  cursor = nil, --- @type [integer, integer]?
   last_request_time = nil, --- @type integer?
   pending_requests = {}, --- @type function[]
   isIncomplete = false,
@@ -83,7 +82,6 @@ end
 
 --- @nodoc
 function Context:reset()
-  -- Note that the cursor isn't reset here, it needs to survive a `CompleteDone` event.
   self.isIncomplete = false
   self.last_request_time = nil
   self:cancel_pending()
@@ -246,6 +244,19 @@ local function apply_defaults(item, defaults)
       textEdit.replace = defaults.editRange.replace
     end
   end
+end
+
+--- Leading bytes of `prefix` the server rewrites (e.g. `p.`->`p->`).
+---
+--- @param prefix string boundary..cursor text
+--- @param new_text string textEdit.newText
+--- @return integer # 0 if newText already starts with the operator
+local function rewritten_op_len(prefix, new_text)
+  local op = prefix:match('^[^%w_]*') --[[@as string]]
+  if op == '' or vim.startswith(new_text, op) then
+    return 0
+  end
+  return #op
 end
 
 --- @param result vim.lsp.CompletionResult
@@ -418,26 +429,21 @@ function M._lsp_to_complete_items(
     return {}
   end
 
-  ---@type fun(item: lsp.CompletionItem):boolean
-  local matches
-  if not prefix:find('%w') then
-    matches = function(_)
+  --- @param item lsp.CompletionItem
+  --- @param pfx string
+  --- @return boolean, integer?
+  local function matches(item, pfx)
+    if not pfx:find('%w') then
       return true
     end
-  else
-    ---@param item lsp.CompletionItem
-    matches = function(item)
-      if item.filterText then
-        return match_item_by_value(item.filterText, prefix)
-      end
-
-      if item.textEdit and not item.textEdit.newText then
-        -- server took care of filtering
-        return true
-      end
-
-      return match_item_by_value(item.label, prefix)
+    if item.filterText then
+      return match_item_by_value(item.filterText, pfx)
     end
+    if item.textEdit and not item.textEdit.newText then
+      -- server took care of filtering
+      return true
+    end
+    return match_item_by_value(item.label, pfx)
   end
 
   local candidates = {}
@@ -447,25 +453,43 @@ function M._lsp_to_complete_items(
   local client = client_id and vim.lsp.get_client_by_id(client_id)
   local server_supports_resolve = client and client:supports_method('completionItem/resolve')
   for _, item in ipairs(items) do
-    local match, score = matches(item)
-    if match then
-      local word = get_completion_word(item, prefix, match_item_by_value)
+    local item_start_byte --- @type integer?
+    local op_len = 0
 
-      if server_start_boundary and line and lnum and encoding and item.textEdit then
-        --- @type integer?
-        local item_start_char
-        if item.textEdit.range and item.textEdit.range.start.line == lnum then
-          item_start_char = item.textEdit.range.start.character
-        elseif item.textEdit.insert and item.textEdit.insert.start.line == lnum then
-          item_start_char = item.textEdit.insert.start.character
+    if server_start_boundary and line and lnum and encoding and item.textEdit then
+      local range = item.textEdit.range or item.textEdit.insert
+      if range and range.start.line == lnum and range['end'].line == lnum then
+        item_start_byte = vim.str_byteindex(line, encoding, range.start.character, false)
+        local item_end_byte = vim.str_byteindex(line, encoding, range['end'].character, false)
+        if
+          item.textEdit.newText
+          and item_start_byte == server_start_boundary
+          and item_end_byte > item_start_byte
+        then
+          op_len = rewritten_op_len(prefix, item.textEdit.newText)
         end
+      end
+    end
 
-        if item_start_char then
-          local item_start_byte = vim.str_byteindex(line, encoding, item_start_char, false)
-          if item_start_byte > server_start_boundary then
-            local missing_prefix = line:sub(server_start_boundary + 1, item_start_byte)
-            word = missing_prefix .. word
-          end
+    -- filter only on the part of `prefix` the item covers; drop the operator it rewrites
+    local effective_prefix = prefix
+    if item_start_byte and item_start_byte > server_start_boundary then
+      effective_prefix = prefix:sub(item_start_byte - server_start_boundary + 1)
+    elseif op_len > 0 then
+      effective_prefix = prefix:sub(op_len + 1)
+    end
+
+    local match, score = matches(item, effective_prefix)
+    if match then
+      local word --- @type string
+      if op_len > 0 then
+        -- keep the operator so the leader matches; newText is applied on accept
+        word = prefix:sub(1, op_len) .. (item.filterText or vim.trim(item.label))
+      else
+        word = get_completion_word(item, effective_prefix, match_item_by_value)
+        if line and item_start_byte and item_start_byte > server_start_boundary then
+          -- item starts right of the boundary; keep the gap
+          word = line:sub(server_start_boundary + 1, item_start_byte) .. word
         end
       end
 
@@ -862,24 +886,54 @@ local function on_complete_done()
   local position_encoding = client.offset_encoding or 'utf-16'
   local resolve_provider = (client.server_capabilities.completionProvider or {}).resolveProvider
 
-  -- Keep reference to avoid race where completion/resolve response arrives after on_insert_leave
-  -- and Context.cursor got cleared before clear_word() gets called
-  local context_cursor = assert(Context.cursor)
+  -- inserted `word` ending at the cursor, so it started at cursor_col - #word
+  local word_start_col = math.max(0, cursor_col - #completed_item.word)
+
+  --- textEdit start col on the current line, or nil if outside the replaced span.
+  --- @return integer?
+  local function edit_start_col()
+    local te = completion_item.textEdit
+    local range = te and (te.range or te.insert)
+    if not range or range.start.line ~= cursor_row then
+      return nil
+    end
+    local col = vim.str_byteindex(
+      api.nvim_get_current_line(),
+      position_encoding,
+      range.start.character,
+      false
+    )
+    -- must be inside the replaced span
+    if col < word_start_col or col > cursor_col then
+      return nil
+    end
+    return col
+  end
+
+  -- span end is the cursor, not range['end'] (user may have typed more since the request)
+  local start_col = edit_start_col() or word_start_col
+
+  -- If the inserted word isn't newText (multi-line, filterText-as-word, or an operator
+  -- rewrite `p.x`->`p->x`), apply newText now -- not in the resolve callback below, which
+  -- may be slow or never arrive.
+  local text_edit = not expand_snippet and completion_item.textEdit or nil
+  local new_text = text_edit and text_edit.newText
+  if new_text and api.nvim_get_current_line():sub(start_col + 1, cursor_col) ~= new_text then
+    local lines = vim.split((new_text:gsub('\r\n?', '\n')), '\n', { plain = true })
+    api.nvim_buf_set_text(bufnr, cursor_row, start_col, cursor_row, cursor_col, lines)
+    local last = lines[#lines]
+    api.nvim_win_set_cursor(0, {
+      cursor_row + #lines,
+      #lines == 1 and start_col + #last or #last,
+    })
+  end
 
   local function clear_word()
     if not expand_snippet then
       return nil
     end
-
-    -- Remove the already inserted word.
-    api.nvim_buf_set_text(
-      bufnr,
-      context_cursor[1] - 1,
-      context_cursor[2] - 1,
-      cursor_row,
-      cursor_col,
-      { '' }
-    )
+    -- remove the inserted word; snippet.expand() puts newText in its place
+    api.nvim_buf_set_text(bufnr, cursor_row, start_col, cursor_row, cursor_col, { '' })
   end
 
   local function apply_snippet_and_command()
@@ -1048,7 +1102,6 @@ local function trigger(bufnr, clients, ctx)
     end
 
     local start_col = (server_start_boundary or word_boundary) + 1
-    Context.cursor = { cursor_row, start_col }
     if #matches > 0 and has_completeopt('popup') then
       local group = get_augroup(bufnr)
       if #api.nvim_get_autocmds({ buf = bufnr, event = 'CompleteChanged', group = group }) == 0 then
@@ -1111,7 +1164,6 @@ end
 
 local function on_insert_leave()
   reset_timer()
-  Context.cursor = nil
   Context:reset()
 end
 
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -1066,6 +1066,7 @@ describe('vim.lsp.completion: protocol', function()
     local item = completion_list.items[1]
     exec_lua(function()
       vim.v.completed_item = {
+        word = 'hello',
         user_data = {
           nvim = {
             lsp = { client_id = client_id, completion_item = item },
@@ -1108,6 +1109,7 @@ describe('vim.lsp.completion: protocol', function()
     local item = completion_list.items[1]
     exec_lua(function()
       vim.v.completed_item = {
+        word = 'hello',
         user_data = {
           nvim = {
             lsp = { client_id = client_id, completion_item = item },
@@ -1751,6 +1753,147 @@ describe('vim.lsp.completion: integration', function()
         return vim.fn.complete_info({ 'selected' }).selected
       end)
     )
+  end)
+
+  it('dot-to-arrow with isIncomplete', function()
+    local completion_list = {
+      isIncomplete = true,
+      items = {
+        {
+          detail = 'handle_T',
+          filterText = 'handle',
+          insertText = '->handle',
+          insertTextFormat = 1,
+          kind = 5,
+          label = ' handle',
+          score = 0.43193808197975,
+          sortText = '4122d903handle',
+          textEdit = {
+            newText = '->handle',
+            range = {
+              ['end'] = {
+                character = 3,
+                line = 0,
+              },
+              start = {
+                character = 2,
+                line = 0,
+              },
+            },
+          },
+        },
+        {
+          detail = 'int',
+          filterText = 'w_alt_fnum',
+          insertText = '->w_alt_fnum',
+          insertTextFormat = 1,
+          kind = 5,
+          label = ' w_alt_fnum',
+          score = 0.43193808197975,
+          sortText = '4122d903w_alt_fnum',
+          textEdit = {
+            newText = '->w_alt_fnum',
+            range = {
+              ['end'] = {
+                character = 3,
+                line = 0,
+              },
+              start = {
+                character = 2,
+                line = 0,
+              },
+            },
+          },
+        },
+        {
+          detail = 'pointer',
+          filterText = 'get',
+          insertText = '.get()',
+          insertTextFormat = 2,
+          kind = 2,
+          label = ' get',
+          labelDetails = {
+            detail = '() const',
+          },
+          score = 0.46563199162483,
+          sortText = '411198afget',
+          textEdit = {
+            newText = '.get()',
+            range = {
+              ['end'] = {
+                character = 4,
+                line = 0,
+              },
+              start = {
+                character = 2,
+                line = 0,
+              },
+            },
+          },
+        },
+      },
+    }
+    exec_lua(function()
+      vim.o.completeopt = 'menu,menuone,noinsert'
+    end)
+    create_server('dummy', completion_list)
+
+    feed('iwp.<c-x><c-o>')
+    wait_for_pum()
+
+    feed('w')
+    wait_for_pum()
+
+    -- accept textEdit
+    feed('<C-y>')
+    eq(
+      { 'wp->w_alt_fnum', { 1, 14 } },
+      exec_lua(function()
+        return { vim.api.nvim_get_current_line(), vim.api.nvim_win_get_cursor(0) }
+      end)
+    )
+
+    n.command('set completeopt+=longest') --- #39001
+    feed('<ESC>Swp-><C-x><C-O>')
+    wait_for_pum()
+    feed('<C-N><C-y>')
+    eq('wp.get()', n.api.nvim_get_current_line())
+  end)
+
+  it('accept text.Edit when end_col equals the column at the end of the line.', function()
+    create_server('dummy1', {
+      isIncomplete = false,
+      items = {
+        {
+          detail = 'runtime/lua/vim/treesitter.lua',
+          insertTextFormat = 1,
+          kind = 17,
+          label = 'vim.treesitter',
+          sortText = '0082',
+          textEdit = {
+            newText = 'vim.treesitter',
+            range = {
+              ['end'] = {
+                character = 13,
+                line = 0,
+              },
+              start = {
+                character = 9,
+                line = 0,
+              },
+            },
+          },
+        },
+      },
+    })
+    exec_lua(function()
+      vim.o.completeopt = 'menu,menuone,noinsert'
+    end)
+    n.api.nvim_buf_set_lines(0, 0, -1, false, { 'require("vim.")' })
+    feed('A<left><left><C-x><C-O>')
+    wait_for_pum()
+    feed('<C-Y>')
+    eq('require("vim.treesitter")', exec_lua([[return vim.api.nvim_get_current_line()]]))
   end)
 end)
 
--- a/runtime/doc/news.txt
+++ b/runtime/doc/news.txt
@@ -248,6 +248,8 @@ LSP
   |vim.lsp.buf.type_definition()|, and |vim.lsp.buf.implementation()|
   now supports passing arbitrary positions.
 • Support for nested snippets.
+• |lsp-completion| applies `textEdit` when it differs from the typed word,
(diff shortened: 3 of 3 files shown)


przepompownia:
When looked if this PR can be related to #30905 I see (as a side effect) that multiline snippets from textEdit are not expanded here. Tried on
<details>

<summary>init.lua</summary>

```lua
local function create_server(opts)
  opts = opts or {}
  local server = {}
  server.messages = {}

  function server.cmd(dispatchers)
    local closing = false
    local handlers = opts.handlers or {}
    local srv = {}

    function srv.request(method, params, callback)
      table.insert(server.messages, {
        method = method,
        params = params,
      })
      if handler then
        handler(method, params, callback)
      elseif method == 'initialize' then
        callback(nil, {
          capabilities = opts.capabilities or {},
        })
      elseif method == 'shutdown' then
        callback(nil, nil)
      end
      local request_id = #server.messages
      return true, request_id
    end

    function srv.notify(method, params)
      table.insert(server.messages, {
        method = method,
        params = params,
      })
      if method == 'exit' then
        dispatchers.on_exit(0, 15)
      end
    end

    function srv.is_closing()
      return closing
    end

    function srv.terminate()
      closing = true
    end

    return srv
  end

  return server
end

local fun = table.concat({
  '${1|public,private,protected|} static function ${2:name}($3): ${4:void}',
  '{$0',
  '}',
}, '\n')

local textEdit = {
  range = {
    start = {line = 0, character = 0},
    ['end'] = {line = 0, character = 2},
  },
  newText = fun,
}

local completion_result = {
  isIncomplete = false,
  items = {
    {
      label = '/**',
      insertText = '/** ${1:hello} */',
      insertTextFormat = 2,
      kind = 15,
      detail = 'Snippet with non-word label',
    },
    {
      label = 'f',
      insertText = fun,
      insertTextFormat = 2,
      kind = 15,
      detail = 'Multiline snippet',
    },
    {
      label = 'fu',
      kind = vim.lsp.protocol.CompletionItemKind.Snippet,
      insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet,
      detail = 'Multiline with textEdit',
      insertText = insertText,
      textEdit = textEdit,
    }
  },
}

local server = create_server({
  capabilities = {
    completionProvider = {},
  },
  handlers = {
    ['textDocument/completion'] = function (_, _, callback)
      callback(nil, completion_result)
    end,
  },
})

-- vim.o.completeopt = 'menuone,noinsert'
vim.o.completeopt = 'menuone,noselect'
vim.fn.complete = vim.schedule_wrap(vim.fn.complete)

vim.api.nvim_create_autocmd('UIEnter', {
  callback = function ()
    vim.lsp.start({
      name = 'dummy',
      cmd = server.cmd,
      on_attach = function (client, bufnr0)
        vim.lsp.completion.enable(true, client.id, bufnr0, {})
      end,
    })
  end
})
```

</details>


v0.11.0-dev-1702+g44740e561f (master):
image

v0.11.0-dev-1608+g1540d90db1 (this PR):
image

glepnir:
idk i haven't look at  #30905  but i will take a look at it later.

przepompownia:
> now the fu can expand correctly

Works for me too, thanks

runtime/lua/vim/lsp/completion.lua
```diff
@@ -441,7 +445,7 @@ local function trigger(bufnr, clients)
   reset_timer()
   Context:cancel_pending()
 
-  if tonumber(vim.fn.pumvisible()) == 1 and not Context.isIncomplete then
+  if tonumber(vim.fn.pumvisible()) == 1 then
```
mfussenegger, line 448:
```suggestion
  if tonumber(vim.fn.pumvisible()) == 1 and not Context.isIncomplete then
```

This looks unrelated?

glepnir: if keep there , continuing input will clean pum `wp.` then input w. pum is cleaned

mfussenegger: But isn't removing it breaking the functionality to fetch more results if the initial results were incomplete?
Is there some state hiccup that causes isIncomplete to have the wrong value?

mfussenegger: Maybe worth reporting upstream with clangd? In the meantime we could instead guard the `vim.fn.complete(start_col, matches)` and call it only if there are matches given that we rely on the pum mechanism to do the filtering we don't need a response to "wipe out" any shown candidates.

glepnir: Wait, let me research it.

glepnir: Resolved. and another question here is `Context.isIncomplete` should bind with client ? if one client return isIncomplete false do we want still send with all clients ?

glepnir: > (e.g. remove previous results for client-2, add new results for client-2, but keep results from client-1)

Can do in C side I prefer to do a soft delete here, that is, set an unavailable flag in the compl_T structure. Otherwise, there will be some memory allocation efficiency issues.

> Independent of that, do I understand it right that this PR is otherwise good to go?

I changed isInComplete to a table storing clientid. Do I need to revert to the original state? So far no tests have been broken. But maybe that means no catch.

przepompownia:
For some reason I cannot get completion result on `vim.api.` with LuaLs whilst I have no problem on master with it. If you cannot reproduce it, I can try to provide some example.

glepnir:
I can't reproduce do you have mutliple clients on lua file ?

przepompownia:
Yes, there is none-ls providing completion too, but even after stopping it I still cannot get that list. 

I may have no time to prepare repro now, but see that the luals completion works if Nvim starts without none-ls.

glepnir:
Okay, I reverted the change of isInComplete. Maybe I can reprocess it after extending complete_add.

przepompownia:
> Okay, I reverted the change of isInComplete. Maybe I can reprocess it after extending complete_add.

Now I see results again with two completion providers :+1:

mfussenegger:
Looks like this causes issues with lemminx. In a java pom.xml typing `<d` shows `dependency` as candidate and ends up completing `<<dependency></dependency>`. Without this change it correctly completes it as `<dependency></dependency>`

Item returned from the server is:

```
 data = {
   offset = 3539,
   participantId = "org.eclipse.lemminx.services.extensions.completion.ElementEndTagCompletionResolver",
   updateEndTagName = false,
   uri = "file:///path/to/pom.xml"
 },
 documentation = {
   kind = "markdown",
   value = "3.0.0+\n\nThe `<dependency>` element contains information about a dependency of the project.\n\nSource: [maven-4.0.0.xsd](...)"
 },
 filterText = "<dependency",
 insertTextFormat = 2,
 kind = 10,
 label = "dependency",
 sortText = "dependency",
 textEdit = {
   newText = "<dependency>$1</dependency>$0",
   range = {
     ["end"] = {
         character = 6,
         line = 113
     },
     start = {
         character = 4,
         line = 113
     }
   }
 },
 textEditText = "<dependency>$1</dependency>$0"
```

przepompownia:
Confirmed for lemminx (while `<` does not belong to `iskeyword`).

glepnir:
how about now ?

przepompownia:
Now it's ok at least on my repro and lemminx. Both have single completion provider. 

For two providers I still get incorrectly recognized start of word but if I understood, at the moment (after [0efb2c4](https://github.com/neovim/neovim/commit/0efb2c4a81f32b2563dcf159e5b2ce08d1fbb611)) that case fell out of scope of this PR, is that right?

przepompownia:
...and again completion on luals fails with none-ls. Maybe we need to add a test that covers such case here.

glepnir:
emmm I don't modified `isInComplete` just change the prefix value when context.cursor have value ..how can i reproduce it ? i don' use none-ls. at least it works fine when i use `phoenix` a server which provider path/word a fake lsp and luals

przepompownia:
I'll then try to create repro with two completion in free time, maybe with looking how none-ls (especially its adapter I wrote and use) does it. You can look [there](https://github.com/nvimtools/none-ls.nvim/blob/main/lua/null-ls/builtins/completion/nvim_snippets.lua) before - maybe you'll find something specific sooner than me.

glepnir:
Aha i know what happen here. seems like Context.cursor does not reset when complete done.. @mfussenegger  when we trigger completedone there still need reset Context.cursor and timer right that mean calling `on_insert_leave` in CompleteDone

przepompownia:
Thanks. Now lemminx, my recent repro and completion list works. Only double providers require fix, still not sure if in this PR (and I sitll have a plan to prepare some minimal example for that, without none-ls).

przepompownia:
I went through double server case and found that the following issue
image
is reproducible also with my recent (single client) repro, even simplified, but with indentation. 

<details>

<summary>init.lua</summary>

```lua
local function create_server(opts)
  opts = opts or {}
  local server = {}
  server.messages = {}

  function server.cmd(dispatchers)
    local closing = false
    local handlers = opts.handlers or {}
    local srv = {}

    function srv.request(method, params, callback)
      table.insert(server.messages, {
        method = method,
        params = params,
      })
      if handler then
        handler(method, params, callback)
      elseif method == 'initialize' then
        callback(nil, {
          capabilities = opts.capabilities or {},
        })
      elseif method == 'shutdown' then
        callback(nil, nil)
      end
      local request_id = #server.messages
      return true, request_id
    end

    function srv.notify(method, params)
      table.insert(server.messages, {
        method = method,
        params = params,
      })
      if method == 'exit' then
        dispatchers.on_exit(0, 15)
      end
    end

    function srv.is_closing()
      return closing
    end

    function srv.terminate()
      closing = true
    end

    return srv
  end

  return server
end

local completion_result = {
  isIncomplete = false,
  items = {
    {
      label = '/**',
      insertText = '/** ${1:hello} */',
      insertTextFormat = 2,
      kind = 15,
      detail = 'Snippet with non-word label',
    },
    {
      label = 'return',
      insertText = 'return',
      insertTextFormat = 2,
      kind = 15,
      detail = 'return ...',
    },
  },
}

local server = create_server({
  capabilities = {
    completionProvider = {},
  },
  handlers = {
    ['textDocument/completion'] = function (_, _, callback)
      callback(nil, completion_result)
    end,
  },
})

vim.o.completeopt = 'menuone,noselect'
vim.fn.complete = vim.schedule_wrap(vim.fn.complete)

vim.api.nvim_create_autocmd('UIEnter', {
  callback = function ()
    vim.lsp.start({
      name = 'dummy',
      cmd = server.cmd,
      on_attach = function (client, bufnr)
        vim.lsp.completion.enable(true, client.id, bufnr, {})
        vim.bo[bufnr].indentexpr = 'GetLuaIndent()'
      end,
    })
  end
})
```

</details>

glepnir:
Because I reverted the changes. I know what's going on here. But it's a bit complicated and I need to think about it for a while.

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -1220,4 +1244,104 @@ describe('vim.lsp.completion: integration', function()
...
+
+    -- accept textEdit
+    feed('<C-y>')
+    eq(
+      { 'wp->handle', { 1, 10 } },
+      exec_lua(function()
+        return { vim.api.nvim_get_current_line(), vim.api.nvim_win_get_cursor(0) }
+      end)
+    )
+  end)
```
przepompownia, line 1346:
```suggestion
  end)
  
  
  it('accept nonword label', function()
    local completion_list = {
      isIncomplete = true,
      items = {
        {
          label = '/**',
          insertText = '/** @example */',
          insertTextFormat = 2,
          kind = 1,
        },
      },
    }
    exec_lua(function()
      vim.o.completeopt = 'menu,menuone,noinsert'
    end)
    create_server('dummy', completion_list)

    feed('i/**<C-x><C-o>')
    retry(nil, nil, function()
      eq(
        1,
        exec_lua(function()
          return vim.fn.pumvisible()
        end)
      )
    end)

    feed('<C-y>')
    eq(
      { '/** @example */', { 1, 13 } },
      exec_lua(function()
        return { vim.api.nvim_get_current_line(), vim.api.nvim_win_get_cursor(0) }
      end)
    )
  end)
```

```
./test/functional/plugin/lsp/completion_spec.lua:1376: Expected objects to be the same.
Passed in:
(table: 0x7fc856c4ee18) {
 *[1] = '/**/** @example */'
  [2] = {
    [1] = 1
    [2] = 18 } }
Expected:
(table: 0x7fc856c46090) {
 *[1] = '/** @example */'
  [2] = {
    [1] = 1
    [2] = 13 } }
```

glepnir: emm This should be irrelevant to this PR. I think you need to manually set iskeyword to infer word boundaries

przepompownia: I didn't sure if the scope of this PR covers it but wanted to avoid next human mistake so finally wrote that case. 

Extending `isk` has too much unwanted side effects - maybe we need some extended way to determine completion word boundaries (user defined callback?), where `isk` can be used as fallback.

glepnir: Maybe some local modifications can be done?  `if somefiletype and line:sub(vim.fn.col('.') - 3, vim.fn.col('.'):match('^xxx) then save iskeyword , modified iskeyword` and restore iskeyword in CompleteDone..

justinmk:
Do we have `vim.completion.trigger()` tests that check candiates with whitespace (and multilines)? E.g. `foo bar`, and `foo bar\nbaz buz\n`.

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -1220,4 +1244,104 @@ describe('vim.lsp.completion: integration', function()
...
+          },
+        },
+      },
+    }
+    exec_lua(function()
+      vim.o.completeopt = 'menu,menuone,noinsert'
+    end)
+    create_server('dummy', completion_list)
+
+    feed('iwp.<c-x><c-o>')
```
justinmk, line 1328: Would `trigger_at_pos()` also work here?

glepnir: Nope in different describe

test/functional/plugin/lsp/completion_spec.lua
glepnir: btw I think I have found a way to handle what you got..but it's C side work please wait :)

przepompownia: Hm, I lost idea what could be done both in my particular case and in general - probably I need some time to sort out that topic (offtopic here). I noticed sideways that `iskeyword` buffer scope can be too wide if could use a value specific to injected language. Maybe we need something like `iskeywordexpr` (keeping compatibility with Vim) that could be context aware. 

Thanks for precising the PR title.

glepnir: I have add a option named  `completetrigger` and `complete_startcol` in vim ...like `set ctg=k,/,*,.` k mean keyword  pattern other just normal character and `complete_startcol` will range ctg to find the startcol

przepompownia:
I get the following error when expanding any luals snippet:
```
Error executing vim.schedule lua callback: ...om-src-dev/share/nvim/runtime/lua/vim/lsp/completion.lua:586: attempt to index field 'cursor' (a nil value)
stack traceback:
        ...om-src-dev/share/nvim/runtime/lua/vim/lsp/completion.lua:586: in function 'clear_word'
        ...om-src-dev/share/nvim/runtime/lua/vim/lsp/completion.lua:618: in function 'handler'
        ...s/from-src-dev/share/nvim/runtime/lua/vim/lsp/client.lua:679: in function ''
```

przepompownia:
Hm, did you include fix in the recent big commit that merges master into this PR? Currently I cannot reproduce that bug.

glepnir:
Yes it solves the problem temporarily. But I have an autocomplete script that does it at every character. If Context.cursor is not cleaned it won't work.

```lua
    api.nvim_create_autocmd(InsertCharPre, {
      buffer = bufnr,
      group = g,
      callback = function()
        if tonumber(pumvisible()) == 1 then
          return
        end
          client,
          'server_capabilities',
          'completionProvider',
          'triggerCharacters'
        ) or {}
        if vim.v.char:match('[%w_]') and not vim.list_contains(triggerchars, vim.v.char) then
          vim.schedule(function()
            completion.trigger()
          end)
        end
      end,
      desc = 'completion on character which not exist in lsp client triggerCharacters',
    })
```

glepnir:
currently doesn't have

glepnir:
@przepompownia  so far everything seems to work. even with my scripts. could you give it a test ?

przepompownia:
I don't experience the recently reported bug. On the other side, `require('vim.trees<cursor>')` gives the correct list, but the replacement after both selection and confirmation is not correct - not sure if it's it in the scope of this PR.

image
(`completeopt=popup,menuone,noinsert,noselect,fuzzy`)

glepnir:
fixed

przepompownia:
https://github.com/user-attachments/assets/df0ff553-a05a-469e-9366-d49c36753780

przepompownia:
My quick manual test on Lua module completion shows fixed selection and text replacement after confirm - thank you!

glepnir:
Thanks for helping me test

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -1289,4 +1310,163 @@ describe('vim.lsp.completion: integration', function()
...
+    feed('<C-y>')
+    eq(
+      { 'wp->w_arg_idx', { 1, 13 } },
+      exec_lua(function()
+        return { vim.api.nvim_get_current_line(), vim.api.nvim_win_get_cursor(0) }
+      end)
+    )
+  end)
+
+  it('accept text.Edit when end_col equals the column at the end of the line.', function()
```
glepnir, line 1430: @mfussenegger  PTAL, there may be some places you need to look at like add char before item.filterText and adjust_end is for there avoid the text of tail delete by `apply_text_edits`

przepompownia:
@glepnir could you merge the current master here please?

glepnir:
I suddenly noticed this section. I’m not sure if I had simply overlooked it before 😮‍💨. After reading it and comparing it with completion.lua, I feel that our current behavior may not match what the specification describes.

>A completion item provides additional means to influence filtering and sorting. They are expressed by either creating a [CompletionItem](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem) with a insertText or with a textEdit. The two modes differ as follows:

> Completion item provides an insertText / label without a text edit: in the model the client should filter against what the user has already typed using the word boundary rules of the language (e.g. resolving the word under the cursor position). The reason for this mode is that it makes it extremely easy for a server to implement a basic completion list and get it filtered on the client.

> Completion Item with text edits: in this mode the server tells the client that it actually knows what it is doing. If you create a completion item with a text edit at the current cursor position no word guessing takes place and no automatic filtering (like with an insertText) should happen. This mode can be combined with a sort text and filter text to customize two things. If the text edit is a replace edit then the range denotes the word used for filtering. If the replace changes the text it most likely makes sense to specify a filter text to be used.


According to the spec, when a completion item provides a `textEdit`, the client should not try to guess word boundaries and should not perform automatic filtering. But in the current implementation, we always run `adjust_start_col`, iterating over all items to compute `server_start_boundary`. This leads to the original issue I encountered: when clangd sets `editsNearCursor = true`, the returned `range.start.character` for a pointer member access triggered by a dot includes the dot itself—meaning it points to the column before the cursor. As a result, fn.complete uses the computed `server_start_boundary`, and the later `compl_leader` built from `startcol + offset` becomes something like `.w`. Since the candidate word does not include a dot, the pum fails to build correctly. After that, even isIncomplete cannot proceed.

another problem: `adjust_start_col` performs a full traversal of all items . Then `lsp_complete` performs another full traversal and filtering pass. In practice, completion item lists are often large, and when multiple language servers are active, the overall time complexity becomes O(M × N). With two passes, it effectively becomes O(N²) in many real scenarios.

I rewrote part of the code to experiment with a new approach. It seems to work well so far, but the implementation is still rough. Before refining it,  need your @mfussenegger  thoughts.. If any part of my understanding is wrong, I’d appreciate it if you could point it out.

test4

justinmk:
If it passes tests and fixes bugs, without adding complexity, it's probably an improvement.

> it effectively becomes O(N²) in many real scenarios.

Is that the new code or the old code? Doesn't sound attractive.

glepnir:
> Is that the new code or the old code? Doesn't sound attractive.

old code runs one full loop in adjust_start_col, and then lsp_to_complete performs another full loop. That’s the worst-case scenario. The main issue here is the word boundary and startcol.

mfussenegger:
> According to the spec, when a completion item provides a textEdit, the client should not try to guess word boundaries and should not perform automatic filtering. But in the current implementation, we always run adjust_start_col, iterating over all items to compute server_start_boundary. This leads to the original issue I encountered: when clangd sets editsNearCursor = true, the returned range.start.character for a pointer member access triggered by a dot includes the dot itself—meaning it points to the column before the cursor. As a result, fn.complete uses the computed server_start_boundary,


Not sure I understand where you see the discrepancy. The `server_start_boundary` is the word boundary based on the text edits, instead of the boundary guessed by the client. This is needed because `vim.fn.complete` must have a starting point to which it will base it's filtering on when typing additional characters once the popup menu is open. This seems in line with what the spec is dictating.

The filtering aspect should also be in line with the spec:

https://github.com/neovim/neovim/blob/3e3624db64778b33d868be32ca64e8b67faf3da8/runtime/lua/vim/lsp/completion.lua?plain=1#L304-L307

In that we don't pre-filter based on a prefix. Typing characters will additionally filter based on the completeopt settings.


>  In practice, completion item lists are often large, and when multiple language servers are active, the overall time complexity becomes O(M × N). With two passes, it effectively becomes O(N²) in many real scenarios.


It's not O(N²), but currently the items are processed 4 times: Once to apply defaults, once to get the word boundary, once to filter and convert to complete-item, once to sort.

Could probably optimize that a bit if profiling shows it is a problem. I'd think that for most servers the number of completion items is actually rather small. tailwind probably being an exception with it's thousands of css classes.


---

But one issue the current implementation does have is if a server were to provide several textEdits each with a different starting word boundary. `vim.fn.complete` needs the same starting point for all items. We could address that by normalizing (trimming or extending) the completion items relative to the left (or right?)-most starting point.
I hadn't seen any server which returns entries with mixed starting locations when I created lsp-compl (on which this code was based on). So I neglected that case.

But note that to fix that, we'd have to keep iterating twice, because you need to figure out the left/right most position to be able to normalize the other entries. That requires 2 iterations.

glepnir:
> the server_start_boundary is the word boundary based on the text edits, instead of the boundary guessed by the client. This is needed because vim.fn.complete must have a starting point to which it will base it's filtering on when typing additional characters once the popup menu is open. This seems in line with what the spec is dictating.

doc says that when a textEdit is provided, the server already knows what it is doing, so boundary guessing is unnecessary right .. Also, the startcol returned from fn.complete only indicates where the popup menu should appear. The text used for filtering can be obtained by storing this server_start_boundary and filtering based on it, instead of modifying startcol. Changing startcol causes the C-side compl_leader to include the trigger character. 🤔

If I understand correctly, when a textEdit is present we don’t need boundary guessing, so there’s no need to run adjust_start_col every time...but this is not main point. It depends on the previous.

mfussenegger:
> Also, the startcol returned from fn.complete only indicates where the popup menu should appear. The text used for filtering can be obtained by storing this server_start_boundary and filtering based on it, instead of modifying startcol. Changing startcol causes the C-side compl_leader to include the trigger character.

My understanding is that `vim.fn.complete` doesn't use it only for deciding where to show the popup but also for subsequent filtering once you type another character. And one limitation here is that `completion-items` only has a `word` that controls both, what text to insert, and the filtering. On the LSP side it can be split into filterText and textEdit.
To avoid that we could set `equal=1` in the completion items, but then we'd have to do manual filtering on each new character. 

I'll try to have a closer look at the `editsNearCursor = true` case in the coming days. I hope I'll then be able to give better inputs on how to deal with it.

glepnir:
Ok.. compl_leader starts from start_col.  https://github.com/yegappan/lsp/blob/main/autoload/lsp/completion.vim seems like there also  don't change start_col.  As the docs explain, when a textEdit is present, the server already knows the context, so there’s no need to guess word boundaries.  I feel like this part of the implementation is kinda off and feels weird.

justinmk:
this surely requires a docs update? marking as draft

glepnir:
> this surely requires a docs update? marking as draft

why ?  It is just a bug fix, no new API or user-facing options added.

test/functional/plugin/lsp/completion_spec.lua
```diff
@@ -1547,6 +1549,144 @@ describe('vim.lsp.completion: integration', function()
...
+            },
+          },
+        },
+      },
+    }
+    exec_lua(function()
+      vim.o.completeopt = 'menu,menuone,noinsert'
+    end)
+    create_server('dummy', completion_list, {
+      forIncomplete = {
```
justinmk: this `forIncomplete` seems like a hack, can it be avoided

runtime/lua/vim/lsp/completion.lua
```diff
@@ -843,6 +867,34 @@ local function on_complete_done()
...
+    local range = completion_item.textEdit.range or completion_item.textEdit.insert
+    if range then
+      needs_text_edit = range.start.character ~= range['end'].character
+        or range.start.line ~= range['end'].line
+    end
+  end
+
+  if needs_text_edit and Context.cursor then
+    api.nvim_buf_set_text(
+      bufnr,
```
justinmk: lots of new code without any comments giving insight. 

can any of the new code be extracted to local functions?

justinmk:
This is a lot of new code and the benefit is not clear.
What is a small before/after example of how this changes behavior?

Going to assume this is unwanted unless there is review/approval from @neovim/lsp

glepnir:
> What is a small before/after example of how this changes behavior?

did you see the linked issue and test case ? 

before 
test5


after
test4


> Going to assume this is unwanted 

?

justinmk:
I know it improves some cases for completion, but looking for a small self-contained before/after example in text form that we can mention in news.txt .

PR #32148, opened 2025-01-21
Add tar.lz4 and tlz4 support

corpulentrobin:
Adding tar.lz4 and .tlz4 support for native viewing of lz4 compressed tarballs

Test results pending


corpulentrobin:
> You might wanna send this PR to vim first, they basically own the runtime files and we just copy it from them.

Thanks for the early review. I'll do that first and circle back for any docs I need to update here.

clason:
Yep. These changes _must_ go through Vim.

PR #32149, opened 2025-01-21
fix(editor): avoid scrolling substitute confirm message

### LINKED ISSUE #32147: Text of last substitution is not refreshed
### Problem

`:%s/qweqwe/asdasd/g`

Image

### Steps to reproduce

Run substitute command, answer `y` for each question. After last substitution neovim shows message `<n> substitutions on <n> lines`.
Error: the text of last substitution is not refreshed until the message is not closed.

### Expected behavior

Text should be refreshed immediately after substitute.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1609+g05435bf105

-

### Operating system/version

Ubuntu 20.04

### Terminal name/version

kitty 0.38.1, urxvt 9.22

xterm-kitty (kitty), rxvt-unicode-256color (urxvt)

### Installation

nvim.appimage

luukvbaal:
Regression from 48e2a73.
Fix #32147.

### DIFF
--- a/src/nvim/ex_cmds.c
+++ b/src/nvim/ex_cmds.c
@@ -3799,6 +3799,8 @@ static int do_sub(exarg_T *eap, const proftime_T timeout, const int cmdpreview_n
               typed = prompt_for_input(p, HLF_R, true, NULL);
               xfree(p);
 
+              msg_didout = false;                 // don't scroll up
+              gotocmdline(true);
               p_lz = save_p_lz;
               RedrawingDisabled = temp;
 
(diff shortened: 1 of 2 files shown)


zeertzjq (commented):
The number of substitutions needs to be larger than the `'report'` option to trigger the original issue. The test only uses two substitutions, and doesn't fail without the fix.

luukvbaal:
> The number of substitutions needs to be larger than the `'report'` option to trigger the original issue. The test only uses two substitutions, and doesn't fail without the fix.

Regardless of `'report'`, from `--clean`, pressing <kbd>y</kbd> in the substitute confirm prompt the first time already causes the message area to scroll. Not sure why it doesn't in the test.

PR #32150, opened 2025-01-21
feat(vim.fs.find): add option to switch to depth-first search [skip ci]

dundargoc:
It's needed to replace util.root_patterns in lspconfig.

### DIFF
--- a/runtime/lua/vim/fs.lua
+++ b/runtime/lua/vim/fs.lua
@@ -219,6 +219,10 @@ end
 --- Follow symbolic links.
 --- (default: `true`)
 --- @field follow? boolean
+---
+--- Use DFS
+--- (default: `false`)
+--- @field dfs? boolean
 
 --- Find files or directories (or other items as specified by `opts.type`) in the given path.
 ---
@@ -266,6 +270,7 @@ function M.find(names, opts)
   vim.validate('type', opts.type, 'string', true)
   vim.validate('limit', opts.limit, 'number', true)
   vim.validate('follow', opts.follow, 'boolean', true)
+  vim.validate('dfs', opts.dfs, 'boolean', true)
 
   if type(names) == 'string' then
     names = { names }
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/fs.lua
```diff
@@ -266,6 +266,7 @@ function M.find(names, opts)
   vim.validate('type', opts.type, 'string', true)
   vim.validate('limit', opts.limit, 'number', true)
   vim.validate('follow', opts.follow, 'boolean', true)
+  vim.validate('dfs', opts.dfs, 'boolean', false)
```
dundargoc, line 269: Placeholder name. Python uses the term "topdown" for this option but I kinda don't get it, so don't like it. Might go with "depth", "depth_first" or "order".

runtime/lua/vim/fs.lua
```diff
@@ -275,6 +280,24 @@ function M.find(names, opts)
   local stop = opts.stop
   local limit = opts.limit or 1
 
+  if opts.dfs then
+    local matches = {}
+    if type(names) == 'function' then
+      error("Can't use dfs with functions")
+    end
```
gpanders, line 287: What's the reason for this restriction?

runtime/lua/vim/fs.lua
```diff
@@ -275,6 +280,24 @@ function M.find(names, opts)
...
+    for _, name in ipairs(names) do
+      opts.dfs = false
+      local match = vim.fs.find(name, opts)
```
gpanders, line 290: I'm not sure implementing this in terms of recursively calling `vim.fs.find` is the way to go. I think we should implement the DFS behavior directly. Will probably need to refactor the separate DFS and BFS behavior into separate functions and then have `vim.fs.find` call the appropriate one to avoid this function getting too complicated.

runtime/lua/vim/fs.lua
dundargoc: It didn't seem to make sense, because if `names` is a function then `dfs` should not affect it? Although we could still allow it tbf.

runtime/lua/vim/fs.lua
dundargoc: Couldn't get it to work. Unsure if it'll be any easier than this, at least with this approach the two methods are completely separated which helps make it easier.

dundargoc: I will make you regret this. You will get what you wish for. Prepare

PR #32151, opened 2025-01-21
fix(tui): ensure all pending escape sequences are processed before exiting

### LINKED ISSUE #31885: neovim prints key release event characters after exiting in kitten ssh session
### Problem

 When exiting neovim in a remote kitten ssh session, sometimes the characters "3;1:3u" shows in terminal.

More information can be found at https://github.com/kovidgoyal/kitty/issues/8200

Not sure if https://github.com/neovim/neovim/pull/31727 is related?

### Steps to reproduce

1.  ssh to a remote machine using kitten ssh
2. start nvim nvim -u NONE
3. exit nvim
4. See the sequence "3;1:3u" (the key release event sequence)

### Expected behavior

No characters are printed

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1465+ga09c7a5d57

no, vim 8.2.2121

### Operating system/version

Ubuntu 22.04.3 LTS

### Terminal name/version

kitty 0.38.1 (d23adce11c) created by Kovid Goyal

xterm-kitty

### Installation

Nightly releases

### LINKED ISSUE #32769: terminal escapes output to shell if nvim is closed too quickly (race?)
### Problem

Sometimes quickly I realize I open the wrong file, and close nvim very fast. Rarely this will emit some escapes to the terminal:
```console
think:~$ nvim --clean -c 'q!'
think:~$ nvim --clean -c 'q!'
think:~$ nvim --clean -c 'q!'
^[[48;58;213;1044;1917tthink:~$ 8;58;213;1044;1917t
```

It doesn't reproduce every time with the `foot` terminal, but if repeated enough, it happens for me regardless of `TERM` value: `foot`, `xterm-256color`, etc.

### Steps to reproduce

# you may need to loop this:
nvim --clean -c 'q!'

### Expected behavior

terminal output is clean after exit

### Nvim version (nvim -v)

NVIM v0.11.0-dev-1912+g0c0352783f

I can't reproduce with vim-git from the AUR but I'm not confident on this.

### Operating system/version

linux 6.13.4

### Terminal name/version

foot version: 1.20.2 +pgo +ime +graphemes +toplevel-icon -assertions

xterm-256color

### Installation

AUR

gpanders:
Neovim disables a number of terminal modes when it exits, some of which cause the terminal to send asynchronous events to Neovim. It's possible that Neovim exits before the terminal has received and processed all of the sequences to disable these modes, causing the terminal to emit one of these asynchronous sequences after Neovim has already exited. If this happens, then the sequence is received by the user's shell (or some other program that is not Neovim).

When Neovim exits, it now emits a Device Attributes request (DA1) after disabling all of the different modes. When the terminal responds to this request we know that it has already received all of our other sequences disabling the other modes. At that point, it should not be emitting any further asynchronous sequences. This means the process of exiting Neovim is now asynchronous as well since it depends on receiving the DA1 response from the terminal.

Fixes: #31885
Fixes: #32769

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -18,6 +18,7 @@
 #include "nvim/cursor_shape.h"
 #include "nvim/event/defs.h"
 #include "nvim/event/loop.h"
+#include "nvim/event/multiqueue.h"
 #include "nvim/event/signal.h"
 #include "nvim/event/stream.h"
 #include "nvim/globals.h"
@@ -45,6 +46,10 @@
 #ifdef MSWIN
 # include "nvim/os/os_win_console.h"
 #endif
+
+// Maximum amount of time (in ms) to wait to receive a Device Attributes
+// response before exiting.
+#define EXIT_TIMEOUT_MS 1000
 
 #define OUTBUF_SIZE 0xffff
 
@@ -100,6 +105,7 @@ struct TUIData {
   bool bce;
   bool mouse_enabled;
   bool mouse_move_enabled;
+  bool mouse_enabled_save;
   bool title_enabled;
   bool sync_output;
   bool busy, is_invisible, want_invisible;
@@ -287,7 +293,8 @@ void tui_enable_extended_underline(TUIData *tui)
 static void tui_query_kitty_keyboard(TUIData *tui)
   FUNC_ATTR_NONNULL_ALL
 {
-  tui->input.waiting_for_kkp_response = true;
+  // Set the key encoding whenever the Device Attributes (DA1) response is received.
+  tui->input.callbacks.primary_device_attr = tui_set_key_encoding;
   out(tui, S_LEN("\x1b[?u\x1b[c"));
 }
 
@@ -514,8 +521,8 @@ static void terminfo_start(TUIData *tui)
   xfree(term_program_version_env);
 }
 
-/// Disable the alternate screen and prepare for the TUI to close.
-static void terminfo_stop(TUIData *tui)
+/// Disable various terminal modes and other features.
+static void terminfo_disable(TUIData *tui)
 {
   // Disable theme update notifications. We do this first to avoid getting any
   // more notifications after we reset the cursor and any color palette changes.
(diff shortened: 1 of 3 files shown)


src/nvim/tui/tui.c
```diff
@@ -287,10 +288,30 @@ void tui_enable_extended_underline(TUIData *tui)
...
+  }
+
+  // Send a DA1 request after disabling key events. When we receive the
+  // response, we know there are no more key events to receive.
+  tui->input.callbacks.primary_device_attr = tui_reset_kitty_keyboard;
+  out(tui, S_LEN("\x1b[<u\x1b[c"));
+
+  // Immediately flush the buffer and wait for the DA1 response.
+  flush_buf(tui);
+  LOOP_PROCESS_EVENTS_UNTIL(tui->loop, NULL, 500, tui->input.callbacks.primary_device_attr == NULL);
```
gpanders, line 312: This doesn't work when suspending (Ctrl-Z) because the main event loop runs recursively.

AThePeanut4:
Can confirm this fixes #31885 for me, when exiting with `<C-w><C-q>` over ssh.

Suspending with `CTRL-Z` crashes with a SIGABRT.

gpanders:
Not 100% working yet but it's getting closer. There's still an issue with the cursor position when suspending, and I also still need to implement a timeout for the DA1 response (so that if the terminal never responds for some reason, Neovim doesn't hang trying to exit).

justinmk (approved):

src/nvim/tui/tui.c
```diff
@@ -592,37 +612,57 @@ static void tui_terminal_after_startup(TUIData *tui)
...
 
-void tui_error_exit(TUIData *tui, Integer status)
-{
-  tui->seen_error_exit = (int)status;
-}
+  tui->input.callbacks.primary_device_attr = tui_stop_cb;
+  terminfo_disable(tui);
+
+  // Wait until DA1 response is received
+  LOOP_PROCESS_EVENTS_UNTIL(tui->loop, tui->loop->events, EXIT_TIMEOUT_MS, tui->stopped);
```
justinmk, line 635: Maybe as a followup, if we continue to get reports, we could print something like this if the timeout is reached:

    Nvim: timeout waiting for DA1 response

justinmk:
Since this change, CI now always shows these log messages:

```
$NVIM_LOG_FILE: /home/runner/work/neovim/neovim/build/.nvimlog
(last 100 lines)
ERR 2025-04-22T15:20:50.546 ui/c/T5408.22274.0 tui_stop:626: TUI already stopped (race?)
ERR 2025-04-22T15:20:50.736 ui/c/T5409.22280.0 tui_stop:626: TUI already stopped (race?)
```

gpanders:
@justinmk Tracked here https://github.com/neovim/neovim/issues/33599

PR #32153, opened 2025-01-22
fix(search): avoid quadratic time complexity when computing fuzzy score

zeertzjq:


zeertzjq:
cc @seandewar

seandewar (approved):
Oops! Good spot :see_no_evil:

Issue #32162, opened 2025-01-22
More structured health checks


clason:
### Problem

Neovim provides a `:checkhealth` feature for plugins and Neovim itself to show "self-diagnostics". This allows users to
1. quickly see detectable problems and misconfigurations ("runtime is missing", "parsers are broken", "server not executable", "TERM misconfigured")
2. information about the internal state (loaded treesitter parsers, active and inactive LSP configurations)

While this information is very useful for bug reports and understanding why your Neovim behaves as it does, the increasing amount of information (especially around (nvim-)treesitter and LSP) makes it increasingly hard to navigate. So we need to add ways of structuring and navigating this information.

### Expected behavior

There are several (independent) improvements that can be made, roughly in order of ascending complexity:

* [ ] Add a table of contents (`gO`, already supported for `help` buffers, which `checkhealth` inherits from).
* [ ] Add section navigation via `]]` (which would also be useful for `help` buffers)
* [ ] Add color to section headers (white/green for "all clear" or info only, orange if the section contains a warning, red if the section contains an error)
* [ ] Add folding to sections (foldtext should be section header, colored as above). This is a bit controversial but a net win if
   - `foldenable` is respected (so people can opt out of all folding);
   - sections are either all folded or not folded by default;
   - folding is not enabled if only a single section is shown (e.g., `:check lsp`).
* [ ]  Only show the most important, actionable, health checks by default and reserve the full information for an "extended view" that can be toggled by either pressing, say, `D` in a `checkhealth` buffer or calling `:checkhealth!`.
(This could be implemented by adding a flag to `vim.health.info` to indicate that the message should only be shown in extended mode. Bonus points for evaluating this lazily -- "Treesitter parsers", I'm thinking of you!)

yochem:
Maybe have `:checkhealth!` only show warnings and errors (i.e. `vim.health.warning` and `vim.health.error`)? Or is that what you meant with adding a flag to `vim.health.info?

bonderado:
It would be preferable to mark warnings and errors also with some formatting if available (bold, italic, underline etc): there are more color blind persons that one would expect, as I learned when my university professor could not understand a thing in my color coded presentation.

lithammer:
Just throwing it out there. But would it make sense to have something like `:verbose checkhealth`? Not sure exactly how they would differ though.

VanshajSaxena:
`:checkhealth` currently freezes the UI. Would it be possible to make it asynchronous to improve responsiveness?

[...]

PR #32165, opened 2025-01-22
feat(checkhealth): enable gO to show table of contents [skip ci]

dundargoc:
Work on https://github.com/neovim/neovim/issues/32162.

### DIFF
--- a/runtime/ftplugin/checkhealth.vim
+++ b/runtime/ftplugin/checkhealth.vim
@@ -15,3 +15,9 @@ if exists("b:undo_ftplugin")
 else
   let b:undo_ftplugin = "setl wrap< bri< lbr< kp< isk< list<"
 endif
+
+lua << EOF
+vim.keymap.set('n', 'gO', function()
+  require('vim.vimhelp').show_toc()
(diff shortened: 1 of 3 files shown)


justinmk (approved):
Just for info: it's likely that checkhealth will use ft=markdown instead of ft=help in the future. Doesn't block small improvements like this, just something to consider before big changes.

clason:
If you want to do this "by the (Vim) book", you could move the keymap before the undo dance and add a `keymap.del` in there to clean up.

(Also, why is this file not in Lua 😠 )

siddhantdev:
> Just for info: it's likely that checkhealth will use ft=markdown instead of ft=help in the future. Doesn't block small improvements like this, just something to consider before big changes.

Maybe it would be nice to allow the `show_toc()` method to take in a treesitter query, so that a TOC could be built for arbritary filetypes by passing a query to the method? Then making a toc for markdown files for example would be just writing a treesitter query.

dundargoc:
Superseded by https://github.com/neovim/neovim/pull/32282

PR #32167, opened 2025-01-22
fix(messages): avoid empty msg_showmode with 'noshowmode'

### LINKED ISSUE #32164: vim.ui_attach: `msg_showmode` data goes to its callback also with `noshowmode`
### Problem

If I'm correct `msg_showmode` should not appear in the callback of `vim.ui_attach` if `showmode` is disabled. 

### Steps to reproduce

init.lua:
```lua
local api = vim.api
vim.o.showmode = false
local ns = api.nvim_create_namespace('messages')
local count = 0

  if event ~= 'msg_showmode' then return end
  count = count + 1

  vim.schedule(function ()
    api.nvim_buf_set_lines(1, 0, 1, true, {tostring(count)})
  end)
end)
```
- `nvim --clean -u init.lua`
- press `:<Esc>` and see the counter value

### Expected behavior

If `noshowmode` is set, `ui_attach` callback isn't called with `msg_showmode` related data.

### Nvim version (nvim -v)

v0.11.0-dev-1617+gaf0ef2ac9d

n/a

### Operating system/version

Debian Sid

### Terminal name/version

alacritty

alacritty

### Installation

from repo

luukvbaal:
Fix #32164

### DIFF
--- a/src/nvim/message.c
+++ b/src/nvim/message.c
@@ -3183,7 +3183,11 @@ void msg_ext_flush_showmode(void)
 {
   // Showmode messages doesn't interrupt normal message flow, so we use
   // separate event. Still reuse the same chunking logic, for simplicity.
-  if (ui_has(kUIMessages)) {
+  // This is called unconditionally; check if we are emitting, or have
+  // emitted non-empty "content".
+  static bool clear = false;
+  if (ui_has(kUIMessages) && (msg_ext_last_attr != -1 || clear)) {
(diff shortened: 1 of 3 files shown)


luukvbaal:
> does not appear also when vim.o.showmode = true - not sure if intentionally.

Intentional; the PR avoids previously redundant `msg_showmode` events, also seen in the changed `messages_spec.lua` test. `:<Esc>` does not print a mode message without `ext_messages` and should not emit a showmode event.

PR #32174, opened 2025-01-23
build(deps): bump luv to v1.50.0-1

clason:
https://github.com/luvit/luv/releases/tag/v1.50.0-0
https://github.com/luvit/luv/releases/tag/v1.50.0-1

### DIFF
--- a/cmake.deps/deps.txt
+++ b/cmake.deps/deps.txt
@@ -10,8 +10,8 @@ LUA_SHA256 2640fc56a795f29d28ef15e13c34a47e223960b0240e8cb0a82d9b0738695333
 UNIBILIUM_URL https://github.com/neovim/unibilium/archive/v2.1.2.tar.gz
 UNIBILIUM_SHA256 370ecb07fbbc20d91d1b350c55f1c806b06bf86797e164081ccc977fc9b3af7a
 
-LUV_URL https://github.com/luvit/luv/releases/download/1.48.0-2/luv-1.48.0-2.tar.gz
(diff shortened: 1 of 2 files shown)


cmake.deps/deps.txt
```diff
@@ -10,8 +10,8 @@ LUA_SHA256 2640fc56a795f29d28ef15e13c34a47e223960b0240e8cb0a82d9b0738695333
 UNIBILIUM_URL https://github.com/neovim/unibilium/archive/v2.1.2.tar.gz
 UNIBILIUM_SHA256 370ecb07fbbc20d91d1b350c55f1c806b06bf86797e164081ccc977fc9b3af7a
 
-LUV_URL https://github.com/luvit/luv/releases/download/1.48.0-2/luv-1.48.0-2.tar.gz
-LUV_SHA256 2c3a1ddfebb4f6550293a40ee789f7122e97647eede51511f57203de48c03b7a
+LUV_URL https://github.com/luvit/luv/archive/7aeed6f089e85c3e81aa99a391ee9d48db266e21.tar.gz
```
clason, line 13: @dundargoc I used `bump_deps.lua luv --ref v0.15.0-0` here -- any idea why it put the commit hash instead of the version number here?

dundargoc: v1.15.0-0 doesn't exist. It doesn't put the commit hash for me but it also doesn't throw an error that the ref doesn't exist, which it should.

clason: Yeah, I always forget whether it's that or `1.15.0-0`. Probably the reason we had the weird `version` code in the original script.

dundargoc: https://github.com/neovim/neovim/pull/32193

Issue #31925, opened 2025-01-09
Crash on tab switching autocmd after refactor on options


przepompownia:
### Problem

With `au FileType help wincmd T` I get crash on `:help`.

```cpp
#0  use_tabpage (tp=0x0) at .../src/nvim/window.c:4028
4028      topframe = curtab->tp_topframe;
(gdb) bt
#0  use_tabpage (tp=0x0) at .../src/nvim/window.c:4028
#1  aucmd_prepbuf_win (aco=0x7fff285a1f70, buf=0x5562554ceef0, win=0x5562554cc450) at .../src/nvim/autocmd.c:1312
#2  0x000055623d64d7d9 in apply_optionset_autocmd (errmsg=0x0, opt_idx=676994928, opt_flags=1033903008, oldval=..., oldval_g=..., oldval_l=..., newval=..., ctx=...) at .../src/nvim/option.c:1883
#3  set_option_for (opt_idx=opt_idx@entry=kOptFiletype, value=..., ctx=..., opt_flags=opt_flags@entry=2, set_sid=set_sid@entry=0, direct=direct@entry=false, value_replaced=<optimized out>, errbuf=<optimized out>, errbuflen=<optimized out>)
    at .../src/nvim/option.c:3787
#4  0x000055623d64e99e in set_option_value_for (opt_idx=kOptFiletype, value=..., ctx=..., opt_flags=2) at .../src/nvim/option.c:3862
#5  0x000055623d64eb20 in set_option_value (opt_idx=<optimized out>, value=..., opt_flags=<optimized out>) at .../src/nvim/option.c:3874
#6  set_option_value_give_err (opt_idx=<optimized out>, value=..., opt_flags=<optimized out>) at .../src/nvim/option.c:3925
#7  0x000055623d54a38d in ex_setfiletype (eap=0x7fff285a2990) at .../src/nvim/ex_docmd.c:7635
#8  ex_setfiletype (eap=0x7fff285a2990) at .../src/nvim/ex_docmd.c:7624
#9  0x000055623d78c3d5 in execute_cmd0.constprop.0 (retv=retv@entry=0x7fff285a2264, eap=eap@entry=0x7fff285a2990, errormsg=errormsg@entry=0x7fff285a2268, preview=preview@entry=false)
    at .../src/nvim/ex_docmd.c:1711
[...]

Issue #31923, opened 2025-01-09
Cursor Jumps to 0:0 after first entering insert mode.


m4r1vs:
### Problem

After the second most recent commit #c4b658fe (https://github.com/neovim/neovim/pull/31112), my cursor jumps to the very top after opening a file and inserting a character.

Not reproducible without any plugins but reproducible with only nvim-cmp (https://github.com/hrsh7th/nvim-cmp).

### Steps to reproduce

1. Have nvim-cmp (https://github.com/hrsh7th/nvim-cmp) installed with minimal config (only buffer completion).
2. Checkout commit  #c4b658fe or later.
3. Build neovim.
4. Run `nvim foo.bar`
5. Enter insert-mode in a spot different than 1:1 and insert a character.
6. notice the cursor jumping to 0:0 immediately.

### Expected behavior

I expect the cursor to not jump randomly.

### Nvim version (nvim -v)

NVIM v0.11.0-dev-c4b658fed

### Vim (not Nvim) behaves the same?

no

### Operating system/version

Linux 6.12.8-arch1-1

### Terminal name/version

alacritty and ghostty, does not seem to matter

### $TERM environment variable

xterm-256color

### Installation

Initially nvim-git from AUR and now reproduced from source manually.

m4r1vs:
specifically introduced by https://github.com/neovim/neovim/pull/31112/commits/6257270040bc5c61a489f7fb9d4102223c36cf89

PR #32179, opened 2025-01-23
[WIP] refactor(options): set option value for non-current context directly

famiu:
Problem: Currently, we use switch_option_context to temporarily switch the current option context before setting an option for a different buffer / window. This is not ideal because we already support getting and setting option values for non-current contexts in the underlying implementation.

Solution: Set option value for non-current context by passing the context directly to the lower level functions. Also introduce a new OptCtx struct to store option context information, this will scale much better if we add more option scopes and other context information in the future.

Second attempt at #31112 which was reverted by #31924 due to #31923 and #31925


justinmk:
Not sure this is feasible. Also need to profile performance before going in this direction.

Meanwhile, context-switching is now owned by `ctx_switch`.

PR #32182, opened 2025-01-23
fix(column): apply custom highlight to last 'statuscolumn' segment

luukvbaal:


gpanders (approved):

test/functional/ui/statuscolumn_spec.lua
```diff
@@ -227,6 +227,14 @@ describe('statuscolumn', function()
       {1: }{8:8│}aaaaa                                             |
                                                            |
     ]])
+    -- Last segment and fillchar is highlighted properly
```
zeertzjq, line 230:
```suggestion
    -- Last segment and fillchar are highlighted properly
```

PR #32183, opened 2025-01-23
replace deprecated:ts_node_child_containing_descendent with ts_node_child_with_descendent

voidstar-null:
https://github.com/tree-sitter/tree-sitter/commit/344a88c4fb968cc789049a378e0cc988c4253751 removed the `ts_node_child_containing_descendent` method from `<tree_sitter/api.h>` in favor of `ts_node_child_with_descendent`.

This was causing build errors on Gentoo, so just replacing this with the alternative worked.
I'm sorry if the PR's lacking anything, I"m new to this thing.


the-fuckin-nobody:
Just had to make some changes to the ebuild, no pain at all.

PR #26921, opened 2024-01-06
fix(colorscheme): default statusline groups usability

echasnovski:
Problem: Current values of `StatusLine` and `StatusLineNC` are currently
  designed to be visually distinctive while being not intrusive.
  However, the compromise was more shifted towards "not intrusive".
  After the feedback, statusline highlight groups should be designed to:
  - Make current window clearly noticeable. Meaning `StatusLine` and `StatusLineNC` should obviously differ.
  - Make non-current windows clearly separable. Meaning `StatusLineNC` and `Normal`/`NormalNC` should obviously differ.

Solution:
  - Update `StatusLineNC` to have more visible background.
  - Update `StatusLine` to be inverted variant of `StatusLineNC`.
  - Update `WinBar` and `WinBarNC` to not link to `StatusLine` and `StatusLineNC` because it makes two goals harder to achieve.
  - Update `TabLine` to link to `StatusLineNC` instead of `StatusLine` to not be very visually intrusive.

------

Screenshots:

<details><summary>Many windows with tabline and winbar</summary>

screenshot_2024-01-06_17:43:27
screenshot_2024-01-06_17:43:41

</details>

<details><summary>Single window with tabline and winbar</summary>

screenshot_2024-01-06_17:43:51
screenshot_2024-01-06_17:44:00

</details>

This is a follow-up to [this comment](https://github.com/neovim/neovim/pull/26540#issuecomment-1874465785). Some notes:

- The `NvimLightGrey3` and `NvimDarkGrey3` are chosen again to strike a balance between not very intrusive while clearly distinguishable. The contrast ratio is ~7.96 which is a AAA type of accessibility.
- @justinmk, I really hope that using different color for `StatusLineNC` is enough for making windows separable. Using underlined text makes not much distinction to `Normal` and it would still need some sort of foreground/background adjustment if clear window separation is a goal. Besides, it is less likely to have a good rendering across terminals than different background color.

PR #32186, opened 2025-01-23
fix(highlight): `Folded` and `CursorLine` hl for default color scheme

### LINKED ISSUE #30612: StatusLineNC, Folded, and CursorLine highlights are indistinguishable in default colorscheme
### Problem

This can make the UI confusing, and difficult to tell what is what at a glance, especially if you have a lot of windows open and/or folded lines.

### Steps to reproduce

Here's an example with the following highlights being displayed right next to each other

1. StatusLineNC
2. Folded
3. CursorLine

image

### Expected behavior

Here is the same example where the highlights are different making it easier/quicker to tell what is what at a glance:

image

### Nvim version (nvim -v)

NVIM v0.11.0-dev-4005+g69553f7bf5-Homebrew

no, vim 9.1

### Operating system/version

macOS 14.5

### Terminal name/version

kitty 0.36.4

xterm-kitty

### Installation

homebrew

lsvmello:
Resolves #30612

Changed `Folded` and `CursorLine` to a differents shade of gray so they don't "blend" with `StatusLineNC`.

image


echasnovski:
Thanks for the PR!

All chosen shades of grey for these groups was a result of a deliberate compromise based on the following arguments:
- Those are UI elements and have to have backgrounds as shades of grey.
- There are only four shades of grey available for background (in each dark and light variants):
    - Grey1 is used for floating windows and is deliberately more extreme than background (darker for dark theme and lighter for light theme). Setting it as background to either `Folded`, `CursorColumn`, or something else will make it invisible in floating windows.
    - Grey2 is used for Normal. Setting it as background will make it invisible in normal windows.
    - Grey3 is used for all other common UI elements: `CursorLine`, `CursorColumn`, `Folded`. Currently it is also used for `StatusLineNC`, but initially it was using Grey1. One of the reasons for this was *this specific issue* of not being the same to `CursorLine`, but it was deemed as bad usability and so Grey3 was chosen as the closest shade of Grey. See #26921.
    - Grey4 is mostly reserved for `Visual` as it being different from `CursorLine` is pretty crucial for usability: it has to be more noticeable for easier visual selection. Besides, having `CursorLine` be lighter (in dark theme) will hurt accessibility via lower contrast ratio.

So with this in mind:
- Changing `CursorLine` to Grey1 will make it invisible in floating windows. Navigating inside floating windows is most certainly a more common scenario than cursor being near statusline in inactive normal window and *probably* more common than it being near the fold.
- Changing `Folded` to Grey4 is possible but I personally think that it makes folds be too visible, as Grey4 was *designed* to be more visible.

All in all, I think it's better to close this PR. I'll also add this to #30612 with the intention of closing that one too.

justinmk:
> * Grey4 is mostly reserved for `Visual` as it being different from `CursorLine` is pretty crucial for usability: it has to be more noticeable for easier visual selection. Besides, having `CursorLine` be lighter (in dark theme) will hurt accessibility via lower contrast ratio.

Would it help if we switched `Visual` to `gui=reverse`[1] ? I think the current `Visual` could use much more contrast. And it sounds like that would also allow the use of Grey4 for something else.

[1] I know I objected to "reverse" initially, but at this point I'd prefer it to the low-contrast `Visual`. Actually I would prefer `Visual guifg=bg guibg=fg` but I assume that's too controversial.

echasnovski:
> Would it help if we switched `Visual` to `gui=reverse`[1] ? I think the current `Visual` could use much more contrast. And it sounds like that would also allow the use of Grey4 for something else.

I don't think it would be much help. Mostly because the whole grey palette [was designed](https://github.com/neovim/neovim/blob/851137f67905f6038e51b5b7d1490fbedea4faaa/runtime/doc/dev_theme.txt#L45-L47) around the concept of four lightness values for background with each level reserved for `NormalFloat`, `Normal`, `CursorLine`, and `Visual`. It basically means that Grey4 is designed to draw (mostly temporary) attention through background but still have reasonable contrast ratio with all foreground colors (4 shades of grey and six syntax colors). It is already used as background to serve that purpose where it needs to: `Visual`, `ColorColumn`, `DiffChange`, `MatchParen`. The `Folded` and `StatusLineNC` groups don't feel like they need to draw attention, but on the contrary need to indicate their presence without drawing attention (which is what Grey3 is designed for).

Although the root issue from #30612 is unfortunate, there are other visual clues to distinguish lines:
- Fold has special 'foldtext' to indicate fold, which usually and by default makes it clear that that is fold.
- Inactive statusline when next to the cursor line in active window is always the one at the top.

> [1] I know I objected to "reverse" initially, but at this point I'd prefer it to the low-contrast `Visual`. Actually I would prefer `Visual guifg=bg guibg=fg` but I assume that's too controversial.

If by "low-contrast" you mean contrast ratio between `Visual` and `Normal` background, then it is a respectable 2.3. Many most popular Neovim color schemes have significantly lower score. With this contrast even text is fairly readable, as can be seen with non-current Line numbers (as [`LineNr` uses DarkGrey4 for foreground](https://github.com/neovim/neovim/blob/851137f67905f6038e51b5b7d1490fbedea4faaa/src/nvim/highlight_group.c#L447) with `Normal`'s DarkGrey2 as background). If you mean "low-contrast" as in "text has low contrast inside Visual selection", then it is a very good 6.1-6.3 for foreground colors and not-so-good 2.9 for comments (which should be okay as visual selection is temporary).

Using background-only `Visual` highlight group was one of the main constraints during default color scheme building, because users *mostly* prefer and expect having syntax highlighting present during visual selection, as this is how it is done in other editors. So using `guifg=bg guibg=fg` (or more realistically `guifg=NvimDarkGrey2 guibg=NvimLightGrey2` for dark background) doesn't achieve that, but indeed delivers better contrast ratio.

Using "reverse" half-achieves syntax color preservation, but in a very attention taxing way. I personally prefer "reverse in selection" in terminals when there are only 16 colors available, but not if more colors are available. And I don't think using "reverse" approach is what's more commonly expected for selection with 'termguicolors'.

------

The default color scheme was initially designed to be minimal, usable, and "Neovim branded". Being minimal mostly means using as few colors as possible (both in palette and its application to highlight groups) but being usable mostly means have as much visual feedback as possible. Striking a balance between the two so that on average a color scheme is a good *default* is hard, but I think the current design and palette fits it pretty decently. It can't be perfect in every way, unfortunately.

It also means that for each user it is likely that there will be *something* off, in which case the usual suggestion of "configure it the way you like it" feels appropriate here.

echasnovski:
@lsvmello, after rethinking, it was fixed in #32256 with a slightly different mix of shades of grey. Thanks anyways for the effort!

PR #32187, opened 2025-01-23
feat(diagnostic): add `current_line` option for `virtual_text` handler

MariaSolOs:
As suggested by @gpanders in https://github.com/neovim/neovim/pull/31959#discussion_r1927069813.

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -190,6 +190,10 @@ end
 --- severity |diagnostic-severity|
 --- @field severity? vim.diagnostic.SeverityFilter
 ---
+--- Only show diagnostics for the current line.
+--- (default `false`)
+--- @field current_line? boolean
+---
 --- Include the diagnostic source in virtual text. Use `'if_many'` to only
 --- show sources if there is more than one diagnostic source in the buffer.
 --- Otherwise, any truthy value means to always show the diagnostic source.
@@ -628,6 +632,26 @@ local function diagnostic_lines(diagnostics)
     table.insert(line_diagnostics, diagnostic)
   end
   return diagnostics_by_line
+end
+
+--- @param diagnostics table<integer, vim.Diagnostic[]>
+--- @return vim.Diagnostic[]
+local function diagnostics_at_cursor(diagnostics)
+  local lnum = api.nvim_win_get_cursor(0)[1] - 1
+
+  if diagnostics[lnum] ~= nil then
+    return diagnostics[lnum]
+  end
+
+  local cursor_diagnostics = {}
+  for _, line_diags in pairs(diagnostics) do
+    for _, diag in ipairs(line_diags) do
+      if diag.end_lnum and lnum >= diag.lnum and lnum <= diag.end_lnum then
+        table.insert(cursor_diagnostics, diag)
+      end
+    end
+  end
+  return cursor_diagnostics
 end
 
 --- @param namespace integer
@@ -1570,6 +1594,28 @@ M.handlers.underline = {
   end,
 }
 
+--- @param namespace integer
+--- @param bufnr integer
+--- @param diagnostics table<integer, vim.Diagnostic[]>
+--- @param opts vim.diagnostic.Opts.VirtualText
+local function render_virtual_text(namespace, bufnr, diagnostics, opts)
+  api.nvim_buf_clear_namespace(bufnr, namespace, 0, -1)
+
+  for line, line_diagnostics in pairs(diagnostics) do
+    local virt_texts = M._get_virt_text_chunks(line_diagnostics, opts)
+
+    if virt_texts then
+      api.nvim_buf_set_extmark(bufnr, namespace, line, 0, {
+        hl_mode = opts.hl_mode or 'combine',
+        virt_text = virt_texts,
+        virt_text_pos = opts.virt_text_pos,
+        virt_text_hide = opts.virt_text_hide,
+        virt_text_win_col = opts.virt_text_win_col,
+      })
+    end
+  end
+end
+
 M.handlers.virtual_text = {
   show = function(namespace, bufnr, diagnostics, opts)
     vim.validate('namespace', namespace, 'number')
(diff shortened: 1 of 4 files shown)


test/functional/lua/diagnostic_spec.lua
```diff
@@ -2117,6 +2117,25 @@ describe('vim.diagnostic', function()
         end)
       )
     end)
+
+    it('can only show virtual_text for the current line', function()
```
justinmk, line 2121:
```suggestion
    it('virtual_text.current_line shows only current line', function()
```

justinmk (approved):

gpanders (approved):

runtime/lua/vim/diagnostic.lua
```diff
@@ -1577,6 +1581,12 @@ M.handlers.virtual_text = {
 
     local virt_text_ns = ns.user_data.virt_text_ns
     local buffer_line_diagnostics = diagnostic_lines(diagnostics)
+
+    if opts.virtual_text.current_line == true then
+      local lnum = api.nvim_win_get_cursor(0)[1] - 1
+      buffer_line_diagnostics = { [lnum] = buffer_line_diagnostics[lnum] }
+    end
```
gpanders, line 1588: Is it worth moving the call to `diagnostic_lines` into the `else` branch of this condition so that we don't throw away a bunch of data we don't need? In the `then` branch we'd need to compute the buffer line diagnostics "manually", but that is easy (simple for loop + table.insert).

MariaSolOs: Good idea!

justinmk (approved):

runtime/lua/vim/diagnostic.lua
```diff
@@ -1543,6 +1548,46 @@ M.handlers.underline = {
   end,
 }
 
+---@param bufnr integer
+---@param ns integer
+---@param diagnostics table<integer, vim.Diagnostic[]>
+---@param opts vim.diagnostic.Opts.VirtualText
+local function render_virtual_text(bufnr, ns, diagnostics, opts)
+  api.nvim_buf_clear_namespace(bufnr, ns, 0, -1)
```
MariaSolOs, line 1556: I'm not sure if this approach (which is the same as what I'm doing in #31959) is compatible with the diagnostic caching in this module.

MariaSolOs:
I'll move this to a draft since it has quite a bit of overlap with #31959 so I'll just wait for that to get merged and refactor common logic.

runtime/lua/vim/diagnostic.lua
```diff
@@ -1592,23 +1643,43 @@ M.handlers.virtual_text = {
...
+      api.nvim_create_autocmd('CursorMoved', {
+        buffer = bufnr,
+        group = M.handlers.virtual_text._augroup,
+        callback = function()
+          render_virtual_text_at_current_line(
+            bufnr,
+            ns.user_data.virt_text_ns,
+            diagnostics,
+            opts.virtual_text
+          )
```
gpanders, line 1663: If I'm understanding how this code works, then we are iterating through the list of diagnostics to find the diagnostics for the current cursor line every time the cursor moves, even when the list of diagnostics hasn't changed and the current cursor line hasn't changed.

It would increase the complexity of the handler a bit, but we could cache the result of `diagnostic_lines` whenever `show` is called, and then the CursorMoved autocommand can index into that table based on the current cursor line (so we get the cursor line diagnostics in O(1) time instead of O(N)).

The cursor moves a lot, so it's probably worth it (or at the very least only recompute diagnostics when the cursor's line number changes).

tris203:
I have been using the similar implementation in `virtual_lines` and I think there is an issue where due to the deleting and recreating of autocmds it only shows diagnostics for the current line in the last edited buffer

If you open a split of two buffers with multiple errors each and switch between them. Only the one you updated last will draw correctly, otherwise it gets 'stuck' in the last drawn state

Making a change or entering insert will kick it back to life

MariaSolOs:
> I have been using the similar implementation in `virtual_lines` and I think there is an issue where due to the deleting and recreating of autocmds it only shows diagnostics for the current line in the last edited buffer
> 
> If you open a split of two buffers with multiple errors each and switch between them. Only the one you updated last will draw correctly, otherwise it gets 'stuck' in the last drawn state
> 
> Making a change or entering insert will kick it back to life

Could you please open an issue with repro steps so that we can fix this? :)

runtime/lua/vim/diagnostic.lua
MariaSolOs: Good idea. The same improvement would also apply with `virtual_lines` so double win.

MariaSolOs: Implementing this in https://github.com/neovim/neovim/pull/32288.

MariaSolOs:
@gpanders @justinmk I've updated this PR to include the perf improvement from #32288 and the fix from @tris203 (#32275).

runtime/lua/vim/diagnostic.lua
```diff
@@ -1809,22 +1885,8 @@ end
...
-          table.insert(cursor_diagnostics, diag)
-        end
-      end
-    end
-  end
-
-  render_virtual_lines(namespace, bufnr, cursor_diagnostics)
+  diagnostics = diagnostics_at_cursor(diagnostics)
+  render_virtual_lines(namespace, bufnr, diagnostics)
 end
```
gpanders, line 1890: Can we remove this function completely and use

```lua
render_virtual_lines(namespace, bufnr, diagnostics_at_cursor(diagnostics))
```

instead?

gpanders (approved):

runtime/lua/vim/diagnostic.lua
MariaSolOs: Hmm given that we use this function in 2 places and the name makes the code clearer, I prefer keeping it.

MariaSolOs: It also aligns with how we have a `render_virtual_text_at_current_line`.

The name is a bit long though, I admit. I'm open to renaming it to something less verbose.

gpanders: > Hmm given that we use this function in 2 places and the name makes the code clearer, I prefer keeping it.

I don't actually think it makes the code any clearer. Consider:

```lua
render_virtual_lines(namespace, bufnr, diagnostics_at_cursor(diagnostics))
```

vs

```lua
render_virtual_lines_at_current_line(diagnostics, namespace, bufnr)
```

If anything, I'd argue the extra function makes it less clear because there are two different functions with similar names that do almost the same thing. Having fewer functions that compose well is better than many functions with specific uses.

MariaSolOs: Okay. I don't feel strongly about this so I'll remove the `*_at_current_line` functions if it bothers you.

justinmk (approved):
needs a rebase

KarlEricsson:
If possible, it would be great to also have a setting that disables virtual_text for the current line. My use case is using the new virtual_lines feature with current_line set to true, while keeping virtual_text enabled (filtered to show errors only). Having both virtual_text and virtual_lines enabled on the current line creates a bit of visual clutter and duplication of messages. Apologies if this isn't the right forum for this feature request. Thanks for your work on this great feature!

MariaSolOs:
@KarlEricsson so if I understand what you want correctly (sorry if I'm missing something) is only showing `virtual_lines` for the current line and `virtual_text` everywhere else.

Then you can have something like:
```lua
vim.diagnostic.config {
  virtual_text = {
    format = function(diag)
      local lnum = api.nvim_win_get_cursor(0)[1] - 1
      -- Don't use the virtual text handler for diagnostics at the current line.
      if diag.lnum == lnum then return nil end
      return diag.message
    end
  },
  virtual_lines = { current_line = true }
}
```

Note that I haven't tested this, but I expect it to work.

sagg0t:
@MariaSolOs I was trying to achieve the discussed behavior today and came up with pretty much the same code snippet, however the issue with it is that it hides the virtual_text diagnostic even when cursor moves away from the line. It gets update only when document changes or other event that refreshes the diagnostic list.

However the desired behavior is rather opposite to virtual_lines diagnostic configured with current_line=true. So in this case virtual_lines appear only when you hover the line and disappear when you move away from it, and @KarlEricsson (as well as I) would like to be able to be able to hide virtual_text on the line when the cursor is placed on the line with it.

Sorry id it's too much details, it's late in my timezone and I wanted to clarify the subject as much as possible 😅
Anyway, thanks for your contribution!

KarlEricsson:
> @KarlEricsson so if I understand what you want correctly (sorry if I'm missing something) is only showing `virtual_lines` for the current line and `virtual_text` everywhere else.
> 
> Then you can have something like:
> 
> ```lua
> vim.diagnostic.config {
>   virtual_text = {
>     format = function(diag)
>       local lnum = api.nvim_win_get_cursor(0)[1] - 1
>       -- Don't use the virtual text handler for diagnostics at the current line.
>       if diag.lnum == lnum then return nil end
>       return diag.message
>     end
>   },
>   virtual_lines = { current_line = true }
> }
> ```
> 
> Note that I haven't tested this, but I expect it to work.

I tried this solution when I started experimenting with this feature but as @sagg0t mentions this only hides the virtual_text where the cursor happens to be positioned at the moment the diagnostics are updated. If I later move in the document without refreshing diagnostics, it will not hide the virtual_text if I move the cursor to another line with diagnostics. What I'm asking is as mentioned above basically an the exact inverse of what current_line true does. Ie hide the diagnostic at current line and show all other (and update on cursor movement, not on diagnostics refresh).

MariaSolOs:
@sagg0t @KarlEricsson ohhhh I think I understand the problem now. Wow okay I think that in order to achieve this behaviour we would need to call `vim.diagnostic.show` for each `CursorMoved` event, which would be quite expensive.

@gpanders I'm curious to hear if you have an idea of how to have this functionality (if possible).

khaneliman:
> > @KarlEricsson so if I understand what you want correctly (sorry if I'm missing something) is only showing `virtual_lines` for the current line and `virtual_text` everywhere else.
> > Then you can have something like:
> > ```lua
> > vim.diagnostic.config {
> >   virtual_text = {
> >     format = function(diag)
> >       local lnum = api.nvim_win_get_cursor(0)[1] - 1
> >       -- Don't use the virtual text handler for diagnostics at the current line.
> >       if diag.lnum == lnum then return nil end
> >       return diag.message
> >     end
> >   },
> >   virtual_lines = { current_line = true }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note that I haven't tested this, but I expect it to work.
> 
> I tried this solution when I started experimenting with this feature but as @sagg0t mentions this only hides the virtual_text where the cursor happens to be positioned at the moment the diagnostics are updated. If I later move in the document without refreshing diagnostics, it will not hide the virtual_text if I move the cursor to another line with diagnostics. What I'm asking is for, as mentioned above, basically the exact inverse of what current_line = true does. Ie hide the diagnostic at current line and show all other (and update on cursor movement, not on diagnostics refresh).

Dang, came hoping I'd found a solution to this use case... :p Liked the idea of saving vertical space with virtual_text and swapping to lines when cursor is on the line.

PR #32191, opened 2025-01-24
docs(support): update tested macOS and FreeBSD versions

zeertzjq:


clason:
Macos runners: "Oh no you don't!"

PR #32193, opened 2025-01-24
build(bump_deps): abort if archive doesn't exist

dundargoc:
Also use git tag archive over commit sha if possible.

### DIFF
--- a/scripts/bump_deps.lua
+++ b/scripts/bump_deps.lua
@@ -3,6 +3,7 @@
 -- Usage:
 --    ./scripts/bump_deps.lua -h
 
+assert(vim.fn.executable('gh') == 1)
 assert(vim.fn.executable('sed') == 1)
 
 local required_branch_prefix = 'bump-'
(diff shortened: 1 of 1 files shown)


scripts/bump_deps.lua
```diff
@@ -70,9 +70,8 @@ local function get_archive_info(repo, ref)
   local archive_path = temp_dir .. '/' .. archive_name
   local archive_url = 'https://github.com/' .. repo .. '/archive/' .. archive_name
 
-  vim.fs.rm(archive_path, { force = true })
   run_die(
-    { 'curl', '-sL', archive_url, '-o', archive_path },
+    { 'curl', '-sfL', archive_url, '-o', archive_path },
```
clason, line 74: Probably good to add `--show-error`, too.

(*Ceterum censeo* that exactly this flagpalooza is why a `vim.net.download()` would be valuable.)

dundargoc: Doesn't make a difference in the output. I think it's because we don't use `stderr` from `vim.system`.

dundargoc:
> But that won't fix the `--ref vX.Y.Z` ending up as a hash instead of the version ref in the `deps.txt`, right?

Should be fixed now

PR #32194, opened 2025-01-24
ci(release): add linux-arm64 appimage

### LINKED ISSUE #15143: provide AppImage release for arm / aarch64
- `nvim --version`: N/A
- Operating system/version: Debian testing, arm64

### Actual behaviour

there is no AppImage for aarch64 platforms. this means users like me who do not use x86_64 and run a distribution without the latest version in its repositories, will have to build it ourselves if we want the latest.

### Expected behaviour

an AppImage for aarch64 is provided.

clason:
Problem: No releases for ARM Linux.

Solution: Provide appimages and tarballs for `linux-arm64`. Rename x86 releases to `linux-x86_64` for consistency.

https://github.com/clason/neovim/releases/tag/nightly


.github/workflows/release.yml
```diff
@@ -39,10 +39,21 @@ jobs:
...
+          - runner: ubuntu-20.04
+            arch: x86_64
+            cc: gcc-10
+          - runner: ubuntu-24.04-arm
+            arch: aarch64
+    runs-on: ${{ matrix.runner }}
     env:
-      CC: gcc-10
+      CC: ${{ matrix.cc }}
+      LDAI_NO_APPSTREAM: 1 # skip checking (broken) AppStream metadata for issues
```
clason, line 56: Note: This is necessary to make the current `linuxdeploy` (only on Ubuntu-24.04) not bail out on our broken AppStream metadata.

.github/workflows/release.yml
```diff
@@ -39,10 +39,21 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ ubuntu-20.04, ubuntu-24.04-arm ]
+        include:
+          - runner: ubuntu-20.04
+            arch: x86_64
+            cc: gcc-10
+          - runner: ubuntu-24.04-arm
+            arch: aarch64
```
clason, line 52: using `aarch64` here, there, and everywhere since that's what the appimage builder gets from `$ARCH`.

.github/workflows/release.yml
jamessan: Why not fix our AppStream metadata?

zeertzjq:
Since `nvim-linux64` is renamed to `nvim-linux-x86_64` here, is this a breaking change?

INSTALL.md
```diff
@@ -118,24 +119,24 @@ Then add this to your shell config (`~/.bashrc`, `~/.zshrc`, ...):
 
 ### AppImage ("universal" Linux package)
 
-The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppImage](https://appimage.org) that runs on most Linux systems. No installation is needed, just download `nvim.appimage` and run it. (It might not work if your Linux distribution is more than 4 years old.)
+The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppImage](https://appimage.org) that runs on most Linux systems. No installation is needed, just download `nvim.appimage` and run it. (It might not work if your Linux distribution is more than 4 years old.) The following instructions assume an `x86_64` architecture; on ARM Linux replace with `aarch64`.
```
zeertzjq, line 122:
```suggestion
The [Releases](https://github.com/neovim/neovim/releases) page provides an [AppImage](https://appimage.org) that runs on most Linux systems. No installation is needed, just download `nvim-linux-x86_64.appimage` and run it. (It might not work if your Linux distribution is more than 4 years old.) The following instructions assume an `x86_64` architecture; on ARM Linux replace with `aarch64`.
```

.github/workflows/notes.md
```diff
@@ -34,25 +34,43 @@ Note: On Windows "Server" you may need to [install vcruntime140.dll](https://lea
...
 #### Tarball
 
-1. Download **nvim-linux64.tar.gz**
-2. Extract: `tar xzvf nvim-linux64.tar.gz`
-3. Run `./nvim-linux64/bin/nvim`
+1. Download **nvim-linux-x86_64.tar.gz**
+2. Extract: `tar xzvf nvim-linux-x86_64.tar.gz`
+3. Run `./nvim-linux-x86_64/bin/nvim`
+
+### Linux (aarch64)
```
justinmk, line 57: for the macos builds we call them `macos-arm64`, wouldn't it make sense to be consistent with how we describe "arm architecture"?

.github/workflows/release.yml
justinmk: why does it matter what appimage gets? our interface is the names we provide. it's unnecessary entropy to use different names for "arm 64" for different OSes, at the layer that we control.

.github/workflows/notes.md
clason: See above. There's a technical reason for that.

.github/workflows/release.yml
clason: Because I don't control what `$ARCH` is on the runners, and I rely on that to dispatch to the right `linuxdeploy` binary in `genappimage.sh`.

clason: And they're completely different platforms with different naming conventions, so I don't see why they need to be the same (especially since there are technical reasons _not_ to make them the same).

clason:
> Since `nvim-linux64` is renamed to `nvim-linux-x86_64` here, is this a breaking change?

Hmm, that's a good point. I should probably publish the `x86_64` under the old name as well for compatibility. Maybe we can upload but not advertise them to avoid instantly breaking scripts.

.github/workflows/release.yml
dundargoc: I think aarch and arm are the same platform if I understand correctly. The annoying part is that cmake sometimes returns `arm64` and sometimes returns `aarch64`. I think it's up to each cmake module to define the value as they see fit. Cmake might accept a PR to fix this inconsistency, but in the meantime there's a simple workaround if you want to use that:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,10 @@ if(XCODE)
   message(FATAL_ERROR [[Xcode generator is not supported. Use "Ninja" or "Unix Makefiles" instead]])
 endif()
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+  set(CMAKE_SYSTEM_PROCESSOR arm64)
+endif()
+
 # Point CMake at any custom modules we may ship
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
```

This will always set it to `arm64`.

clason: That doesn't fix it for `$ARCH`, though, which is still `aarch64`.

dundargoc:
Continued in https://github.com/neovim/neovim/pull/32207

.github/workflows/release.yml
clason: (Also, scripts can use `$ARCH` to download the right appimage if we stick to the standard.)

justinmk: > scripts can use `$ARCH` to download the right appimage if we stick to the standard.)

That's a good reason, if scripts are doing that. Do you have an example? The example I've seen is our own node-client which points to various architectures, and it sure looks silly that "x86" and "arm" are spelled different ways for different OSes: https://github.com/neovim/node-client/blob/aa8fff0b7e486caf66a95befeed5805ae7e458f1/.github/workflows/ci.yml#L33-L43

jamessan: > Any scripts that download all the release archives for whatever reason have to treat each name specially, they can't "calculate" the name using simple logic.

They already can't do that, because the name people will think of depends on the context they're coming from. [Arch Linux ARM](https://archlinuxarm.org/) calls it aarch64. Debian's official name for the [port](https://release.debian.org/testing/arch_qualify.html) is arm64 but the [install guide](https://www.debian.org/releases/stable/installmanual) mentions aarch64.

At least it's only 2 variants, unlike the [4 variants](https://en.wikipedia.org/wiki/X86-64) that get used for x86_64 (and I recall being debated when it was introduced to Debian).

justinmk: > Debian's official name for the [port](https://release.debian.org/testing/arch_qualify.html) is arm64 but the [install guide](https://www.debian.org/releases/stable/installmanual) mentions aarch64.

So Debian agrees with me :) 

On the other hand, if Nvim maintainers don't agree with me, then we absolutely shouldn't change the existing macOS name, which would pointlessly break consumers.

clason: > That's a good reason, if scripts are doing that. Do you have an example?

Not yet, because we don't offer different appimages (and the ARM runners were just released).

For the record, the typical (semi-official, as in some Apple tools use that) distinction is "Intel" vs. "Apple (silicon)"; people also used "M1" but that is now obsolete. "arm" is understood, too (and `uname` does return `arm64`; the point is simply "obsolete" vs. "new" since Intel hardware (and OS versions supporting it) are quickly vanishing. 

On Linux, both `aarch64` (ARM's official name for the architecture) and `arm64` are used (interchangeably). The sticker here is that on the runner we (and downstream projects) use, both `$ARCH` and CMake spit out `aarch64` when queried. So we get internal(!) consistency for free, and changing that requires manual intervention which I just don't see the value of.

clason: Fine; if that's what it takes to get this over with.

justinmk (approved):

clason:
@dundargoc cleanup welcome

Issue #32097, opened 2025-01-19
Undefined variable: s:termguicolors in .dircolors files


mhdna:
### Problem

Opening any .dircolors file containing some `dircolors` syntax causes an error message upon any cursor movement in that file:

```
Error detected while processing CursorMoved Autocommands for "<buffer=1>"..function <SNR>22_preview_color[41]..<SNR>22_get_hi_str:
line    3:
E121: Undefined variable: s:termguicolors
``` 

Surprisingly, opening a file not named `.dircolors` that contains some `dircolors` syntax and then enabling `:setfiletype dircolors` doesn't cause any errors on cursor movement and colors in that file are rendered just fine.

### Steps to reproduce

echo "DIR 01;38;5;7;48;5;4" > /tmp/.dircolors
nvim --clean /tmp/.dircolors
move the cursor around

### Expected behavior

No error messages

### Nvim version (nvim -v)

v0.10.3

### Vim (not Nvim) behaves the same?

no, vim9.1.1006

### Operating system/version

Arch Linux

### Terminal name/version

xterm

### $TERM environment variable

xterm-256color

### Installation

system package manager

zeertzjq:
Related #29056

juditnovak:
A PR addressing the issue: https://github.com/neovim/neovim/pull/32209
Detailed description of underlying reasons of the bug is available in the PR **Description**

juditnovak:
Thanks much @justinmk for merging and closing the issue :-)

PR #32201, opened 2025-01-25
[WIP][IGNORE][FIX] #32097 Dircolors, termguicolors variable undefined

juditnovak:
This fix is targeting Issue https://github.com/neovim/neovim/issues/32097

(NOTE: It's the first occasion that I interact with the `nvim` codebase. The hint represented here may be superficial or misguided. Furthermore I apologize in case I may have overlooked any of the CONTRIBUTING guidelines.)

I believe that the issue was slightly misleading, as script variable `termguicolors` was overlapping with the global variable `termguicolors`. The script variable was even different in type, as it is an array (while global `termguicolors` is a `bool`).

Renaming the script variable to `termguicolors_array` allows to see where the missing definition comes from, as the local variable wasn't initialized in the `Cursor*` autocommands.

### DIFF
--- a/runtime/syntax/dircolors.vim
+++ b/runtime/syntax/dircolors.vim
@@ -44,24 +44,24 @@ highlight default link dircolorsExtension Identifier
 highlight default link dircolorsEscape    Special
 
 function! s:set_guicolors() abort
-    let s:termguicolors = {}
+    let s:termguicolors_array = {}
 
-    let s:termguicolors[0]  = "Black"
-    let s:termguicolors[1]  = "DarkRed"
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but that file is taken from Vim, so this should ideally be fixed there.

juditnovak:
Sure, thanks much @clason for pointing me to the right direction :-)

juditnovak:
@clason : In fact, `neovim` is adding a small amount of code to the original `vim` syntax file. I made a 2nd attempt (https://github.com/neovim/neovim/pull/32209), taking those changes as the base. (More details on that PR **Description**.)

PR #32207, opened 2025-01-25
ci(release): add linux-aarch64 appimage and tarball [skip ci]

dundargoc:
Problem: No releases for ARM Linux.

Solution: Provide appimages and tarballs for `linux-aarch64`. Rename
x86 releases to `linux-x86_64` for consistency.


clason:
(Nobody uses "aarch64" for macOS...)

dundargoc:
Should we use `arm` for both instead then? I think windows also uses the moniker "arm", and the github runners also uses "arm".

clason:
You know my opinion on this.

.github/workflows/notes.md
```diff
@@ -27,32 +27,56 @@ Note: On Windows "Server" you may need to [install vcruntime140.dll](https://lea
 3. Extract: `tar xzvf nvim-macos-x86_64.tar.gz`
 4. Run `./nvim-macos-x86_64/bin/nvim`
 
-### macOS (arm64)
+### macOS (aarch64)
```
justinmk, line 30: Why rename this? The new linux arm build can be called `arm64`. Then they'll be consistent. And it's more discoverable for users. And easier for scripts that download our releases (used very commonly in CI).

justinmk: Continuing https://github.com/neovim/neovim/pull/32194#discussion_r1929566813 : 

> > scripts can use `$ARCH` to download the right appimage if we stick to the standard.)
> 
> That's a good reason, if scripts are doing that. Do you have an example? The example I've seen is our own node-client which points to various architectures, and it sure looks silly that "x86" and "arm" are spelled different ways for different OSes: https://github.com/neovim/node-client/blob/aa8fff0b7e486caf66a95befeed5805ae7e458f1/.github/workflows/ci.yml#L33-L43
> 
> Any scripts that download all the release archives for whatever reason have to treat each name specially, they can't "calculate" the name using simple logic.

clason: There's no good reason for renaming the macOS artifacts (which will break downstream scripts).

dundargoc:
Re-continues in https://github.com/neovim/neovim/pull/32194

PR #8175, opened 2018-03-25
syntax: fix dircolors plugin to support termguicolors

msva:

PR #32209, opened 2025-01-25
[FIX] #32097 Dircolors, termguicolors variable undefined

### LINKED ISSUE #32097: Undefined variable: s:termguicolors in .dircolors files

juditnovak:
fix https://github.com/neovim/neovim/issues/32097

`neovim` added `termguicolors` support, on top of `vim`'s `dircolors` syntax in contribution  **syntax: fix dircolors plugin to support termguicolors** https://github.com/neovim/neovim/pull/8175 (2018).

The currently proposed fix is aiming to introduce minimal changes on top of that enhancement.

## Issue

Current `dircolors` syntax setup doesn't seem prepared for `termguicolors` settings changing at runtime.

### Details

The syndrome described in the Bug Report is triggered by `termguicolors` being altered by automated settup taking effect after `runtime/syntax/dircolors.vim` was initially evaluated.

1. Script-local array variable `s:termguicolors` (*not* to be confused with the global `termguicolors` option) is [set up](https://github.com/neovim/neovim/blob/master/runtime/syntax/dircolors.vim#L208) when the script is loaded, depending on the value of the global `termguicolors` setting -- at the time. 
2. However, the global `termguicolors` settings can be altered altered later.
   - Other default setup (like  [this one](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/_defaults.lua#L708)) may apply
   - [Introducting](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/_defaults.lua#L566) a `VimEnter` autocommand, to set `termguicolors`
   - NOTE: This autocommand is executed after `dircolors` syntax initialization.
3. Later on, when `dircolors` syntax internal colors may be referred, the script-local (`s:termguicolors`) variable is not defined

## Solution

Keeping in mind that `termguicolors` may change on the fly, initializing internal structures before referred -- in case it was not yet initialized.

### DIFF
--- a/runtime/syntax/dircolors.vim
+++ b/runtime/syntax/dircolors.vim
@@ -85,6 +85,9 @@ endfunction
 function! s:get_hi_str(color, place) abort
     if a:color >= 0 && a:color <= 255
         if has('gui_running') || &termguicolors
+            if ! exists("s:termguicolors")
+              call s:set_guicolors()
+            endif
             return ' gui' . a:place . '=' . s:termguicolors[a:color]
         elseif a:color <= 7 || &t_Co == 256 || &t_Co == 88
             return ' cterm' . a:place . '=' . a:color


justinmk:
Ok. What's remaining to do here?

juditnovak:
@justinmk I was thinking to provide tests, however I'm not sure what may be the `neovim` policy for this case (i.e. fix on a `vim` syntax file extension). Could you please suggest? Thanks very much.

(If I'm not mistaken, test coverage may be missing for the feature, meaning original PR https://github.com/neovim/neovim/pull/8175 . Of course I'm glad to provide coverage for the feature itself, if needed, not just the fix :-) )

justinmk:
syntax plugins and ftplugins generally don't have tests.

PR #32219, opened 2025-01-27
Minor fixes for macOS

barracuda156:
This fixes two build errors on macOS 10.6.


clason:
Thanks, but we don't support that platform (see `:h supported`).

PR #32224, opened 2025-01-27
docs(treesitter): fix TSNode:range() type signature

ribru17:
Uses an overload to properly annotate the different return type based on the input parameter.


lewis6991 (approved):
I assume this doesn't create any LuaLS warnings anywhere.

PR #32226, opened 2025-01-27
ci(release)!: remove backwards compatible releases

dundargoc:
Remove `nvim-linux64.tar.gz` and `nvim.appimage` as maintaining
these is too much work.

Also fix directory names to be consistent.


CMakeLists.txt
```diff
@@ -22,6 +22,10 @@ endif()
 # Point CMake at any custom modules we may ship
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+  set(CMAKE_SYSTEM_PROCESSOR arm64)
```
justinmk, line 26: According to https://stackoverflow.com/a/70498851/152142 , this is probably fine. Though it mentions ARMClang args may be affected, and in theory, `Modules/CMake<LANG>Information.cmake` files are selected based on this env var.

To mitigate "risk", does it work if we only set this in `cmake.packaging/CMakeLists.txt` or somehow restrict it to the "packaging step" (cpack)?

scripts/genappimage.sh
```diff
@@ -75,6 +76,13 @@ chmod 755 AppRun
 
 cd "$APP_BUILD_DIR" || exit # Get out of AppImage directory.
 
+# We want to be consistent, so always use arm64 over aarch64
```
justinmk, line 79: nit: if `ARCH_ORIGINAL=$ARCH` is done just before this, then we can delete the other `ARCH_ORIGINAL=...` assignments.

scripts/genappimage.sh
```diff
@@ -9,6 +9,7 @@
 # App arch, used by generate_appimage.
 if [ -z "$ARCH" ]; then
   ARCH="$(arch)"
+  ARCH_ORIGINAL=$ARCH
   export ARCH
 fi
```
justinmk, line 15: nit: could put `ARCH_ORIGINAL="$ARCH"` here, then don't need other `ARCH_ORIGINAL=...` assignments.

CMakeLists.txt
dundargoc: Those files are selected before though, in the `project` function call. But we can set it in packaging as well and it should work, so let's do that.

justinmk (approved):

scripts/genappimage.sh
```diff
@@ -9,6 +9,7 @@
...
```
justinmk, line 14:
```suggestion
  export ARCH
fi
ARCH_ORIGINAL=$ARCH
```

PR #32227, opened 2025-01-27
fix(filetype): normalize full path before matching with patterns

markstegeman:
Problem:
On Windows, treesitter query files are not recognized as such when opened from inside their directory, because the full path returned from fnamemodify(_, ':p') contains backslashes, while the filetype patterns expect forward slashes.

Solution:
Normalize the result of fnamemodify(_, ':p') before trying to match it to filetype patterns.

### DIFF
--- a/runtime/lua/vim/filetype.lua
+++ b/runtime/lua/vim/filetype.lua
@@ -2873,7 +2873,7 @@ function M.match(args)
     name = normalize_path(name)
 
     -- First check for the simple case where the full path exists as a key
-    local path = fn.fnamemodify(name, ':p')
+    local path = normalize_path(fn.fnamemodify(name, ':p'))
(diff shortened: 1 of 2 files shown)


justinmk (approved):

runtime/lua/vim/filetype.lua
```diff
@@ -2565,6 +2565,15 @@ local function normalize_path(path, as_pattern)
   return normal
 end
 
+local abspath = function(x)
+  return fn.fnamemodify(x, ':p')
+end
+if fn.has('win32') == 1 then
+  abspath = function(x)
+    return (fn.fnamemodify(x, ':p'):gsub('\\', '/'))
```
justinmk, line 2573: After discussion, since match() is a performance-sensitive call, we decided to define a local `abspath()` which does the minimum work necessary to address the issue.

PR #32231, opened 2025-01-28
fix(terminal): support InsertCharPre in terminal mode

### LINKED ISSUE #5018: terminal-mode events: InsertCharPre, TextChangedI, CursorMovedI
### Expected behaviour

- When `:term` buffer contents change while user is in "terminal mode" (insert mode), insert-mode events such as ~~`TextChangedI`, `CursorMovedI`,~~ `InsertCharPre` should fire.
  - See also http://stackoverflow.com/a/38247473/152142.

Update: it makes sense to implement `InsertCharPre` because Nvim knows exactly what the user sends. But `TextChangedI`  and `CursorMovedI` are less obvious, because the terminal program may change those arbitrarily.

hgranthorner:
Closes: neovim/neovim#5018

Problem:
InsertCharPre events were not handled while inserting characters in terminal mode.

Solution:
Port the existing `do_insert_char_pre` procedure for insert mode from `edit.c` to `terminal.c` for terminal mode.

Notes:
This is my first PR for neovim, so please let me know if anything is missing. I also wasn't totally sure if this is considered a `fix` or a `feat`.

This solution is pretty copy pasta, but I'm not sure how bad of a thing that is. It also wasn't clear to me where where to put a factored out function, since the `autocmd.c` file seemed to be more about generic autocmd functions. Insert mode also seems to have a custom autocmd application function that I wasn't sure how to parameterize. Happy to factor it out if requested, and if some more experienced folks could give me a few pointers.

There may also need to be more tests required for edge cases I'm not aware of.


zeertzjq:
I don't think this is a good idea. It will break existing autocommands.

test/functional/autocmd/termxx_spec.lua
```diff
@@ -224,3 +224,14 @@ describe('autocmd TextChangedT', function()
...
 end)
+
+it('autocmd InsertCharPre', function()
+  clear()
+  command('autocmd InsertCharPre * ++once let g:called = 1')
+  command('terminal')
+  feed('i')
+  feed('q')
+  retry(nil, nil, function()
+    eq(1, api.nvim_get_var('called'))
```
justinmk, line 235: Need to test a few more cases such as rapid input, backspace, etc.

zeertzjq:
What is the use case of this autocommand? I don't see a use case mentioned in #5018.

haolian9:
hi all, i just walked in here and got reminded of [a pain point i met before](https://github.com/haolian9/zongzi/blob/master/nvim/.config/nvim/natives/lua/digits/Git.lua#L209).
so i currently simulate a 'cbreak' mode in a terminal buffer which runs `git add --patch a-file`, by `tnoremap a-z a-z<cr>`. with this
autocmd available, maybe i can do it in a clean way, like:
```
aucmd('InsertCharPre', {buffer = termbuf, callback = function()
    vim.schedule(function() feedkeys('<cr>', 'n') end)
end)
```

vlelo:
> What is the use case of this autocommand? I don't see a use case mentioned in #5018.

As a very banal example, software CapsLock:
https://github.com/barklan/capslock.nvim
https://github.com/tpope/vim-capslock

(This, of course, has its issues, like for example CapsLock being enabled in a TUI not being an intuitive behaviour)

haolian9:
i feel i should say it again (i once deleted my comment),
this feature can be used to simulate 'cbreak' mode in terminal buffer in a brief way (compare to `tmap <buf> a-z`:

```
if termspec.cbreak then
  aug:repeats("InsertCharPre", {
    callback = function()
      if vim.v.char == "\r" then return end
      vim.v.char = vim.v.char .. '\r'
    end,
  })
end
```

this 'cbreak' mode can make `:terminal git add -p` more convinent.

haolian9:
(sorry for the anonying updates caused by my commits, i tried to rephrase the commit message to make github not recognize them but failed)

justinmk:
Closing since comments were not addressed, but also will mention that `aucmd_defer` from https://github.com/neovim/neovim/pull/35793 greatly reduces the risk of adding new events.

That may be useful for this event, if anyone wants to continue this.

PR #32234, opened 2025-01-28
fix(treesitter) Set modeline=false in TSHighlighter:destroy

### LINKED ISSUE #32220: b:undo_ftplugin unmapping fails when filetype is changed from help set via modeline
### Problem

If a file has a modeline `vim:ft=help`, changing filetype raises E31 errors:
```
Error detected while processing FileType Autocommands for "*"..function <SNR>1_LoadFTPlugin[2]..modelines[1]..FileType Autocommands for "*"..function <SNR>1_LoadFTPlugin:
line    2:
E31: No such mapping
E31: No such mapping
```

which must come from https://github.com/neovim/neovim/blob/c7d4a77ff9040c8b242f9d12a98472fe6217328c/runtime/ftplugin/help.lua#L85

This does not occur when ft is set to help but not by modeline.

### Steps to reproduce

```
echo 'vim:ft=help' | nvim --clean -
:set ft=
```

### Expected behavior

E31 is not issued.

### Nvim version (nvim -v)

v0.11.0-dev-c7d4a77ff904

no, vim 9.1.1043

### Operating system/version

EndeavourOS

### Terminal name/version

GNOME Terminal 3.54.3 using VTE 0.78.3 +BIDI +GNUTLS +ICU +SYSTEMD

xterm-256color

### Installation

build from repo

daniel-petrovic:
Fix #32220

Problem:
  TSHighlighter:destroy() may cause double-processing of the modeline
  and failure of b:undo_ftplugin.

Solution:
  As suggested by @seandewar 

  Disable modeline in TSHighlighter:destroy() by setting modeline=false if executing syntaxset autocommands
  for the FileType event.

### DIFF
--- a/runtime/lua/vim/treesitter/highlighter.lua
+++ b/runtime/lua/vim/treesitter/highlighter.lua
@@ -160,7 +160,10 @@ function TSHighlighter:destroy()
     vim.bo[self.bufnr].spelloptions = self.orig_spelloptions
     vim.b[self.bufnr].ts_highlight = nil
     if vim.g.syntax_on == 1 then
-      api.nvim_exec_autocmds('FileType', { group = 'syntaxset', buffer = self.bufnr })
+      api.nvim_exec_autocmds(
+        'FileType',
+        { group = 'syntaxset', buffer = self.bufnr, modeline = false }
+      )
     end
   end
 end


justinmk:
needs a problem/solution that explains the change. https://github.com/neovim/neovim/issues/32220#issuecomment-2618683233 clearly explained the solution.

explaing the change is the main work here, not a 1-line code change.

PR #32237, opened 2025-01-28
fix(ui): avoid redundant ext_cmdline events

luukvbaal:
Problem:  cmdline_show is emitted unnecessarily each event
          loop iteration, because `cmdline_was_last_drawn` is never set.
Solution: Keep track of whether the cmdline was last drawn to avoid
          unnecessarily emitting `cmdline_show`. Set `redraw_state` to
          emit `cmdline_pos` when emitting `CursorMovedC`. Only emit
          `cmdline_pos` when cmdline was last drawn.

### DIFF
--- a/src/nvim/drawscreen.c
+++ b/src/nvim/drawscreen.c
@@ -692,8 +692,11 @@ int update_screen(void)
 
   decor_providers_invoke_end();
 
-  // either cmdline is cleared, not drawn or mode is last drawn
-  cmdline_was_last_drawn = false;
+  // Either cmdline is cleared, not drawn or mode is last drawn.
+  // This does not (necessarily) overwrite an external cmdline.
+  if (!ui_has(kUICmdline)) {
+    cmdline_was_last_drawn = false;
+  }
   return OK;
 }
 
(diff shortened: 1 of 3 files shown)


justinmk (approved):
🚀

folke:
Just tested this PR and seems to be fixed indeed!

folke (approved):

Sam-programs:
> Set redraw_state to emit cmdline_pos when emitting CursorMovedC. Only emit
cmdline_pos when cmdline was last drawn.

Hey, this doesn't seem to work as intended. Currently neovim sends `cmdline_pos` whenever the cursor moves when both `ext_messages` and `ext_cmdline` are active, but when only `ext_cmdline` is active, it sends `cmdline_pos` once if the cursor moves after the command-line changes:

https://github.com/user-attachments/assets/6af75106-2505-4673-aabc-50f7517990f7

During both tests I type "asdasd" then press the arrow keys a few times. the cursor was flickering because I was moving it. Once a test is done I press `<ctrl-b>` to check the output. Here's the init file:
```lua
-- init.lua
local ns = vim.api.nvim_create_namespace("Foo")                        
local i = 0
vim.ui_attach(ns, {ext_cmdline = true, }, function(event, ...)
    if event == "cmdline_pos" then
        vim.print("Cursor moved " .. i)
        i = i + 1
    end
end)

vim.keymap.set("n", '<c-b>', function() 
    vim.ui_detach(ns)
end)
```

PR #32243, opened 2025-01-29
Fix "vim_snprintf() positional arguments" failures on 32-bit systems

jamessan:
ERROR    test/unit/testutil.lua @ 785: vim_snprintf() positional arguments
    test/unit/testutil.lua:757: test/unit/testutil.lua:741: (string) '
    test/unit/strings_spec.lua:143: Expected objects to be the same.
    Passed in:
    (number) 6400
    Expected:
    (number) 6'
    exit code: 256

When 9.0.1856 was ported, the numbers being formatted as binary were cast
to "unsigned int" rather than uvarnumber_T, as is done upstream.

### DIFF
--- a/test/unit/strings_spec.lua
+++ b/test/unit/strings_spec.lua
@@ -140,7 +140,13 @@ end)
 
 describe('vim_snprintf()', function()
   local function a(expected, buf, bsize, fmt, ...)
-    eq(#expected, strings.vim_snprintf(buf, bsize, fmt, ...))
+    local args = { ... }
+    local ctx = string.format('snprintf(buf, %d, "%s"', bsize, fmt)
+    for _, x in ipairs(args) do
+      ctx = ctx .. ', ' .. tostring(x)
+    end
+    ctx = ctx .. string.format(') = %s', expected)
+    eq(#expected, strings.vim_snprintf(buf, bsize, fmt, ...), ctx)
     if bsize > 0 then
       local actual = ffi.string(buf, math.min(#expected + 1, bsize))
       eq(expected:sub(1, bsize - 1) .. '\0', actual)
--- a/test/unit/strings_spec.lua
+++ b/test/unit/strings_spec.lua
@@ -1,12 +1,19 @@
 local t = require('test.unit.testutil')
 local itp = t.gen_itp(it)
 
+local child_call_once = t.child_call_once
 local cimport = t.cimport
 local eq = t.eq
 local ffi = t.ffi
 local to_cstr = t.to_cstr
 
 local strings = cimport('stdlib.h', './src/nvim/strings.h', './src/nvim/memory.h')
+
+local UVARNUM_TYPE
+
+child_call_once(function()
+  UVARNUM_TYPE = ffi.typeof('uvarnumber_T')
+end)
 
 describe('vim_strsave_escaped()', function()
   local vim_strsave_escaped = function(s, chars)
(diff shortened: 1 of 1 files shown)


zeertzjq:
Why is it called `BLOB_TYPE`?

jamessan:
> Why is it called `BLOB_TYPE`?

I had it in my head that this was specific to blobs, for some reason. Changed it to `UVARNUM_TYPE`.

nmeum:
This fixes test failures on the i386 32-bit musl arch. Sadly, on the armhf and armv7 32-bit musl arches these tests still fail with these patches applied.

Without this patch applied:

```
ERROR    test/unit/testutil.lua @ 790: vim_snprintf() positional arguments
test/unit/testutil.lua:762: test/unit/testutil.lua:746: (string) '
test/unit/strings_spec.lua:143: Expected objects to be the same.
Passed in:
(number) 4
Expected:
(number) 6'
exit code: 256
stack traceback:
	test/unit/testutil.lua:762: in function 'itp_parent'
	test/unit/testutil.lua:800: in function <test/unit/testutil.lua:790>
```

With this patch applied:

```
ERROR    test/unit/testutil.lua @ 790: vim_snprintf() truncation
test/unit/testutil.lua:762: test/unit/testutil.lua:746: (string) '
test/unit/strings_spec.lua:153: Expected objects to be the same.
Passed in:
(string) '000'
Expected:
(string) '001'
exit code: 256
stack traceback:
	test/unit/testutil.lua:762: in function 'itp_parent'
	test/unit/testutil.lua:800: in function <test/unit/testutil.lua:790>
ERROR    test/unit/testutil.lua @ 790: vim_snprintf() positional arguments
test/unit/testutil.lua:762: test/unit/testutil.lua:746: (string) '
test/unit/strings_spec.lua:153: Expected objects to be the same.
Passed in:
(string) '000'
Expected:
(string) '001'
exit code: 256

stack traceback:
	test/unit/testutil.lua:762: in function 'itp_parent'
	test/unit/testutil.lua:800: in function <test/unit/testutil.lua:790>
```

test/unit/strings_spec.lua
```diff
@@ -140,13 +147,22 @@ end)
...
+    local args = { ... }
+    local ctx = string.format('snprintf(buf, %d, "%s"', bsize, fmt)
+    for _, x in ipairs(args) do
+      ctx = ctx .. ', ' .. tostring(x)
+    end
+    ctx = ctx .. string.format(') = %s', expected)
+    eq(#expected, strings.vim_snprintf(buf, bsize, fmt, ...), ctx)
     if bsize > 0 then
       local actual = ffi.string(buf, math.min(#expected + 1, bsize))
       eq(expected:sub(1, bsize - 1) .. '\0', actual)
```
nmeum, line 159: Probably also want to provide that `ctx` information for this assertion as well.

nmeum:
> I assume this doesn't happen on the 32-bit arm Debian builders?

Ah! Looks like this test case was disabled https://salsa.debian.org/vim-team/neovim/-/commit/3b97fb41503dc7dd78bb840557e6586d999ff2ad

test/unit/strings_spec.lua
jamessan: Yes, I ran into the same issue and was going to add that in a follow-up PR. However, I haven't had time to figure out the fix for the other issues you (and I) ran into.

Any help is appreciated.

PR #32249, opened 2025-01-29
ci!: store artifact shasums in a single shasum.txt file

dundargoc:
Users can parse this file to get the shasum they require.


justinmk:
Great idea!

Issue #7769, opened 2017-12-26
Multiline completion (also affects LSP textDocument/completion results)


hauleth:
Currently Vim (and by bloodline also NeoVim) allows only single line completions. This is quite limiting as there is also no way to determine beginning and end of the completed text so this leaves us in the limbo where we need to do some hackish things to achieve things like multiline snippets.

Most obvious way to do such thing would be to make (`:h complete-items`):

```viml
{ 'word': "aaa\nbbb" }
```

to insert 2 lines:

```
aaa
bbb
```

Currently it inserts

```
aaa^@bbb
```

Where `^@` is NUL character (`:h NL-used-for-NUL`).

The problems with multiline completion is that there will be open question how to handle indentation.

Alternatively we could allow `List` as `'word'` parameter where each line would be one item.

justinmk:
> Alternatively we could allow List as 'word' parameter where each line would be one item.

Something like that sounds reasonable.

hauleth:
@justinmk probably, but the first proposal is on par with `'info'` field of `complete-items`. The only problem is that in such case you cannot complete entries containing `NUL` character.

joereynolds:
Thinking out loud but if this made it in, it's not really that much different from the floating windows that are also incoming is it?

Would there be plans to make the omnicompletion popup use a floating window?

hauleth:
@joereynolds this isn't related to #6619 in any way, as this is only about inserting text, not about displaying it in popup window.

hauleth:
I have also created vim/vim#2505 issue with proposition of such feature.

justinmk:
# Status

From https://github.com/vim/vim/issues/2505 : 
> the current behavior of inserting `^@` presumably is just an implementation artifact, not useful behavior.



- work started in https://github.com/vim/vim/pull/15373
- workaround: https://github.com/vim/vim/issues/2505#issuecomment-354175243

justinmk:
Fixed by

- #32377

PR #32251, opened 2025-01-30
vim-patch:9.1.1056: Vim doesn't highlight to be inserted text when completing

zeertzjq:
#### vim-patch:9.1.1056: Vim doesn't highlight to be inserted text when completing

Problem:  Vim doesn't highlight to be inserted text when completing
Solution: Add support for the "preinsert" 'completeopt' value
          (glepnir)

Support automatically inserting the currently selected candidate word
that does not belong to the latter part of the leader.

closes: vim/vim#16403

https://github.com/vim/vim/commit/edd4ac3e895ce16034c7e098f1d68e0155d97886

Co-authored-by: glepnir <glephunter@gmail.com>


justinmk:
This is cool! Some improvements:

- default theme doesn't highlight `ComplMatchIns`. cc @echasnovski 
- this feature will be much more useful after https://github.com/neovim/neovim/issues/7769

echasnovski:
>     * default theme doesn't highlight `ComplMatchIns`. cc @echasnovski

This is [intended](https://github.com/neovim/neovim/pull/31615#issuecomment-2551350979). I think having it be completely invisible by default and strictly opt-in for color scheme is the most expected behavior here. There might be a merit of adding a comment that it is intended, not sure if it is needed though?

justinmk:
> `ComplMatchIns` should be completely invisible by default and strictly opt-in for color schemes.

Is that because we expect most people want "syntax highlighting" of the inserted text?

It seems like an improvement for the user to see which text is being inserted.

echasnovski:
> > `ComplMatchIns` should be completely invisible by default and strictly opt-in for color schemes.
> 
> Is that because we expect most people want "syntax highlighting" of the inserted text?
> 
> It seems like an improvement for the user to see which text is being inserted.

Mostly to preserve syntax highlighting, yes. It is also usually visually clear which part is being inserted, as it is the same as the selected entry in the popup. There is already a default visual feedback about difference between present text and what-to-be-inserted via bold `PmenuMatch` and `PmenuMatchSel`, so duplicating it by default seems to be too much.

glepnir:
description is a bit strange. CompMatchIns has been added before. This highlight is applied to all parts that are not leaders. Preinsert inserts the parts that are not leaders but keeps the cursor position.

PR #32256, opened 2025-01-30
fix(colorscheme): make `CursorLine`, `Folded`, `StatusLineNC` different

### LINKED ISSUE #30612: StatusLineNC, Folded, and CursorLine highlights are indistinguishable in default colorscheme

echasnovski:
Problem: currently `CursorLine`, `Folded`, `StatusLineNC` highlight
  groups have the same background color in default color scheme (Grey3).
  This is a result of optimizing their highlighting to be different from
  `Normal` but not draw too much attention.

  However, this design has a usability issue as those groups are
  sometimes appear together which can make it hard (but not impossible)
  to differentiate between them.

  This was previously partially resolved with `StatusLineNC` using Grey1
  as background (and thus be different from `CursorLine` but closer to
  `Normal`), but optimizing more towards it being a visible separator
  between windows was decided to be more important.

Solution: make `Folded` use Grey1 and `StatusLineNC` use Grey4. This
  makes all three highlight groups have different backgrounds with the
  following consequences:
    - Folds now have the same background as floating windows. It makes them there differentiable only by the value of 'foldtext' (which is usually enough). Optimizing for the case "cursor line is next to the fold" seems to be more useful than for the case "make folds in floating window differ by background".
    - Statusline of inactive windows now draw more attention to them. The benefits are that they are different from cursor column and are better window separators.
    - Inactive tabline (both `TabLine` and `TabLineFill`) now also draws a bit more attention to it (as they are linked to `StatusLineNC`) but with the benefit of also being different from `CursorLine`.

------

Resolve #30612 

------

Comparisons.

Current `master`:

screenshot_2025-01-30_11:25:14

screenshot_2025-01-30_11:25:03

After this PR:

screenshot_2025-01-30_11:24:25

screenshot_2025-01-30_11:24:12


echasnovski:
I don't think this requires a 'news.txt' entry, so there was nothing added there.

echasnovski:
Failed tests don't look related. The `ubuntu arm gcc functionaltest` has `treesitter/fold_spec.lua @ 720` failing with different fold text, not highlighting. Probably a result of async tree-sitter folds?

clason:
No, just (increasingly) flaky tests.

justinmk:
This is perfect.

justinmk:
I've been using underdashed for `CursorLineNC`, it really helps. WDYT?

image

echasnovski:
> I've been using underdashed for `CursorLineNC`, it really helps. WDYT?

There is no such highlight group :thinking:
And even if it were, my first impression is that it introduces more noise instead of reducing it.

PR #32266, opened 2025-01-31
fix(lsp): autotrigger should only trigger on client's triggerCharacters

rmuir:
Problem: `autotrigger` option of `vim.lsp.completion.enable()` would trigger all clients, as long as it matched at least one client's `triggerCharacters`.

Solution: trigger only the clients with triggerCharacters matching the character. overtriggering still happens if any client returns `isIncomplete=true` (this case is more involved).

For example: jdtls is really slow client especially on a big java codebase. All responses from it are laggy, if you send it requests too fast or for strange triggerCharacters, it will destabilize, start returning errors, etc. But it is worth the pain to trigger it for its triggerCharacters (e.g. `.`).

I have "fast clients" I'd like to use on other trigger characters, e.g. `[a-zA-Z0-9_]`, which just do treesitter locals, snippets, etc. But currently, I can't use them without also triggering the slow jdtls, too. With this patch, everything works together wonderfully, as clients are only triggered on the characters they should be.

This issue might just be a simple oversight: the code goes to great efforts to track a mapping of `char -> client[]` but then doesn't use the return value when actually triggering.

I could use guidance on lua testing: I understand the existing tests, but I don't see how to test the `autotrigger`.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -518,11 +518,14 @@ local function on_insert_char_pre(handle)
   end
 
   local char = api.nvim_get_vvar('char')
-  if not completion_timer and handle.triggers[char] then
+  local matched_clients = handle.triggers[char]
+  if not completion_timer and matched_clients then
     completion_timer = assert(vim.uv.new_timer())
     completion_timer:start(25, 0, function()
       reset_timer()
-      vim.schedule(M.trigger)
+      vim.schedule(function()
+        trigger(api.nvim_get_current_buf(), matched_clients)
+      end)
     end)
   end
 end
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -770,13 +770,14 @@ end)
 
 --- @param name string
 --- @param completion_result lsp.CompletionList
+--- @param trigger_chars? string[]
 --- @return integer
-local function create_server(name, completion_result)
+local function create_server(name, completion_result, trigger_chars)
   return exec_lua(function()
     local server = _G._create_server({
       capabilities = {
         completionProvider = {
-          triggerCharacters = { '.' },
+          triggerCharacters = trigger_chars or { '.' },
         },
       },
       handlers = {
@@ -793,6 +794,7 @@ local function create_server(name, completion_result)
       cmd = server.cmd,
       on_attach = function(client, bufnr0)
         vim.lsp.completion.enable(true, client.id, bufnr0, {
+          autotrigger = trigger_chars ~= nil,
           convert = function(item)
             return { abbr = item.label:gsub('%b()', '') }
           end,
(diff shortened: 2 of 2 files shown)


MariaSolOs:
The change looks correct to me, but @mfussenegger should have the final say as our completion expert.

mfussenegger (commented):
One problem with this approach is that the subsequent characters won't trigger the remaining clients because of the `pumvisible()` condition. If you type `a.` and `a` is a trigger character for client-1 and `.` for client-2, then the `.` won't trigger client-2.

Always triggering all clients was a bit of a simplification for that case, but also because it might confuse users if some candidates don't show up. Not opposed to change that behavior, but the case above would need to be handled.

rmuir:
@mfussenegger hmm, that is my use-case that is working perfectly with this change. Wouldn't client-1 have to actually return an actual completion of `a.` for that to happen?

In my use case (which I guess I thought would be the typical one, maybe not?) client-1 is trying to complete identifiers, it is the fast, dumb one, but then I want to use client-2 (heavy-duty LSP) to trigger on  `.`

when i type the `.` the slower client is triggered correctly.

Here's what it looks like right after i press `c`:
Screen_Shot_2025-02-01_at_15 28 02

Then i press the `.` and jdtls is triggered correctly:
Screen_Shot_2025-02-01_at_15 28 24

My issue with the current code is that I have no way to prevent the completion calls from being sent to the slow jdtls. it slows down the whole neovim, because it waits on responses from all the servers.

On the other hand, with this patch, if the user wants to configure "union of all trigger characters" for client-1 and client-2, they can configure that easily via `triggerCharacters` in their LSP configuration and still have the previous behavior?

rmuir:
I understand what you are getting, if `client-1` returns `a.` (with isIncomplete=false).

In such a case, where the `.` is already part of "complete" completion response and pum is still visible, it would be nice to just treat it as if it were `incomplete=true`, which will trigger all clients.

I do think this seems pretty odd/rare though.

rmuir:
@mfussenegger I added a commit with a stab at handling that case. It adds some complexity and makes things a bit ugly: I'd appreciate any advice on simplifying it.

runtime/lua/vim/lsp/completion.lua
```diff
@@ -501,7 +504,22 @@ end
...
+
   if tonumber(vim.fn.pumvisible()) == 1 then
+    -- completion from clientA could have a trigger of clientB:
+    -- fall back to retriggering all clients (treat as incomplete)
+    if
+      matched_clients
+      and matched_clients ~= Context.current_clients
+      and vim.iter(matched_clients):any(function(client)
+        return not vim.tbl_contains(Context.current_clients, client)
+      end)
```
rmuir, line 518: I'm not happy about this piece for the corner case, but my knowledge of lua is extremely limited and it shows :)

my goto here normally, would be to track client IDs (not clients), and to use bitset for the trigger mapping and this special check (and-not).

mfussenegger:
After sleeping on this and seeing the impact I'd almost tend to go with your initial approach for now, because the new one also has the downside that it re-requests from the first server - which could be the slower one instead of the faster one. For example with your snippet server triggering on a, e, i, o, u and jdtls triggering on `.` if you type `x.e`, the `.` would trigger jdtls and `e` would then re-trigger both jdtls and snippets.

To solve that we'd have to carry-over the first result, or be able to extend the already shown candidates.

cc @glepnir do you think `complete_add` could be extended to allow adding additional entries to the menu created via `complete` ? And not just as string, but with the full dict/table structure.

glepnir:
`.` and `e` are two different completion cycles? `e` will input as a new leader to filter completion items from `compl_match_array` . if adding new completion from result which send by `e` , the startcol still use the col of `.`?

rmuir:
I rolled this one back to the original approach.

glepnir:
@mfussenegger  I have tried add a new function `complete_append` instead of extending complete_add , this is test script

```vim
function! AppendCandidate() abort
  if v:char == 'b'
    call timer_start(100, {-> complete_append(1, [ {'word': 'bool'}, {'word': 'balon'}, {'word': 'foo'} ]) })
    echomsg "done"
  endif
endfunction

inoremap <buffer> <f5> <cmd>call complete(1, ["red", "blue"])<cr>

autocmd InsertCharPre * :call AppendCandidate()
```

it works like . how do you feel

test

mfussenegger:
@rmuir thanks I added a test case and merged this.

@glepnir `complete_append` looks promising. I created https://github.com/neovim/neovim/issues/32428 to track that

rmuir:
@mfussenegger thank you for writing the test!

Issue #31936, opened 2025-01-09
Inconsistent behavior of registers when clipboard provider not found


UnkwUsr:
### Problem

When clipboard provider not found, trying to write to registers `"+` and `"*` shows error, but still this registers gets created and new content gets written to them. We can prove this by checking `:registers` output. But trying to paste from this registers results in pasting from `""`.

### Steps to reproduce

(`PATH=` used to avoid detecting clipboard provider. But better to also check manually that it is not set: `:checkhealth provider.clipboard`)

```
echo -e 'aaa\nbbb' | PATH= /bin/nvim --clean
"+yy
j
yy
"+p
:registers
```

At then end `:registers` shows that `"+` contains `aaa` but we clearly see that when we did `"+p`, it pasted content of `""`, not `"+`.

## Solution 1

Make `"+` and `"*` registers act as any other named register if clipboard provider not found. So it will write and read registers without falling back to `""` as it does now.

## Solution 2

Prevent creating (and writing to) registers `"+` and `"*` when clipboard provider not found. So, after `"+yy`, doing `"+p` should not paste anything.

### Nvim version (nvim -v)

v0.10.3

### Vim (not Nvim) behaves the same?

no, gvim 9.1

### Operating system/version

archlinux

### Terminal name/version

kitty

### $TERM environment variable

xterm-kitty

### Installation

pacman

UnkwUsr:
Behavior in vim 9.1 (with +clipboard compiled feature): it does not write to those registers, but on reading it fallbacks to `""` the same as nvim

AbhijithKumble:
I want to try solving this issue, how do i start solving this issue ?

justinmk:
@AbhijithKumble read https://github.com/neovim/neovim/blob/master/CONTRIBUTING.md 

you should start by try things out and ask specific questions. 

> how do i start solving this issue ?

answering that question is about the same amount of work as fixing the issue.

AbhijithKumble:
So which of those two proposed solutions should i implement ?

> Expected behavior
>
> Solution 1
> 
> Make "+ and "* registers act as any other named register if clipboard provider not found. So it will write and read registers without falling back to "" as it does now.
>
> Solution 2
> 
> Prevent creating (and writing to) registers "+ and "* when clipboard provider not found. So, after "+yy, doing "+p should not paste anything.

justinmk:
Solution 1 sounds like what I would favor, unless others have comments.

UnkwUsr:
[...]

PR #32269, opened 2025-01-31
fix(yank): ensure registers "+ and "* behave consistently without clipboard provider

AbhijithKumble:
Problem: When the clipboard provider is unavailable, writing to `"+` and `"*` registers shows an error but still stores data. However, pasting from them (`"+p`) unexpectedly falls back to `""`, leading to inconsistent behavior.

Solution: Ensure `"+` and `"*` registers behave like normal named registers when the clipboard provider is unavailable, preventing fallback to `""` and ensuring predictable yank and paste operations.

### DIFF
--- a/src/nvim/ops.c
+++ b/src/nvim/ops.c
@@ -914,16 +914,6 @@ yankreg_T *get_yank_register(int regname, int mode)
       && get_clipboard(regname, &reg, false)) {
     // reg is set to clipboard contents.
     return reg;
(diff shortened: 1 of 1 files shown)


AbhijithKumble:
fixes #31936

justinmk:
> preventing fallback to `""`

https://github.com/neovim/neovim/issues/31936#issuecomment-2662145289 mentions that as vim-compatible behavior.

meanwhile, CI needs a lot of work here.

PR #32275, opened 2025-01-31
fix(diagnostic): improve current_line refresh logic

### LINKED ISSUE #32241: `current_line` diagnostic handler does not update correctly
### Problem

Image

with 
```lua
vim.diagnostic.config({
  virtual_lines = { current_line = true },
})
```
Due to the ordering of the auto command being destroyed on `hide` the diagnostic only updates for the most recently edited buffer and doesn't track the cursor correctly.

### Steps to reproduce

create two errors in file
create split
create two errors in second file
switch between the two splits (using `<c-w>` or `mouse` and observe that the diagnostic only redraws for the most recently edited buffer

### Expected behavior

error to follow cursor

### Nvim version (nvim -v)

HEAD

no

### Operating system/version

Linux

### Terminal name/version

Kitty

xterm-256

### Installation

build from source

tris203:
Problem:
The current implementation uses a global augroup for virtual lines in diagnostics, which can lead to conflicts and unintended behavior when multiple namespaces/buffers are involved.

Solution:
Refactor the code to use a namespace-specific augroup for virtual lines. This ensures that each namespace has its own augroup. Scope the clear commands to only the relevant buffer.

Closes #32241

### DIFF
--- a/runtime/lua/vim/diagnostic.lua
+++ b/runtime/lua/vim/diagnostic.lua
@@ -344,7 +344,6 @@ local global_diagnostic_options = {
 --- @class (private) vim.diagnostic.Handler
 --- @field show? fun(namespace: integer, bufnr: integer, diagnostics: vim.Diagnostic[], opts?: vim.diagnostic.OptsResolved)
 --- @field hide? fun(namespace:integer, bufnr:integer)
---- @field _augroup? integer
 
 --- @nodoc
 --- @type table<string,vim.diagnostic.Handler>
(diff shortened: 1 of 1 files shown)


tris203:
cc @MariaSolOs

MariaSolOs:
Are you sure that this fixes https://github.com/neovim/neovim/issues/32241? Because I had tried this earlier this week locally and I could still repro the issue.

MariaSolOs:
Also for reference, this is the fix suggested by https://lists.sr.ht/~whynothugo/lsp_lines.nvim/patches/47413

tris203:
> Are you sure that this fixes #32241? Because I had tried this earlier this week locally and I could still repro the issue.

I can't reproduce the error not tracking the cursor with this change

MariaSolOs:
> I can't reproduce the error not tracking the cursor with this change

Thanks for confirming!

MariaSolOs (approved):

PR #32281, opened 2025-02-01
fix(ftplugin): change PHP commentstring to '// %s'

brianhuster:
Problem:
Currently, `'commentstring'` of PHP is set to `/* %s */`. However, this doesn't play nice with Vim's builtin comment plugin because the plugin is based on line commenting instead of block commenting. And when writing PHP, people often use `// %s` for line commenting instead of `/* %s */`

Solution:
Change `'commentstring'` to `// %s` for PHP.

### DIFF
--- a/runtime/ftplugin/php.lua
+++ b/runtime/ftplugin/php.lua
@@ -1,0 +1,3 @@
+vim.bo.commentstring = '// %s'
+
+vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '') .. '\n setl commentstring<'


clason:
It does play nice with _Neovim's_ builtin comment plugin, though, so I don't see the need for deviating from upstream.

runtime/ftplugin/php.lua
```diff
@@ -0,0 +1,5 @@
+-- These are the default option values in Vim, but not in Nvim, so must be set explicitly.
```
clason, line 1: That is simply not correct.

clason:
Sorry, but these files are not maintained by us, so any change has to go through the listed filetype maintainer.

brianhuster:
> It does play nice with _Neovim's_ builtin comment plugin, though, so I don't see the need for deviating from upstream.

Because Neovim's builtin comment plugin works based on comment line instead of comment block. For comment line in PHP, people usually use `// %s` or `# %s` (I think the first is more popular). I have never seen anyone using `/* %s */` for just single line of comment.

Also see #29085 . I don't understand the commentstring has been changed for C, why it can't be changed for PHP too?

clason:
Because 1. this change was made in the connection with removing the default (C) commentstring, which was necessary for treesitter injection support in the comment plugin and 2. Neovim is written in C and we don't use `/* */` comments at all (unlike Vim). So that is a very different situation.

brianhuster:
The maintainer of PHP ftplugin maintainer [has no strong opinion one way or the other](https://github.com/vim/vim/pull/16507#issuecomment-2609432972). But anyway I failed to persuade Vim maintainer for the change. 

But ok, if that's your policy, I have no objection

clason:
Yes, sorry about that. You can always add that to your personal `ftplugin`, of course.

chrisbra:
> The maintainer of PHP ftplugin maintainer https://github.com/vim/vim/pull/16507#issuecomment-2609432972. But anyway I failed to persuade Vim maintainer for the change.

Yes that also means he **does not support for the change** right now. In any case, you haven't yet made convincing arguments to change the default for other users plus there was not a lot of support for this change. We may reconsider if more people vote for this change, but currently there is not a convincing argument for the change, but one argument against the change (that is: changing the default will annoy people used to it).

PR #32282, opened 2025-02-01
feat(treesitter): table of contents for checkhealth, markdown

clason:
Problem: No table of contents in health checks and Markdown files.

Solution: Refactor `show_toc()` function to support multiple queries. Add Markdown query. Add buffer-local `gO` mapping in `checkhealth` and `markdown` ftplugin.

Also rename (internal) `vim.vimhelp` module to `vim.render` as itcontains generic functions.

---

Problem: Cannot jump between sections with `[[` and `]]`.

Solution: Refactor `show_toc` function and implement `jump` function for
text buffers. Map to `[[`/`]]` in `help` and `checkhealth` ftplugin.

(Turns out markdown already has these mappings in its ftplugin file. We could still override those since the treesitter functions are bound to be more robust than the simple regex searches.)

---

Possible follow-ups:
* refactor the `g==` function and add it for markdown files (not done because it's too noisy right now).

close #32165


runtime/ftplugin/help.lua
```diff
@@ -82,5 +91,6 @@ vim.keymap.set('n', 'g==', function()
 end, { buffer = true })
 
 vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '')
-  .. '\n exe "nunmap <buffer> gO" | exe "nunmap <buffer> g=="'
+  .. '\n exe "nunmap <buffer> gO" | exe "nunmap <buffer> g==" | '
+  .. '\n exe "nunmap <buffer> ]]" | exe "nunmap <buffer> [["'
```
zeertzjq, line 95: Why are the second and third commands separated by both a `|` and a `\n`?

runtime/ftplugin/checkhealth.lua
```diff
@@ -0,0 +1,14 @@
...
+vim.keymap.set('n', ']]', function()
+  require('vim.render').jump({ count = 1, level = 1 })
+end, { buffer = 0, silent = false, desc = 'Jump to next section' })
+vim.keymap.set('n', '[[', function()
+  require('vim.render').jump({ count = -1, level = 1 })
+end, { buffer = 0, silent = false, desc = 'Jump to previous section' })
+
+vim.b.undo_ftplugin = (vim.b.undo_ftplugin or '')
+  .. '\n exe "nunmap <buffer> gO"'
+  .. '\n | exe "nunmap <buffer> ]]" | exe "nunmap <buffer> [["'
```
zeertzjq, line 14: Same here

runtime/ftplugin/help.lua
clason: eh, better safe than sorry

runtime/doc/news.txt
```diff
@@ -379,6 +379,9 @@ TREESITTER
   activated by passing the `on_parse` callback parameter.
 • |vim.treesitter.query.set()| can now inherit and/or extend runtime file
   queries in addition to overriding.
+• |gO| now works in `help`, `checkhealth`, and `markdown` buffers.
+• Jump between sections in `help` and `checkhealth` buffers with `[[` and
+  `]]`.
```
justinmk, line 384: maybe put this in the EDITOR section? the ts part is a implementation detail.

runtime/lua/vim/render.lua
```diff
@@ -0,0 +1,162 @@
+-- Extra functionality for rendering rich text buffers (vim help, markdown)
```
justinmk, line 1: shouldn't this stuff live in `vim.ui` ?

runtime/lua/vim/render.lua
```diff
@@ -0,0 +1,162 @@
...
+
+local M = {}
+
+--- Apply current colorscheme to lists of default highlight groups
+---
+--- Note: {patterns} is assumed to be sorted by occurrence in the file.
+--- @param patterns {start:string,stop:string,match:string}[]
+function M.highlight_groups(patterns)
```
justinmk, line 9: should this live in `vim.hl` ?

```suggestion
function M.hl_groups(patterns)
```

runtime/lua/vim/render.lua
```diff
@@ -0,0 +1,162 @@
...
+  vim.fn.setloclist(0, {}, 'a', { title = 'Help TOC' })
+  vim.cmd.lopen()
+end
+
+--- Jump to section
+--- @param opts table jump options
+---  - count integer direction to jump (>0 forward, <0 backward)
+---  - level integer only consider headings up to level
+--- todo(clason): support count
+function M.jump(opts)
```
justinmk, line 133: doesn't seem like `vim.render` or `vim.ui` is the place for this.

justinmk (commented):
unclear if `vim.render` is intended to be public, but the naming questions apply in any case

clason:
It's not meant to be public, and these functions should stay together since they are used in the same context (and for the same reason) but they should not be part of a `ftplugin` (since they're shared across three or more). 

If you have a better name for this internal module, please tell me and I will use it.

runtime/lua/vim/render.lua
clason: "Highlight" is the verb here :) So it should be `highlight_hl_groups`.

runtime/lua/vim/render.lua
clason: Again, internal utility function. Would love to have a generic `jump` or `motion` module, but that is a job for somebody else.

justinmk:
Don't we usually give some signal that a module is private? Seems like the pattern is to name it with underscore. If these can't live in `vim.ui` then perhaps `vim._ui`, I don't want "render" to gain traction/mindshare.

`vim.ui.show_toc` could make sense as a public function but can revisit that later.

clason:
Sure, I'll rename it to `_textutils.lua`. I don't want this mixed into _any_ public-facing API; it's really just a set of common ftplugin utility functions.

(I had the idea of integrating this with our current markdown renderer we use for LSP, and moving vimdoc rendering into Lua, but I see that this is not an acceptable direction so I will drop this idea.)

justinmk:
> Sure, I'll rename it to `_textutils.lua`.


We have `vim.text`, if this is truly text-related stuff that should be private, it can live in `vim._text` or underscore-prefixed functions in `vim.text`.

Nvim is a TUI so everything is "text", but the "text" aspects of these functions is secondary.

These functions are about highlighting, jump (navigation), and UI (show_toc). So `vim.ui` or `vim._ui` seems more appropriate.

clason:
> We have vim.text, if this is truly text-related stuff that should be private, it can live in vim._text or underscore-prefixed functions in vim.text.

I thought about that, but those are text (string) _manipulation_ functions, so this seemed like a poor fit.

Is `vim.ui._text` acceptable to get this merged? Since it's not public API, you can rename and refactor to your heart's content afterwards.

clason:
Or you can also close this if you don't think the feature is worth polluting the namespace.

runtime/ftplugin/help.lua
```diff
@@ -4,37 +4,45 @@ vim.treesitter.start()
...
+vim.keymap.set('n', '[[', function()
+  require('_textutils').jump({ count = -1 })
+end, { buffer = 0, silent = false, desc = 'Jump to previous section' })
 
 -- Add "runnables" for Lua/Vimscript code examples.
 ---@type table<integer, { lang: string, code: string }>
 local code_blocks = {}
-local tree = vim.treesitter.get_parser():parse()[1]
+local bufnr = vim.api.nvim_get_current_buf()
+local parser = assert(vim.treesitter.get_parser(bufnr, 'vimdoc', { error = false }))
```
MariaSolOs, line 45:
```suggestion
local parser = assert(vim.treesitter.get_parser(nil, 'vimdoc', { error = false }))
```

The default is the current buffer.

runtime/lua/_textutils.lua
```diff
@@ -0,0 +1,164 @@
...
+local function hash_tick(bufnr)
+  return tostring(vim.b[bufnr].changedtick)
+end
+
+--- Extract headings from buffer
+--- @param bufnr integer buffer to extract headings from
+--- @return table[]? headings each entry has keys `bufnr`, `lnum`, `capture`, `text`
+local get_headings = vim.func._memoize(hash_tick, function(bufnr)
+  local lang = vim.treesitter.language.get_lang(vim.bo[bufnr].filetype)
+  if not lang then
```
MariaSolOs, line 77: In which case would `lang` be nil?

runtime/lua/_textutils.lua
```diff
@@ -0,0 +1,164 @@
...
+    for _, heading in ipairs(headings) do
+      if heading.lnum > curpos and heading.level <= maxlevel then
+        vim.api.nvim_win_set_cursor(winid, { heading.lnum, 0 })
+        return
+      end
+    end
+  elseif opts.count < 0 then
+    for i = #headings, 1, -1 do
+      if headings[i].lnum < curpos and headings[i].level <= maxlevel then
+        print(headings[i].lnum)
```
MariaSolOs, line 156: Is this `print` left from debugging?

clason: Yes, thanks.

runtime/lua/_textutils.lua
clason: If there's no `filetype`.

justinmk:
> they're just internal utility functions shared by two or more ftplugins. We have yet to come up with a pattern for such. Would it be better if I moved them out of the `vim` namespace altogether?)

Yeah, either `_shared.lua` or `_util.lua` I think conveys the fact that these are just random shared internal functions.

runtime/lua/_textutils.lua
MariaSolOs: Right, I got that bit. But is that possible in a Markdown or Vimdoc file?

I ask this in case that this might silently ignore an error, and maybe `assert` would be more appropriate.

clason: Well, this is primarily intended to shut up luals, and short-circuiting in this case seemed the most reasonable approach.

But maybe returning an empty table is even more reasonable given the purpose here and simplifies the check later.

lewis6991 (approved):

lewis6991 (approved):

justinmk:
> either `_shared.lua` or `_util.lua` I think conveys the fact that these are just random shared internal functions.

👀

clason:
> > either `_shared.lua` or `_util.lua` I think conveys the fact that these are just random shared internal functions.
> 
> 👀

I don't want to give this a too generic name, because I don't want a single dumping ground for all possible shared files. (We have pushed back on this in other contexts.) There's also performance implications since this module is required in ftplugins.

If the current `_textutils.lua` is unacceptable, I'd appreciate a better reasoned suggestion. If it's about top-level files, I can do `shared/_text.lua` or similar.

clason:
OK, new plan (and final offer): 
1. inline the vimdoc specific colorizer function in the help `ftplugin`;
2. move the treesitter navigation functions to a private(!) `vim.treesitter._headings` module.

Follow-up can turn this into a proper module with `headings.scm` queries and an API so other filetypes (asciidoc, rst, LaTeX, ....) can partake.

PR #32303, opened 2025-02-02
vim-patch:9.1.1070: Cannot control cursor positioning of getchar()

zeertzjq:
#### vim-patch:9.1.1070: Cannot control cursor positioning of getchar()

Problem:  Cannot control cursor positioning of getchar().
Solution: Add "cursor" flag to {opts}, with possible values "hide",
          "keep" and "msg".

related: vim/vim#10603
closes: vim/vim#16569

https://github.com/vim/vim/commit/edf0f7db28f87611368e158210e58ed30f673098

### DIFF
--- a/test/functional/vimscript/getchar_spec.lua
+++ b/test/functional/vimscript/getchar_spec.lua
@@ -1,0 +1,92 @@
+local n = require('test.functional.testnvim')()
+local Screen = require('test.functional.ui.screen')
+
+local clear = n.clear
+local exec = n.exec
+local feed = n.feed
+local async_command = n.async_meths.nvim_command
+
+describe('getchar()', function()
+  before_each(clear)
+
+  -- oldtest: Test_getchar_cursor_position()
+  it('cursor positioning', function()
+    local screen = Screen.new(40, 6)
+    exec([[
+      call setline(1, ['foobar', 'foobar', 'foobar'])
+      call cursor(3, 6)
+    ]])
+    screen:expect([[
+      foobar                                  |*2
+      fooba^r                                  |
+      {1:~                                       }|*2
+                                              |
+    ]])
+
+    -- Default: behaves like "msg" when immediately after printing a message,
+    -- even if :sleep moved cursor elsewhere.
+    for _, cmd in ipairs({
+      'echo 1234 | call getchar()',
+      'echo 1234 | call getchar(-1, {})',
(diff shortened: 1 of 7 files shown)


test/functional/vimscript/getchar_spec.lua
```diff
@@ -0,0 +1,92 @@
...
+        1234                                    |
+      ]])
+    end
+
+    -- Default: behaves like "keep" when not immediately after printing a message.
+    for _, cmd in ipairs({
+      'call getchar()',
+      'call getchar(-1, {})',
+      "call getchar(-1, #{cursor: 'keep'})",
+      "echo 1234 | sleep 10m | call getchar(-1, #{cursor: 'keep'})",
```
justinmk, line 56: Testing e32f2b4606f0 locally, it seems `sleep` is necessary to get the desired cursor "keep" effect, even when `getchar(-1, #{cursor: 'keep'})` is called from a mapping.

Would it make sense for `cursor='keep'` to take care of that internally? Else we should probably document it.

justinmk (approved):

test/functional/vimscript/getchar_spec.lua
zeertzjq: What exactly did you test?

justinmk: I tested this, and it seems to be working now...

    nnoremap s <cmd>echo getchar(-1, #{cursor: 'keep'})<cr>

zeertzjq: If the cursor is already in the message area, `getchar()` will not move it back to the current window. I'm thinking about adding another value for `cursor` that forces `getchar()` to move cursor to the current window, but can't decide how to name it.

Or maybe allow `cursor` to also take a List that can specify an arbitrary position?

justinmk: > thinking about adding another value for `cursor` that forces `getchar()` to move cursor to the current window

`win` or maybe `clear` ?

zeertzjq: Actually, I previously thought that this should be achieved by adding a separate functionality solely focusing on moving cursor, and when combined with `getchar(-1, {'cursor': 'keep'})` it can achieve that. There is `cursor` flag of `nvim__redraw` that can already achieve this, but in Vim this can only be done with `:sleep`.

cc @luukvbaal what do you thin?

PR #32310, opened 2025-02-03
fix(tui): include ST to get Cs (set_cursor_color) capability augmented

### LINKED ISSUE #32217: 'guicursor' color change not working in Simple Terminal (suckless ST)
## Issue Description
I have NeoVim configured to change the cursor's color depending on the mode. See exert from my configuration in Lua:
```
vim.cmd.highlight({"iCursor", "guibg=#a9b665"}) -- gruvbox green
vim.cmd.highlight({"rCursor", "guibg=#e78a4e"}) -- gruvbox orange
vim.cmd.highlight({"vCursor", "guibg=#ea6962"}) -- gruvbox red
vim.cmd.highlight({"cCursor", "guibg=#7daea3"}) -- gruvbox blue
vim.opt.guicursor = "a:block-Cursor,i:block-iCursor,v:block-vCursor,c-ci-cr:block-cCursor,r:block-rCursor"
```
This works as expected in WezTerm, as does the following simplified test. 
However in ST,  the cursor color never changes from the terminal set default #f8f8f2.

## Steps to reproduce
`nvim --clean`
`:set termguicolors`
`:highlight Cursor guibg=black`
`:highlight iCursor guibg=green`
`:set guicursor=n-v-c:block-Cursor,i:block-iCursor`

Enter *Insert Mode*
Exit *Insert Mode*
### Expected behavior
Cursor is black in *Normal Mode*, turns green upon entering *Insert Mode*, and returns back to black on exiting *Insert Mode*.
### Actual behavior
Cursor remains a #f8f8f2 colored block throughout, as defined by in ST's config.h:
```
/* special colors */
[256] = "#282a36", /* background */
[257] = "#f8f8f2", /* foreground */
```
### Method A 
`TERM=xterm-256color`
Setting $TERM to xterm-256color in ST.
#### Outcome
When preforming the **Steps to reproduce** above, behavior now matches the **Expected behavior**. However, after exiting NeoVim via 
`:q` the cursor remains black and all terminal text is colored black too, which should not happen.

### Method B 
Looking for a relevant ST patch
#### Outcome
Could not find one.

### Method C 
Verify via `printf "\e]12;#1793d0\007"` that ST "actually supports termcodes to change the cursor color" [#29498](https://github.com/neovim/neovim/issues/29498#issuecomment-2322500302)
#### Outcome
Terminal text and cursor turn blue (#1793d0).
EDIT: Previously tested with `printf "\033[31m"` which would only change the terminal text color.

## Thoughts
Since the **Outcome** of **Method C** indicates ST supports termcodes for cursor color change, and **Method A** enables the expected behavior within NeoVim, I hope there may be a way to natively enable NeoVim to change ST's cursor color.

## Terminfo
Terminfo section from:
`nvim -V3log` and then `:q`
[log.txt](https://github.com/user-attachments/files/18552252/log.txt)

### Relevant software info

OS: Arch/ Linux 6.12.10-arch1-1 x86_64
`nvim --version`: NVIM v0.10.3 -- installed via `pacman -S neovim`
[ST](https://st.suckless.org/) version: [0.9.2](https://dl.suckless.org/st/st-0.9.2.tar.gz) 
_(for reference) WezTerm: 20240203-110809-5046fc22_

### $TERM
$TERM=st-256color

#### Vim (not Nvim) behaves the same?

N/A, **Steps to reproduce** with vim 9.1, `vim -u DEFAULTS` do not working in ST *or* WezTerm

uio23:
Problem:
Nvim's 'guicursor' does not change the cursor color in ST because Nvim's builtin terminfo for ST lacks a Cs entry, even though ST does support the cursor color to be set via termcodes.

Solution:
Thomas Dickey has added a to-do item to add the Cs entry for st into ncurses, from which Nvim's builtin terminfos are generated.  In the mean time, this fix includes ST as a separate condition for a branch in `augment_terminfo()` that manually adds the Cs extended unibilium capability.

Test coverage:
Ran `make test`, some variable amount (4-6) tests failed, but this was also the case on the master branch and none related to tui.c

fix #32217

### DIFF
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -2382,7 +2382,7 @@ static void augment_terminfo(TUIData *tui, const char *term, int vte_version, in
     } else if ((xterm || hterm || rxvt || tmux || alacritty || st)
                && (vte_version == 0 || vte_version >= 3900)) {
       // Supported in urxvt, newer VTE.
-      // Supported in st, but currently lacks entry in ncurses definitions  #32217
+      // Supported in st, but currently missing in ncurses definitions. #32217
       tui->unibi_ext.set_cursor_color = (int)unibi_add_ext_str(ut, "ext.set_cursor_color",
                                                                "\033]12;%p1%s\007");
     }
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -2379,8 +2379,8 @@ static void augment_terminfo(TUIData *tui, const char *term, int vte_version, in
       // would use a tmux control sequence and an extra if(screen) test.
       tui->unibi_ext.set_cursor_color =
         (int)unibi_add_ext_str(ut, NULL, TMUX_WRAP(tmux, "\033]Pl%p1%06x\033\\"));
-    } else if (((xterm || hterm || rxvt || tmux || alacritty)
-               && (vte_version == 0 || vte_version >= 3900)) || st) {
+    } else if ((xterm || hterm || rxvt || tmux || alacritty || st)
+               && (vte_version == 0 || vte_version >= 3900)) {
       // Supported in urxvt, newer VTE.
       // Supported in st, but currently lacks entry in ncurses definitions  #32217
       tui->unibi_ext.set_cursor_color = (int)unibi_add_ext_str(ut, "ext.set_cursor_color",
--- a/src/nvim/tui/tui.c
+++ b/src/nvim/tui/tui.c
@@ -2294,6 +2294,7 @@ static void augment_terminfo(TUIData *tui, const char *term, int vte_version, in
   bool putty = terminfo_is_term_family(term, "putty");
   bool screen = terminfo_is_term_family(term, "screen");
   bool tmux = terminfo_is_term_family(term, "tmux") || !!os_getenv("TMUX");
+  bool st = terminfo_is_term_family(term, "st");
   bool iterm = terminfo_is_term_family(term, "iterm")
                || terminfo_is_term_family(term, "iterm2")
                || terminfo_is_term_family(term, "iTerm.app")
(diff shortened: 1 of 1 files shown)


src/nvim/tui/tui.c
```diff
@@ -2378,9 +2379,10 @@ static void augment_terminfo(TUIData *tui, const char *term, int vte_version, in
       // would use a tmux control sequence and an extra if(screen) test.
       tui->unibi_ext.set_cursor_color =
         (int)unibi_add_ext_str(ut, NULL, TMUX_WRAP(tmux, "\033]Pl%p1%06x\033\\"));
-    } else if ((xterm || hterm || rxvt || tmux || alacritty)
-               && (vte_version == 0 || vte_version >= 3900)) {
+    } else if (((xterm || hterm || rxvt || tmux || alacritty)
+               && (vte_version == 0 || vte_version >= 3900)) || st) {
```
justinmk, line 2383:
```suggestion
    } else if ((xterm || hterm || rxvt || tmux || alacritty || st)
               && (vte_version == 0 || vte_version >= 3900)) {
```

src/nvim/tui/tui.c
```diff
@@ -2378,9 +2379,10 @@ static void augment_terminfo(TUIData *tui, const char *term, int vte_version, in
       // would use a tmux control sequence and an extra if(screen) test.
       tui->unibi_ext.set_cursor_color =
         (int)unibi_add_ext_str(ut, NULL, TMUX_WRAP(tmux, "\033]Pl%p1%06x\033\\"));
-    } else if ((xterm || hterm || rxvt || tmux || alacritty)
+    } else if ((xterm || hterm || rxvt || tmux || alacritty || st)
                && (vte_version == 0 || vte_version >= 3900)) {
       // Supported in urxvt, newer VTE.
+      // Supported in st, but currently lacks entry in ncurses definitions  #32217
```
justinmk, line 2385:
```suggestion
      // Supported in st, but currently missing in ncurses definitions. #32217
```

PR #32316, opened 2025-02-03
fix(ci): only run vim-patches workflow on neovim/neovim

### LINKED ISSUE #32313: Run vim-patches workflow only on neovim/neovim
### Problem

People that have forked the Neovim repo get a workflow run failed error every day:

Image

### Steps to reproduce

1. Fork neovim
2. Wait a day

### Expected behavior

I'm not 100% sure, but I think this workflow is only needed on the main repository: neovim/neovim. I would expect the workflow to have a check if the owner of the repo is the neovim org.

### Nvim version (nvim -v)

Currently the forked master branch is at https://github.com/yochem/neovim/commit/92d3bf101d07d52cfbd4c22e6e06251897adda34

yochem:
Problem: The default for forks is to run all workflows, which results in
a "failed workflow run" error every day

Solution: check if the repository owner is the Neovim organisation.

Fixes: #32313.

I'n not 100% sure this works, but the [Github docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution#example-only-run-job-for-specific-repository) make it seems like this would work. Maybe CI-legend @dundargoc can step in :)

### DIFF
--- a/.github/workflows/vim_patches.yml
+++ b/.github/workflows/vim_patches.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   update-vim-patches:
+    if: github.repository_owner == 'neovim'
     runs-on: ubuntu-latest
     permissions:
       contents: write


jamessan:
Workflows aren't enabled by default when you fork a repository. If you choose to enable them, you can also later disable them.

workflow-fork

clason:
I think that was changed at some point, which means if you have a vintage fork you could still have these enabled by default. (But then making this change won't magically fix that; you'd have to rebase, and at that point you could just as well disable manually.)

dundargoc:
Eh, this workflow also fails in the neovim repo, this has nothing to do with whether it's run in or outside neovim repo: https://github.com/neovim/neovim/actions/workflows/vim_patches.yml. I will send a patch now to take care of this.

PR #32326, opened 2025-02-04
feat(treesitter): show which nodes are missing in InspectTree

ribru17:
Now `:InspectTree` will show missing nodes as e.g. `(MISSING identifier)` or `(MISSING ";")` rather than just `(identifier)` or `";"`. This is doable because the `MISSING` keyword is now valid query syntax.

### DIFF
--- a/runtime/lua/vim/treesitter/dev.lua
+++ b/runtime/lua/vim/treesitter/dev.lua
@@ -224,9 +224,12 @@ function TSTreeView:draw(bufnr)
 
     local text ---@type string
     if item.node:named() then
-      text = string.format('(%s', item.node:type())
+      text = string.format('(%s%s', item.node:missing() and 'MISSING ' or '', item.node:type())
     else
       text = string.format('%q', item.node:type()):gsub('\n', 'n')
+      if item.node:missing() then
+        text = string.format('(MISSING %s)', text)
+      end
     end
     if item.field then
       text = string.format('%s: %s', item.field, text)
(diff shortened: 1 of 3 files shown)


runtime/lua/vim/treesitter/dev.lua
```diff
@@ -224,9 +224,13 @@ function TSTreeView:draw(bufnr)
 
     local text ---@type string
     if item.node:named() then
-      text = string.format('(%s', item.node:type())
+      local missing_prefix = item.node:missing() and 'MISSING ' or ''
```
clason, line 227: Maybe shorter to just inline?

justinmk (approved):

runtime/lua/vim/treesitter/dev.lua
```diff
@@ -224,9 +224,12 @@ function TSTreeView:draw(bufnr)
 
     local text ---@type string
     if item.node:named() then
-      text = string.format('(%s', item.node:type())
+      text = string.format('(%s', (item.node:missing() and 'MISSING ' or '') .. item.node:type())
```
clason, line 227:
```suggestion
      text = string.format('(%s%s', item.node:missing() and 'MISSING ' or '', item.node:type())
```
Does this work?

ribru17: Sure :+1:

MariaSolOs:
> This is doable because the `MISSING` keyword is now valid query syntax.

@ribru17 for the sake of Github history (and my own learning), could you please link the change that introduced this new syntax?

ribru17:
Sure, the tree-sitter change is here: https://github.com/tree-sitter/tree-sitter/pull/3887 and the grammar change is here: https://github.com/tree-sitter-grammars/tree-sitter-query/pull/46

PR #32342, opened 2025-02-05
ci(arm64): skip installing npm to prevent segmentation faults

### LINKED ISSUE #32339: ci: ARM jobs are unreliable
### Problem

arm jobs often fail on a `npm` segfault?

```
Install neovim npm package
/home/runner/work/_temp/7ec7ce0e-6780-4fc3-88d6-8077daf0b7e0.sh: line 5:  6604 Segmentation fault      (core dumped) npm install -g neovim
```

### Expected behavior

just skip the `npm install` for arm. it's not that important. green CI is much more important.

dundargoc:
Installing npm on linux arm64 causes intermittent segmentation faults
for unknown reasons.

Closes https://github.com/neovim/neovim/issues/32339.

### DIFF
--- a/.github/scripts/install_deps.sh
+++ b/.github/scripts/install_deps.sh
@@ -9,8 +9,9 @@ while (($# > 0)); do
   esac
 done
 
-os=$(uname -s)
-if [[ $os == Linux ]]; then
+OS=$(uname -s)
+ARCH=$(uname -m)
+if [[ $OS == Linux ]]; then
   sudo apt-get update
   sudo apt-get install -y build-essential cmake curl gettext ninja-build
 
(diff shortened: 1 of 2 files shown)


dundargoc:
Future TODO: that `CC=cc` is some jank BS, maybe remove and test if issue persist

``bash
CC=cc python3 -m pip -q install --user --upgrade --break-system-packages pynvim
```

zeertzjq:
They also sometimes fail earlier at the checkout step

dundargoc:
Not much to do about that except skipping arm ubuntu completely. Is it bad enough to skip it compeltely?

zeertzjq:
run the checkout step multiple times?

HoangLe734:
do you know how to use the config because when I open nvim after clone the repository, it kepts telling error

dundargoc:
Tbh I've only noticed segfaults from npm, it seems like the segfaults from checkout is much more rare. Maybe we can get away with just skipping npm for the time being and then reevaluate?

PR #32353, opened 2025-02-06
build: add luals check

dundargoc:
This automatically downloads and uses the correct luals binary for the
currently used system. `make luals` will run luals on all lua files in
`runtime`.

We download lua-language-server manually instead of relying on
contributors downloading it on their own (like with stylua) as
lua-language-server is updated frequently which may cause unnecessary
friction. Therefore, we download a pinned version of luals which we then
can manually bump when needed. This can be re-evaluated if luals becomes
more stable in the future.

Currently this is not run when using `make lint` since cmake style "file
caching" doesn't seem possible at the moment. This is because checking a
single file doesn't seem to work.

Work on https://github.com/neovim/neovim/issues/24563.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,10 @@ jobs:
         run: cmake --build build --target lintlua-stylua
 
       - if: success() || failure() && steps.abort_job.outputs.status == 'success'
+        name: luals
+        run: cmake --build build --target luals
+
+      - if: success() || failure() && steps.abort_job.outputs.status == 'success'
         name: luacheck
         run: cmake --build build --target lintlua-luacheck
 
--- a/runtime/autoload/ccomplete.lua
+++ b/runtime/autoload/ccomplete.lua
@@ -5,7 +5,8 @@
 
 -- Ignore "value assigned to a local variable is unused" because
 --  we can't guarantee that local variables will be used by plugins
--- luacheck: ignore 311
+-- luacheck: ignore
+--- @diagnostic disable
 
 local vim9 = require('_vim9script')
 local M = {}
--- a/runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
+++ b/runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
@@ -5,7 +5,8 @@
 
 -- Ignore "value assigned to a local variable is unused" because
 --  we can't guarantee that local variables will be used by plugins
--- luacheck: ignore 311
+-- luacheck: ignore
+--- @diagnostic disable
 
 local vim9 = require('_vim9script')
 local M = {}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,3 +321,33 @@ if(USE_BUNDLED_BUSTED)
 else()
   add_custom_target(lua_dev_deps)
 endif()
+
+if (CMAKE_SYSTEM_PROCESSOR MATCHES arm64)
+  set(LUALS_ARCH arm64)
+else()
+  set(LUALS_ARCH x64)
+endif()
+
+set(LUALS_VERSION 3.13.6)
+set(LUALS "lua-language-server-${LUALS_VERSION}-${CMAKE_SYSTEM_NAME}-${LUALS_ARCH}")
+set(LUALS_TARBALL ${LUALS}.tar.gz)
+set(LUALS_URL https://github.com/LuaLS/lua-language-server/releases/download/${LUALS_VERSION}/${LUALS_TARBALL})
+
+ExternalProject_Add(download_luals
+  URL ${LUALS_URL}
+  DOWNLOAD_DIR ${DEPS_DOWNLOAD_DIR}/luals
+  SOURCE_DIR ${DEPS_BIN_DIR}/luals
+  CONFIGURE_COMMAND ""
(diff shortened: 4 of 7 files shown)


dundargoc:
Problems:
- Me and clason get completely different number of errors when we run `make luals`, he gets around 550 and I get around 450.
- Clason gets warnings from runtime/lua/vim/shared.lua but I don't.
- The number of errors is not deterministic. It gives out different number of problems for the same command. Both clason and I can recreate this problem.

Conclusion: S-rank special grade cursed tool.

clason:
Difference is from a flaky false positive on `lua/vim/shared.lua`, which seems to be a race condition in inference. "Fixed" by forcing the type via annotation.

Similar issue in `lua/coxpcall.lua`; here an annotation didn't help and I just disabled diagnostics on the affected line. (It's not our code anyway.)

clason:
I would not spend time fixing up the generator scripts in this PR; that's way too much work. Better to restrict the lint step to `runtime/lua` for now.

dundargoc:
Let's go with all lua files in `runtime` at least since all lua files there pass the check now after vim9jit updates.

.github/workflows/test.yml
```diff
@@ -77,6 +77,10 @@ jobs:
         name: uncrustify
         run: cmake --build build --target lintc-uncrustify
 
+      - if: success() || failure() && steps.abort_job.outputs.status == 'success'
```
clason, line 80: Maybe move that up before luacheck, to keep the checks organized by language?

runtime/autoload/ccomplete.lua
```diff
@@ -26,7 +27,7 @@ local SearchMembers = nil
 -- # Vim completion script
 -- # Language:	C
 -- # Maintainer:	The Vim Project <https://github.com/vim/vim>
--- # Last Change:	2023 Aug 10
+-- # Last Change:	2024 Jun 06
```
clason, line 30: 👀

runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
```diff
@@ -90,7 +91,7 @@ Qf_filter = function(qf, searchpat, bang)
   end
 
   items = vim9.fn_mut('filter', { Xgetlist(), Cond }, { replace = 0 })
-  title = cmd .. ' /' .. pat .. '/'
+  title = string.format([=[{cmd} /{pat}/]=])
```
zeertzjq, line 94: ?

runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
```diff
@@ -48,7 +49,7 @@ Qf_filter = function(qf, searchpat, bang)
     Xsetlist = function(...)
       return vim.fn['setqflist'](...)
     end
-    cmd = ':Cfilter' .. bang
+    cmd = string.format([=[:Cfilter{bang}]=])
```
clason, line 52: Since these files are generated, I'd prefer not to make any code changes here. 

Simply ignore the whole file.

runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
dundargoc: That is what vim9jit generated.

clason: Did you regenerate the files in this PR, or why was this changed?

runtime/autoload/ccomplete.lua
dundargoc: I think it just uses the the same date that's in the vim file used to generate it: https://github.com/dundargoc/vim/blob/76bdb82527a13b5b2baa8f7d7ce14b4d5dc05b82/runtime/autoload/ccomplete.vim#L6

clason: Yes, but why was this changed _in this PR_?

runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
zeertzjq: Ah, vim9jit is broken then

runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
dundargoc: OK, I'll just keep the disabling parts.

clason (approved):

justinmk (approved):

lewis6991 (approved):

runtime/pack/dist/opt/cfilter/plugin/cfilter.lua
dundargoc: > Did you regenerate the files in this PR, or why was this changed?

I did at first, but I reverted it (except I kept the lines that disabled lints)

clason:
Follow-up: track version in `deps.txt` and allow bumping with `bump_deps.lua`.

dundargoc:
> Follow-up: track version in `deps.txt` and allow bumping with `bump_deps.lua`.

Not possible unfortunately. The URL is decided programatically depending on the current platform and can't be hardcoded. At most we can list every URL for each platform but then we'd need to update all of them at the same time which isn't worth it.

PR #32376, opened 2025-02-09
test(fix): reset of cmdline abort state only when the expectation has  finished

fredizzimo:
Without this, it's reset when intermediate unexpected states are received.

Pulled out from https://github.com/neovim/neovim/pull/31837 for easier review.

### DIFF
--- a/test/functional/ui/screen.lua
+++ b/test/functional/ui/screen.lua
@@ -655,12 +655,6 @@ screen:redraw_debug() to show all intermediate screen states.]]
       end
     end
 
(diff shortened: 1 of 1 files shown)


dundargoc:
🤓 type nitpick: the commit type should be `test:` or `test(fix):` as the `fix:` type is for user bugs, and if we use `fix(test):` this change will show up in the "Bug fixes" section of the changelog.

fredizzimo:
I fixed the commits and descriptions of all three of my recent pull requests

PR #32378, opened 2025-02-09
fix(defaults): improve visual search mappings

neeshy:
Problem:  The behavior of the visual search mappings aren't consistent with their normal mode counterparts.
  - The count isn't considered
  - Searching with an empty selection will match every character in the buffer
  - Searching backwards only jumps back when the cursor is positioned at the start of the selection.

Solution:
  - Issue `n` `v:count1` times
  - Error out and exit visual mode when the selection is empty
  - Detect when the cursor is at the start of the selection, and adjust the count accordingly

Also, use the search register instead of the more error-prone approach of feeding the entire search string as an expression

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -32,27 +32,53 @@ do
   ---
   --- See |v_star-default| and |v_#-default|
   do
-    local function _visual_search(cmd)
-      assert(cmd == '/' or cmd == '?')
-      local chunks =
-        vim.fn.getregion(vim.fn.getpos('.'), vim.fn.getpos('v'), { type = vim.fn.mode() })
+    local function _visual_search(forward)
+      assert(forward == 0 or forward == 1)
+      local pos = vim.fn.getpos('.')
+      local vpos = vim.fn.getpos('v')
+      local mode = vim.fn.mode()
+      local chunks = vim.fn.getregion(pos, vpos, { type = mode })
       local esc_chunks = vim
         .iter(chunks)
         :map(function(v)
-          return vim.fn.escape(v, cmd == '/' and [[/\]] or [[?\]])
+          return vim.fn.escape(v, [[\]])
         end)
         :totable()
       local esc_pat = table.concat(esc_chunks, [[\n]])
-      local search_cmd = ([[%s\V%s%s]]):format(cmd, esc_pat, '\n')
-      return '\27' .. search_cmd
+      if #esc_pat == 0 then
+        vim.api.nvim_echo({ { 'E348: No string under cursor' } }, true, { err = true })
+        return '<Esc>'
+      end
+      local search = [[\V]] .. esc_pat
+
+      vim.fn.setreg('/', search)
+      vim.fn.histadd('/', search)
+      vim.v.searchforward = forward
+
+      -- The count has to be adjusted when searching backwards and the cursor
+      -- isn't positioned at the beginning of the selection
+      local count = vim.v.count1
+      if forward == 0 then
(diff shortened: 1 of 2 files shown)


runtime/lua/vim/_defaults.lua
```diff
@@ -32,27 +32,59 @@ do
...
+            return line == vline and col > vcol
+          elseif mode == 'V' then
+            return col ~= 1
+          elseif mode == '\22' then
+            return col > vcol
+          end
+        end)()
+      then
+        count = count + 1
+      end
```
zeertzjq, line 78: Maybe it's better to use `o` and `O` to move cursor to the start of the selection?

And to test for this you can remove the `o` in the tests.

neeshy: That sounds like it would involve more logic, especially with <C-V> (since it would have to use different combinations of `o` and `O` for each of the three corners which aren't the top-left). Incrementing the count means all those cases can be collapsed into one.

runtime/lua/vim/_defaults.lua
```diff
@@ -32,27 +32,59 @@ do
...
+      return '<Esc>' .. count .. 'n'
     end
 
     vim.keymap.set('x', '*', function()
```
asmodeus812, line 82: Not sure if this is intended, but the usual * and # ones work on a word boundaries, adding < and > to the pattern, and the g* and g# variations do not, do we need to support visual g* and g# ?

neeshy: I think it makes sense to not use word boundaries since visual `*`/`#` is much more deliberate (you see the boundaries of the selection before you search) than normal mode `*`/`#`.

justinmk (approved):
LGTM, though it's unfortunate that `?` requires all this extra code. Is it worth adding another test?

neeshy:
I figured removing the `o` from https://github.com/neovim/neovim/blob/ac207c3ac20026baa2a4561f5365c2af6a38a357/test/functional/editor/defaults_spec.lua#L142-L151 would make for an adequate test, but that fails: 

```
$ TEST_FILTER='handle various chars properly' make functionaltest
...
FAILED   test/functional/editor/defaults_spec.lua @ 98: default key mappings Visual mode search mappings handle various chars properly
test/functional/editor/defaults_spec.lua:123: Row 1 did not match.
Expected:
  |*{3:testing <CR> /?\!}1                                          |
  |*{4:^testing <CR> /?\!}2                                          |
  |*{3:testing <CR> /?\!}3                                          |
  |*{3:testing <CR> /?\!}4                                          |
  |{1:~                                                           }|
  |{1:~                                                           }|
  |{2:[No Name] [+]                             2,1            All}|
  |*/\Vtesting <CR> \/?\\!                    [2/4]             |
Actual:
  |*{3:t}es{3:t}ing <CR> /?\!1                                          |
  |*{4:^t}es{3:t}ing <CR> /?\!2                                          |
  |*{3:t}es{3:t}ing <CR> /?\!3                                          |
  |*{3:t}es{3:t}ing <CR> /?\!4                                          |
  |{1:~                                                           }|
  |{1:~                                                           }|
  |{2:[No Name] [+]                             2,1            All}|
  |*/\Vt                                      [3/8]             |
```

Which is bizarre; it only searches for the `t`. The same failure can be seen on master if you remove the `o`'s as well. So maybe it's an issue with the test framework?

@zeertzjq do you know what could be causing this?

justinmk:
You could try:

```
n.feed('G0vf!')
n.poke_eventloop()
n.feed('#')
```

neeshy:
OK, that works. Pushed.

justinmk (approved):

PR #32382, opened 2025-02-09
fix(lsp): Show correct activeParameter on vim.lsp.buf.signature_help

### LINKED ISSUE #32381: LSP: `signature_help` shows incorrect active parameter
### Problem

It seems like `vim.lsp.buf.signature_help` always highlights the first parameter regardless of the current cursor position.

Tested on nightly: af0cd9a9075117319808b8c0515cd86628992baf

### Example

```go
// main.go
package main

import (
	"context"
	"fmt"
)

func foo(
	_ context.Context,
	message string,
	id int,
) string {
	return fmt.Sprintf("ID: %d, Message: %s", id, message)
}

func main() {
	ctx := context.Background()
	msg := "Hello, World!"

	res := foo(ctx, msg, 1)

	fmt.Println(res)
}
```

### Steps

* Using the default `C-S` binding in insert mode.
* The first argument is highlighted regardless of the cursor position.
* Tested with `gopls` and `vtsls`.

I used a minimal `nvim` setup using the `NVIM_APPNAME`

```sh
NVIM_APPNAME=tmpvim nvim main.go
```

### Configuration

```lua
-- ~/.config/tmpvim/init.lua
vim.lsp.config["gopls"] = {
	cmd = { "gopls", "-logfile=/tmp/gopls.log", "serve", "--debug=localhost:6060" },
	filetypes = { "go", "gomod" },
}

vim.lsp.enable("gopls")
```

### Video

[asciicast](https://asciinema.org/a/702395)

### Nvim version (nvim -v)

af0cd9a9075117319808b8c0515cd86628992baf

gopls

### Operating system/version

macOS 15

### Log file

_No response_

shantanuraj:
- On some lsps the `textDocument/signatureHelp` response, only includes the `activeParameter` field on the `lsp.SignatureHelp` type.
```lua
{
  {
    result = {
      activeParameter = 2,
      signatures = {
        {
          documentation = {
            kind = "markdown",
            value = ""
          },
          label = "getBuyers(ctx context.Context, orderDB boil.ContextExecutor, supplierID string) ([]*BuyerWithLocation, error)",
          parameters = {
            {
              label = "ctx context.Context"
            },
            {
              label = "orderDB boil.ContextExecutor"
            },
            {
              label = "supplierID string"
            }
          }
        }
      }
    }
  }
}
```

- This change ensures we retain this information before showing the signature information

### Video

* Using the reproduction steps from the linked issue.
[asciicast](https://asciinema.org/a/702394)

### Context

Closes #32381

### DIFF
--- a/runtime/lua/vim/lsp/buf.lua
+++ b/runtime/lua/vim/lsp/buf.lua
@@ -315,6 +315,7 @@ local function process_signature_help_results(results)
       local result = r.result --- @type lsp.SignatureHelp
       if result and result.signatures and result.signatures[1] then
         for _, sig in ipairs(result.signatures) do
+          sig.activeParameter = sig.activeParameter or result.activeParameter
           signatures[#signatures + 1] = { client, sig }
         end
       end


justinmk:
Thanks! This was really bugging me.

justinmk:
Merging this because it's urgent. 

But if you can, adding a test will help a lot in the future. It looks like we have no test coverage for `signature_help` currently. `vim.lsp.buf.signature_help()` is a public function, so it's pretty easy to directly test it. Based on the existing tests, this would imply a new `describe()` block after this line: https://github.com/neovim/neovim/blob/8c2571991ad748b0ee71529951e1502e553aa8ba/test/functional/plugin/lsp_spec.lua#L5228

  describe('lsp.buf.signature_help', function()

You can [run the test](https://github.com/neovim/neovim/blob/master/test/README.md#filter-by-file) like this:

    TEST_FILE=test/functional/plugin/lsp_spec.lua TEST_FILTER=signature_help TEST_TAG= make functionaltest

shantanuraj:
Sure, I'll open another PR to add a test case.

PR #32387, opened 2025-02-09
feat(lsp): add select kind in showMessageRequest

MariaSolOs:
^^

Motivation: I use `dressing.nvim`, which allows one to use [different `select` implementations](https://github.com/stevearc/dressing.nvim?tab=readme-ov-file#advanced-configuration). I would like to customize the picker when coming from this request, but it's a bit tricky because it doesn't include the `kind` (and filtering on the `prompt` is a bit flaky IMO).

### DIFF
--- a/runtime/lua/vim/lsp/handlers.lua
+++ b/runtime/lua/vim/lsp/handlers.lua
@@ -90,6 +90,7 @@ RSC[ms.window_showMessageRequest] = function(_, params)
   local co, is_main = coroutine.running()
   if co and not is_main then
     local opts = {
+      kind = 'lsp_message',
       prompt = params.message .. ': ',
       format_item = function(action)
         return (action.title:gsub('\r\n', '\\r\\n')):gsub('\n', '\\n')


mfussenegger (approved):
lgtm but sounds more like `feat:`?

runtime/lua/vim/lsp/handlers.lua
```diff
@@ -90,6 +90,7 @@ RSC[ms.window_showMessageRequest] = function(_, params)
   local co, is_main = coroutine.running()
   if co and not is_main then
     local opts = {
+      kind = 'lsp_message',
```
justinmk, line 93: Not sure what the existing precedent is, should we use dot instead? 

```suggestion
      kind = 'lsp.message',
```

MariaSolOs: We don't have a precedent, other `kind`s simply don't use a word delimiter (e.g. `codeaction`). Maybe we can set this to `lspmessage`?

justinmk: Not sure. Not a blocker.

MariaSolOs: In that case are we okay with merging this? :)

PR #32391, opened 2025-02-10
refactor(defaults): remove need of ctx from enable_ctx_menu method

siddhantdev:
There seems to be no need to call the method with the ctx parameter when it can be calculated in the method itself.

Relevant discussion: https://github.com/neovim/neovim/pull/32122#discussion_r1948301935

### DIFF
--- a/runtime/lua/vim/_defaults.lua
+++ b/runtime/lua/vim/_defaults.lua
@@ -428,7 +428,7 @@ do
     anoremenu PopUp.How-to\ disable\ mouse  <Cmd>help disable-mouse<CR>
   ]])
 
-  local function enable_ctx_menu(ctx)
+  local function enable_ctx_menu()
     vim.cmd([[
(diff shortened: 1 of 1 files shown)


justinmk (approved):

runtime/lua/vim/_defaults.lua
```diff
@@ -437,9 +437,13 @@ do
...
 
-    if ctx == 'url' then
+    local urls = require('vim.ui')._get_urls()
+    local url = vim.startswith(urls[1], 'http')
+    if url then
       vim.cmd([[amenu enable PopUp.Open\ in\ web\ browser]])
-    elseif ctx == 'lsp' then
+    end
+
+    if vim.lsp.get_clients({ bufnr = 0 })[1] then
```
justinmk, line 446:
```suggestion
    elseif vim.lsp.get_clients({ bufnr = 0 })[1] then
```

PR #32393, opened 2025-02-10
fix(lsp): clear word when expand multi-lines word

glepnir:
Problem: When expanding a completion item that contains a multi-line word, the word is not deleted correctly.

Solution: If the word contains a line break, delete the text from Context.cursor to the current cursor position.

### DIFF
--- a/runtime/lua/vim/lsp/completion.lua
+++ b/runtime/lua/vim/lsp/completion.lua
@@ -493,6 +493,7 @@ local function trigger(bufnr, clients)
       end
     end
     local start_col = (server_start_boundary or word_boundary) + 1
+    Context.cursor = { cursor_row, start_col }
     vim.fn.complete(start_col, matches)
   end)
 
@@ -569,8 +570,14 @@ local function on_complete_done()
     end
 
     -- Remove the already inserted word.
-    local start_char = cursor_col - #completed_item.word
-    api.nvim_buf_set_text(bufnr, cursor_row, start_char, cursor_row, cursor_col, { '' })
+    api.nvim_buf_set_text(
+      bufnr,
+      Context.cursor[1] - 1,
+      Context.cursor[2] - 1,
+      cursor_row,
+      cursor_col,
+      { '' }
+    )
   end
 
   local function apply_snippet_and_command()
--- a/test/functional/plugin/lsp/completion_spec.lua
+++ b/test/functional/plugin/lsp/completion_spec.lua
@@ -1078,4 +1078,42 @@ describe('vim.lsp.completion: integration', function()
       end)
     )
   end)
+
+  it('#clear multiple-lines word', function()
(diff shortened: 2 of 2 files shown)


runtime/lua/vim/lsp/completion.lua
```diff
@@ -493,6 +493,7 @@ local function trigger(bufnr, clients)
       end
     end
     local start_col = (server_start_boundary or word_boundary) + 1
+    Context.cursor = { api.nvim_win_get_cursor(0)[1], start_col }
```
glepnir, line 496: @mfussenegger  `Context.cursor` is used to record the starting position of the completion, right? I don't see this field being assigned and used. So I use it here

runtime/lua/vim/lsp/completion.lua
```diff
@@ -569,8 +570,19 @@ local function on_complete_done()
...
+        Context.cursor[1] - 1,
+        Context.cursor[2] - 1,
+        cursor_row,
+        cursor_col,
+        { '' }
+      )
+    else
+      local start_char = cursor_col - #completed_item.word
+      api.nvim_buf_set_text(bufnr, cursor_row, start_char, cursor_row, cursor_col, { '' })
+    end
```
mfussenegger, line 585: Wouldn't the new logic work for both cases? E.g.:

```suggestion
    api.nvim_buf_set_text(
      bufnr,
      Context.cursor[1] - 1,
      Context.cursor[2] - 1,
      cursor_row,
      cursor_col,
      { '' }
    )
```

runtime/lua/vim/lsp/completion.lua
mfussenegger: I think `Context.cursor` is a leftover from the initial PR which had functionality to re-trigger completion on backspace.

That was derived from lsp-compl and set whenever the cursor changed (https://github.com/mfussenegger/nvim-lsp-compl/blob/e51383b609f56b45fc47d3477127805b713d0e2a/lua/lsp_compl.lua#L509-L526)

Seems okay to use it like this for now, but could re-use the local variables from above:

```suggestion
    Context.cursor = { cursor_row, cursor_col }
```

glepnir: cursor_col will not work it should start_col

mfussenegger:
Looks like this caused some test flakyness:

```
test/functional/plugin/lsp/completion_spec.lua:1109: Expected objects to be the same.
Passed in:
(number) 0
Expected:
(number) 13

stack traceback:
	test/functional/plugin/lsp/completion_spec.lua:1109: in function <test/functional/plugin/lsp/completion_spec.lua:1074>
```

Could you have a look?

glepnir:
how you trigger this failed

justinmk:
@glepnir it's a random CI failure.

